PR #18459, opened 2022-12-15
Remove a redundant check on the signed_distance < influence_distance in MinimumDistanceConstraint.

hongkai-dai:
The original code is correct, but the check on `signed_distance_pair.distance < influence_distance` is redundant because `signed_distance_pairs` is computed with `max_distance = influence_distance`. https://github.com/RobotLocomotion/drake/blob/4443764321ac2dcad19fc796b5ce685916b045c1/multibody/inverse_kinematics/minimum_distance_constraint.cc#L33-L35

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459)

This was motivated by the stack overflow question https://stackoverflow.com/questions/74793241/multiple-minimumdistanceconstraint-constraints

### DIFF
--- a/multibody/inverse_kinematics/minimum_distance_constraint.cc
+++ b/multibody/inverse_kinematics/minimum_distance_constraint.cc
@@ -34,32 +34,26 @@ VectorX<S> Distances(const MultibodyPlant<T>& plant,
       query_object.ComputeSignedDistancePairwiseClosestPoints(
           influence_distance);
   VectorX<S> distances(signed_distance_pairs.size());
-  int distance_count{0};
-  for (const auto& signed_distance_pair : signed_distance_pairs) {
-    if (signed_distance_pair.distance < influence_distance) {
-      const geometry::SceneGraphInspector<T>& inspector =
-          query_object.inspector();
-      const geometry::FrameId frame_A_id =
-          inspector.GetFrameId(signed_distance_pair.id_A);
-      const geometry::FrameId frame_B_id =
-          inspector.GetFrameId(signed_distance_pair.id_B);
-      const Frame<T>& frameA =
-          plant.GetBodyFromFrameId(frame_A_id)->body_frame();
-      const Frame<T>& frameB =
-          plant.GetBodyFromFrameId(frame_B_id)->body_frame();
-      internal::CalcDistanceDerivatives(
-          plant, *context, frameA, frameB,
-          // GetPoseInFrame() returns RigidTransform<double> -- we can't
-          // multiply across heterogeneous scalar types; so we cast the double
-          // to T.
-          inspector.GetPoseInFrame(signed_distance_pair.id_A)
-                  .template cast<T>() *
-              signed_distance_pair.p_ACa,
-          signed_distance_pair.distance, signed_distance_pair.nhat_BA_W, q,
-          &distances(distance_count++));
-    }
+  for (int i = 0; i < static_cast<int>(signed_distance_pairs.size()); ++i) {
+    const geometry::SceneGraphInspector<T>& inspector =
+        query_object.inspector();
+    const geometry::FrameId frame_A_id =
+        inspector.GetFrameId(signed_distance_pairs[i].id_A);
+    const geometry::FrameId frame_B_id =
+        inspector.GetFrameId(signed_distance_pairs[i].id_B);
+    const Frame<T>& frameA = plant.GetBodyFromFrameId(frame_A_id)->body_frame();
+    const Frame<T>& frameB = plant.GetBodyFromFrameId(frame_B_id)->body_frame();
+    internal::CalcDistanceDerivatives(
+        plant, *context, frameA, frameB,
+        // GetPoseInFrame() returns RigidTransform<double> -- we can't
+        // multiply across heterogeneous scalar types; so we cast the double
+        // to T.
+        inspector.GetPoseInFrame(signed_distance_pairs[i].id_A)
+                .template cast<T>() *
+            signed_distance_pairs[i].p_ACa,
+        signed_distance_pairs[i].distance, signed_distance_pairs[i].nhat_BA_W,
+        q, &distances(i));
   }
-  distances.resize(distance_count);
   return distances;
 }
 
--- a/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc
+++ b/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc
@@ -6,6 +6,7 @@
 #include "drake/common/default_scalars.h"
 #include "drake/common/find_resource.h"
 #include "drake/multibody/parsing/parser.h"
+#include "drake/multibody/plant/coulomb_friction.h"
 #include "drake/systems/framework/diagram_builder.h"
 
 using drake::geometry::SceneGraph;
--- a/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h
+++ b/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h
@@ -3,6 +3,7 @@
 #include <limits>
 #include <memory>
 #include <string>
+#include <vector>
 
 #include <gtest/gtest.h>
 
--- a/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc
+++ b/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc
@@ -307,5 +307,100 @@ TEST_F(BoxSphereTest, Test) {
                                 gradient_tol);
   }
 }
+
+/**
+ Constructs a diagram that contains N free-floating spheres.
+ */
+template <typename T>
+class NFreeSpheresModel {
+ public:
+  explicit NFreeSpheresModel(int num_spheres) : num_spheres_{num_spheres} {
+    systems::DiagramBuilder<T> builder;
+    std::tie(plant_, scene_graph_) = AddMultibodyPlantSceneGraph(&builder, 0.0);
+    const double mass{1};
+    const math::RigidTransformd X_BS{};
+    const Eigen::Vector3d p_AoAcm_A = X_BS.translation();
+    const RotationalInertia<double> I_AAcm_A =
+        UnitInertia<double>::SolidSphere(radius_);
+    const SpatialInertia<double> M_AAo_A =
+        SpatialInertia<double>::MakeFromCentralInertia(mass, p_AoAcm_A,
+                                                       I_AAcm_A);
+    for (int i = 0; i < num_spheres; ++i) {
+      auto& body = plant_->AddRigidBody("sphere" + std::to_string(i), M_AAo_A);
+      plant_->RegisterCollisionGeometry(body, X_BS, geometry::Sphere(radius_),
+                                        fmt::format("sphere{}_collision", i),
+                                        multibody::CoulombFriction<double>());
+      sphere_frame_indices_.push_back(body.body_frame().index());
+    }
+    plant_->Finalize();
+
+    diagram_ = builder.Build();
+
+    diagram_context_ = diagram_->CreateDefaultContext();
+  }
+
+  double radius() const { return radius_; }
+
+  const systems::Diagram<T>& diagram() const { return *diagram_; }
+
+  const MultibodyPlant<T>& plant() const { return *plant_; }
+
+  const systems::Context<T>& plant_context() const {
+    return diagram_->GetSubsystemContext(*plant_, *diagram_context_);
+  }
+
+  systems::Context<T>& get_mutable_plant_context() {
+    return plant_->GetMyMutableContextFromRoot(diagram_context_.get());
+  }
+
+ private:
+  int num_spheres_;
+  const double radius_{0.01};
+  std::unique_ptr<systems::Diagram<T>> diagram_;
+  // This plant points to a sub-system in diagram_.
+  MultibodyPlant<T>* plant_{nullptr};
+  // This scene_graph points to a sub-system in diagram_.
+  geometry::SceneGraph<T>* scene_graph_{nullptr};
+
+  std::vector<FrameIndex> sphere_frame_indices_;
+
+  std::unique_ptr<systems::Context<T>> diagram_context_;
+};
+
+GTEST_TEST(ThreeSpheresTest, SomeLargerThanInfluenceSomeSmallerThanMinimum) {
+  // Test the case with three spheres. Some pair of spheres have distance >
+  // d_influence, and some pairs have distance < d_min.
+  NFreeSpheresModel<double> three_spheres(3);
+  const double d_min = 0.05;
+  const double d_influence = 0.06;
+  // Indices into the q vector for each sphere's position.
+  Eigen::Index kSpheres[] = {4, 11, 18};
+  Eigen::VectorXd q =
+      three_spheres.plant().GetPositions(three_spheres.plant_context());
+  // Position for sphere 0.
+  q.segment<3>(kSpheres[0]) << 0, 0, 0;
+  // Position for sphere 1.
+  q.segment<3>(kSpheres[1]) << 0, 0, 0.05;
+  // Position for sphere 2.
+  q.segment<3>(kSpheres[2]) << 0, 0, 0.1;
+  // Make sure that distance(sphere0, sphere1) < d_min.
+  ASSERT_LT((q.segment<3>(kSpheres[0]) - q.segment<3>(kSpheres[1])).norm() -
+                2 * three_spheres.radius(),
+            d_min);
+  // Make sure distance(sphere0, sphere2) > d_influence.
+  ASSERT_GT((q.segment<3>(kSpheres[2]) - q.segment<3>(kSpheres[0])).norm() -
+                2 * three_spheres.radius(),
+            d_influence);
+  for (MinimumDistancePenaltyFunction penalty_function :
+       {QuadraticallySmoothedHingeLoss, ExponentiallySmoothedHingeLoss}) {
+    MinimumDistanceConstraint dut(&(three_spheres.plant()), d_min,
+                                  &(three_spheres.get_mutable_plant_context()),
+                                  penalty_function, d_influence - d_min);
+    Eigen::VectorXd y_val;
+    dut.Eval(q, &y_val);
+    EXPECT_TRUE((y_val.array() < dut.lower_bound().array()).any() ||
+                (y_val.array() > dut.upper_bound().array()).any());
+  }
+}
 }  // namespace multibody
 }  // namespace drake


hongkai-dai (commented):
+@seancurtis-tri for feature review please, thanks!

cc @russtedrake 

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459#-:-NJIh13p8IY3wdkJMbcC:b-3gl7dn)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
First pass done.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459#-:-NJLQp3T10z-65jgU2GL:b-te4kzy)* status: 16 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/inverse_kinematics/minimum_distance_constraint.cc` line 37 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLGHUh5m1hL-i1QBRz:-NJLGHUh5m1hL-i1QBS-:b8m1232) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/minimum_distance_constraint.cc#L37)):*
> ```cpp
>           influence_distance);
>   VectorX<S> distances(signed_distance_pairs.size());
>   int distance_count{0};
> ```

BTW `distance_count` no longer serves the same double role as before. Previously, it was the count *and* the index. Now it's just an index. You might consider renaming it.

This might be a case where looping over the index value may be better than range iteration:

```c++
for (int i = 0; i < signed_distance_pairs.size(); ++i) {
  const geometry::SignedDistancePair<T>& signed_distance_pair =
      signed_distance_pairs[i];
...
```

Or, if you want to get *really* aggressive:

```c++
  for (int i = -1; const auto& signed_distance_pair : signed_distance_pairs) {
    ...
    internal::CalcDistanceDerivatives(
        plant, *context, frameA, frameB,
        inspector.GetPoseInFrame(signed_distance_pair.id_A).template cast<T>() *
            signed_distance_pair.p_ACa,
        signed_distance_pair.distance, signed_distance_pair.nhat_BA_W, q,
        &distances(++i));
  }
```

Note: this is a C++20 functionality that is available with our current compilers (much like we can use designated initializers). Then you get both index and value together in the for loop declaration. Both variables nicely scoped.

(If you hate initializing it to -1, you can switch to i++ in references `distances` but I have a pathological aversion to post-incrementing.)
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 6 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLL5H6CI4BXJEOir8o:-NJLL5H71oMiItGhCh72:b8h60xf) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L6)):*
> ```objc
> #include <memory>
> #include <string>
> ```

nit: Other lint: `#include <vector>`
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 139 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLK985Cbc-QJj_4uuP:-NJLK985Cbc-QJj_4uuQ:bxejamo) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L139)):*
> ```objc
> };
> 
> template <typename T>
> ```

BTW This would benefit with a brief summary of *what* this does (the name isn't overly suggesting) and *why* it does it.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLKJLm1IFhd_vxKq2c:-NJLKJLn6Spy8pcVgJEZ:byza5gn) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*
> ```objc
> 
> template <typename T>
> class NFreeSpheres {
> ```

BTW This might benefit from a name change `NFreeSpheresModel`.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLMcPaB4kiri4FXSN4:-NJLMcPb1MGlBbOs7gTI:b-csgzkj) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*
> ```objc
> 
> template <typename T>
> class NFreeSpheres {
> ```

Do you have an explicit plan to use this in more tests than just the minimum distance constraint test? Right now, it doesn't feel justified to be defined in *this* file as it only has just that single use. If there's a known trajectory to greater usage, I can be ok with that.

This also relates to the note on unused APIs down below.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 142 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLKrsm6JxaD2Qq3kKz:-NJLKrsm6JxaD2Qq3kL-:b-k8xn9w) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L142)):*
> ```objc
> class NFreeSpheres {
>  public:
>   NFreeSpheres(int num_spheres);
> ```

nit: This is one of your lint pieces: `explicit NFreeSpherse(int num_spheres);`.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 148 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLM3RR1shBapADj6ab:-NJLM3RR1shBapADj6ac:btuxb83) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L148)):*
> ```objc
>   const systems::Diagram<T>& diagram() const { return *diagram_; }
> 
>   geometry::GeometryId GetSphereGeometryId(FrameIndex sphere_index) const {
> ```

nit: I'd request the following change (the API you use is something I'm desperately trying to eliminate, so removing new uses will help that).

```c++
  geometry::GeometryId GetSphereGeometryId(FrameIndex sphere_index) const {
    const geometry::FrameId frame_id = plant_->GetBodyFrameIdOrThrow(
        plant_->get_frame(sphere_index).body().index());
    return scene_graph_->model_inspector().GetGeometries(
        frame_id, geometry::Role::kProximity)[0];
  }
```

This gets *simpler* if you pass a `BodyIndex` instead of a `FrameIndex`. Also, you don't actually *use* this function. So, maybe it all goes away.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 167 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLNRDG0KHrdf8PRt06:-NJLNRDG0KHrdf8PRt07:b-n6vag9) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L167)):*
> ```objc
>   }
> 
>   systems::Context<T>& get_mutable_plant_context() { return *plant_context_; }
> ```

nit: As near as I can tell, the following APIs are unused:

1. `diagram()`
2. `GetSphereGeometryId()`
3. `scene_graph()`
4. `sphere_frame_indices()`
5. `X_BS()`
6. `get_mutable_diagram_context()`

Is this left over? Used in a future PR?
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 171 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLOvSmCZTqwyb-Btel:-NJLOvSmCZTqwyb-Btem:bw145d4) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L171)):*
> ```objc
>  private:
>   int num_spheres_;
>   double radius_{0.01};
> ```

nit: Currently, there's no reason to not makes this `constexpr double kRadius{0.01}`. Again, unless you're planning on allowing for variable radii sizes -- and then, the single double is insufficient.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 179 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLP9ui1hpOlR5z938h:-NJLP9ui1hpOlR5z938i:be9m9yh) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L179)):*
> ```objc
> 
>   // The pose of sphere i's collision geometry in sphere i's body frame.
>   math::RigidTransformd X_BS_;
> ```

A single transform for N spheres? That can't be right.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 145 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLPuH2-36teekfKgoD:-NJLPuH2-36teekfKgoE:b-udu07c) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L145)):*
> ```cpp
> template <typename T>
> NFreeSpheres<T>::NFreeSpheres(int num_spheres)
>     : num_spheres_{num_spheres}, X_BS_{Eigen::Vector3d(0.01, 0.02, 0.03)} {
> ```

nit: Why not at the Body origin?
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLQDGvDz-pyqWo5okX:-NJLQDGvDz-pyqWo5okY:b-p63jeb) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L149)):*
> ```cpp
>   std::tie(plant_, scene_graph_) = AddMultibodyPlantSceneGraph(&builder, 0.0);
>   const double mass{1};
>   const Eigen::Vector3d p_AoAcm_A(0, 0, 0);
> ```

Given X_BS_ not being zero, this center of mass seems like a misleading lie. Where the sphere is is not where the mass is, apparently. Your test doesn't *depend* on the mass, but why introduce surprises?
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 150 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLQ1MuEa_ICO2sQlFc:-NJLQ1MuEa_ICO2sQlFd:b-xp19f3) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L150)):*
> ```cpp
>   const double mass{1};
>   const Eigen::Vector3d p_AoAcm_A(0, 0, 0);
>   const RotationalInertia<double> I_AAcm_A{0.001, 0.001, 0.001};
> ```

BTW You could use `UnitInertia::SolidSphere(radius);`.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 165 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLQRCX3sig0cBpNqZ3:-NJLQRCX3sig0cBpNqZ4:b-knxel3) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L165)):*
> ```cpp
> 
>   diagram_context_ = diagram_->CreateDefaultContext();
>   plant_context_ =
> ```

BTW The preferred spelling:

```c++
plant_context_ = 
    &plant_->GetMyMutableContextFromRoot(diagram_context_.get());
```
___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 318 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLNA2J7DuNzR0Vq8tP:-NJLNA2J7DuNzR0Vq8tQ:b-qww3wl) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L318)):*
> ```cpp
>   const double d_influence = 0.06;
>   Eigen::VectorXd q = three_spheres.plant().GetPositions(
>       three_spheres.get_mutable_plant_context());
> ```

BTW This being a `mutable_plant_context()` is misleading; you aren't doing any mutating.
___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 319 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLPKOn6-5Xj6lz5Wsr:-NJLPKOn6-5Xj6lz5Wss:b-rmgsdy) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L319)):*
BTW I'd recommend a note regarding your magical indices: 4, 11, 18. And, something about the *purpose* of the poses of the sphere bodies. 

> Pose it so that the middle sphere is within the influence distance of each of its neighbors, but the outside neighbors are too far.

hongkai-dai (commented):
Thanks a lot for the thorough review!

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459#-:-NJMGpRxBfQbbSy44u_k:btlm0i5)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`multibody/inverse_kinematics/minimum_distance_constraint.cc` line 37 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLGHUh5m1hL-i1QBRz:-NJMH85C8CtdUeunBlqN:b-nnzyt4) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/minimum_distance_constraint.cc#L37)):*

BTW `distance_count` no longer serves the same double role as before. Previously, it was the count *and* the index. Now it's just an index. You might consider renaming it.

This might be a case where looping over the index value may be better than range iteration:

```c++
for (int i = 0; i < signed_distance_pairs.size(); ++i) {
  const geometry::SignedDistancePair<T>& signed_distance_pair =
      signed_distance_pairs[i];
...
```

Or, if you want to get *really* aggressive:

```c++
  for (int i = -1; const auto& signed_distance_pair : signed_distance_pairs) {
    ...
    internal::CalcDistanceDerivatives(
        plant, *context, frameA, frameB,
        inspector.GetPoseInFrame(signed_distance_pair.id_A).template cast<T>() *
            signed_distance_pair.p_ACa,
        signed_distance_pair.distance, signed_distance_pair.nhat_BA_W, q,
        &distances(++i));
  }
```

Note: this is a C++20 functionality that is available with our current compilers (much like we can use designated initializers). Then you get both index and value together in the for loop declaration. Both variables nicely scoped.

(If you hate initializing it to -1, you can switch to i++ in references `distances` but I have a pathological aversion to post-incrementing.)
</blockquote></details>

Done, I switched to iterating over i.

Thanks Sean, TIL new C++ 20 functionality. It gets more similar to python, with `enumerate` function.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 139 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLK985Cbc-QJj_4uuP:-NJMN7jf2XBnAfblAIjS:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L139)):*

BTW This would benefit with a brief summary of *what* this does (the name isn't overly suggesting) and *why* it does it.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLKJLm1IFhd_vxKq2c:-NJNH9oe00tbFwLuaE7G:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*

BTW This might benefit from a name change `NFreeSpheresModel`.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLMcPaB4kiri4FXSN4:-NJNHF1g34PSUM6eAXQY:b8o569v) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*

Do you have an explicit plan to use this in more tests than just the minimum distance constraint test? Right now, it doesn't feel justified to be defined in *this* file as it only has just that single use. If there's a known trajectory to greater usage, I can be ok with that.

This also relates to the note on unused APIs down below.
</blockquote></details>

I imagine this class to be used for other kinematic constraint that involve multiple (>=2 ) bodies. For example in the future we might want to write a constraint that a camera (body1) looks at the gripper (body2) while the line of sight is not occluded by obstacles (body3).
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 142 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLKrsm6JxaD2Qq3kKz:-NJNHvjeCkZpWakKw7hq:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L142)):*

nit: This is one of your lint pieces: `explicit NFreeSpherse(int num_spheres);`.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 148 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLM3RR1shBapADj6ab:-NJNHyIY21HPdf7gSkmb:bnhl1kz) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L148)):*

nit: I'd request the following change (the API you use is something I'm desperately trying to eliminate, so removing new uses will help that).

```c++
  geometry::GeometryId GetSphereGeometryId(FrameIndex sphere_index) const {
    const geometry::FrameId frame_id = plant_->GetBodyFrameIdOrThrow(
        plant_->get_frame(sphere_index).body().index());
    return scene_graph_->model_inspector().GetGeometries(
        frame_id, geometry::Role::kProximity)[0];
  }
```

This gets *simpler* if you pass a `BodyIndex` instead of a `FrameIndex`. Also, you don't actually *use* this function. So, maybe it all goes away.
</blockquote></details>

Done. You are right. I will just remove this function
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 167 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLNRDG0KHrdf8PRt06:-NJNII-8-d-At98iphS3:b-c119yd) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L167)):*

nit: As near as I can tell, the following APIs are unused:

1. `diagram()`
2. `GetSphereGeometryId()`
3. `scene_graph()`
4. `sphere_frame_indices()`
5. `X_BS()`
6. `get_mutable_diagram_context()`

Is this left over? Used in a future PR?
</blockquote></details>

I typically just add getters for the class, hence `diagram(), scene_graph(), X_BS()`. I removed function 2, 4, 6
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 171 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLOvSmCZTqwyb-Btel:-NJNId3wCi-xtlR_yNn7:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L171)):*

nit: Currently, there's no reason to not makes this `constexpr double kRadius{0.01}`. Again, unless you're planning on allowing for variable radii sizes -- and then, the single double is insufficient.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 179 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLP9ui1hpOlR5z938h:-NJNImGw5hY7a_tMj18u:bysokq7) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L179)):*

A single transform for N spheres? That can't be right.
</blockquote></details>

Done. I added more documentation.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 145 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLPuH2-36teekfKgoD:-NJNJ-NHE0CXSqWWlB7Z:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L145)):*

nit: Why not at the Body origin?
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLQDGvDz-pyqWo5okX:-NJNJBxp654qHy13Z8Cn:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.cc#L149)):*

Given X_BS_ not being zero, this center of mass seems like a misleading lie. Where the sphere is is not where the mass is, apparently. Your test doesn't *depend* on the mass, but why introduce surprises?
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 319 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLPKOn6-5Xj6lz5Wsr:-NJNK9cUEfsnm-JZ48WA:b-q9oyrl) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L319)):*

BTW I'd recommend a note regarding your magical indices: 4, 11, 18. And, something about the *purpose* of the poses of the sphere bodies. 

> Pose it so that the middle sphere is within the influence distance of each of its neighbors, but the outside neighbors are too far.
</blockquote></details>

Done. I changed documentation below.

SeanCurtis-TRI (commented):
I'm still not happy with the N-sphere model. However, I'm marking myself satisified and will let the platform reviewer weigh in.

:LGTM:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459#-:-NJQIZFaDVVVN9gSU5cn:b-mv9lun)* status: LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLMcPaB4kiri4FXSN4:-NJQGfbM1LDoMIDdS_Fl:bglv9jp) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*

I imagine this class to be used for other kinematic constraint that involve multiple (>=2 ) bodies. For example in the future we might want to write a constraint that a camera (body1) looks at the gripper (body2) while the line of sight is not occluded by obstacles (body3).
</blockquote></details>

"I imagine" feels very speculative. If you had a concrete plan, its' easier to be forgiving. But this sounds like a vague speculation.

Its presence here is *particularly* odd because for several reasons:

1. It's only used on one specific test, but has been hoisted to the general location - without any specific plan on where else it will be used.
2. We have the two-sphere model test. It feels like in an ideal world the two-sphere test is a special case of the n-sphere test.
3. This class *still* has unused APIs (albeit with some better documentation).

I can't make a strict legal argument for why this is defective, but it all smells bad. Therefore, I'm marking myself as satisified.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 179 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLP9ui1hpOlR5z938h:-NJQHOb98qPlQ4OkTYGw:b-515d9k) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L179)):*

Done. I added more documentation.
</blockquote></details>

The documentation helps....but....(see note above). And now it's simply the identity....

1) It can't be configured.
2) It never gets queried
3) It's simply the identity.

Feels quite superfluous.

ggould-tri (commented):
A few issues below; as Sean says `NFreeSpheresModel` is a bit messy.  I've put in some suggestions for simplifying it below.
:lgtm: pending those issues.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459#-:-NJf-Lx11Dw0WkawGiZh:br1d7gx)* status: 6 unresolved discussions, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 155 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfGp0SDSfmND1N6eCQ:-NJfGp0SDSfmND1N6eCR:bdjpjkc) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L155)):*
> ```objc
>   const MultibodyPlant<T>& plant() const { return *plant_; }
> 
>   const geometry::SceneGraph<T>& scene_graph() const { return *scene_graph_; }
> ```

minor: unused
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 167 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfGB4nCLVpQPp4a7zp:-NJfGB4nCLVpQPp4a7zq:b-f9k7dp) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L167)):*
> ```objc
>   const double radius_{0.01};
>   std::unique_ptr<systems::Diagram<T>> diagram_;
>   MultibodyPlant<T>* plant_{nullptr};
> ```

minor:  These bare pointer members with no clear ownership are a bad sign.
Assuming that computing them at their use site is impractical, so I suggest commenting that they point to elements of `diagram_`.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 172 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfF4-46ge5AaZFIjUV:-NJfF4-46ge5AaZFIjUW:b8a9ykt) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L172)):*
> ```objc
>   std::vector<FrameIndex> sphere_frame_indices_;
> 
>   // The pose of sphere i's collision geometry in sphere i's body frame, every
> ```

nit: grammar



_Suggestion:_
```objc
  // The pose of sphere i's collision geometry in sphere i's body frame; every
  // sphere has the same pose.
```
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 174 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfFgH3DNXrqRjBJ67n:-NJfFgH3DNXrqRjBJ67o:b2m5j8g) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L174)):*
> ```objc
>   // The pose of sphere i's collision geometry in sphere i's body frame, every
>   // sphere share the same pose.
>   math::RigidTransformd X_BS_;
> ```

minor:  Unused?  Seems like this is always identity.  And if it were used it would probably be confusing anyway.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 177 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfHTAz2iiHtZrJAC7i:-NJfHTAz2iiHtZrJAC7j:bbfdvwp) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L177)):*
> ```objc
> 
>   std::unique_ptr<systems::Context<T>> diagram_context_;
>   systems::Context<T>* plant_context_{nullptr};
> ```

minor:  Is there an advantage to storing this pointer rather than just calling `GetMyMutableContextFromRoot` in `GetMutablePlantContext`?  There's no interesting performance gain that I can see, and the extra level of indirection and stored bare pointer seem unnecessary.
___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 323 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfIfLQ0xC8Z1yRlpLu:-NJfIfLQ0xC8Z1yRlpLv:bsxvnks) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L323)):*
> ```cpp
>   // Position for sphere 1.
>   q.segment<3>(11) << 0, 0, 0.05;
>   // Position for sphere 2.
> ```

minor:  This could be clearer if you used a variable name instead of comments to explain your magic numbers.  For instance:



_Suggestion:_
```cpp
  // Indexes into the q vector for each sphere's position.
  Eigen::Index kSpheres[] = {4, 11, 18};
  q.segment<3>(kSpheres[0]) << 0, 0, 0;
  q.segment<3>(kSpheres[1]) << 0, 0, 0.05;
  q.segment<3>(kSpheres[2]) << 0, 0, 0.1;
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459)* status: 6 unresolved discussions, missing label for release notes (waiting on @ggould-tri and @hongkai-dai)

___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 323 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfIfLQ0xC8Z1yRlpLu:-NJfpv961SHSwje4vSv5:b-5a8b28) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L323)):*

minor:  This could be clearer if you used a variable name instead of comments to explain your magic numbers.  For instance:
</blockquote></details>

BTW IN drake we prefer "indices" to "indexes".

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459)* status: 6 unresolved discussions, missing label for release notes (waiting on @hongkai-dai and @SeanCurtis-TRI)

___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 323 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfIfLQ0xC8Z1yRlpLu:-NJg-G3t70QVrV-W8Pm6:b4ygvbe) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L323)):*

BTW IN drake we prefer "indices" to "indexes". 
</blockquote></details>

colon facepalm emoji colon.  I think the general point still stands, though :-)

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18459)* status: missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLMcPaB4kiri4FXSN4:-NJgNNaE6zQZ4f9O6R2h:b1dazcr) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L140)):*

"I imagine" feels very speculative. If you had a concrete plan, its' easier to be forgiving. But this sounds like a vague speculation.

Its presence here is *particularly* odd because for several reasons:

1. It's only used on one specific test, but has been hoisted to the general location - without any specific plan on where else it will be used.
2. We have the two-sphere model test. It feels like in an ideal world the two-sphere test is a special case of the n-sphere test.
3. This class *still* has unused APIs (albeit with some better documentation).

I can't make a strict legal argument for why this is defective, but it all smells bad. Therefore, I'm marking myself as satisified.
</blockquote></details>

Done. I moved this class to minimum_distance_constraint_test.cc
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 179 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJLP9ui1hpOlR5z938h:-NJgPAoq8Xcvv4LhVXpv:b-hhi2w9) ([raw file](https://github.com/robotlocomotion/drake/blob/ff7d8f9b68f04f2b8aea9cff4ce7ddb80355788e/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L179)):*

The documentation helps....but....(see note above). And now it's simply the identity....

1) It can't be configured.
2) It never gets queried
3) It's simply the identity.

Feels quite superfluous.
</blockquote></details>

Done. I removed X_BS_.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 167 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfGB4nCLVpQPp4a7zp:-NJgNf4vCCxfRP7qNwh4:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L167)):*

minor:  These bare pointer members with no clear ownership are a bad sign.
Assuming that computing them at their use site is impractical, so I suggest commenting that they point to elements of `diagram_`.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 172 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfF4-46ge5AaZFIjUV:-NJgP7fJ4gB42A2_zHdF:bqvv7ac) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L172)):*

nit: grammar
</blockquote></details>

Done. I removed `X_BS_` from the class.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 174 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfFgH3DNXrqRjBJ67n:-NJgPE9R8zfplizV4VVT:b-hhi2w9) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L174)):*

minor:  Unused?  Seems like this is always identity.  And if it were used it would probably be confusing anyway.
</blockquote></details>

Done. I removed X_BS_.
___
*[`multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h` line 177 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfHTAz2iiHtZrJAC7i:-NJgPvoH2JnGAc4jLwqI:b-olsewu) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h#L177)):*

minor:  Is there an advantage to storing this pointer rather than just calling `GetMyMutableContextFromRoot` in `GetMutablePlantContext`?  There's no interesting performance gain that I can see, and the extra level of indirection and stored bare pointer seem unnecessary.
</blockquote></details>

Done. I removed `plant_context_`.
___
*[`multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc` line 323 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18459#-NJfIfLQ0xC8Z1yRlpLu:-NJgTMF83qUmLPdfDaYF:b-yh16l6) ([raw file](https://github.com/robotlocomotion/drake/blob/a995bd3a700bf7b04bbb2aa68e9fc312871d7656/multibody/inverse_kinematics/test/minimum_distance_constraint_test.cc#L323)):*

colon facepalm emoji colon.  I think the general point still stands, though :-)
</blockquote></details>

Done. Good call.

PR #18470, opened 2022-12-15
[multibody] Add 3-dof ball constraints to SAP

joemasterjohn:
Implements support for ball constraints when using the SAP solver.
With this PR the user can set a ball constraint with the call to `MultibodyPlant::AddBallConstraint()`

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470)

### DIFF
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -473,6 +473,16 @@ drake_cc_googletest(
 )
 
 drake_cc_googletest(
+    name = "sap_driver_ball_constraints_test",
+    deps = [
+        ":compliant_contact_manager_tester",
+        ":plant",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
     name = "sap_driver_joint_limits_test",
     data = [
         "//manipulation/models/iiwa_description:models",
--- a/multibody/plant/constraint_specs.h
+++ b/multibody/plant/constraint_specs.h
@@ -66,6 +66,28 @@ struct DistanceConstraintSpecs {
   double damping{0.0};  // Constraint damping c in N⋅s/m.
 };
 
+// Struct to store the specification for a ball constraint. A ball
+// constraint is modeled as a holonomic constraint:
+//   p_PQ_W(q) = 0
+// P is a point rigidly affixed to body A and Q is a point rigidly affixed to
+// body B. p_PQ_W(q) denotes the relative position of point Q with respect to
+// point P, expressed in the world frame, as a function of the configuration of
+// the model q. Imposing this constraint forces P and Q to be coincident, but
+// does not restrict the rotational degrees of freedom.
+//
+// @pre body_A != body_B. @see IsValid().
+struct BallConstraintSpecs {
+  // Returns `true` iff `this` specification is valid to define a ball
+  // constraint. A ball constraint specification is considered to be valid iff:
+  //   body_A != body_B.
+  bool IsValid() { return body_A != body_B; }
+
+  BodyIndex body_A;      // Index of body A.
+  Vector3<double> p_AP;  // Position of point P in body frame A.
+  BodyIndex body_B;      // Index of body B.
+  Vector3<double> p_BQ;  // Position of point Q in body frame B.
+};
+
 }  // namespace internal
 }  // namespace multibody
 }  // namespace drake
--- a/multibody/plant/discrete_update_manager.cc
+++ b/multibody/plant/discrete_update_manager.cc
@@ -166,6 +166,13 @@ DiscreteUpdateManager<T>::distance_constraints_specs() const {
 }
 
 template <typename T>
+const std::vector<internal::BallConstraintSpecs>&
+DiscreteUpdateManager<T>::ball_constraints_specs() const {
+  return MultibodyPlantDiscreteUpdateManagerAttorney<T>::ball_constraints_specs(
+      *plant_);
+}
+
+template <typename T>
 BodyIndex DiscreteUpdateManager<T>::FindBodyByGeometryId(
     geometry::GeometryId geometry_id) const {
   return MultibodyPlantDiscreteUpdateManagerAttorney<T>::FindBodyByGeometryId(
--- a/multibody/plant/discrete_update_manager.h
+++ b/multibody/plant/discrete_update_manager.h
@@ -248,6 +248,9 @@ class DiscreteUpdateManager : public ScalarConvertibleComponent<T> {
   const std::vector<internal::DistanceConstraintSpecs>&
   distance_constraints_specs() const;
 
+  const std::vector<internal::BallConstraintSpecs>& ball_constraints_specs()
+      const;
+
   BodyIndex FindBodyByGeometryId(geometry::GeometryId geometry_id) const;
   /* @} */
 
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -392,6 +392,7 @@ MultibodyPlant<T>::MultibodyPlant(const MultibodyPlant<U>& other)
 
     coupler_constraints_specs_ = other.coupler_constraints_specs_;
     distance_constraints_specs_ = other.distance_constraints_specs_;
+    ball_constraints_specs_ = other.ball_constraints_specs_;
 
     // cache_indexes_ is set in DeclareCacheEntries() in
     // DeclareStateCacheAndPorts() in FinalizePlantOnly().
@@ -500,6 +501,49 @@ ConstraintIndex MultibodyPlant<T>::AddDistanceConstraint(
   const ConstraintIndex constraint_index(num_constraints());
 
   distance_constraints_specs_.push_back(spec);
+
+  return constraint_index;
+}
+
+template <typename T>
+ConstraintIndex MultibodyPlant<T>::AddBallConstraint(
+    const Body<T>& body_A, const Vector3<double>& p_AP, const Body<T>& body_B,
+    const Vector3<double>& p_BQ) {
+  // N.B. The manager is set up at Finalize() and therefore we must require
+  // constraints to be added pre-finalize.
+  DRAKE_MBP_THROW_IF_FINALIZED();
+
+  if (!is_discrete()) {
+    throw std::runtime_error(
+        "Currently ball constraints are only supported for discrete "
+        "MultibodyPlant models.");
+  }
+
+  // TAMSI does not support ball constraints. For all other solvers, we let
+  // the discrete update manager throw an exception at finalize time.
+  if (contact_solver_enum_ == DiscreteContactSolver::kTamsi) {
+    throw std::runtime_error(
+        "Currently this MultibodyPlant is set to use the TAMSI solver. TAMSI "
+        "does not support ball constraints. Use "
+        "set_discrete_contact_solver(DiscreteContactSolver::kSap) to use the "
+        "SAP solver instead. For other solvers, refer to "
+        "DiscreteContactSolver.");
+  }
+
+  internal::BallConstraintSpecs spec{body_A.index(), p_AP, body_B.index(),
+                                     p_BQ};
+  if (!spec.IsValid()) {
+    const std::string msg = fmt::format(
+        "Invalid set of parameters for constraint between bodies '{}' and "
+        "'{}'. For a ball constraint, points P and Q must be on two distinct "
+        "bodies, i.e. body_A != body_B must be satisfied.",
+        body_A.name(), body_B.name());
+    throw std::logic_error(msg);
+  }
+
+  const ConstraintIndex constraint_index(num_constraints());
+
+  ball_constraints_specs_.push_back(spec);
 
   return constraint_index;
 }
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -1195,7 +1195,8 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   /// Returns the total number of constraints specified by the user.
   int num_constraints() const {
     return static_cast<int>(coupler_constraints_specs_.size() +
-           distance_constraints_specs_.size());
+                            distance_constraints_specs_.size() +
+                            ball_constraints_specs_.size());
   }
 
   /// Defines a holonomic constraint between two single-dof joints `joint0`
@@ -1276,6 +1277,23 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
       const Vector3<double>& p_BQ, double distance,
       double stiffness = std::numeric_limits<double>::infinity(),
       double damping = 0.0);
+
+  /// Defines a constraint such that point P affixed to body A is coincident at
+  /// all times with point Q affixed to body B, effectively modeling a
+  /// ball-and-socket joint.
+  ///
+  /// @param[in] body_A Body to which point P is rigidly attached.
+  /// @param[in] p_AP Position of point P in body A's frame.
+  /// @param[in] body_B Body to which point Q is rigidly attached.
+  /// @param[in] p_BQ Position of point Q in body B's frame.
+  /// @returns the index to the newly added constraint.
+  ///
+  /// @throws std::exception if bodies A and B are the same body.
+  /// @throws std::exception if the %MultibodyPlant has already been finalized.
+  ConstraintIndex AddBallConstraint(const Body<T>& body_A,
+                                    const Vector3<double>& p_AP,
+                                    const Body<T>& body_B,
+                                    const Vector3<double>& p_BQ);
 
   /// <!-- TODO(xuchenhan-tri): Add getters to interrogate existing constraints.
   /// -->
@@ -5246,6 +5264,9 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
 
   // Vector of distance constraints specifications.
   std::vector<internal::DistanceConstraintSpecs> distance_constraints_specs_;
+
+  // Vector of ball constraint specifications.
+  std::vector<internal::BallConstraintSpecs> ball_constraints_specs_;
 
   // All MultibodyPlant cache indexes are stored in cache_indexes_.
   CacheIndexes cache_indexes_;
--- a/multibody/plant/multibody_plant_discrete_update_manager_attorney.h
+++ b/multibody/plant/multibody_plant_discrete_update_manager_attorney.h
@@ -119,6 +119,11 @@ class MultibodyPlantDiscreteUpdateManagerAttorney {
     return plant.distance_constraints_specs_;
   }
 
+  static const std::vector<internal::BallConstraintSpecs>&
+  ball_constraints_specs(const MultibodyPlant<T>& plant) {
+    return plant.ball_constraints_specs_;
+  }
+
   static BodyIndex FindBodyByGeometryId(const MultibodyPlant<T>& plant,
                                         geometry::GeometryId geometry_id) {
     return plant.FindBodyByGeometryId(geometry_id);
--- a/multibody/plant/sap_driver.cc
+++ b/multibody/plant/sap_driver.cc
@@ -467,6 +467,105 @@ void SapDriver<T>::AddDistanceConstraints(const systems::Context<T>& context,
 }
 
 template <typename T>
+void SapDriver<T>::AddBallConstraints(
+    const systems::Context<T>& context,
+    contact_solvers::internal::SapContactProblem<T>* problem) const {
+  DRAKE_DEMAND(problem != nullptr);
+
+  // Ball constraints do not have impulse limits, they are bi-lateral
+  // constraints. Each ball constraint introduces three constraint
+  // equations.
+  constexpr double kInfinity = std::numeric_limits<double>::infinity();
+  const Vector3<T> gamma_lower(-kInfinity, -kInfinity, -kInfinity);
+  const Vector3<T> gamma_upper(kInfinity, kInfinity, kInfinity);
+
+  // Stiffness and dissipation are set so that the constraint is in the
+  // "near-rigid" regime, [Castro et al., 2022].
+  const Vector3<T> stiffness(kInfinity, kInfinity, kInfinity);
+  const Vector3<T> relaxation_time = plant().time_step() * Vector3<T>::Ones();
+
+  const int nv = plant().num_velocities();
+  Matrix3X<T> Jv_WAp_W(3, nv);
+  Matrix3X<T> Jv_WBq_W(3, nv);
+  MatrixX<T> Jv_ApBq_W = MatrixX<T>::Zero(3, nv);
+
+  const Frame<T>& frame_W = plant().world_frame();
+  for (const BallConstraintSpecs& specs :
+       manager().ball_constraints_specs()) {
+    const Body<T>& body_A = plant().get_body(specs.body_A);
+    const Body<T>& body_B = plant().get_body(specs.body_B);
+
+    const math::RigidTransform<T>& X_WA =
+        plant().EvalBodyPoseInWorld(context, body_A);
+    const math::RigidTransform<T>& X_WB =
+        plant().EvalBodyPoseInWorld(context, body_B);
+    const Vector3<T> p_WP = X_WA * specs.p_AP.cast<T>();
+    const Vector3<T> p_WQ = X_WB * specs.p_BQ.cast<T>();
+
+    const Vector3<T> p_PQ_W = p_WQ - p_WP;
+
+    // Dense Jacobian.
+    // d(p_PQ_W)/dt = Jv_ApBq_W * v.
+    manager().internal_tree().CalcJacobianTranslationalVelocity(
+        context, JacobianWrtVariable::kV, body_A.body_frame(), frame_W, p_WP,
+        frame_W, frame_W, &Jv_WAp_W);
+    manager().internal_tree().CalcJacobianTranslationalVelocity(
+        context, JacobianWrtVariable::kV, body_B.body_frame(), frame_W, p_WQ,
+        frame_W, frame_W, &Jv_WBq_W);
+    Jv_ApBq_W = (Jv_WBq_W - Jv_WAp_W);
+
+    // TODO(amcastro-tri): consider exposing this parameter.
+    const double beta = 0.1;
+    const typename SapHolonomicConstraint<T>::Parameters parameters{
+        gamma_lower, gamma_upper, stiffness, relaxation_time, beta};
+
+    const TreeIndex treeA_index =
+        tree_topology().body_to_tree_index(specs.body_A);
+    const TreeIndex treeB_index =
+        tree_topology().body_to_tree_index(specs.body_B);
+
+    // TODO(joemasterjohn): Move this exception up to the plant level so that it
+    // fails as fast as possible. Currently, the earliest this can happen is
+    // in MbP::Finalize() after the topology has been finalized.
+    if (!treeA_index.is_valid() && !treeB_index.is_valid()) {
+      const std::string msg = fmt::format(
+          "Creating a ball Constraint between bodies '{}' and '{}' where both "
+          "are welded to the world is not allowed.",
+          body_A.name(), body_B.name());
+      throw std::runtime_error(msg);
+    }
+
+    // Both bodies A and B belong to the same tree or one of them is the world.
+    const bool single_tree = !treeA_index.is_valid() ||
+                             !treeB_index.is_valid() ||
+                             treeA_index == treeB_index;
+
+    // Constraint function at current time step.
+    const Vector3<T> g0 = p_PQ_W;
+    if (single_tree) {
+      const TreeIndex tree_index =
+          treeA_index.is_valid() ? treeA_index : treeB_index;
+      const MatrixX<T> J = Jv_ApBq_W.middleCols(
+          tree_topology().tree_velocities_start(tree_index),
+          tree_topology().num_tree_velocities(tree_index));
+
+      problem->AddConstraint(std::make_unique<SapHolonomicConstraint<T>>(
+          tree_index, g0, std::move(J), parameters));
+    } else {
+      const MatrixX<T> JA = Jv_ApBq_W.middleCols(
+          tree_topology().tree_velocities_start(treeA_index),
+          tree_topology().num_tree_velocities(treeA_index));
+      const MatrixX<T> JB = Jv_ApBq_W.middleCols(
+          tree_topology().tree_velocities_start(treeB_index),
+          tree_topology().num_tree_velocities(treeB_index));
+      problem->AddConstraint(std::make_unique<SapHolonomicConstraint<T>>(
+          treeA_index, treeB_index, g0, std::move(JA), std::move(JB),
+          parameters));
+    }
+  }
+}
+
+template <typename T>
 void SapDriver<T>::CalcContactProblemCache(
     const systems::Context<T>& context, ContactProblemCache<T>* cache) const {
   SapContactProblem<T>& problem = *cache->sap_problem;
@@ -483,6 +582,7 @@ void SapDriver<T>::CalcContactProblemCache(
   AddLimitConstraints(context, problem.v_star(), &problem);
   AddCouplerConstraints(context, &problem);
   AddDistanceConstraints(context, &problem);
+  AddBallConstraints(context, &problem);
 }
 
 template <typename T>
--- a/multibody/plant/sap_driver.h
+++ b/multibody/plant/sap_driver.h
@@ -144,6 +144,14 @@ class SapDriver {
       const systems::Context<T>& context,
       contact_solvers::internal::SapContactProblem<T>* problem) const;
 
+  // Adds holonomic constraint equations to model ball constraints specified in
+  // the MultibodyPlant.
+  // @throws std::exception if both bodies have invalid tree indices from the
+  // tree topology (i.e. both bodies are welded to `world`).
+  void AddBallConstraints(
+      const systems::Context<T>& context,
+      contact_solvers::internal::SapContactProblem<T>* problem) const;
+
   // This method takes SAP results for a given `problem` and loads forces due to
   // contact only into `contact_results`. `contact_results` is properly resized
   // on output.
--- a/multibody/plant/test/sap_driver_ball_constraints_test.cc
+++ b/multibody/plant/test/sap_driver_ball_constraints_test.cc
@@ -1,0 +1,286 @@
+#include <memory>
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
+#include "drake/math/rigid_transform.h"
+#include "drake/multibody/contact_solvers/sap/sap_contact_problem.h"
+#include "drake/multibody/contact_solvers/sap/sap_holonomic_constraint.h"
+#include "drake/multibody/plant/compliant_contact_manager.h"
+#include "drake/multibody/plant/multibody_plant.h"
+#include "drake/multibody/plant/sap_driver.h"
+#include "drake/multibody/plant/test/compliant_contact_manager_tester.h"
+
+/* @file This file tests SapDriver's support for ball constraints. */
+
+using drake::math::RigidTransformd;
+using drake::multibody::contact_solvers::internal::SapContactProblem;
+using drake::multibody::contact_solvers::internal::SapHolonomicConstraint;
+using drake::systems::Context;
+using Eigen::MatrixXd;
+using Eigen::Vector3d;
+using Eigen::VectorXd;
+
+constexpr double kInfinity = std::numeric_limits<double>::infinity();
+
+namespace drake {
+namespace multibody {
+namespace internal {
+
+// Friend class used to provide access to a selection of private functions in
+// SapDriver for testing purposes.
+class SapDriverTest {
+ public:
+  static const ContactProblemCache<double>& EvalContactProblemCache(
+      const SapDriver<double>& driver, const Context<double>& context) {
+    return driver.EvalContactProblemCache(context);
+  }
+};
+
+struct TestConfig {
+  // This is a gtest test suffix; no underscores or spaces.
+  std::string description;
+  bool bodyA_anchored{};
+};
+
+// This provides the suffix for each test parameter: the test config
+// description.
+std::ostream& operator<<(std::ostream& out, const TestConfig& c) {
+  out << c.description;
+  return out;
+}
+
+// Fixture that sets a MultibodyPlant model of two bodies A and B with a
+// ball constraint connecting point P on body A and point Q on body B.
+class TwoBodiesTest : public ::testing::TestWithParam<TestConfig> {
+ public:
+  // Makes the model described in the fixture's documentation.
+  // @param[in] anchor_bodyA If true, body A will be anchored to the world.
+  // Otherwise body A has 6 DOFs as body B does.
+  void MakeModel(bool anchor_bodyA) {
+    plant_.set_discrete_contact_solver(DiscreteContactSolver::kSap);
+
+    // Arbitrary inertia values only used by the driver to build a valid contact
+    // problem.
+    const double mass = 1.5;
+    const double radius = 0.1;
+    const SpatialInertia<double> M_Bo =
+        SpatialInertia<double>::MakeFromCentralInertia(
+            mass, Vector3d::Zero(),
+            UnitInertia<double>::SolidSphere(radius) * mass);
+
+    bodyA_ = &plant_.AddRigidBody("A", M_Bo);
+    bodyB_ = &plant_.AddRigidBody("B", M_Bo);
+    if (anchor_bodyA) {
+      plant_.WeldFrames(plant_.world_frame(), bodyA_->body_frame());
+    }
+
+    plant_.AddBallConstraint(*bodyA_, p_AP_, *bodyB_, p_BQ_);
+
+    plant_.Finalize();
+
+    auto owned_contact_manager =
+        std::make_unique<CompliantContactManager<double>>();
+    manager_ = owned_contact_manager.get();
+    plant_.SetDiscreteUpdateManager(std::move(owned_contact_manager));
+    // Model with a single ball constraint.
+    EXPECT_EQ(plant_.num_constraints(), 1);
+
+    context_ = plant_.CreateDefaultContext();
+  }
+
+  const SapDriver<double>& sap_driver() const {
+    return CompliantContactManagerTester::sap_driver(*manager_);
+  }
+
+ protected:
+  MultibodyPlant<double> plant_{0.01};  // Discrete model.
+  const RigidBody<double>* bodyA_{nullptr};
+  const RigidBody<double>* bodyB_{nullptr};
+  CompliantContactManager<double>* manager_{nullptr};
+  std::unique_ptr<Context<double>> context_;
+
+  // Parameters of the problem. Arbitrary and non-zero.
+  Vector3d p_AP_{1.0, -2.0, 3.0};
+  Vector3d p_BQ_{-5.0, 4.0, -6.0};
+  const Vector3d kOffset_{0.1, 0.2, 0.3};
+};
+
+// This test configures a single ball constraint in variety of ways (causing
+// differing numbers of cliques and varying constraint properties). It then
+// examines the newly added constraint to confirm that its instantiation
+// reflects the specification.
+TEST_P(TwoBodiesTest, ConfirmConstraintProperties) {
+  const TestConfig& config = GetParam();
+
+  MakeModel(config.bodyA_anchored);
+
+  const int expected_num_velocities = config.bodyA_anchored ? 6 : 12;
+  EXPECT_EQ(plant_.num_velocities(), expected_num_velocities);
+
+  // Place Bo at known offset from Ao; this does *not* satisfy the
+  // constraint. We'll observe a non-zero value when evaluating the constraint
+  // function.
+  plant_.SetFreeBodyPoseInWorldFrame(context_.get(), *bodyB_,
+                                     RigidTransformd(kOffset_));
+
+  const ContactProblemCache<double>& problem_cache =
+      SapDriverTest::EvalContactProblemCache(sap_driver(), *context_);
+  const SapContactProblem<double>& problem = *problem_cache.sap_problem;
+
+  // Verify the expected number of constraints and equations for a single
+  // ball constraint.
+  EXPECT_EQ(problem.num_constraints(), 1);
+  EXPECT_EQ(problem.num_constraint_equations(), 3);
+
+  const auto* constraint = dynamic_cast<const SapHolonomicConstraint<double>*>(
+      &problem.get_constraint(0));
+  // Verify it is a SapHolonomicConstraint as expected.
+  ASSERT_NE(constraint, nullptr);
+
+  // One clique if body A is anchored, two if not.
+  const int expected_num_cliques = config.bodyA_anchored ? 1 : 2;
+  EXPECT_EQ(constraint->num_cliques(), expected_num_cliques);
+  EXPECT_EQ(constraint->first_clique(), 0);
+  if (expected_num_cliques == 1) {
+    EXPECT_THROW(constraint->second_clique(), std::exception);
+  } else {
+    EXPECT_EQ(constraint->second_clique(), 1);
+  }
+
+  // Verify parameters.
+  const SapHolonomicConstraint<double>::Parameters p = constraint->parameters();
+  EXPECT_EQ(p.num_constraint_equations(), 3);
+  // bi-lateral constraint with no impulse bounds.
+  EXPECT_EQ(p.impulse_lower_limits(),
+            Vector3d(-kInfinity, -kInfinity, -kInfinity));
+  EXPECT_EQ(p.impulse_upper_limits(),
+            Vector3d(kInfinity, kInfinity, kInfinity));
+
+  EXPECT_EQ(p.stiffnesses(), Vector3d(kInfinity, kInfinity, kInfinity));
+  EXPECT_EQ(p.relaxation_times(), plant_.time_step() * Vector3d::Ones());
+
+  // This value is hard-coded in the source. This test serves as a brake to
+  // prevent the value changing without notification. Changing this value
+  // would lead to a behavior change and shouldn't happen silently.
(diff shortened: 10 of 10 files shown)


joemasterjohn (commented):
+@xuchenhan-tri for feature review, please. cc: @amcastro-tri 

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470#-:-NJMx1fK1W_uigpay6iG:b-qnj9qd)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @xuchenhan-tri)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @joemasterjohn and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.cc` line 522 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJN2jVY8dXAOEZHSXN5:-NJN2jVY8dXAOEZHSXN6:bl38780) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L522)):*
> ```cpp
>   }
> 
>   // TAMSI does not support distance constraints. For all other solvers, we let
> ```

drive by paste-o: distance -> ball

xuchenhan-tri (commented):
I made a pass for non-testing code. Mostly minor comments along with a sign convention quibble and a question about jacobian calculation.

BTW, the deformable rigid weld constraint I'm working on is very similar to this!

> Reviewed 9 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470#-:-NJNYp6p9NCbJBdhQ2oc:blotxng)* status: 18 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @joemasterjohn)

___
*[`multibody/plant/constraint_specs.h` line 71 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNXUHdAG5LELx1zo1_:-NJNXUHdAG5LELx1zo1a:bdk26b5) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L71)):*
> ```objc
> // Struct to store the specification for a ball constraint. A ball
> // constraint is modeled as a holonomic constraint:
> //   p_QP_W(q) = 0
> ```

BTW, can I persuade you to do `p_PQ_W` instead of `p_QP_W`. This makes the sign convention the same as in contact and distance constraints and creates less mental dissonance (at least for myself).
___
*[`multibody/plant/constraint_specs.h` line 72 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNQZmN7TwAWnXWCly9:-NJNQZmN7TwAWnXWClyA:b-ingyet) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L72)):*
> ```objc
> // constraint is modeled as a holonomic constraint:
> //   p_QP_W(q) = 0
> // We use p_QP_W(q) to denote the relative pose of point P
> ```

nit probably better to say "position" instead of "pose" as there's no orientation involved.

Also, weird line break.



_Suggestion:_
```objc
position
```
___
*[`multibody/plant/constraint_specs.h` line 73 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNRHTlDomX6EtA2TOt:-NJNRHTlDomX6EtA2TOu:b-e86k6c) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L73)):*
> ```objc
> //   p_QP_W(q) = 0
> // We use p_QP_W(q) to denote the relative pose of point P
> // with respect to point Q, rigidly affixed to bodies A and B respectively,
> ```

Probably worth saying a few words about what P and Q are. I'm assuming the ball constraint imposed is forcing them to be coincident. If so, please make that explicit.
___
*[`multibody/plant/constraint_specs.h` line 78 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNQywBD4IoOFBhXQtP:-NJNQywBD4IoOFBhXQtQ:b-1bv7sw) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L78)):*
> ```objc
> //
> // @pre body_A != body_B. @see IsValid().
> struct BallConstraintSpecs {
> ```

BTW, why do we name these struct `FooSpecs` instead of `FooSpec`? Each struct represents a single specification.
___
*[`multibody/plant/constraint_specs.h` line 80 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNRc4d5xcqfOqxhuHD:-NJNRc4d5xcqfOqxhuHE:b2ni8l0) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L80)):*
> ```objc
> struct BallConstraintSpecs {
>   // Returns `true` iff `this` specification is valid to define a ball
>   // constraint. A ball constraints specification is considered to be valid iff:
> ```

nit



_Suggestion:_
```objc
A ball constraint specification is considered to be valid iff:
```
___
*[`multibody/plant/multibody_plant.h` line 1281 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNS-RQ6Y6Nh0vJtIe9:-NJNS-RQ6Y6Nh0vJtIeA:bpp0dpm) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L1281)):*
> ```objc
>       double damping = 0.0);
> 
>   /// Defines a ball constraint between a point P on a body A and a point Q
> ```

BTW, should we define what a ball constraint is here?



_Code quote:_
```objc
a ball constraint
```
___
*[`multibody/plant/multibody_plant.h` line 1291 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNSNH37tw2v3aU6kwU:-NJNSNH37tw2v3aU6kwV:bu3t94) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L1291)):*
> ```objc
>   ///
>   /// @warning Currently, it is the user's responsibility to initialize the
>   /// model's context in a configuration compatible with the newly added
> ```

It's not clear what configuration is compatible with the constraint just from reading the doc.

Unlike the distance constraint, I don't think there's a invalid config for ball constraint. Sure one can add constraints so that they fight each other, but I wouldn't call that "incompatible".



_Code quote:_
```objc
  /// @warning Currently, it is the user's responsibility to initialize the
  /// model's context in a configuration compatible with the newly added
  /// constraint.
```
___
*[`multibody/plant/multibody_plant.h` line 5280 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNT4i-CmlG2L8-lHfE:-NJNT4i-CmlG2L8-lHfF:bl5xrz) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L5280)):*
> ```objc
>   std::vector<internal::DistanceConstraintSpecs> distance_constraints_specs_;
> 
>   // Vector of ball constraints specifications.
> ```

nit, grammar



_Suggestion:_
```objc
// Vector of ball constraint specifications.
```
___
*[`multibody/plant/multibody_plant.cc` line 512 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNTIcP7A5pUY62fsx9:-NJNTIcP7A5pUY62fsxA:b2ni8l0) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L512)):*
> ```cpp
>     const Body<T>& body_A, const Vector3<double>& p_AP, const Body<T>& body_B,
>     const Vector3<double>& p_BQ) {
>   // N.B. The manager is setup at Finalize() and therefore we must require
> ```

nit



_Suggestion:_
```cpp
  // N.B. The manager is set up at Finalize() and therefore we must require
  // constraints to be added pre-finalize.
```
___
*[`multibody/plant/multibody_plant.cc` line 518 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNTNN5DZWXv04nl-g2:-NJNTNN5DZWXv04nl-g3:b-s5y7zm) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L518)):*
> ```cpp
>   if (!is_discrete()) {
>         "Currently distance constraints are only supported for discrete "
> ```

nit copypasta



_Suggestion:_
```cpp
ball constraints
```
___
*[`multibody/plant/multibody_plant.cc` line 523 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNTQf2El0eG5smZRdO:-NJNTQf2El0eG5smZRdP:bb7phvt) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L523)):*
> ```cpp
> 
>   // TAMSI does not support distance constraints. For all other solvers, we let
>   // the discrete update manger throw an exception at finalize time.
> ```

typo



_Suggestion:_
```cpp
manager
```
___
*[`multibody/plant/multibody_plant.cc` line 533 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNTciRDyGnJ3hMtV-F:-NJNTciRDyGnJ3hMtV-G:bj4lqo7) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L533)):*
> ```cpp
>   }
> 
>   DRAKE_THROW_UNLESS(body_A.index() != body_B.index());
> ```

We already check this in `spec.IsValid()`. We could delete this and provide a more useful throw message below because there's only one reason that the constraint spec might not be valid .
___
*[`multibody/plant/sap_driver.cc` line 484 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNUZrJ6WcKCPXCjhYi:-NJNUZrJ6WcKCPXCjhYj:bqcitxb) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L484)):*
> ```cpp
>   // Stiffness and dissipation are set so that the constraint is in the
>   // "near-rigid" regime, [Castro et al., 2022].
>   const Vector3<T> stiffness(kInfinity, kInfinity, kInfinity);
> ```

What's the rational behind not allowing more compliant constraint as we do in the distance constraint case?
___
*[`multibody/plant/sap_driver.cc` line 504 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNVAO63DgsOVf6_qo8:-NJNVAO63DgsOVf6_qo9:b-iqevqy) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L504)):*
> ```cpp
>     const math::RigidTransform<T>& X_WB =
>         plant().EvalBodyPoseInWorld(context, body_B);
>     const Vector3<T>& p_WP = X_WA * specs.p_AP.cast<T>();
> ```

BTW, alternatively with less magic



_Suggestion:_
```cpp
    const Vector3<T> p_WP = X_WA * specs.p_AP.cast<T>();
    const Vector3<T> p_WQ = X_WB * specs.p_BQ.cast<T>();
```
___
*[`multibody/plant/sap_driver.cc` line 530 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNW9zO6FSivi_a-ZHj:-NJNW9zO6FSivi_a-ZHk:bw53v5e) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L530)):*
> ```cpp
> 
>     // Sanity check at least one body is not the world.
>     DRAKE_DEMAND(treeA_index.is_valid() || treeB_index.is_valid());
> ```

This check should probably be moved to the time when the constraint is added.
___
*[`multibody/plant/sap_driver.cc` line 550 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNYH2a1unQjIMnmz-u:-NJNYH2a1unQjIMnmz-v:bbpm58z) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L550)):*
> ```cpp
>     } else {
>       const MatrixX<T> JA = Jv_BqAp_W.middleCols(
>           tree_topology().tree_velocities_start(treeA_index),
> ```

This looks suspicious.

If A and B belong to the same tree, the jacobian would always be the same? I think this should instead be the following. Please double check.



_Suggestion:_
```cpp
      const MatrixX<T> JA = -Jv_WAp_W.middleCols(
          tree_topology().tree_velocities_start(treeA_index),
          tree_topology().num_tree_velocities(treeA_index));
```
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 72 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNY_7e0DJ8DMhLOeJG:-NJNY_7e0DJ8DMhLOeJH:bbdwg6y) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/test/sap_driver_ball_constraints_test.cc#L72)):*
> ```cpp
>             UnitInertia<double>::SolidSphere(radius) * mass);
> 
>     bodyA_ = &plant_.AddRigidBody("A", M_Bo);
> ```

How hard would it be to add a case where body A and body B are in the same tree? It should stress test the jacobian calculation a bit more.

xuchenhan-tri (commented):
I realized I still owe you the review for the tests. Sorry about the delay. I have now made a full pass.
The main concern is still the Jacobian calculation. Back to you now.

> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470#-:-NLStxPu7i7_m7JtKFsI:b-nmmj5k)* status: 19 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @joemasterjohn)

___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 104 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NLSsFHuC0DaPATk8eyX:-NLSsFHuC0DaPATk8eyY:b2yqywy) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/test/sap_driver_ball_constraints_test.cc#L104)):*
> ```cpp
> 
>   // Parameters of the problem.
>   Vector3d p_AP_{0.0, 0.0, 0.0};
> ```

nit, can we change these to non-zero and non-equal values?

amcastro-tri (commented):
pass by since I was cc'ed. Great work @joemasterjohn and thanks for the thorough review @xuchenhan-tri.

> Reviewed 6 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470#-:-NM4EIzqA2Fn2Merh7jN:be1mgrs)* status: 21 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @joemasterjohn and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 1281 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNS-RQ6Y6Nh0vJtIe9:-NM49oYC4rSV2kws88OL:b-wl7y9z) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L1281)):*

BTW, should we define what a ball constraint is here?
</blockquote></details>

Maybe you can paraphrase completely. My try: "Defines a constraint such that point P affixed to body A is coincident at all times with point Q affixed to body B, effectively modeling a ball-and-socket joint." 
___
*[`multibody/plant/sap_driver.cc` line 484 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNUZrJ6WcKCPXCjhYi:-NM4Bg6b6BEfPhL6z86C:b-b533iy) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L484)):*

What's the rational behind not allowing more compliant constraint as we do in the distance constraint case?
</blockquote></details>

I believe here we want do do the same we did for coupler constraints. The idea there was to simplify the setup for users, who most likely wanted to model a "hard" constraint. 
We did something special for distance constraints because our first users (Michael Posa's group at UPenn) already needed compiant distance constraint to model spring elements for Cassie.
___
*[`multibody/plant/sap_driver.cc` line 507 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NM4CmAN41PtQVDHnjYB:-NM4CmAN41PtQVDHnjYC:bunxfwd) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L507)):*
> ```cpp
>     const Vector3<T>& p_WQ = X_WB * specs.p_BQ.cast<T>();
> 
>     const Vector3<T> p_QP_W = p_WQ - p_WP;
> ```

this is actually reversed. This is the position vector from point P to Q, i.e. `p_PQ_W`
___
*[`multibody/plant/sap_driver.cc` line 517 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NM4DQAS5z_KxrezrON3:-NM4DQAS5z_KxrezrON4:b-3gvt3b) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L517)):*
> ```cpp
>         context, JacobianWrtVariable::kV, body_B.body_frame(), frame_W, p_WQ,
>         frame_W, frame_W, &Jv_WBq_W);
>     Jv_BqAp_W = (Jv_WBq_W - Jv_WAp_W);
> ```

nit, monogram notation also reversed here.
___
*[`multibody/plant/sap_driver.cc` line 550 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNYH2a1unQjIMnmz-u:-NM4E7zfB8cXj0zRJcYF:blvy482) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L550)):*

This looks suspicious.

If A and B belong to the same tree, the jacobian would always be the same? I think this should instead be the following. Please double check.
</blockquote></details>

I believe this is correct @xuchenhan-tri. Joe already knows he's got two different trees since withing this scope `single_tree = false`.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470)* status: 19 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @amcastro-tri and @joemasterjohn)

___
*[`multibody/plant/sap_driver.cc` line 484 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNUZrJ6WcKCPXCjhYi:-NM5m7h5BFaC9jRhSIyn:b-a7jufr) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L484)):*

I believe here we want do do the same we did for coupler constraints. The idea there was to simplify the setup for users, who most likely wanted to model a "hard" constraint. 
We did something special for distance constraints because our first users (Michael Posa's group at UPenn) already needed compiant distance constraint to model spring elements for Cassie.
</blockquote></details>

I see. Thanks for the explanation.
___
*[`multibody/plant/sap_driver.cc` line 550 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNYH2a1unQjIMnmz-u:-NM5nuIB2jO6X294h8Ok:b-lkknc3) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L550)):*

I believe this is correct @xuchenhan-tri. Joe already knows he's got two different trees since withing this scope `single_tree = false`.
</blockquote></details>

Sorry, I had somehow missed that! Retracted.

joemasterjohn (commented):
Thank you both for the reviews. Everything addressed, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-:-NM5nz0YAhLEVNsJovkD:b-jeo9q7)* status: 7 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @amcastro-tri and @xuchenhan-tri)

___
*[`multibody/plant/constraint_specs.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNXUHdAG5LELx1zo1_:-NJxJy7t9pxJ9ES0S5VZ:b-o3s9mv) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L71)):*

BTW, can I persuade you to do `p_PQ_W` instead of `p_QP_W`. This makes the sign convention the same as in contact and distance constraints and creates less mental dissonance (at least for myself).
</blockquote></details>

Done. Sure thing, keeping the sign convention with the existing constraints is a strong enough argument.
___
*[`multibody/plant/constraint_specs.h` line 72 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNQZmN7TwAWnXWCly9:-NM4bdbHEaCIfVQfL6XN:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L72)):*

nit probably better to say "position" instead of "pose" as there's no orientation involved.

Also, weird line break.
</blockquote></details>

Done.
___
*[`multibody/plant/constraint_specs.h` line 73 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNRHTlDomX6EtA2TOt:-NM4dzpNBPvq_mILS2zS:bteix6e) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L73)):*

Probably worth saying a few words about what P and Q are. I'm assuming the ball constraint imposed is forcing them to be coincident. If so, please make that explicit.
</blockquote></details>

Done. Reworded the comment based on your suggestion.
___
*[`multibody/plant/constraint_specs.h` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNQywBD4IoOFBhXQtP:-NM4eA0yE6nWaEB6JuPY:b9h7blj) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L78)):*

BTW, why do we name these struct `FooSpecs` instead of `FooSpec`? Each struct represents a single specification.
</blockquote></details>

I had a little cognitive dissonance grammar-wise as well, but I'm just following the previous convention here. We can follow up with a PR that just changes names though. 
___
*[`multibody/plant/constraint_specs.h` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNRc4d5xcqfOqxhuHD:-NM4ecAe9uTmZBtZ3Nh9:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L80)):*

nit
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.h` line 1281 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNS-RQ6Y6Nh0vJtIe9:-NM4essV2dWdoE_ydcFA:bpavnb6) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L1281)):*

Maybe you can paraphrase completely. My try: "Defines a constraint such that point P affixed to body A is coincident at all times with point Q affixed to body B, effectively modeling a ball-and-socket joint." 
</blockquote></details>

Done. Thanks for the suggestion Alejandro.
___
*[`multibody/plant/multibody_plant.h` line 1291 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNSNH37tw2v3aU6kwU:-NM4fwXE84YS1md9LBGB:b-64xewc) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L1291)):*

It's not clear what configuration is compatible with the constraint just from reading the doc.

Unlike the distance constraint, I don't think there's a invalid config for ball constraint. Sure one can add constraints so that they fight each other, but I wouldn't call that "incompatible".
</blockquote></details>

True. I think I misinterpreted this comment when copy/pasting it to mean "we don't guarantee that the default configuration in a default context satisfies the constraint. It is the user's responsibility to set the configuration in the context such that the constraint is satisfied." But it's talking about the invalid config where the Jacobian of the distance constraint blows up.

I'll remove the comment here since all configurations are valid for the ball constraint.
___
*[`multibody/plant/multibody_plant.h` line 5280 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNT4i-CmlG2L8-lHfE:-NM4mKWe3B60GOksPETk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.h#L5280)):*

nit, grammar
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 533 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNTciRDyGnJ3hMtV-F:-NM4nc-i82ILFmYvffoX:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/multibody_plant.cc#L533)):*

We already check this in `spec.IsValid()`. We could delete this and provide a more useful throw message below because there's only one reason that the constraint spec might not be valid .
</blockquote></details>

Done. Good call.
___
*[`multibody/plant/sap_driver.cc` line 484 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNUZrJ6WcKCPXCjhYi:-NM5GSUdARAr3fgZPByw:b-pgstgf) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L484)):*

I believe here we want do do the same we did for coupler constraints. The idea there was to simplify the setup for users, who most likely wanted to model a "hard" constraint. 
We did something special for distance constraints because our first users (Michael Posa's group at UPenn) already needed compiant distance constraint to model spring elements for Cassie.
</blockquote></details>

Yes, the suggestion came from Alejandro to simplify to just offer a hard constraint. 
___
*[`multibody/plant/sap_driver.cc` line 504 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNVAO63DgsOVf6_qo8:-NM4nr-J-oE7iLBbix3S:boyexbg) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L504)):*

BTW, alternatively with less magic
</blockquote></details>

Done. Thanks.
___
*[`multibody/plant/sap_driver.cc` line 507 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM4CmAN41PtQVDHnjYB:-NM4nydo9r90nVk9Lko_:bb13ws) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L507)):*

this is actually reversed. This is the position vector from point P to Q, i.e. `p_PQ_W`
</blockquote></details>

Done. Thanks for catching this. I believe the rest of the math is correct and consistent but I just have the monogram notation flipped.
___
*[`multibody/plant/sap_driver.cc` line 517 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM4DQAS5z_KxrezrON3:-NM4oJAo-_qF6325ZlxH:boyexbg) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L517)):*

nit, monogram notation also reversed here.
</blockquote></details>

Done. Thanks.
___
*[`multibody/plant/sap_driver.cc` line 530 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNW9zO6FSivi_a-ZHj:-NM58mWvEqUiicY8nEoR:b-8jlzzz) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L530)):*

This check should probably be moved to the time when the constraint is added.
</blockquote></details>

Actually this condition is already satisfied implicitly by `body_A != body_B`, checked in `spec.IsValid()`. So this sanity check isn't even needed.
___
*[`multibody/plant/sap_driver.cc` line 550 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNYH2a1unQjIMnmz-u:-NM5F6Cs4sjGEsL_bSqb:bt8a0e) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/sap_driver.cc#L550)):*

I believe this is correct @xuchenhan-tri. Joe already knows he's got two different trees since withing this scope `single_tree = false`.
</blockquote></details>

Like Alejandro said this case deals with two different trees. And in this case:
```
Jv_BqAp_W.middleCols(
          tree_topology().tree_velocities_start(treeA_index),
          tree_topology().num_tree_velocities(treeA_index));
```
is equivalent to:
```
-Jv_WAp_W.middleCols(
          tree_topology().tree_velocities_start(treeA_index),
          tree_topology().num_tree_velocities(treeA_index));
```
because the non-zero portions of `Jv_WBq_W` only occur in indices belonging to treeB, which don't overlap those of treeA.
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 72 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNY_7e0DJ8DMhLOeJG:-NM5lwQ-Cqhyp9NhPPyL:b-kgeglm) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/test/sap_driver_ball_constraints_test.cc#L72)):*

How hard would it be to add a case where body A and body B are in the same tree? It should stress test the jacobian calculation a bit more.
</blockquote></details>

Not too hard, but it would just shuffle the values around in the non-sparse jacobian really. The portion for A and B would then be organized adjacently. Check out the modification I made to the test though to stress test it a bit more.
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 104 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NLSsFHuC0DaPATk8eyX:-NM5GMeI3-bP2ZQ4yHF6:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/test/sap_driver_ball_constraints_test.cc#L104)):*

nit, can we change these to non-zero and non-equal values?
</blockquote></details>

Done.

xuchenhan-tri (commented):
Feature :lgtm: +@sherm1 for platform review per rotation please.
BTW this PR still needs a rebase.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-:-NM5pzjA5vsvsMv9yEDC:b-yf7sm2)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform) (waiting on @amcastro-tri)

___
*[`multibody/plant/constraint_specs.h` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NJNQywBD4IoOFBhXQtP:-NM5oSjh-_XjogP8uR56:bigavl9) ([raw file](https://github.com/RobotLocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L78)):*

I had a little cognitive dissonance grammar-wise as well, but I'm just following the previous convention here. We can follow up with a PR that just changes names though. 
</blockquote></details>

Sounds good.
cc the original author @amcastro-tri.

sherm1 (commented):
This is very nice! Platform :lgtm_strong: with a few comments

> Reviewed 5 of 10 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-:-NM6WHUv-8SNe2YiFuLB:b-1g00bd)* status: 10 unresolved discussions (waiting on @amcastro-tri and @joemasterjohn)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6J_DrA72S0DpEcAeO):*
FYI labeling the constraint by the number of dofs only works well for ball constraints!  More generally it's better to describe them by the number of constraint equations they introduce (also 3 here so it doesn't matter). A distance constraint (for example) permits 5 dofs because it only introduces one constraint equation but describing it as a "5 dof constraint" doesn't seem helpful. A constraint that produces the effect of a revolute joint introduces 5 constraint equations, leaving 1 dof if the constrained pair had 6 to start with.

In general there can be redundant constraint equations in a system in which case there are fewer restrictions on motion than would be expected. So you can always say how many equations there are, but it's hard to say how many dofs are left.
___
*[`multibody/plant/multibody_plant.cc` line 541 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6MS6tEQNBGUcCooYu:-NM6MS6tEQNBGUcCooYv:bksie4z) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/multibody_plant.cc#L541)):*
> ```cpp
>         "bodies, i.e. body_A != body_B must be satisfied.",
>         body_A.name(), body_B.name());
>     throw std::runtime_error(msg);
> ```

BTW I would call that an "std::logic_error" since the programmer who made this illegitimate call could have avoided it (fixing is optional since we only promise that an std::exception of some kind will be thrown)
___
*[`multibody/plant/sap_driver.h` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6NUgE8v6-P8na9Mj0:-NM6NUgE8v6-P8na9Mj1:brnv1gx) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.h#L147)):*
nit: consider saying it this way:
> Adds holonomic constraint _equations_ to model ball constraints ...

That is, avoid using "constraint" to mean both a single constraint equation and the Constraint model element that can be a collection of equations. My preference would be to capitalize Constraint whenever we're referring to one of the model elements but I think that horse is out of the barn :(
___
*[`multibody/plant/sap_driver.cc` line 531 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6PM9i4qFuoHu3Arda:-NM6PM9i4qFuoHu3Ardb:bolbzke) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L531)):*
> ```cpp
>                              !treeB_index.is_valid() ||
>                              treeA_index == treeB_index;
> ```

BTW consider asserting that at least one of the tree indexes is valid.
___
*[`multibody/plant/sap_driver.cc` line 537 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Po1HDi1NsbNraBg8:-NM6Po1HDi1NsbNraBg9:be6mr9k) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L537)):*
> ```cpp
>       const TreeIndex tree_index =
>           treeA_index.is_valid() ? treeA_index : treeB_index;
>       const MatrixX<T> J = Jv_ApBq_W.middleCols(
> ```

BTW it seems a shame that we have to make a copy of the "middle cols" of this matrix. Seems like the block returned by middleCols() could be used as-is. That's a bigger issue than just this constraint though. WDYT @amcastro-tri ? Maybe a TODO here or elsewhere?
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 180 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6U88A3-Psuy_MdzNz:-NM6U88A3-Psuy_MdzO-:bag4kmc) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L180)):*
> ```cpp
>   // We exploit internal knowledge of the fact that the state stores angular
>   // velocities first, followed by translational velocities (and the
>   // relative positions of the bodies).
> ```

BTW I couldn't quite follow the above paragraph. It's right that the generalized velocities for a floating joint are as described. But what do you mean by the "relative positions" of the bodies? Can you make the comment clearer?
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 182 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6UcoO9qy5x9GY5c2w:-NM6UcoO9qy5x9GY5c2x:bsnhkuj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L182)):*
> ```cpp
>   // relative positions of the bodies).
> 
>   // We know by construction that the 6 generalized coordinates for a floating
> ```

nit: coordinates -> velocities
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Vijz9ueE1-Er-Wpq:-NM6Vijz9ueE1-Er-Wpr:b8c5c1a) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L224)):*
> ```cpp
>   }
> }
> ```

BTW is there any reasonable way to test that the constraint produces the right reaction forces? It would be easy to set up -- e.g. apply a body force at point Q for the 1-clique case and verify that the reaction force in the constraint is exactly opposite that force. Optional since I presume a properly set-up constraint in SAP should always work if SAP does, but it would be reassuring!
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 237 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Rks_A91pptOSX-un:-NM6Rksa2ySAxrKcm2XH:b-lwrxp7) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L237)):*
> ```cpp
> 
> // TODO(joemasterjohn): implement unit tests verifying:
> //  - attempt to use ball constraint with continuous system or TAMSI
> ```

BTW reminder to add these (small) tests. I'd suggest adding them to this PR so the whole thing is done at merge

sherm1 (commented):
(Needs a manual rebase to fix merge conflicts)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-:-NM6WPSd4OGmFScNVO5u:b-rk111j)* status: 10 unresolved discussions (waiting on @amcastro-tri and @joemasterjohn)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470)* status: 9 unresolved discussions (waiting on @joemasterjohn and @sherm1)

___
*[`multibody/plant/constraint_specs.h` line 78 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NJNQywBD4IoOFBhXQtP:-NMALVA126TNlma4JPna:b-1v98r8) ([raw file](https://github.com/robotlocomotion/drake/blob/28c38d6bb7aff25bbf60b4b35fef8b232085674a/multibody/plant/constraint_specs.h#L78)):*

Sounds good.
cc the original author @amcastro-tri.
</blockquote></details>

In my mind each of the fields in these structs are a single spec and therefore why I used plural.
That being said, I am probably the last person to ask about grammar though. Feel free to make the update to singular.
___
*[`multibody/plant/sap_driver.cc` line 537 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NM6Po1HDi1NsbNraBg8:-NMAMM-_0DHw-goPbgN8:b-rjjr4d) ([raw file](https://github.com/robotlocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L537)):*

BTW it seems a shame that we have to make a copy of the "middle cols" of this matrix. Seems like the block returned by middleCols() could be used as-is. That's a bigger issue than just this constraint though. WDYT @amcastro-tri ? Maybe a TODO here or elsewhere?
</blockquote></details>

maybe. My impression is that the cost of these unavoidable allocations (someone must allocate this block at some point) is negligible. We'll know more after our campaign to optimize/profile the multibody engine this year.
Now, what we can do, is to avoid allocating more than once. In this case, I believe we could `std::move`  `J` into `SapHolonomicConstraint`'s constructor.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18470)* status: 9 unresolved discussions (waiting on @joemasterjohn)

___
*[`multibody/plant/sap_driver.cc` line 537 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18470#-NM6Po1HDi1NsbNraBg8:-NMAXpvLADT8b334ukBG:b-76qph4) ([raw file](https://github.com/robotlocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L537)):*

maybe. My impression is that the cost of these unavoidable allocations (someone must allocate this block at some point) is negligible. We'll know more after our campaign to optimize/profile the multibody engine this year.
Now, what we can do, is to avoid allocating more than once. In this case, I believe we could `std::move`  `J` into `SapHolonomicConstraint`'s constructor.
</blockquote></details>

Oh, yes definitely J should be std::move'd there right now rather than making yet another copy! (Those parameter are taken by value @joemasterjohn)

But that block is already allocated within Jv_ApBq_W as a contiguous set of columns. That could easily be read as-is without copying. Why does it have to be allocated and copied at all? Is that because SapHolonomicConstraint needs a matrix it can stomp on in place?

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18470)* status: 4 unresolved discussions (waiting on @amcastro-tri, @sherm1, and @xuchenhan-tri)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6J_DrA72S0DpEcAeO):*

FYI labeling the constraint by the number of dofs only works well for ball constraints!  More generally it's better to describe them by the number of constraint equations they introduce (also 3 here so it doesn't matter). A distance constraint (for example) permits 5 dofs because it only introduces one constraint equation but describing it as a "5 dof constraint" doesn't seem helpful. A constraint that produces the effect of a revolute joint introduces 5 constraint equations, leaving 1 dof if the constrained pair had 6 to start with.

In general there can be redundant constraint equations in a system in which case there are fewer restrictions on motion than would be expected. So you can always say how many equations there are, but it's hard to say how many dofs are left.
</blockquote></details>

Ah that's true, and also recently came up when computing the number of constrained dofs on the strandbeest. I'll change the language everywhere to refer to num constraint equations rather than num dofs.
___
*[`multibody/plant/multibody_plant.cc` line 541 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6MS6tEQNBGUcCooYu:-NMZ_lQGBSueJ87VVroa:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/multibody_plant.cc#L541)):*

BTW I would call that an "std::logic_error" since the programmer who made this illegitimate call could have avoided it (fixing is optional since we only promise that an std::exception of some kind will be thrown)
</blockquote></details>

Done.
___
*[`multibody/plant/sap_driver.h` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6NUgE8v6-P8na9Mj0:-NMZa8je6zkIfA8H4aF-:b5i2r74) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.h#L147)):*

nit: consider saying it this way:
> Adds holonomic constraint _equations_ to model ball constraints ...

That is, avoid using "constraint" to mean both a single constraint equation and the Constraint model element that can be a collection of equations. My preference would be to capitalize Constraint whenever we're referring to one of the model elements but I think that horse is out of the barn :(
</blockquote></details>

Done. I believe all the existing APIs and data structures are in line with your preference. We could do a cleanup PR to standardize comments and variable names.
___
*[`multibody/plant/sap_driver.cc` line 531 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6PM9i4qFuoHu3Arda:-NM_7hGYB2xRYsggsl66:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L531)):*

BTW consider asserting that at least one of the tree indexes is valid.
</blockquote></details>

Done.
___
*[`multibody/plant/sap_driver.cc` line 537 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Po1HDi1NsbNraBg8:-NM_GDBJ9XneVan8unZU:b-393m6q) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/sap_driver.cc#L537)):*

Oh, yes definitely J should be std::move'd there right now rather than making yet another copy! (Those parameter are taken by value @joemasterjohn)

But that block is already allocated within Jv_ApBq_W as a contiguous set of columns. That could easily be read as-is without copying. Why does it have to be allocated and copied at all? Is that because SapHolonomicConstraint needs a matrix it can stomp on in place?
</blockquote></details>

OK I've std::move'd them into the constraint's constructor. The lifetime of `Jv_ApBq` is this function, so at least one copy has to happen for the constraint to keep the data around. If I std::move here (and from `SapHolonomicConstraint`'s constructor to `SapConstraint`'s constructor) then we've done the minimal amount of copying, I think?
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 180 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6U88A3-Psuy_MdzNz:-NM_7w1N1BJtw6KyPE1_:b-m7kmpp) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L180)):*

BTW I couldn't quite follow the above paragraph. It's right that the generalized velocities for a floating joint are as described. But what do you mean by the "relative positions" of the bodies? Can you make the comment clearer?
</blockquote></details>

Sorry, that was a leftover copy/paste. Removed and clarified in the comment below.
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 182 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6UcoO9qy5x9GY5c2w:-NM_8Grb47omKBidIvyP:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L182)):*

nit: coordinates -> velocities
</blockquote></details>

Done.
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Vijz9ueE1-Er-Wpq:-NM_DIfr7ole6pSG9mBU:bcipwfw) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L224)):*

BTW is there any reasonable way to test that the constraint produces the right reaction forces? It would be easy to set up -- e.g. apply a body force at point Q for the 1-clique case and verify that the reaction force in the constraint is exactly opposite that force. Optional since I presume a properly set-up constraint in SAP should always work if SAP does, but it would be reassuring!
</blockquote></details>

Going to punt on this one. 
___
*[`multibody/plant/test/sap_driver_ball_constraints_test.cc` line 237 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18470#-NM6Rks_A91pptOSX-un:-NM_AcZD-MJWK9zYIL-U:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9767e6d60ea7de0c6847656b3ac968ad718e45bb/multibody/plant/test/sap_driver_ball_constraints_test.cc#L237)):*

BTW reminder to add these (small) tests. I'd suggest adding them to this PR so the whole thing is done at merge
</blockquote></details>

Done.

PR #18471, opened 2022-12-15
When we document "_absolute_ file path", enforce it

ggould-tri:
Glory in the number of these diff lines where the word "absolute" appears within a hundred characters of a relative path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18471)

### DIFF
--- a/bindings/pydrake/systems/test/planar_scenegraph_visualizer_test.py
+++ b/bindings/pydrake/systems/test/planar_scenegraph_visualizer_test.py
@@ -176,13 +176,13 @@ class TestPlanarSceneGraphVisualizer(unittest.TestCase):
                 scene_graph, substitute_collocated_mesh_files=False)
 
         # This should report that the file does not exist.
-        scene_graph = scene_graph_with_mesh("garbage.obj")
+        scene_graph = scene_graph_with_mesh("/garbage.obj")
         with self.assertRaises(FileNotFoundError):
             PlanarSceneGraphVisualizer(scene_graph)
 
         # This should report that the extension was wrong and no .obj was
         # found.
-        scene_graph = scene_graph_with_mesh("garbage.STL")
+        scene_graph = scene_graph_with_mesh("/garbage.STL")
         with self.assertRaises(RuntimeError):
             PlanarSceneGraphVisualizer(scene_graph)
 
--- a/bindings/pydrake/test/geometry_common_test.py
+++ b/bindings/pydrake/test/geometry_common_test.py
@@ -323,8 +323,8 @@ class TestGeometryCore(unittest.TestCase):
             mut.Capsule(radius=1.0, length=2.0),
             mut.Ellipsoid(a=1.0, b=2.0, c=3.0),
             mut.HalfSpace(),
-            mut.Mesh(absolute_filename="arbitrary/path", scale=1.0),
-            mut.Convex(absolute_filename="arbitrary/path", scale=1.0),
+            mut.Mesh(absolute_filename="/arbitrary/path", scale=1.0),
+            mut.Convex(absolute_filename="/arbitrary/path", scale=1.0),
             mut.MeshcatCone(height=1.23, a=3.45, b=6.78)
         ]
         for shape in shapes:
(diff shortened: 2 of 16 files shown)


ggould-tri (commented):
+@seancurtis-tri

I suspected this (incorrectly?) in a bug downstream in anzu and spent too long chasing it in drake; better to enforce the right answer where we document it.

FYI @rpoyner-tri as we were discussing :-)

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18471#-:-NJN4puDAdKklL7K7Rr8:b-digax4)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
First pass done.

You'll see my notes below on the efficacy of the current approach. It doesn't quite do what it intends.

I wasn't privy to the conversation and the bug so I'm about to propose an alternate "solution" which may not actually fit the bill:

Remove the documented requirement that the path be absolute. Change it to saying that if the path isn't absolute already, it will be *made* absolute based on the current working directory. At least then we'll always be passing around absolute paths that can be recognized as such -- i.e., you can see the whole path and think, "oops! I forgot to send in an absolute path". This wouldn't be a "breaking" change because every previous absolute path will continue working unchanged, and any relative path will simply be made more explicit. The failure still comes where we attempt to *read* the file, but instead of the message, "Can't load 'foo.txt'" we'll get the message, "Can't load '/This/can't/possibly/be/the/right/path/foo.txt'."

That can be achieved simply by using `std::filesystem::absolute()` on the input path blindly and none of the tests have to change, etc.

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18471#-:-NJQR_je0KYFenn_y78Y:bhkfy04)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @ggould-tri)

___
*[`geometry/shape_specification.cc` line 78 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18471#-NJQLy1p6b50f5aUlpHZ:-NJQLy1p6b50f5aUlpH_:b893vvn) ([raw file](https://github.com/robotlocomotion/drake/blob/342f6d0a37f6d2c1187b9c84c4e9f263c70ad61c/geometry/shape_specification.cc#L78)):*
> ```cpp
>     : Shape(ShapeTag<Convex>()), filename_(absolute_filename), scale_(scale) {
>   auto path = std::filesystem::path(absolute_filename);
>   if (path.is_relative()) {
> ```

BTW When I first read this, I was unsure if `path.is_relative()` was synonymous with `!path.is_absolute()` (my mind was imagining file types for which, perhaps, that might not be the case). I looked at documentation and believe it is. Perhaps it would help future readers if this didn't rely on that level of indirection and simply said what we're interested in?
___
*[`geometry/shape_specification.cc` line 83 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18471#-NJQKbXaEFqr6VpCvzzb:-NJQKbXaEFqr6VpCvzzc:bgp84ij) ([raw file](https://github.com/robotlocomotion/drake/blob/342f6d0a37f6d2c1187b9c84c4e9f263c70ad61c/geometry/shape_specification.cc#L83)):*
> ```cpp
>       absolute_filename));
>   }
>   if (!path.has_filename()) {
> ```

I don't think this is the test you want. For example:

```c++
  fs::path p("/test/..");
  std::cout << std::boolalpha;
  std::cout << p << "\n"
	    << "  is_absolute: " << p.is_absolute() << "\n"
	    << "  has_filename: " << p.has_filename() << "\n";
```

produces the following output:

```
"/test/.."
  is_absolute: true
  has_filename: true
```

I fear determining the provided path is both absolute and a file is not quite straightforward. It also raises the question of "what about sockets?" Or other esoteric file types whose exact semantics escape me.

I suspect we'll be better off making this a free function with a return enumeration such that these two call sites can still differentiate error messages. But it'll allow us to isolate and document the weirdness required to accomplish this.

jwnimmer-tri:
If I understand correctly, the approach in #18536 was chosen instead and we should close this PR?

SeanCurtis-TRI:
That's my understanding as well.

Issue #18334, opened 2022-11-16
Auto-reload in ModelVisualizer


jwnimmer-tri:
I common use case for ModelVisualizer is creating a new model from scratch.

When doing that, the user will typically
- edit their model file(s) in their IDE, then
- re-launch the ModelVisualizer to see the effect, then
- close the visualizer, then repeat.

It would help speed up their workflow if they didn't need to re-launch the visualizer every time.  Ideally, it would monitor the input file(s) for changes, and if there was a change either (1) automatically reload the model, or (2) provide a "reload" button for the user to click at their leisure.  If we do (2), maybe the button could always be there (i.e., without trying to detect changes).

SeanCurtis-TRI:
With the presence of the button, it would be *nice* if the visualizer reported that there *was* a change or not. Updating files in the wrong directory is a common enough mistake, it would be nice to help users realize that they don't see any changes because the visualizer says there *aren't* any changes.

trowell-tri:
That's a good idea. We should have some kind of error reporting anyway, for malformed model files (maybe it already does this, I haven't tried) so this is just an extension of that.

jwnimmer-tri:
At the moment, there is no way to ask the Parser "which files did you load?".  (Recall that files can include other files or meshes, so monitoring only the one filename we pass to the parser is _not_ sufficient on its own.)  We'd need to add that feature before ModelVisualizer could report whether or not anything changed.

trowell-tri:
I've added an optional (for now, via a command-line switch) reload button so we can try this out and discuss future directions.

jwnimmer-tri:
We discussed this a bit today.

My decision is that for now, we'll de-scope "auto" as the feature.  I think the "Reload" button is very useful on its own, and should suit this use case OK.

Trying to instrument "auto" would mean that the Parser needs to reach down into the SceneGraph to understand which exact mesh / material / texture files it loaded.  That level of implementation cost doesn't seem worth the benefit of the "auto" feature.

PR #18488, opened 2022-12-20
Enable manual model reloading in ModelVisualizer

trowell-tri:
Adds a Reload Model File button to the ModelVisualizer control panel to allow live model editing and re-visualization. Also improves error feedback when the number of position arguments does not match that in the model, and normalizes some docstring style.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)

### DIFF
--- a/bindings/pydrake/visualization/_model_visualizer.py
+++ b/bindings/pydrake/visualization/_model_visualizer.py
@@ -1,3 +1,4 @@
+from enum import Enum
 import time
 import webbrowser
 
@@ -7,9 +8,6 @@ from pydrake.geometry import (
     Cylinder,
     GeometryInstance,
     MakePhongIllustrationProperties,
-    MeshcatVisualizer,
-    MeshcatVisualizerParams,
-    Role,
     StartMeshcat,
 )
 from pydrake.math import RigidTransform, RotationMatrix
@@ -25,6 +23,14 @@ from pydrake.visualization import (
     VisualizationConfig,
     ApplyVisualizationConfig,
 )
+
+
+class RunResult(Enum):
+    """Result code indicating how Run() terminated."""
+    KEEP_GOING = 0  # Note that this is never returned but is used internally.
+    LOOP_ONCE = 1
+    STOPPED = 2
+    RELOAD = 3
 
 
 class ModelVisualizer:
@@ -86,6 +92,10 @@ class ModelVisualizer:
         self._browser_new = browser_new
         self._pyplot = pyplot
         self._meshcat = meshcat
+
+        # This is both a button label and a flag to enable the button.
+        self._reload_button_name = None
+        self._model_filenames = list()
 
         # The following fields remain valid for this object's lifetime after
         # Finalize() has been called.
@@ -151,6 +161,7 @@ class ModelVisualizer:
           filename: the name of a file containing one or more models.
         """
         self._parser.AddModels(filename)
+        self._model_filenames.append(filename)
 
     def Finalize(self, position=None):
         """
@@ -197,6 +208,11 @@ class ModelVisualizer:
         self._meshcat.Delete()
         self._meshcat.DeleteAddedControls()
 
+        # We want to place this button far away from the Stop Running button,
+        # hence the work to do this here.
+        if self._reload_button_name:
+            self._meshcat.AddButton(self._reload_button_name)
+
         # Connect to drake_visualizer, meldis, and meshcat.
         # Meldis and meshcat provide simultaneous visualization of
         # illustration and proximity geometry.
@@ -214,10 +230,6 @@ class ModelVisualizer:
             "joint_sliders", JointSliders(meshcat=self._meshcat,
                                           plant=self._plant))
 
-        if self._browser_new:
-            url = self._meshcat.web_url()
-            webbrowser.open(url=url, new=self._browser_new)
-
         # Connect to PyPlot.
         if self._pyplot:
             ConnectPlanarSceneGraphVisualizer(self._builder, self._scene_graph)
@@ -226,9 +238,16 @@ class ModelVisualizer:
         self._context = self._diagram.CreateDefaultContext()
         self._plant_context = self._plant.GetMyContextFromRoot(self._context)
 
-        if position:
+        # We don't just test 'position' because NumPy does weird things with
+        # the truth values of arrays.
+        if position is not None and len(position) > 0:
+            self._raise_if_invalid_positions(position)
             self._plant.SetPositions(self._plant_context, position)
             self._sliders.SetPositions(position)
+
+        if self._browser_new:
+            url = self._meshcat.web_url()
+            webbrowser.open(url=url, new=self._browser_new)
 
         # Use Simulator to dispatch initialization events.
         # TODO(eric.cousineau): Simplify as part of #13776 (was #10015).
@@ -258,10 +277,13 @@ class ModelVisualizer:
             7 positions corresponding to any model bases that aren't welded to
             the world.
           loop_once: a flag that exits the evaluation loop after one pass.
+
+        Returns:
+          A RunResult code indicating why the method stopped.
         """
         if self._builder:
             self.Finalize(position=position)
-        elif position is not None:
+        elif position is not None and len(position) > 0:
             self._plant.SetPositions(self._plant_context, position)
             self._sliders.SetPositions(position)
             self._diagram.ForcedPublish(self._context)
@@ -279,25 +301,143 @@ class ModelVisualizer:
             self._meshcat)])
 
         # Wait for the user to cancel us.
-        button_name = "Stop Running"
+        stop_button_name = "Stop Running"
         if not loop_once:
-            print(f"Click '{button_name}' or press Esc to quit")
+            print(f"Click '{stop_button_name}' or press Esc to quit")
 
+        loop_result = RunResult.KEEP_GOING
         try:
-            self._meshcat.AddButton(button_name, "Escape")
+            self._meshcat.AddButton(stop_button_name, "Escape")
+
+            def has_clicks(button_name):
+                if not button_name:
+                    return False
+                return self._meshcat.GetButtonClicks(button_name) > 0
 
             sliders_context = self._sliders.GetMyContextFromRoot(self._context)
-            while True:
+            while loop_result == RunResult.KEEP_GOING:
                 time.sleep(1 / 32.0)
                 q = self._sliders.get_output_port().Eval(sliders_context)
                 self._plant.SetPositions(self._plant_context, q)
                 self._diagram.ForcedPublish(self._context)
-                if loop_once or self._meshcat.GetButtonClicks(button_name) > 0:
-                    return
+                if loop_once:
+                    loop_result = RunResult.LOOP_ONCE
+                if has_clicks(stop_button_name):
+                    loop_result = RunResult.STOPPED
+                if has_clicks(self._reload_button_name):
+                    loop_result = RunResult.RELOAD
         except KeyboardInterrupt:
-            pass
+            loop_result = RunResult.STOPPED
         finally:
-            self._meshcat.DeleteButton(button_name)
+            self._meshcat.DeleteButton(stop_button_name)
+            if self._reload_button_name:
+                self._meshcat.DeleteButton(self._reload_button_name)
+
+        return loop_result
+
+    def RunWithReload(self, position=None):
+        """
+        Runs the model with an option to reload.
+        Neither Finalize() nor Run() may be called before this method.
+
+        Will iterate until the user clicks the Stop Running button. All model
+        files will be reloaded and meshcat will be reset if the reload button
+        is clicked. If pyplot is being used, clicking reload will create a
+        new pyplot window.
+
+        Only models added using AddModels() will be properly reloaded.
+        If the parser() was accessed to add models directly to the parser
+        then those models will not be properly reloaded by this method.
+
+        Args:
+          position: an ndarray-like list of slider positions for the model(s);
+            must match the number of positions in all model(s), including the
+            7 positions corresponding to any model bases that aren't welded to
+            the world.
+        """
+        # This method uses this outer visualizer (self) to track options
+        # and manage meshcat (and its window, if applicable). We then
+        # create an inner visualizer each time through the reload loop.
+
+        # This outer visualizer must not have been Finalized or Run.
+        if self._builder is None:
+            raise RuntimeError("Finalize has already been called.")
+        assert all([x is not None for x in (
+            self._plant,
+            self._scene_graph,
+            self._parser)])
+        assert all([x is None for x in (
+            self._diagram,
+            self._sliders,
+            self._context,
+            self._plant_context)])
+
+        # We don't just test 'position' because NumPy does weird things with
+        # the truth values of arrays.
+        if position is not None and len(position) > 0:
+            # If we have a position list to validate, finalize the outer
+            # visualizer's plant only so we can validate positions.
+            self._plant.Finalize()
+            self._raise_if_invalid_positions(position)
+
+        if self._browser_new:
+            url = self.meshcat().web_url()
+            webbrowser.open(url=url, new=self._browser_new)
+
+        model_suffix = "s" if len(self._model_filenames) != 1 else ""
+        reload_button_name = "Reload Model File%s" % model_suffix
+
+        while True:
+            inner_visualizer = ModelVisualizer(
+                visualize_frames=self._visualize_frames,
+                triad_length=self._triad_length,
+                triad_radius=self._triad_radius,
+                triad_opacity=self._triad_opacity,
+                browser_new=False,
+                pyplot=self._pyplot,
+                meshcat=self._meshcat)
+
+            inner_visualizer._reload_button_name = reload_button_name
+            for filename in self._model_filenames:
+                inner_visualizer.AddModels(filename)
+
+            # Finalize the inner visualizer outside of Run so that we can
+            # make sure the positions we saved from the prior iteration
+            # are still usable with the reloaded model. This is necessary
+            # so that changes in the number of joints doesn't just exit.
+            inner_visualizer.Finalize()
+
+            # We don't just test 'position' because NumPy does weird things
+            # with the truth values of arrays.
+            if position is not None and len(position) > 0:
+                # Don't exit if the number of positions in the model changes.
+                # TODO(trowell-tri) Examine the plant more thoroughly
+                # and properly restore position values to the same joints.
+                if inner_visualizer._plant.num_positions() != len(position):
+                    position = None
+
+            result = inner_visualizer.Run(position)
+            if result != RunResult.RELOAD:
+                return
+
+            position = inner_visualizer._sliders.get_output_port().Eval(
+                inner_visualizer._sliders.GetMyContextFromRoot(
+                    inner_visualizer._context))
+
+    def _raise_if_invalid_positions(self, position):
+        """
+        Validate the position argument.
+
+        Raises:
+          ValueError: if the length of the position list does not match
+        the number of positions in the plant.
+        """
+        assert self._plant is not None
+        if self._plant.num_positions() != len(position):
+            message = ("Number of passed positions (%d) does not match "
+                       "the number in the model (%d).")
+            raise ValueError(
+                message % (len(position), self._plant.num_positions()))
 
     def _add_triad(
         self,
@@ -315,7 +455,7 @@ class ModelVisualizer:
         The axes point in +x, +y and +z directions, respectively.
         Based on [code permalink](https://github.com/RussTedrake/manipulation/blob/5e59811/manipulation/scenarios.py#L367-L414).# noqa
         Args:
-          frame_id: A geometry::frame_id registered with scene_graph.
+          frame_id: a geometry::frame_id registered with scene_graph.
           length: the length of each axis in meters.
           radius: the radius of each axis in meters.
           opacity: the opacity of the coordinate axes, between 0 and 1.
--- a/bindings/pydrake/visualization/model_visualizer.py
+++ b/bindings/pydrake/visualization/model_visualizer.py
@@ -54,8 +54,12 @@ def _main():
     assert defaults["pyplot"] is False
     args_parser.add_argument(
         "--pyplot", action="store_true",
-        help="Opens a pyplot figure for rendering using "
+        help="Open a pyplot figure for rendering using "
              "PlanarSceneGraphVisualizer.")
+    # N.B.: There's no default for enable_reload; it's not an init argument.
+    args_parser.add_argument(
+        "-r", "--enable_reload", action="store_true",
+        help="Enable reloading of the model.")
     # TODO(russt): Consider supporting the PlanarSceneGraphVisualizer
     #  options as additional arguments.
     assert defaults["visualize_frames"] is False
@@ -122,7 +126,11 @@ def _main():
         args_parser.error(f"File does not exist: {filename}")
 
     visualizer.AddModels(filename)
-    visualizer.Run(position=args.position, loop_once=args.loop_once)
+    if not args.loop_once and args.enable_reload:
+        # TODO(trowell-tri) Consider enabling reload by default in the future.
+        visualizer.RunWithReload(position=args.position)
+    else:
+        visualizer.Run(position=args.position, loop_once=args.loop_once)
 
 
 if __name__ == '__main__':


trowell-tri (commented):
Towards #18334
+(release notes: feature)

You can test that the prior functionality remains by running
```bazel run //tools:model_visualizer -- package://drake/manipulation/models/iiwa_description/iiwa7/iiwa7_with_box_collision.sdf -w -q 1 0 0 0 0 0 0 -2.5 1 0 0 0 0 0```

Also try ```bazel run //tutorials:authoring_multibody_simulation``` to see that the relevant tutorial isn't broken.

To test the new functionality, add `-r` or `--enable_reload`. An easy edit to make is to change the joint names.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJlL-N17FkLTKJkZh41:b9ztkbw)* status: needs platform reviewer assigned, needs at least two assigned reviewers

**Review by trowell-tri — COMMENTED** [review]
+@seancurtis-tri for feature review, please.
FYI @jshep-tri 

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJld03Y8Hf6Togg_OON:b-7hymsw)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

**Review by SeanCurtis-TRI — COMMENTED** [review]
First pass done -- I've got some fundamental questions. See below.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJpN8Kp5RHnEC8HMiCo:b-ibngur)* status: 9 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @trowell-tri)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp6HP306n1IYOAh6aD:-NJp6HP4DQEqHL1BQVRJ:bc0nl8s) (no related file):*
I'm a bit confused as to the design of the feature:

1. A user has to *opt* in to have the reload button. Why isn't it simply always available?
2. In order to enable the reload button, the user has to provide positions. Why doesn't it default to the zero configuration in the absence of an appropriately sized position vector?
3. It's not clear to me why "Run" and "Run with reload" are so different. Admittedly, I didn't spend a great deal of time exploring that because of the previous two design questions. I'd imagine depending on how those get resolved, the distinction between the two modes may evaporate.  I'll delay on elaborating the most conspicuous differences until I see if they survive.

___
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 272 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpE2_o7qAAne6XPPPq:-NJpE2_o7qAAne6XPPPr:b-cc3bef) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L272)):*
> ```Python
> 
>         Returns:
>           The slider positions if the visualizer is internally set to allow
> ```

BTW Why not just always return the slider values? The caller isn't obliged to capture them. Looking at the for loop, it's clear that you always have access to them so there's no particular cost.

Is it because of the `KeyboardInterrupt`? Seems to me that one more call to

```py
return self._sliders.get_output_port().Eval(sliders_context)
```

in the finally block would simplify the return value.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 273 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp9OHg8bz4a5sr2zC1:-NJp9OHg8bz4a5sr2zC2:bxikxt3) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L273)):*
> ```Python
>         Returns:
>           The slider positions if the visualizer is internally set to allow
>         reloading model filesand the user clicks the Reload button.
> ```

nit s/filesand/files and
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 274 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp9__cEcEsQQqLjXpy:-NJp9__cEcEsQQqLjXpz:b3f7e47) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L274)):*
> ```Python
>           The slider positions if the visualizer is internally set to allow
>         reloading model filesand the user clicks the Reload button.
>         Otherwise, always returns None.
> ```

nit: The word "always" doesn't belong. This text, as it stands, basically says "It always does A, unless it doesn't."
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJpFFjg9CPzb5dWKzBV:b-5nnpre) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*
> ```Python
>             self._plant_context)])
> 
>         if position:
> ```

BTW Is it worth distinguishing between an empty list of positions and no positions passed at all? Or are they, as this code implies, equivalent?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJpKgD7D1cpujMh9ZbT:b-fqpey7) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*
> ```Python
>             self._plant_context)])
> 
>         if position:
> ```

Passing `--enable_reload` without passing explicit positions crashes. That seems strange to me. Not just strange, but flawed. For example, if I'm developing a model for the first time and I don't have a strong sense of what all of the position values are, how can I pass them until I've seen the model and its sliders?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 389 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpHkpIBq7PWewN6URp:-NJpHkpJ2w4XmoC2jU6E:bu3xhqm) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L389)):*
> ```Python
> 
>             # Don't exit if the number of positions in the model changes.
>             # TODO(trowell-tri) Examine the scene graph more thoroughly
> ```

nit s/scene graph/plant
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 404 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpHd4n-BHC9hYFImtI:-NJpHd4n-BHC9hYFImtJ:b9zdk9s) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L404)):*
> ```Python
>         Raises:
>           ValueError: if the length of the position list does not match
>         the number of positions in the scene graph.
> ```

nit s/scene graph/plant.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 127 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKyLRA6yEq-zQcqFT:-NJpKyLRA6yEq-zQcqFU:b-10ueeb) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/model_visualizer.py#L127)):*
> ```Python
>         args_parser.error(f"File does not exist: {filename}")
> 
>     if args.enable_reload:
> ```

This introduces inconsistency on whose job it is to call `AddModels`. I'd hope that either this file is consistently responsible for doing so, or `_ModelVisualizer` is. But this split duty has a bad smell to it.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 6 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp6HP306n1IYOAh6aD:-NJpNv600GVNLm9s1_CC:b-z2xj7v) (no related file):*

I'm a bit confused as to the design of the feature:

1. A user has to *opt* in to have the reload button. Why isn't it simply always available?
2. In order to enable the reload button, the user has to provide positions. Why doesn't it default to the zero configuration in the absence of an appropriately sized position vector?
3. It's not clear to me why "Run" and "Run with reload" are so different. Admittedly, I didn't spend a great deal of time exploring that because of the previous two design questions. I'd imagine depending on how those get resolved, the distinction between the two modes may evaporate.  I'll delay on elaborating the most conspicuous differences until I see if they survive.

</blockquote></details>

1. I wanted to keep the effect of this change as small as I could for now, so e.g. I didn't break other users of ModelVisualizer like I have in the past. This is meant to be the first step, and to create discussion (which it's doing!) but I didn't necessarily think I could resolve them all those discussions this go-around. Plus this feature doesn't make much sense in some cases, e.g. within a Notebook or where you're using the parser to add models in some more elaborate way. It sounds like this needs more documentation.
2. You don't need to provide positions to use this. Is there a bad comment somewhere that suggests this?
3. The reload and run_once modes are (to my thinking) in conflict enough that I wanted to keep them separate, plus see item 1.

___
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 272 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpE2_o7qAAne6XPPPq:-NJpOeV81beVp-KNi8v3:bk87yu) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L272)):*

BTW Why not just always return the slider values? The caller isn't obliged to capture them. Looking at the for loop, it's clear that you always have access to them so there's no particular cost.

Is it because of the `KeyboardInterrupt`? Seems to me that one more call to

```py
return self._sliders.get_output_port().Eval(sliders_context)
```

in the finally block would simplify the return value.
</blockquote></details>

No, I just wanted to keep the modes functionally separate. There's no real reason to not always return the positions other than "it might be confusing" and "I'm not sure if this might break something that's not in a test suite".
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJpPV5U8Zeisdal1ubj:bktoldk) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

BTW Is it worth distinguishing between an empty list of positions and no positions passed at all? Or are they, as this code implies, equivalent?
</blockquote></details>

As near as I can tell, they're always equivalent. Do you see something I missed?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJpPqza63YXG1wFnjSQ:b-ppwyzd) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

Passing `--enable_reload` without passing explicit positions crashes. That seems strange to me. Not just strange, but flawed. For example, if I'm developing a model for the first time and I don't have a strong sense of what all of the position values are, how can I pass them until I've seen the model and its sliders?
</blockquote></details>

Now I understand a comment from above, but it doesn't crash for me. Can you give me your invocation (and the stack trace)?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 404 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpHd4n-BHC9hYFImtI:-NJpQDJA1-DkinTPh7_6:b-1jl00q) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L404)):*

nit s/scene graph/plant.
</blockquote></details>

Fixed. Thank you.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 127 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKyLRA6yEq-zQcqFT:-NJpQKn48Xf_pHLdP7JJ:buctkj0) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/model_visualizer.py#L127)):*

This introduces inconsistency on whose job it is to call `AddModels`. I'd hope that either this file is consistently responsible for doing so, or `_ModelVisualizer` is. But this split duty has a bad smell to it.
</blockquote></details>

Originally I kept the `AddModels` here in both cases. I decided to push it into the reload version in this case to nudge users away from doing manual model handling at all, and especially using the parser to do fancy things because that will break this functionality.

So, either there is a problem with the documentation, or I should revert to allowing `AddModels` in the reload case. Thoughts on which is better? Sounds like you might prefer the latter?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @trowell-tri)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp6HP306n1IYOAh6aD:-NJpXyqs0zAIlig2s-hJ:b-eud10d) (no related file):*

1. I wanted to keep the effect of this change as small as I could for now, so e.g. I didn't break other users of ModelVisualizer like I have in the past. This is meant to be the first step, and to create discussion (which it's doing!) but I didn't necessarily think I could resolve them all those discussions this go-around. Plus this feature doesn't make much sense in some cases, e.g. within a Notebook or where you're using the parser to add models in some more elaborate way. It sounds like this needs more documentation.
2. You don't need to provide positions to use this. Is there a bad comment somewhere that suggests this?
3. The reload and run_once modes are (to my thinking) in conflict enough that I wanted to keep them separate, plus see item 1.
</blockquote></details>

I'm going to defer responding til I've had a chance to discuss just so I can see how many of my thoughts arise from my own ignorance.

___
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 272 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpE2_o7qAAne6XPPPq:-NJpU1jZ40WGJQi76-AA:b-o97gau) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L272)):*

No, I just wanted to keep the modes functionally separate. There's no real reason to not always return the positions other than "it might be confusing" and "I'm not sure if this might break something that's not in a test suite".
</blockquote></details>

Given that ,in the past, it always returned an undocumented `None`, it's hard to imagine that any existing code would be harmed by returning the slider values. If they're capturing the returned `None` now, then, shame on them.

It has the benefit of simplifying the semantics of this function

-------------------------------

Follow up

Looking at the code, I think I better understand what's going on.

One can't infer why it makes sense for `Run` to sometimes return `q` and sometimes not by looking at its documentation. What's really happening is that it's serving multiple roles, because we might want to leave the run loop for various reeasons.

1. Run was set to single loop.
2. Run was exited (via Esc, Ctrl-C, or the "stop" button).
3. The reload button was hit.

Conditions (1) and (2) are lumped together and signaled by returning None. Condition (3) is signaled by returning `q`.

The key here is that, ultimately, the return value is indicating something about how the run loop exited.

I suspect we'd better off sending an explicit signal, differentiating between those three signals. Then any other code that uses the `Run` loop can decide what it means to exit the loop in various ways. If we did so in a way where the `Run()` loop didn't have to have knowledge on what the stopping criteria are, so much the better.

Imagine enumerations: KeepGoing, RunOnce, Quit, Reload. Some function called in the run loop is responsible for examining the state of the visualizer and reporting one of those four values. `Run()` looks for `KeepGoing` to stay in its loop. It returns all other values directly. This allows us to change stopping criteria arbitrarily without worrying about how we encode needs in the return value.

`RunWithReload()`, given a signal that indicates "Reload was clicked" can grab the slider values itself and, therefore, `Run` doesn't need to know anything about what the caller's needs are. Better separation of concerns.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJpTU8WELSTUsn8uU7j:b3ic9m3) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

Now I understand a comment from above, but it doesn't crash for me. Can you give me your invocation (and the stack trace)?
</blockquote></details>

All I did was remove the `-q` parameter from what you suggested.

```bash
bazel run //tools:model_visualizer -- package://drake/manipulation/models/iiwa_description/iiwa7/iiwa7_with_box_collision.sdf -w --enable_reload
INFO: Invocation ID: 25cd758b-cd96-4c17-8e96-67b3b049441a
INFO: Analyzed target //tools:model_visualizer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //bindings/pydrake/visualization:model_visualizer up-to-date:
  bazel-bin/bindings/pydrake/visualization/model_visualizer
INFO: Elapsed time: 0.436s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/bindings/pydrake/visualization/model_visualizer package://drake/manINFO: Build completed successfully, 1 total action
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/model_visualizer.py", line 135, in <module>
    _main()
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/model_visualizer.py", line 128, in _main
    visualizer.RunWithReload([filename], position=args.position)
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/_model_visualizer.py", line 364, in RunWithReload
    self._raise_if_invalid_positions(position)
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/_model_visualizer.py", line 410, in _raise_if_invalid_positions
    raise ValueError(
ValueError: Number of passed positions (1) does not match the number in the model (14).
```
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJpWptr0qmhfk6wrFN4:bvongqc) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

As near as I can tell, they're always equivalent. Do you see something I missed?
</blockquote></details>

Usually, when the parameter gets defaulted to `None`, that implies an important distinction. Hence my question. I don't have a firm practice philosophy on this particular aspect because I don't spend much time in python nowadays and I agree that there doesn't seem to be much practical distinction in the code right now.

If I had to make a distinction I'd say,

- `None` means that you should do whatever behavior is appropriate when no position data has been specified. None gets most naturally passed when the parameter is omitted (although a user could choose to pass it in explicitly).
- An empty list means that the caller explicitly passed a list that is *bad*. That may deserve feedback saying, "Rather than deferring to the default behavior, you've asked for the impossible".
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 127 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKyLRA6yEq-zQcqFT:-NJpXuE831kM7AuIoIxw:b-xk48wf) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/model_visualizer.py#L127)):*

Originally I kept the `AddModels` here in both cases. I decided to push it into the reload version in this case to nudge users away from doing manual model handling at all, and especially using the parser to do fancy things because that will break this functionality.

So, either there is a problem with the documentation, or I should revert to allowing `AddModels` in the reload case. Thoughts on which is better? Sounds like you might prefer the latter?
</blockquote></details>

Probably best addressing in a face-to-face discussion.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 272 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpE2_o7qAAne6XPPPq:-NJpzGQZ-DJmX6L2pzpi:bbaylni) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L272)):*

Given that ,in the past, it always returned an undocumented `None`, it's hard to imagine that any existing code would be harmed by returning the slider values. If they're capturing the returned `None` now, then, shame on them.

It has the benefit of simplifying the semantics of this function

-------------------------------

Follow up

Looking at the code, I think I better understand what's going on.

One can't infer why it makes sense for `Run` to sometimes return `q` and sometimes not by looking at its documentation. What's really happening is that it's serving multiple roles, because we might want to leave the run loop for various reeasons.

1. Run was set to single loop.
2. Run was exited (via Esc, Ctrl-C, or the "stop" button).
3. The reload button was hit.

Conditions (1) and (2) are lumped together and signaled by returning None. Condition (3) is signaled by returning `q`.

The key here is that, ultimately, the return value is indicating something about how the run loop exited.

I suspect we'd better off sending an explicit signal, differentiating between those three signals. Then any other code that uses the `Run` loop can decide what it means to exit the loop in various ways. If we did so in a way where the `Run()` loop didn't have to have knowledge on what the stopping criteria are, so much the better.

Imagine enumerations: KeepGoing, RunOnce, Quit, Reload. Some function called in the run loop is responsible for examining the state of the visualizer and reporting one of those four values. `Run()` looks for `KeepGoing` to stay in its loop. It returns all other values directly. This allows us to change stopping criteria arbitrarily without worrying about how we encode needs in the return value.

`RunWithReload()`, given a signal that indicates "Reload was clicked" can grab the slider values itself and, therefore, `Run` doesn't need to know anything about what the caller's needs are. Better separation of concerns.
</blockquote></details>

Returning an enum was my original implementation and perhaps I got too minimalistic. I actually prefer returning a stop reason and have added that.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJpqC6VD1XJDEMnBviF:b1axlzg) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

All I did was remove the `-q` parameter from what you suggested.

```bash
bazel run //tools:model_visualizer -- package://drake/manipulation/models/iiwa_description/iiwa7/iiwa7_with_box_collision.sdf -w --enable_reload
INFO: Invocation ID: 25cd758b-cd96-4c17-8e96-67b3b049441a
INFO: Analyzed target //tools:model_visualizer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //bindings/pydrake/visualization:model_visualizer up-to-date:
  bazel-bin/bindings/pydrake/visualization/model_visualizer
INFO: Elapsed time: 0.436s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/bindings/pydrake/visualization/model_visualizer package://drake/manINFO: Build completed successfully, 1 total action
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/model_visualizer.py", line 135, in <module>
    _main()
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/model_visualizer.py", line 128, in _main
    visualizer.RunWithReload([filename], position=args.position)
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/_model_visualizer.py", line 364, in RunWithReload
    self._raise_if_invalid_positions(position)
  File "/home/seancurtis/.cache/bazel/_bazel_seancurtis/c482a09839de82c7b8c8a508d26ab2c3/execroot/drake/bazel-out/k8-opt/bin/bindings/pydrake/visualization/model_visualizer.runfiles/drake/bindings/pydrake/visualization/_model_visualizer.py", line 410, in _raise_if_invalid_positions
    raise ValueError(
ValueError: Number of passed positions (1) does not match the number in the model (14).
```
</blockquote></details>

Interesting; so I'd think you should get a similar error message if you turned off the reloading. Do you? I cannot make it do this and we may need to debug to figure out how you're getting a single value. Are you on Jammy?

I cannot for the life of me figure out where you're getting a single position from … we may need to do something interactive.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJpuYif5sgPtrFPixBp:b-9fvbvu) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

Usually, when the parameter gets defaulted to `None`, that implies an important distinction. Hence my question. I don't have a firm practice philosophy on this particular aspect because I don't spend much time in python nowadays and I agree that there doesn't seem to be much practical distinction in the code right now.

If I had to make a distinction I'd say,

- `None` means that you should do whatever behavior is appropriate when no position data has been specified. None gets most naturally passed when the parameter is omitted (although a user could choose to pass it in explicitly).
- An empty list means that the caller explicitly passed a list that is *bad*. That may deserve feedback saying, "Rather than deferring to the default behavior, you've asked for the impossible".
</blockquote></details>

I can't default the parameter to an empty list, because then modifying the value in the method would modify the default object. I could have a separate input parameter that I don't modify I think this is pretty standard idiom.

Regardless, I'm switching to adding models via `AddModels` for both paths, so this is kind of moot.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 127 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKyLRA6yEq-zQcqFT:-NJpyvpiB6B0UuTX-dS4:b-rmyaur) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/model_visualizer.py#L127)):*

Probably best addressing in a face-to-face discussion.
</blockquote></details>

Per discussion, I've switched to always using `AddModels` and added more documentation telling people to not do the other thing.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJqAqhXBIhReTzvcoMJ:b-trqxhj) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

Interesting; so I'd think you should get a similar error message if you turned off the reloading. Do you? I cannot make it do this and we may need to debug to figure out how you're getting a single value. Are you on Jammy?

I cannot for the life of me figure out where you're getting a single position from … we may need to do something interactive.
</blockquote></details>

I tried it again on the newest version and it's happy.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpKgD7D1cpujMh9ZbS:-NJqB8juAkvFAojUWoU9:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

I tried it again on the newest version and it's happy.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJqBXr6Cbm7jq9TzP40:bchf71) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

I can't default the parameter to an empty list, because then modifying the value in the method would modify the default object. I could have a separate input parameter that I don't modify I think this is pretty standard idiom.

Regardless, I'm switching to adding models via `AddModels` for both paths, so this is kind of moot.
</blockquote></details>

I wasn't advocating changing the default value. Merely interpreting it. In essence, it would change `if position` to `if position is not None`.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJqEW_W5Sxd9YS9dx_1:b-tice85) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

I wasn't advocating changing the default value. Merely interpreting it. In essence, it would change `if position` to `if position is not None`.
</blockquote></details>

Fair. This made me realize we should check this in a few other places, but I think both the `None` and "empty list" cases are the same in that we do not want to call `SetPositions` passing an empty list.

SeanCurtis-TRI (commented):
last round of comments

Mostly some straightforward nits and a question or two.

:LGTM: 

But feel free to pass it on to platform once addressed.

> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJqUeqUB83mtvbNsLqa:b-n96s5x)* status: 7 unresolved discussions, needs at least two assigned reviewers (waiting on @trowell-tri)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp6HP306n1IYOAh6aD:-NJqRj6m8LqmVsYqy6Zg:b-yleort) (no related file):*

I'm going to defer responding til I've had a chance to discuss just so I can see how many of my thoughts arise from my own ignorance.
</blockquote></details>

After our face to face an my ruminations:

1. I'd argue that in the *long run* "reload off" by default is not the right thing. However, in the short-term, I respect the author's discretion of valuing avoiding any possible disruption.
2. Whatever caused my observed bug (documented below) has gone away (if it was ever more than some strange quirk on my machine). This is implicitly resolved.
3. I think we've introduced some redundancy that muddies clarity with the explosion. However, that can be resolved in the long run with our best effort now of flagging technical debt for future work. Short term, this feature works.

___
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 272 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpE2_o7qAAne6XPPPq:-NJqSJ1G2AUo2s0psahF:bshzwtd) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L272)):*

Returning an enum was my original implementation and perhaps I got too minimalistic. I actually prefer returning a stop reason and have added that.
</blockquote></details>

I'm marking myself satisified. Worth noting that the value `KEEP_GOING` will never be returned?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJqRM_eB5taBT2_K0XU:bramvji) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

Fair. This made me realize we should check this in a few other places, but I think both the `None` and "empty list" cases are the same in that we do not want to call `SetPositions` passing an empty list.
</blockquote></details>

I'd emphasize the semantic distinction -- if we get an empty list, its because *someone passed it*. If an empty position list represents an incorrectly sized position vector, the user deserves feedback. Simply ignoring it means the user won't necessarily know they passed something empty. If the *mean* to pass something empty, they should pass None.  

Also see my note that `if position:` is equivalent to what you've written here.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 51 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqSheOBZdjpJuszA-e:-NJqSheOBZdjpJuszA-f:b792f9z) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L51)):*
> ```Python
>     # Note: this class uses C++ method names to ease future porting.
> 
>     class RunResult(Enum):
> ```

nit: [Pep8](https://peps.python.org/pep-0008/#constants) suggests that "constants" should be defined at the module level. And PEP8 is what serves as the [drake python style guide](https://drake.mit.edu/code_style_guide.html).
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 283 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqT0fb2bmQfjfEwBEx:-NJqT0fb2bmQfjfEwBEy:b68mhvp) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L283)):*
> ```Python
>         if self._builder:
>             self.Finalize(position=position)
>         elif position is not None and len(position) > 0:
> ```

BTW The equivalent spelling of this is:

```py
elif position:
```

However, I'm not sure what inspired the concern about explicitly testing the position length here... It's not like a non-empty position list is particularly more compatible with `SetPositions` than an empty one. it has to be just the right length and that's not caught in these conditions.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 327 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqT_841le-tSd_H2d2:-NJqT_841le-tSd_H2d3:b-dt1xbe) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L327)):*
> ```Python
>                     loop_result = self.RunResult.RELOAD
>         except KeyboardInterrupt:
>             pass
> ```

nit: I believe you need to replace this `pass` with `loop_result = RunReslut.STOPPED`.  Yes?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 374 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqTsyOEvzY-yQgLbZC:-NJqTsyOEvzY-yQgLbZD:baqadqr) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L374)):*
> ```Python
>         if position is not None and len(position) > 0:
>             # Finalize the plant so we can validate positions.
>             self._plant.Finalize()
> ```

This code will *always* strike me as odd. It *looks* like if we don't provide positions we don't finalize the plant. But that seems ridiculous. How can we do meaningful visualization if that were the case? How would we be able to pose it? I suspect, this isn't what you think it is. (At the very least, it's not what *I* think it is.)

This is the kind of thing that really confuses me about understanding what the data flow is.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 399 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqUCxx4CyXEUog9QHg:-NJqUCxx4CyXEUog9QHh:b-umjlyo) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L399)):*
> ```Python
>             inner_visualizer.Finalize()
> 
>             if position is not None and len(position) > 0:
> ```

nit: Same note here.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 59 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqUZXKBHWd1sB8IFAe:-NJqUZXKBHWd1sB8IFAf:bcq154z) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/model_visualizer.py#L59)):*
> ```Python
>         help="Open a pyplot figure for rendering using "
>              "PlanarSceneGraphVisualizer.")
>     args_parser.add_argument(
> ```

nit: To match the surrounding pattern, should this be included in the defaults and also with an assert?

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJp6HP306n1IYOAh6aD:-NJqcENW16U0hlqTTQQI:b6fm8m3) (no related file):*

After our face to face an my ruminations:

1. I'd argue that in the *long run* "reload off" by default is not the right thing. However, in the short-term, I respect the author's discretion of valuing avoiding any possible disruption.
2. Whatever caused my observed bug (documented below) has gone away (if it was ever more than some strange quirk on my machine). This is implicitly resolved.
3. I think we've introduced some redundancy that muddies clarity with the explosion. However, that can be resolved in the long run with our best effort now of flagging technical debt for future work. Short term, this feature works.
</blockquote></details>

I've added a TODO.

___
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJqYcdjEI5xiV9W5gU-:b-1e2qwn) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

I'd emphasize the semantic distinction -- if we get an empty list, its because *someone passed it*. If an empty position list represents an incorrectly sized position vector, the user deserves feedback. Simply ignoring it means the user won't necessarily know they passed something empty. If the *mean* to pass something empty, they should pass None.  

Also see my note that `if position:` is equivalent to what you've written here.
</blockquote></details>

On the principle that we should be generous in what we accept, I wanted to treat an empty list the same as None. Those were the prior semantics and some users are students. But I don't feel strongly about this either way.

However, note my prior comments about the truth values of NumPy arrays making the other test impossible.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 51 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqSheOBZdjpJuszA-e:-NJqWgdG8H4qJIvw4aqi:b-rxzb2b) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L51)):*

nit: [Pep8](https://peps.python.org/pep-0008/#constants) suggests that "constants" should be defined at the module level. And PEP8 is what serves as the [drake python style guide](https://drake.mit.edu/code_style_guide.html).
</blockquote></details>

OK. I saw some code that defined them at the class level so I did that. Will switch.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 283 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqT0fb2bmQfjfEwBEx:-NJqXUSu7z_QdZ_9WFe9:b-2eedw9) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L283)):*

BTW The equivalent spelling of this is:

```py
elif position:
```

However, I'm not sure what inspired the concern about explicitly testing the position length here... It's not like a non-empty position list is particularly more compatible with `SetPositions` than an empty one. it has to be just the right length and that's not caught in these conditions.
</blockquote></details>

I tried this (because it is what I'd expect as well) but it trips up with `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`, I guess because this is a NumPy array? I've added a comment.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 327 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqT_841le-tSd_H2d2:-NJqYD0x4TaGZvYvsfsN:bb6un1j) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L327)):*

nit: I believe you need to replace this `pass` with `loop_result = RunReslut.STOPPED`.  Yes?
</blockquote></details>

Done
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 374 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqTsyOEvzY-yQgLbZC:-NJqZaZP78f8GiysLMiY:bond5h0) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L374)):*

This code will *always* strike me as odd. It *looks* like if we don't provide positions we don't finalize the plant. But that seems ridiculous. How can we do meaningful visualization if that were the case? How would we be able to pose it? I suspect, this isn't what you think it is. (At the very least, it's not what *I* think it is.)

This is the kind of thing that really confuses me about understanding what the data flow is.
</blockquote></details>

Here we're using the outer visualizer merely as scaffolding, and in this particular stanza using it to validate the positions. I've added a comment; does that help?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 399 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqUCxx4CyXEUog9QHg:-NJqahln3Emj-f2sdUgI:b1gf7fk) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L399)):*

nit: Same note here.
</blockquote></details>

I think I understand the issue, and have added a comment.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 59 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqUZXKBHWd1sB8IFAe:-NJqbOF46NEtXiR7WQy7:b-knz5m3) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/model_visualizer.py#L59)):*

nit: To match the surrounding pattern, should this be included in the defaults and also with an assert?
</blockquote></details>

No, because `enable_reload` isn't a constructor argument. I've added a comment.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 359 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJpFFjg9CPzb5dWKzBU:-NJqfJL-DSGQUzlBGH2p:b-qnkbz) ([raw file](https://github.com/robotlocomotion/drake/blob/be9999a51b47ab2f3570111747f27a0191531e0a/bindings/pydrake/visualization/_model_visualizer.py#L359)):*

On the principle that we should be generous in what we accept, I wanted to treat an empty list the same as None. Those were the prior semantics and some users are students. But I don't feel strongly about this either way.

However, note my prior comments about the truth values of NumPy arrays making the other test impossible.
</blockquote></details>

Generosity is an interesting concept. Generous "in what we accept" is one form. Generous in giving feedback is another. As this is a subjective design decision, I'll defer to the author's discretion.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 283 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqT0fb2bmQfjfEwBEx:-NJqf431-IEv4ON_BuMF:b-2m7ttc) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L283)):*

I tried this (because it is what I'd expect as well) but it trips up with `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`, I guess because this is a NumPy array? I've added a comment.
</blockquote></details>

Right...I'd lost track of the fact that these were arrays and not lists.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 374 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqTsyOEvzY-yQgLbZC:-NJqfVFc3hBVSiSdOJZn:b-xvnvxo) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L374)):*

Here we're using the outer visualizer merely as scaffolding, and in this particular stanza using it to validate the positions. I've added a comment; does that help?
</blockquote></details>

It's worth noting that we're not finalizing *only* to validate positions. We'll have to eventually finalize no matter what. Perhaps:

```py
# To validate positions as early as possible, we'll do early plant finalization so that
# we can perform the validation.
```

(Unless I'm wrong in my understanding that this plant won't inevitably be finalized otherwise.)
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 59 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqUZXKBHWd1sB8IFAe:-NJqg31H5S6QvzSMQcmX:biriklk) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/model_visualizer.py#L59)):*

No, because `enable_reload` isn't a constructor argument. I've added a comment.
</blockquote></details>

Riiiiiiiight.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 374 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqTsyOEvzY-yQgLbZC:-NJqgmcFEeZduJNAXNv9:bacya91) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L374)):*

It's worth noting that we're not finalizing *only* to validate positions. We'll have to eventually finalize no matter what. Perhaps:

```py
# To validate positions as early as possible, we'll do early plant finalization so that
# we can perform the validation.
```

(Unless I'm wrong in my understanding that this plant won't inevitably be finalized otherwise.)
</blockquote></details>

Your parenthetical is correct: we never Finalize the **outer** visualizer (the self object) if we don't need it to validate. It's the **inner** visualizer (the one in the loop, that we recreate every reload) that needs to get Finalized as usual. We never really use the other infrastructure in that outer visualizer other than as a place to keep out init params and the meshcat — and the plant for validation only. I was hoping the comments would clarify? Still no?

trowell-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJqmu4cD0GThWoOm46X:bppyoqq)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform) (waiting on @SeanCurtis-TRI and @sherm1)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488)* status: LGTM missing from assignee sherm1(platform) (waiting on @sherm1)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 374 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18488#-NJqTsyOEvzY-yQgLbZC:-NJqw7L-ApWn9hO2LwKe:bwqm0q4) ([raw file](https://github.com/robotlocomotion/drake/blob/b7e5a85d0260cb31426e164958feae96c3e8cb16/bindings/pydrake/visualization/_model_visualizer.py#L374)):*

Your parenthetical is correct: we never Finalize the **outer** visualizer (the self object) if we don't need it to validate. It's the **inner** visualizer (the one in the loop, that we recreate every reload) that needs to get Finalized as usual. We never really use the other infrastructure in that outer visualizer other than as a place to keep out init params and the meshcat — and the plant for validation only. I was hoping the comments would clarify? Still no?
</blockquote></details>

Sounds good.  Thanks for the elucidation.

sherm1 (commented):
A design question (probably you already discussed this somewhere): why is this new feature optional? Wouldn't it be better always to have a Reload button available? I'm not certain about the use case but if I understand it correctly users would have to (a) know this feature exists, and (b) plan ahead to get the button. If it was there by default new users could discover it and could use it without planning ahead.

Anyway, what's here :lgtm: but I'd be happy to see this without the new command line parameters with an omnipresent Reload button.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJqy5Eg3nLEHqdW8xMm:b-omotk6)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),sherm1(platform) (waiting on @trowell-tri)

SeanCurtis-TRI (commented):
I had the same question. The understanding that I bought was:

There's a balance between the long-term benefit of having it always on and the short term benefit of not accidentally breaking something that might be out there (Russ's lectures notes, for example). Author's discretion favored the "non-disruptive" value over the "long-term" value. It can always be turned on by default later.

As it had been considered by the author, I was content to accept author's choice.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18488#-:-NJr0ZRE31WS1DPRGWkc:bpyk6gw)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),sherm1(platform) (waiting on @trowell-tri)

trowell-tri:
There is some prior discussion around these parts about why this is optional for now, but to recap:

1. This is just a minimum useful version of this feature and we're going to iterate to figure out it should best work, but I wanted to make it available to seek feedback. There was a slack thread discussing this feature a while back as well as discussion on the GH issue but I don't feel it really settled and I don't yet have a strong vision for what the answers should be myself, hence I wanted to be as un-intrusive as possible.

2. I wanted to keep the effect of this change small for now so it can't possibly break other users of `ModelVisualizer`, especially Jupyter notebooks. And in the case of embedded code and particularly notebooks I think it's often true that a reload button won't always make sense — the model may not be in a file but rather a buffer, or the caller may have done various tricks with the `parser` to add the models and there's currently no way to find out from the parser what was actually loaded. (You see, the `ModelVisualizer` class does double-duty both providing the functionality of the `model_visualizer` script tool but also as an easy way to visualize models from other code.)

PR #18496, opened 2022-12-27
[C-IRIS] Consolidate to the formulation using Chebyshev center for polytope

hongkai-dai:
If I were to use the formulation that doesn't introduce Chebyshev, namely a'*x + b >= 1, then this causes problem in the bilinear iteration. Since when we search for the separating margin, we use the formulation a'*x + b - margin - lambda(s) * (d - C*s) is sos. Then in the next iteration when we search for C and d, we don't need the margin and use the formulation a'*x + b - 1 - lambda(s) * (d-C*s) is sos. This causes the problem that the solution a, b, lambda in the first iteration with margin is NOT a feasible solution for the next iteration when we search for C and d but fix lambda(s) (as very likely the margin is < 1).

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496)

### DIFF
--- a/geometry/optimization/dev/collision_geometry.cc
+++ b/geometry/optimization/dev/collision_geometry.cc
@@ -29,8 +29,7 @@ struct ReifyData {
           m_X_AB_multilinear,
       const multibody::RationalForwardKinematics& m_rational_forward_kin,
       const std::optional<symbolic::Variable>& m_separating_margin,
-      PlaneSide m_plane_side, GeometryType m_other_side_geometry_type,
-      GeometryId m_geometry_id,
+      PlaneSide m_plane_side, GeometryId m_geometry_id,
       std::vector<symbolic::RationalFunction>* m_rationals,
       std::optional<VectorX<symbolic::Polynomial>>* m_unit_length_vector)
       : a{&m_a},
@@ -39,7 +38,6 @@ struct ReifyData {
         rational_forward_kin{&m_rational_forward_kin},
         separating_margin{&m_separating_margin},
         plane_side{m_plane_side},
-        other_side_geometry_type{m_other_side_geometry_type},
         geometry_id{m_geometry_id},
         rationals{m_rationals},
         unit_length_vector{m_unit_length_vector} {}
@@ -53,7 +51,6 @@ struct ReifyData {
   const multibody::RationalForwardKinematics* rational_forward_kin;
   const std::optional<symbolic::Variable>* separating_margin;
   const PlaneSide plane_side;
-  const GeometryType other_side_geometry_type;
   const GeometryId geometry_id;
   std::vector<symbolic::RationalFunction>* rationals;
   std::optional<VectorX<symbolic::Polynomial>>* unit_length_vector;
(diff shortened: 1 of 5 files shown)


hongkai-dai (commented):
+@alexandreamice for review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496#-:-NKHIeud5T2dnKB3zM4V:b-kengf8)* status: LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

AlexandreAmice (commented):
Two things to consider.
1) Is there any reason to make the separating margin optional now? It may simplify to just remove the option all together
2) Is there a way to make the multiple `0.5` in `0.5r` a named constant? It might be less brittle this way.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496#-:-NKJWP4WDsdfccMLaYIc:b7hnx7c)* status: LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
Currently I still like to keep the separating margin optional. For the polytope case, when we have the separating margin, the constraint is
```
a.dot(p_AV) + b >= margin
a.dot(p_AC) + b >= 0.5 * radius
|a| <= 1
```
When we don't have the margin, the constraint is
```
a.dot(p_AV) +  b >= 0
a.dot(p_AC + b >= 0.5 * radius
```
The major difference is that when we have the margin, we introduce the constraint `|a|<=1`, which leads to a matrix-sos constraint.

> Is there a way to make the multiple 0.5 in 0.5r a named constant? It might be less brittle this way.

Done, I added a named constant.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496#-:-NKJonazAArY_0W2SZ6-:b4q8dlh)* status: LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

AlexandreAmice (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496#-:-NKJqLsj3uq8VwxA3oDI:bnfp4nl)* status: 1 unresolved discussion, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/collision_geometry.cc` line 113 at r2](https://reviewable.io/reviews/robotlocomotion/drake/18496#-NKJqARx45D41Ofx_A3W:-NKJqARx45D41Ofx_A3X:bgc44ms) ([raw file](https://github.com/robotlocomotion/drake/blob/6daadf6fdcedf8a42e54cd0991534f9ff1ba7cde/geometry/optimization/dev/collision_geometry.cc#L113)):*
> ```cpp
>   //   aᵀ*p_AC+b ≤ -k*r
>   // where C is the Chebyshev center of the polytope, and r is the distance
>   // from C to the polytope boundary. k is a positive scalar between (0.5 in
> ```

nit: `k is a positive scalar in (0,1), defined in PolytopeChebyshevRadiusMultiplier`

AlexandreAmice (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18496#-:-NKK9A9qAz7gXnSBi6eJ:bnfp4nl)* status: missing label for release notes (waiting on @hongkai-dai)

PR #18505, opened 2022-12-31
[C-IRIS] Find C-space polytope through binary search.

hongkai-dai:
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)

### DIFF
--- a/geometry/optimization/dev/cspace_free_polytope.cc
+++ b/geometry/optimization/dev/cspace_free_polytope.cc
@@ -1324,7 +1324,18 @@ HPolyhedron CspaceFreePolytope::GetPolyhedronWithJointLimits(
   return HPolyhedron(A, b);
 }
 
-std::optional<CspaceFreePolytope::BilinearAlternationResult>
+void CspaceFreePolytope::SearchResult::SetSeparatingPlanes(
+    std::vector<std::optional<SeparationCertificateResult>>
+        certificates_result) {
+  a.clear();
+  b.clear();
+  for (const auto& certificate : certificates_result) {
+    a.emplace(certificate->plane_index, std::move(certificate->a));
+    b.emplace(certificate->plane_index, std::move(certificate->b));
+  }
+}
+
+std::optional<CspaceFreePolytope::SearchResult>
 CspaceFreePolytope::SearchWithBilinearAlternation(
     const IgnoredCollisionPairs& ignored_collision_pairs,
     const Eigen::Ref<const Eigen::MatrixXd>& C_init,
@@ -1332,7 +1343,7 @@ CspaceFreePolytope::SearchWithBilinearAlternation(
     const BilinearAlternationOptions& options) const {
   DRAKE_DEMAND(C_init.rows() == d_init.rows());
   DRAKE_DEMAND(C_init.cols() == this->rational_forward_kin_.s().rows());
-  std::optional<CspaceFreePolytope::BilinearAlternationResult> ret{
+  std::optional<CspaceFreePolytope::SearchResult> ret{
       std::nullopt};
   int iter = 0;
   // When we search for the C-space polytope {s |C*s<=d, s_lower<=s<=s_upper},
@@ -1393,10 +1404,7 @@ CspaceFreePolytope::SearchWithBilinearAlternation(
       // Copy the result to ret.
       ret->C = C;
       ret->d = d;
-      for (const auto& certificate : certificates_result) {
-        ret->a.emplace(certificate->plane_index, std::move(certificate->a));
-        ret->b.emplace(certificate->plane_index, std::move(certificate->b));
-      }
+      ret->SetSeparatingPlanes(certificates_result);
     }
     // Now fix the Lagrangian and search for C-space polytope and separating
     // planes.
@@ -1431,6 +1439,86 @@ CspaceFreePolytope::SearchWithBilinearAlternation(
     }
     ++iter;
   }
+  return ret;
+}
+
+std::optional<CspaceFreePolytope::SearchResult>
+CspaceFreePolytope::BinarySearch(
+    const IgnoredCollisionPairs& ignored_collision_pairs,
+    const Eigen::Ref<const Eigen::MatrixXd>& C,
+    const Eigen::Ref<const Eigen::VectorXd>& d_init,
+    const Eigen::Ref<const Eigen::VectorXd>& s_center,
+    const BinarySearchOptions& options) const {
+  DRAKE_DEMAND(options.scale_max >= options.scale_min &&
+               options.scale_min >= 0);
+  DRAKE_DEMAND(((C * s_center).array() <= d_init.array()).all());
+  DRAKE_DEMAND((s_center.array() >= s_lower_.array()).all());
+  DRAKE_DEMAND((s_center.array() <= s_upper_.array()).all());
+  CspaceFreePolytope::SearchResult ret;
+
+  const Eigen::ArrayXd C_row_norm = C.rowwise().norm().array();
+  if ((C_row_norm == 0).any()) {
+    throw std::runtime_error(
+        "C contains rows with all 0 entries. Please remove these rows.");
+  }
+
+  // Determines if we can certify the scaled C-space polytope {s |C*s<=d,
+  // s_lower<=s<=s_upper} is collision free or not. Also update `ret` if eps is
+  // feasible.
+  auto is_scale_feasible = [this, &ignored_collision_pairs, &C, &d_init,
+                            &s_center, &options, &ret](double scale) {
+    // (d - C*s_center) / |C| = scale * (d_init - C*s_center) / |C|, hence d =
+    // scale * d_init + (1-scale) * C * s_center.
+    const Eigen::VectorXd d = scale * d_init + (1 - scale) * C * s_center;
+    const auto certificates_result =
+        this->FindSeparationCertificateGivenPolytope(
+            ignored_collision_pairs, C, d, false /* search_separating_margin */,
+            options.find_lagrangian_options);
+    if (std::any_of(
+            certificates_result.begin(), certificates_result.end(),
+            [](const std::optional<SeparationCertificateResult>& certificate) {
+              return !certificate.has_value();
+            })) {
+      return false;
+    } else {
+      ret.C = C;
+      ret.d = d;
+      ret.SetSeparatingPlanes(std::move(certificates_result));
+      return true;
+    }
+  };
+
+  if (!is_scale_feasible(options.scale_min)) {
+    drake::log()->error(
+        "CspaceFreePolytope::BinarySearch(): scale_min={} is infeasible.",
+        options.scale_min);
+    return std::nullopt;
+  }
+  if (is_scale_feasible(options.scale_max)) {
+    drake::log()->error(
+        "CspaceFreePolytope::BinarySearch(): scale_max={} is feasible.",
+        options.scale_max);
+    ret.num_iter = 0;
+    return ret;
+  }
+  double scale_min = options.scale_min;
+  double scale_max = options.scale_max;
+  int iter = 0;
+  while (scale_max - scale_min > options.convergence_tol &&
+         iter < options.max_iter) {
+    const double scale = (scale_max + scale_min) / 2;
+    if (is_scale_feasible(scale)) {
+      drake::log()->info(
+          "CspaceFreePolytope::BinarySearch(): scale={} is feasible", scale);
+      scale_min = scale;
+    } else {
+      drake::log()->info(
+          "CspaceFreePolytope::BinarySearch(): scale={} is infeasible", scale);
+      scale_max = scale;
+    }
+    ++iter;
+  }
+  ret.num_iter = iter;
   return ret;
 }
 
--- a/geometry/optimization/dev/cspace_free_polytope.h
+++ b/geometry/optimization/dev/cspace_free_polytope.h
@@ -306,7 +306,7 @@ class CspaceFreePolytope {
     bool search_s_bounds_lagrangians{true};
   };
 
-  struct BilinearAlternationResult {
+  struct SearchResult {
     Eigen::MatrixXd C;
     Eigen::VectorXd d;
     // a[i].dot(x) + b[i]=0 is the separation plane for separating_planes()[i].
@@ -314,6 +314,10 @@ class CspaceFreePolytope {
     std::unordered_map<int, symbolic::Polynomial> b;
     // The number of iterations at termination.
     int num_iter;
+
+    void SetSeparatingPlanes(
+        std::vector<std::optional<SeparationCertificateResult>>
+            certificates_result);
   };
 
   struct BilinearAlternationOptions {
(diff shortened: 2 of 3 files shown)


hongkai-dai:
@drake-jenkins-bot linux-focal-clang-bazel-experimental-everything-debug please
@drake-jenkins-bot inux-focal-clang-bazel-experimental-everything-thread-sanitizer please
@drake-jenkins-bot linux-focal-clang-bazel-experimental-everything-valgrind-memcheck please
@drake-jenkins-bot linux-focal-gcc-bazel-experimental-everything-debug please
@drake-jenkins-bot linux-focal-unprovisioned-clang-bazel-experimental-everything-debug please
@drake-jenkins-bot linux-focal-unprovisioned-gcc-bazel-experimental-everything-debug please
@drake-jenkins-bot linux-jammy-unprovisioned-clang-bazel-experimental-everything-debug please
@drake-jenkins-bot linux-jammy-unprovisioned-gcc-bazel-experimental-everything-debug please
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-everything-address-sanitizer please

AlexandreAmice (commented):
+@alexandreamice for feature review

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505#-:-NKjV06G0qLAK_H40_iG:b-4t6ud0)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 298 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjVJq80o1fjYHwBT2D:-NKjVJq80o1fjYHwBT2E:bd4kf74) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L298)):*
> ```objc
>     int num_iter;
> 
>     void SetSeparationPlanes(
> ```

nit: name "SetSeparatingPlanes"
___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 343 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjW1XoEM8hdBG5FGNR:-NKjW1XoEM8hdBG5FGNS:bxai3f0) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L343)):*
> ```objc
> 
>   /** Binary search on ε such that the C-space polytope {s | C*s<=d+ε,
>    * s_lower<=s<=s_upper} is collision free.
> ```

I have been thinking about this a little bit. Would it not be better to center the polytope around the seed point or its chebyshev center? We would still search for epsilon, but the meaning of the epsilon would be more intuitive. Right now, this scales the polytope in a weird way.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 343 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjW1XoEM8hdBG5FGNR:-NKjlNN282G17cYLH55u:b3pqbm2) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L343)):*

I have been thinking about this a little bit. Would it not be better to center the polytope around the seed point or its chebyshev center? We would still search for epsilon, but the meaning of the epsilon would be more intuitive. Right now, this scales the polytope in a weird way.
</blockquote></details>

I like the idea of making this epsilon geometrically more meaningful!

Using Chebyshev center is a bit tricky. A polytope's Chebyshev center is not always unique. For example, consider a 2D box -2 <= x <= 2, -1 <= y <= 1. Its Chebyshev center could be any point in the line segment -1 <= x <= 1, y = 0.

I wonder if we really need to specify a center of the polytope here. I think we can say push each face inward or outward by epsilon, that means to change the right-hand side of each face C.row(i) <= d(i) as C.row(i) <= d(i) + epsilon * |C.row(i)|. This way it is independent of the center, and also that epsilon has a geometric meaning. What do you think?

The downside is that if C.row(i) = 0, namely one row of C is just zero vector, then we cannot scale it using this definition, as d(i) + epsilon * |C.row(i)| is still d(i).

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 343 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjW1XoEM8hdBG5FGNR:-NKjp1V0CySwih0c7hMn:b-uvwl36) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L343)):*

I like the idea of making this epsilon geometrically more meaningful!

Using Chebyshev center is a bit tricky. A polytope's Chebyshev center is not always unique. For example, consider a 2D box -2 <= x <= 2, -1 <= y <= 1. Its Chebyshev center could be any point in the line segment -1 <= x <= 1, y = 0.

I wonder if we really need to specify a center of the polytope here. I think we can say push each face inward or outward by epsilon, that means to change the right-hand side of each face C.row(i) <= d(i) as C.row(i) <= d(i) + epsilon * |C.row(i)|. This way it is independent of the center, and also that epsilon has a geometric meaning. What do you think?

The downside is that if C.row(i) = 0, namely one row of C is just zero vector, then we cannot scale it using this definition, as d(i) + epsilon * |C.row(i)| is still d(i).
</blockquote></details>

I forgot to mention, the proposal above means that we should throw a warning if one row of C is zero vector (which is meaningless anyway).

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 343 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjW1XoEM8hdBG5FGNR:-NKjnR4TD-KCBhpi1rw5:b-zdgpgt) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L343)):*

I forgot to mention, the proposal above means that we should throw a warning if one row of C is zero vector (which is meaningless anyway).
</blockquote></details>

Yes, but in the case that C.row(i) = 0, then that constraint is either vacuous, or defines the empty set. I would say that in this case, the constraint is meaningless anyway and shifting it does nothing (in the certification code, it would be thrown out as redundant I hope)

While shifting each face by epsilon as you suggest is intuitive, I worry that it can introduce some unintended subtleties into the code. Namely, shifting each face by epsilon may change the face lattice of the polytope, i.e. which constraints are redundant/irredundant, as well as accidentally make the polytope empty.

That is why I think picking some point in the interior of the polytope is a more sensible thing to do. This scaling shouldn't change the face lattice. The point about which we scale will affect the geometric meaning, i.e. if the seed point isn't near the chebyshev center then scaling with respect to either point will result in very different polytopes. 

The point about which to scale is certainly a design choice. I am in favor of the seed point, or the average of the points we require the polytope to contain as we already have deemed these points as important, and we want our final polytope to contain them.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 343 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NKjW1XoEM8hdBG5FGNR:-NKkQ6NC6Q5WWs5_-dza:bgfnbl5) ([raw file](https://github.com/robotlocomotion/drake/blob/ca3e70a58600bba241f9687f744488ac0eb895b0/geometry/optimization/dev/cspace_free_polytope.h#L343)):*

Yes, but in the case that C.row(i) = 0, then that constraint is either vacuous, or defines the empty set. I would say that in this case, the constraint is meaningless anyway and shifting it does nothing (in the certification code, it would be thrown out as redundant I hope)

While shifting each face by epsilon as you suggest is intuitive, I worry that it can introduce some unintended subtleties into the code. Namely, shifting each face by epsilon may change the face lattice of the polytope, i.e. which constraints are redundant/irredundant, as well as accidentally make the polytope empty.

That is why I think picking some point in the interior of the polytope is a more sensible thing to do. This scaling shouldn't change the face lattice. The point about which we scale will affect the geometric meaning, i.e. if the seed point isn't near the chebyshev center then scaling with respect to either point will result in very different polytopes. 

The point about which to scale is certainly a design choice. I am in favor of the seed point, or the average of the points we require the polytope to contain as we already have deemed these points as important, and we want our final polytope to contain them.
</blockquote></details>

Done. I changed to search for the scaling factor. The user will need to provide the center point.

AlexandreAmice (commented):
> Reviewed 2 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/cspace_free_polytope.cc` line 1471 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NL7Glo-Cuwnh0fxgzl-:-NL7Glo0-jJt24NNBAHn:b-2aqlu1) ([raw file](https://github.com/robotlocomotion/drake/blob/1092df0cb74879b4b7d73ef54011fc4a709c0674/geometry/optimization/dev/cspace_free_polytope.cc#L1471)):*
> ```cpp
>                             &s_center, &options, &ret](double scale) {
>     // (d - C*s_center) / |C| = scale * (d_init - C*s_center) / |C|, hence d =
>     // scale * d_init + (1-scale) * C * s_center.
> ```

I disagree about the sign of this. I think it should be  `d + C*s_center`.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.cc` line 1471 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NL7Glo-Cuwnh0fxgzl-:-NL8E06R3cDfeU42ri-I:b-sz92ef) ([raw file](https://github.com/robotlocomotion/drake/blob/1092df0cb74879b4b7d73ef54011fc4a709c0674/geometry/optimization/dev/cspace_free_polytope.cc#L1471)):*

I disagree about the sign of this. I think it should be  `d + C*s_center`. 
</blockquote></details>

So our C-space polytope is defined as {s | C*s<=d, s_lower <= s <= s_upper}. The distance from a face `C.row(i) *s <= d(i)` to a point s_center is computed as `(d(i) - C.row(i) * s_center) / |C.row(i)|`, hence the derivation above.

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 355 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NLC6TXI818MJteJRC8O:-NLC6TXI818MJteJRC8P:b-ih5268) ([raw file](https://github.com/robotlocomotion/drake/blob/004939c5c8eb2a502bb2d7e532d2d40c6d958002/geometry/optimization/dev/cspace_free_polytope.h#L355)):*
> ```objc
> 
>   struct BinarySearchOptions {
>     double scale_max{10};
> ```

BTW: Should this scale now be in (0,1) by default? It depends on whether we are looking for a uniform contraction or expansion of our polytope.
___
*[`geometry/optimization/dev/cspace_free_polytope.cc` line 1471 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NL7Glo-Cuwnh0fxgzl-:-NLC6i1f2vL_00ZHA2zo:b-tzfn3i) ([raw file](https://github.com/robotlocomotion/drake/blob/1092df0cb74879b4b7d73ef54011fc4a709c0674/geometry/optimization/dev/cspace_free_polytope.cc#L1471)):*

So our C-space polytope is defined as {s | C*s<=d, s_lower <= s <= s_upper}. The distance from a face `C.row(i) *s <= d(i)` to a point s_center is computed as `(d(i) - C.row(i) * s_center) / |C.row(i)|`, hence the derivation above.
</blockquote></details>

I see. That is a reasonable way to perform the scaling. My derivation was
1) Center the polytope (i.e. rewrite the polytope so that `s_center` is the origin)
2) scale the RHS by epsilon
3) Return the polytope so that 0 is the origin.

Your method is reasonable so I am marking myself satisfied.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 355 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NLC6TXI818MJteJRC8O:-NLD_ZZe8MCQ2wjgAEu6:b-a3pjao) ([raw file](https://github.com/robotlocomotion/drake/blob/004939c5c8eb2a502bb2d7e532d2d40c6d958002/geometry/optimization/dev/cspace_free_polytope.h#L355)):*

BTW: Should this scale now be in (0,1) by default? It depends on whether we are looking for a uniform contraction or expansion of our polytope.
</blockquote></details>

I think the math for expansion (scale > 1) and contraction (scale < 1) is the same, hence I don't want to restrict us to (0, 1) or (1, inf). We almost always need to manually set this `scale_max` and `scale_min`, so I just put some non-zero number here as a default.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_polytope.h` line 355 at r4](https://reviewable.io/reviews/robotlocomotion/drake/18505#-NLC6TXI818MJteJRC8O:-NLE2Xpm8R9WNbMm3IJy:b-857ik0) ([raw file](https://github.com/robotlocomotion/drake/blob/004939c5c8eb2a502bb2d7e532d2d40c6d958002/geometry/optimization/dev/cspace_free_polytope.h#L355)):*

I think the math for expansion (scale > 1) and contraction (scale < 1) is the same, hence I don't want to restrict us to (0, 1) or (1, inf). We almost always need to manually set this `scale_max` and `scale_min`, so I just put some non-zero number here as a default.
</blockquote></details>

After thinking more about this, I think I see your point. One typical use case for the binary search, is to start with a C-space polytope found through nonlinear IRIS. The C-IRIS should provide a region smaller than the nonlinear IRIS, hence we would like to contract the polytope.

AlexandreAmice (commented):
:lgtm:

> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18505#-:-NLLp5w5-6P7Oh04KLLD:bnfp4nl)* status: missing label for release notes (waiting on @hongkai-dai)

PR #18460, opened 2022-12-15
Remove unneeded CRTP parameter from MultibodyElement

sherm1:
For some reason lost to history the internal class MultibodyElement used a CRTP pattern yet never actually referenced the recursive template at all. This PR removes that parameter and hopefully some unnecessary confusion along with it.

**Release note**: this is technically a breaking change since MultibodyElement is the non-internal base class for all the elements, both internal and visible. However it is very unlikely that anyone is referencing the base class directly. If they are, they'll have to make the same (easy) change to their code also.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18460)

PR #18522, opened 2023-01-03
[multibody] Repair some physical dependency defects in MbT

jwnimmer-tri:
The `MultibodyElement` header depends on `MultibodyTree`, so should be part of the main library, instead of pretending to be usable in isolation. This will allow for it to have a `*.cc` file down the road.

The `JointImplementBuilder` is a Tree--Joint interface so should not pollute `MultibodyElement`.

I noticed this while reading #18460.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18522)

### DIFF
--- a/multibody/tree/BUILD.bazel
+++ b/multibody/tree/BUILD.bazel
@@ -18,7 +18,6 @@ drake_cc_package_library(
     deps = [
         ":articulated_body_inertia",
         ":geometry_spatial_inertia",
-        ":multibody_element",
         ":multibody_tree_caches",
         ":multibody_tree_core",
         ":multibody_tree_indexes",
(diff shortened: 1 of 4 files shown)


jwnimmer-tri:
+@sherm1 for both(?) reviews, please.

sherm1 (commented):
:lgtm: x 2, thanks Jeremy!

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18522#-:-NKtrsa90dd_KaK3P3WL:b-s5kyip)* status: :shipit: complete! all discussions resolved, LGTM from assignee sherm1(platform) (waiting on @jwnimmer-tri)

PR #18529, opened 2023-01-04
[pydrake] Add __repr__ to several classes

### LINKED ISSUE #13987: pydrake doc's default arguments have memory hex addresses
A few places in our generated docs have hex memory addresses from `__str__` or `__repr__` embedded in the API signature:
- [FiniteHorizonLinearQuadraticRegulator](https://drake.mit.edu/pydrake/pydrake.systems.controllers.html#pydrake.systems.controllers.FiniteHorizonLinearQuadraticRegulator)
- [RgbdSensor](https://drake.mit.edu/pydrake/pydrake.systems.sensors.html?r#pydrake.systems.sensors.RgbdSensor.__init__)
- [RegionOfAttraction](https://drake.mit.edu/pydrake/pydrake.systems.analysis.html#pydrake.systems.analysis.RegionOfAttraction)
- etc.

It's usually from pybound C++ structs that don't have good `__str__` or `__repr__`.

We should make the objects print correctly, so the docs don't have line noise.  (The line noise churns our deployment diffs, also.)

jwnimmer-tri:
Closes #13987.

`RenderLabel` gets a round-trip `repr()`.

`NumericalGradientOption` lacks a getter function for one of its constructor arguments, so can only provide an incomplete `repr()`.

`SolverOptions` only provides a no-argument constructor, so cannot offer round-trip.  For the purpose of cleaning up our generated documentation, it's sufficient to just hard-code a return value in order to omit the hex address that would otherwise show up in our docs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529)

### DIFF
--- a/bindings/pydrake/geometry_py_render.cc
+++ b/bindings/pydrake/geometry_py_render.cc
@@ -300,6 +300,25 @@ void DoScalarIndependentDefinitions(py::module m) {
         .def(py::init<int>(), py::arg("value"), doc.RenderLabel.ctor.doc_1args)
         .def("is_reserved", &RenderLabel::is_reserved)
         .def("__int__", [](const RenderLabel& self) -> int { return self; })
+        .def("__repr__",
+            [](const RenderLabel& self) -> std::string {
+              if (self == RenderLabel::kEmpty) {
+                return "RenderLabel.kEmpty";
+              }
+              if (self == RenderLabel::kDoNotRender) {
+                return "RenderLabel.kDoNotRender";
+              }
+              if (self == RenderLabel::kDontCare) {
+                return "RenderLabel.kDontCare";
+              }
+              if (self == RenderLabel::kUnspecified) {
+                return "RenderLabel.kUnspecified";
+              }
+              if (self == RenderLabel::kMaxUnreserved) {
+                return "RenderLabel.kMaxUnreserved";
+              }
+              return fmt::format("RenderLabel({})", int{self});
+            })
         // EQ(==).
         .def(py::self == py::self)
         .def(py::self == int{})
--- a/bindings/pydrake/math_py.cc
+++ b/bindings/pydrake/math_py.cc
@@ -565,6 +565,10 @@ void DoScalarIndependentDefinitions(py::module m) {
         .def(
             "__repr__", [](const NumericalGradientOption& self) -> std::string {
               py::object method = py::cast(self.method());
+              // This is a minimal implementation that serves to avoid
+              // displaying memory addresses in pydrake docs and help strings.
+              // In the future, we should enhance this to display all of the
+              // information.
               return fmt::format(
                   "<NumericalGradientOption({})>", py::repr(method));
             });
--- a/bindings/pydrake/math_py.cc
+++ b/bindings/pydrake/math_py.cc
@@ -561,7 +561,13 @@ void DoScalarIndependentDefinitions(py::module m) {
             py::arg("function_accuracy") = 1E-15, cls_doc.ctor.doc)
         .def("NumericalGradientMethod", &Class::method, cls_doc.method.doc)
         .def("perturbation_size", &Class::perturbation_size,
-            cls_doc.perturbation_size.doc);
+            cls_doc.perturbation_size.doc)
+        .def(
+            "__repr__", [](const NumericalGradientOption& self) -> std::string {
+              py::object method = py::cast(self.method());
+              return fmt::format(
+                  "<NumericalGradientOption({})>", py::repr(method));
+            });
   }
 
   m.def(
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -524,6 +524,9 @@ void BindSolverInterfaceAndFlags(py::module m) {
       .def("get_print_to_console", &SolverOptions::get_print_to_console,
           doc.SolverOptions.get_print_to_console.doc)
       .def("__repr__", [](const SolverOptions&) -> std::string {
+        // This is a minimal implementation that serves to avoid displaying
+        // memory addresses in pydrake docs and help strings. In the future,
+        // we should enhance this to provide more details.
         return "<SolverOptions>";
       });
 
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -522,7 +522,10 @@ void BindSolverInterfaceAndFlags(py::module m) {
       .def("get_print_file_name", &SolverOptions::get_print_file_name,
           doc.SolverOptions.get_print_file_name.doc)
       .def("get_print_to_console", &SolverOptions::get_print_to_console,
-          doc.SolverOptions.get_print_to_console.doc);
+          doc.SolverOptions.get_print_to_console.doc)
+      .def("__repr__", [](const SolverOptions&) -> std::string {
+        return "<SolverOptions>";
+      });
 
   py::enum_<CommonSolverOption>(
       m, "CommonSolverOption", doc.CommonSolverOption.doc)
--- a/bindings/pydrake/test/geometry_render_test.py
+++ b/bindings/pydrake/test/geometry_render_test.py
@@ -78,6 +78,25 @@ class TestGeometryRender(unittest.TestCase):
         # Confirm value instantiation.
         Value[mut.render.RenderLabel]
 
+    def test_render_label_repr(self):
+        RenderLabel = mut.render.RenderLabel
+
+        # Special labels should use a non-numeric spelling.
+        special_labels = [
+            RenderLabel.kEmpty,
+            RenderLabel.kDoNotRender,
+            RenderLabel.kDontCare,
+            RenderLabel.kUnspecified,
+        ]
+        for label in special_labels:
+            self.assertIn("RenderLabel.k", repr(label))
+
+        # Any label should round-trip via 'eval'.
+        all_labels = special_labels + [RenderLabel(10)]
+        for label in all_labels:
+            roundtrip = eval(repr(label), dict(RenderLabel=RenderLabel))
+            self.assertEqual(label, roundtrip)
+
     def test_render_engine_api(self):
         class DummyRenderEngine(mut.render.RenderEngine):
             """Mirror of C++ DummyRenderEngine."""
--- a/bindings/pydrake/test/math_test.py
+++ b/bindings/pydrake/test/math_test.py
@@ -504,6 +504,8 @@ class TestMath(unittest.TestCase):
             method=mut.NumericalGradientMethod.kCentral,
             function_accuracy=1E-15)
 
+        self.assertIn("kCentral", repr(option))
+
         def foo(x):
             return np.array([x[0] ** 2, x[0] * x[1]])
 


SeanCurtis-TRI (commented):
:LGTM: With a couple of questions.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529#-:-NKz149iELOhiSju8PbW:b-ti15gs)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/math_py.cc` line 569 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKyyfgb7EoPxbjEA9zV:-NKyyfgb7EoPxbjEA9zW:b-uns2nm) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/math_py.cc#L569)):*
> ```cpp
>               py::object method = py::cast(self.method());
>               return fmt::format(
>                   "<NumericalGradientOption({})>", py::repr(method));
> ```

btw: The `function_accuracy` could be reverse engineered.

You do have the getter for `perturbation_size()` and the input parameter is either that quantity squared or cubed, depending on the value of `method`.

Is there a reason to not do that?  (Of course, it'd be better if `Numerical GradientOption` did that heavy lifting *for* you. But, at the very least, a TODO would be nice.) Admittedly, someone who uses the default accuracy will find themselves happy with the round-trip result.

It seems this partial `repr()` implementation lands in an unhappy spot. If a full `__repr__` were impossible, I could fully get on board.  But, as it's merely inconvenient, it's not as obvious that the inconvenience justifies an incomplete implementation.
___
*[`bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc` line 527 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKz-6yP2QiRLiIxPKQb:-NKz-6yP2QiRLiIxPKQc:b-yqpvhm) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc#L527)):*
> ```cpp
>           doc.SolverOptions.get_print_to_console.doc)
>       .def("__repr__", [](const SolverOptions&) -> std::string {
>         return "<SolverOptions>";
> ```

nit: This seems a strange choice; `repr(SolverOptions())` returns as a *type* indicator rather than an expression for creating an instance.

This satisfies a method for cleaning up the documentation, but why *this* method? It's not obvious why `<SolverOptions>` is superior to `SolverOptions()`. Is that a convention that should generally be clear and I'm just missing something, or does this bespoke representation deserve a comment?

SeanCurtis-TRI (commented):
+@seancurtis-tri 

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529#-:-NKz2bBLEPt3R18tuwqP:b-l9yj1x)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529)* status: 2 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/math_py.cc` line 569 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKyyfgb7EoPxbjEA9zV:-NKzC7VrDWrnAA6TiDaC:bea53dq) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/math_py.cc#L569)):*
> The function_accuracy could be reverse engineered.

Yes.  Doing that reverse math _in the bindings_ though would be a defect -- arms-length (brittle) coupling.  Someone could change the C++ code and fail to update the bindings (maybe depending on the unit test it might be found or missed).

So that leaves us with adding a C++ getter function on the class, to match the constructor argument.  That would be a fine thing, but beyond the scope of removing spam from the pydrake help docs.

> ... not as obvious that the inconvenience justifies an incomplete implementation.

It's a step in the right direction -- displaying the enum value is the piece users will care about the most.

If the original author didn't see the need to add a C++ getter to replay ctor argument, that's evidence that it's unimporant to print it back to the user.  I suppose that's an argument for adding the getter that we _do_ have to the repr, but _shrug_.  Again, my only goal is to remove hex spam from help strings, and this accomplishes it.
___
*[`bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc` line 527 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKz-6yP2QiRLiIxPKQb:-NKz7uNy3KrWEmrpb9TC:b-sjiaga) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc#L527)):*

nit: This seems a strange choice; `repr(SolverOptions())` returns as a *type* indicator rather than an expression for creating an instance.

This satisfies a method for cleaning up the documentation, but why *this* method? It's not obvious why `<SolverOptions>` is superior to `SolverOptions()`. Is that a convention that should generally be clear and I'm just missing something, or does this bespoke representation deserve a comment?
</blockquote></details>

There is a very particular contract that `repr()` is supposed to meet:

https://docs.python.org/3/reference/datamodel.html#object.__repr__

https://docs.python.org/3/library/functions.html#repr

SeanCurtis-TRI (commented):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529)* status: needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/math_py.cc` line 569 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKyyfgb7EoPxbjEA9zV:-NL13Ry_ATMzbKImerlo:btmygd0) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/math_py.cc#L569)):*
From the link:

>  If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

I'd point out a couple of things:

1. "If __*at all*__ possible...used to recreate an object with the same value..." (emphasis mine) is an interesting turn of phrase. It suggests a strong obligation towards extraordinary effort (whether the authors *intended* that implication or not is a different debate). If one were to subscribe to this point of view, then adding the getter to the underlying class may be the right way to go.
2. Arguably, the current value is a "useful description" in place of the recreating expression. I question that utility. I find this implementation misleading. Sometimes the expression is sufficient, sometimes not. But its *format* suggests that it always is with. I'd vote we push it to one side or the other -- either return a type like `SolverOptions` or add the getter on the C++ side to support proper `repr()` implementation.

That said, the documentation implies the missing parameter is seldom the non-default value and this implementation provides utility for the common case. I can't judge that assertion. I've made some arguments and leave it to your discretion.
___
*[`bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc` line 527 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKz-6yP2QiRLiIxPKQb:-NL1I-gs1_r5JP6kme1f:bo2dzym) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc#L527)):*

There is a very particular contract that `repr()` is supposed to meet:

https://docs.python.org/3/reference/datamodel.html#object.__repr__

https://docs.python.org/3/library/functions.html#repr

</blockquote></details>

I can fully support this. It doesn't pretend to generate an instance with the same values because such an act is impossible.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529)* status: needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/math_py.cc` line 569 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKyyfgb7EoPxbjEA9zV:-NL1Q9t74LNLJB-ZTV13:b-3z3wwe) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/math_py.cc#L569)):*
> ... at all possible ... adding the getter to the underlying class may be the right way to go.

I agree that the current implementation is incomplete, and does not meet the goal of what a `repr()` should do to truly be Pythonic.  It's just _less_ incomplete than before, so to my view it's worth landing an incremental improvement, even if it's not perfect yet.

> But its format suggests that it always is ...

You're saying that Python users might not realize that the `<>` means the information is incomplete?  I suppose that's possible at would be the best reason to do more work here.

I think my premise is that basically few to no users will ever print out this object, really the only place it matters is docs, which is basically fine at the moment.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529)* status: needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/math_py.cc` line 569 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18529#-NKyyfgb7EoPxbjEA9zV:-NL1Ugup08BqIUWZWm_8:b3o9gj1) ([raw file](https://github.com/robotlocomotion/drake/blob/17e2c18e9a010eaa554cb09a12e473e447143aba/bindings/pydrake/math_py.cc#L569)):*

> ... at all possible ... adding the getter to the underlying class may be the right way to go.

I agree that the current implementation is incomplete, and does not meet the goal of what a `repr()` should do to truly be Pythonic.  It's just _less_ incomplete than before, so to my view it's worth landing an incremental improvement, even if it's not perfect yet.

> But its format suggests that it always is ...

You're saying that Python users might not realize that the `<>` means the information is incomplete?  I suppose that's possible at would be the best reason to do more work here.

I think my premise is that basically few to no users will ever print out this object, really the only place it matters is docs, which is basically fine at the moment.
</blockquote></details>

Oops...I'd somehow glossed over the fact that you've got both `<>` and `Foo(value)`. I'm retracting my comment. My statement was predicated on the idea that it was a legitimately a string that would *sometimes* recreate the instance. Obviously, with `<>` that's never true. So, as such, it *is* useful and only misleading to the people who don't read carefully -- certainly, no one who tried `eval()`ing the result of `repr()` would be able to maintain that misconception (should they happen to stupidly misread it in the first place).

jwnimmer-tri:

sammy-tri (commented):
:lgtm:

> Reviewed 3 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18529#-:-NL2IwNvAaG8zQ20dtPD:bnfp4nl)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

PR #18537, opened 2023-01-05
[workspace] Fix conex build with modern compilers

jwnimmer-tri:
At the bottom of [conex/debug_macros.h](https://github.com/ToyotaResearchInstitute/conex/blob/main/conex/debug_macros.h), we see a use of `std::array` without a matching `#include` atop the file.

See also https://github.com/ToyotaResearchInstitute/conex/pull/3.

My repro on Jammy was `bazel test --config=clang --compilation_mode=dbg --config=tsan //geometry/optimization:iris_test` but I think most Jammy Clang Debug builds will hit this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18537)

### DIFF
--- a/tools/workspace/conex/patches/debug_macros.patch
+++ b/tools/workspace/conex/patches/debug_macros.patch
@@ -1,0 +1,12 @@
+Add a missing #include statement
+
+See https://github.com/ToyotaResearchInstitute/conex/pull/3
+
+--- conex/debug_macros.h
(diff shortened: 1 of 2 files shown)


jwnimmer-tri:
+@rpoyner-tri for both reviews, please.

\CC @frankpermenter FYI

PR #18538, opened 2023-01-05
[multibody] Fix misuse of forward-declared classes

jwnimmer-tri:
In the member functions of a non-templated class, we must see definitions for all the classes used as member fields prior to defining the member functions that interact with them.

To break the cyclic dependency, move several function definitions to the cc file.

My repro on Jammy was `bazel test --config=clang --compilation_mode=dbg --config=tsan //geometry/optimization:iris_test` but I think most Jammy Clang Debug builds will hit this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18538)

### DIFF
--- a/multibody/topology/multibody_graph.cc
+++ b/multibody/topology/multibody_graph.cc
@@ -9,6 +9,11 @@ namespace drake {
 namespace multibody {
 namespace internal {
 
+MultibodyGraph::MultibodyGraph(const MultibodyGraph&) = default;
+MultibodyGraph& MultibodyGraph::operator=(const MultibodyGraph&) = default;
+MultibodyGraph::MultibodyGraph(MultibodyGraph&&) = default;
+MultibodyGraph& MultibodyGraph::operator=(MultibodyGraph&&) = default;
+
 MultibodyGraph::MultibodyGraph() {
   RegisterJointType(weld_type_name());
   // Verify invariant promised to users in the documentation.
@@ -159,6 +164,28 @@ JointIndex MultibodyGraph::AddJoint(const std::string& name,
 
 int MultibodyGraph::num_joint_types() const {
   return static_cast<int>(joint_type_name_to_index_.size());
+}
+
+int MultibodyGraph::num_bodies() const {
+  return static_cast<int>(bodies_.size());
+}
+
+int MultibodyGraph::num_joints() const {
+  return static_cast<int>(joints_.size());
+}
+
+const MultibodyGraph::Body& MultibodyGraph::get_body(BodyIndex index) const {
+  DRAKE_THROW_UNLESS(index < num_bodies());
+  return bodies_[index];
+}
+
+MultibodyGraph::Body& MultibodyGraph::get_mutable_body(BodyIndex body_index) {
+  return bodies_[body_index];
+}
+
+const MultibodyGraph::Joint& MultibodyGraph::get_joint(JointIndex index) const {
+  DRAKE_THROW_UNLESS(index < num_joints());
+  return joints_[index];
 }
 
 JointTypeIndex MultibodyGraph::RegisterJointType(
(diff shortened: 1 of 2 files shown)


sherm1 (commented):
:lgtm: x 2 pending CI
+@sherm1 for both reviews

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18538#-:-NL1pYLS0fSZ18rdDM7s:b-zdmr6m)* status: 2 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`multibody/topology/multibody_graph.h` line 41 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18538#-NL1ohpE1qYps5im8mvL:-NL1ohpE1qYps5im8mvM:bxbd2p0) ([raw file](https://github.com/robotlocomotion/drake/blob/8161280adbd4a92cb46a66d0eabfc4dc9fdda472/multibody/topology/multibody_graph.h#L41)):*
> ```objc
>   MultibodyGraph& operator=(const MultibodyGraph&);
>   MultibodyGraph(MultibodyGraph&&);
>   MultibodyGraph& operator=(MultibodyGraph&&);
> ```

BTW consider moving this block to where the macro was to emphasize that it is just the "declare" version of the conventional macro. (If we had a suitable DRAKE_DECLARE... macro that's where we'd put it [not suggesting we need one!])
___
*[`multibody/topology/multibody_graph.cc` line 183 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18538#-NL1pNHt3-DDJqva7N1j:-NL1pNHt3-DDJqva7N1k:b7wc0wv) ([raw file](https://github.com/robotlocomotion/drake/blob/8161280adbd4a92cb46a66d0eabfc4dc9fdda472/multibody/topology/multibody_graph.cc#L183)):*
> ```cpp
> 
> MultibodyGraph::Body& MultibodyGraph::get_mutable_body(BodyIndex body_index) {
>   return bodies_[body_index];
> ```

BTW odd that this doesn't have the same error check as the const case.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18538)* status: :shipit: complete! all discussions resolved, LGTM from assignee sherm1(platform) (waiting on @sherm1)

___
*[`multibody/topology/multibody_graph.h` line 41 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18538#-NL1ohpE1qYps5im8mvL:-NL27cNb4KfI5Zdw46UF:b-896fix) ([raw file](https://github.com/robotlocomotion/drake/blob/8161280adbd4a92cb46a66d0eabfc4dc9fdda472/multibody/topology/multibody_graph.h#L41)):*

BTW consider moving this block to where the macro was to emphasize that it is just the "declare" version of the conventional macro. (If we had a suitable DRAKE_DECLARE... macro that's where we'd put it [not suggesting we need one!])
</blockquote></details>

Done.
___
*[`multibody/topology/multibody_graph.cc` line 183 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18538#-NL1pNHt3-DDJqva7N1j:-NL26SBhEVd5IL7jqZWu:bxk62uv) ([raw file](https://github.com/robotlocomotion/drake/blob/8161280adbd4a92cb46a66d0eabfc4dc9fdda472/multibody/topology/multibody_graph.cc#L183)):*

BTW odd that this doesn't have the same error check as the const case.
</blockquote></details>

Yup.  Maybe the thought was that it's `private:` so it's always pre-validated somehow.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18538)* status: :shipit: complete! all discussions resolved, LGTM from assignee sherm1(platform) (waiting on @jwnimmer-tri)

Issue #18091, opened 2022-10-12
Remaining Jammy CI builds


jwnimmer-tri:
#16217 tracked adding Jammy support. In that issue, passing jobs were moved into production.  This issue covers the remainder.

We still have four Jammy jobs that are excluded from Production:

```
linux-jammy-clang-bazel-weekly-non-production-gurobi-thread-sanitizer
linux-jammy-clang-bazel-weekly-non-production-mosek-thread-sanitizer
linux-jammy-clang-bazel-weekly-non-production-snopt-thread-sanitizer
linux-jammy-clang-bazel-weekly-non-production-mosek-valgrind-memcheck
```

We should work to either enable them, or delete them as out-of-scope.

I suspect that #17562 is the only blocker for turning on the thread sanitizer builds.

I'm not sure if we've looked into the mosek-valgrind one.

BetsyMcPhail:
When the DockerHub password was updated this summer, only the production jobs were updated. https://github.com/RobotLocomotion/drake-jenkins-jobs/pull/58 also updates the the non-production jobs.

jwnimmer-tri:
See #18098 for a possible fix for TSan.

BetsyMcPhail:
There is one failing test on `linux-jammy-clang-bazel-weekly-non-production-mosek-valgrind-memcheck` - [//solvers:mosek_solver_internal_test](https://drake-cdash.csail.mit.edu/viewTest.php?onlyfailed&buildid=1659179). With the goal of moving Jammy jobs to production, @williamjallen is going to skip memcheck on that test.

BetsyMcPhail:
The `linux-jammy-clang-bazel-weekly-non-production-mosek-valgrind-memcheck` job has been moved to production.

@jwnimmer-tri are we ready to move the other jobs into production? (I'm updating the Jammy AMI  now for #18541)

jwnimmer-tri:
> @jwnimmer-tri are we ready to move the other jobs into production?

We should just run them in CI and see whether they have errors (after #18541 is in effect).

BetsyMcPhail:
Experimental jobs using updated Jammy AMI:

https://drake-jenkins.csail.mit.edu/view/Linux%20Jammy/job/linux-jammy-clang-bazel-experimental-gurobi-thread-sanitizer/2/
https://drake-jenkins.csail.mit.edu/view/Linux%20Jammy/job/linux-jammy-clang-bazel-experimental-mosek-thread-sanitizer/2/
[...]

PR #18539, opened 2023-01-05
Suppress Valgrind false positive in mosek_solver_internal_test

williamjallen:
The test `solvers/mosek_solver_internal_test` is currently failing under Valgrind on Jammy.  To speed our migration to Jammy, Valgrind has been disabled for this test.

This change fixes one of the issues listed in https://github.com/RobotLocomotion/drake/issues/18091.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539)

### DIFF
--- a/tools/dynamic_analysis/valgrind.supp
+++ b/tools/dynamic_analysis/valgrind.supp
@@ -19,7 +19,6 @@
     MOSEK invalid memory access #3
     Memcheck:Leak
     match-leak-kinds: possible
-    match-leak-kinds: possible
     ...
     fun:moseklicense_checkout
 }
--- a/tools/dynamic_analysis/valgrind.supp
+++ b/tools/dynamic_analysis/valgrind.supp
@@ -14,6 +14,15 @@
     fun:checkout_from_flexlm
 }
 
+
+{
+    MOSEK invalid memory access #3
+    Memcheck:Leak
+    match-leak-kinds: possible
+    ...
+    fun:_ZN7testing4Test3RunEv
+}
+
 {
    MOSEK use of uninitialized CPU condition code #1
    Memcheck:Cond
--- a/tools/dynamic_analysis/valgrind.supp
+++ b/tools/dynamic_analysis/valgrind.supp
@@ -19,7 +19,20 @@
     MOSEK invalid memory access #3
     Memcheck:Leak
     match-leak-kinds: possible
-    ...
+    match-leak-kinds: possible
+    fun:calloc
+    fun:calloc
+    fun:allocate_dtv
+    fun:_dl_allocate_tls
+    fun:allocate_stack
+    fun:lc_checkout
+    fun:moseklicense_flexlmcheckout
+    fun:moseklicense_checkout
+    fun:MSKP_islicensed
+    fun:MSK_optimize
+    fun:MSK_optimizetrm
+    fun:_ZN5drake7solvers8internal38AddQuadraticCostAsLinearCost_Test_Test8TestBodyEv
+    fun:_ZN7testing8internal35HandleExceptionsInMethodIfSupportedINS_4TestEvEET0_PT_MS4_FS3_vEPKc
     fun:_ZN7testing4Test3RunEv
 }
 
--- a/tools/dynamic_analysis/valgrind.supp
+++ b/tools/dynamic_analysis/valgrind.supp
@@ -20,20 +20,8 @@
     Memcheck:Leak
     match-leak-kinds: possible
     match-leak-kinds: possible
-    fun:calloc
-    fun:calloc
-    fun:allocate_dtv
-    fun:_dl_allocate_tls
-    fun:allocate_stack
-    fun:lc_checkout
-    fun:moseklicense_flexlmcheckout
+    ...
     fun:moseklicense_checkout
-    fun:MSKP_islicensed
-    fun:MSK_optimize
-    fun:MSK_optimizetrm
-    fun:_ZN5drake7solvers8internal38AddQuadraticCostAsLinearCost_Test_Test8TestBodyEv
-    fun:_ZN7testing8internal35HandleExceptionsInMethodIfSupportedINS_4TestEvEET0_PT_MS4_FS3_vEPKc
-    fun:_ZN7testing4Test3RunEv
 }
 
 {


williamjallen:
@drake-jenkins-bot linux-jammy-clang-bazel-experimental-mosek-valgrind-memcheck please.

williamjallen:
@drake-jenkins-bot linux-jammy-clang-bazel-experimental-mosek-valgrind-memcheck please

williamjallen (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539)* status: 1 unresolved discussion, LGTM missing from assignee BetsyMcPhail, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @BetsyMcPhail)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/18539#-NL2jubI9WXng44UDDSP:-NL2jubI9WXng44UDDSQ:b9rse0m) (no related file):*
+@betsymcphail for review

___

BetsyMcPhail (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539#-:-NL2lguI6ZjvER09UdRN:bnfp4nl)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

BetsyMcPhail (commented):
+@sammy-tri for platform review please

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539#-:-NL2mIwrE3RE-pRJjqs_:b-6xrt9v)* status: 1 unresolved discussion, LGTM missing from assignee sammy-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @sammy-tri)

sammy-tri (commented):
We have several mosek suppressions in `tools/dynamic_analysis/valgrind.supp`.  Do we think they're still valid?  Also what's the rationale for disabling memcheck entirely instead of adding a new suppression?

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539#-:-NL2nu0l4ASCGfqbGd20:buin65e)* status: LGTM missing from assignee sammy-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @williamjallen)

BetsyMcPhail (commented):
Thanks for the suggestion, I completely forgot about the suppression file when working on this. That does seem like the better approach.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539#-:-NL2q24E-995Tncolfir:bg4s2nx)* status: LGTM missing from assignee sammy-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @williamjallen)

williamjallen:
@drake-jenkins-bot linux-jammy-clang-bazel-experimental-mosek-valgrind-memcheck please

jwnimmer-tri:
*[`tools/dynamic_analysis/valgrind.supp` line 23 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18539#-NL3FP-062myXPOrZQQ9:-NL3FP-062myXPOrZQQA:bb41k37) ([raw file](https://github.com/robotlocomotion/drake/blob/b80569b7e35c91a1f0a8fedb51a68cc2ae7268ce/tools/dynamic_analysis/valgrind.supp#L23)):*
> ```
>     match-leak-kinds: possible
>     ...
>     fun:_ZN7testing4Test3RunEv
> ```

This matching pattern is way too wide.  It will in effect disable Memcheck on every single test in Drake.

The match should be on something like `fun:moseklicense_checkout`, or anything else in https://drake-cdash.csail.mit.edu/test/924782530 that seems specific to this exact control flow path.

williamjallen:
@drake-jenkins-bot linux-jammy-clang-bazel-experimental-mosek-valgrind-memcheck please

jwnimmer-tri:
*[`tools/dynamic_analysis/valgrind.supp` line 23 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18539#-NL3FP-062myXPOrZQQ9:-NL3LHIr8GPxq1VsBybj:brs5d06) ([raw file](https://github.com/robotlocomotion/drake/blob/b80569b7e35c91a1f0a8fedb51a68cc2ae7268ce/tools/dynamic_analysis/valgrind.supp#L23)):*

This matching pattern is way too wide.  It will in effect disable Memcheck on every single test in Drake.

The match should be on something like `fun:moseklicense_checkout`, or anything else in https://drake-cdash.csail.mit.edu/test/924782530 that seems specific to this exact control flow path.
</blockquote></details>

Now it's the other end of Goldilocks.  The exact function names of "allocate_stack, allocate_rls, allocate_dtv" are subject to change as libc6 evolves, or different heap conditions are in effect.  Never mind that if a developer renames the test case, this suppression will no longer match but nobody will notice until the nightly build tips over.

The goal of a suppression should be to have _as few frames as possible_ while matching _only the problematic conditions_.  In this case that means something like the 1 fun I mentioned above -- any leaks _that happen within the license manager_ are of no consequence for us, since we can't actually fix them (MOSEK is closed-source).  It doesn't matter which test case name first called into the license manager -- we should suppress the message for all test cases.

williamjallen (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539)* status: 1 unresolved discussion, LGTM missing from assignee sammy-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @BetsyMcPhail, @jwnimmer-tri, and @sammy-tri)

___
*[`tools/dynamic_analysis/valgrind.supp` line 23 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18539#-NL3FP-062myXPOrZQQ9:-NL3QCah7Y8R7sgmCxqz:b-eqf1n4) ([raw file](https://github.com/robotlocomotion/drake/blob/b80569b7e35c91a1f0a8fedb51a68cc2ae7268ce/tools/dynamic_analysis/valgrind.supp#L23)):*

Now it's the other end of Goldilocks.  The exact function names of "allocate_stack, allocate_rls, allocate_dtv" are subject to change as libc6 evolves, or different heap conditions are in effect.  Never mind that if a developer renames the test case, this suppression will no longer match but nobody will notice until the nightly build tips over.

The goal of a suppression should be to have _as few frames as possible_ while matching _only the problematic conditions_.  In this case that means something like the 1 fun I mentioned above -- any leaks _that happen within the license manager_ are of no consequence for us, since we can't actually fix them (MOSEK is closed-source).  It doesn't matter which test case name first called into the license manager -- we should suppress the message for all test cases.
</blockquote></details>

My thought was that we would want to restrict it to the most minimal code path possible.  I now see that it's a library call which could get called again (and then need to be ignored again) at some point in the future.  I have made the suggested updates.

williamjallen:
@drake-jenkins-bot linux-jammy-clang-bazel-experimental-mosek-valgrind-memcheck please

jwnimmer-tri (commented):
I can do platform +@jwnimmer-tri -@sammy-tri.

:lgtm: modulo CI results.

> Reviewed 1 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18539#-:-NL3U4JS6zSRu0gUaGQG:b-a275o0)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @williamjallen)

___
*[`tools/dynamic_analysis/valgrind.supp` line 23 at r3](https://reviewable.io/reviews/robotlocomotion/drake/18539#-NL3FP-062myXPOrZQQ9:-NL3Ts3e-44x1JrR1HV2:b-k7gkyh) ([raw file](https://github.com/robotlocomotion/drake/blob/b80569b7e35c91a1f0a8fedb51a68cc2ae7268ce/tools/dynamic_analysis/valgrind.supp#L23)):*
<details><summary><i>Previously, williamjallen (William Allen) wrote…</i></summary><blockquote>

My thought was that we would want to restrict it to the most minimal code path possible.  I now see that it's a library call which could get called again (and then need to be ignored again) at some point in the future.  I have made the suggested updates.
</blockquote></details>

Yup. We want it specific enough to not risk false negatives, but also general enough to avoid foreseeable false positives down the road.

PR #18356, opened 2022-11-23
[framework] Add unittest showing confusing discrete initialization

EricCousineau-TRI:
Relates #12649

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18356)

PR #18551, opened 2023-01-08
[WIP] Handling "convergent" discrete initialization

EricCousineau-TRI:
One possible venue towards suggestion in #18356
FYI @sherm1 @siyuanfeng-tri

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18551)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,7 @@
 # -*- python -*-
 
+DISABLE CI
+
 # This file marks a workspace root for the Bazel build system. see
 # https://bazel.build/ .
 
--- a/systems/analysis/simulator.cc
+++ b/systems/analysis/simulator.cc
@@ -77,14 +77,34 @@ SimulatorStatus Simulator<T>::Initialize(const InitializeParams& params) {
 
   // Process all the initialization events.
   merged_events_ = system_.AllocateCompositeEventCollection();
+
   if (!params.suppress_initialization_events) {
     system_.GetInitializationEvents(*context_, merged_events_.get());
+  } else {
+    DRAKE_THROW_UNLESS(!params.toposort_initialization_events);
+    DRAKE_DEMAND(merged_events_->Empty());
   }
 
-  // Do unrestricted updates first.
-  HandleUnrestrictedUpdate(merged_events_->get_unrestricted_update_events());
-  // Do restricted (discrete variable) updates next.
-  HandleDiscreteUpdate(merged_events_->get_discrete_update_events());
+  if (!params.toposort_initialization_events) {
+    // Do unrestricted updates first.
+    HandleUnrestrictedUpdate(merged_events_->get_unrestricted_update_events());
+    // Do restricted (discrete variable) updates next.
+    HandleDiscreteUpdate(merged_events_->get_discrete_update_events());
+  } else {
+    std::vector<std::unique_ptr<CompositeEventCollection<T>>>
+        event_phases;
+    // How to recurse into inner diagrams?
+    // How to control level of recursion?
+    system_->ToposortAndPartitionInitializationEvents(
+        *merged_events_,
+        param.toposort_initialization_strict,
+        &event_phases);
+    // Process phases.
+    for (auto& event_phase : event_phases) {
+      HandleUnrestrictedUpdate(event_phases->get_unrestricted_update_events());
+      HandleDiscreteUpdate(event_phases->get_discrete_update_events());
+    }
+  }
 
   // Gets all per-step events to be handled.
   per_step_events_ = system_.AllocateCompositeEventCollection();
--- a/systems/analysis/simulator.h
+++ b/systems/analysis/simulator.h
@@ -38,6 +38,19 @@ struct InitializeParams {
   /// Whether to trigger initialization events. Events are triggered by
   /// default; it may be useful to suppress them when reusing a simulator.
   bool suppress_initialization_events{false};
+
+  // TODO(eric.cousineau): Is there a way to make this more granular, or happen
+  // outside of Simulator (e.g. at different Diagram levels)?
+  /// Perform topological sort of discrete and unrestricted initialization
+  /// events to allow for simultaneous initialization events. This is done
+  /// by partitioning said events into a set of (minimum) phases to process in
+  /// order.
+  bool toposort_initialization_events{false};
+
+  /// Require topological sort be free of all cycles. If true, will throw an
+  /// error. If false, partitioning may have simultaneous, interdependent
+  /// events that initialize the system to an invalid state.
+  bool toposort_initialization_strict{true};
 };
 
 /** @ingroup simulation


sherm1 (commented):
A few questions:
- Would this be applied only to initialization events but not other simultaneous event circumstances (e.g. multiple events with the same period)? If so, why?
- Can this be implemented with the information we have? We don't know much about what gets affected by an Event. Or is this just sorting by System input/output relationships? If so, what happens in the common case where there are loops in that relationship, e.g. for MbP and SG?
- What are the downsides if any of this? Currently we have the promise that every simultaneous event sees the same Context values  when it is handled, with the Context updated only after all events have had their say. I presume the intention here is that the Context would be updated multiple times and the events handled sequentially. Is that always the best policy? 
- Maybe it would be helpful to provide a real example where the current approach doesn't work and the proposed one is better. IIRC #18356 was a case where initialization events had not been used properly but could have been. (ZOH in particular isn't set up to use initialization events but should be.)

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18551#-:-NLIYuXX8h-6SQ70P7om:b-8qvng9)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

sherm1:
(Discussed briefly with @edrumwri -- we both have concerns about changing the underlying event system abstractions which were set after several years of struggling to find the right ones.)

If it does turn out that we really need to provide for a sequenced initialization (that is, occurring at the same time but _not_ simultaneous) let's consider a simpler mechanism that doesn't mess with the basic Event abstractions. Would something like this work?
- User declares Initialization Events with an optional sequence number (a small integer)
- Simulator initialization handles all of a Diagram's "sequence 0" Events simultaneously, then updates Context
- Repeat that for "sequence 1" Events, etc. up to the highest sequence number that appears in the Diagram

That's a more explicit mechanism that assumes a user has a particular initialization sequence in mind and can define what should come first. It would avoid a search and problems with loops and would retain the current treatment of simultaneity. It could also be generalized to sequentialize other simultaneous events (non-initialization I mean) if that turned out to be useful.

sherm1:
I recognize the hypothetical problem with user-assigned sequence numbers but I'm wondering whether that would be a problem in practice. For example, if a typical application has a bunch of independent initializations and a few that need to wait, the sequence numbers might only be 0 and 1. My thought is that the situation might be clear for any given Diagram so the small integers would be sufficient, but I don't know. Might be worth trying first.

jwnimmer-tri:
In the specific cases where the current initialization sequence doesn't work for Eric and Evan (i.e., where certain subsystems' events need to see the post-initializaion-event output of other systems), are all of the initialization event handlers idempotent?  Assuming yes, would a simple solution (or at lease work-around) be for those users to invoke the Diagram initialization a few times in a row, so that it will reach a fixed point?

sherm1:
That's a good thought -- I do think multiple calls to Simulator::Initialize() would achieve the right end. Possible glitches:
- later-stage initializations might at first see NaNs or other invalid inputs and would need to be tolerant until those are corrected
- simulations that don't make explicit Initialize() calls will get a single call only at the first AdvanceTo() call
- users of a Diagram must know how many times to call Initialize(). A few extra calls wouldn't hurt though so maybe just routinely making several calls is good enough

WDYT @EricCousineau-TRI and @edrumwri ?

jwnimmer-tri:
I should also probably clarify my framing -- my thought is that if we're considering an incomplete / unsound solution like the magic integers, the "call Initialize a few times" might be a _simpler_ incomplete solution.

There might still be a need for the whole shebang where initialization events declare dependencies (and especially, explicitly state whether they want the pre-init vs post-init data present on their input ports).  But for that level of complexity, we need to renew the call for fully-worked, real-world examples to be presented.

edrumwri:
> In the specific cases where the current initialization sequence doesn't work for Eric and Evan (i.e., where certain subsystems' events need to see the post-initializaion-event output of other systems), are all of the initialization event handlers idempotent? Assuming yes, would a simple solution (or at least work-around) be for those users to invoke the Diagram initialization a few times in a row, so that it will reach a fixed point?

It's an interesting idea, but it wouldn't work for us out of the box: we'd need to do other changes on our end to make it viable. The main problem is that we have some initialization events that send hardware signals and should only be run once. So we actually allow initialization to only happen once in our code.

EricCousineau-TRI:
> Digressive complaint: [...]

My use case is simple unilateral (or bilateral) coupling, e.g. I have one system taking actual values in, producing desired values out, then a controller that takes those desired values in. I implement "discreteness" in the same vein as `RgbdSensor` - the actual system is nominally direct feedforward, and state is stored on ZOH on the output.

In this case, both systems should run at a discrete interval, thus there is an effective second-order cascade of ZOH as shown in #18356.

> [...] are all of the initialization event handlers idempotent [...]

Yeah, mentioned that in [this discussion](https://github.com/RobotLocomotion/drake/pull/18356#pullrequestreview-1196759542), "Dumbest approach is keep initializing until all values stop changing [...]"

It sounds OK for vector-based state. However, I am also using abstract values, and disprefer "reshaping" my data as a vector.

If we have something to enable `AbstractValue.EqualTo(other)` in a way that isn't a burden to implement, then sure.
However, I think it's instead better to be able to partition updates.

Regarding approaches to implement partitioning, I guess the "prescribe magic numbers; add option to automatic magic numbers" sounds OK to me; ultimately, I don't care, but I would like some affordance to:

- Know that there is possibly a cascade of dependent initializations; and
- Have some mechanism to toposort / whatever to make those initializations consistent.

I think we could do a first step which is simply diagram-level cascade analysis; and I believe it could be coded up to then admit more granular dependency carving.

jwnimmer-tri:
> It sounds OK for vector-based state. However, I am also using abstract values, and disprefer "reshaping" my data as a vector.

In my suggestion for a work-around, I didn't mean that the _framework_ automatically finds the fixed point.  I meant that your end-user application calls `Initialize` in a `for` loop, manually. Then you can still use abstract state, you just need to know that _in your case_ it will take 2 or 3 or whatever updates for everything to flow though -- or just run it 10 times and assume that's plenty.  That's basically the same kind of limitation as giving magic numbers to events, which is why I suggested it as a simpler workaround than the `int` tagging.

> Have some mechanism to toposort / whatever to make those initializations consistent.

If this ever starts into a design phase, keep in mind that not all initialization events necessarily want the post-init version of their input ports.  When adding the "declared dependencies" concept to an initialization event, that declaration needs to admit both pre-image and post-image dependencies for the inputs.

EricCousineau-TRI:
> [...] I meant that your end-user application calls Initialize in a for loop, manually. [...]

Will try that. I thought that initialization events would "statefully exhaust", e.g. I call `Initialize()` twice but effects are only seen first time.

That being said, good to simplify confirm!

EricCousineau-TRI:
Ooh, that worked! Confirmed in 8cd8a6ae19 in #18356.

Given this, I'd be happy to "just use this", without needing anything fancier. (didn't realize you could re-call `.Initialize()`).

I think the only nicety that'd be nice to have is:
- Provide analysis for *conservative* lower bound of discrete initialization cascade
- Consider warning (or failing) if this discrete initialization cascade is >1
  - Alternatively, just fold this directly into `simulator.Initialize()`

What do y'all think?

jwnimmer-tri:
> Provide analysis for conservative lower bound on order of discrete initialization cascade.

I'm not sure how to do this without first adding declared dependencies to initialization events.  I suppose one way would be to provide a getter for the total count of declared initialization events within a diagram (transitively)?  In the absence of any cycles, I _think_ that would be safe.  It seems like it would be easy enough to implement.

>  Alternatively, just fold this directly into simulator.Initialize()

I'm skeptical about running initialization events multiple times _by default_.  I think users would expect them to only be run once.  Only in the case where they are idempotent and always want the post-image flavor of their inputs (i.e., like your ZOH cascade) does the "run them a bunch of times" work-around apply.

EricCousineau-TRI:
>I suppose one way would be to provide a getter for the total count of declared initialization events within a diagram (transitively)?

Yup, that'd satisfy the (very) conservative bound.

> I'm skeptical about running initialization events multiple times by default.

That makes sense, but I think it is still good to possibly warn users a about the possible foot gun of cascaded discrete initialization.

Perhaps, though, that's only a useful (not noisy) analysis when we have declared event dependencies for more granular analysis?

> Maybe we could offer a function with a different name for that, e.g., `InitializeRepeatedly()`

The only concern here is if there is discrete event that (for whatever reason) has an external event, e.g. publishing a LCM or ROS 2 message.

However, I guess the correct solution here is to
- only publish to middleware via (discrete) publish events
- warn or fail if `publish_at_initialization_ = true`?

Does that sound practical?

Evan + Sherm, WDYT?

sherm1:
My proposal now on reading through the above discussion would be just to update the documentation to discuss the sequential-initialization problem and suggest repeated calls to Simulator::Initialize(). We could revisit if that turns out to be insufficient. 

It might also be helpful to upgrade ZOH to initialize properly, WDYT?

jwnimmer-tri:
@sherm1 I think that's an improvement, but remember than Evan said repeated Initialize would not work for him (since his events have side effects beyond changing the state).

sherm1:
True, but I also believe he wasn't asking for a change in the current behavior -- is that right @edrumwri ?

EricCousineau-TRI:
> [...] but remember than Evan said repeated Initialize would not work for him (since his events have side effects beyond changing the state).

From my naive perspective, that seems like a defect in the implementation of Evan's events (see my above comment about  middleware)

Evan, (or Jeremy / Sherm if you know), is there something that prevents you from decoupling initialization updates from initialization publishes (side effects)?

EricCousineau-TRI:
I've been exercising the repeated initialization a bit more, and have noted the following:
- `Initiaize()` will trigger discrete publishes on things like `LcmPublisherSystem`. That is not desirable; see [Drake Slack thread](https://drakedevelopers.slack.com/archives/C2CHRT98E/p1674250843246739)
- It's nice to have a generic flag for systems to (somehow) let them know when a called initialization event is a "warm-up" event, or the "final" event. See [rough example Python code](https://github.com/EricCousineau-TRI/repro/blob/6be62282021af9cf591e2b72d7b40ace31fccd0d/drake_stuff/system_init/systems.py#L44-L112)

If we can solve these in some form and provide user guidance, then I think handling start-up transients will become much simpler and ergonomic (possibly moreso than Simulink, at least as far as I remember in 2017)

jwnimmer-tri:
Closing as abandoned. When the work unpauses, feel free to reopen it, or open a new PR.

PR #18546, opened 2023-01-06
[skylark] Set num_threads for test concurrency control

jwnimmer-tri:
Tests that do not opt-in to concurrency will be limited to 1 thread.

---

Reviewable r1 contains code copied from Anzu; r2 and onward contain Drake's integration changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18546)

PR #18560, opened 2023-01-09
[tools] Improve documentation of num_threads

jwnimmer-tri:
Follow-up from #18546.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18560)

### DIFF
--- a/tools/skylark/README.md
+++ b/tools/skylark/README.md
@@ -23,13 +23,15 @@ Can either be None, or else an integer.
 
 When None or 1, adds `OMP_NUM_THREADS="1"` to `env` to disable OpenMP.
 
-When N>1, adds `OMP_NUM_THREADS="N"` to `env` to establish OpenMP's
-desired level of parallelism, and also adds "cpu:N" to `tags` to reserve
-sufficient CPUs for the test.
+When N>1, adds `OMP_NUM_THREADS="N"` to `env` to allow for the desired level of
+parallelism, and also adds "cpu:N" to `tags` to reserve sufficient CPUs for the
+test. Those changes are a *necessary* but not *sufficient* condition to enable
+actual CPU parallelism while running the test. In addition, the overall build
+must also have OpenMP enabled, via --config=omp or --config=everything.
 
-Note that setting N>1 will also be applied to any sub-processes that
-are launched by your test program.  Ask for help on Slack if you need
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
+@SeanCurtis-TRI for review, please.

I'm OK if you want to take both reviews, or else you could tap Rico for platform as a follow-up.

SeanCurtis-TRI (commented):
+(status: single reviewer ok) 

:LGTM:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18560#-:-NLN8yLt7ssEm-5t5lkz:bamq21o)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`tools/skylark/README.md` line 30 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18560#-NLN97c68PHj_vdYKyeH:-NLN97c68PHj_vdYKyeI:b-zfv0f7) ([raw file](https://github.com/robotlocomotion/drake/blob/d4989b74f19617061c64384d9e6ca87fd3ec3031/tools/skylark/README.md#L30)):*
> ```Markdown
> test. Those changes are a *necessary* but not *sufficient* condition to enable
> actual CPU parallelism while running the test. In addition, the overall build
> must also have OpenMP enabled, via --config=omp or --config=everything.
> ```

BTW As you'd pointed out earlier, all of these steps are necessary but *still* may not be sufficient. While it seems unlikely that the hardware constraint is going to be a common tripping hazard, is it still worth mentioning for completeness?

jwnimmer-tri:
*[`tools/skylark/README.md` line 30 at r1](https://reviewable.io/reviews/robotlocomotion/drake/18560#-NLN97c68PHj_vdYKyeH:-NLN9zXj10x19jVHkGeF:b-ex3fe5) ([raw file](https://github.com/robotlocomotion/drake/blob/d4989b74f19617061c64384d9e6ca87fd3ec3031/tools/skylark/README.md#L30)):*

BTW As you'd pointed out earlier, all of these steps are necessary but *still* may not be sufficient. While it seems unlikely that the hardware constraint is going to be a common tripping hazard, is it still worth mentioning for completeness?
</blockquote></details>

I don't think it will cause a problem in practice, and therefore adding more text here is additional confusion and complexity without any actual benefit.

PR #18566, opened 2023-01-11
[examples] Add main and yaml parsing to python hardware_sim

jwnimmer-tri:
Towards #17112.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18566)

Issue #17112, opened 2022-05-05
Add hardware_station_simulation capability


jwnimmer-tri:
Details to follow.

(Cross-reference for Anzu issue 8760.)

jwnimmer-tri:
One question here is how to provide a demo of this feature, with an example binary.  If we can figure it out, it might be nice to replace `kuka_simulation` with this program, if we can?  @sammy-tri has some ideas.

@calderpg-tri also suggests that we offer a 2-arm config file in the demo as well.

mpetersen94:
Just found this issue as I was trying to implement a similar setup for my lab's research.  Would love even a suggestion of good ways to organize the config file so that the simulation/communication layer is adjustable for different numbers/types of robots!

jwnimmer-tri:
We are getting very close to publishing this software (active work for the past ~month).  In the meantime, here's a sample config files for what we use:  https://gist.github.com/jwnimmer-tri/b087cc01f41296d637fe33c032762298

The glue code that supports `model_drivers` and `lcm_buses` will be some of the first to arrive.

mpetersen94:
The config file is already super helpful! Does the directive file have all the models to load into the simulator (i.e. arms, grippers, welded objects and floating objects)? I'm also assuming the model drivers section is telling what control stack (that would run on the real robot) to spoof for the simulation? Feel free not to answer if any of this is proprietary.

jwnimmer-tri:
Yes, exactly right on both counts.

For loading models into the plant (environment, robots, manipulands), we tend to use cascading directives files -- the typical top-level one for a station includes another one (e.g., our standard iiwa or panda) which includes another one (how the gripper is typically welded), etc. That allows for some amount of configuration re-use.

The model_drivers basically wire up the systems such as https://github.com/RobotLocomotion/drake/tree/master/manipulation/kuka_iiwa along with inverse dynamics to respond to lcm commands and report status, just like the real robot.

jwnimmer-tri:
See #17615 for a demo program underway.

mpetersen94:
Now that #17615 has merged, what is the missing glue that is needed to implement a similar hardware simulator on the python side?

jwnimmer-tri:
I am not entirely sure, but I do plan to work on it (and have a Python twin to that demo as an example).

[...]

PR #18571, opened 2023-01-12
[geometry] Add array-based getters and setters to Rgba

jwnimmer-tri:
Rewrite serialization to add python bindings.

Towards https://github.com/RobotLocomotion/drake/pull/18566 (and therefore https://github.com/RobotLocomotion/drake/issues/17112).

Requires #18572 first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)

### DIFF
--- a/bindings/pydrake/BUILD.bazel
+++ b/bindings/pydrake/BUILD.bazel
@@ -497,6 +497,7 @@ drake_py_unittest(
     name = "geometry_common_test",
     deps = [
         ":geometry_py",
+        "//bindings/pydrake/common",
         "//bindings/pydrake/common/test_utilities",
         "//bindings/pydrake/multibody:parsing_py",
         "//bindings/pydrake/multibody:plant_py",
--- a/bindings/pydrake/geometry_py_common.cc
+++ b/bindings/pydrake/geometry_py_common.cc
@@ -9,6 +9,7 @@
 #include "drake/bindings/pydrake/common/default_scalars_pybind.h"
 #include "drake/bindings/pydrake/common/deprecation_pybind.h"
 #include "drake/bindings/pydrake/common/identifier_pybind.h"
+#include "drake/bindings/pydrake/common/serialize_pybind.h"
 #include "drake/bindings/pydrake/common/value_pybind.h"
 #include "drake/bindings/pydrake/documentation_pybind.h"
 #include "drake/common/drake_deprecated.h"
@@ -321,20 +322,34 @@ void DoScalarIndependentDefinitions(py::module m) {
     cls  // BR
         .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init<double, double, double, double>(), py::arg("r"),
-            py::arg("g"), py::arg("b"), py::arg("a") = 1.,
+            py::arg("g"), py::arg("b"), py::arg("a") = 1.0,
             cls_doc.ctor.doc_4args)
         .def("r", &Class::r, cls_doc.r.doc)
         .def("g", &Class::g, cls_doc.g.doc)
         .def("b", &Class::b, cls_doc.b.doc)
         .def("a", &Class::a, cls_doc.a.doc)
-        .def("set", &Class::set, py::arg("r"), py::arg("g"), py::arg("b"),
-            py::arg("a") = 1., cls_doc.set.doc)
+        .def("set",
+            py::overload_cast<double, double, double, double>(&Class::set),
+            py::arg("r"), py::arg("g"), py::arg("b"), py::arg("a") = 1.0,
+            cls_doc.set.doc_4args)
+        .def("set",
+            py::overload_cast<const Eigen::Ref<const Eigen::VectorXd>&>(
+                &Class::set),
+            py::arg("rgba"), cls_doc.set.doc_1args)
         .def(py::self == py::self)
         .def(py::self != py::self)
         .def("__repr__", [](const Class& self) {
           return py::str("Rgba(r={}, g={}, b={}, a={})")
               .format(self.r(), self.g(), self.b(), self.a());
         });
+    DefAttributesUsingSerialize(&cls);
+    cls.def_property("rgba",
+        // The Serialize-based binding skips the validity checking; we'll
+        // add it back here by re-binding the property getter and setter.
+        &Class::rgba,
+        py::overload_cast<const Eigen::Ref<const Eigen::VectorXd>&>(
+            &Class::set),
+        "The RGBA value as a property (as np.ndarray).");
     DefCopyAndDeepCopy(&cls);
     AddValueInstantiation<Rgba>(m);
   }
--- a/bindings/pydrake/test/geometry_common_test.py
+++ b/bindings/pydrake/test/geometry_common_test.py
@@ -10,6 +10,7 @@ from pydrake.common.test_utilities import numpy_compare
 from pydrake.common.test_utilities.deprecation import catch_drake_warnings
 from pydrake.common.test_utilities.pickle_compare import assert_pickle
 from pydrake.common.value import AbstractValue, Value
+from pydrake.common.yaml import yaml_load_typed
 from pydrake.math import RigidTransform
 from pydrake.multibody.plant import CoulombFriction
 
@@ -299,21 +300,73 @@ class TestGeometryCore(unittest.TestCase):
     def test_rgba_api(self):
         default_white = mut.Rgba()
         self.assertEqual(default_white, mut.Rgba(1, 1, 1, 1))
-        r, g, b, a = 0.75, 0.5, 0.25, 1.
+        r, g, b, a = 0.75, 0.5, 0.25, 1.0
         color = mut.Rgba(r=r, g=g, b=b)
         self.assertEqual(color.r(), r)
         self.assertEqual(color.g(), g)
         self.assertEqual(color.b(), b)
         self.assertEqual(color.a(), a)
         self.assertEqual(color, mut.Rgba(r, g, b, a))
-        self.assertNotEqual(color, mut.Rgba(r, g, b, 0.))
+        self.assertNotEqual(color, mut.Rgba(r, g, b, 0.0))
         self.assertEqual(
             repr(color),
             "Rgba(r=0.75, g=0.5, b=0.25, a=1.0)")
-        color.set(r=1., g=1., b=1., a=0.)
-        self.assertEqual(color, mut.Rgba(1., 1., 1., 0.))
+        color.set(r=1.0, g=1.0, b=1.0, a=0.0)
+        self.assertEqual(color, mut.Rgba(1.0, 1.0, 1.0, 0.0))
+        color.set(rgba=[0.75, 0.5, 0.25])
+        self.assertEqual(color, mut.Rgba(0.75, 0.5, 0.25, 1.0))
+
+        # Property read/write.
+        color.rgba = [0.1, 0.2, 0.3, 0.4]
+        self.assertEqual(color.r(), 0.1)
+        self.assertEqual(color.g(), 0.2)
+        self.assertEqual(color.b(), 0.3)
+        self.assertEqual(color.a(), 0.4)
+        color.rgba = [0.5, 0.6, 0.7]
+        self.assertEqual(color.r(), 0.5)
+        self.assertEqual(color.g(), 0.6)
+        self.assertEqual(color.b(), 0.7)
+        self.assertEqual(color.a(), 1.0)
+        self.assertEqual(color.rgba[0], 0.5)
+        self.assertEqual(color.rgba[1], 0.6)
+        self.assertEqual(color.rgba[2], 0.7)
+        self.assertEqual(color.rgba[3], 1.0)
+        with self.assertRaisesRegex(RuntimeError, ".*range.*"):
+            color.rgba = [-1.0] * 4
+        with self.assertRaisesRegex(RuntimeError, ".*3 or 4.*"):
+            color.rgba = [1.0] * 2
+        with self.assertRaisesRegex(RuntimeError, ".*3 or 4.*"):
+            color.rgba = [1.0] * 5
+
         # Confirm value instantiation.
         Value[mut.Rgba]
+
+    def test_rgba_yaml(self):
+        yaml = "rgba: [0.1, 0.2, 0.3, 0.4]"
+        dut = yaml_load_typed(schema=mut.Rgba, data=yaml)
+        self.assertEqual(dut.r(), 0.1)
+        self.assertEqual(dut.g(), 0.2)
+        self.assertEqual(dut.b(), 0.3)
+        self.assertEqual(dut.a(), 0.4)
+
+        yaml = "rgba: [0.1, 0.2, 0.3]"
+        dut = yaml_load_typed(schema=mut.Rgba, data=yaml)
+        self.assertEqual(dut.r(), 0.1)
+        self.assertEqual(dut.g(), 0.2)
+        self.assertEqual(dut.b(), 0.3)
+        self.assertEqual(dut.a(), 1.0)
+
+        yaml = "rgba: []"
+        with self.assertRaisesRegex(RuntimeError, ".*3 or 4.*"):
+            yaml_load_typed(schema=mut.Rgba, data=yaml)
+
+        yaml = "rgba: [0, 1, 2, 3, 4, 5]"
+        with self.assertRaisesRegex(RuntimeError, ".*3 or 4.*"):
+            yaml_load_typed(schema=mut.Rgba, data=yaml)
+
+        yaml = "rgba: [0, 0, 0, -1]"
+        with self.assertRaisesRegex(RuntimeError, ".*range.*"):
+            yaml_load_typed(schema=mut.Rgba, data=yaml)
 
     def test_shape_constructors(self):
         shapes = [
--- a/common/yaml/test/example_structs.h
+++ b/common/yaml/test/example_structs.h
@@ -230,6 +230,7 @@ bool operator==(const EigenStruct<Rows, Cols>& a,
 
 using EigenVecStruct = EigenStruct<Eigen::Dynamic, 1>;
 using EigenVec3Struct = EigenStruct<3, 1>;
+using EigenVecUpTo3Struct = EigenStruct<Eigen::Dynamic, 1, 3, 1>;
 using EigenMatrixStruct = EigenStruct<Eigen::Dynamic, Eigen::Dynamic>;
 using EigenMatrix34Struct = EigenStruct<3, 4>;
 using EigenMatrix00Struct = EigenStruct<0, 0>;
--- a/common/yaml/test/yaml_read_archive_test.cc
+++ b/common/yaml/test/yaml_read_archive_test.cc
@@ -536,8 +536,11 @@ TEST_P(YamlReadArchiveTest, EigenVector) {
                        const Eigen::VectorXd& expected) {
     const auto& vec = AcceptNoThrow<EigenVecStruct>(LoadSingleValue(value));
     const auto& vec3 = AcceptNoThrow<EigenVec3Struct>(LoadSingleValue(value));
+    const auto& upto3 =
+        AcceptNoThrow<EigenVecUpTo3Struct>(LoadSingleValue(value));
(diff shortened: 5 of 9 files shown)


jwnimmer-tri:

Note that this will not pass all of its tests until #18572 merges.  I'm OK if you want to wait for that to land, but in case you'd like to get started now the code should be in practically its final form.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLanRkK4Awa7hOwnb_r:-NLanRkK4Awa7hOwnb_s:b-cxrucs) (no related file):*
Working

Check Anzu CI.

___

SeanCurtis-TRI (commented):
A few small clarity things.

:LGTM:

> Reviewed 4 of 7 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-:-NLbFGGtBDq7at3BB1p2:bjvda3o)* status: 5 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`common/yaml/yaml_read_archive.h` line 379 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCIek--PQnRbbsauN:-NLbCIek--PQnRbbsauO:b-rqro89) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/common/yaml/yaml_read_archive.h#L379)):*
> ```objc
>     if (max_size.has_value() && size > *max_size) {
>       ReportError(fmt::format(
>           "has too many array elements ({}); the maximum size is {} in the",
> ```

nit: Confused about the trailling "in the". The other `ReportError` usages have complete error descriptions without any trailing artifacts.
___
*[`geometry/rgba.h` line 94 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLb9W1n9x5iKn0_56YD:-NLb9W1n9x5iKn0_56YE:bpso7e5) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/rgba.h#L94)):*
> ```objc
>     // We want to be sure that value_ remains well-formed even if set() throws,
>     // so we'll clear it and then reset it.
>     auto new_value = value_;
> ```

nit: This `auto` feels dubious. Is this simply to avoid typing the whole templated declaration?
___
*[`geometry/rgba.h` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCp0677O3l-rMkker:-NLbCp0677O3l-rMkkes:b-to9c96) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/rgba.h#L105)):*
> ```objc
>   // need to declare it as Eigen::Dynamic for compatibility with Serialize.
>   // As documented above, default-constructed Rgba is opaque white.
>   Eigen::Matrix<double, Eigen::Dynamic, 1, 0, 4, 1> value_{
> ```

nit: This particular incantation will be hard to grok. *What* is being done and *why* it is being done requires sleuthing.

Figuring out the "what" requires exploring the semantics of the template parameters that few drake developers ever look at and recognizing we've set max rows.

Figuring out *why* we're doing that requires understanding how it plays with yaml.

As this second point is *particularly* baroque, I think it'll help maintainers to be pointed in that direction.
___
*[`geometry/test/rgba_test.cc` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbAEbQ2u7Nqo8WhEX-:-NLbAEbQ2u7Nqo8WhEX0:b6lz9z9) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/test/rgba_test.cc#L112)):*
> ```cpp
>   DRAKE_EXPECT_THROWS_MESSAGE(
>       yaml::LoadYamlString<Rgba>("rgba: [1, 1, 1, 1, 1]"),
>       ".*maximum size is 4.*");
> ```

BTW This error message in this test file may surprise readers (it surprised me). Given that `Rgba` explicitly validates vector size in `set()` and `set()` is called in archiving, one would not expect *this* message. Probably a comment saying indicating that the yaml error mechanism got first whack at the problem might help.

Exacerbated by the fact that for a vector that is too small, we get the expected response.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`common/yaml/yaml_read_archive.h` line 379 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCIek--PQnRbbsauN:-NLbHZom9IitOlS1pFBU:b-9olzfw) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/common/yaml/yaml_read_archive.h#L379)):*

nit: Confused about the trailling "in the". The other `ReportError` usages have complete error descriptions without any trailing artifacts.
</blockquote></details>

Here is an example error message when the array is too big (taken from the rgba unit test):

`YAML node of type Mapping (with size 1 and keys {rgba}) has too many array elements (5); the maximum size is 4 in the entry for Eigen::Matrix<double,-1,1,0,4,1> rgba.`

So two questions:

(1) Is the message on its own a good one?

(2) Should the source code here somehow be made more clear about what's going on?
___
*[`geometry/rgba.h` line 94 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLb9W1n9x5iKn0_56YD:-NLbFrd04x1KN47zs6ZJ:b-ni9wxb) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/rgba.h#L94)):*

nit: This `auto` feels dubious. Is this simply to avoid typing the whole templated declaration?
</blockquote></details>

Yes.  Typing it again longform would be too brittle, which means we'd need a typedef, and that seemed like overkill.  My hypothesis was that a typical reader would not struggle too much to understand this formulation.
___
*[`geometry/rgba.h` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCp0677O3l-rMkker:-NLbJGqd71o-MiJ64-Kv:bbgm45i) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/rgba.h#L105)):*

nit: This particular incantation will be hard to grok. *What* is being done and *why* it is being done requires sleuthing.

Figuring out the "what" requires exploring the semantics of the template parameters that few drake developers ever look at and recognizing we've set max rows.

Figuring out *why* we're doing that requires understanding how it plays with yaml.

As this second point is *particularly* baroque, I think it'll help maintainers to be pointed in that direction.
</blockquote></details>

Hopefully this rewrite is more clear?
___
*[`geometry/test/rgba_test.cc` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbAEbQ2u7Nqo8WhEX-:-NLbGuaf0Ai0VnULXFKl:b24ttf1) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/test/rgba_test.cc#L112)):*

BTW This error message in this test file may surprise readers (it surprised me). Given that `Rgba` explicitly validates vector size in `set()` and `set()` is called in archiving, one would not expect *this* message. Probably a comment saying indicating that the yaml error mechanism got first whack at the problem might help.

Exacerbated by the fact that for a vector that is too small, we get the expected response.
</blockquote></details>

Done.

I also reordered the tests a bit, hopefully improved clarity as well.

jwnimmer-tri:
+@EricCousineau-TRI for platform review per schedule, please.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 3 unresolved discussions, LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLanRkK4Awa7hOwnb_r:-NLbNlUgBPwT71i2HSHM:b-l28nif) (no related file):*

Working

Check Anzu CI.
</blockquote></details>

Passed.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI and @jwnimmer-tri)

___
*[`common/yaml/yaml_read_archive.h` line 379 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCIek--PQnRbbsauN:-NLbLHgn1n9gzwFdqbpM:b-s4glz1) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/common/yaml/yaml_read_archive.h#L379)):*

Here is an example error message when the array is too big (taken from the rgba unit test):

`YAML node of type Mapping (with size 1 and keys {rgba}) has too many array elements (5); the maximum size is 4 in the entry for Eigen::Matrix<double,-1,1,0,4,1> rgba.`

So two questions:

(1) Is the message on its own a good one?

(2) Should the source code here somehow be made more clear about what's going on?
</blockquote></details>

When we visit array, we would return something like:

```
YAML node of type Mapping (with size 1 and keys {foo") has
7-size entry (wanted 5-size) entry for Sometype bar.
```
(I think.) The flow into the word entry is clunky. I imagine it could be the same for vector.

```
YAML node of type Mapping (with size 1 and keys {rgba}) has
5-size entry (wanted up to 4-size) entry for
Eigen::Matrix<double,-1,1,0,4,1> rgba.
```
That doesn't flow any better than for array, but doesn't seem to flow any worse either.

I suspect that the `ReportError` function wasn't optimized for discussing array elements and we're running into that.

So, regarding the current state:

1. I believe that the resulting message given the current code is superior to what I understand we'd get for a malformed array. (So, making max-size vector match array in some sense would be a step backwards in message quality). 
2. Tweaking error reporting so that it flows better across the various types seems reasonably outside the scope of this PR.
3. The simplest solution is a note at this site indicating why it looks aberrant and that's ok.
___
*[`geometry/rgba.h` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCp0677O3l-rMkker:-NLbKmUr9CvRMXmiHoIZ:b-4z13qv) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/geometry/rgba.h#L105)):*

Hopefully this rewrite is more clear?
</blockquote></details>

Works for me.  Thanks.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: LGTM missing from assignee EricCousineau-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

___
*[`common/yaml/yaml_read_archive.h` line 379 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbCIek--PQnRbbsauN:-NLbQ-X58Q-sirwCcOUG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/70f8512fb0d37bf17ece7546e530481c37bcd1bb/common/yaml/yaml_read_archive.h#L379)):*

When we visit array, we would return something like:

```
YAML node of type Mapping (with size 1 and keys {foo") has
7-size entry (wanted 5-size) entry for Sometype bar.
```
(I think.) The flow into the word entry is clunky. I imagine it could be the same for vector.

```
YAML node of type Mapping (with size 1 and keys {rgba}) has
5-size entry (wanted up to 4-size) entry for
Eigen::Matrix<double,-1,1,0,4,1> rgba.
```
That doesn't flow any better than for array, but doesn't seem to flow any worse either.

I suspect that the `ReportError` function wasn't optimized for discussing array elements and we're running into that.

So, regarding the current state:

1. I believe that the resulting message given the current code is superior to what I understand we'd get for a malformed array. (So, making max-size vector match array in some sense would be a step backwards in message quality). 
2. Tweaking error reporting so that it flows better across the various types seems reasonably outside the scope of this PR.
3. The simplest solution is a note at this site indicating why it looks aberrant and that's ok.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: LGTM missing from assignee EricCousineau-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
:lgtm: platform with two nits
- concern about non-parity bindings with C++ in terms of `.set(rgba=...)` and `.rgba = property()` vs. `.rgba()`.
- Making `value_` a type of `Vec3or4` seems not as good as `Vector4 value_` and `Vec3or4 temp`.

Will not block if y'all think it's better this way, though.

> Reviewed 3 of 7 files at r1, 3 of 5 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-:-NLbvJa5DjiP63MNcVF7:b-cdmzu7)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`geometry/rgba.h` line 109 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbvaNV2SFY8kJ5xi1a:-NLbvaNW8VmnlXX1d-VS:bic7hb4) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/geometry/rgba.h#L109)):*
> ```objc
>   // We'll also set the template argument `MaxRowsAtCompileTime == 4` so that
>   // our data is stored inline as part of this object, instead of on the heap.
>   Eigen::Matrix<double, Eigen::Dynamic, 1, 0, 4, 1> value_ =
> ```

nit Unclear why we want to the direct storage to be `Vec3or4`-ish and then add in the "canoicalizing" / "re-validating" step.

Why not use a direct temporary in serialization? e.g.
```
Vec3or4 temp = value_;
a->Visit(MakeNameValue("rgba", &temp));
value_.setZero();
set(temp);
```
___
*[`bindings/pydrake/geometry_py_common.cc` line 342 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbuxq40dTe5P9firU7:-NLbuxq40dTe5P9firU8:b-w9ize4) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/bindings/pydrake/geometry_py_common.cc#L342)):*
> ```cpp
>         });
>     DefAttributesUsingSerialize(&cls);
>     cls.def_property("rgba",
> ```

nit Not sure if I've seen a more Pythonic approach used in lieu of the exact same C++ API. I believe I've seen it used in addition to, instead.

Consider explicitly binding `.set(rgba=...)`?
Additionally, concerned that this is not the exact same as C++ `my_color.rgba()`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`geometry/rgba.h` line 109 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbvaNV2SFY8kJ5xi1a:-NLby5Zx7QHnPu1fN5v9:b4aajpe) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/geometry/rgba.h#L109)):*

nit Unclear why we want to the direct storage to be `Vec3or4`-ish and then add in the "canoicalizing" / "re-validating" step.

Why not use a direct temporary in serialization? e.g.
```
Vec3or4 temp = value_;
a->Visit(MakeNameValue("rgba", &temp));
value_.setZero();
set(temp);
```
</blockquote></details>

The `DefAttributesUsingSerialize` is incompatible with temporaries, it must only ever use member fields.  (I'll add a comment here.)
___
*[`bindings/pydrake/geometry_py_common.cc` line 342 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbuxq40dTe5P9firU7:-NLbyIbs9v9oBJLszZVb:b-94aqxw) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/bindings/pydrake/geometry_py_common.cc#L342)):*

nit Not sure if I've seen a more Pythonic approach used in lieu of the exact same C++ API. I believe I've seen it used in addition to, instead.

Consider explicitly binding `.set(rgba=...)`?
Additionally, concerned that this is not the exact same as C++ `my_color.rgba()`.
</blockquote></details>

I guess I can bind the `set` overload.

But for deserialization we _need_ to have the settable property, that's how deserialization works (using setattr).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

___
*[`geometry/rgba.h` line 109 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbvaNV2SFY8kJ5xi1a:-NLc-DIA4g7o7rjSzCJ_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/geometry/rgba.h#L109)):*

The `DefAttributesUsingSerialize` is incompatible with temporaries, it must only ever use member fields.  (I'll add a comment here.)
</blockquote></details>

Done.
___
*[`bindings/pydrake/geometry_py_common.cc` line 342 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLbuxq40dTe5P9firU7:-NLc-Eu98zmFVSEpX41F:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d5f0e4849f703fef2b116ed558bcaef3b1ecea8/bindings/pydrake/geometry_py_common.cc#L342)):*

I guess I can bind the `set` overload.

But for deserialization we _need_ to have the settable property, that's how deserialization works (using setattr).
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18571)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`geometry/rgba.h` line 91 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18571#-NLc2_TFABoMrbvTplfV:-NLc2_TFABoMrbvTplfW:bwpziyt) ([raw file](https://github.com/RobotLocomotion/drake/blob/cc25d3791253e5010b65df6b01123d54687696db/geometry/rgba.h#L91)):*
> ```objc
>     //
>     // N.B. For the Python binding use of DefAttributesUsingSerialize, this
>     // must refer a member field; we cannot use a stack temporary here.
> ```

nit:  "...must refer __*to*__ a member field...

EricCousineau-TRI:
Ah, didn't realize that about serialization + property/attr. Thanks!
And derp on the temporary setup as well :facepalm:

PR #18584, opened 2023-01-13
CompareMatrices takes RigidTransform and RotationMatrix directly

SeanCurtis-TRI:
This changes calls like:

```c++
EXPECT_TRUE(CompareMatrices(X_AB.GetAsMatrix34(),
                            X_AB_expected.GetAsMatrix34()));
```

to

```c++
EXPECT_TRUE(CompareMatrices(X_AB, X_AB_expected));
```

And the same for rotations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584)

### DIFF
--- a/common/test_utilities/BUILD.bazel
+++ b/common/test_utilities/BUILD.bazel
@@ -188,6 +188,7 @@ drake_cc_googletest(
     deps = [
         ":eigen_matrix_compare",
         "//common:essential",
+        "//math:geometric_transform",
     ],
 )
 
--- a/common/test_utilities/eigen_matrix_compare.h
+++ b/common/test_utilities/eigen_matrix_compare.h
@@ -113,4 +113,45 @@ template <typename DerivedA, typename DerivedB>
                                        << m2;
 }
 
+/** Specialization for comparing rotation matrices. The enabling criteria is
+ a shibboleth that should uniquely apply to rotation matrices. This allows us
+ to directly compare RotationMatrixes without explicitly pulling in drake::math
+ as a dependency. I.e.,
+
+ RotationMatrixd R_AB_dut = ...;
+ RotationMatrixd R_AB_expected = ...;
+ EXPECT_TRUE(CompareMatrices(R_AB_dut, R_AB_expected));
+ */
+template <template <typename> typename RotMatrixType, typename T>
+[[nodiscard]] std::enable_if_t<
+    std::is_same_v<decltype(RotMatrixType<T>::MakeXRotation(0)),
+                   RotMatrixType<T>>,
+    ::testing::AssertionResult>
+CompareMatrices(const RotMatrixType<T>& m1, const RotMatrixType<T>& m2,
+                double tolerance = 0.0,
+                MatrixCompareType compare_type = MatrixCompareType::absolute) {
+  return CompareMatrices(m1.matrix(), m2.matrix(), tolerance, compare_type);
+}
+
+/** Specialization for comparing rigid transforms. The enabling criteria is
+ a shibboleth that should uniquely apply to rigid transforms. This allows us
+ to directly compare RigidTransforms without explicitly pulling in drake::math
+ as a dependency. I.e.,
+
+ RigidTransformd X_AB_dut = ...;
+ RigidTransformd X_AB_expected = ...;
+ EXPECT_TRUE(CompareMatrices(X_AB_dut, X_AB_expected));
+ */
+template <template <typename> typename TransformType, typename T>
+[[nodiscard]] std::enable_if_t<
+    std::is_same_v<decltype(TransformType<T>().GetAsMatrix34()),
+                   Eigen::Matrix<T, 3, 4>>,
+    ::testing::AssertionResult>
+CompareMatrices(const TransformType<T>& m1, const TransformType<T>& m2,
+                double tolerance = 0.0,
+                MatrixCompareType compare_type = MatrixCompareType::absolute) {
+  return CompareMatrices(m1.GetAsMatrix34(), m2.GetAsMatrix34(), tolerance,
+                         compare_type);
+}
+
 }  // namespace drake
--- a/common/test_utilities/test/eigen_matrix_compare_test.cc
+++ b/common/test_utilities/test/eigen_matrix_compare_test.cc
@@ -2,6 +2,9 @@
 
 #include <gtest/gtest.h>
 
+#include "drake/math/rigid_transform.h"
+#include "drake/math/rotation_matrix.h"
+
 namespace drake {
 namespace {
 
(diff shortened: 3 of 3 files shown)


SeanCurtis-TRI (commented):
@jwnimmer-tri Is there any reason *not* to do this?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NLg7lnN7lliMrezDmFc:b-r3aj86)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri:
> @jwnimmer-tri Is there any reason not to do this?

As implemented currently, it adds spurious dependencies.  Now anytime `rotation_matrix.h` changes, more unit tests will get recompiled than are strictly necessary.  That's easily enough solved by moving this new content to a different filename (presuming under `math/...` somewhere).  The question would be whether a new include path is worth the changes at call sites.  Assuming we might add more transform-comparison sugar down the road (e.g., rotation tolerance vs translation tolerance), it might very well be worth it.

Bigger picture, though, it's also doubling down on the "AssertionResult" idiom, which we should disprefer in favor of the more flexible architecture of gtest "matchers".  It's not necessarily a defect to incrementally improve the old way, but if I were doing this I wouldn't invest my time in the old way, rather I'd work on doing better in the new way.

SeanCurtis-TRI (commented):
I completely agree with the first point. Dragging in those dependencies can easily be bad. Everywhere geometric transforms are in question, we're already dragging them in.

However, I find the second point less compelling. The matching API is ill-suited to these types.

1. Because of the macro based structure, we can't simply overload the `Eq` matcher. We'd have to have `RotationEq` and `TransformEq` (or some such thing and doubling up for the `Near` variant). This is cumbersome, but not a deal killer.
2. While we'd like the name`RotationEq` to help guide the user to make sure they only pass the right type (e.g., `RotationMatrix`), failure to do so will lead to compilation errors inside the matcher implementation and not in the unit test. (This is also an argument to naming it `RotationMatrixEq` to be more strongly suggestive of type and not semantics). Again, not a deal killer.
3. In the case of failure, we've got limited control over what is printed in the console. 

```c++
EXPECT_THAT(R, RotationMatrixEq(R_expected));
```

produces an output of the form

```
Failure
value of: R
Expected: rotation matrix eq (R_expected <string representation of R_expected>
Actual:  <string representation of R>
```

We have *some* limited control over formatting. By defining a `PrintTo` in with the matchers and playing with `result_listener`, I can get a message of the form:

```
Failure
Value of: R
Expected: rotation eq (R_ref: 
        0  0.877583 -0.479426
        1         0         0
        0  0.479426  0.877583)
  Actual: 
        1         0         0
        0  0.877583 -0.479426
        0  0.479426  0.877583, 
where the difference is:
        1 -0.877583  0.479426
       -1  0.877583 -0.479426
        0         0         0
```

But this still puts limits on my formatting. Note that the dut symbol (R) is separated from its value).  Furthermore, text and math are comingled. Contrast that with the result that comes from overloading `CompareMatrices`. 

```
Value of: CompareMatrices(R, R_ref)
  Actual: false (Values at (0, 0) exceed tolerance: 1 vs. 0, diff = 1, tolerance = 0
m1 =
                   1                    0                    0
                   0  0.87758256189037276 -0.47942553860420301
                   0  0.47942553860420301  0.87758256189037276
m2 =
                   0  0.87758256189037276 -0.47942553860420301
                   1                    0                    0
                   0  0.47942553860420301  0.87758256189037276
delta=
                   1 -0.87758256189037276  0.47942553860420301
                  -1  0.87758256189037276 -0.47942553860420301
                   0                    0                    0)
```

I believe that because the `RotationMatrix` and `RigidTransform` screen output generally requires some very specific, multi-line formatting to be helpful, it fights the matcher, making failures far less helpful (which seems to be 90% the point of `CompareMatrices`).

Thoughts?

P.S. [My branch](https://github.com/SeanCurtis-TRI/drake/tree/PR_geometric_transform_matcher) contains a brief sally in the matcher direction. Maybe I misconstrued the limited documentation and missed some liberating feature.

+(priority: backlog) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NM-uqCGBom5-u7_eZ1K:bzfy2wd)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri:
The matcher hypothesis goes like this:

By induction, we already have good a matcher infrastructure for multidimensional matrices, with configuration options and diagnostic output no worse than CompareMatrices.  Since we need to compare matrix data beyond just RigidTransform and/or RotationMatrix, we will certainly need this anyway.  Essentially the task to make this happen is to port CompareMatrices to the matcher idiom.  It's akin to the upstream collection matchers already, except for a multi-dimentional collection, and with better support for tolerances.  Possibly we should look into whether anticipating https://en.cppreference.com/w/cpp/container/mdspan would make sense.

Given that multidimensional matcher, for RigidTransform and RotationMatrix we just need to guide it to unwrap them into their matrix form when comparing.  That's akin to the higher-order matchers already like http://google.github.io/googletest/reference/matchers.html#composite-matchers and http://google.github.io/googletest/reference/matchers.html#adapters-for-matchers so we have something like `EXPECT_THAT(rotmat, AsMatrix(MdElementsAreArray({{1, 2, 3}, {4, 5, 6}, {7, 8, 9}, tol)));`.

sherm1 (commented):
FWIW I really like the CompareMatrices(R, R_expected) overload and its output!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NM0GrAx8BqkZiN4JaHD:btvee64)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
Turning `CompareMatrices` into a matcher is straight forward. Certainly, it would be sensitive to all the same differences between matrices.

Where the loss comes in would be:

1. The failure message would be degraded (arguably significantly so).
2. The overload-like behavior we get with default parameters would be supplanted by a family of matchers (one unique matcher per collection of parameters).

(P.S. Your hypothetical example uses `AsMatrix` as the second parameter of `EXPECT_THAT`. The name `AsMatrix` suggests it's not an matcher -- returning a bool -- but, instead, produces a value. That means you're using `EXPECT_THAT` as a replacement for `EXPECT_EQ` which gmock advises against. Presumably, it should be `MatrixEq` or some such thing? I mention this is a footnote of what I interpret as vague rumination pinned down in text and not a concrete expectation.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NM0H3jc6WkYjwufnQqp:bfvc95f)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri:
> The failure message would be degraded (arguably significantly so).

But why?  The failure message for matrix comparison failures should be good.  Both values should be printed in full, and the index that differs should receive a mention.

I think you are misunderstanding what I have in mind overall, but that's fine.  There's no real need for me to explain it: adding ADL overloads for `CompareMatrices` in distinct header files is not defective.

SeanCurtis-TRI (commented):
My understanding of how matchers work is that the error message is largely controlled by gtest with only a couple of levers to pull by the user (see the example up above).

I tried something clever/tricky/stupid on the new revision. The goal was two fold:

1. Eliminate explicit dependency on `//math:geometric_transform`.
2. Make the overloads immediately available to anyone currently using eigen matrix compare.

I can imagine a difference in compile time. Thoughts? Too esoteric? 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NM0n5OiDCDeHyQdeb_W:b-gvvjwm)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
+@jwnimmer-tri for feature review. There are enough votes (at least around HQ) that I think it's worth going forward with this version.

+(release notes: none) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-:-NMUTJdF12llzSCcc8ns:b-7m9egi)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-NMUzr7h80Jvktq4ewTo:-NMUzr7iCPF2fUsvGl96:b-eunr6w) (no related file):*
For changes that propose new sugar, reviewers should be shown examples of the effect of the sugar so that they can understand its efficacy -- either by adding some examples in the same PR here, or else via an overview hyperlink to sample changes that would land later in the PR train.  (The code snippet in the overview is not enough; it's important to see the example(s) in-situ.)

In particular, I'm unconvinced that for a RigidTransform the matrix comparisons of the rotation matrix and translation vector should use the same tolerance for each when the tolerance is non-zero.  My understanding is that those two parts do not have comparable numerical scaling, in general.

___
___
*[`common/test_utilities/eigen_matrix_compare.h` line 17 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-NMV-cc205dGw2p7tqcr:-NMV-cc205dGw2p7tqcs:b8oscmq) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d65ad35009a73765c2a3d04e1fa80cb87606c70/common/test_utilities/eigen_matrix_compare.h#L17)):*
> ```objc
> 
> /**
>  * Compares two matrices to determine whether they are equal to within a certain
> ```

nit The new and old comment styles are not consistent within this file.  Either the old should lose the line-leading `*`s or the new should add the `*`s.
___
*[`common/test_utilities/eigen_matrix_compare.h` line 146 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-NMV1xrB5Ih3IVs8k8lb:-NMV1xrC19ojjyPKLrL-:b-b6xetl) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d65ad35009a73765c2a3d04e1fa80cb87606c70/common/test_utilities/eigen_matrix_compare.h#L146)):*
> ```objc
>  */
> template <template <typename> typename TransformType, typename T>
> [[nodiscard]] std::enable_if_t<
> ```

I am OK with the shibboleth idea for solving dependency inversion, but we should make the match as tight as possible.  We don't want to accidentally call some type that _happens_ to have a `GetAsMatrix34` function in cases where the that function does not correctly biject the equality state of the object.

One way to make this much tighter would be to guard the size of the object vs `sizeof(double) * 12`.  That would substantially limit the false positives, probably to the point where a false positive is vanishingly small.

The same goes for the RotationMatrix shibboleth (with 9 doubles).

Another way to make it tighter would be mixing in even more functions, like requiring `InvertAndCompose` and `GetAsMatrix34` both.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-NMUzr7h80Jvktq4ewTo:-NMV44hu9Dds7aoh4-uT:bbih61) (no related file):*

For changes that propose new sugar, reviewers should be shown examples of the effect of the sugar so that they can understand its efficacy -- either by adding some examples in the same PR here, or else via an overview hyperlink to sample changes that would land later in the PR train.  (The code snippet in the overview is not enough; it's important to see the example(s) in-situ.)

In particular, I'm unconvinced that for a RigidTransform the matrix comparisons of the rotation matrix and translation vector should use the same tolerance for each when the tolerance is non-zero.  My understanding is that those two parts do not have comparable numerical scaling, in general.
</blockquote></details>

In particular, for the common case of zero-tolerance absolute comparisons, wouldn't it even even better if this did something reasonable?

```
EXPECT_EQ(X_AB, X_AB_expected);
```

Maybe all of the call sites calling `GetAsMatrix34` in that case are silly in the first place?  Maybe it's silly that `RigidTransform::operator==` is not defined, since clearly we frequently want to compare them for exact equality?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18584)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18584#-NMUzr7h80Jvktq4ewTo:-NMV6H5N-PeYAEKc4CjS:b-knzusp) (no related file):*

In particular, for the common case of zero-tolerance absolute comparisons, wouldn't it even even better if this did something reasonable?

```
EXPECT_EQ(X_AB, X_AB_expected);
```

Maybe all of the call sites calling `GetAsMatrix34` in that case are silly in the first place?  Maybe it's silly that `RigidTransform::operator==` is not defined, since clearly we frequently want to compare them for exact equality?
</blockquote></details>

Or perhaps put another more actionable way -- when providing examples of using this sugar, I'd like to consider the zero-tolerance an nonzero-tolerance cases separately, so please highlight examples of both.

___

PR #18587, opened 2023-01-13
[pydrake] Introduce name mangling for template classes

jwnimmer-tri:
For template instantations, use Unicode mangling instead of square brackets for the underlying class name, and similarly for commas and periods within the template param. Python class names are not allowed to use brackets; doing so cases headaches for tools like Mypy. On the other hand, for the website we always want the demangled names so turn off name mangling during docs generation via a giant hack.

When displaying possibly-mangled class names (e.g., in `repr`), as much as practical in the bindings use the demangling function to provide the more natural expression for users to see.

Add a regression test that Mypy can parse our `*.pyi` files (modulo one file that still needs some grooming).

---

Towards #18580.  Only a few `solvers` typos remain before Mypy is working.

Requires:
- [x] #18588
- [x] #18589
- [x] #18591
- [x] #18592
- [x] #18593
- [x] #18595

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18587)

Issue #18580, opened 2023-01-12
Incompatibility between pydrake and Mypy


jwnimmer-tri:
See #16987 for some background.

In our `pydrake` releases, we install `*.pyi` files so that people who use VSCode will have auto-complete enabled.

However, our `*.pyi` files are currently malformed in several ways, which means that if someone tries to run `mypy` type checking with `pydrake` in the loop, it will error out with syntax errors.

The work-around for the moment is to delete the `*.pyi` files if you intent to use `mypy`.

(Details to follow.)

jwnimmer-tri:
Here's a tip from a user who found another work-around: if you move the `*.pyi` files out of your venv and into a `typings` directory, apparently VSCode will still find them there, but it won't confuse Mypy.

PR #18593, opened 2023-01-16
[multibody] Remove markup that destroys Python

jwnimmer-tri:
In the `shape` overload of `CalcSpatialInertia`, the hyperlink to the other overload just below on the page ends up unparseable in Python. Reword to use slightly less precise (but simpler) phrasing.

(Ideally, our API documentation parsing tools would be able to handle cross-references like this, but there are so many layers of parsing, transformation, reparsing, retransformation, etc. that sometimes discretion is the better part of valor.)

Towards https://github.com/RobotLocomotion/drake/issues/18580.
Required for the Mypy regression test in #18587 to pass.

---

FYI here are the current Python docs on master, showing the unparsed cross-reference:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18593)

### DIFF
--- a/multibody/tree/geometry_spatial_inertia.h
+++ b/multibody/tree/geometry_spatial_inertia.h
@@ -18,9 +18,8 @@ namespace multibody {
  (and, therefore, Bo). These are the shapes that have symmetry across So along
  each of the axes Sx, Sy, Sz (e.g., geometry::Box, geometry::Sphere, etc.) For
  meshes, it depends on how the mesh is defined. For more discussion on the
- nuances of geometry::Mesh and geometry::Convex calculations
- @ref CalcSpatialInertia(const geometry::TriangleSurfaceMesh<double>&,double)
- "see below".
+ nuances of geometry::Mesh and geometry::Convex calculations refer to the
+ `mesh` overload, below.
 
  Note: Spatial inertia calculations for the geometry::Convex type do not
  currently require that the underlying mesh actually be convex. Although certain
@@ -29,8 +28,7 @@ namespace multibody {
 
  @retval M_BBo_B The spatial inertia of the hypothetical body implied by the
                  given `shape`.
- @throws std::exception if `shape` is an instance of geometry::HalfSpace or
-                        geometry::MeshcatCone.
+ @throws std::exception if `shape` is not supported (e.g., a half space)
  @pydrake_mkdoc_identifier{shape} */
 SpatialInertia<double> CalcSpatialInertia(const geometry::Shape& shape,
                                           double density);


jwnimmer-tri:
+@SeanCurtis-TRI for both reviews, please.

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18593#-:-NM-f1TzDH6UmPNLDEZk:bmxilld)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`multibody/tree/geometry_spatial_inertia.h` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18593#-NM-ek-yDxL0mIr7boLl:-NM-ek-yDxL0mIr7boLm:bipyqey) ([raw file](https://github.com/RobotLocomotion/drake/blob/6bc9cf0f29624cc0b7e635e74342f2f87018ce98/multibody/tree/geometry_spatial_inertia.h#L32)):*
> ```objc
>                  given `shape`.
>  @throws std::exception if `shape` is an instance of geometry::HalfSpace or
>                         geometry::MeshcatCone.
> ```

BTW As long as you're here, it turns out the pydrake documentation treats this hard-line break badly as well. This second line gets interpreted as a unique bullet point. :(

I have to say, for anyone who cares about layout, our documentation mechanism is a source of torment.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18593)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`multibody/tree/geometry_spatial_inertia.h` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18593#-NM-ek-yDxL0mIr7boLl:-NM-lMNT34YckfIhfD0K:b6wsr4r) ([raw file](https://github.com/RobotLocomotion/drake/blob/6bc9cf0f29624cc0b7e635e74342f2f87018ce98/multibody/tree/geometry_spatial_inertia.h#L32)):*

BTW As long as you're here, it turns out the pydrake documentation treats this hard-line break badly as well. This second line gets interpreted as a unique bullet point. :(

I have to say, for anyone who cares about layout, our documentation mechanism is a source of torment.
</blockquote></details>

It wasn't the hard line break causing the problem, it was the class names like `geometry::Foo` being misinterpreted.  Saying "half space" instead of the hyperlink phrase solves the problem.

I also elected to remove the full list of all unsupported shapes.  To me, that kind of list is too difficult to maintain over time, e.g., if we add new kinds of shapes.  Also in the case of MeshcatCone, a _promise_ to throw means that adding an implementation for it in the future would mean _withdrawing_ promise to throw. We don't want users relying on that promise, so we shouldn't make it.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18593)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`multibody/tree/geometry_spatial_inertia.h` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18593#-NM-ek-yDxL0mIr7boLl:-NM0AjUg-OhbfQhMW9iP:b-jv4e7b) ([raw file](https://github.com/RobotLocomotion/drake/blob/6bc9cf0f29624cc0b7e635e74342f2f87018ce98/multibody/tree/geometry_spatial_inertia.h#L32)):*

It wasn't the hard line break causing the problem, it was the class names like `geometry::Foo` being misinterpreted.  Saying "half space" instead of the hyperlink phrase solves the problem.

I also elected to remove the full list of all unsupported shapes.  To me, that kind of list is too difficult to maintain over time, e.g., if we add new kinds of shapes.  Also in the case of MeshcatCone, a _promise_ to throw means that adding an implementation for it in the future would mean _withdrawing_ promise to throw. We don't want users relying on that promise, so we shouldn't make it.
</blockquote></details>

OK I can kind of see that. However, the user has no way of knowing what is unsupported short of simply calling and finding out if it explodes. That feels dissatisfying.  In this case, I feel the enumerated exceptions are just that, exceptions. They're not representative of a collection that will grow in the future (in fact, I would hope that it shrinks).

That said, the supportability is challenging. Adding a new shape can have far reaching tendrils. Part of what I'm doing in my effort to add a Cone is to make a checklist for adding a new shape, so that it can be accounted for across the board.

Nevertheless, while this change isn't lossless, I'd certainly deem it reasonable. It's worth noting that these are not the only namespace-qualified classes mentioned in throw declarations (I assume). Lots of other problems remain lurking.

jwnimmer-tri:
*[`multibody/tree/geometry_spatial_inertia.h` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18593#-NM-ek-yDxL0mIr7boLl:-NM0FdzM8sWhVK2lg4PH:b-y8c5wh) ([raw file](https://github.com/RobotLocomotion/drake/blob/6bc9cf0f29624cc0b7e635e74342f2f87018ce98/multibody/tree/geometry_spatial_inertia.h#L32)):*
> However, the user has no way of knowing what is unsupported short of simply calling and finding out if it explodes. That feels dissatisfying.

I think we differ in our priors for how likely users are to read documentation with such care.  I am OK to defer to your preference, though.

To document the limitations, however, `@throws` is not the correct markup.  Markup along the lines of a TODO or Warning is what we should use.  Documented exceptions are a post-condition of the function's API.

PR #13417, opened 2020-05-27
Add GetInfeasibleConstraintBindings.

hongkai-dai:
returns a vector of Binding objects for the infeasible constraints, useful when debugging infeasible mathematical optimization problems.

As mentioned in #13091

PR #18607, opened 2023-01-18
[pydrake] Fix MathematicalProgramResult.num_suboptimal_solution binding

jwnimmer-tri:
Amends #13417.  Towards #18580.


jwnimmer-tri:
+@sherm1 for both reviews per schedule, please.

Issue #14431, opened 2020-12-08
Make TRI motion planning tools available in drake


calderpg-tri:
TRI uses a number of motion planning tools, both external (public) and internal that we would like to make available in drake. These tools include:

- Sampling-based motion planners (constrained and unconstrained RRT/BiRRT variants, PRM variants)
- Path and trajectory processing (shortcut and b-spline smoothing, TOTP [Time-optimal trajectory parametrization])
- Voxel-based geometry (dense voxelized occupancy and signed distance fields for collision checking/gradients)
- GPU-accelerated pointcloud->voxelized occupancy
- Collision checking interfaces and abstractions for parallelized collision checking, IK, planning, and optimization
- Additional constraints for collision avoidance
- Utilities used in planning problems (kNN, A*, some forms of clustering, etc)

Since the TRI-internal tools include interfaces with ROS for visualization, we cannot simply port these tools over until drake-ROS integration is more developed. Therefore, we propose to bring in the existing public tools into drake as externals and make the ROS-free subset of internal tools available in drake (i.e. in a new directory such as `drake/multibody/planning`) via a tool such as [copybara](https://github.com/google/copybara). Using (for example) copybara, the TRI-internal version would remain the authoritative version while contributions would be accepted from both public drake and TRI-internal.

Roughly speaking, the steps to complete this:

1. Bring in [common_robotics_utilities](https://github.com/calderpg/common_robotics_utilities) and [voxelized_geometry_tools](https://github.com/calderpg/voxelized_geometry_tools) as externals. The limited ROS-dependent parts of these packages will not be built.
2. Add build support for OpenMP (directive-based CPU parallelization) and OpenCL (cross-platform GPU offload) to drake. [Note that while both OpenGL and OpenCL are officially deprecated by Apple, they have remained available for several releases]
3. Export the ROS-free subset of `anzu/planning` to `drake/multibody/planning` using a tool like copybara.
4. Provide additional public documentation and demos of the new planning tools.

Farther in the future, once drake-ROS integration has developed and sufficient visualization capabilities are exposed in drake itself, these TRI-internal planning tools will be moved completely to drake.

cc @ggould-tri @jwnimmer-tri @avalenzu @RussTedrake

jamiesnape:
If our AWS instances do not do support OpenCL, is that going to be a blocker?

calderpg-tri:
[...]

PR #18558, opened 2023-01-09
[planning] Add CollisionChecker

SeanCurtis-TRI:
Towards #14431.

|  Topic  | Total lines | Reviewer | Code sections |
| -------| -----------| ------------| ------------- |
| Contexts | 498 | @jwnimmer-tri | `collision_checker.h`:25-148, 1029-1053, 1137-1148, 1202-1317 <br> `collision_checker.cc`: 177-201, 814-848, 858-862, 1074-1111 <br> `collision_checker_test.cc`: 469-480, 567-594, 787-812 |
| Base Class | 395 | @sammy-tri | `collision_checker.h`: 149-223, 1014-1028, 1128-1131 <br> `collision_checker.cc`: 155-176, 769-813, 849-853 <br> `collision_checker_test.cc`: 77-124, 352-468, 778-786 <br> `BUILD.bazel`: all lines |
| Shapes | 229 | @sherm1 | `collision_checker.h`: 276-385 <br> `collision_checker.cc`: 202-290 <br> `collision_checker_test.cc`: 748-777 |
| Collision filters | 720 | @EricCousineau-TRI  | `collision_checker.h`: 537-651, 1149-1180 <br> `collision_checker.cc`: 417-471, 863-1013 <br> `collision_checker_test.cc`: 835-856 |
| Collision state | 336 | @sherm1 | `collision_checker.h`: 652-685, 950-1013 <br> `collision_checker.cc`: 472-501, 78-154, 728-768 <br> `collision_checker_test.cc`: 481-526, 813-834, 835-856 |
| Edge checks | 1317 | @xuchenhan-tri | `collision_checker.h`: 686-949 <br> `collision_checker.cc`: 42-77, 502-727 <br> `collision_checker_test.cc`: 22-40, 1224-1906 <br> `edge_measure.h`: all lines |
|Padding | 520 | @ggould-tri  | `collision_checker.h`: 386-536, 1132-1136, 1181-1201 <br> `collision_checker.cc`: 291-416, 854-857, 1014-1073 <br> `collision_checker_test.cc`: 696-747 |
|Unimplemented | 178 | @sammy-tri | `unimplemented*`: all |

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18558)

PR #18608, opened 2023-01-18
[planning] Add ComputeCollisionAvoidanceDisplacement

jwnimmer-tri:
Towards #14431.

+@sherm1 for platform review per schedule, please.  This code has already been feature-reviewed in Anzu.

\CC @calderpg-tri @rpoyner-tri @SeanCurtis-TRI

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)

### DIFF
--- a/planning/BUILD.bazel
+++ b/planning/BUILD.bazel
@@ -15,6 +15,7 @@ drake_cc_package_library(
     visibility = ["//visibility:public"],
     deps = [
         ":body_shape_description",
+        ":collision_avoidance",
         ":collision_checker",
         ":collision_checker_context",
         ":collision_checker_params",
@@ -34,6 +35,15 @@ drake_cc_library(
         "//common:essential",
         "//geometry",
         "//multibody/plant",
+    ],
+)
+
+drake_cc_library(
+    name = "collision_avoidance",
+    srcs = ["collision_avoidance.cc"],
+    hdrs = ["collision_avoidance.h"],
+    deps = [
+        ":collision_checker",
     ],
 )
 
@@ -139,6 +149,16 @@ drake_cc_googletest(
         ":robot_diagram_builder",
         "//multibody/parsing",
         "//multibody/plant",
+    ],
+)
+
+drake_cc_googletest(
+    name = "collision_avoidance_test",
+    deps = [
+        ":collision_avoidance",
+        ":robot_diagram_builder",
+        ":unimplemented_collision_checker",
+        "//common/test_utilities:eigen_matrix_compare",
     ],
 )
 
--- a/planning/collision_avoidance.cc
+++ b/planning/collision_avoidance.cc
@@ -1,0 +1,48 @@
+#include "planning/collision_avoidance.h"
+
+#include <algorithm>
+#include <cmath>
+
+namespace anzu {
+namespace planning {
+
+using drake::planning::CollisionChecker;
+using drake::planning::CollisionCheckerContext;
+using drake::planning::RobotClearance;
+
+Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
+    const CollisionChecker& checker, const Eigen::VectorXd& q,
+    double max_penetration, double max_clearance,
+    CollisionCheckerContext* context) {
+  DRAKE_THROW_UNLESS(max_penetration <= 0.0);
+  DRAKE_THROW_UNLESS(std::isfinite(max_penetration));
+  DRAKE_THROW_UNLESS(max_clearance >= 0.0);
+  DRAKE_THROW_UNLESS(std::isfinite(max_clearance));
+  DRAKE_THROW_UNLESS(max_clearance > max_penetration);
+
+  const double penetration_range = max_clearance - max_penetration;
+  // Collect the distances and Jacobians.
+  const RobotClearance robot_clearance =
+      context == nullptr
+          ? checker.CalcRobotClearance(q, max_clearance)
+          : checker.CalcContextRobotClearance(context, q, max_clearance);
+
+  const int num_rows = robot_clearance.size();
+
+  if (num_rows == 0) {
+    // No collisions within the safety range; so no displacement required.
+    return Eigen::VectorXd::Zero(q.size());
+  }
+
+  // Compute weights based on distance.
+  Eigen::VectorXd weights(num_rows);
+  for (int row = 0; row < num_rows; ++row) {
+    const double distance = robot_clearance.distances()(row);
+    const double penetration = max_clearance - distance;
+    weights(row) = std::clamp(penetration / penetration_range, 0.0, 1.0);
+  }
+
+  return robot_clearance.jacobians().transpose() * weights;
+}
+}  // namespace planning
+}  // namespace anzu
--- a/planning/collision_avoidance.cc
+++ b/planning/collision_avoidance.cc
@@ -1,14 +1,11 @@
-#include "planning/collision_avoidance.h"
+#include "drake/planning/collision_avoidance.h"
 
 #include <algorithm>
 #include <cmath>
 
-namespace anzu {
+namespace drake {
 namespace planning {
-
-using drake::planning::CollisionChecker;
-using drake::planning::CollisionCheckerContext;
-using drake::planning::RobotClearance;
+namespace internal {
 
 Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
     const CollisionChecker& checker, const Eigen::VectorXd& q,
@@ -27,22 +24,30 @@ Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
           ? checker.CalcRobotClearance(q, max_clearance)
           : checker.CalcContextRobotClearance(context, q, max_clearance);
 
-  const int num_rows = robot_clearance.size();
+  const int num_measurements = robot_clearance.size();
 
-  if (num_rows == 0) {
+  if (num_measurements == 0) {
     // No collisions within the safety range; so no displacement required.
     return Eigen::VectorXd::Zero(q.size());
   }
 
   // Compute weights based on distance.
-  Eigen::VectorXd weights(num_rows);
-  for (int row = 0; row < num_rows; ++row) {
+  Eigen::VectorXd weights(num_measurements);
+  for (int row = 0; row < num_measurements; ++row) {
     const double distance = robot_clearance.distances()(row);
     const double penetration = max_clearance - distance;
     weights(row) = std::clamp(penetration / penetration_range, 0.0, 1.0);
   }
 
+  // Here we compute ∇f = Σᵢ(wᵢ⋅Jqᵣ_ϕᵢ)ᵀ. Per the RobotClearance API docs we
+  // have ϕᵢ as the iᵗʰ distance measurement and Jqᵣ_ϕᵢ as the partial of ϕᵢ
+  // with respect to qᵣ, the robot state; wᵢ is our weight for iᵗʰ measurement.
+  //
+  // Note that Jqᵣ_ϕ is dimensioned as (num_measurements, nq) and weights is
+  // (num_measurements, 1), so we can use Jᵀ as a convenient way to calculate
+  // the sum in a single pass.
   return robot_clearance.jacobians().transpose() * weights;
 }
+}  // namespace internal
 }  // namespace planning
-}  // namespace anzu
+}  // namespace drake
--- a/planning/collision_avoidance.h
+++ b/planning/collision_avoidance.h
@@ -1,0 +1,60 @@
+#pragma once
+
+#include "drake/planning/collision_checker.h"
+
+namespace anzu {
+namespace planning {
+
+/** Computes a displacement in configuration space, Δq, that moves the provided
+ configuration `q` into a "clearer" state, reducing the likelihood of collision.
+
+ The displacement vector Δq = Σᵢ(wᵢ⋅Jqᵣ_ϕᵢ), where
+
+   - The iᵗʰ term comes from the iᵗʰ clearance measurement on the robot. It may
+     measure the clearance between robot and environment or robot and robot.
+   - The number of measurements depends on how the robot is represented
+     geometrically in `checker`.
+   - ϕᵢ is the iᵗʰ measured distance and Jq_ϕᵢ is the partial derivative of ϕᵢ
+     with respect to qᵣ, the robot state.
+   - wᵢ is the weight for iᵗʰ measurement, defined as wᵢ = (c - ϕᵢ) / (c - p),
+     where c = `max_clearance` and p = `max_penetration` and c > p.
+
+ Intuitively, the displacement vector Δq tries to reconcile multiple
+ simultaneous potential collisions via a penalty method. The closer to
+ penetration, the more weight that potential collision has. No measurements
+ greater than `max_clearance` contribute to the calculation, and any distances
+ less than `max_penetration` are all treated equally, regardless of the degree
+ of penetration.
+
+ @param checker             The checker used to define ϕᵢ and Jqᵣ_ϕᵢ.
+ @param q                   The robot configuration at which we compute Δq.
+ @param max_penetration     The bottom of the range; a non-positive value
+                            representing the level of penetration that saturates
+                            the weight. Penetration occurs inside an object
+                            where ϕ < 0. Setting the value to zero will give all
+                            penetration equal weight, regardless of depth.
+                            Making this value *more* negative won't increase the
+                            weight of deep penetrations, but it will reduce the
+                            weight of shallow penetrations.
+ @param max_clearance       The top of the range; a non-negative value beyond
+                            which possible collisions do not contribute. Two
+                            objects are "clear" where ϕ > 0. Setting the value
+                            to zero will only consider collisions and not nearby
+                            objects. As the value increases, more and more
+                            distant objects will influence the calculation.
+ @param context             An optional collision checker context. If none is
+                            provided, the checker's context for the current
+                            thread is used.
+ @retval Δq the configuration displacement to increase robot clearance.
+ @pre q.size() == checker.GetZeroConfiguration().size().
+ @pre max_penetration <= 0.
+ @pre max_clearance >= 0.
+ @pre max_clearance > max_penetration.
+ @pre if context != nullptr, it is a context managed by checker. */
+Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
+    const drake::planning::CollisionChecker& checker, const Eigen::VectorXd& q,
+    double max_penetration, double max_clearance,
+    drake::planning::CollisionCheckerContext* context = nullptr);
+
+}  // namespace planning
+}  // namespace anzu
--- a/planning/collision_avoidance.h
+++ b/planning/collision_avoidance.h
@@ -2,32 +2,24 @@
 
 #include "drake/planning/collision_checker.h"
 
-namespace anzu {
+namespace drake {
 namespace planning {
+namespace internal {
 
-/** Computes a displacement in configuration space, Δq, that moves the provided
- configuration `q` into a "clearer" state, reducing the likelihood of collision.
+/* Computes the gradient ∇f = (∂f/∂q)ᵀ of an unspecifed distance function "f".
 
- The displacement vector Δq = Σᵢ(wᵢ⋅Jqᵣ_ϕᵢ), where
+ "f" combines the distances ϕᵢ of the given collision checker's RobotClearance
+ data, reconciling multiple simultaneous potential collisions via a penalty
+ method. The closer to penetration, the more weight that potential collision
+ has. No measurements greater than `max_clearance` contribute to the
+ calculation, and any distances less than `max_penetration` are all treated
+ equally, regardless of the degree of penetration.
 
-   - The iᵗʰ term comes from the iᵗʰ clearance measurement on the robot. It may
-     measure the clearance between robot and environment or robot and robot.
-   - The number of measurements depends on how the robot is represented
-     geometrically in `checker`.
-   - ϕᵢ is the iᵗʰ measured distance and Jq_ϕᵢ is the partial derivative of ϕᵢ
-     with respect to qᵣ, the robot state.
-   - wᵢ is the weight for iᵗʰ measurement, defined as wᵢ = (c - ϕᵢ) / (c - p),
-     where c = `max_clearance` and p = `max_penetration` and c > p.
+ Therefore, the gradient ∇f points `q` into a "clearer" state, reducing the
+ likelihood of collision.
 
- Intuitively, the displacement vector Δq tries to reconcile multiple
- simultaneous potential collisions via a penalty method. The closer to
- penetration, the more weight that potential collision has. No measurements
- greater than `max_clearance` contribute to the calculation, and any distances
- less than `max_penetration` are all treated equally, regardless of the degree
- of penetration.
-
- @param checker             The checker used to define ϕᵢ and Jqᵣ_ϕᵢ.
- @param q                   The robot configuration at which we compute Δq.
+ @param checker             The checker used to calculate clearance data.
+ @param q                   The robot configuration at which we compute ∂f/∂q.
  @param max_penetration     The bottom of the range; a non-positive value
                             representing the level of penetration that saturates
                             the weight. Penetration occurs inside an object
@@ -45,16 +37,25 @@ namespace planning {
  @param context             An optional collision checker context. If none is
                             provided, the checker's context for the current
                             thread is used.
- @retval Δq the configuration displacement to increase robot clearance.
  @pre q.size() == checker.GetZeroConfiguration().size().
  @pre max_penetration <= 0.
  @pre max_clearance >= 0.
  @pre max_clearance > max_penetration.
  @pre if context != nullptr, it is a context managed by checker. */
 Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
-    const drake::planning::CollisionChecker& checker, const Eigen::VectorXd& q,
+    const CollisionChecker& checker, const Eigen::VectorXd& q,
     double max_penetration, double max_clearance,
-    drake::planning::CollisionCheckerContext* context = nullptr);
+    CollisionCheckerContext* context = nullptr);
 
+// TODO(jwnimmer-tri) Before we promote the above function out of the internal
+// namespace, consider renaming "Displacement" to "Gradient" to better reflect
+// the units on the return type.
+
+// TODO(jwnimmer-tri) Before we promote the above function out of the internal
+// namespace, consider changing its optional `context` parameter above into a
+// distinct function named ComputeContextCollisionAvoidanceDisplacement, for
+// consistency with how CollisionChecker names its member functions.
+
+}  // namespace internal
 }  // namespace planning
-}  // namespace anzu
+}  // namespace drake
--- a/planning/test/collision_avoidance_test.cc
+++ b/planning/test/collision_avoidance_test.cc
@@ -1,0 +1,213 @@
+#include "planning/collision_avoidance.h"
+
+#include <memory>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/planning/robot_diagram_builder.h"
+#include "drake/planning/unimplemented_collision_checker.h"
+
+namespace anzu {
+namespace planning {
+namespace {
+
+using drake::CompareMatrices;
+using drake::multibody::BodyIndex;
+using drake::multibody::default_model_instance;
+using drake::multibody::SpatialInertia;
+using drake::planning::CollisionCheckerContext;
+using drake::planning::RobotClearance;
+using drake::planning::RobotCollisionType;
+using drake::planning::RobotDiagram;
+using drake::planning::RobotDiagramBuilder;
+using drake::planning::UnimplementedCollisionChecker;
+using Eigen::VectorXd;
+using std::move;
+using std::unique_ptr;
+
+double ZeroDistanceFunc(const VectorXd&, const VectorXd&) { return 0.0; }
+
+/* We simply need a diagram with a non-zero number of dofs; a single floating
+ body will give us 7 -- the semantics are unimportant. */
+unique_ptr<RobotDiagram<double>> MakeRobot() {
+  RobotDiagramBuilder<double> builder;
+  const auto model_instance = drake::multibody::default_model_instance();
+  builder.mutable_plant().AddRigidBody("floater", model_instance,
+                                       SpatialInertia<double>::MakeUnitary());
+  return builder.BuildDiagram();
+}
+
+/* A minimum implementation of the CollisionChecker, but one that allows us to
+ explicitly specify the result of CalcRobotClearance(). */
+class DummyCollisionChecker final : public UnimplementedCollisionChecker {
+ public:
+  /* I'm passing garbage parameters to UnimplementedCollisionChecker. That's ok;
+   the only API we're exercising is CalcRobotClearance() directly and
+   DoCalcContextRobotClearance() and DoUpdateContextPositions() indirectly.
+   We've overridden the latter so they don't care about the garbage. */
+  DummyCollisionChecker()
+      : UnimplementedCollisionChecker(
+            {.model = MakeRobot(),
+             .robot_model_instances = {default_model_instance()},
+             .configuration_distance_function = ZeroDistanceFunc,
+             .edge_step_size = 0.1}, false /* supports_parallel */),
+        data_(plant().num_positions()) {
+    AllocateContexts();
+  }
+
+  /* This clearance data will define the distance/jacobian data that will be
+   given back to ComputeCollisionAvoidanceDisplacement() when invoking
+   CalcRobotClearance(). */
+  void set_robot_clearance(RobotClearance data) {
+    data_ = move(data);
+  }
+
+ private:
+  void DoUpdateContextPositions(
+      CollisionCheckerContext* model_context) const final {
+    /* Invoked indirectly; but we won't do any work. */
+    return;
+  }
+
+  RobotClearance DoCalcContextRobotClearance(
+      const CollisionCheckerContext& model_context,
+      double influence_distance) const final {
+    return data_;
+  }
+
+  RobotClearance data_;
+};
+
+/* ComputeCollisionAvoidanceDisplacement() is uniquely responsible for handling:
+    - Calling the right CollisionChecker API based on presence of a context
+      (i.e., context provenance).
+    - Empty distance measurements produces zero vector.
+    - Weights are properly computed.
+
+Each of those gets us own unit test case in the below. */
+
+GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, ContextProvenance) {
+  DummyCollisionChecker checker;
+  /* Has to match the 7 dofs in the robot. */
+  constexpr int kQSize = 7;
+  const VectorXd q1 = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
+  const VectorXd q2 = (VectorXd(kQSize) << 9, 8, 7, 6, 5, 4, 3).finished();
+
+  /* Confirm context provenance. We'll show that the expected context has been
+   written to as evidence of provenance. */
+
+  /* Implicit context. */
+  EXPECT_FALSE(CompareMatrices(
+      checker.plant().GetPositions(checker.plant_context()), q1));
+  EXPECT_NO_THROW(ComputeCollisionAvoidanceDisplacement(checker, q1, 0, 1));
+  EXPECT_TRUE(CompareMatrices(
+      checker.plant().GetPositions(checker.plant_context()), q1));
+
+  /* Explicit context. */
+  auto context = checker.MakeStandaloneModelContext();
+  EXPECT_FALSE(CompareMatrices(
+      checker.plant().GetPositions(context->plant_context()), q2));
+  EXPECT_NO_THROW(
+      ComputeCollisionAvoidanceDisplacement(checker, q2, 0, 1, context.get()));
+  EXPECT_TRUE(CompareMatrices(
+      checker.plant().GetPositions(context->plant_context()), q2));
+}
+
+GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, NoDistanceMeasurements) {
+  DummyCollisionChecker checker;
+  /* Has to match the 7 dofs in the robot. */
+  constexpr int kQSize = 7;
+  const VectorXd q = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
+
+  /* The checker returns an empty RobotClearance by default. This should lead to
+   zero-valued Δq. We're passing in an arbitrary, non-zero, q to eliminate the
+   appearance that the return value copies the input value. */
+  EXPECT_TRUE(CompareMatrices(
+      ComputeCollisionAvoidanceDisplacement(checker, q, 0, 1),
+      VectorXd::Zero(kQSize)));
+}
+
+GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
+  DummyCollisionChecker checker;
+  /* Has to match the 7 dofs in the robot. */
+  constexpr int kQSize = 7;
+  const VectorXd q0 = VectorXd::Zero(kQSize);
+  const VectorXd q1 = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
+  const VectorXd q2 = (VectorXd(kQSize) << 9, 8, 7, 6, 5, 4, 3).finished();
+
+  /* Confirm weight calculation. Using an arbitrary safety range, we'll specify
+   distances around the range and confirm the Jacobians have been properly
+   scaled. */
+  const double kMin = -0.25;
+  const double kMax = 0.75;
+  const double kRange = kMax - kMin;
+
+  /* We'll use a robot-env collision; ComputeCollisionAvoidanceDisplacement()
+   ignores this field. So, robot-env and robot-robot measurements should be
+   treated identically. */
+  const auto& plant = checker.plant();
+  const BodyIndex robot_body = plant.GetBodyByName("floater").index();
+  const BodyIndex env_body = plant.GetBodyByName("world").index();
+  const auto env_type = RobotCollisionType::kEnvironmentCollision;
+
+  /* In the following tests, the q passed to
+   ComputeCollisionAvoidanceDisplacement() doesn't matter. Only the qs in
+   the collision avoidance data matter. */
+
+  {
+    /* Deeper than kMin (max_penetration), saturates to a weight of 1. */
+    RobotClearance clearance(kQSize);
+    clearance.Append(robot_body, env_body, env_type, kMin - 0.1,
+                     q1.transpose());
+    checker.set_robot_clearance(move(clearance));
+    const VectorXd delta_q =
+        ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
+    EXPECT_TRUE(CompareMatrices(delta_q, q1));
+  }
+
+  {
+    /* Farther than kMax (max_penetration), saturates to a weight of 0. */
+    RobotClearance clearance(kQSize);
+    clearance.Append(robot_body, env_body, env_type, kMax + 0.1,
+                     q1.transpose());
+    checker.set_robot_clearance(move(clearance));
+    const VectorXd delta_q =
+        ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
+    EXPECT_TRUE(CompareMatrices(delta_q, q0));
+  }
+
+  {
+    /* Pick an arbitrary weight between 0 and 1. */
+    constexpr double kWeight = 1.0 / 3.0;
+    RobotClearance clearance(kQSize);
+    clearance.Append(robot_body, env_body, env_type, kMax - kRange * kWeight,
+                     q1.transpose());
+    checker.set_robot_clearance(move(clearance));
+    const VectorXd delta_q =
+        ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
+    EXPECT_EQ(delta_q.size(), kQSize);
+    EXPECT_TRUE(CompareMatrices(delta_q, kWeight * q1));
+  }
+
+  {
+    /* Two arbitrary weights combine. */
+    constexpr double kWeight1 = 1.0 / 3.0;
+    constexpr double kWeight2 = 3.0 / 5.0;
+    RobotClearance clearance(kQSize);
+    clearance.Append(robot_body, env_body, env_type, kMax - kRange * kWeight1,
+                     q1.transpose());
+    clearance.Append(robot_body, env_body, env_type, kMax - kRange * kWeight2,
+                     q2.transpose());
+    checker.set_robot_clearance(move(clearance));
+    const VectorXd delta_q =
+        ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
+    EXPECT_EQ(delta_q.size(), kQSize);
+    EXPECT_TRUE(CompareMatrices(delta_q, kWeight1 * q1 + kWeight2 * q2));
+  }
+}
+
+}  // namespace
+}  // namespace planning
+}  // namespace anzu
--- a/planning/test/collision_avoidance_test.cc
+++ b/planning/test/collision_avoidance_test.cc
@@ -9,20 +9,14 @@
 #include "drake/planning/robot_diagram_builder.h"
 #include "drake/planning/unimplemented_collision_checker.h"
 
-namespace anzu {
+namespace drake {
 namespace planning {
+namespace internal {
 namespace {
 
-using drake::CompareMatrices;
 using drake::multibody::BodyIndex;
 using drake::multibody::default_model_instance;
 using drake::multibody::SpatialInertia;
-using drake::planning::CollisionCheckerContext;
-using drake::planning::RobotClearance;
-using drake::planning::RobotCollisionType;
-using drake::planning::RobotDiagram;
-using drake::planning::RobotDiagramBuilder;
-using drake::planning::UnimplementedCollisionChecker;
 using Eigen::VectorXd;
 using std::move;
 using std::unique_ptr;
@@ -38,6 +32,9 @@ unique_ptr<RobotDiagram<double>> MakeRobot() {
                                        SpatialInertia<double>::MakeUnitary());
   return builder.BuildDiagram();
 }
+
+/* Has to match the 7 dofs in the robot. */
+constexpr int kQSize = 7;
 
 /* A minimum implementation of the CollisionChecker, but one that allows us to
  explicitly specify the result of CalcRobotClearance(). */
@@ -90,8 +87,6 @@ Each of those gets us own unit test case in the below. */
 
 GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, ContextProvenance) {
   DummyCollisionChecker checker;
-  /* Has to match the 7 dofs in the robot. */
-  constexpr int kQSize = 7;
   const VectorXd q1 = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
   const VectorXd q2 = (VectorXd(kQSize) << 9, 8, 7, 6, 5, 4, 3).finished();
 
@@ -117,8 +112,6 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, ContextProvenance) {
 
 GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, NoDistanceMeasurements) {
   DummyCollisionChecker checker;
-  /* Has to match the 7 dofs in the robot. */
-  constexpr int kQSize = 7;
   const VectorXd q = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
 
   /* The checker returns an empty RobotClearance by default. This should lead to
@@ -131,8 +124,6 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, NoDistanceMeasurements) {
 
 GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
   DummyCollisionChecker checker;
-  /* Has to match the 7 dofs in the robot. */
-  constexpr int kQSize = 7;
   const VectorXd q0 = VectorXd::Zero(kQSize);
   const VectorXd q1 = (VectorXd(kQSize) << 1, 2, 3, 4, 5, 6, 7).finished();
   const VectorXd q2 = (VectorXd(kQSize) << 9, 8, 7, 6, 5, 4, 3).finished();
@@ -162,9 +153,9 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
     clearance.Append(robot_body, env_body, env_type, kMin - 0.1,
                      q1.transpose());
     checker.set_robot_clearance(move(clearance));
-    const VectorXd delta_q =
+    const VectorXd grad =
         ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
-    EXPECT_TRUE(CompareMatrices(delta_q, q1));
+    EXPECT_TRUE(CompareMatrices(grad, q1));
   }
 
   {
@@ -173,9 +164,9 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
     clearance.Append(robot_body, env_body, env_type, kMax + 0.1,
                      q1.transpose());
     checker.set_robot_clearance(move(clearance));
-    const VectorXd delta_q =
+    const VectorXd grad =
         ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
-    EXPECT_TRUE(CompareMatrices(delta_q, q0));
+    EXPECT_TRUE(CompareMatrices(grad, q0));
   }
 
   {
@@ -185,10 +176,10 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
     clearance.Append(robot_body, env_body, env_type, kMax - kRange * kWeight,
                      q1.transpose());
     checker.set_robot_clearance(move(clearance));
-    const VectorXd delta_q =
+    const VectorXd grad =
         ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
-    EXPECT_EQ(delta_q.size(), kQSize);
-    EXPECT_TRUE(CompareMatrices(delta_q, kWeight * q1));
+    EXPECT_EQ(grad.size(), kQSize);
+    EXPECT_TRUE(CompareMatrices(grad, kWeight * q1));
   }
 
   {
@@ -201,13 +192,34 @@ GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, WeightedCombinations) {
     clearance.Append(robot_body, env_body, env_type, kMax - kRange * kWeight2,
                      q2.transpose());
     checker.set_robot_clearance(move(clearance));
-    const VectorXd delta_q =
+    const VectorXd grad =
         ComputeCollisionAvoidanceDisplacement(checker, q0, kMin, kMax);
-    EXPECT_EQ(delta_q.size(), kQSize);
-    EXPECT_TRUE(CompareMatrices(delta_q, kWeight1 * q1 + kWeight2 * q2));
+    EXPECT_EQ(grad.size(), kQSize);
+    EXPECT_TRUE(CompareMatrices(grad, kWeight1 * q1 + kWeight2 * q2));
   }
 }
 
+GTEST_TEST(ComputeCollisionAvoidanceDisplacementTest, Errors) {
+  DummyCollisionChecker checker;
+  const VectorXd q = VectorXd::Zero(kQSize);
+
+  // The max_penetration cannot be positive.
+  EXPECT_THROW(
+      ComputeCollisionAvoidanceDisplacement(checker, q, 0.1, 1),
+      std::exception);
+
+  // The max_clearance cannot be negative.
+  EXPECT_THROW(
+      ComputeCollisionAvoidanceDisplacement(checker, q, -1, -0.1),
+      std::exception);
+
+  // They cannot be both zero.
+  EXPECT_THROW(
+      ComputeCollisionAvoidanceDisplacement(checker, q, 0, 0),
+      std::exception);
+}
+
 }  // namespace
+}  // namespace internal
 }  // namespace planning
-}  // namespace anzu
+}  // namespace drake


sherm1 (commented):
Platform review checkpoint. I'm having difficulty with the basics and would appreciate some clarification before I go any further, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM4yQs8D6H492CKV-cz:b-6qyxey)* status: 5 unresolved discussions, LGTM missing from assignee sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM4nKSl3PQL__lrwN53:b-6mesez) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*
> ```objc
>  configuration `q` into a "clearer" state, reducing the likelihood of collision.
> 
>  The displacement vector Δq = Σᵢ(wᵢ⋅Jqᵣ_ϕᵢ), where
> ```

This expression doesn't pass dimensional analysis. Assuming revolute joints, Δq should be measured in radians. w is dimensionless, so we need Jqᵣ_ϕᵢ to be in radians. Instead Jᵢ=∂ϕᵢ/∂qᵣ has units of length/radian. The equation would have to look something more like Δq = Σᵢ(wᵢ⋅ϕᵢ⋅Jᵢ⁻¹) to have the right dimensions, but since J isn't invertible that would be a pseudoinverse.

I'm not proposing an algorithmic change, but the documentation must be changed to make sense (or to explain why I'm wrong). (And radians being "unitless" is not the answer!)
___
*[`planning/collision_avoidance.h` line 18 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4lF_AEuc5AywMak6n:-NM4lF_AEuc5AywMak6o:b-1cteln) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L18)):*
> ```objc
>      geometrically in `checker`.
>    - ϕᵢ is the iᵗʰ measured distance and Jq_ϕᵢ is the partial derivative of ϕᵢ
>      with respect to qᵣ, the robot state.
> ```

typo: Jq_ϕᵢ -> Jqᵣ_ϕᵢ
___
*[`planning/collision_avoidance.h` line 20 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4qtdv7YiqaUAV0BS7:-NM4qtdv7YiqaUAV0BS8:b9x3852) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L20)):*
> ```objc
>      with respect to qᵣ, the robot state.
>    - wᵢ is the weight for iᵗʰ measurement, defined as wᵢ = (c - ϕᵢ) / (c - p),
>      where c = `max_clearance` and p = `max_penetration` and c > p.
> ```

BTW is there something missing from the weight definition to keep it in bounds?
___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM4uyLbEaoPE6uX3Gmf:bsqg7nx) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*
> ```objc
>     const CollisionChecker& checker, const Eigen::VectorXd& q,
>     double max_penetration, double max_clearance,
>     CollisionCheckerContext* context = nullptr);
> ```

BTW I prefer this signature but feel compelled to note that it is inconsistent with the approach taken in #18558 where there were two distinct function names depending on whether an internal or external Context is used. Calder provided a vigorous (though IMO unconvincing) defense of the other strategy during #18558 review around collision_checker.h:1010 in Reviewable. (Not sure how to link to it)

Why isn't the same style followed in both places?
___
*[`planning/collision_avoidance.cc` line 24 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4wPCB3zqQD5iVB16V:-NM4wPCB3zqQD5iVB16W:b-cuosxr) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L24)):*
> ```cpp
>       context == nullptr
>           ? checker.CalcRobotClearance(q, max_clearance)
>           : checker.CalcContextRobotClearance(context, q, max_clearance);
> ```

FYI this style is what I was referring to
___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM4xD9H7INPBBhC-NTG:bagd2lx) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*
> ```cpp
>   }
> 
>   return robot_clearance.jacobians().transpose() * weights;
> ```

Hmmm. So we really use Jᵀ here which has the right shape but still the wrong units. Even if we pretend  Jᵀ has units of radian/length like J⁻¹ would, there is still a missing length-unit factor (presumably the current gap) to get the final answer measured in radians.

Is there some justification for this? If so it should be documented.

jwnimmer-tri:
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM50lll34B8gcDLicbo:by4u6ia) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

This expression doesn't pass dimensional analysis. Assuming revolute joints, Δq should be measured in radians. w is dimensionless, so we need Jqᵣ_ϕᵢ to be in radians. Instead Jᵢ=∂ϕᵢ/∂qᵣ has units of length/radian. The equation would have to look something more like Δq = Σᵢ(wᵢ⋅ϕᵢ⋅Jᵢ⁻¹) to have the right dimensions, but since J isn't invertible that would be a pseudoinverse.

I'm not proposing an algorithmic change, but the documentation must be changed to make sense (or to explain why I'm wrong). (And radians being "unitless" is not the answer!)
</blockquote></details>

@SeanCurtis-TRI I will probably need your help here.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @SeanCurtis-TRI and @sherm1)

___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM4z7iA2d92DvCqmHBP:b-brx0rz) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*

BTW I prefer this signature but feel compelled to note that it is inconsistent with the approach taken in #18558 where there were two distinct function names depending on whether an internal or external Context is used. Calder provided a vigorous (though IMO unconvincing) defense of the other strategy during #18558 review around collision_checker.h:1010 in Reviewable. (Not sure how to link to it)

Why isn't the same style followed in both places?
</blockquote></details>

The member functions on CollisionChecker will be heavily used by a wide audience (hopefully most of Drake's audience), some of whom are relative novices.  This function will be much less widely used; it's intended for a very narrow audience of experts, in low-level control code where they are already intimately familiar with how the threads are managed.  (It's not actually used in Anzu as of today, though it was in the past.)

I've updated the class overview to start with "(Advanced)" to help steer novices away.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @SeanCurtis-TRI and @sherm1)

___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM52Nlq2goSdail4cpv:b9ebopv) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

Hmmm. So we really use Jᵀ here which has the right shape but still the wrong units. Even if we pretend  Jᵀ has units of radian/length like J⁻¹ would, there is still a missing length-unit factor (presumably the current gap) to get the final answer measured in radians.

Is there some justification for this? If so it should be documented.
</blockquote></details>

See `anzu/planning/collision_avoidance_benchmark.cc` for an example use.

It seems to me like in practice the expected use of Δq is to use only its infinitesimal "direction", not its magnitude.  The example code in the benchmark normalizes Δq to a configurable magnitude (with a norm of, e.g., 0.05 radians), applies that to the current position, and then loops over those steps repeatedly until q is collision-free.

jwnimmer-tri:
I'll ask +@calderpg-tri to feature-review the edits made during platform review.  Here, that will be everything after r2.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM58B9iAHnpLuxsclb6:badvh5s) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*

The member functions on CollisionChecker will be heavily used by a wide audience (hopefully most of Drake's audience), some of whom are relative novices.  This function will be much less widely used; it's intended for a very narrow audience of experts, in low-level control code where they are already intimately familiar with how the threads are managed.  (It's not actually used in Anzu as of today, though it was in the past.)

I've updated the class overview to start with "(Advanced)" to help steer novices away.
</blockquote></details>

On the other hand, for my part I also wouldn't necessarily mind adding the second function here with the `Context` in the function name just for consistency, even if the hazard for experts is thought to be low.  Happy to take input from @calderpg-tri and @SeanCurtis-TRI on that.  I don't think this facet got discussed in 9564 when this change happened.

SeanCurtis-TRI (commented):
One quick question:

The easy path out is to not port this code at this time. The only reason I think it was included in the port list is because it was originally a member function of `CollisionChecker` before it got separated. Do we have a plan for this particular function in Drake (i.e., does an imminent code port depend on this)?  Or can we simply close the PR and leave it in Anzu until more fully motivated?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM5LKfJ2o4RjJPvvJ3T:b-ri4a3t)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5GWue-Q0pFuGefD7l:b8efhg9) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

@SeanCurtis-TRI I will probably need your help here.
</blockquote></details>

For the record, I cannot resolve Sherm's concerns rigorously. But nevertheless, here's my best take. (Hand wavy all the way down.)

First, as a clarification, we don't assume heterogeneous units across qᵣ. So, ∂ϕᵢ/∂qᵣ is just a vector in some high dimensional space with no guarantee that the units of one element match the units of another. In order to do this operation, we push the whole problem into a high-dimensional space such that the units get discarded out during the projection process.

It just so happens that doing some kind of weighted gradient descent in that huge space has a nice effect in Euclidian space -- the robot configuration moves in a way to avoid collision (maximizing distance between measurements).

This is a horrible justification of a very common *heuristic* in motion planning. The heuristic ostensibly subsumes the units to make it work (buried in the weights).

(As a foot note: I think of it akin to how we map a single error tolerance parameter into error-checking integrators. There is an implicit unit reconciliation between that unitless parameter and the actual error tolerances applied to each dof.)
___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM5KZQO-gi9ROWECJBD:bmz1g6g) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*

On the other hand, for my part I also wouldn't necessarily mind adding the second function here with the `Context` in the function name just for consistency, even if the hazard for experts is thought to be low.  Happy to take input from @calderpg-tri and @SeanCurtis-TRI on that.  I don't think this facet got discussed in 9564 when this change happened.
</blockquote></details>

It was touched on. Everyone acknowledged it was different. But as with the other form of the APIs, the reviewers didn't feel sufficiently strongly to force authors to change. Certainly inconsistency is regrettable and possibly sufficiently regrettable to undo it. I'd rather *not* see an explosion of APIs (but that merely reveals my side on the religious war and doesn't actually provide an objective basis for deciding).

If I needed consistency with a minimum of code churn, I'd say we double up this API. If we needed consistency where cost were no obstacle, I'd change `CollisionChecker` (where the blank check would be necessary to cover a huge effort).

At the end of the day, I vote "present" and support whatever the consensus view is.

calderpg-tri (commented):
I expect to use it very soon as part of generating near-contact C-IRIS regions, for which being in Drake makes some things much easier.

A bit more context about the method generally: the role of `ComputeCollisionAvoidanceDisplacement` is primarily for use in fast project-out-of-collision operations, such as what would be used in making a near-contact or collision-free sampler for motion planning.

Compared with using the normal collision avoidance constraint + optimization to perform this type of projection, the displacement may be somewhat less capable (fewer colliding configurations projected to a collision-free configuration), but significantly faster in terms of runtime performance (~10x faster in many cases).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM5K2oy7irq8LGzKyCE:brrdp3h)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM5GIhI1VVAij8mAqp9:buv8udf) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*

It was touched on. Everyone acknowledged it was different. But as with the other form of the APIs, the reviewers didn't feel sufficiently strongly to force authors to change. Certainly inconsistency is regrettable and possibly sufficiently regrettable to undo it. I'd rather *not* see an explosion of APIs (but that merely reveals my side on the religious war and doesn't actually provide an objective basis for deciding).

If I needed consistency with a minimum of code churn, I'd say we double up this API. If we needed consistency where cost were no obstacle, I'd change `CollisionChecker` (where the blank check would be necessary to cover a huge effort).

At the end of the day, I vote "present" and support whatever the consensus view is.
</blockquote></details>

I'd be in favor of having a second function, although it is a little more annoying to write outside of `CollisionChecker` without access to mutable pool contexts, so it would likely end up looking like
```
namespace {
Eigen::VectorXd DoComputeCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    double max_penetration, double max_clearance,
    CollisionCheckerContext* context) {
  ...
}
}  // namespace

Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    const double max_penetration, const double max_clearance) {
  return DoComputeCollisionAvoidanceDisplacement(
      checker, q, max_penetration, max_clearance, nullptr);
}

Eigen::VectorXd ComputeContextCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    const double max_penetration, const double max_clearance,
    CollisionCheckerContext* context) {
  return DoComputeCollisionAvoidanceDisplacement(
      checker, q, max_penetration, max_clearance, context);
}
```
___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM5K0NI2XCvqU8k9ARL:b-eq2cmu) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

See `anzu/planning/collision_avoidance_benchmark.cc` for an example use.

It seems to me like in practice the expected use of Δq is to use only its infinitesimal "direction", not its magnitude.  The example code in the benchmark normalizes Δq to a configurable magnitude (with a norm of, e.g., 0.05 radians), applies that to the current position, and then loops over those steps repeatedly until q is collision-free.
</blockquote></details>

A bit more context is that Jᵀ is better behaved than J⁻¹ in terms of behavior around singularities, and in our experience with this method using Jᵀ works much better (cheaper per iteration, significantly fewer iterations needed).

jwnimmer-tri:
The CollisionChecker abstract unit test suite depends on this function (because it has tests for it).  See [slack discussion](https://app.slack.com/client/T0PTZL6RE/C02KXJ6LM54/thread/C02KXJ6LM54-1672961405.885089).  We could imagine some refactors that let us defer this work, but I'd like to see if we can resolve the reviews here quickly and keep moving.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5O7Q18whuLmDRP7R9:b-hbpn4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

For the record, I cannot resolve Sherm's concerns rigorously. But nevertheless, here's my best take. (Hand wavy all the way down.)

First, as a clarification, we don't assume heterogeneous units across qᵣ. So, ∂ϕᵢ/∂qᵣ is just a vector in some high dimensional space with no guarantee that the units of one element match the units of another. In order to do this operation, we push the whole problem into a high-dimensional space such that the units get discarded out during the projection process.

It just so happens that doing some kind of weighted gradient descent in that huge space has a nice effect in Euclidian space -- the robot configuration moves in a way to avoid collision (maximizing distance between measurements).

This is a horrible justification of a very common *heuristic* in motion planning. The heuristic ostensibly subsumes the units to make it work (buried in the weights).

(As a foot note: I think of it akin to how we map a single error tolerance parameter into error-checking integrators. There is an implicit unit reconciliation between that unitless parameter and the actual error tolerances applied to each dof.)
</blockquote></details>

Sounds like I should resolve this via a comment here along the lines of:

This math a useful heuristic for motion planning.  The dimensional analysis doesn't add insight, so readers should just ignore that it seems wrong.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @SeanCurtis-TRI)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5N-IqBJitGlT5v7aq:b-vq3cbi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Sounds like I should resolve this via a comment here along the lines of:

This math a useful heuristic for motion planning.  The dimensional analysis doesn't add insight, so readers should just ignore that it seems wrong.
</blockquote></details>

Thanks for the valiant attempt, Sean! IMO it doesn't clarify anything though. The mixed units are actually not a problem here because the individual derivatives all have the right units -- either ∂ϕ/∂θ for revolute entries or ∂ϕ/∂x for prismatic ones. So whatever hacks are being done here they aren't being done to fuzz over the units. And the weights are always unitless by construction -- they are length/length.

Jeremy, I don't think that's enough to allow this to land in Drake. If the algorithm's sole justification is "we've seen it work" then it should be documented like that. The pseudo-math justification seems out of place to me.

jwnimmer-tri:
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5PXwF5Mdmk3Qp-RkB:bf98uy2) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Thanks for the valiant attempt, Sean! IMO it doesn't clarify anything though. The mixed units are actually not a problem here because the individual derivatives all have the right units -- either ∂ϕ/∂θ for revolute entries or ∂ϕ/∂x for prismatic ones. So whatever hacks are being done here they aren't being done to fuzz over the units. And the weights are always unitless by construction -- they are length/length.

Jeremy, I don't think that's enough to allow this to land in Drake. If the algorithm's sole justification is "we've seen it work" then it should be documented like that. The pseudo-math justification seems out of place to me. 
</blockquote></details>

The _math_ is fine, and explains what this is computing.  Units are different than math.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @SeanCurtis-TRI)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5P6NNEya_qCODnVxM:b-d0c80h) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

The _math_ is fine, and explains what this is computing.  Units are different than math.
</blockquote></details>

My suggestion would be for the author to give some more thought to why this works if it does (better if there is a literature reference for it). Having it come out with the wrong units is just about the worst thing I can think of. If it really does work then perhaps the math behind it is not what is shown. For example, if there is actually a missing factor with length units, perhaps there is a good argument that a typical length is around 1 meter which is why it usually works. In that case it would be better to include the factor but note that we estimate it to have a magnitude of 1.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @SeanCurtis-TRI)

___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM5RPTd7wmpAjo0HBBZ:b-mlh7f6) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

A bit more context is that Jᵀ is better behaved than J⁻¹ in terms of behavior around singularities, and in our experience with this method using Jᵀ works much better (cheaper per iteration, significantly fewer iterations needed).
</blockquote></details>

Since J isn't generally invertible, are you saying you tried it with a pseudoinverse? Also is there a way you can use Jᵀ and make the units come out right? There should be some argument that explains how we're updating angles and lengths with δangles and δlengths.

calderpg-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM5S4qT69SqEKfL5fCg:b-vdet1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

Since J isn't generally invertible, are you saying you tried it with a pseudoinverse? Also is there a way you can use Jᵀ and make the units come out right? There should be some argument that explains how we're updating angles and lengths with δangles and δlengths.
</blockquote></details>

Yes, by J⁻¹ I mean pseudoinverse J⁺. Both J⁺ and Jᵀ were tried in the early days (and in predecessor formulations). There's a long tradition in control and planning of using Jᵀ over J⁺ for its stability in "get a direction + take a small step" operations like this.

jwnimmer-tri:
Hopefully by tomorrow sometime @calderpg-tri can offer some thoughts to help Sherm with math, to tune things up.

If not, then my fallback plan is to demote this function to `namespace internal`, add a TODO about the confusing documentation which captures the review threads, and bash on.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5jpok4eu1tNVsO7Ph:bu7cgxg) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

My suggestion would be for the author to give some more thought to why this works if it does (better if there is a literature reference for it). Having it come out with the wrong units is just about the worst thing I can think of. If it really does work then perhaps the math behind it is not what is shown. For example, if there is actually a missing factor with length units, perhaps there is a good argument that a typical length is around 1 meter which is why it usually works. In that case it would be better to include the factor but note that we estimate it to have a magnitude of 1.
</blockquote></details>

Ok, Sherm, try this on for size.

We have parameters (as documented):
  C ≥ 0 (farthest clearance)
  P ≤ 0 (deepest penetration)

And define the clearance function (named A for lack of letter creativity):

  A(q) = ϕ(q)·C / (C - P) - ϕ(q)²/[2(C - P)]

It is only defined for P ≤ ϕ(q) ≤ C (although, in the implementation we fudge this by clamping to that range; but we can limit analysis to this range).

A(q) has its maximum value at qc and its minimum value at qp (where qc and qp are defined such that ϕ(qc) = C and ϕ(qp) = P, respectively -- I've done some algebra to prove that A(qc) > A(qp) but I haven't confirmed that it is monotonically decreasing in the range, but I suspect it is). Therefore ∂A/∂q
will provide a vector in q-space toward increasing A (increasing clearance).

If we calculate ∂A/∂q, we get: Σwᵢ·∂ϕᵢ/∂q. where  wᵢ = [C - ϕᵢ(q)] / (C - P).

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 3 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @SeanCurtis-TRI)

___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM5jLPjD47H-ksjn3t2:bbiu47d) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

Yes, by J⁻¹ I mean pseudoinverse J⁺. Both J⁺ and Jᵀ were tried in the early days (and in predecessor formulations). There's a long tradition in control and planning of using Jᵀ over J⁺ for its stability in "get a direction + take a small step" operations like this.
</blockquote></details>

Can you provide a literature reference? It's not obvious to me how Jᵀ is even a direction. I can think of hand-wavy justification, like if J were orthogonal or nearly so then Jᵀ ≈  J⁻¹. I don't know of an equivalent for rectangular matrices but there probably is something.

Let me state my own naive impression so we can talk in the same terms:
I would expect that you are trying to solve an equation that starts out looking like J Δq = ϕ. That has perfectly fine units since J=∂ϕ/∂q. I see you have a method for weighted combination of individual Jacobians to produce a composite J; makes perfect sense. Now if you have an approximation of the inverse you could write
 Δq ≈ J⁻¹ϕ. (Even if we allow for  Jᵀ ≈  J⁻¹ the multiplication by ϕ is confusingly absent in the current code but is necessary for the units to make sense.)

Then since the original problem is nonlinear you probably wouldn't want to jump by the entire Δq so you'd have a "step size" reduction factor α < 1 somewhere. Maybe that's accounted for in the weights? I can't tell.

Anyway, what I would like to see is some explanation for how what appears at first to be nonsense is actually a well-thought-out approximation, or something like that.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5pJa-695-Re7QA3DF:b-b7ph8d) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Sorry, A(q) should be a summation of ϕ, but the analysis doesn't change.
</blockquote></details>

Makes sense, but I don't have a problem with ∂A/∂q being a direction, I just don't understand why it is on the RHS rather than on the LHS where it makes sense, i.e.
(∂A/∂q) Δq = ϕ. How do you get it over on the right without an inversion? And what happened to the gap function on the RHS?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri, @jwnimmer-tri, and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5qtS4DuVnXp9SLQk6:bneuu5c) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Makes sense, but I don't have a problem with ∂A/∂q being a direction, I just don't understand why it is on the RHS rather than on the LHS where it makes sense, i.e.
(∂A/∂q) Δq = ϕ. How do you get it over on the right without an inversion? And what happened to the gap function on the RHS?
</blockquote></details>

Are you referring to the code that does J.transpose() * w?

If so, that is just a mechanically simple way to do the summation. If one buys this A(q) function explanation, the transpose is not a pseudo inverse. It's simply using the linear algebra mechanism to sum up the weighted Js.

(And I think we're making progress; I've eliminated the unit concerns as A(q) is unitless.)

sherm1 (commented):
cc @ggould-tri: If you're willing, I suspect you have the right combination of mathematical rigor and robotics knowledge to help resolve the discussion below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM5r2Z_3hSpu8xxYMtE:bc037l6)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @jwnimmer-tri)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM5w6nn2O7B7JxdGbPU:b36585b) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Are you referring to the code that does J.transpose() * w?

If so, that is just a mechanically simple way to do the summation. If one buys this A(q) function explanation, the transpose is not a pseudo inverse. It's simply using the linear algebra mechanism to sum up the weighted Js.

(And I think we're making progress; I've eliminated the unit concerns as A(q) is unitless.)
</blockquote></details>

Wait, how is A(q) unitless??? ϕ has units of length, C/(C-P) is unitless, so A(q) has units of length.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM6IPX0CYyfMCU2uwbS:b-frtfpu) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Wait, how is A(q) unitless??? ϕ has units of length, C/(C-P) is unitless, so A(q) has units of length.
</blockquote></details>

I'll let Calder correct me, but here's how I understand it now:

What Sherm is saying is that we're calculating ∂A/∂q (as per Sean's exposition), but we've named our return value Δq -- it somehow has q the numerator now instead of the denominator, has different units.

The call site in Anzu immediately calls `stableNormalized` on the result of this function, thus turning it into a dimensionless direction vector.  (It assumes that all joints are of similar scale; you could imagine allowing for custom per-joint weights being desirable for certain kinds of robots.)  Then it multiples that vector by a constant (0.05 radians) to compute the _actual_ Δq that it applies to get out of collision.  (Again, you could imagine per-joint step size customization for other kinds of robots.)

So to my eye, that leaves us with a choice here:

(1) Move the norm + rescale code from the Anzu call site into this function, leaving the documentation as Δq.  (Possibly we would need to add a parameter for joint scaling, or a TODO.)

(2) Change the documentation to say we return ∂A/∂q and explain to callers how to get a Δq from it.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM6QyQh6okz7OlDgO8-:b-h5qz1e) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

I'll let Calder correct me, but here's how I understand it now:

What Sherm is saying is that we're calculating ∂A/∂q (as per Sean's exposition), but we've named our return value Δq -- it somehow has q the numerator now instead of the denominator, has different units.

The call site in Anzu immediately calls `stableNormalized` on the result of this function, thus turning it into a dimensionless direction vector.  (It assumes that all joints are of similar scale; you could imagine allowing for custom per-joint weights being desirable for certain kinds of robots.)  Then it multiples that vector by a constant (0.05 radians) to compute the _actual_ Δq that it applies to get out of collision.  (Again, you could imagine per-joint step size customization for other kinds of robots.)

So to my eye, that leaves us with a choice here:

(1) Move the norm + rescale code from the Anzu call site into this function, leaving the documentation as Δq.  (Possibly we would need to add a parameter for joint scaling, or a TODO.)

(2) Change the documentation to say we return ∂A/∂q and explain to callers how to get a Δq from it.
</blockquote></details>

Yep, thanks! Either of those would be fine.

SeanCurtis-TRI:
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM7Cxp80r9Tl_MoV99b:b-kprjvo) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Yep, thanks! Either of those would be fine.
</blockquote></details>

Correct -- I erred in claiming it was unitless.

SeanCurtis-TRI:
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM7DdVj-iD3XeZVtUWR:bilhdfd) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

Correct -- I erred in claiming it was unitless.
</blockquote></details>

Some thoughts on the listed options:

(1) Losing the original ∂A/∂q value (in place of a normalized, or normalized and scaled alternative) loses the information about what the magnitude actually is. It's conceivable that this quantity could be used (i.e., if the value were actually quite small, then normalizing it and multiplying it by a constant that is larger than the original vector's magnitude would have overshot implications near asymptotic regions). In other words, when taking a step, the magnitude of the derivative can be just as important as the direction.
(2) The single biggest challenge of *documenting* A(q) is providing *any* insight as to why it is formulated the way it is. From my perspective, my definition of A(q) as given above is merely a *post hoc* justification. The vector we're providing is a penalty-based combination of disparate distance constraints. It's a heuristic to reconcile a collection of constraints without any collision free guarantees. Maybe it would be better to leave the details of the math hidden and just indicate that there is *some* "clearance" function and we return its gradient which points in the direction of maximally increasing clearance. Then we can simply return that gradient and given the guidance Jeremy suggested.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM7Hqtm2RioMLnWmlVl:b7tj68p) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*
> It's conceivable that this quantity could be used ...

This is a small wrapper function around an already-public API.  My inclination is that here we should err on the side of making it most convenient to use and explain, i.e., something along the lines of my option (1).  That's the only motivating use case for this function.

Users who want to play with custom math on the intermediate values still have that option.

I'm also leaning towards marking it "namespace internal" for now, so that we can develop a more clear demo program that shows its efficacy before we advertise it further.

calderpg-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @jwnimmer-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM7Jy-e5pQpdz_9OboZ:bju2i4g) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

> It's conceivable that this quantity could be used ...

This is a small wrapper function around an already-public API.  My inclination is that here we should err on the side of making it most convenient to use and explain, i.e., something along the lines of my option (1).  That's the only motivating use case for this function.

Users who want to play with custom math on the intermediate values still have that option.

I'm also leaning towards marking it "namespace internal" for now, so that we can develop a more clear demo program that shows its efficacy before we advertise it further.
</blockquote></details>

I would go with (2), documenting it as the gradient of some clearance function (it was originally called "collision avoidance gradient" before it became "collision avoidance displacement").

I don't think wrapping it only in the way it's currently used is a great choice, since applications to robots with non-revolute joints would likely treat ∂A/∂q differently.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4nKSl3PQL__lrwN52:-NM7SK2w3Cx5YXZ75Ae_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L11)):*

I would go with (2), documenting it as the gradient of some clearance function (it was originally called "collision avoidance gradient" before it became "collision avoidance displacement").

I don't think wrapping it only in the way it's currently used is a great choice, since applications to robots with non-revolute joints would likely treat ∂A/∂q differently.
</blockquote></details>

Done.
___
*[`planning/collision_avoidance.h` line 57 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4uyLbEaoPE6uX3Gme:-NM7U3ue7PfhtFuoiz3V:b-j3fr4k) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.h#L57)):*

I'd be in favor of having a second function, although it is a little more annoying to write outside of `CollisionChecker` without access to mutable pool contexts, so it would likely end up looking like
```
namespace {
Eigen::VectorXd DoComputeCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    double max_penetration, double max_clearance,
    CollisionCheckerContext* context) {
  ...
}
}  // namespace

Eigen::VectorXd ComputeCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    const double max_penetration, const double max_clearance) {
  return DoComputeCollisionAvoidanceDisplacement(
      checker, q, max_penetration, max_clearance, nullptr);
}

Eigen::VectorXd ComputeContextCollisionAvoidanceDisplacement(
    const CollisionChecker& checker, const Eigen::VectorXd& q,
    const double max_penetration, const double max_clearance,
    CollisionCheckerContext* context) {
  return DoComputeCollisionAvoidanceDisplacement(
      checker, q, max_penetration, max_clearance, context);
}
```
</blockquote></details>

Done (via TODO).
___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM7SOsLDbo-Qc-8efT3:b1f5a4k) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

Can you provide a literature reference? It's not obvious to me how Jᵀ is even a direction. I can think of hand-wavy justification, like if J were orthogonal or nearly so then Jᵀ ≈  J⁻¹. I don't know of an equivalent for rectangular matrices but there probably is something.

Let me state my own naive impression so we can talk in the same terms:
I would expect that you are trying to solve an equation that starts out looking like J Δq = ϕ. That has perfectly fine units since J=∂ϕ/∂q. I see you have a method for weighted combination of individual Jacobians to produce a composite J; makes perfect sense. Now if you have an approximation of the inverse you could write
 Δq ≈ J⁻¹ϕ. (Even if we allow for  Jᵀ ≈  J⁻¹ the multiplication by ϕ is confusingly absent in the current code but is necessary for the units to make sense.)

Then since the original problem is nonlinear you probably wouldn't want to jump by the entire Δq so you'd have a "step size" reduction factor α < 1 somewhere. Maybe that's accounted for in the weights? I can't tell.

Anyway, what I would like to see is some explanation for how what appears at first to be nonsense is actually a well-thought-out approximation, or something like that.
</blockquote></details>

I think this discussion is done now?  We're computing with ∂𝒜/∂q in batch, not doing anything with inverses.

sherm1 (commented):
Great, all the badness is gone! NVM Grant.  The changes revealed one more discrepancy, PTAL.

> Reviewed 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM7ZwlOEoGPhXIBkgwB:bdn6m8s)* status: 1 unresolved discussion, LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri and @jwnimmer-tri)

___
*[`planning/collision_avoidance.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM4xD9H7INPBBhC-NTF:-NM7ZlaU6PqRSmbR-Hgl:b83q0x8) ([raw file](https://github.com/RobotLocomotion/drake/blob/b3f9644c42b439778f846a27a444d638595018e4/planning/collision_avoidance.cc#L41)):*

I think this discussion is done now?  We're computing with ∂𝒜/∂q in batch, not doing anything with inverses.
</blockquote></details>

Yes, lovely!
___
*[`planning/collision_avoidance.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM7Y23t5v25KqqofwIC:-NM7Y23t5v25KqqofwID:bxvrkkw) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.cc#L42)):*
> ```cpp
>   }
> 
>   // ∂𝐀/∂q = Σᵢ(wᵢ⋅Jqᵣ_ϕᵢ) where wᵢ is the weight for iᵗʰ measurement.
> ```

nit: the definition of Jqᵣ_ϕᵢ (=∂ϕᵢ/∂qᵣ) is gone from the intro comments now so should be mentioned here.

minor: it's not clear why there is a transpose() in the code below. If the robot_clearance Jacobian really is as documented above, then we wouldn't want to transpose it to return ∂𝐀/∂q. Or does this function actually return (∂𝐀/∂q)ᵀ ? There should be a comment here explaining why the transpose is necessary since  on the surface it seems that the code and documentation disagree.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 1 unresolved discussion, LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM7Y23t5v25KqqofwIC:-NM99pwJ5lJKDxWLrYfh:by2e9pk) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.cc#L42)):*

nit: the definition of Jqᵣ_ϕᵢ (=∂ϕᵢ/∂qᵣ) is gone from the intro comments now so should be mentioned here.

minor: it's not clear why there is a transpose() in the code below. If the robot_clearance Jacobian really is as documented above, then we wouldn't want to transpose it to return ∂𝐀/∂q. Or does this function actually return (∂𝐀/∂q)ᵀ ? There should be a comment here explaining why the transpose is necessary since  on the surface it seems that the code and documentation disagree.
</blockquote></details>

It seems that the primary motivation is to return ∂𝐀/∂q as a column vector - as vectors in Drake are generally treated as such.

So, technically, to achieve this, we could write it fully explicitly:

```c++
return (weights.transpose() * robot_clearance.jacobians()).transpose();
```

or more compactly (relying on Eigen to convert row to column on return):

```c++
return weights.transpose() * robot_clearance.jacobians();
```

All three approaches (including the current approach) accomplish the same thing. The current approach would benefit from a comment. The most verbose would be less mathematically deceptive and require the least amount of documentation (possibly none).

I know when I saw the transpose originally, I had to reason about it before I recognized it as simply the weighted sum of the various ∂ϕᵢ/∂qᵣ. So, I'd be in favor of some clarification.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: LGTM missing from assignees calderpg-tri,sherm1(platform) (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM7Y23t5v25KqqofwIC:-NM9tU-t5Y4GtITR2f1L:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.cc#L42)):*

I agree that some clarifying documentation is the likely cure here.

BTW a matrix of partial derivatives is normally laid out in rows. You can see that from the fact that we can write `∂ϕ/∂q Δq = ε`. Since Δq is a column vector, the matrix on the left must have rows of the same length.

A perversion of that convention (as I understand it) is that a _gradient_ (partial derivatives of a scalar function) is a column! That is, ∇f = (∂f/∂x)ᵀ for scalar function f(x).

So IMO if we claim to return a matrix of partial derivatives we shouldn't return its transpose. But fine if we document that we're returning the transpose instead.
</blockquote></details>

Done.

sherm1 (commented):
Platform :lgtm: pending some doc & naming adjustments

> Reviewed 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM9yCSACz9nGR75B7xE:bwr2o07)* status: 4 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 9 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9vV1fEo_yskIzFfI7:-NM9vV1fEo_yskIzFfI8:b4fq7yd) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L9)):*
nit: because 𝐀(q) is not a scalar, this is a Jacobian rather than a gradient. Wikipedia has a nice picture showing the relationship [here](https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant).

typo: unspecifed -> unspecified

Since we are apparently returning the transpose of ∂𝐀/∂q we should say that, e.g.:
> Computes the Jacobian J=∂𝐀/∂q of an unspecified distance function 𝐀 and returns Jᵀ.

(If you don't like calling this a Jacobian you could just say "matrix of partial derivatives")
___
*[`planning/collision_avoidance.h` line 40 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9w_jb9ajR1Niwutlp:-NM9w_jb9ajR1Niwutlq:blflqp7) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L40)):*
> ```objc
>                             provided, the checker's context for the current
>                             thread is used.
>  @retval ∂𝐀/∂q the gradient indicating increased robot clearance.
> ```

minor: (∂𝐀/∂q)ᵀ transpose of Jacobian indicating increased robot clearance
___
*[`planning/test/collision_avoidance_test.cc` line 159 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9x60qBlF_iSCzUH4s:-NM9x60qBlF_iSCzUH4t:bbk62zc) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/test/collision_avoidance_test.cc#L159)):*
> ```cpp
>                      q1.transpose());
>     checker.set_robot_clearance(move(clearance));
>     const VectorXd delta_q =
> ```

nit: this name is now inconsistent with the documentation -- I believe this is now the transpose of the Jacobian and since there is only a scalar distance here it is the gradient. So you could call this `grad_A` now.
___
*[`planning/test/collision_avoidance_test.cc` line 204 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9wwnPEkWkObGwj9lt:-NM9wwnPEkWkObGwj9lu:barjcg8) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/test/collision_avoidance_test.cc#L204)):*
> ```cpp
>   }
> }
> ```

minor: should have some quick "throw" tests of bad parameter values to the API

sherm1 (commented):
Also needs another lgtm from a feature reviewer

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NM9yRtJ7rHdyMBw0AZH:b-i1pous)* status: 4 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @jwnimmer-tri)

calderpg-tri (commented):
> Reviewed 1 of 3 files at r2, 1 of 3 files at r4, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 4 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @jwnimmer-tri)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 4 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @jwnimmer-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 9 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9vV1fEo_yskIzFfI7:-NMA182s0YZl_ENbSCWy:b-z8l9pp) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L9)):*

nit: because 𝐀(q) is not a scalar, this is a Jacobian rather than a gradient. Wikipedia has a nice picture showing the relationship [here](https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant).

typo: unspecifed -> unspecified

Since we are apparently returning the transpose of ∂𝐀/∂q we should say that, e.g.:
> Computes the Jacobian J=∂𝐀/∂q of an unspecified distance function 𝐀 and returns Jᵀ.

(If you don't like calling this a Jacobian you could just say "matrix of partial derivatives")
</blockquote></details>

A(q) *is* a scalar. phi is a scalar, and we're taking the set of phi_i and summing them together. So, it's scalar.

jwnimmer-tri (commented):
(Calder is assigned for feature review.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NMA0tq1DNt-CYdSf-uD:byi43be)* status: 4 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @SeanCurtis-TRI and @sherm1)

___
*[`planning/collision_avoidance.h` line 9 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9vV1fEo_yskIzFfI7:-NMA0xlU2eGX9ENQISLr:bv5c4bb) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L9)):*

A(q) *is* a scalar. phi is a scalar, and we're taking the set of phi_i and summing them together. So, it's scalar.
</blockquote></details>

Right, it is a scalar.  (Maybe I shouldn't have used a bold letter for it here?)
___
*[`planning/collision_avoidance.h` line 40 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9w_jb9ajR1Niwutlp:-NMA1AEc2wpp1V3caYLz:b-ysyu0s) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L40)):*

minor: (∂𝐀/∂q)ᵀ transpose of Jacobian indicating increased robot clearance
</blockquote></details>

The `@retval` is no longer here; maybe this review discussion is vestigial?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions, LGTM missing from assignee calderpg-tri (waiting on @calderpg-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`planning/test/collision_avoidance_test.cc` line 159 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9x60qBlF_iSCzUH4s:-NMA4ivp7db5hUx8RRg9:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/test/collision_avoidance_test.cc#L159)):*

nit: this name is now inconsistent with the documentation -- I believe this is now the transpose of the Jacobian and since there is only a scalar distance here it is the gradient. So you could call this `grad_A` now.
</blockquote></details>

Done.
___
*[`planning/test/collision_avoidance_test.cc` line 204 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9wwnPEkWkObGwj9lt:-NMA4kNU0jw_vMlzjWB6:b-90ioj1) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/test/collision_avoidance_test.cc#L204)):*

minor: should have some quick "throw" tests of bad parameter values to the API
</blockquote></details>

Done (for some plausible error cases).  Since this function will only be used for drake-internal unit tests and benchmarking, we can defer any extra polishing for later when it becomes promoted to public.

calderpg-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-:-NMA5gDoB770W9CYJc5a:bnfp4nl)* status: 2 unresolved discussions (waiting on @SeanCurtis-TRI and @sherm1)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: :shipit: complete! all discussions resolved, LGTM from assignees calderpg-tri,sherm1(platform), base commit is latest master (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: :shipit: complete! all discussions resolved, LGTM from assignees calderpg-tri,sherm1(platform), base commit is latest master (waiting on @calderpg-tri and @sherm1)

___
*[`planning/collision_avoidance.h` line 9 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NM9vV1fEo_yskIzFfI7:-NMANuQlDhrmH7IPU2tR:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0ea87a454a3da515e6631264a68b6fe3abd484a0/planning/collision_avoidance.h#L9)):*

Oh, I missed that the rectangular Jacobian gets is rows weight-summed together to end up with a gradient. Apologies! Consider using a lower-case letter for the function name to suggest a scalar (upper case A suggests a matrix).
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 2 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`planning/collision_avoidance.h` line 10 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NMAOi0b1rqt_64Gst7Y:-NMAOi0cAnADe7QXvOsp:ba8sa95) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9dbfec4fbd8b163771e737a6f725b0a29a99cd5/planning/collision_avoidance.h#L10)):*
BTW This could be both more precise (in some ways) and more honest (in others): 

The definition given in the other conversation (corrected here) `f(q) = Σᵢ[ϕᵢ(q)·C / (C - P) - ϕᵢ(q)²/[2(C - P)]]` is not obviously a weighted combination of anything (it's certainly more complex than the phrase implies). Certainly, the gradient of f is well implemented as a weighted combination of the gradients.

Perhaps it would be better to say:

> The value of "f" is a function of the distances provided in CollisionChecker's RobotClerance data quantifying the amount of robot clearance (larger values implies greater clearance), and so its gradient...
___
*[`planning/collision_avoidance.h` line 14 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NMAQHaHEoA5Snwk3_da:-NMAQHaHEoA5Snwk3_db:bk3vu95) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9dbfec4fbd8b163771e737a6f725b0a29a99cd5/planning/collision_avoidance.h#L14)):*
> ```objc
>  reducing the likelihood of collision.
> 
>  The gradient reconciles multiple simultaneous potential collisions via a
> ```

`f` defines the reconciliation, not the gradient.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`planning/collision_avoidance.h` line 14 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18608#-NMAQHaHEoA5Snwk3_da:-NMASBI_CE9gTKYqlLXc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9dbfec4fbd8b163771e737a6f725b0a29a99cd5/planning/collision_avoidance.h#L14)):*

`f` defines the reconciliation, not the gradient.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18608)* status: :shipit: complete! all discussions resolved, LGTM from assignees calderpg-tri,sherm1(platform), base commit is latest master (waiting on @jwnimmer-tri)

Issue #17231, opened 2022-05-19
Use hidden visibility when compiling externals


jwnimmer-tri:
When building private externals from source (i.e., anything other than #7451), we should use hidden symbol visibility, so that downstream users of Drake can use their own, different version of that code without any link-time or load-time conflicts.

For C code, we can use `-fvisibility=hidden`.  For C++ code, can use `//tools/workspace:vendor_cxx` if it compiles.

Checklist:
- [x] #16774
- [x] #18678 (ccd)
- [x] #19928
- [x] #19893
- [x] conex
   - [x] #19883
   - [x] #19892
- [x] #19894
- [x] #17755
- [x] #18678 (fcl)
- [x] #17834
- [x] #20245
- [x] #18559
- [x] Ipopt
   - [x] #20429
   - [x] #22331
- [x] Mumps
   - [x] To be dropped as a dependency (rather than vendored).
- [x] #17288
- [x] #18690 (osqp)
- [x] #19890
  - [x] #19882
- [x] #19879
- [x] #18690 (qdldl)
- [x] #17229 
- [x] #18690 (scs)
- [x] #17230
  - [x] ignition_math
  - [x] ignition_utils
- [x] #18610
- [x] #19880
- [x] #17830
- [x] #18690 (suitesparse)
- [x] #19889 (usockets)
- [ ] uwebsockets
   - [x] #19889
   - [ ] Switch to hidden namespace
- [x] #20387
- [x] #16502
- [x] #16756
- [x] #20276
- [x] #20331
- [x] #20343

Tricky, possibly unsolveable:
- [x] lcm

The overall plan for all externals lives in [this spreadsheet](https://docs.google.com/spreadsheets/d/1WNeCU4errZk5UpyxefowBmFwhI5b27UfAiqG2mKFuaQ/edit).

Closing the loop:
- [x] #20279
- [ ] Ongoing for the list of "known bad symbols" in that new linter.
   - ...

Additional tasks:
- [ ] Documentation in `tools/workspace` about the current policy, how to add a new external, etc.
- [ ] Check if `vendor_cxx.bzl` can offer any more sugar macro(s) to reduce boilerplate.

jwnimmer-tri:
We are getting really close here!

PR #18610, opened 2023-01-18
[workspace] Deprecate the stduuid external

jwnimmer-tri:
Fork it to `@stduuid_internal` instead.
Fix the new copy to actually install its license file.
Adjust the new copy to use a numbered version now that upstream uses tags.
Adjust the new copy to be vendored and linker-hidden.
Rework `gsl` build rule and note that `gsl` is soon-to-be unused.

Towards #17231.

Note that Focal users will still be at ODR-risk for `gsl::span`, since we don't vendor that here.  It doesn't seem worth the effort since Focal support will eventually evaporate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18610)

### DIFF
--- a/geometry/BUILD.bazel
+++ b/geometry/BUILD.bazel
@@ -471,7 +471,7 @@ drake_cc_library(
         "//common:unused",
         "@common_robotics_utilities",
         "@msgpack_internal//:msgpack",
-        "@stduuid",
+        "@stduuid_internal//:stduuid",
         "@uwebsockets",
     ],
 )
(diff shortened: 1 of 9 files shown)


jwnimmer-tri:
+@rpoyner-tri for feature review, please.  (Or both reviews, if you like.)

Note that some of the code in `workspace/stduuid_internal` here is copied from the original `workspace/stduuid`, though it's been altered just enough that reviewing as-new without worrying about the copy diffs is probably just as well.

rpoyner-tri (commented):
:lgtm:
+(status: single reviewer ok) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18610#-:-NM5ONY49zEQF45ttXmT:b39zuqs)* status: :shipit: complete! all discussions resolved, LGTM from assignee rpoyner-tri(platform) (waiting on @jwnimmer-tri)

PR #18622, opened 2023-01-19
[C-IRIS] Do not use all threads in cspace_free_polytope_test

hongkai-dai:
From slack thread https://drakedevelopers.slack.com/archives/C270MN28G/p1674142473151599

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18622)

### DIFF
--- a/geometry/optimization/dev/BUILD.bazel
+++ b/geometry/optimization/dev/BUILD.bazel
@@ -77,6 +77,9 @@ drake_cc_googletest(
 drake_cc_googletest(
     name = "cspace_free_polytope_test",
     timeout = "long",
+    # This test launches 2 threads to test both serial and parallel code paths
+    # in CspaceFreePolytope.
+    tags = ["cpu:2"],
     deps = [
         ":c_iris_test_utilities",
         ":cspace_free_polytope",
--- a/geometry/optimization/dev/test/cspace_free_polytope_test.cc
+++ b/geometry/optimization/dev/test/cspace_free_polytope_test.cc
@@ -21,6 +21,7 @@ namespace drake {
 namespace geometry {
 namespace optimization {
 const double kInf = std::numeric_limits<double>::infinity();
+const int kTestConcurrency = 2;
 
 // This is a friend class of CspaceFreePolytope, we use it to expose the private
 // functions in CspaceFreePolytope for unit testing.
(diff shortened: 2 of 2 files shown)


hongkai-dai:
@drake-jenkins-bot linux-focal-clang-bazel-experimental-everything-valgrind-memcheck please.

hongkai-dai (commented):
When we do memcheck, it does memory check into Mosek solver, this significantly increases the computation time. I tested some of the optimization problem in cspace_free_polytope_test.cc, that running the program in the debug mode, without memcheck the Mosek solver time is about 1 second, with memcheck it takes about 1500 seconds. I am surprised that memcheck would slow down the computation by this much.

@jwnimmer-tri I don't think it makes sense to run memcheck into Mosek optimizer. Is it possible to run memcheck only on Drake code, but not into Mosek's code?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18622#-:-NMC81LZ7UjfdSYvbpJD:bajo81b)* status: needs at least one assigned reviewer, missing label for release notes

jwnimmer-tri:
> Is it possible to run memcheck only on Drake code, but not into Mosek's code?

I don't know.  If Mosek and Drake pass memory back and forth to each other, it's probably not possible.

Typically, we never run memcheck on tests that use Mosek.  Adding `mosek_test_tags()` to the build rule for the test adds the tag "mosek" ...

https://github.com/RobotLocomotion/drake/blob/8a0f43bb5d652f4b76eda9e167812bf875c2d98b/tools/skylark/test_tags.bzl#L42-L43

... and our filter for Memcheck builds disables those tests ...

https://github.com/RobotLocomotion/drake/blob/8a0f43bb5d652f4b76eda9e167812bf875c2d98b/tools/dynamic_analysis/bazel.rc#L230

This test looks like maybe it has some Mosek-required code and some okay-without-Mosek code.  It would be better if you could split those test cases into different test programs, tag the mosek half as `mosek_test_tags()`, and remove the `solver.available()` guards.

hongkai-dai (commented):
Thanks Jeremy, I will file a PR to split the test cases into two files, after PR #18627 is merged (which touches the same file).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18622#-:-NMCIdDQAUWuUZLVWKyW:btxajvf)* status: needs at least one assigned reviewer, missing label for release notes

Issue #18059, opened 2022-10-10
Parser: "singular" methods are obsolete and should go


rpoyner-tri:
The "singular" methods are deeply awkward to implement for multi-model capable formats, and perhaps have outlived their usefulness with the ability to scope-name models rather than renaming them. Consider deprecation and removal.

See discussions in #17623.

rpoyner-tri:
We should do #18060 before deprecating the singular methods.

rpoyner-tri:
Partially addressed in #18110.

rpoyner-tri:
Draft/prototype is at #18157. That needs to become a PR train and get some further design review. #18266 tries to dispose of some of the boring parts. More to come.

rpoyner-tri:
More boring parts swept into #18277.

rpoyner-tri:
Adding a concern here, first discovered in #18878. The dmd.yaml format doesn't currently have a multi-model aware file loading statement. If we hope to wipe out the singular methods entirely, dmd.yaml will need some changes (add_models statement?, options for prefixes, auto-renaming, etc.). Also, places in other code where dmd.yaml has been used to retain forced model renaming would have to be reviewed and possibly rewritten.

jwnimmer-tri:
My hypothesis is that we could have the `add_model:` yaml syntax call `AddModels` plural.  I think all API promises would still hold, and it would not generate any implementation complexity.  The `default_joint_positions` could either apply to only the first model, or we could search all of the added models.

rpoyner-tri:
At the very least, the `name` field would have to be deprecated. The entire point of this odyssey was to move away from whack-a-mole model renaming.

jwnimmer-tri:
Yes, we should look at deprecating `name`, but it's also simple to make an improvement here without doing that, if we decided it was a good trade-off:
- if `name` is not given (or empty), use the plural parser;
- if `name` is given, use the singular parser.

Maybe that is too confusing, and we need to create `struct AddModels` (with an "s") for the dmd schema?  I imagined that would be a lot of pasta though, for little benefit.

Making the name optional seemed to me like it might be the smoothest ramp towards removing the renaming entirely.  Possibly this deserves some prototyping versus Anzu, to see what ends up working best.

EricCousineau-TRI:
[...]

PR #18625, opened 2023-01-19
[parsing] Add model name prefix constructors to Parser

rpoyner-tri:
This patch introduces a new means to disambiguate parsed copies of the same model: Parser constructors that take model_name_prefix. The new mechanism will supplant the old scheme of renaming a single model during parsing.

This patch only introduces the new API elements; later patches will port call sites and deprecate the APIs that implemented parse-time renaming.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)

### DIFF
--- a/bindings/pydrake/multibody/parsing_py.cc
+++ b/bindings/pydrake/multibody/parsing_py.cc
@@ -70,9 +70,14 @@ PYBIND11_MODULE(parsing, m) {
     constexpr auto& cls_doc = doc.Parser;
     auto cls = py::class_<Class>(m, "Parser", cls_doc.doc);
     cls  // BR
-        .def(py::init<MultibodyPlant<double>*, SceneGraph<double>*>(),
+        .def(py::init<MultibodyPlant<double>*, SceneGraph<double>*,
+                 std::string_view>(),
             py::arg("plant"), py::arg("scene_graph") = nullptr,
-            cls_doc.ctor.doc)
+            py::arg("model_name_prefix") = "",
+            cls_doc.ctor.doc_3args_plant_scene_graph_model_name_prefix)
+        .def(py::init<MultibodyPlant<double>*, std::string_view>(),
+            py::arg("plant"), py::arg("model_name_prefix"),
+            cls_doc.ctor.doc_2args_plant_model_name_prefix)
         .def("plant", &Class::plant, py_rvp::reference_internal,
             cls_doc.plant.doc)
         .def("package_map", &Class::package_map, py_rvp::reference_internal,
--- a/bindings/pydrake/multibody/test/parsing_test.py
+++ b/bindings/pydrake/multibody/test/parsing_test.py
@@ -26,6 +26,7 @@ import unittest
 
 from pydrake.common import FindResourceOrThrow
 from pydrake.common.test_utilities.deprecation import catch_drake_warnings
+from pydrake.geometry import SceneGraph
 from pydrake.multibody.tree import (
     ModelInstanceIndex,
 )
@@ -119,6 +120,26 @@ class TestParsing(unittest.TestCase):
             file_contents=sdf_contents, file_type="sdf")
         self.assertIsInstance(results[0], ModelInstanceIndex)
 
+    def test_parser_prefix_constructors(self):
+        model = "<robot name='r'><link name='a'/></robot>"
+        plant = MultibodyPlant(time_step=0.0)
+        scene_graph = SceneGraph()
+
+        Parser(plant=plant).AddModelsFromString(model, "urdf")
+
+        # Reload the same model, via a different parser constructor. Catch the
+        # name collision.
+        with self.assertRaisesRegex(RuntimeError, r'.*names must be unique.*'):
+            Parser(plant=plant, scene_graph=scene_graph).AddModelsFromString(
+                model, "urdf")
+
+        # Reload the same model, but use model_name_prefix to avoid name
+        # collisions.
+        Parser(plant=plant, model_name_prefix="prefix1").AddModelsFromString(
+            model, "urdf")
+        Parser(plant=plant, scene_graph=scene_graph,
+               model_name_prefix="prefix2").AddModelsFromString(model, "urdf")
+
     def test_strict(self):
         model = """<robot name='robot' version='0.99'>
             <link name='a'/>
@@ -133,13 +154,8 @@ class TestParsing(unittest.TestCase):
         plant = MultibodyPlant(time_step=0.01)
         parser = Parser(plant=plant)
         parser.SetStrictParsing()
-        with self.assertRaises(RuntimeError) as e:
-            result = parser.AddModelsFromString(
-                file_contents=model, file_type='urdf')
-        pattern = r'.*version.*ignored.*'
-        message = str(e.exception)
-        match = re.match(pattern, message)
-        self.assertTrue(match, f'"{message}" does not match "{pattern}"')
+        with self.assertRaisesRegex(RuntimeError, r'.*version.*ignored.*'):
+            parser.AddModelsFromString(file_contents=model, file_type='urdf')
 
     def test_model_instance_info(self):
         """Checks that ModelInstanceInfo bindings exist."""
--- a/multibody/parsing/parser.cc
+++ b/multibody/parsing/parser.cc
@@ -20,11 +20,23 @@ using internal::ParserInterface;
 using internal::ParsingWorkspace;
 using internal::SelectParser;
 
-Parser::Parser(
-    MultibodyPlant<double>* plant,
-    geometry::SceneGraph<double>* scene_graph)
+Parser::Parser(MultibodyPlant<double>* plant,
+               geometry::SceneGraph<double>* scene_graph)
+    : Parser(plant, scene_graph, {}) {}
+
+Parser::Parser(MultibodyPlant<double>* plant,
+               std::string_view model_name_prefix)
+    : Parser(plant, nullptr, model_name_prefix) {}
+
+Parser::Parser(MultibodyPlant<double>* plant,
+               geometry::SceneGraph<double>* scene_graph,
+               std::string_view model_name_prefix)
     : plant_(plant) {
   DRAKE_THROW_UNLESS(plant != nullptr);
+
+  if (!model_name_prefix.empty()) {
+    model_name_prefix_ = std::string(model_name_prefix);
+  }
 
   if (scene_graph != nullptr && !plant->geometry_source_is_registered()) {
     plant->RegisterAsSourceForSceneGraph(scene_graph);
@@ -47,7 +59,8 @@ std::vector<ModelInstanceIndex> Parser::AddModels(
   DataSource data_source(DataSource::kFilename, &filename_string);
   ParserInterface& parser = SelectParser(diagnostic_policy_, file_name);
   auto composite = internal::CompositeParse::MakeCompositeParse(this);
-  return parser.AddAllModels(data_source, {}, composite->workspace());
+  return parser.AddAllModels(data_source, model_name_prefix_,
+                             composite->workspace());
 }
 
 std::vector<ModelInstanceIndex> Parser::AddAllModelsFromFile(
@@ -61,7 +74,8 @@ std::vector<ModelInstanceIndex> Parser::AddModelsFromString(
   const std::string pseudo_name(data_source.GetStem() + "." + file_type);
   ParserInterface& parser = SelectParser(diagnostic_policy_, pseudo_name);
   auto composite = internal::CompositeParse::MakeCompositeParse(this);
-  return parser.AddAllModels(data_source, {}, composite->workspace());
+  return parser.AddAllModels(data_source, model_name_prefix_,
+                             composite->workspace());
 }
 
 ModelInstanceIndex Parser::AddModelFromFile(
@@ -71,8 +85,8 @@ ModelInstanceIndex Parser::AddModelFromFile(
   ParserInterface& parser = SelectParser(diagnostic_policy_, file_name);
   auto composite = internal::CompositeParse::MakeCompositeParse(this);
   std::optional<ModelInstanceIndex> maybe_model;
-  maybe_model = parser.AddModel(data_source, model_name, {},
-                                 composite->workspace());
+  maybe_model = parser.AddModel(data_source, model_name, model_name_prefix_,
+                                composite->workspace());
   if (!maybe_model.has_value()) {
     throw std::runtime_error(
         fmt::format("{}: parsing failed", file_name));
@@ -89,8 +103,8 @@ ModelInstanceIndex Parser::AddModelFromString(
   ParserInterface& parser = SelectParser(diagnostic_policy_, pseudo_name);
   auto composite = internal::CompositeParse::MakeCompositeParse(this);
   std::optional<ModelInstanceIndex> maybe_model;
-  maybe_model = parser.AddModel(data_source, model_name, {},
-                                 composite->workspace());
+  maybe_model = parser.AddModel(data_source, model_name, model_name_prefix_,
+                                composite->workspace());
   if (!maybe_model.has_value()) {
     throw std::runtime_error(
         fmt::format("{}: parsing failed", pseudo_name));
--- a/multibody/parsing/parser.h
+++ b/multibody/parsing/parser.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <filesystem>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -55,6 +56,30 @@ class CompositeParse;
 /// absence of units specified by the format itself. This includes the literals
 /// in the explicitly specified files as well as referenced files such as OBJ
 /// or other data file formats.
+///
+/// MultibodyPlant requires that model instances have unique names. To support
+/// loading multiple instances of the same model file(s) into a plant, Parser
+/// offers constructors that take a model name prefix, which gets applied to
+/// all models loaded with that Parser instance. The resulting workflow makes
+/// multiple parsers to build models for a single plant:
+/// @code
+///  Parser left_parser(plant, "left");
+///  Parser right_parser(plant, "right");
+///  left_parser.AddModels(arm_model);  // "left::arm"
+///  right_parser.AddModels(arm_model);  // "right::arm"
+///  left_parser.AddModels(gripper_model);  // "left::gripper"
+///  right_parser.AddModels(gripper_model);  // "right::gripper"
+/// @endcode
+///
+/// (Advanced) In the rare case where the user is parsing into a MultibodyPlant
+/// and SceneGraph but has created them one at a time instead of using the more
+/// convenient AddMultibodyPlant() or AddMultibodyPlantSceneGraph() functions,
+/// the Parser constructors accept an optional SceneGraph pointer to specify
+/// which SceneGraph to parse into. If it is provided and non-null and the
+/// MultibodyPlant is not registered as a source, the Parser will perform the
+/// SceneGraph registration into the given plant. We describe this option only
+/// for completeness; we strongly discourage anyone from taking advantage of
+/// this feature.
 class Parser final {
  public:
   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(Parser)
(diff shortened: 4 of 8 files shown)


rpoyner-tri (commented):
Toward #18059.

As I type this, I've gotten a request to add more examples to this PR. Those will appear shortly.
+@jwnimmer-tri for feature review; might as well wait until I paste in the requested examples.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-:-NMB4KQJAfX6ZeiAt7w6:biek0q2)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

rpoyner-tri (commented):
Ok, should be ready for review. Still `r1`, since no comments landed on any code in the mean time.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-:-NMBDqbkCkVZadZiIq47:bqihicf)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
:lgtm: feature.

+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-:-NMJvQMz4pODEooYpmIT:blgxg03)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMJsqzV3k8IskKZhtjH:-NMJsqzV3k8IskKZhtjI:b-yvxw3x) ([raw file](https://github.com/RobotLocomotion/drake/blob/08dc4dca6af2f922a77af9c1b6e8858de6284150/multibody/parsing/parser.h#L86)):*
> ```objc
>   ///   May be nullptr.
>   /// @param model_name_prefix A string that will be added as a scoped name
>   ///   prefix to the names of any models loaded by this parser.
> ```

nit Missing caveat about the empty string.

Ditto below (for the third overload).



_Suggestion:_
```objc
prefix to the names of any models loaded by this parser;
when empty, no scoping will be added.
```
___
*[`tutorials/rendering_multibody_plant.ipynb` line 213 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMJs-KJ84V1hhA1EOeD:-NMJs-KK1o9DPMhrsAr3:b-xr8i05) ([raw file](https://github.com/RobotLocomotion/drake/blob/08dc4dca6af2f922a77af9c1b6e8858de6284150/tutorials/rendering_multibody_plant.ipynb#L213)):*
> ```
>    "outputs": [],
>    "source": [
>     "(iiwa_1,) = Parser(plant, \"1\").AddModels(iiwa_file)\n",
> ```

nit Since the comment just above already talks about "left" vs "right", I think we should improve on the prior `"iiwa_1"` name here by using a clear and meaningful name, `left::iiwa`.

Ditto below.



_Suggestion:_
```
left
```
___
*[`bindings/pydrake/multibody/parsing_py.cc` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMJtJHD4Ep3CCZ5lzlj:-NMJtJHD4Ep3CCZ5lzlk:bf0bzo8) ([raw file](https://github.com/RobotLocomotion/drake/blob/08dc4dca6af2f922a77af9c1b6e8858de6284150/bindings/pydrake/multibody/parsing_py.cc#L80)):*
> ```cpp
>         .def(py::init<MultibodyPlant<double>*, SceneGraph<double>*>(),
>             py::arg("plant"), py::arg("scene_graph") = nullptr,
>             cls_doc.ctor.doc_2args_plant_scene_graph)
> ```

nit To be a little closer to Pythonic, I don't see any need to bind this middle overload.  Instead, in the first overload above we can default the `scene_graph` to none and the `model_name_prefix` to empty. That will subsume the set of function matched by the second overload. Having defaulted kwargs is more Pythonic than a lot of overloads, and the fewer overloads we have here, the clearer the help string will be, likewise any error messages for "cannot resolve overload" when calling this function will be simpler.

rpoyner-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-:-NMUkKu9DDiqzJT4ajJe:b5r1yx7)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM: A few cosmetic things.

> Reviewed 5 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-:-NMVVKNt7Yh11tO5NaQE:beacqjy)* status: 9 unresolved discussions (waiting on @rpoyner-tri)

___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVApZ0CmImXBZACJRM):*
nit: Note to yourself? Otherwise, dressing it up in a sentence would be good.
___
*[`multibody/parsing/parser.h` line 75 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVC6ogCv8NIyFMiUl-:-NMVC6ogCv8NIyFMiUl0:b-vmr5md) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L75)):*
> ```objc
>                   geometry::SceneGraph<double>* scene_graph = nullptr);
> 
>   /// Creates a Parser that adds models to the given plant and scene_graph. The
> ```

BTW These new APIs imply a workflow in which a new parser is created for each desired prefix. Is that left as a job for the reader to infer by design?
___
*[`multibody/parsing/parser.h` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVBE3m9F2LSJhHIjQy:-NMVBE3m9F2LSJhHIjQz:b-q8558e) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L81)):*
> ```objc
>   /// @param plant A pointer to a mutable MultibodyPlant object to which parsed
>   ///   model(s) will be added; `plant->is_finalized()` must remain `false` for
>   ///   as long as the @p plant is in used by `this`.
> ```

nit s/used/use

(The next function as well.)
___
*[`multibody/parsing/parser.h` line 84 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVBUwu3moxv2Je_CQH:-NMVBUwu3moxv2Je_CQI:b-gb8stm) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L84)):*
> ```objc
>   /// @param scene_graph A pointer to a mutable SceneGraph object used for
>   ///   geometry registration (either to model visual or contact geometry).
>   ///   May be nullptr.
> ```

BTW This isn't unique to these new functions; but it doesn't seem the requirements on the relationship between `plant` and `scene_graph` are documented.
___
*[`multibody/parsing/parser.h` line 99 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVBv1R5OWdaQaIV8OF:-NMVBv1R5OWdaQaIV8OG:b-zc46er) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L99)):*
> ```objc
>   ///   model(s) will be added; `plant->is_finalized()` must remain `false` for
>   ///   as long as the @p plant is in used by `this`.
>   /// @param model_name_prefix A string that will be added as a scoped name
> ```

BTW Is the phrase "scoped name" understood by the user for what that means vis a vis model instance names?
___
*[`multibody/parsing/test/parser_test.cc` line 46 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVS6Fa1dv41wzBfsf-:-NMVS6Fa1dv41wzBfsf0:beajpna) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/test/parser_test.cc#L46)):*
> ```cpp
>     EXPECT_EQ(dut.AddModels(sdf_name).size(), 1);
>     dut.AddModelFromFile(sdf_name, "foo");
>     Parser(&plant, "prefix").AddModels(sdf_name);
> ```

nit: Other than confirming that the invocation doesn't throw/crash, I'm not sure what's being tested here. I'd expect a follow up test that confirms that the prefix exists. This note applies to the first three blocks.

The last two blocks *do* test for model instance names. It's really not clue why these don't.
___
*[`multibody/test_utilities/add_fixed_objects_to_plant.h` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVTeLPAKv-g1hJj5AL:-NMVTeLPAKv-g1hJj5AM:b-hzq2rc) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/test_utilities/add_fixed_objects_to_plant.h#L36)):*
> ```objc
>   // Load a model of a table for a robot.
>   const ModelInstanceIndex robot_table_model =
>       Parser(plant, "robot").AddModels(table_sdf_path).at(0);
> ```

BTW Why the change from `robot_table` to `robot`?
___
*[`multibody/test_utilities/add_fixed_objects_to_plant.h` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVTqLg4n9DqcHQbETV:-NMVTqLg4n9DqcHQbETW:bivviv5) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/test_utilities/add_fixed_objects_to_plant.h#L41)):*
> ```objc
> 
>   // Load a second table for objects.
>   Parser parser(plant, "objects");
> ```

BTW I know it would introduce more changes to this file, but calling this `object_parser` would probably serve as a good example to others.
___
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMVUN9796AZYMwuT5cM:b-l1569s) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*
> ```cpp
>             py::arg("model_name_prefix") = "",
>             cls_doc.ctor.doc_3args_plant_scene_graph_model_name_prefix)
>         .def(py::init<MultibodyPlant<double>*, std::string_view>(),
> ```

BTW What's the value of binding this second flavor? Given our preference for keyword arguments, exercising *this* function looks just like exercising the previous. In fact, I suspect that you can't even *call* this function unless you omit the keyword arguments.

jwnimmer-tri:
*[`multibody/test_utilities/add_fixed_objects_to_plant.h` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVTeLPAKv-g1hJj5AL:-NMV_wUj5Mml0npE1UDT:bhy1uj7) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/test_utilities/add_fixed_objects_to_plant.h#L36)):*

BTW Why the change from `robot_table` to `robot`?
</blockquote></details>

The new name will be `robot::extra_heavy_duty_table`, which seems fine.  `robot_table::extra_heavy_duty_table` seems silly.

SeanCurtis-TRI:
*[`multibody/test_utilities/add_fixed_objects_to_plant.h` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVTeLPAKv-g1hJj5AL:-NMVjpK2DCBlmkD_aAsG:b-o6n3e) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/test_utilities/add_fixed_objects_to_plant.h#L36)):*

The new name will be `robot::extra_heavy_duty_table`, which seems fine.  `robot_table::extra_heavy_duty_table` seems silly.
</blockquote></details>

Also, I just realized it's in a `test_utilities` folder; so we don't have to worry about stability issues.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 1 unresolved discussion (waiting on @jwnimmer-tri, @rpoyner-tri, and @SeanCurtis-TRI)

___
*[`multibody/parsing/parser.h` line 75 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVC6ogCv8NIyFMiUl-:-NMVzFRv2hwxQDuEIJeq:b-xvujkf) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L75)):*

BTW These new APIs imply a workflow in which a new parser is created for each desired prefix. Is that left as a job for the reader to infer by design?
</blockquote></details>

Added some class doc; PTAL.
___
*[`multibody/parsing/parser.h` line 84 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVBUwu3moxv2Je_CQH:-NMW23SZE8Fx6ZObVc3D:b-kuj42l) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L84)):*

BTW This isn't unique to these new functions; but it doesn't seem the requirements on the relationship between `plant` and `scene_graph` are documented.
</blockquote></details>

Added some class doc; PTAL. It occurred to me to deprecate taking a scene graph pointer at all, but that is a battle for another day.
___
*[`multibody/parsing/parser.h` line 99 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVBv1R5OWdaQaIV8OF:-NMW2OltCNBs7yKk06X3:b5xp9t) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/multibody/parsing/parser.h#L99)):*

BTW Is the phrase "scoped name" understood by the user for what that means vis a vis model instance names?
</blockquote></details>

See if the code example in class doc helps. If not, I can probably come up with some other documentation to link to.
___
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMW6Bbr6ZfntYGF57yR:b-y3kz39) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*

BTW What's the value of binding this second flavor? Given our preference for keyword arguments, exercising *this* function looks just like exercising the previous. In fact, I suspect that you can't even *call* this function unless you omit the keyword arguments.
</blockquote></details>

I'll look into this a bit more. I know that I tried deleting it after `r1` and something broke; I'll try to be a bit more precise about why it is here.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 2 unresolved discussions (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 81 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMW6pLO8F70AC5CAsWJ:-NMW6pLP7X_co-HnKNUK:b9bbwm5) ([raw file](https://github.com/RobotLocomotion/drake/blob/c429ce79d23d12734597137c19fab91f0f3533aa/multibody/parsing/parser.h#L81)):*
> ```objc
> /// separately), then subsequent parsing will register geometries (visual and
> /// collision), and apply collision filters. Otherwise, geometry and collision
> /// filter inputs will be ignored.
> ```

The prior two paragraphs are not relevant to 99% of our users.  Presenting them in the same level and tone as other, more user-focused elements of the contract does our new users a disservice.

We should somehow make it clear in the markup that nobody needs to care at all about any of this new text, e.g., by marking the first paragraph as "(Advanced)" and smooshing the two paragraphs together.  Or maybe there is some other Doxygen markup to help clarify its prominence.

(For clarity: users should usually call any of the setup sugar functions like AddMultibodyPlant or AddMultibodyPlantSceneGraph, at which point none of this text is relevant to them; they should not pass a `scene_graph` pointer, and geometry will always get registered.)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/parser.h` line 81 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMW6pLO8F70AC5CAsWJ:-NMW8WZE34xiq-kLyime:bj445h6) ([raw file](https://github.com/RobotLocomotion/drake/blob/c429ce79d23d12734597137c19fab91f0f3533aa/multibody/parsing/parser.h#L81)):*

The prior two paragraphs are not relevant to 99% of our users.  Presenting them in the same level and tone as other, more user-focused elements of the contract does our new users a disservice.

We should somehow make it clear in the markup that nobody needs to care at all about any of this new text, e.g., by marking the first paragraph as "(Advanced)" and smooshing the two paragraphs together.  Or maybe there is some other Doxygen markup to help clarify its prominence.

(For clarity: users should usually call any of the setup sugar functions like AddMultibodyPlant or AddMultibodyPlantSceneGraph, at which point none of this text is relevant to them; they should not pass a `scene_graph` pointer, and geometry will always get registered.)
</blockquote></details>

I don't understand which text you want changed or how.
___
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMW8b_z9pX1sjZpzhBU:b-trw63n) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*

I'll look into this a bit more. I know that I tried deleting it after `r1` and something broke; I'll try to be a bit more precise about why it is here.
</blockquote></details>

Recall that pybind resolution takes parameter types into account. Without the second binding, it is impossible to write the following line of Python without expanding all the parameter names:
```
(left_iiwa,) = Parser(plant, "left").AddModels(iiwa_file)
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 1 unresolved discussion (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 81 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMW6pLO8F70AC5CAsWJ:-NMW9VOW77J52tZgqXrX:b-jcebas) ([raw file](https://github.com/RobotLocomotion/drake/blob/c429ce79d23d12734597137c19fab91f0f3533aa/multibody/parsing/parser.h#L81)):*

I don't understand which text you want changed or how.
</blockquote></details>

The problem is with the last two new paragraphs here.  Here's my attempt at a rewording to subsume both paragraphs:

(Advanced) In the rare case where the user is parsing into a MultibodyPlant and SceneGraph but has created them one at a time instead of using the more convenient AddMultibodyPlant() or AddMultibodyPlantSceneGraph() functions, the Parser constructors accept an optional SceneGraph pointer to specify which SceneGraph to parse into. If it is provided and non-null and the MultibodyPlant is not registered as a source, the Parser will perform the SceneGraph registration into the given plant.  We describe this option only for completeness; we strongly discourage anyone from taking advantage of this feature.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 1 unresolved discussion (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 81 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMW6pLO8F70AC5CAsWJ:-NMWAow1-8Sv_so3pJKO:b-4q7ify) ([raw file](https://github.com/RobotLocomotion/drake/blob/c429ce79d23d12734597137c19fab91f0f3533aa/multibody/parsing/parser.h#L81)):*

The problem is with the last two new paragraphs here.  Here's my attempt at a rewording to subsume both paragraphs:

(Advanced) In the rare case where the user is parsing into a MultibodyPlant and SceneGraph but has created them one at a time instead of using the more convenient AddMultibodyPlant() or AddMultibodyPlantSceneGraph() functions, the Parser constructors accept an optional SceneGraph pointer to specify which SceneGraph to parse into. If it is provided and non-null and the MultibodyPlant is not registered as a source, the Parser will perform the SceneGraph registration into the given plant.  We describe this option only for completeness; we strongly discourage anyone from taking advantage of this feature.
</blockquote></details>

(Really, our overall goal should be to deprecate the `scene_graph` argument, rather than document it; but for now documenting is better than the status quo, I suppose.)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18625)* status: 1 unresolved discussion (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 81 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMW6pLO8F70AC5CAsWJ:-NMWB72c5ZYn-or5VMwP:b-8gsj3c) ([raw file](https://github.com/RobotLocomotion/drake/blob/c429ce79d23d12734597137c19fab91f0f3533aa/multibody/parsing/parser.h#L81)):*

(Really, our overall goal should be to deprecate the `scene_graph` argument, rather than document it; but for now documenting is better than the status quo, I suppose.)
</blockquote></details>

Agreed on deprecation.

SeanCurtis-TRI:
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMX0XRr7QWiE-Y2HW-E:b-urnhxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*

Recall that pybind resolution takes parameter types into account. Without the second binding, it is impossible to write the following line of Python without expanding all the parameter names:
```
(left_iiwa,) = Parser(plant, "left").AddModels(iiwa_file)
```
</blockquote></details>

Yeah. That was what I meant by, "I suspect that you can't even call this function unless you omit the keyword arguments." So, it does have a place but it's kind of counter to the practice we foster.

SeanCurtis-TRI:
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMX0fm0BZCDfpWmT4Xg:bxuhj4q) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*

Yeah. That was what I meant by, "I suspect that you can't even call this function unless you omit the keyword arguments." So, it does have a place but it's kind of counter to the practice we foster.
</blockquote></details>

Does that mean that binding the keyword arguments is misleading?

jwnimmer-tri:
*[`bindings/pydrake/multibody/parsing_py.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18625#-NMVUN9796AZYMwuT5cL:-NMXBqaL8eSqnmhlCMim:b-y50smz) ([raw file](https://github.com/RobotLocomotion/drake/blob/ce062f605389ba4ed144089427a1f247d579d2a0/bindings/pydrake/multibody/parsing_py.cc#L78)):*

Does that mean that binding the keyword arguments is misleading?
</blockquote></details>

The names show up in the documentation and help strings, so are probably at least somewhat useful.

PR #18634, opened 2023-01-22
[py yaml] Add explicit type checking errors for `yaml_load*(data)`

EricCousineau-TRI:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634)

### DIFF
--- a/bindings/pydrake/common/test/yaml_test.py
+++ b/bindings/pydrake/common/test/yaml_test.py
@@ -26,6 +26,20 @@ class TestYaml(unittest.TestCase):
             actual = func(data="{foo: bar}")
             self.assertDictEqual(actual, {"foo": "bar"})
 
+    def test_negative(self):
+        with self.assertRaises(RuntimeError) as cm:
+            yaml_load()
+        self.assertRegex(str(cm.exception), "Neither.*but one should be")
+        with self.assertRaises(RuntimeError) as cm:
+            yaml_load(data="1.0", filename="/not_a_real_file")
+        self.assertRegex(str(cm.exception), "Both.*but only one should be")
+        not_str_or_file = {"foo": 1.0}
+        with self.assertRaises(RuntimeError) as cm:
+            yaml_load(data=not_str_or_file)
+        self.assertEqual(
+            str(cm.exception), "Must specify str or file-like object"
+        )
+
     def test_load_merge_keys(self):
         """Sanity check support for merge keys."""
         # See https://yaml.org/type/merge.html for details; pyyaml already
--- a/bindings/pydrake/common/yaml.py
+++ b/bindings/pydrake/common/yaml.py
@@ -35,6 +35,12 @@ class _SchemaLoader(yaml.loader.SafeLoader):
 _SchemaLoader.add_multi_constructor("", _SchemaLoader._handle_multi_variant)
 
 
+def _is_maybe_file_like(f):
+    # N.B. `isinstance(f, typing.IO)` does not seem to work, at least on
+    # CPython 3.10.
+    return hasattr(f, "read")
+
+
 def yaml_load_data(data, *, private=False):
     """Loads and returns the given `data` str as a yaml object, while also
     accounting for variant-like type tags.  Any tags are reported as an
@@ -50,6 +56,8 @@ def yaml_load_data(data, *, private=False):
     without any schema checking nor default values. To load with respect to
     a schema with defaults, see ``yaml_load_typed()``.
     """
+    if not isinstance(data, str) and not _is_maybe_file_like(data):
+        raise RuntimeError("Must specify str or file-like object")
     result = yaml.load(data, Loader=_SchemaLoader)
     if not private:
         keys_to_remove = [
@@ -83,11 +91,20 @@ def yaml_load(*, data=None, filename=None, private=False):
     without any schema checking nor default values. To load with respect to
     a schema with defaults, see ``yaml_load_typed()``.
     """
-    if sum(bool(x) for x in [data, filename]) != 1:
-        raise RuntimeError("Must specify exactly one of data= and filename=")
-    if data:
+    is_data = data is not None
+    is_filename = filename is not None
+    if is_data and is_filename:
+        raise RuntimeError(
+            "Both `data` and `filename` are present, but only one should be"
+        )
+    if not is_data and not is_filename:
+        raise RuntimeError(
+            "Neither `data` nor `filename` are present, but one should be"
+        )
+    if is_data:
         return yaml_load_data(data, private=private)
     else:
+        assert is_filename
         return yaml_load_file(filename, private=private)
 
 


EricCousineau-TRI (commented):
+@jwnimmer-tri for both reviews, please +(status: single reviewer ok) 
Small UX and dev improvement

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-:-NMP1Acg46tCh3aWbIei:b-nsw7r3)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
For tricky logic like this, I think we should also engage a platform reviewer for extra eyes; -(status: single reviewer ok).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-:-NMPH8tK2x23Mi0IcVwG:b-856ppp)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/yaml.py` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPFNPJ7ito5JU6Iav2:-NMPFNPJ7ito5JU6Iav3:b-ojpfzh) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/yaml.py#L94)):*
> ```Python
>     a schema with defaults, see ``yaml_load_typed()``.
>     """
>     if data is None and filename is None:
> ```

I can buy getting rid of the `sum` phrasing as being too confusing, but there are several problems with the new condition:

(1) When _both_ the `data` and `filename` are non-None, we need to fail.  Otherwise, the user will be confused when their `filename` is ignored.

(2) When `data` is false-ish (e.g., the empty string), we need to fail (instead of, e.g., crashing during `_private` handling).  It would be OK to lift this check into `yaml_load_data` instead of this wrapper, though, if you want to switch to using `is None` as the guard here.

My suggestion for utmost clarity would be like so:

```py
is_data = data is not None
is_filename = filename is not None
if is_data and is_filename:
    ... error, cannot pass both ...
if not is_data and not is_filename:
    ... error, missing data or filename ...
if is_data:
    ... yaml_load_data
else:
    assert is_filename
    ... yaml_load_file
```
___
*[`bindings/pydrake/common/yaml.py` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPFzXz3kQ3gt2eureu:-NMPFzXz3kQ3gt2eurev:b-wu63fk) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/yaml.py#L96)):*
> ```Python
>     if data is None and filename is None:
>         raise RuntimeError("Must specify exactly one of data= and filename=")
>     if data:
> ```

nit If we're switching our guard above to check `None` vs false-ish, then we need `data is None` here to be consistent.
___
*[`bindings/pydrake/common/test/yaml_test.py` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPH-qz8h3Zdb5q4wGP:-NMPH-r-Cr14ng9ps0rD:bhg6vcz) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/test/yaml_test.py#L32)):*
> ```Python
>         with self.assertRaises(RuntimeError) as cm:
>             yaml_load()
>         self.assertIn(str(cm.exception), "exactly one of")
> ```

nit When testing for error strings, use `assertRaisesRegex` for clarity and compactness.

Ditto below.

EricCousineau-TRI (commented):
Aye, thanks for catching the bugs! +@SeanCurtis-TRI for review, please!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-:-NMT1bUpBaKLHOCf_yRP:b4yaxeg)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/common/yaml.py` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPFNPJ7ito5JU6Iav2:-NMT1Tto89lcAEZc0kcT:b-6a69ir) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/yaml.py#L94)):*

I can buy getting rid of the `sum` phrasing as being too confusing, but there are several problems with the new condition:

(1) When _both_ the `data` and `filename` are non-None, we need to fail.  Otherwise, the user will be confused when their `filename` is ignored.

(2) When `data` is false-ish (e.g., the empty string), we need to fail (instead of, e.g., crashing during `_private` handling).  It would be OK to lift this check into `yaml_load_data` instead of this wrapper, though, if you want to switch to using `is None` as the guard here.

My suggestion for utmost clarity would be like so:

```py
is_data = data is not None
is_filename = filename is not None
if is_data and is_filename:
    ... error, cannot pass both ...
if not is_data and not is_filename:
    ... error, missing data or filename ...
if is_data:
    ... yaml_load_data
else:
    assert is_filename
    ... yaml_load_file
```
</blockquote></details>

Done, oops and thanks!
___
*[`bindings/pydrake/common/yaml.py` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPFzXz3kQ3gt2eureu:-NMT1UeP67edPqGs8FE8:b-qsx7uj) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/yaml.py#L96)):*

nit If we're switching our guard above to check `None` vs false-ish, then we need `data is None` here to be consistent.
</blockquote></details>

Done. Oops!
___
*[`bindings/pydrake/common/test/yaml_test.py` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPH-qz8h3Zdb5q4wGP:-NMT1Xyg-Vn1UWJk29ff:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/test/yaml_test.py#L32)):*

nit When testing for error strings, use `assertRaisesRegex` for clarity and compactness.

Ditto below.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634)* status: 2 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

___
*[`bindings/pydrake/common/yaml.py` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMPFNPJ7ito5JU6Iav2:-NMU9f8h0DFGJLENaABT:b-6043be) ([raw file](https://github.com/RobotLocomotion/drake/blob/5635ec1c415c95945d635ce821f3a7b05aca6a3e/bindings/pydrake/common/yaml.py#L94)):*

Done, oops and thanks!
</blockquote></details>

Part (2) got missed.

This new test case:

```py
    def test_empty_str(self):
        for func in yaml_load, yaml_load_data:
            actual = func(data="")
```

... produces the following confusing output:

```
bindings/pydrake/common/yaml.py", line 64, in yaml_load_data
    key for key in result.keys()
AttributeError: 'NoneType' object has no attribute 'keys'
```

... whereas previously it was more helpful:

```RuntimeError: Must specify exactly one of data= and filename=```
___
*[`bindings/pydrake/common/test/yaml_test.py` line 33 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMU8zfu12iaophjPxff:-NMU8zfvAR_Ypy0BSNGK:b2mdbpr) ([raw file](https://github.com/RobotLocomotion/drake/blob/0d920cbd980e1a7692f84a6fbed7b4b974794221/bindings/pydrake/common/test/yaml_test.py#L33)):*
> ```Python
>             yaml_load()
>         self.assertRegex(str(cm.exception), "Neither.*but one should be")
>         with self.assertRaises(RuntimeError) as cm:
> ```

nit When testing for error strings, use assertRaisesRegex for clarity and compactness.

Ditto below.

**Review by SeanCurtis-TRI — COMMENTED** [review]
I agree with @jwnimmer-tri's noted defect regarding empty strings and tests.

> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-:-NMUGzR83KVTvY2ToU-p:b-xw6cm)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/yaml.py` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMUEl3T8N3_W5riwpl7:-NMUEl3U3OzT8AlaY2gy:b-z5kqlu) ([raw file](https://github.com/RobotLocomotion/drake/blob/0d920cbd980e1a7692f84a6fbed7b4b974794221/bindings/pydrake/common/yaml.py#L39)):*
> ```Python
> 
> def _is_maybe_file_like(f):
>     # N.B. `isinstance(f, typing.IO)` does not seem to work, at least on
> ```

BTW This comment would be more helpful if we knew in what way it failed: false positives or false negatives?

**Review by jwnimmer-tri — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

___
*[`bindings/pydrake/common/yaml.py` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMUEl3T8N3_W5riwpl7:-NMUHYwB35vQp12WZc_I:bjsnayr) ([raw file](https://github.com/RobotLocomotion/drake/blob/0d920cbd980e1a7692f84a6fbed7b4b974794221/bindings/pydrake/common/yaml.py#L39)):*

BTW This comment would be more helpful if we knew in what way it failed: false positives or false negatives?
</blockquote></details>

In case it helps reword the comment (maybe just by removing it entirely), I would defect any us of `isinstance` in our logic here.

It's very important that we allow for duck-typed, file-like things, and my interpretation of `typing` is that it does not allow for such cases.  Checking for `hasattr("read")` is exactly the correct solution.  If there is a way to use `typing` to enforce that then that's great, but demanding any particular runtime type is wrong.

In any case, to me the most Pythonic thing would be to allow the object to reach `yaml.load` directly without us doing any type-checks, since that function is the final arbiter of what's okay.

It's message isn't even that bad:

```
  File "/usr/lib/python3/dist-packages/yaml/reader.py", line 178, in update_raw
    data = self.stream.read(size)
AttributeError: '...' object has no attribute 'read'
```

I supposed Eric must have hit a case where that message wasn't enough, so I can accept the new ahead-of-time sanity checks in `yaml_load_data`, but I wouldn't nominally expect them to have been necessary.

**Review by SeanCurtis-TRI — COMMENTED** [review]
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @EricCousineau-TRI and @jwnimmer-tri)

___
*[`bindings/pydrake/common/yaml.py` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18634#-NMUEl3T8N3_W5riwpl7:-NMUJwWc-1eYkeVIsAMQ:b-h28i6) ([raw file](https://github.com/RobotLocomotion/drake/blob/0d920cbd980e1a7692f84a6fbed7b4b974794221/bindings/pydrake/common/yaml.py#L39)):*

In case it helps reword the comment (maybe just by removing it entirely), I would defect any us of `isinstance` in our logic here.

It's very important that we allow for duck-typed, file-like things, and my interpretation of `typing` is that it does not allow for such cases.  Checking for `hasattr("read")` is exactly the correct solution.  If there is a way to use `typing` to enforce that then that's great, but demanding any particular runtime type is wrong.

In any case, to me the most Pythonic thing would be to allow the object to reach `yaml.load` directly without us doing any type-checks, since that function is the final arbiter of what's okay.

It's message isn't even that bad:

```
  File "/usr/lib/python3/dist-packages/yaml/reader.py", line 178, in update_raw
    data = self.stream.read(size)
AttributeError: '...' object has no attribute 'read'
```

I supposed Eric must have hit a case where that message wasn't enough, so I can accept the new ahead-of-time sanity checks in `yaml_load_data`, but I wouldn't nominally expect them to have been necessary.
</blockquote></details>

I certainly buy into that reasoning and would support a distillation. I'm good either way.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18634)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform), missing label for release notes (waiting on @EricCousineau-TRI)

PR #18157, opened 2022-10-21
[parsing] Deprecate parser.Add*FromFile() 

rpoyner-tri:
Draft for further discussion. Usual caveats (stale/missing doc/tests, style sins) apply.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18157)

PR #17623, opened 2022-07-25
[parsing] Add model directives support to Parser

rpoyner-tri:
Relevant to #17390. Usual prototype caveats apply.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/17623)

PR #18646, opened 2023-01-24
Port to parser.AddModels, part 7

rpoyner-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646)

### DIFF
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -181,9 +181,8 @@ class TestMeldis(unittest.TestCase):
             "drake/examples/manipulation_station/models/sphere.sdf")
         builder = DiagramBuilder()
         plant, scene_graph = AddMultibodyPlantSceneGraph(builder, 0.001)
-        parser = Parser(plant=plant)
-        sphere1_model = parser.AddModelFromFile(sdf_file, "sphere1")
-        sphere2_model = parser.AddModelFromFile(sdf_file, "sphere2")
+        sphere1_model, = Parser(plant, "sphere1").AddModels(sdf_file)
+        sphere2_model, = Parser(plant, "sphere2").AddModels(sdf_file)
         body1 = plant.GetBodyByName("base_link", sphere1_model)
         body2 = plant.GetBodyByName("base_link", sphere2_model)
         plant.AddJoint(PrismaticJoint(
--- a/geometry/test/meshcat_visualizer_test.cc
+++ b/geometry/test/meshcat_visualizer_test.cc
@@ -316,10 +316,10 @@ GTEST_TEST(MeshcatVisualizerTest, MultipleModels) {
   std::string urdf = FindResourceOrThrow(
       "drake/manipulation/models/iiwa_description/urdf/"
       "iiwa14_no_collision.urdf");
-  auto iiwa0 = multibody::Parser(&plant).AddModelFromFile(urdf);
+  auto iiwa0 = multibody::Parser(&plant).AddModels(urdf).at(0);
   plant.WeldFrames(plant.world_frame(),
                    plant.GetBodyByName("base", iiwa0).body_frame());
-  auto iiwa1 = multibody::Parser(&plant).AddModelFromFile(urdf, "second_iiwa");
+  auto iiwa1 = multibody::Parser(&plant, "second").AddModels(urdf).at(0);
   plant.WeldFrames(plant.world_frame(),
                    plant.GetBodyByName("base", iiwa1).body_frame());
   plant.Finalize();
@@ -339,18 +339,18 @@ GTEST_TEST(MeshcatVisualizerTest, MultipleModels) {
   auto context = diagram->CreateDefaultContext();
 
   EXPECT_FALSE(meshcat->HasPath("/drake/visualizer/iiwa14"));
-  EXPECT_FALSE(meshcat->HasPath("/drake/visualizer/second_iiwa"));
+  EXPECT_FALSE(meshcat->HasPath("/drake/visualizer/second/iiwa14"));
 
   diagram->ForcedPublish(*context);
 
   EXPECT_TRUE(meshcat->HasPath("/drake/visualizer/iiwa14"));
-  EXPECT_TRUE(meshcat->HasPath("/drake/visualizer/second_iiwa"));
+  EXPECT_TRUE(meshcat->HasPath("/drake/visualizer/second/iiwa14"));
   for (int link = 0; link < 8; link++) {
     EXPECT_NE(meshcat->GetPackedTransform(
                   fmt::format("/drake/visualizer/iiwa14/iiwa_link_{}", link)),
               "");
     EXPECT_NE(meshcat->GetPackedTransform(fmt::format(
-                  "/drake/visualizer/second_iiwa/iiwa_link_{}", link)),
+                  "/drake/visualizer/second/iiwa14/iiwa_link_{}", link)),
               "");
   }
 }
--- a/multibody/meshcat/test/contact_visualizer_test.cc
+++ b/multibody/meshcat/test/contact_visualizer_test.cc
@@ -36,15 +36,14 @@ class ContactVisualizerTest : public ::testing::Test {
         multibody::AddMultibodyPlantSceneGraph(&builder, 0.001);
 
     // Add the point contact spheres and joints.
-    multibody::Parser parser(&plant);
     const std::string sdf = FindResourceOrThrow(
         "drake/examples/manipulation_station/models/sphere.sdf");
-    auto sphere1_model = parser.AddModelFromFile(sdf, "sphere1");
+    auto sphere1_model = Parser(&plant, "sphere1").AddModels(sdf).at(0);
     const auto& sphere1 = plant.GetBodyByName("base_link", sphere1_model);
     plant.AddJoint<multibody::PrismaticJoint>(
         "sphere1_x", plant.world_body(), std::nullopt, sphere1, std::nullopt,
         Eigen::Vector3d::UnitX());
-    auto sphere2_model = parser.AddModelFromFile(sdf, "sphere2");
+    auto sphere2_model = Parser(&plant, "sphere2").AddModels(sdf).at(0);
     const auto& sphere2 = plant.GetBodyByName("base_link", sphere2_model);
     plant.AddJoint<multibody::PrismaticJoint>(
         "sphere2_x", plant.world_body(), std::nullopt, sphere2, std::nullopt,
@@ -59,7 +58,8 @@ class ContactVisualizerTest : public ::testing::Test {
     // Add the hydroelastic spheres and joints between them.
     const std::string hydro_sdf = FindResourceOrThrow(
         "drake/multibody/meshcat/test/hydroelastic.sdf");
-    parser.AddModelFromFile(hydro_sdf);
+    multibody::Parser parser(&plant);
+    parser.AddModels(hydro_sdf);
     const auto& body1 = plant.GetBodyByName("body1");
     plant.AddJoint<multibody::PrismaticJoint>(
         "body1", plant.world_body(), std::nullopt, body1, std::nullopt,
@@ -112,10 +112,12 @@ class ContactVisualizerTest : public ::testing::Test {
     diagram_->ForcedPublish(*context_);
     if (expect_geometry_names) {
       EXPECT_TRUE(meshcat_->HasPath(
-          "contact_forces/point/sphere1.base_link+sphere2.base_link.bonus"));
+          "contact_forces/point/sphere1::sphere.base_link+"
+          "sphere2::sphere.base_link.bonus"));
     } else {
       EXPECT_TRUE(meshcat_->HasPath(
-          "contact_forces/point/sphere1.base_link+sphere2.base_link"));
+                      "contact_forces/point/sphere1::sphere.base_link+"
+                      "sphere2::sphere.base_link"));
     }
 
     // If the query object port is not connected, the geometry names will
--- a/multibody/meshcat/test/joint_sliders_test.cc
+++ b/multibody/meshcat/test/joint_sliders_test.cc
@@ -33,8 +33,8 @@ class JointSlidersTest : public ::testing::Test {
 
   void Add(const std::string& resource_path,
            const std::string& model_name = {}) {
-    Parser parser(&plant_);
-    parser.AddModelFromFile(FindResourceOrThrow(resource_path), model_name);
+    Parser parser(&plant_, model_name);
+    parser.AddModels(FindResourceOrThrow(resource_path));
   }
 
   void AddAcrobot() {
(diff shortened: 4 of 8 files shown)


rpoyner-tri (commented):
+@SeanCurtis-TRI for feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NM_XANt5InE5pA7rNGS:b-gpqagm)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

rpoyner-tri (commented):
This is a particularly well-worked-over slice from PR #18157.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NM_XDiW5zXPkbZ80y2U:bqz2b0n)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
The name of this PR reminded me of the move title,  "Leonard, Part 6".

:LGTM: For what this is and what it does.

As you peruse my limited comments below, you'll see that I was surprised by seeing the new API running around in the wild (particularly how it interacts with `MeshCat`). My discontent is wholly constrained to the "upgrade" of `AddModel` singular. When we provided a model name under the old regime, the old name got *replaced*. This led to a compact name (and all of the downstream implications). The new prefix is not so friendly. What used to be "sphere2" is now "sphere2::sphere". We've got verbosity and redundancy with deeper paths in `MeshCat`. Sadly, using only a prefix, there aren't many nice alternate names because "2::sphere" would look like "2::table" in `MeshCat` (vis a vis the path). I.e., I'd have a `sphere` and a `table` as children of the same parent node.  :( So, you're kind of stuck with the redundancy if you don't want to lump the first `foo` with the first `bar`.  :(

Perhaps this has been discussed and I'm just late to the party, but, to coin a phrase, I am sadness.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NM_fyX8BjzeVWsTc9IW:b-4inhaj)* status: 5 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 353 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_gzYD76teA9eUW7Bb:-NM_gzYD76teA9eUW7Bc:b-4eo9dr) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/geometry/test/meshcat_visualizer_test.cc#L353)):*
> ```cpp
>               "");
>     EXPECT_NE(meshcat->GetPackedTransform(fmt::format(
>                   "/drake/visualizer/second/iiwa14/iiwa_link_{}", link)),
> ```

This is an interesting implication. Was this explicitly desired/discussed? This buries things one layer deeper whereas the oldest just changed the name of the root folder. Are we all happy with that?
___
*[`multibody/meshcat/test/contact_visualizer_test.cc` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_hgK64YQO3DDcrPBQ:-NM_hgK64YQO3DDcrPBR:by0fkla) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/meshcat/test/contact_visualizer_test.cc#L115)):*
> ```cpp
>     if (expect_geometry_names) {
>       EXPECT_TRUE(meshcat_->HasPath(
>           "contact_forces/point/sphere1::sphere.base_link+"
> ```

1) It's weird that the *path* to this would be `drake/visualizer/sphere1/sphere/base_link` but its named here as `sphere1::sphere.base_link`.
2) It's also mildly unfortunate that under the old regime, the model "root" `sphere` got *replaced* with `sphere1`, but here, it's extended. That said, it may well be that the previous behavior was too aggressive.

The test would probably look less silly if the prefixes weren't "sphere1" and "sphere2", but "first" and "second" (or "1" and "2"). It's just the "sphere-sphere" pattern that makes it stand out. So, it's less about how prefixes are interpreted and more about serving as a good example of how to name prefixes (like when `iiwa_table` became, simply, `table` -- knowing that "iiwa" was the very next symbol).
___
*[`multibody/meshcat/test/joint_sliders_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_ikvB8DHbaKafqdy3:-NM_ikvB8DHbaKafqdy4:b-d8nfo6) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/meshcat/test/joint_sliders_test.cc#L215)):*
> ```cpp
> 
>   // Confirm that the names are unique.
>   const std::string alpha = "/alpha::acrobot";
> ```

This feels particularly odd where the *path* is `/alpha/acrobot` for the slider name to be named `/alpha::acrobot`. Is the replacement of path with `::` a common thing that I'm not aware of?

Presumably, if there were only a single acrobot without prefix, the slider would say `/acrobot`, right? Which *would* match the path.
___
*[`multibody/plant/test/internal_geometry_names_test.cc` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_jI-u7GjgoGkwj_lO:-NM_jI-u7GjgoGkwj_lP:bfbvss4) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/plant/test/internal_geometry_names_test.cc#L142)):*
> ```cpp
>   const GeometryId id = GetSoleGeometryId("base_link", "sphere1::sphere");
>   const Entry& entry = dut.Find(id);
>   EXPECT_EQ(entry.model_instance_name, "sphere1::sphere");
> ```

BTW Continuing the previous theme, these tests underscore how the new prefix mechanism feels like a step down in this common use case.  :( I hadn't fully grokked the implications from the previous PR.
___
*[`multibody/plant/test/multibody_plant_introspection_test.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_jbVV1JbPbdY72mJP:-NM_jbVV1JbPbdY72mJQ:baruaji) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/plant/test/multibody_plant_introspection_test.cc#L47)):*
> ```cpp
>   Parser parser(&plant);
>   const ModelInstanceIndex robot_table_model =
>       parser.AddModels(table_sdf_path).at(0);
> ```

BTW Presumably you judged that the model name was irrelevant to the test?

jwnimmer-tri (commented):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 353 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_gzYD76teA9eUW7Bb:-NM_mTil2XFbM3zpSsT_:bsdrdhu) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/geometry/test/meshcat_visualizer_test.cc#L353)):*

This is an interesting implication. Was this explicitly desired/discussed? This buries things one layer deeper whereas the oldest just changed the name of the root folder. Are we all happy with that?
</blockquote></details>

Whether adding or removing a prefix is a good idea is out of scope for this test.

The point of this test is to check what the visualizer does when the user _chooses_ to add a model name prefix.

For my taste, I do want model name prefixes to show up as slashes instead of colons.  I want "/left/iiwa" and "/left/gripper" under a single scene tree checkbox "/left" so that I can turn off the whole left-arm at once with a single display checkbox.  If I want to address them one by one, I can just expand the tree widget one more level.
___
*[`multibody/meshcat/test/contact_visualizer_test.cc` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_hgK64YQO3DDcrPBQ:-NM_oi2F58ZmUn6sgJuv:bz6n7qy) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/meshcat/test/contact_visualizer_test.cc#L115)):*

1) It's weird that the *path* to this would be `drake/visualizer/sphere1/sphere/base_link` but its named here as `sphere1::sphere.base_link`.
2) It's also mildly unfortunate that under the old regime, the model "root" `sphere` got *replaced* with `sphere1`, but here, it's extended. That said, it may well be that the previous behavior was too aggressive.

The test would probably look less silly if the prefixes weren't "sphere1" and "sphere2", but "first" and "second" (or "1" and "2"). It's just the "sphere-sphere" pattern that makes it stand out. So, it's less about how prefixes are interpreted and more about serving as a good example of how to name prefixes (like when `iiwa_table` became, simply, `table` -- knowing that "iiwa" was the very next symbol).
</blockquote></details>

Reading the test again, I think the important point for these is that they are the "point contact" (i.e., non-hydroelastic) geometries.  Maybe reflecting that into the name would be most clear, maybe `pointcontact1::sphere` and ``pointcontact2::sphere` or something?  The hydro body being called just "body1" isn't doing us any favors, either. In the big picture, I'm not sure that this test is defectively worse than it was before.
___
*[`multibody/meshcat/test/joint_sliders_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_ikvB8DHbaKafqdy3:-NM_nMvS5v-rpEgf9Pqv:b-35it45) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/meshcat/test/joint_sliders_test.cc#L215)):*

This feels particularly odd where the *path* is `/alpha/acrobot` for the slider name to be named `/alpha::acrobot`. Is the replacement of path with `::` a common thing that I'm not aware of?

Presumably, if there were only a single acrobot without prefix, the slider would say `/acrobot`, right? Which *would* match the path.
</blockquote></details>

I could go either way with the colons vs slashes here.  The colons match what the plant is doing, but the slashes are more consistent with how Meshcat typically displays things.

The only goal here is to choose something familiar to the user so that they know which slider is for which joint.  Probably slashes are a little bit more familiar and compact in that case?

We just never tested with hierarchical model instances names previously, so didn't notice.

rpoyner-tri (commented):
There are reasons we got to this particular place. I may not give a good or complete account, but here's a shot.

* model name replacement by the parser becomes fraught when nearly all parsed formats can present with multiple models. Which one do we replace? How much spelunking do we do to figure that out? See Issue #18059 and prior PR #17623. In particular, your comment https://github.com/RobotLocomotion/drake/pull/17623#pullrequestreview-1135157181 is proof you are not late to the party.
* Redundancy: it's not atypical to pay a redundancy price to avoid ambiguity or collisions. Is there a better/fancier/magic design for that? Maybe. I don't think this solution solves all cases, but it is simple, and doesn't preclude other approaches. In particular, @jwnimmer-tri has proposed auto-salting names, in discussion of draft PR #18157. As for the redundant names in this PR, Nimmer requested them specifically, as opposed to just plain integer prefixes: see https://github.com/RobotLocomotion/drake/pull/18157#pullrequestreview-1267948479.
* The prefixes don't have have any model-theoretic meaning, as you note. It's up to the user to work out naming and de-confliction of their models. But this is no different from the prior scheme of overwriting names. The prefix scheme scales a bit better, but probably not to dozens or hundreds of objects.
* Meshcat paths: meshcat path length is going to be some function of scene complexity no matter what we do. (Maybe O(logN)?)
* '/' vs. '::': This is sadness built into the world by SDFormat, and carried forward in Drake and Anzu over a long period of time. I don't think there is anything worth doing to change the meshcat path notation or the SDFormat notation. In any case, this patch didn't start the fire.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NM_lsx28R6RjbEdlDpH:bxgd6fg)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/meshcat/test/joint_sliders_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_ikvB8DHbaKafqdy3:-NM_thHHA3lE6Uts0zC-:b-k6me86) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/meshcat/test/joint_sliders_test.cc#L215)):*

I could go either way with the colons vs slashes here.  The colons match what the plant is doing, but the slashes are more consistent with how Meshcat typically displays things.

The only goal here is to choose something familiar to the user so that they know which slider is for which joint.  Probably slashes are a little bit more familiar and compact in that case?

We just never tested with hierarchical model instances names previously, so didn't notice.
</blockquote></details>

(Since this is just uncovering a latent bug, a TODO here that "we'd probably prefer slashes for names nesting on sliders labels" would be one way to conclude this thread.)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`multibody/plant/test/multibody_plant_introspection_test.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NM_jbVV1JbPbdY72mJP:-NM_tRZD8D3B6pEbuZl_:bcozo3m) ([raw file](https://github.com/RobotLocomotion/drake/blob/731f7b398fedf0306707dac03459b0c06666a311/multibody/plant/test/multibody_plant_introspection_test.cc#L47)):*

BTW Presumably you judged that the model name was irrelevant to the test?
</blockquote></details>

They often are when we go to the trouble of saving the model instance index. The old name was set (and could have been deleted); it was never referred to again. Does some other context suffer from the loss of that name? I don't know; I'm not sure that it matters much for this test.

jwnimmer-tri (commented):
I don't mind reviewing but I'm about to time out for the day, and I have a lot of interrupts tomorrow.  I'd be OK if you guys wanted to flip it to Xuchen, depending on how quickly you want a response.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NMa2LUc4KDzXqnTBBjX:b-foca0y)* status: LGTM missing from assignee jwnimmer-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+@xuchenhan-tri for platform review. I might be slow to respond with patch changes as today is my Puget=>Jammy upgrade day.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NMdbgySASQ5Tz_rpJTI:bfcxlwq)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @rpoyner-tri and @xuchenhan-tri)

xuchenhan-tri (commented):
:lgtm: and :crossed_fingers: for your upgrade.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-:-NMe8CZmA5vhXWkkNAnJ:bjx9sze)* status: 1 unresolved discussion (waiting on @rpoyner-tri)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 319 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18646#-NMe7Fza1Q2a5vayM0yn:-NMe7Fza1Q2a5vayM0yo:b5qct13) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fe251cf37bff48c8f5592d5d517f7fb63b19d37/geometry/test/meshcat_visualizer_test.cc#L319)):*
> ```cpp
>       "drake/manipulation/models/iiwa_description/urdf/"
>       "iiwa14_no_collision.urdf");
>   auto iiwa0 = multibody::Parser(&plant).AddModels(urdf).at(0);
> ```

BTW, it looks like there's many repetitions of the same pattern in this PR where an sdf/urdf containing a single model is parsed and we extract the only model instance via `AddModels(file).at(0)`. Has a sugar for that been considered?

PR #18402, opened 2022-12-05
[multibody] Add ContactResults calculation to DiscreteUpdateManager

joemasterjohn:
This PR adds the calculation of `ContactResults` for both point contact and hydroelastic contact to the discrete update manager. Eventually, `MultibodyPlant` will defer computation of contact results entirely to the discrete update manager when in discrete mode.

We add the NVI:
- `DiscreteUpdateManager::CalcContactResults()`
 
And the corresponding implementation:
- `CompliantContactManager::DoCalcContactResults()`

along with various helper methods.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18402)

PR #18265, opened 2022-11-04
Implements TamsiDriver

amcastro-tri:
Towards https://github.com/RobotLocomotion/drake/issues/16955 and https://github.com/RobotLocomotion/drake/issues/13888.

`TamsiDriver` split from the bigger PR #18170 where all of the TAMSI code in `MultibodyPlant` is moved into the `TamsiDriver`. 
Most of the code in `TamsiDriver` introduced in this PR is simply a copy from already tested `MultibodyPlant` code.
A simple unit test verifies we can properly use the `TamsiDriver`, though as soon as #18170, all of the MultibodyPlant unit tests using TAMSI will go through `TamsiDriver`, effectively adding more coverage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18265)

PR #18170, opened 2022-10-24
WIP: Move TAMSI related code out of MultibodyPlant

amcastro-tri:
Towards #16955.

PR #18121, opened 2022-10-17
Removes experimental ContactSolver workflow from MultibodyPlant

amcastro-tri:
Towards #16955.

This PR effectively removes the experimental workflow of setting a ContactSolver.
This will allow for a cleaner transition of TAMSI code into its own TamsiDriver as documented in #17741. 
We can reconsider later if we'd like to bring the ContactSolver workflow back or revamp it to a more useful form.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18121)

PR #18647, opened 2023-01-24
Consolidate contact results

amcastro-tri:
Towards #16955. The result of work that started long ago, #17741, #18121, #18170, #18265, #18402.

This PR moves the responsibility of computing contact results from MultibodyPlant into the DiscreteUpdateManager. This allows for the consistent computation of contact results according to the underlying contact model (regularization with TAMSI, convex relaxation with SAP).

This is done in two parts (two commits):
  1. TamsiDriver computes a consistent contact Jacobian from the pair kinematics computed by the manager. Therefore now both SAP and TAMSI always use consistent sign conventions hardwired in the Jacobian.
  2. Removal of unused Jacobian computation code in the MultibodyPlant.
  
Unit tests already in master verify that this refactoring of the code does not break existing functionality. 

## Note
Thus far the reporting of contact results in discrete mode has been an approximation. We've approximated contact results using the same state dependent computation we use in continuous mode, for both point and hydro models. This approximation has proven useful for most quasi-satic sims we run in the past, but it can lead to observable inconsistencies specially during transients.
This PR fixes that (therefore the release notes label). Now contact results are consistent for either choice of the contact solver. cc'ing @edrumwri here since I believe this was the problem you recently reported.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)

### DIFF
--- a/examples/multibody/inclined_plane_with_body/inclined_plane_with_body.cc
+++ b/examples/multibody/inclined_plane_with_body/inclined_plane_with_body.cc
@@ -8,6 +8,7 @@
 #include "drake/lcm/drake_lcm.h"
 #include "drake/multibody/benchmarks/inclined_plane/inclined_plane_plant.h"
 #include "drake/multibody/plant/contact_results_to_lcm.h"
+#include "drake/multibody/plant/multibody_plant_config_functions.h"
 #include "drake/systems/analysis/simulator.h"
 #include "drake/systems/framework/diagram_builder.h"
 
@@ -58,14 +59,21 @@ DEFINE_bool(is_inclined_plane_half_space, true,
             "Is inclined plane a half-space (true) or box (false).");
 DEFINE_string(bodyB_type, "sphere", "Valid body types are "
               "'sphere', 'block', or 'block_with_4Spheres'");
+DEFINE_string(contact_solver, "tamsi", "Options are: "
+              "'tamsi', 'sap'");
 
 using drake::multibody::MultibodyPlant;
 
 int do_main() {
   // Build a generic multibody plant.
   systems::DiagramBuilder<double> builder;
-  auto [plant, scene_graph] = AddMultibodyPlantSceneGraph(
-      &builder, std::make_unique<MultibodyPlant<double>>(FLAGS_time_step));
+
+  MultibodyPlantConfig plant_config;
+  plant_config.time_step = FLAGS_time_step;
+  plant_config.stiction_tolerance = FLAGS_stiction_tolerance;
+  plant_config.discrete_contact_solver = FLAGS_contact_solver;
+  auto [plant, scene_graph] =
+      multibody::AddMultibodyPlant(plant_config, &builder);
 
   // Set constants that are relevant whether body B is a sphere or block.
   const double massB = 0.1;       // Body B's mass (kg).
--- a/multibody/plant/compliant_contact_manager.cc
+++ b/multibody/plant/compliant_contact_manager.cc
@@ -798,7 +798,17 @@ void CompliantContactManager<T>::CalcHydroelasticContactInfo(
   DRAKE_DEMAND(vn.size() == num_contacts);
   DRAKE_DEMAND(vt.size() == 2 * num_contacts);
 
-  int num_point_contacts = plant().EvalPointPairPenetrations(context).size();
+  // If the point contact model is used, hydroelastic contact pairs are appended
+  // after point contact pairs.
+  // TODO(amcastro-tri): right now EvalPointPairPenetrations() throws an
+  // exception if only hydroelastic is used. Consider a solution in which this
+  // method always returns a valid result, possibly empty.
+  // A possible solution could be moving this method into DiscreteUpdateManager
+  // if the plant no longer uses it.
+  const int num_point_contacts =
+      plant().get_contact_model() == ContactModel::kHydroelastic
+          ? 0
+          : plant().EvalPointPairPenetrations(context).size();
   const int num_surfaces = all_surfaces.size();
 
   std::vector<SpatialForce<T>> F_Ao_W_per_surface(num_surfaces,
--- a/multibody/plant/contact_jacobians.h
+++ b/multibody/plant/contact_jacobians.h
@@ -10,16 +10,15 @@ namespace drake {
 namespace multibody {
 namespace internal {
 
-// Stores the computed contact Jacobians when a point contact model is used.
+// Stores the computed contact Jacobians when a discrete contact model is used.
 // At a given state of the multibody system, there will be `nc` contact pairs.
 // For each penetration pair involving bodies A and B a contact frame C is
 // defined by the rotation matrix `R_WC = [Cx_W, Cy_W, Cz_W]` where
 // `Cz_W = nhat_BA_W` equals the normal vector pointing from body B into body
-// A, expressed in the world frame W. See PenetrationAsPointPair for further
+// A, expressed in the world frame W. See DiscreteContactPair for further
 // details on the definition of each contact pair. Versors `Cx_W` and `Cy_W`
 // constitute a basis of the plane normal to `Cz_W` and are arbitrarily chosen.
 // Below, v denotes the vector of generalized velocities, of size `nv`.
-// @see MultibodyPlant::EvalContactJacobians().
 template <class T>
 struct ContactJacobians {
   // Normal contact Jacobian.
--- a/multibody/plant/discrete_update_manager.cc
+++ b/multibody/plant/discrete_update_manager.cc
@@ -91,14 +91,6 @@ DiscreteUpdateManager<T>::EvalContactSolverResults(
 }
 
 template <typename T>
-const internal::ContactJacobians<T>&
-DiscreteUpdateManager<T>::EvalContactJacobians(
-    const systems::Context<T>& context) const {
-  return MultibodyPlantDiscreteUpdateManagerAttorney<T>::EvalContactJacobians(
-      plant(), context);
-}
-
-template <typename T>
 const std::vector<geometry::ContactSurface<T>>&
 DiscreteUpdateManager<T>::EvalContactSurfaces(
     const systems::Context<T>& context) const {
--- a/multibody/plant/discrete_update_manager.h
+++ b/multibody/plant/discrete_update_manager.h
@@ -216,9 +216,6 @@ class DiscreteUpdateManager : public ScalarConvertibleComponent<T> {
   // N.B. Keep the spelling and order of declarations here identical to the
   // MultibodyPlantDiscreteUpdateManagerAttorney spelling and order of same.
 
-  const internal::ContactJacobians<T>& EvalContactJacobians(
-      const systems::Context<T>& context) const;
-
   const std::vector<geometry::ContactSurface<T>>& EvalContactSurfaces(
       const systems::Context<T>& context) const;
 
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -22,7 +22,6 @@
 #include "drake/math/rotation_matrix.h"
 #include "drake/multibody/contact_solvers/sparse_linear_operator.h"
 #include "drake/multibody/hydroelastics/hydroelastic_engine.h"
-#include "drake/multibody/plant/discrete_contact_pair.h"
 #include "drake/multibody/plant/externally_applied_spatial_force.h"
 #include "drake/multibody/plant/hydroelastic_traction_calculator.h"
 #include "drake/multibody/plant/make_discrete_update_manager.h"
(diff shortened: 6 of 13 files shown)


amcastro-tri (commented):
CI was failing due to a bug in the compliant contact manager. I fixed it and pushed additional unit tests. 
PTAL @joemasterjohn 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-:-NMdc9u54bhWQBlENq52:bv6rndg)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @joemasterjohn)

joemasterjohn (commented):
Thanks! :lgtm: as is, but I left a few suggestions.

> Reviewed 9 of 9 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-:-NMeEAwCCsBmFcUzLWX6:b-2sx8v0)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri)

___
*[`multibody/plant/compliant_contact_manager.cc` line 803 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdtwo03fPSnWBwQuxq:-NMdtwo03fPSnWBwQuxr:b4cknts) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/compliant_contact_manager.cc#L803)):*
> ```cpp
>   // If the point contact model is used, hydroelastic contact pairs are appended
>   // after point contact pairs.
>   const int num_point_contacts =
> ```

Ah good catch there, but it seems to me like `EvalPointPairPenetrations()` should be aware of the contact model and report 0 when in pure hydroelastic mode. WDYT?
___
*[`multibody/plant/contact_jacobians.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMddkzV7SqGO96uQQLS:-NMddkzWAZpWeyHY8qJU:b-8r8v07) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac8722c2efa24479828f1f6e6e5eed9c1ccb14/multibody/plant/contact_jacobians.h#L21)):*
> ```objc
> // details on the definition of each contact pair. Versors `Cx_W` and `Cy_W`
> // constitute a basis of the plane normal to `Cz_W` and are arbitrarily chosen.
> // Below, v denotes the vector of generalized velocities, of size `nv`.
> ```

nit: This comment doesn't need to be removed.



_Code quote:_
```objc
// Below, v denotes the vector of generalized velocities, of size `nv`.
```
___
*[`multibody/plant/tamsi_driver.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdvzoF5_f5N-lD_OUE:-NMdvzoF5_f5N-lD_OUF:b6ftmuz) ([raw file](https://github.com/RobotLocomotion/drake/blob/840d9cecfa031acdb0e376de7670fcfb3caf297e/multibody/plant/tamsi_driver.cc#L50)):*
> ```cpp
>     }
>     contact_jacobians.Jt.middleRows(2 * i, 2) =
>         contact_jacobians.Jc.middleRows(3 * i, 2);
> ```

nit: Is there something stopping us from making `Jt` and `Jn` views on `Jc` using `Eigen::Map` or `Eigen::Ref`? They seem like perfect candidates to me and would hopefully avoid having to do copies like this.

A small but simple optimization you could do right now though would be to copy to the sub-blocks you need in the loop rather than the whole rows of the dense Jacobian (particularly when `(treeA_nv + treeB_nv) << nv`):

```
      contact_jacobians.Jt.block(2 * i, col_offset, 2, tree_nv) =
          tree_jacobian.J.middleRows(0, 2);
      contact_jacobians.Jn.block(i, col_offset, 1, tree_nv) =
          tree_jacobian.J.row(2);
```



_Code quote:_
```cpp
    contact_jacobians.Jt.middleRows(2 * i, 2) =
        contact_jacobians.Jc.middleRows(3 * i, 2);
    contact_jacobians.Jn.row(i) = contact_jacobians.Jc.row(3 * i + 2);
```

joemasterjohn (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-:-NMeEJ_G7ybaVU8YMzZ2:b223nro)* status: 3 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @xuchenhan-tri)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @joemasterjohn and @xuchenhan-tri)

___
*[`multibody/plant/compliant_contact_manager.cc` line 803 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdtwo03fPSnWBwQuxq:-NMegdhqADrYdTw9GMav:b-evkz9e) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/compliant_contact_manager.cc#L803)):*

Ah good catch there, but it seems to me like `EvalPointPairPenetrations()` should be aware of the contact model and report 0 when in pure hydroelastic mode. WDYT?
</blockquote></details>

I considered this, but I really wanted to avoid additional churn on MultibodyPlant. 
I'll take a look, but I'll say no for this PR, unless I find it's a trivial change.
___
*[`multibody/plant/tamsi_driver.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdvzoF5_f5N-lD_OUE:-NMeh3zOE2JXarYnagq1:bj8n6ee) ([raw file](https://github.com/RobotLocomotion/drake/blob/840d9cecfa031acdb0e376de7670fcfb3caf297e/multibody/plant/tamsi_driver.cc#L50)):*
> Is there something stopping us from making Jt and Jn views on Jc using Eigen::Map or Eigen::Ref?

They are not contiguous chunks of Jc, but rather there is a stride.

In any case, here we'd be optimizing something that costs nothing, on a piece of old code (TAMSI) that will soon cease to exist.

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @xuchenhan-tri)

___
*[`multibody/plant/compliant_contact_manager.cc` line 803 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdtwo03fPSnWBwQuxq:-NMeqy1pC2F8cLpBENQX:bb73djm) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/compliant_contact_manager.cc#L803)):*

I considered this, but I really wanted to avoid additional churn on MultibodyPlant. 
I'll take a look, but I'll say no for this PR, unless I find it's a trivial change.
</blockquote></details>

SGTM
___
*[`multibody/plant/tamsi_driver.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdvzoF5_f5N-lD_OUE:-NMeqGnx7ZwreBmkCXTR:b36yqen) ([raw file](https://github.com/RobotLocomotion/drake/blob/840d9cecfa031acdb0e376de7670fcfb3caf297e/multibody/plant/tamsi_driver.cc#L50)):*

> Is there something stopping us from making Jt and Jn views on Jc using Eigen::Map or Eigen::Ref?

They are not contiguous chunks of Jc, but rather there is a stride.

In any case, here we'd be optimizing something that costs nothing, on a piece of old code (TAMSI) that will soon cease to exist.
</blockquote></details>

That's fair. Do we have an end-of-life policy for TAMSI and TAMSI related issues? Because even if it goes away someday on master, I'm sure there will be users lagging behind for at least 6 months as we've seen before.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @joemasterjohn and @xuchenhan-tri)

___
*[`multibody/plant/compliant_contact_manager.cc` line 803 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdtwo03fPSnWBwQuxq:-NMesksa4YvZkbLNYB4T:b-4fbu9r) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/compliant_contact_manager.cc#L803)):*

SGTM
</blockquote></details>

on a first try, it is not trivial. This having to do with the fact that we'd need to return a reference to an object in the cache. It turns out that there are no guarantees on the penetration pairs having zero size, even if using hydro only or hydro with fallback. 
___
*[`multibody/plant/tamsi_driver.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdvzoF5_f5N-lD_OUE:-NMesao51C0jNSohPit-:ba975fa) ([raw file](https://github.com/RobotLocomotion/drake/blob/840d9cecfa031acdb0e376de7670fcfb3caf297e/multibody/plant/tamsi_driver.cc#L50)):*

That's fair. Do we have an end-of-life policy for TAMSI and TAMSI related issues? Because even if it goes away someday on master, I'm sure there will be users lagging behind for at least 6 months as we've seen before.
</blockquote></details>

not yet. First we have to make sure SAP becomes the default. It'll probably be a while until this code gets officially removed.

xuchenhan-tri (commented):
Platform :lgtm: Thanks for the clean up!

> Reviewed 8 of 9 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-:-NMexLWF3auSHYt-0BCD:b-6ygdia)* status: 8 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @joemasterjohn)

___
*[`-- commits` line 3 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMexBr-9vQzb72ay1Fs):*
BTW, I'd recommend squashing this into a single commit for ease of merging.
___
*[`multibody/plant/compliant_contact_manager.cc` line 803 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdtwo03fPSnWBwQuxq:-NMev5Tu-iHfOP7Dxoob:b-gg6hqx) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/compliant_contact_manager.cc#L803)):*

on a first try, it is not trivial. This having to do with the fact that we'd need to return a reference to an object in the cache. It turns out that there are no guarantees on the penetration pairs having zero size, even if using hydro only or hydro with fallback. 
</blockquote></details>

It seems like a TODO is appropriate here. If the plan is for MbP to eventually delegate `EvalPointPairPenetrations` to the manager, it can be recorded in the TODO.
___
*[`multibody/plant/contact_jacobians.h` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMetzMb5XcjgFZuBBGN:-NMetzMb5XcjgFZuBBGO:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/contact_jacobians.h#L19)):*
> ```objc
> // `Cz_W = nhat_BA_W` equals the normal vector pointing from body B into body
> // A, expressed in the world frame W. See DiscreteContactPair for further
> // details on the definition of each contact pair. Versors `Cx_W` and `Cy_W`
> ```

typo



_Code quote:_
```objc
Versors
```
___
*[`multibody/plant/multibody_plant.h` line 5009 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMethM62kQaZTuTxtAl:-NMethM77QOJli9UPJYf:bdmjeec) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac8722c2efa24479828f1f6e6e5eed9c1ccb14/multibody/plant/multibody_plant.h#L5009)):*
> ```objc
>   // results.
>   //
>   // See ContactJacobians for specifics on the returned data storage.
> ```

minor, it might be worth it to migrate this reference to `ContactJacobians` along with the function to `tamsi_driver.h`
___
*[`multibody/plant/tamsi_driver.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMdvzoF5_f5N-lD_OUE:-NMesLZC1zIYywUK7tu4:b-pjk3x9) ([raw file](https://github.com/RobotLocomotion/drake/blob/840d9cecfa031acdb0e376de7670fcfb3caf297e/multibody/plant/tamsi_driver.cc#L50)):*

That's fair. Do we have an end-of-life policy for TAMSI and TAMSI related issues? Because even if it goes away someday on master, I'm sure there will be users lagging behind for at least 6 months as we've seen before.
</blockquote></details>

The default storage order for eigen is column major, so any row data is not contiguous in memory.
___
*[`multibody/plant/tamsi_driver.cc` line 110 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMevn6380O8srjhMbID:-NMevn6380O8srjhMbIE:bjw0s7c) ([raw file](https://github.com/RobotLocomotion/drake/blob/dbe0afc4ec429277768f6372372417c94b0439cc/multibody/plant/tamsi_driver.cc#L110)):*
> ```cpp
> 
>   // Compute normal and tangential velocity Jacobians at t0.
>   const internal::ContactJacobians<T>& contact_jacobians =
> ```

nit, no need to keep the reference here now that we switched to a Calc function.



_Suggestion:_
```cpp
  const internal::ContactJacobians<T> contact_jacobians =
      CalcContactJacobians(context);
```
___
*[`multibody/plant/test/compliant_contact_manager_test.cc` line 796 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMevzba13uVW5PhuvFk:-NMevzba13uVW5PhuvFl:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/dbe0afc4ec429277768f6372372417c94b0439cc/multibody/plant/test/compliant_contact_manager_test.cc#L796)):*
> ```cpp
>   // For this case the friction force must be on the friction cone. For TAMSI
>   // accuracy of this prediction depends on the accuracy of the slip speed,
>   // which in turns depends on the convergence tolerance of the solver. For SAP,
> ```

typo



_Suggestion:_
```cpp
// which in turn depends on the convergence tolerance of the solver. For SAP,
```
___
*[`multibody/plant/test/compliant_contact_manager_test.cc` line 842 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMewwAuD5x5KggjMosu:-NMewwAuD5x5KggjMosv:bwc1mru) ([raw file](https://github.com/RobotLocomotion/drake/blob/dbe0afc4ec429277768f6372372417c94b0439cc/multibody/plant/test/compliant_contact_manager_test.cc#L842)):*
> ```cpp
>     // friction cone. i.e. |fₜ| = μ|fₙ|
>     const double tolerance =
>         kRelativeSlipTolerance * kMu_ * F_Ac_W.translational().z();
> ```

nit, is this term guaranteed to be positive?



_Code quote:_
```cpp
F_Ac_W.translational().z()
```
___
*[`multibody/plant/test/compliant_contact_manager_test.cc` line 857 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMewhx05Nlq6-vbbm1-:-NMewhx05Nlq6-vbbm10:b-k3gcts) ([raw file](https://github.com/RobotLocomotion/drake/blob/dbe0afc4ec429277768f6372372417c94b0439cc/multibody/plant/test/compliant_contact_manager_test.cc#L857)):*
> ```cpp
> std::vector<ContactTestConfig> MakeTestCases() {
>   return std::vector<ContactTestConfig>{
>       {.description = "HydroelasticContactWithFallback_SAP",
> ```

nit, the name of the test doesn't match the config.



_Code quote:_
```cpp
HydroelasticContactWithFallback_SAP
```

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)* status: :shipit: complete! all discussions resolved, LGTM from assignees joemasterjohn,xuchenhan-tri(platform) (waiting on @amcastro-tri)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18647)* status: :shipit: complete! all discussions resolved, LGTM from assignees joemasterjohn,xuchenhan-tri(platform) (waiting on @amcastro-tri)

___
*[`multibody/plant/contact_jacobians.h` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18647#-NMetzMb5XcjgFZuBBGN:-NMjdAo4D1Hlv_Ccf2XM:b-ucrww6) ([raw file](https://github.com/RobotLocomotion/drake/blob/70e2544ceb7c99e281079044cf3c323f5b98f40a/multibody/plant/contact_jacobians.h#L19)):*

Didn't know about this word before! TIL.
However, it seems to mean unit quaternions whereas here I believe you are referring to unit vectors.
</blockquote></details>

I remember seeing that term a lot in physics books. I should be able to find a better reference than a wiki article...

PR #7742, opened 2018-01-12
pydrake: Use transitive `py::keep_alive` to aid in REPL-friendly behavior.

EricCousineau-TRI:
Relates #7660.
Per discussions in Platform Reviewer's meeting and after.

\cc @jwnimmer-tri @sammy-tri @ggould-tri @rdeits 

This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/7742)

PR #18650, opened 2023-01-25
[pydrake] Bind planning.RobotDiagram and related

jwnimmer-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)


jwnimmer-tri:

SeanCurtis-TRI (commented):
:LGTM:

Some tiny stuff.

> Reviewed 3 of 7 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-:-NMeoBrL-yXEWRTgyfk8:bswsjof)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/BUILD.bazel` line 13 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMeijf00F5cnrBSaSJX:-NMeijf00F5cnrBSaSJY:b4b53cz) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/BUILD.bazel#L13)):*
> ```Starlark
> load(
>     "@drake//tools/skylark:drake_py.bzl",
>     "drake_py_library",
> ```

BTW Presumably including this is boilerplate because the expectation is that this module must inevitably include a `drake_py_library`?
___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 1 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMeoQhC7cOXq3m1_f0i:-NMeoQhDB1yDVuDvYWwY:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L1)):*
> ```cpp
> #include "pybind11/eigen.h"
> ```

nit: unused
___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMel8QA3FIkOQV7ZdeI:b-7z3fo9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*
> ```cpp
>               cls_doc.IsDiagramBuilt.doc)
>           .def("BuildDiagram", &Class::BuildDiagram,
>               // Keep alive, ownership (tr.): `return` keeps `self` alive.
> ```

This one confuses me. Two reasons:

1. In the actual invocation, the "return" value is in the patient slot and not the nurse slot (contradicting the comment).
2. Second, it's not clear why there should be any continuing relationship between the builder and the diagram after it has been built. Surely, the builder moves all of its assets over to the diagram upon building? And the diagram knows nothing of its builder. So, what would the dependency be?
___
*[`bindings/pydrake/planning/test/robot_diagram_test.py` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMenHtN1wCBbtcVA6bR:-NMemsG44qXfq9DXp5b9:bahv6cs) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/test/robot_diagram_test.py#L24)):*
> ```Python
>             dut.mutable_parser().AddModels(FindResourceOrThrow(
>                 "drake/manipulation/models/iiwa_description/urdf/"
>                 "iiwa14_spheres_dense_collision.urdf"))
> ```

BTW I can't help but wonder for these smoke tests if it wouldn't be better to parse a smaller file. `box.urdf`?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMev6sh406RideMqqWK:bl4ubif) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

This one confuses me. Two reasons:

1. In the actual invocation, the "return" value is in the patient slot and not the nurse slot (contradicting the comment).
2. Second, it's not clear why there should be any continuing relationship between the builder and the diagram after it has been built. Surely, the builder moves all of its assets over to the diagram upon building? And the diagram knows nothing of its builder. So, what would the dependency be?
</blockquote></details>

I attempted to copy-paste the same annotation as `DiagramBuilder::Build` uses, since this has exactly the same semantics.  I'll double-check it now.
___
*[`bindings/pydrake/planning/test/robot_diagram_test.py` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMenHtN1wCBbtcVA6bR:-NMeupx511BT9qkxQEVW:b-sto1u5) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/test/robot_diagram_test.py#L24)):*

BTW I can't help but wonder for these smoke tests if it wouldn't be better to parse a smaller file. `box.urdf`?
</blockquote></details>

This matches what the C++ unit test does.  I suppose it's wasting a few cpu cycles, but it means that the pydrake test is exactly as stable as the C++ test, which is a minor win.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMexNb4-SrBkpZm_Src:be54czm) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*
Done

> It's not clear why there should be any continuing relationship between the builder and the diagram after it has been built. Surely, the builder moves all of its assets over to the diagram upon building? And the diagram knows nothing of its builder. So, what would the dependency be?

We have `Nurse == 1` which is `self` aka `RobotDiagramBuilder`.
We have `Patient == 0` which is `return` aka `RobotDiagram`.

The [pybind11 docs](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive) say:

> ... the [RobotDiagram return] should be kept alive at least until the [RobotDiagramBuilder self] is freed by the garbage collector.

In other words, so long as the user has the builder still in play we shouldn't free the diagram.

The keep-alive on `DiagramBuilder::Build` was added in #7742, and think it still applies here for the same reason.

If someone did this ...

```
def create():
  builder = RobotDiagramBuilder()
  plant = builder.plant()
  builder.Build()
  return plant
```

... then we should _not_ be freeing the diagram, since the plant still refers to it.  In short, since the getters for the builder return internal references to the builder e.g. the MbP, once we transfer the MbP ownership to the diagram then the diagram has to stay around longer than the builder.

> In the actual invocation, the "return" value is in the patient slot and not the nurse slot (contradicting the comment).

Fixed.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMf4FB24U4hUExXT6iJ:b-jo1gxd) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

Done

> It's not clear why there should be any continuing relationship between the builder and the diagram after it has been built. Surely, the builder moves all of its assets over to the diagram upon building? And the diagram knows nothing of its builder. So, what would the dependency be?

We have `Nurse == 1` which is `self` aka `RobotDiagramBuilder`.
We have `Patient == 0` which is `return` aka `RobotDiagram`.

The [pybind11 docs](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive) say:

> ... the [RobotDiagram return] should be kept alive at least until the [RobotDiagramBuilder self] is freed by the garbage collector.

In other words, so long as the user has the builder still in play we shouldn't free the diagram.

The keep-alive on `DiagramBuilder::Build` was added in #7742, and think it still applies here for the same reason.

If someone did this ...

```
def create():
  builder = RobotDiagramBuilder()
  plant = builder.plant()
  builder.Build()
  return plant
```

... then we should _not_ be freeing the diagram, since the plant still refers to it.  In short, since the getters for the builder return internal references to the builder e.g. the MbP, once we transfer the MbP ownership to the diagram then the diagram has to stay around longer than the builder.

> In the actual invocation, the "return" value is in the patient slot and not the nurse slot (contradicting the comment).

Fixed.
</blockquote></details>

This is incredibly troublesome.

1. The source example of this [DiagramBuilder::Build](https://github.com/RobotLocomotion/drake/blob/master/bindings/pydrake/systems/framework_py_semantics.cc#L645-L647) has disagreement between documentation and code. (The same disagreement that this originally had).

Which one is correct? Return keeps self alive or self keeps return alive.

__return keeps self__
 1. I have the residue of a memory about newly generated classes having to keep their factory classes alive (bloating our memory footprint). If that memory is correct *and* if it applies to this, this would be an argument in favor of this example.
2. The Builder::plant() method is bound as `py_rvp::reference_internal`.  In PR #7742, it has the following quote: "py::return_value_policy::reference_internal` implies `py::keep_alive<0, 1>`, which implies "Keep alive, reference: `return` keeps `self` alive".". So, in the example you gave above, the `plant` object is already doing due diligence in keeping the builder alive.

__self keeps return__
Pretend that `Builder::plant()` didn't already keep the builder alive, and the only thing we've said is that the builder keeps the diagram alive. Then, when we exit the `create()` context, nothing keeps the `builder` alive. It gets destroyed, allowing the diagram to get destroyed, leaving the returned `plant` object pointing to a destructed object.

__One last subtlety__
The weirdness that I don't fully grok is reference counting and python-side garbage collection. When we declare self keeps return alive, we mean that the thing an instance returns can't be destroyed until that object is destroyed. That's straightforward. Conceptually, it puts a floor on the reference counter. 

But what is the effect of the python object wrapping that reference? It *seems* that it doesn't participate in reference counts. Such that its existence doesn't prevent destruction of the underlying C++ object if the originating instance gets cleaned up (similarly, the destruction of `plant` object doesn't cause the underlying `MbP` object to be destroyed). That return value is no longer participating in the equation.

Based on all that, I think the defect was not in the comment, but in the keep_alive declaration. I think it truly is the case that return must keep self alive.

@EricCousineau-TRI should weigh in as well.
___
*[`bindings/pydrake/systems/framework_py_semantics.cc` line 646 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMfxKwO9TCY5iYwtxK3:-NMfxKwO9TCY5iYwtxK4:b-6qyawm) ([raw file](https://github.com/RobotLocomotion/drake/blob/a82617361f2d0230147fec1df44ab5f1bd732873/bindings/pydrake/systems/framework_py_semantics.cc#L646)):*
> ```cpp
>           doc.DiagramBuilder.ExportOutput.doc)
>       .def("Build", &DiagramBuilder<T>::Build,
>           // Keep alive, ownership (tr.): `self` keeps `return` alive.
> ```

See my other note.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMfz7bfE9pWcStT_h7L:b-6pk0d3) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*
> Pretend that Builder::plant() didn't already keep the builder alive ... pointing to a destructed object.

That pretending is all a counter-factual.  The plant _does_ keep the builder alive, and that's a fundamental part of the whole reasoning why we have this annotation and why it's correct.  Waving it away in a hypothetical doesn't make sense.  The keep-alive annotations for the builder all work as a team, none of them stand independently.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMgEFhd7u8aVqDKSvtL:b-7kvood) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

> Pretend that Builder::plant() didn't already keep the builder alive ... pointing to a destructed object.

That pretending is all a counter-factual.  The plant _does_ keep the builder alive, and that's a fundamental part of the whole reasoning why we have this annotation and why it's correct.  Waving it away in a hypothetical doesn't make sense.  The keep-alive annotations for the builder all work as a team, none of them stand independently.


</blockquote></details>

I'm afraid that my rhetoric there didn't work. Of all the things I'd hope you respond to, that would be the *one* thing that deserves the least attention and discussion. I wish I hadn't written it. I'd hoped you'd point out the flaw in the *other* points.

I've thought about it and I think I finally see it.

1. Objects resulting from functions tagged as "internal references" (e.g., `plant` in the example) have two things: an `MbP` pointer and a reference to `builder` (keeping it alive for as long as `plant`).
2. As long as we don't call `builder.Build()`, `builder` maintains 100% control over whether the `MbP` is destroyed or not.
    - `plant` can never destroy the pointer it has.
    - However, as long as `plant` keeps `builder` alive, `MbP` is valid.
3. As soon as we call `builder.Build()`, `builder` no longer has any *direct* control over `MbP`'s lifespan. (If, through some machinations, we were to delete `builder` despite reference counting, `MbP` would remain valid.)
   - However, it knows who *does* have such control: the `diagram` it returned.
    - So, it keeps the `diagram` alive, knowing that implies `MbP` will stay alive.
    - This is the "transitive" term used in the text. `plant` keeps `builder` alive. `builder` keeps `diagram` alive, so `plant` always has a valid `MbP`.
    - The chain would get longer if `diagram` could surrender the `MbP` to yet another party.

So, `plant()` should be "return keeps self alive" and `Build` should be "self keeps return alive". And the documentation on the bindings of `DiagramBuilder::Build` was defective.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMhvcE_4CQcSUnfcxCE:b-shhy56) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

I'm afraid that my rhetoric there didn't work. Of all the things I'd hope you respond to, that would be the *one* thing that deserves the least attention and discussion. I wish I hadn't written it. I'd hoped you'd point out the flaw in the *other* points.

I've thought about it and I think I finally see it.

1. Objects resulting from functions tagged as "internal references" (e.g., `plant` in the example) have two things: an `MbP` pointer and a reference to `builder` (keeping it alive for as long as `plant`).
2. As long as we don't call `builder.Build()`, `builder` maintains 100% control over whether the `MbP` is destroyed or not.
    - `plant` can never destroy the pointer it has.
    - However, as long as `plant` keeps `builder` alive, `MbP` is valid.
3. As soon as we call `builder.Build()`, `builder` no longer has any *direct* control over `MbP`'s lifespan. (If, through some machinations, we were to delete `builder` despite reference counting, `MbP` would remain valid.)
   - However, it knows who *does* have such control: the `diagram` it returned.
    - So, it keeps the `diagram` alive, knowing that implies `MbP` will stay alive.
    - This is the "transitive" term used in the text. `plant` keeps `builder` alive. `builder` keeps `diagram` alive, so `plant` always has a valid `MbP`.
    - The chain would get longer if `diagram` could surrender the `MbP` to yet another party.

So, `plant()` should be "return keeps self alive" and `Build` should be "self keeps return alive". And the documentation on the bindings of `DiagramBuilder::Build` was defective.
</blockquote></details>

I think these are all great points; I've got *possible* answers, but I haven't written real code to vet any of this so my answers would only be possibly incorrect noise.
Given this discussion is now resolved, not worth blocking this PR just yet - I've filed #18656 for tracking purposes. Thanks!

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-:-NMiOACw1RSTKLLqhXQj:b223nro)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMiXgWN0Fg79oiqMBzk:b-pl8gbl) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

I think these are all great points; I've got *possible* answers, but I haven't written real code to vet any of this so my answers would only be possibly incorrect noise.
Given this discussion is now resolved, not worth blocking this PR just yet - I've filed #18656 for tracking purposes. Thanks!

</blockquote></details>

Yup, that's the correct explanation.

To me, the bigger problem here is that these chains of reasoning get to be pretty arcane, and also it's difficult to have the keep_alive be precise enough to ensure that actual garbage does get collected.  At some point, I think we need to invest in switching our bindings to use shared_ptr so that C++ and Python easily and obviously agree on reference counts.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`bindings/pydrake/planning/planning_py_robot_diagram.cc` line 93 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-NMel8QA3FIkOQV7ZdeH:-NMind8LD4gm_7tWwYJa:bxjc8ft) ([raw file](https://github.com/RobotLocomotion/drake/blob/1680ac23f9097422b9844cc93b8ec092d1b6c11f/bindings/pydrake/planning/planning_py_robot_diagram.cc#L93)):*

Yup, that's the correct explanation.

To me, the bigger problem here is that these chains of reasoning get to be pretty arcane, and also it's difficult to have the keep_alive be precise enough to ensure that actual garbage does get collected.  At some point, I think we need to invest in switching our bindings to use shared_ptr so that C++ and Python easily and obviously agree on reference counts.
</blockquote></details>

Agreed. I'm a bit mystified on how "switching to use shared_ptr" would work in practice, but look forward to seeing it.

xuchenhan-tri (commented):
:lgtm: 

> Reviewed 3 of 7 files at r1, 2 of 4 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18650#-:-NMjC2HoAJWtFVSpl8-m:bg7igkv)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri)

Issue #18653, opened 2023-01-25
[py value] PyObjectValue::SetFrom is incorrect


EricCousineau-TRI:
### What happened?

I am writing Python code, and have simulations reset as follows:

```py
context_init = context.Clone()
...
run_count = 0
while True:
   if run_count > 0:
      context.SetTimeStateAndParametersFrom(context_init)
   run_count += 1
   simulator.AdvancedTo(some_time)
```

However, when I run this more than twice, I am finding that some state isn't getting reset properly.

After investigation, I believe it is at least because `PyObjectValue::SetFrom` is not implemented:
https://github.com/RobotLocomotion/drake/blob/dc635f1e5d3ea5f7586c47adbc1004e6f7a2099f/bindings/pydrake/common/value_py.cc#L16-L28

Will file fix.

### Version

dc635f1e5d3ea5f7586c47adbc1004e6f7a2099f

EricCousineau-TRI:
Simple workaround is to simply call `context.SetTimeStateAndParametersFrom(context_init.Clone())`

PR #18655, opened 2023-01-26
[py value] Ensure Value[object].SetFrom() deep copies the value

EricCousineau-TRI:
Resovles #18653

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)

### DIFF
--- a/bindings/pydrake/common/cpp_param_pybind.cc
+++ b/bindings/pydrake/common/cpp_param_pybind.cc
@@ -31,6 +31,13 @@ Object& Object::operator=(Object&& other) {
   other.ptr_ = nullptr;
   return *this;
 }
+
+Object Object::Clone() const {
+  py::object py_copy = py::module::import("copy").attr("deepcopy");
+  py::object copied = py_copy(to_pyobject<py::object>());
+  return from_pyobject(copied);
+}
+
 void Object::inc_ref() {
   py::handle(ptr_).inc_ref();
 }
--- a/bindings/pydrake/common/cpp_param_pybind.h
+++ b/bindings/pydrake/common/cpp_param_pybind.h
@@ -62,6 +62,9 @@ class Object {
     return Object(h.ptr());
   }
 
+  /// Provides a deep copy of the referred-to object.
+  Object Clone() const;
+
  private:
   // Increments reference count. See `py::handle::inc_ref()` for more details.
   void inc_ref();
--- a/bindings/pydrake/common/test/value_test.py
+++ b/bindings/pydrake/common/test/value_test.py
@@ -61,27 +61,69 @@ class TestValue(unittest.TestCase):
         # Test docstring.
         self.assertTrue("unique_ptr" in value.set_value.__doc__)
 
-    def test_abstract_value_py_object(self):
-        expected = {"x": 10}
-        value = Value[object](expected)
-        # Value is by reference, *not* by copy.
-        self.assertTrue(value.get_value() is expected)
-        # Update mutable version.
-        value.get_mutable_value()["y"] = 30
-        self.assertEqual(value.get_value(), expected)
-        # Cloning the value should perform a deep copy of the Python object.
-        value_clone = copy.deepcopy(value)
-        self.assertEqual(value_clone.get_value(), expected)
-        self.assertTrue(value_clone.get_value() is not expected)
-        # Using `set_value` on the original value changes object reference.
-        expected_new = {"a": 20}
-        value.set_value(expected_new)
-        self.assertEqual(value.get_value(), expected_new)
-        self.assertTrue(value.get_value() is not expected)
-
     def assert_equal_but_not_aliased(self, a, b):
         self.assertEqual(a, b)
         self.assertIsNot(a, b)
+
+    def test_assert_equal_but_not_aliased(self):
+        """
+        This is a higher-level test for `assert_equal_but_not_aliased`, not for
+        the device under test `Value[]`.
+        """
+        a = {"x": 10}
+        b = copy.deepcopy(a)
+        self.assert_equal_but_not_aliased(a, b)
+        # Confirm that mutation will not affect the other copy.
+        b["x"] = 100
+        self.assertNotEqual(a, b)
+
+        # Negatives.
+        # - Self-identity should fail.
+        with self.assertRaises(AssertionError):
+            self.assert_equal_but_not_aliased(a, a)
+        # - Inequality should fail.
+        with self.assertRaises(AssertionError):
+            self.assert_equal_but_not_aliased(a, {"x": -10})
+
+    def test_abstract_value_py_object(self):
+        # The Value constructor operates by taking a reference, not by copying.
+        initial_dict = {"x": 10}
+        value = Value[object](initial_dict)
+        self.assertIs(value.get_value(), initial_dict)
+        self.assertIs(value.get_mutable_value(), initial_dict)
+
+        # Therefore, mutating via the accessor will also mutate the original.
+        value.get_mutable_value()["y"] = 20
+        self.assertEqual(value.get_value(), {"x": 10, "y": 20})
+        self.assertEqual(initial_dict, {"x": 10, "y": 20})
+
+        # Cloning the value should perform a deep copy of the Python object.
+        cloned_value = copy.deepcopy(value)
+        cloned_dict = cloned_value.get_value()
+        self.assert_equal_but_not_aliased(cloned_dict, initial_dict)
+
+        # For both the original Value[] and the cloned Value[], mutating via
+        # the accessor should not influence the other one.
+        value.get_mutable_value()["y"] = 55
+        cloned_value.get_mutable_value()["y"] = 99
+        self.assertEqual(value.get_value(), {"x": 10, "y": 55})
+        self.assertEqual(initial_dict, {"x": 10, "y": 55})
+        self.assertEqual(cloned_value.get_value(), {"x": 10, "y": 99})
+        self.assertEqual(cloned_dict, {"x": 10, "y": 99})
+
+        # Using `set_value` on the original Value[] should update the object
+        # reference to the new value. The clone is unchanged.
+        new_dict = {"a": 444}
+        value.set_value(new_dict)
+        self.assertEqual(new_dict, {"a": 444})
+        self.assertIs(value.get_value(), new_dict)
+        self.assertIs(value.get_mutable_value(), new_dict)
+        self.assertEqual(cloned_value.get_value(), {"x": 10, "y": 99})
+
+        # Using `SetFrom` should perform a deep copy, not just alias (#18653).
+        value.SetFrom(cloned_value)
+        self.assertIs(cloned_value.get_value(), cloned_dict)
+        self.assert_equal_but_not_aliased(value.get_value(), cloned_dict)
 
     def test_abstract_value_py_list(self):
         value_cls = Value[List[CustomType]]
--- a/bindings/pydrake/common/value_py.cc
+++ b/bindings/pydrake/common/value_py.cc
@@ -15,15 +15,21 @@ namespace {
 
 // Local specialization of C++ implementation for `Value[object]`
 // instantiation.
+// TODO(eric.cousineau): Per discussion in #18655, should consider enforcing
+// value semantics (always deepcopy, even on `set_value()`).
 class PyObjectValue : public drake::Value<Object> {
  public:
   using Base = Value<Object>;
   using Base::Base;
+
   // Override `Clone()` to perform a deep copy on the object.
   std::unique_ptr<AbstractValue> Clone() const override {
-    py::object py_copy = py::module::import("copy").attr("deepcopy");
-    py::object copied = py_copy(get_value().to_pyobject<py::object>());
-    return std::make_unique<PyObjectValue>(Object::from_pyobject(copied));
+    return std::make_unique<PyObjectValue>(get_value().Clone());
+  }
+
+  // Override `SetFrom()` to perform a deep copy on the object.
+  void SetFrom(const AbstractValue& other) override {
+    get_mutable_value() = other.get_value<Object>().Clone();
   }
 };
 
--- a/bindings/pydrake/systems/test/custom_test.py
+++ b/bindings/pydrake/systems/test/custom_test.py
@@ -949,3 +949,62 @@ class TestCustom(unittest.TestCase):
             system.CalcOutput(context, output)
             value = output.get_data(0)
             self.assertEqual(value.get_value(), expected_output_value)
+
+    def assert_equal_but_not_aliased(self, a, b):
+        self.assertEqual(a, b)
+        self.assertIsNot(a, b)
+
+    def test_context_and_value_object_set_from(self):
+        """
+        Shows how `Value[object]` behaves in a context, especially in
+        connection to `Context.SetTimeStateAndParametersFrom()`.
+
+        Helps to highlight failure mode illustrated in #18653.
+        """
+        arbitrary_object = {"key": "value"}
+
+        class SystemWithCacheAndState(LeafSystem):
+            def __init__(self):
+                super().__init__()
+                model_value = AbstractValue.Make(arbitrary_object)
+                self.state_index = self.DeclareAbstractState(model_value)
+
+                def calc_cache_noop(context, abstract_value):
+                    pass
+
+                self.cache_entry = self.DeclareCacheEntry(
+                    description="test",
+                    value_producer=ValueProducer(
+                        allocate=model_value.Clone,
+                        calc=calc_cache_noop,
+                    ),
+                )
+
+            def eval_state(self, context):
+                return context.get_abstract_state(self.state_index).get_value()
+
+        system = SystemWithCacheAndState()
+        context = system.CreateDefaultContext()
+        context_init = context.Clone()
+
+        cache = system.cache_entry.Eval(context)
+        self.assert_equal_but_not_aliased(cache, arbitrary_object)
+        state = system.eval_state(context)
+        self.assert_equal_but_not_aliased(state, arbitrary_object)
+
+        def check_set_from():
+            nonlocal cache, state
+            context.SetTimeStateAndParametersFrom(context_init)
+            # Ensure that we have cloned the object.
+            old_state = state
+            state = system.eval_state(context)
+            self.assert_equal_but_not_aliased(state, old_state)
+            # Warning: Cache objects are not cloned!
+            old_cache = cache
+            cache = system.cache_entry.Eval(context)
+            self.assertIs(cache, old_cache)
+
+        # Check twice. Per #18653, if we did not implement
+        # Value[object].SetFrom() correctly, this would fail the second time.
+        check_set_from()
+        check_set_from()


EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfN4pG5WPTvYCfbnsl:-NMfN4pG5WPTvYCfbnsm:bd5u8s1) (no related file):*
Working: Shows failures illustrated in issue. Applying fix.

___

EricCousineau-TRI (commented):
+@jwnimmer-tri for feature and +@xuchenhan-tri for platform, please!

This fixes behavior that I believe is defective, but changes nothing else.
I believe this should have high priority as this has unexpected consequences, especially when writing controllers with state in Python.

FWIW, I briefly tried making `Object` non-copyable, but could not quickly make it work. I am also concerned about the possibility of behavior change that isn't concretely defective (`set_value()` aliasing).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-:-NMfRKOi1DUYJVv35Bky:b-pwuhy1)* status: LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform), missing label for release notes (waiting on @jwnimmer-tri and @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfN4pG5WPTvYCfbnsl:-NMfRJj78h-oqtrUC5gS:b-896fix) (no related file):*

Working: Shows failures illustrated in issue. Applying fix.
</blockquote></details>

Done.

___

jwnimmer-tri (commented):
Checkpoint.

> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-:-NMfoEHDBXfWb8NEdNvr:by96gdd)* status: 4 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @EricCousineau-TRI and @xuchenhan-tri)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfjwhH6EFZmNNNUtN8):*
typo



_Suggestion:_
```
copy.deepcopy
```
___
*[`bindings/pydrake/common/cpp_param_pybind.h` line 65 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfj73B2KT-abSNnnXF:-NMfj73DA73v9mdTGLBs:b-xr4yuq) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/cpp_param_pybind.h#L65)):*
> ```objc
>   }
> 
>   Object Clone() const;
> ```

nit It's important to specify here that this clone is a deep copy (e.g., not a shallow copy nor a simple reference count increment).
___
*[`bindings/pydrake/common/value_py.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfmWXm2F2EhS6UnGUg:-NMfmWXnAbU5gJ3yQkyq:b-7wyiyp) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L21)):*
> ```cpp
>  public:
>   using Base = Value<Object>;
>   using Base::Base;
> ```

BTW The aliasing in the inherited constructor here isn't doing us any favors either.  (We definitely shouldn't touch that in this PR, and maybe not at all.)

In C++ we do sometimes allow moving into a `Value<>` to construct it, but we generally discourage it, guiding most uses case to _always_ copy when the user provides a value into the framework.  (The framework API disallows `unique_ptr` when giving values. Only the low-level `Value<>` API allows unique_ptr.)  We definitely never give a Value a `shared_ptr` and expect the C++ user to retain a reference count for later; the most the user can do in C++ is retain a borrowed raw pointer from the `unique_ptr` they give us.  It's extremely hazardous for users to keep handles to borrowed pointers into context values (they might mutate them without proper cache control!), so if we could go back in time, we would have been better off eating a clone during construction here.  At this point it might be too risky of a change, though.

The fundamental problem here is that `Value<T>` is only valid for types with _value_ semantics for `T`, but our `pydrake::Object` class has reference semantics!  OMG yikes what a time bomb!  We've worked around that in this PR by overriding `SetFrom` so that things like the Simulator won't be broken, but the fundamental problem is that `Object` **is not a valid type-erased type** in the first place, because it behaves like a reference.

It's probably OK to fix SetFrom now since this is obviously wrong and the surgical fix is low-risk, but I worry that even `set_value()` is completely broken for all Python users.  It's supposed to deepcopy the object (or at minimum, shallow copy it) but all it does is increment a refcount.  Blarg!
___
*[`bindings/pydrake/common/value_py.cc` line 28 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfkNUB7JDPnryjZkVc:-NMfkNUB7JDPnryjZkVd:bwfnwnr) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L28)):*
> ```cpp
>   }
> 
>   // Override `SetFrom()` to perform a deep copy on the object (#18653).
> ```

nit Broadly speaking, we don't leave breadcumbs on lines of code where we fix a bug; that would accumulate cruft very quickly. Unless there the code is acutely unusual, I don't think we need extra remarks, and when we _do_ need remarks we should explain the remarkable thing inline instead of a number.  `Value<>` setting should obviously have copy semantics, so I don't think this override is unusual at all, so it doesn't need any crumbs.



_Suggestion:_
```cpp
// Override `SetFrom()` to perform a deep copy on the object.
```

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfjwhH6EFZmNNNUtN8):*

typo
</blockquote></details>

Done, thanks!
___
*[`bindings/pydrake/common/cpp_param_pybind.h` line 65 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfj73B2KT-abSNnnXF:-NMhyYLB3GXKApBwX__R:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/cpp_param_pybind.h#L65)):*

nit It's important to specify here that this clone is a deep copy (e.g., not a shallow copy nor a simple reference count increment).
</blockquote></details>

Done.
___
*[`bindings/pydrake/common/value_py.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfmWXm2F2EhS6UnGUg:-NMhyZ4Z1lTTMdjkxUFG:b-8n0s5u) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L21)):*
OK Agreed on the perspective of true consistency.

I'm concerned about possible performance implications (e.g. using `Value[object](my_big_torch_tensor).set_value(my_other_torch_tensor)`) and the change in behavior.
However, yeah, we can definitely change it if/when a user (us or others) run into this sticking point.

> OMG yikes what a time bomb!

Indeed! But kinda crazy that I only just ran into this after having introduced this stuff in ~5 years ago in 51ff3e0fe04b6b :sweat_smile: 
___
*[`bindings/pydrake/common/value_py.cc` line 28 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfkNUB7JDPnryjZkVc:-NMi-Eeg-bdL7zhqJZMW:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L28)):*

nit Broadly speaking, we don't leave breadcumbs on lines of code where we fix a bug; that would accumulate cruft very quickly. Unless there the code is acutely unusual, I don't think we need extra remarks, and when we _do_ need remarks we should explain the remarkable thing inline instead of a number.  `Value<>` setting should obviously have copy semantics, so I don't think this override is unusual at all, so it doesn't need any crumbs.
</blockquote></details>

Done.

jwnimmer-tri (commented):
Implementation looks good.  I need to circle back on the tests once I have a bigger block of free time this morning.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-:-NMiZKAU-kkG-Kq-IIpI:b-3apsye)* status: LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`bindings/pydrake/common/value_py.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfmWXm2F2EhS6UnGUg:-NMiYOeDABHQlDqnEvQ6:br2w22h) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L21)):*

OK Agreed on the perspective of true consistency.

I'm concerned about possible performance implications (e.g. using `Value[object](my_big_torch_tensor).set_value(my_other_torch_tensor)`) and the change in behavior.
However, yeah, we can definitely change it if/when a user (us or others) run into this sticking point.

> OMG yikes what a time bomb!

Indeed! But kinda crazy that I only just ran into this after having introduced this stuff in ~5 years ago in 51ff3e0fe04b6b :sweat_smile: 
</blockquote></details>

Maybe we could have `set_value(yuge, copy=False)` and put giant scary documentation that you must not retain a reference to `yuge` in that case.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @EricCousineau-TRI and @xuchenhan-tri)

___
*[`bindings/pydrake/common/test/value_test.py` line 64 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjvQZN70U1tPHBRXxa:-NMjvQZN70U1tPHBRXxb:b-6vexr3) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/common/test/value_test.py#L64)):*
> ```Python
>         self.assertTrue("unique_ptr" in value.set_value.__doc__)
> 
>     def test_abstract_value_py_object(self):
> ```

nit I had a difficult time following this to prove how everything works.

Counter-proposal:
https://github.com/EricCousineau-TRI/drake/pull/16

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`bindings/pydrake/common/test/value_test.py` line 64 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjvQZN70U1tPHBRXxa:-NMjvyEx8ygZ7_kloAyN:b-iz91t7) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/common/test/value_test.py#L64)):*

nit I had a difficult time following this to prove how everything works.

Counter-proposal:
https://github.com/EricCousineau-TRI/drake/pull/16
</blockquote></details>

Working: I'll take some of what you have, and adjust some other parts.

jwnimmer-tri (commented):
:lgtm: feature.

> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-:-NMjwH6nC15EdRVA8LHa:b-n37teh)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI and @xuchenhan-tri)

___
*[`bindings/pydrake/systems/test/custom_test.py` line 308 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjv_GUAF9UBlRO5Rbv:-NMjv_GUAF9UBlRO5Rbw:b-5oddat) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/systems/test/custom_test.py#L308)):*
> ```Python
>         Context.SetTimeStateAndParametersFrom.
>         """
>         pass
> ```

This seems like the test is missing some content.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI and @xuchenhan-tri)

xuchenhan-tri (commented):
> Reviewed 1 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/common/value_py.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMfmWXm2F2EhS6UnGUg:-NMjzX1B3BkY8Kmck8R_:b-415h9) ([raw file](https://github.com/RobotLocomotion/drake/blob/9e199774bd18ae53fc18664fd94abfb688c32e88/bindings/pydrake/common/value_py.cc#L21)):*

Maybe we could have `set_value(yuge, copy=False)` and put giant scary documentation that you must not retain a reference to `yuge` in that case.
</blockquote></details>

OK Aye, but will hold off for now on that front.
___
*[`bindings/pydrake/common/test/value_test.py` line 64 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjvQZN70U1tPHBRXxa:-NMjzG-P7DUF6uYYBE__:b-ei29hl) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/common/test/value_test.py#L64)):*

Working: I'll take some of what you have, and adjust some other parts.
</blockquote></details>

Done.

I wasn't immediately sure the value gain from checking both identity and value/mutation, as well as testing identity against `.get_value()` and `.get_mutable_value()`.
I added testing above to that effect, and then minimized the portions here.
If you think this is a step backwards, I can (partially) revert that commit.
___
*[`bindings/pydrake/systems/test/custom_test.py` line 308 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjv_GUAF9UBlRO5Rbv:-NMjzVL7BGhg7DvfnXyl:b-qz5atz) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/systems/test/custom_test.py#L308)):*

This seems like the test is missing some content.
</blockquote></details>

Done. Derp!

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/test/value_test.py` line 64 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMjvQZN70U1tPHBRXxa:-NMk00bB4_pkH6Sil3jS:b-dl7ugd) ([raw file](https://github.com/RobotLocomotion/drake/blob/ab8fc975c3ca34af02c236a08719090406ded4fd/bindings/pydrake/common/test/value_test.py#L64)):*

Done.

I wasn't immediately sure the value gain from checking both identity and value/mutation, as well as testing identity against `.get_value()` and `.get_mutable_value()`.
I added testing above to that effect, and then minimized the portions here.
If you think this is a step backwards, I can (partially) revert that commit.
</blockquote></details>

OK mostly fine, just one nit below.
___
*[`bindings/pydrake/common/test/value_test.py` line 97 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMk-TsK7KsE_cvc0C-_:-NMk-TsK7KsE_cvc0C-a:btnnsan) ([raw file](https://github.com/RobotLocomotion/drake/blob/ea86f514c6154558e8acd3c510de3519e2c2763f/bindings/pydrake/common/test/value_test.py#L97)):*
> ```Python
>         self.assertIs(value.get_value(), new_dict)
>         self.assertIs(value.get_mutable_value(), new_dict)
>         self.assertEqual(cloned_value.get_value(), {"x": 10, "y": 99})
> ```

nit This line was the test that demonstrated "The clone is unchanged" from the comment paragraph above.  I think it's important to double-check that the clone is unchanged upon a `set_value`.

We don't have any evidence that `set_value` is modifying the same things as `get_mutable_value()` uses, so we can't rely on the prior checks that the `get_value()` objects are at distinct addresses to conclude that the clone is unaffected.
___
*[`bindings/pydrake/common/test/value_test.py` line 68 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMk0REf33q0_usjgCuZ:-NMk0REg9yrWhz2cxEpj:bbmgxp7) ([raw file](https://github.com/RobotLocomotion/drake/blob/9a80f9ca1fd20bd628fca6d4110516e98e5a2594/bindings/pydrake/common/test/value_test.py#L68)):*
> ```Python
>         self.assertIsNot(a, b)
> 
>     def test_assert_equal_but_not_aliased(self):
> ```

nit It's probably worth a little docstring here that it's a higher-order test for the function immediately prior, not a test of the actual device under test (`Value[]`).

xuchenhan-tri (commented):
:lgtm: 

> Reviewed 1 of 2 files at r5, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-:-NMkXqi8-tvdVHgsGYvZ:bg7igkv)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18655)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/test/value_test.py` line 97 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18655#-NMk-TsK7KsE_cvc0C-_:-NMn8mX5CBqQ7bYBi0gd:b-m6pg52) ([raw file](https://github.com/RobotLocomotion/drake/blob/ea86f514c6154558e8acd3c510de3519e2c2763f/bindings/pydrake/common/test/value_test.py#L97)):*

FWIW I was ok with your trimmed down version, just that one line I marked (`self.assertEqual(cloned_value.get_value(), {"x": 10, "y": 99})`) was the only one that seemed like a meaningful loss.
</blockquote></details>

OK Aye, gotcha! Yeah, I had struggled to place that back in w/ my trimming (maybe mind fuzz at the time?), so it was easier to take your suggestions. Sorry about that!

PR #18660, opened 2023-01-26
Notes on the formulation of ball constraints

amcastro-tri:
A few thoughts on the formulation of ball constraints, per conversation with @sherm1.
We should investigate if using the same position (say the point midway of P and Q) leads to better physics (validity of Newton's 3rd law), energy conservation and/or momentum conservation.

@sherm1, you mentioned to me you know of a reference on this?

cc'ing @joemasterjohn

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660)

### DIFF
--- a/multibody/plant/sap_driver.cc
+++ b/multibody/plant/sap_driver.cc
@@ -512,6 +512,13 @@ void SapDriver<T>::AddBallConstraints(
     manager().internal_tree().CalcJacobianTranslationalVelocity(
         context, JacobianWrtVariable::kV, body_B.body_frame(), frame_W, p_WQ,
         frame_W, frame_W, &Jv_WBq_W);
+
+    // TODO(amcastro-tri): Per conversation with Sherm. Consider using the same
+    // point here, affixed to each corresponding body. That is, we could for
+    // instance use the middle point M and write:
+    //   p_WM = 0.5 * (p_WP + p_WQ)
+    //   Compute Jacobian Jv_AmBm_W instead.
+    // Consider conservation of energy and momentum.
     Jv_ApBq_W = (Jv_WBq_W - Jv_WAp_W);
 
     // TODO(amcastro-tri): consider exposing this parameter.


sherm1 (commented):
This kind of detail is not usually covered in published literature. The closest thing to a reference I have is Simbody's implementation of this constraint. Here is the derivation in the code: https://github.com/simbody/simbody/blob/master/Simbody/src/ConstraintImpl.h#L1870
In this case we took the easy route of picking one of the two points rather than the midpoint, but I think the midpoint idea is better.

This was likely worked out long ago between me and @mitiguy at Stanford, so Paul might have some thoughts on the topic.

The TODO :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660#-:-NMjBBWx-rPVjLBvo5qs:b-7tt0zy)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri)

mitiguy (commented):
@amcastro-tri  and @sherm1 
I have been investigating.  From a preliminary simulation of a 4-bar linkage with constraints, using two "half-way" points gets rid of the violation with the "strong form" of  Newton's III law.  However, due to how I implemented the loop constraint, there is considerable amount of work done by the "half-way" points whereas there is no work done when half-way points are not used.    It may be easier to pick up this conversation during the on-site.  It is not clear that one is radically better than the other.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660#-:-NNJvm3X4kiwA0PUe0NV:bnmtr90)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri)

sherm1 (commented):
Is that because you're including a velocity for the half-way point (I mean the time derivative of the point on the bodies)? I think for the constraint to be non-working the contact point has to be considered as a fixed station, as though we always planned on using that point. Then the enforced acceleration-level constraints is instantaneously no-power so integrates to no-work. Or is it worse than that? 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660#-:-NNK5Ab--kXsiy4__Lnu:br6i36n)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri)

mitiguy (commented):
I am considering the half-way point to be a station.  Herein, line L is the line that connect the "original" points that are associated with the constraint.  There is no work done by the constraint forces that are parallel to line L (whether one uses the original or half-way points).  However, due to the ω x r term, there is work done perpendicular to line L by the half-way points whereas there is no work done perpendicular to line L by the original points (since I enforce the constraint on the "original points"). 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660#-:-NNK7B6t8Mk-eFDNMidF:b-j5zx1t)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri)

sherm1 (commented):
Hmmm ... that doesn't sound right. Shouldn't the constraint be enforced at the chosen contact point?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18660#-:-NNKAqYmBbYBWO5Y3v6W:b-urixgw)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri)

amcastro-tri:
This is now resolved with the merging for #19940.
@joemasterjohn did an excellent job coming up with a formulation that is:
  1. Frame invariant
  2. Satisfies Newton's 3rd law
  3. Conserves angular momentum

Moreover, the implementation comes with a beautifully documented formulation.

PR #18666, opened 2023-01-26
RenameModelInstance: feature, or footgun?

rpoyner-tri:
Work on the parser has shown that having model naming joined to parsing is a source of pain, tedium, and disappointment. While the parser will still need to be upgraded to give better naming affordances, users may want different names after the fact, for any number of reasons (preference, presentation displays, etc.). Rather than try to satisfy all constituencies through parser features, this patch allows end users to override model names at any time *before* Finalize().

The embodiment here is a complete implementation (with no examples :) ), but is mostly meant to spur discussion of the feature.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_scene_graph.cc
+++ b/bindings/pydrake/geometry/geometry_py_scene_graph.cc
@@ -97,7 +97,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py_rvp::reference_internal, py::arg("frame_id"),
             cls_doc.GetOwningSourceName.doc_1args_frame_id)
         .def("GetName",
-            overload_cast_explicit<const std::string&, FrameId>(
+            overload_cast_explicit<std::string, FrameId>(
                 &Class::GetName),
             py_rvp::reference_internal, py::arg("frame_id"),
             cls_doc.GetName.doc_1args_frame_id)
@@ -127,7 +127,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def("GetFrameId", &Class::GetFrameId, py::arg("geometry_id"),
             cls_doc.GetFrameId.doc)
         .def("GetName",
-            overload_cast_explicit<const std::string&, GeometryId>(
+            overload_cast_explicit<std::string, GeometryId>(
                 &Class::GetName),
             py_rvp::reference_internal, py::arg("geometry_id"),
             cls_doc.GetName.doc_1args_geometry_id)
--- a/bindings/pydrake/geometry/geometry_py_scene_graph.cc
+++ b/bindings/pydrake/geometry/geometry_py_scene_graph.cc
@@ -97,8 +97,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py_rvp::reference_internal, py::arg("frame_id"),
             cls_doc.GetOwningSourceName.doc_1args_frame_id)
         .def("GetName",
-            overload_cast_explicit<std::string, FrameId>(
-                &Class::GetName),
+            overload_cast_explicit<std::string, FrameId>(&Class::GetName),
             py_rvp::reference_internal, py::arg("frame_id"),
             cls_doc.GetName.doc_1args_frame_id)
         .def("GetFrameGroup", &Class::GetFrameGroup, py::arg("frame_id"),
@@ -127,8 +126,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def("GetFrameId", &Class::GetFrameId, py::arg("geometry_id"),
             cls_doc.GetFrameId.doc)
         .def("GetName",
-            overload_cast_explicit<std::string, GeometryId>(
-                &Class::GetName),
+            overload_cast_explicit<std::string, GeometryId>(&Class::GetName),
             py_rvp::reference_internal, py::arg("geometry_id"),
             cls_doc.GetName.doc_1args_geometry_id)
         .def("GetShape", &Class::GetShape, py_rvp::reference_internal,
--- a/bindings/pydrake/multibody/test/inverse_kinematics_test.py
+++ b/bindings/pydrake/multibody/test/inverse_kinematics_test.py
@@ -428,11 +428,11 @@ class TestInverseKinematics(unittest.TestCase):
         frame_id1 = inspector.GetGeometryIdByName(
             self.plant.GetBodyFrameIdOrThrow(
                 self.plant.GetBodyByName("body1").index()),
-            pydrake.geometry.Role.kProximity, "two_bodies::body1_collision")
+            pydrake.geometry.Role.kProximity, "body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             self.plant.GetBodyFrameIdOrThrow(
                 self.plant.GetBodyByName("body2").index()),
-            pydrake.geometry.Role.kProximity, "two_bodies::body2_collision")
+            pydrake.geometry.Role.kProximity, "body2_collision")
         ik.AddDistanceConstraint(
             geometry_pair=(frame_id1, frame_id2),
             distance_lower=distance_lower, distance_upper=distance_upper)
@@ -690,11 +690,11 @@ class TestConstraints(unittest.TestCase):
         frame_id1 = inspector.GetGeometryIdByName(
             self.plant_f.GetBodyFrameIdOrThrow(
                 self.plant_f.GetBodyByName("body1").index()),
-            pydrake.geometry.Role.kProximity, "two_bodies::body1_collision")
+            pydrake.geometry.Role.kProximity, "body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             self.plant_f.GetBodyFrameIdOrThrow(
                 self.plant_f.GetBodyByName("body2").index()),
-            pydrake.geometry.Role.kProximity, "two_bodies::body2_collision")
+            pydrake.geometry.Role.kProximity, "body2_collision")
         constraint = ik.DistanceConstraint(
             plant=variables.plant, geometry_pair=(frame_id1, frame_id2),
             distance_lower=0.1, distance_upper=0.2,
--- a/bindings/pydrake/multibody/test/inverse_kinematics_test.py
+++ b/bindings/pydrake/multibody/test/inverse_kinematics_test.py
@@ -428,11 +428,11 @@ class TestInverseKinematics(unittest.TestCase):
         frame_id1 = inspector.GetGeometryIdByName(
             self.plant.GetBodyFrameIdOrThrow(
                 self.plant.GetBodyByName("body1").index()),
-            pydrake.geometry.Role.kProximity, "body1_collision")
+            pydrake.geometry.Role.kProximity, "two_bodies::body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             self.plant.GetBodyFrameIdOrThrow(
                 self.plant.GetBodyByName("body2").index()),
-            pydrake.geometry.Role.kProximity, "body2_collision")
+            pydrake.geometry.Role.kProximity, "two_bodies::body2_collision")
         ik.AddDistanceConstraint(
             geometry_pair=(frame_id1, frame_id2),
             distance_lower=distance_lower, distance_upper=distance_upper)
@@ -690,11 +690,11 @@ class TestConstraints(unittest.TestCase):
         frame_id1 = inspector.GetGeometryIdByName(
             self.plant_f.GetBodyFrameIdOrThrow(
                 self.plant_f.GetBodyByName("body1").index()),
-            pydrake.geometry.Role.kProximity, "body1_collision")
+            pydrake.geometry.Role.kProximity, "two_bodies::body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             self.plant_f.GetBodyFrameIdOrThrow(
                 self.plant_f.GetBodyByName("body2").index()),
-            pydrake.geometry.Role.kProximity, "body2_collision")
+            pydrake.geometry.Role.kProximity, "two_bodies::body2_collision")
         constraint = ik.DistanceConstraint(
             plant=variables.plant, geometry_pair=(frame_id1, frame_id2),
             distance_lower=0.1, distance_upper=0.2,
--- a/bindings/pydrake/multibody/test/optimization_test.py
+++ b/bindings/pydrake/multibody/test/optimization_test.py
@@ -261,10 +261,10 @@ class TestContactWrenchFromForceInWorldFrameEvaluator(unittest.TestCase):
         inspector = scene_graph.model_inspector()
         frame_id1 = inspector.GetGeometryIdByName(
             plant.GetBodyFrameIdOrThrow(plant.GetBodyByName("body1").index()),
-            Role.kProximity, "body1_collision")
+            Role.kProximity, "two_bodies::body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             plant.GetBodyFrameIdOrThrow(plant.GetBodyByName("body2").index()),
-            Role.kProximity, "body2_collision")
+            Role.kProximity, "two_bodies::body2_collision")
         dut = ContactWrenchFromForceInWorldFrameEvaluator(
             plant=ad_plant,
             context=ad_plant_context,
--- a/bindings/pydrake/multibody/test/optimization_test.py
+++ b/bindings/pydrake/multibody/test/optimization_test.py
@@ -261,10 +261,10 @@ class TestContactWrenchFromForceInWorldFrameEvaluator(unittest.TestCase):
         inspector = scene_graph.model_inspector()
         frame_id1 = inspector.GetGeometryIdByName(
             plant.GetBodyFrameIdOrThrow(plant.GetBodyByName("body1").index()),
-            Role.kProximity, "two_bodies::body1_collision")
+            Role.kProximity, "body1_collision")
         frame_id2 = inspector.GetGeometryIdByName(
             plant.GetBodyFrameIdOrThrow(plant.GetBodyByName("body2").index()),
-            Role.kProximity, "two_bodies::body2_collision")
+            Role.kProximity, "body2_collision")
         dut = ContactWrenchFromForceInWorldFrameEvaluator(
             plant=ad_plant,
             context=ad_plant_context,
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -408,9 +408,9 @@ class TestMeldis(unittest.TestCase):
 
         # The geometry isn't registered until the load is processed.
         hydro_path = "/CONTACT_RESULTS/hydroelastic/" + \
-                     "body1.body1_collision+body2"
+                     "body1.two_bodies::body1_collision+body2"
         hydro_path2 = "/CONTACT_RESULTS/hydroelastic/" + \
-                      "body1.body1_collision2+body2"
+                      "body1.two_bodies::body1_collision2+body2"
         self.assertEqual(meshcat.HasPath(hydro_path), False)
         self.assertEqual(meshcat.HasPath(hydro_path2), False)
 
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -408,9 +408,9 @@ class TestMeldis(unittest.TestCase):
 
         # The geometry isn't registered until the load is processed.
         hydro_path = "/CONTACT_RESULTS/hydroelastic/" + \
-                     "body1.two_bodies::body1_collision+body2"
+                     "body1.body1_collision+body2"
         hydro_path2 = "/CONTACT_RESULTS/hydroelastic/" + \
-                      "body1.two_bodies::body1_collision2+body2"
+                      "body1.body1_collision2+body2"
         self.assertEqual(meshcat.HasPath(hydro_path), False)
         self.assertEqual(meshcat.HasPath(hydro_path2), False)
 
--- a/bindings/pydrake/visualization/test/triad_test.py
+++ b/bindings/pydrake/visualization/test/triad_test.py
@@ -53,8 +53,9 @@ class TestTriad(unittest.TestCase):
         for i, char in enumerate(("x", "y", "z")):
             geom_id = [x, y, z][i]
             frame_name = inspect.GetName(inspect.GetFrameId(geom_id))
-            self.assertEqual(frame_name, "Link2")
-            self.assertEqual(inspect.GetName(geom_id), f"foo {char}-axis")
+            self.assertEqual(frame_name, "acrobot::Link2")
+            self.assertEqual(inspect.GetName(geom_id),
+                             f"acrobot::foo {char}-axis")
             self.assertEqual(inspect.GetShape(geom_id).length(), 0.2)
             self.assertEqual(inspect.GetShape(geom_id).radius(), 0.001)
             props = inspect.GetIllustrationProperties(geom_id)
--- a/bindings/pydrake/visualization/test/triad_test.py
+++ b/bindings/pydrake/visualization/test/triad_test.py
@@ -53,7 +53,7 @@ class TestTriad(unittest.TestCase):
         for i, char in enumerate(("x", "y", "z")):
             geom_id = [x, y, z][i]
             frame_name = inspect.GetName(inspect.GetFrameId(geom_id))
-            self.assertEqual(frame_name, "acrobot::Link2")
+            self.assertEqual(frame_name, "Link2")
             self.assertEqual(inspect.GetName(geom_id), f"foo {char}-axis")
             self.assertEqual(inspect.GetShape(geom_id).length(), 0.2)
             self.assertEqual(inspect.GetShape(geom_id).radius(), 0.001)
--- a/examples/manipulation_station/manipulation_station.cc
+++ b/examples/manipulation_station/manipulation_station.cc
@@ -232,17 +232,14 @@ multibody::ModelInstanceIndex AddAndWeldModelFrom(
     const RigidTransform<double>& X_PC, MultibodyPlant<T>* plant) {
   DRAKE_THROW_UNLESS(!plant->HasModelInstanceNamed(model_name));
 
-  // Since we need to force the model name here, exploit the fact that model
-  // directives processing can do that.
+  // Since we need to force the model name here, exploit auto-renaming to avoid
+  // name collision errors, and then rename after parsing.
   multibody::Parser parser(plant);
-  multibody::parsing::ModelDirectives directives;
-  multibody::parsing::ModelDirective directive;
-  directive.add_model = multibody::parsing::AddModel{
-      model_url, model_name, {}, {}};
-  directives.directives.push_back(directive);
-  const auto models = ProcessModelDirectives(directives, &parser);
+  parser.SetAutoRenaming(true);
+  const auto models = parser.AddModelsFromUrl(model_url);
   DRAKE_THROW_UNLESS(models.size() == 1);
-  const multibody::ModelInstanceIndex new_model = models[0].model_instance;
+  plant->RenameModelInstance(models[0], model_name);
+  const multibody::ModelInstanceIndex new_model = models[0];
 
   const auto& child_frame = plant->GetFrameByName(child_frame_name, new_model);
   plant->WeldFrames(parent, child_frame, X_PC);
--- a/examples/planar_gripper/gripper_brick.cc
+++ b/examples/planar_gripper/gripper_brick.cc
@@ -86,7 +86,7 @@ GripperBrickHelper<T>::GripperBrickHelper() {
             plant_
                 ->GetBodyByName("finger" + std::to_string(i + 1) + "_tip_link")
                 .index()),
-        geometry::Role::kProximity, "tip_sphere_collision");
+        geometry::Role::kProximity, "planar_gripper::tip_sphere_collision");
   }
   const geometry::Shape& fingertip_shape =
       inspector.GetShape(finger_tip_sphere_geometry_ids_[0]);
@@ -105,7 +105,7 @@ GripperBrickHelper<T>::GripperBrickHelper() {
       inspector.GetShape(inspector.GetGeometryIdByName(
           plant_->GetBodyFrameIdOrThrow(
               plant_->GetBodyByName("brick_link").index()),
-          geometry::Role::kProximity, "box_collision"));
+          geometry::Role::kProximity, "brick::box_collision"));
   brick_size_ = dynamic_cast<const geometry::Box&>(brick_shape).size();
 
   for (int i = 0; i < 3; ++i) {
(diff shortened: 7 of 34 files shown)


rpoyner-tri (commented):
@sherm1 @xuchenhan-tri @SeanCurtis-TRI @jwnimmer-tri @amcastro-tri FYI. I'm mostly interested in whether the idea is any good.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NMjxmBH2EUd7MOlFyFJ:bikj9jq)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

jwnimmer-tri:
There are a few places in MbP that take the model instance name and copy it elsewhere, e.g., `RegisterVisualGeometry`.  If we decide to allow for renaming, I think we'll need to chase all of those down and reflect the new names universally.

rpoyner-tri (commented):
Good point. The particular case of copying has a TODO screed suggesting we stop doing that. I haven't done an exhaustive search for othe problem areas.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NMk2gwC8DCgE1oFeU1j:b-jh3t1u)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

jwnimmer-tri:
In my own quick search, the only hazards were non-logging calls to `GetScopedName` inside `mbp.cc`.  For the multibody elements names we blast copies those names everywhere, but the model instance blast radius _seemed_ to be pretty small.

amcastro-tri (commented):
my 2 cnets. Overall, I am all in favor of allowing users to change their models. This seems like a feature if it solves someone's problem. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NNE7iCYEasH99QyrGBU:b-50pczv)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NSMtAXP4cg8_UoHYhJ4:-NSMtAXQ5K-nH40l2ekX:b-d7226i) (no related file):*
Work in Anzu has added evidence that this feature is maybe useful. Mentioned via TODO in Anzu PR#10342.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NSMtAXP4cg8_UoHYhJ4:-NSMtaDy7iPc5w7aQq5n:b-9mps0) (no related file):*

Work in Anzu has added evidence that this feature is maybe useful. Mentioned via TODO in Anzu PR#10342.
</blockquote></details>

Er, not PR#10342, rather PR#10369.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NSs8B8W4ikzybTGMqI2:-NSrxsVNAbHbk8I2bpVJ:b-yjk5mh) (no related file):*
Name blast radius: I would say it's not actually that small. The worst bit is geometry frame naming, but also implicated are any bits that care about model nesting: collision filters, various SDFormat stuff, etc. Some of these probably don't care about the actual name, and could be implemented some other way.

___

rpoyner-tri (commented):
`r2` (revised) gives a reasonable measure of the blast radius, in that tests have been made to pass. Some of the decisions at this stage are clearly wrong. Time to go through and figure out what is really wanted.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NU3bA5IBPvrbpVmhXTU:b-4ng5mq)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 10 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

___
*[`geometry/drake_visualizer.cc` line 830 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7_kVA-tXDnIA7iY_C:-NU7_kVA-tXDnIA7iY_D:b-x3885g) ([raw file](https://github.com/RobotLocomotion/drake/blob/d50dc887be1e27b4a8a78c86016c6783788b7e4d/geometry/drake_visualizer.cc#L830)):*
> ```cpp
>       std::string name = inspector.GetOwningSourceName(frame_id);
>       const std::string group_name = inspector.GetFrameGroupName(frame_id);
>       if (!group_name.empty()) {
> ```

Relevant to other comments elsewhere, this "frame group name is maybe empty" hack allows some existing tests to pass, but needs to be thought through. Probably better is a rule (similar to some existing ones in drake) that frame groups always have some name, with a scheme for creating default names.
___
*[`geometry/geometry_state.h` line 239 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7a9Mx9U3CWi9yngmM:-NU7a9Mx9U3CWi9yngmN:b-ylqfxm) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/geometry/geometry_state.h#L239)):*
> ```objc
> 
>   /** XXX */
>   std::string GetFullName(GeometryId id) const;
> ```

At some point, this same full-name accessor is likely needed for frames.
___
*[`geometry/geometry_state.h` line 239 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7b_Tw5rsJLJbn_u2W:-NU7b_Tw5rsJLJbn_u2X:b-oai3f9) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/geometry/geometry_state.h#L239)):*
> ```objc
> 
>   /** XXX */
>   std::string GetFullName(GeometryId id) const;
> ```

There is some ambiguity here, that needs to be decided and documented. Some clients want "{FRAME_GROUP_NAME}::{LEAF_THING_NAME}" and some want also want source name in the mix.
___
*[`geometry/geometry_state.h` line 239 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7bvTd7jDJiE_IgtbA:-NU7bvTd7jDJiE_IgtbB:bkrbpjt) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/geometry/geometry_state.h#L239)):*
> ```objc
> 
>   /** XXX */
>   std::string GetFullName(GeometryId id) const;
> ```

As literal uses of "::" proliferate, we should consider (physically) demoting scoped name support to geometry/ or even common/.
___
*[`geometry/geometry_state.cc` line 783 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7ZNjU36687hpY65DV:-NU7ZNjU36687hpY65DW:byxdoas) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/geometry/geometry_state.cc#L783)):*
> ```cpp
>   }
>   FrameNameSet& f_name_set = source_frame_name_map_[source_id];
>   // XXX patchery hackery
> ```

This change needs more study. The old rule was "frame names are source-scope unique", but it only worked because MbP was adding model names into the names seen here. Having removed the model name spam from MbP, we have changed the rule here to "frame names are frame-group-scope unique". That might be ok, but needs to be carried through any other parts of geometry state code that depend on the old rule.
___
*[`geometry/meshcat_visualizer.cc` line 173 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7_0MiBQ86BANpVGKE:-NU7_0MiBQ86BANpVGKF:bd62ywr) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/geometry/meshcat_visualizer.cc#L173)):*
> ```cpp
>             : fmt::format("{}/{}/{}",
>                           params_.prefix,
>                           inspector.GetFrameGroupName(frame_id),
> ```

As seen elsewhere in this patch, frame groups are not yet required to have a name. We should either require they have some name, with default values, or make sure all the paths that use frame group names treat them as optional, to avoid paths with "//" or "::::".
___
*[`multibody/tree/multibody_tree.h` line 783 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7aafU2bqvpCqKGLTY:-NU7aafU2bqvpCqKGLTZ:blkmrhc) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/multibody/tree/multibody_tree.h#L783)):*
> ```objc
>   // See MultibodyPlant method.
>   const std::string& GetModelInstanceName(
>       ModelInstanceIndex model_instance, bool freeze = true) const;
> ```

Of course, all the freeze/frozen bit nonsense goes away before this is a real feature patch. It was mostly useful for tracing out all the places that old hacked-up scoped names were used.
___
*[`bindings/pydrake/multibody/plant_py.cc` line 718 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NU7b3xD4sf-mFH7GoZ7:-NU7b3xD4sf-mFH7GoZ8:b-glgdd) ([raw file](https://github.com/RobotLocomotion/drake/blob/4d8c5281888fd70b9ceb78b6b2280f43a86d7df8/bindings/pydrake/multibody/plant_py.cc#L718)):*
> ```cpp
>             overload_cast_explicit<const string&, ModelInstanceIndex, bool>(
>                 &Class::GetModelInstanceName),
>             py::arg("model_instance"), py::arg("freeze") = false,
> ```

This is more freeze/frozen nonsense that goes away eventually.

rpoyner-tri (commented):
`r5` is getting to the point where I don't totally hate the results.

For consideration at your leisure:

@jwnimmer-tri `r5` now has `geometry_state` doing a slightly worse job of what `internal_geometry_names` does, but at a physical layer I can actually use. I'm open to suggestions for caching, or consolidation, or whatever else you notice.

@SeanCurtis-TRI `r5` clearly demonstrates that `geometry_state/scene_graph_inspector` is taking the biggest change-hit from the proposed change so far. I've sketched out new API, but I'm not married to the details; mostly care about the data flow.

Usual draft caveats apply: incomplete/wrong documentation, incomplete tests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NVGY2etDMmTc6K_D_pa:b-xh8rkr)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+@jwnimmer-tri for design review. As usual for draft, doc and tests are likely not complete.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NXByuwJ9BVArwrNxwdX:bejv69o)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri and @rpoyner-tri)

jwnimmer-tri (commented):
> Reviewed 6 of 22 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-NXHrKloCVU309to7ASl:bpux2op) (no related file):*
The changes related to SceneGraph are not clearly enough expressed yet for me to have a well-formed opinion.

It seems like possible it's changing "name" to mean a more sparse name and introducing "qualified name" to mean the current colon-delimited name, with the associated test case fallout.

If I read that correctly, that seems to severe (breaking).  I am not sure whether the "group name" design is helpful, but assuming it is I don't think we want to change the meaning of "name", rather we could introduce "element name" which is the spare one and "group" which has the scope then "name" is the combination of the two.  That would leave users' code intact, while still teasing apart the two contracts.

___
___
*[`multibody/tree/multibody_tree.cc` line 119 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHqBEbBLg31vdurJzJ:-NXHqBEc5a2S9pMD8SGW:bbprq83) ([raw file](https://github.com/RobotLocomotion/drake/blob/86696782f2b75d597399b45f6ca698996c76146e/multibody/tree/multibody_tree.cc#L119)):*
> ```cpp
>   if (!base_body_index.has_value()) {
>     throw std::logic_error("Model " +
>                            GetModelInstanceName(model_instance) +
> ```

BTW Was there something subtle going on here where the map lookup was bad?  Or was this just a related cleanup while we're nearby anyway?

The index validity has already been checked, so the map lookup would always  succeed, I would have thought.

If this change isn't part of the feature, it would probably help me grok the big picture if cleanups were in a different PR.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-NXLYN0mA-HQleNRdXTF:b-8un9yw) (no related file):*

The changes related to SceneGraph are not clearly enough expressed yet for me to have a well-formed opinion.

It seems like possible it's changing "name" to mean a more sparse name and introducing "qualified name" to mean the current colon-delimited name, with the associated test case fallout.

If I read that correctly, that seems to severe (breaking).  I am not sure whether the "group name" design is helpful, but assuming it is I don't think we want to change the meaning of "name", rather we could introduce "element name" which is the spare one and "group" which has the scope then "name" is the combination of the two.  That would leave users' code intact, while still teasing apart the two contracts.
</blockquote></details>

This is good. I was wondering whether scene-graph Name() ==> "foo::bar" has become stable API; apparently it has. 

___
___
*[`multibody/tree/multibody_tree.cc` line 119 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHqBEbBLg31vdurJzJ:-NXLYfZ25LGTO0VeHT-F:bh6iwy1) ([raw file](https://github.com/RobotLocomotion/drake/blob/86696782f2b75d597399b45f6ca698996c76146e/multibody/tree/multibody_tree.cc#L119)):*

BTW Was there something subtle going on here where the map lookup was bad?  Or was this just a related cleanup while we're nearby anyway?

The index validity has already been checked, so the map lookup would always  succeed, I would have thought.

If this change isn't part of the feature, it would probably help me grok the big picture if cleanups were in a different PR.
</blockquote></details>

Ah, this is cruft from the "freeze/frozen" diagnostic code in very early revs (for example `r2`). I can revert these.

amcastro-tri (commented):
Cleaning my Reviewable dashboard, feel free to re-ping me.
-@amcastro-tri 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-NXM-ZQI9_Yqvw8hjhf2:bqg0wow)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-NXuh-hD4hC9iQG5X7_w:b-uhz4ig) (no related file):*

There's a fair debate whether the exact string content is stable API.  It might be OK to change it (e.g., replacing `::` with `/` or something).

If I've understood the code correctly, it seems the most breaking change here is that the "names" are no longer unique, instead only the "qualified names" are unique.
</blockquote></details>

I will claim that the underlying implementation clarifies implicit uniqueness rules that were already in place, but the various name accessor APIs probably still need work to avoid breaking code that relied on old behavior. Still working on actual code changes to propose.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-NY-dllKCoipZ2Fl8cQK:b81fnbg) (no related file):*

I will claim that the underlying implementation clarifies implicit uniqueness rules that were already in place, but the various name accessor APIs probably still need work to avoid breaking code that relied on old behavior. Still working on actual code changes to propose.
</blockquote></details>

`r6` displays some consequences of a first attempt to treat the legacy name behavior as stable.

In it, we have: GetName({Frame, Geometry}Id) return the "legacy" behavior: "model_name::thing_name". To get the actual name, we now have GetUnqualifiedName(). Finally, GetGeometryByName() will now accept either scoped or unscoped names; the scope namespace is just ignored.

There are at least two defects with this approach:
* GetName(), which used to be `const std::string&`-returning, now returns `std::string`. This is a *very* subtle invitation to UB, since likely many callers were depending on be able to assign the returned thing to either a reference or a string_view, and doing that now will leave them pointing to a long-departed temporary in short order. This could be fixed by forcing geometry_state to have a backing store for the scoped strings, but let's not get ahead of ourselves.
* `triad_test.py` shows the other defect: not all callers of RegisterGeometry are named MultibodyPlant, so attempts to mimic legacy Mbp behavior in geometry names turns up weird side-effects: I registered "foo", and got back "model_name::foo".

___

rpoyner-tri (commented):
`r7` reverts (more of?) cruft that Nimmer had pointed out earlier.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-N_5la1p2chlODA7B3Rc:b-rj9g2s)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
`r7`-prime (232462e0ea) starts to fill in some of the missing API docs.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-:-N_5zt7S0bgxkYr-KvGp:b-k01rxg)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

jwnimmer-tri (commented):
> Reviewed 9 of 30 files at r6, 4 of 14 files at r7.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

SeanCurtis-TRI (commented):
> Reviewed 2 of 22 files at r5, 3 of 30 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 14 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-N_fjZqX3SiDQr8LpVKS:bg82egj) (no related file):*

`r6` displays some consequences of a first attempt to treat the legacy name behavior as stable.

In it, we have: GetName({Frame, Geometry}Id) return the "legacy" behavior: "model_name::thing_name". To get the actual name, we now have GetUnqualifiedName(). Finally, GetGeometryByName() will now accept either scoped or unscoped names; the scope namespace is just ignored.

There are at least two defects with this approach:
* GetName(), which used to be `const std::string&`-returning, now returns `std::string`. This is a *very* subtle invitation to UB, since likely many callers were depending on be able to assign the returned thing to either a reference or a string_view, and doing that now will leave them pointing to a long-departed temporary in short order. This could be fixed by forcing geometry_state to have a backing store for the scoped strings, but let's not get ahead of ourselves.
* `triad_test.py` shows the other defect: not all callers of RegisterGeometry are named MultibodyPlant, so attempts to mimic legacy Mbp behavior in geometry names turns up weird side-effects: I registered "foo", and got back "model_name::foo".
</blockquote></details>

I'm going to go ahead and hit publish on the notes I *had* made. Take them with a grain of salt.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2hVkk7Ehh9SBT0IA4:-N_2hVkk7Ehh9SBT0IA5:b-6cnxat) (no related file):*
Don't forget to update the documentation in the following places (regarding the uniqueness of frame names):

 - `geometry_frame.h`: class overview
 - `scene_graph.h`: docs for `RegisterFrame`.

___
___
*[`geometry/geometry_state.cc` line 419 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2nro8Dm2_Y1nukktX:-N_2nro8Dm2_Y1nukktY:b-o56vf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L419)):*
> ```cpp
> const std::string& GeometryState<T>::GetFrameGroupName(
>     SourceId source_id, int frame_group) const {
>   static const std::string empty;
> ```

nit: needs a `never_destroyed`.
___
*[`geometry/geometry_state.cc` line 438 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2lW-693lejXKX8Yb8:-N_2lW-693lejXKX8Yb9:bjdcemb) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L438)):*
> ```cpp
> 
> template <typename T>
> template <typename ID>
> ```

nit: `Id` (style guide)
___
*[`geometry/geometry_state.cc` line 480 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2oKiE2aVyu2O7lN4s:-N_2oKiE2aVyu2O7lN4t:b-xipql5) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L480)):*
> ```cpp
>     NameQualifiers qualifiers,
>     std::string_view delimiter) const {
>   return BuildQualifiedName<GeometryId>(id, qualifiers, delimiter);
> ```

BTW I would believe that you can omit `<GeometryId>` as that can be inferred by the type of the first argument. Same below.  Which is actually one more argument for just calling `BuildQualifiedName(id, qualifiers, delimiter)` directly instead.
___
*[`geometry/geometry_state.cc` line 554 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2pHcs3IrjaJDK-dNt:-N_2pHcs3IrjaJDK-dNu:bvk9hen) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L554)):*
> ```cpp
>   // the already-implemented parsing logic to analyze the in-bound name.
>   const std::string search_name = [&]() {
>     std::regex pattern(".*::([^:]+):*");
> ```

So, you're anticipating including qualified names as inputs (you'd have to update the documentation in `scene_graph_inspector.h`.

If we do that, it seems we have to disallow our delimiter as being a valid name substring.

Tangentially (and this is related to my note on `GetName(GeometryId)`, but I would've thought that the qualified name of a frame would be the prefix of the qualified name of one of its geometries. That's clearly not the case, but it if it were, this would be a simple prefix-test. As it is, I could pass the name "model1::collision" for a geometry that belongs in a different model instance (i.e., its frame's qualified name is "model2::root").
___
*[`geometry/geometry_state.cc` line 554 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2rUAr4gA5mVgL9rqF:-N_2rUAr4gA5mVgL9rqG:bi5qc7d) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L554)):*
> ```cpp
>   // the already-implemented parsing logic to analyze the in-bound name.
>   const std::string search_name = [&]() {
>     std::regex pattern(".*::([^:]+):*");
> ```

Why zero or more terminal `:`? The zero I get. The multiple? I don't.
___
*[`geometry/geometry_state.cc` line 638 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2py5i4SmA5tM6CCZZ:-N_2py5j-3uOnMBpKuol:b-dv4hzy) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L638)):*
> ```cpp
> template <typename T>
> std::string GeometryState<T>::GetName(GeometryId geometry_id) const {
>   return GetQualifiedName(geometry_id, kQualifierFrameGroup, "::");
> ```

This seems odd. I'd think that a qualified geometry name would *always* include its frame name. Otherwise, if I have a `collision` geometry attached to ten links in a model instance, I'll have ten geometries with identical qualified names.
___
*[`geometry/geometry_state.cc` line 847 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2rfo_05kISpwF8soN:-N_2rfo_05kISpwF8soO:b-pr7umv) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L847)):*
> ```cpp
> void GeometryState<T>::NameFrameGroup(
>     SourceId source_id, int frame_group, std::string_view name) {
>   source_to_frame_group_names_[source_id][frame_group] = name;
> ```

BTW Worried about accidental renaming at all?
___
*[`geometry/geometry_state.cc` line 887 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2ruVn1jwXe4mYMDs8:-N_2ruVn1jwXe4mYMDs9:bb10o84) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/geometry_state.cc#L887)):*
> ```cpp
>   // data structures by (source_id, frame_group)?
>   std::string group_prefix;
>   // Frames in the world instance/frame group don't get prefixes; this flows
> ```

This strikes me as a comment on how `MbP` uses `SceneGraph` and *not* what `SceneGraph` is capable of. I believe `SceneGraph` doesn't care who the parent is in setting the frame group of a particular frame.

For example, the "world instance frame group" is *purely* an `MbP` concept. `SceneGraph` doesn't conflate frame groups with parent frames at all. So, any logic we apply in `SceneGraph` should probably get a bit more space from `MbP` logic to define its behavior (or at least to justify it).



_Code quote:_
```cpp
Frames in the world instance/frame group don't get prefixes;
```
___
*[`geometry/meshcat_visualizer.cc` line 182 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2thm71TosDcAVAYWO:-N_2thm71TosDcAVAYWP:blwg2lq) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/meshcat_visualizer.cc#L182)):*
> ```cpp
>                           params_.prefix,
>                           inspector.GetQualifiedName(
>                               frame_id, kQualifierFrameGroup, "/"));
> ```

BTW Presumably, meshcat doesn't care about `::` delimiters in names. Yes?
___
*[`geometry/name_qualifiers.h` line 8 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-N_2geYz6MzcItYJP7Yo:-N_2geYz6MzcItYJP7Yp:b-iex3gt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db10114a97ef5e8672e0b407df4eb92a0537a1fd/geometry/name_qualifiers.h#L8)):*
> ```objc
> namespace geometry {
> 
> using NameQualifiers = uint8_t;
> ```

nit: Plural doesn't look right.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18666)* status: 14 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18666#-NXHrKln6embDiXASvBt:-N_irr2H6Fyu9gVtf5j8:b-5yufoz) (no related file):*

I'm going to go ahead and hit publish on the notes I *had* made. Take them with a grain of salt.
</blockquote></details>

Thanks for the comments. As per f2f discussion, I will also try writing out Nimmer's idea of no-change to SceneGraph (roughly), just re-spam the implicated names as part of the Mbp model instance rename. That will probably turn up as a separate draft PR.

___

rpoyner-tri:
This PR was superseded by #19939.

PR #18670, opened 2023-01-27
[pydrake] Bind planning.CollisionChecker and related

jwnimmer-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)

### DIFF
--- a/bindings/pydrake/planning/BUILD.bazel
+++ b/bindings/pydrake/planning/BUILD.bazel
@@ -30,11 +30,14 @@ drake_pybind_library(
         "//bindings/pydrake:documentation_pybind",
         "//bindings/pydrake/common:cpp_template_pybind",
         "//bindings/pydrake/common:default_scalars_pybind",
+        "//bindings/pydrake/common:wrap_pybind",
     ],
     cc_so_name = "__init__",
     cc_srcs = [
         "planning_py.h",
         "planning_py.cc",
+        "planning_py_collision_checker.cc",
+        "planning_py_collision_checker_interface_types.cc",
         "planning_py_robot_diagram.cc",
     ],
     package_info = PACKAGE_INFO,
@@ -51,6 +54,18 @@ install(
     targets = [":planning"],
     py_dest = PACKAGE_INFO.py_dest,
     deps = get_drake_py_installs([":planning"]),
+)
+
+drake_py_unittest(
+    name = "collision_checker_test",
+    data = [
+        "//multibody:models",
+        "//planning/test_utilities:collision_ground_plane.sdf",
+    ],
+    deps = [
+        ":planning",
+        "//bindings/pydrake/common/test_utilities:numpy_compare_py",
+    ],
 )
 
 drake_py_unittest(
--- a/bindings/pydrake/planning/planning_py.cc
+++ b/bindings/pydrake/planning/planning_py.cc
@@ -18,6 +18,8 @@ and/or trajectories of dynamical systems.
 
   // The order of these calls matters. Some modules rely on prior definitions.
   internal::DefinePlanningRobotDiagram(m);
+  internal::DefinePlanningCollisionCheckerInterfaceTypes(m);
+  internal::DefinePlanningCollisionChecker(m);
 }
 
 }  // namespace pydrake
--- a/bindings/pydrake/planning/planning_py.h
+++ b/bindings/pydrake/planning/planning_py.h
@@ -12,6 +12,14 @@ namespace drake {
 namespace pydrake {
 namespace internal {
 
+// For simplicity, these declarations are listed in alphabetical order.
+
+/* Defines bindings per planning_py_collision_checker.cc. */
+void DefinePlanningCollisionChecker(py::module m);
+
+/* Defines bindings per planning_py_collision_checker_interface_types.cc. */
+void DefinePlanningCollisionCheckerInterfaceTypes(py::module m);
+
 /* Defines bindings per planning_py_robot_diagram.cc. */
 void DefinePlanningRobotDiagram(py::module m);
 
--- a/bindings/pydrake/planning/planning_py_collision_checker.cc
+++ b/bindings/pydrake/planning/planning_py_collision_checker.cc
@@ -1,0 +1,329 @@
+#include "pybind11/eigen.h"
+#include "pybind11/functional.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+
+#include "drake/bindings/pydrake/common/wrap_pybind.h"
+#include "drake/bindings/pydrake/documentation_pybind.h"
+#include "drake/bindings/pydrake/planning/planning_py.h"
+#include "drake/bindings/pydrake/pydrake_pybind.h"
+#include "drake/planning/collision_checker.h"
+#include "drake/planning/scene_graph_collision_checker.h"
+#include "drake/planning/unimplemented_collision_checker.h"
+
+namespace drake {
+namespace pydrake {
+namespace internal {
+
+using multibody::Body;
+using multibody::BodyIndex;
+using multibody::Frame;
+
+void DefinePlanningCollisionChecker(py::module m) {
+  // NOLINTNEXTLINE(build/namespaces): Emulate placement in namespace.
+  using namespace drake::planning;
+  constexpr auto& doc = pydrake_doc.drake.planning;
+
+  {
+    using Class = CollisionChecker;
+    constexpr auto& cls_doc = doc.CollisionChecker;
+    py::class_<Class> cls(m, "CollisionChecker", cls_doc.doc);
+    cls  // BR
+        .def("model", &Class::model, py_rvp::reference_internal,
+            cls_doc.model.doc)
+        .def("plant", &Class::plant, py_rvp::reference_internal,
+            cls_doc.plant.doc)
+        .def("get_body", &Class::get_body, py::arg("body_index"),
+            py_rvp::reference_internal, cls_doc.get_body.doc)
+        .def("GetScopedName",
+            overload_cast_explicit<std::string, const Frame<double>&>(
+                &Class::GetScopedName),
+            py::arg("frame"), cls_doc.GetScopedName.doc_1args_frame)
+        .def("GetScopedName",
+            overload_cast_explicit<std::string, const Body<double>&>(
+                &Class::GetScopedName),
+            py::arg("body"), cls_doc.GetScopedName.doc_1args_body)
+        .def("robot_model_instances", &Class::robot_model_instances,
+            cls_doc.robot_model_instances.doc)
+        .def("IsPartOfRobot",
+            overload_cast_explicit<bool, const Body<double>&>(
+                &Class::IsPartOfRobot),
+            py::arg("body"), cls_doc.IsPartOfRobot.doc)
+        .def("IsPartOfRobot",
+            overload_cast_explicit<bool, BodyIndex>(&Class::IsPartOfRobot),
+            py::arg("body_index"), cls_doc.IsPartOfRobot.doc)
+        .def("GetZeroConfiguration", &Class::GetZeroConfiguration,
+            cls_doc.GetZeroConfiguration.doc)
+        .def("num_allocated_contexts", &Class::num_allocated_contexts,
+            cls_doc.num_allocated_contexts.doc)
+        .def("model_context", &Class::model_context, py_rvp::reference_internal,
+            cls_doc.model_context.doc)
+        .def("plant_context", &Class::plant_context, py_rvp::reference_internal,
+            cls_doc.plant_context.doc)
+        .def("UpdatePositions", &Class::UpdatePositions,
+            py_rvp::reference_internal, py::arg("q"),
+            cls_doc.UpdatePositions.doc)
+        .def("UpdateContextPositions", &Class::UpdateContextPositions,
+            py::arg("model_context"), py::arg("q"),
+            // Keep alive, ownership: `return` keeps `model_context` alive.
+            py::keep_alive<0, 2>(), py_rvp::reference,
+            cls_doc.UpdateContextPositions.doc)
+        .def("MakeStandaloneModelContext", &Class::MakeStandaloneModelContext,
+            cls_doc.MakeStandaloneModelContext.doc)
+        .def("PerformOperationAgainstAllModelContexts",
+            WrapCallbacks(&Class::PerformOperationAgainstAllModelContexts),
+            py::arg("operation"),
+            cls_doc.PerformOperationAgainstAllModelContexts.doc)
+        .def("AddCollisionShape", &Class::AddCollisionShape,
+            py::arg("group_name"), py::arg("description"),
+            cls_doc.AddCollisionShape.doc)
+        .def("AddCollisionShapes",
+            py::overload_cast<const std::string&,
+                const std::vector<BodyShapeDescription>&>(
+                &Class::AddCollisionShapes),
+            py::arg("group_name"), py::arg("descriptions"),
+            cls_doc.AddCollisionShapes.doc_2args)
+        .def("AddCollisionShapes",
+            py::overload_cast<const std::map<std::string,
+                std::vector<BodyShapeDescription>>&>(
+                &Class::AddCollisionShapes),
+            py::arg("geometry_groups"), cls_doc.AddCollisionShapes.doc_1args)
+        .def("AddCollisionShapeToFrame", &Class::AddCollisionShapeToFrame,
+            py::arg("group_name"), py::arg("frameA"), py::arg("shape"),
+            py::arg("X_AG"), cls_doc.AddCollisionShapeToFrame.doc)
+        .def("AddCollisionShapeToBody", &Class::AddCollisionShapeToBody,
+            py::arg("group_name"), py::arg("bodyA"), py::arg("shape"),
+            py::arg("X_AG"), cls_doc.AddCollisionShapeToBody.doc)
+        .def("GetAllAddedCollisionShapes", &Class::GetAllAddedCollisionShapes,
+            cls_doc.GetAllAddedCollisionShapes.doc)
+        .def("RemoveAllAddedCollisionShapes",
+            py::overload_cast<const std::string&>(
+                &Class::RemoveAllAddedCollisionShapes),
+            py::arg("group_name"),
+            cls_doc.RemoveAllAddedCollisionShapes.doc_1args)
+        .def("RemoveAllAddedCollisionShapes",
+            py::overload_cast<>(&Class::RemoveAllAddedCollisionShapes),
+            cls_doc.RemoveAllAddedCollisionShapes.doc_0args)
+        .def("MaybeGetUniformRobotEnvironmentPadding",
+            &Class::MaybeGetUniformRobotEnvironmentPadding,
+            cls_doc.MaybeGetUniformRobotEnvironmentPadding.doc)
+        .def("MaybeGetUniformRobotRobotPadding",
+            &Class::MaybeGetUniformRobotRobotPadding,
+            cls_doc.MaybeGetUniformRobotRobotPadding.doc)
+        .def("GetPaddingBetween",
+            overload_cast_explicit<double, BodyIndex, BodyIndex>(
+                &Class::GetPaddingBetween),
+            py::arg("bodyA_index"), py::arg("bodyB_index"),
+            cls_doc.GetPaddingBetween.doc_2args_bodyA_index_bodyB_index)
+        .def("GetPaddingBetween",
+            overload_cast_explicit<double, const Body<double>&,
+                const Body<double>&>(&Class::GetPaddingBetween),
+            py::arg("bodyA"), py::arg("bodyB"),
+            cls_doc.GetPaddingBetween.doc_2args_bodyA_bodyB)
+        .def("SetPaddingBetween",
+            py::overload_cast<BodyIndex, BodyIndex, double>(
+                &Class::SetPaddingBetween),
+            py::arg("bodyA_index"), py::arg("bodyB_index"), py::arg("padding"),
+            cls_doc.SetPaddingBetween.doc_3args_bodyA_index_bodyB_index_padding)
+        .def("SetPaddingBetween",
+            py::overload_cast<const Body<double>&, const Body<double>&, double>(
+                &Class::SetPaddingBetween),
+            py::arg("bodyA"), py::arg("bodyB"), py::arg("padding"),
+            cls_doc.SetPaddingBetween.doc_3args_bodyA_bodyB_padding)
+        .def("GetPaddingMatrix", &Class::GetPaddingMatrix,
+            py_rvp::reference_internal, cls_doc.GetPaddingMatrix.doc)
+        .def("SetPaddingMatrix", &Class::SetPaddingMatrix,
+            py::arg("collision_padding"), cls_doc.SetPaddingMatrix.doc)
+        .def("GetLargestPadding", &Class::GetLargestPadding,
+            cls_doc.GetLargestPadding.doc)
+        .def("SetPaddingOneRobotBodyAllEnvironmentPairs",
+            &Class::SetPaddingOneRobotBodyAllEnvironmentPairs,
+            py::arg("body_index"), py::arg("padding"),
+            cls_doc.SetPaddingOneRobotBodyAllEnvironmentPairs.doc)
+        .def("SetPaddingAllRobotEnvironmentPairs",
+            &Class::SetPaddingAllRobotEnvironmentPairs, py::arg("padding"),
+            cls_doc.SetPaddingAllRobotEnvironmentPairs.doc)
+        .def("SetPaddingAllRobotRobotPairs",
+            &Class::SetPaddingAllRobotRobotPairs, py::arg("padding"),
+            cls_doc.SetPaddingAllRobotRobotPairs.doc)
+        .def("GetNominalFilteredCollisionMatrix",
+            &Class::GetNominalFilteredCollisionMatrix,
+            py_rvp::reference_internal,
+            cls_doc.GetNominalFilteredCollisionMatrix.doc)
+        .def("GetFilteredCollisionMatrix", &Class::GetFilteredCollisionMatrix,
+            py_rvp::reference_internal, cls_doc.GetFilteredCollisionMatrix.doc)
+        .def("SetCollisionFilterMatrix", &Class::SetCollisionFilterMatrix,
+            py::arg("filter_matrix"), cls_doc.SetCollisionFilterMatrix.doc)
+        .def("IsCollisionFilteredBetween",
+            overload_cast_explicit<bool, BodyIndex, BodyIndex>(
+                &Class::IsCollisionFilteredBetween),
+            py::arg("bodyA_index"), py::arg("bodyB_index"),
+            cls_doc.IsCollisionFilteredBetween
+                .doc_2args_bodyA_index_bodyB_index)
+        .def("IsCollisionFilteredBetween",
+            overload_cast_explicit<bool, const Body<double>&,
+                const Body<double>&>(&Class::IsCollisionFilteredBetween),
+            py::arg("bodyA"), py::arg("bodyB"),
+            cls_doc.IsCollisionFilteredBetween.doc_2args_bodyA_bodyB)
+        .def("SetCollisionFilteredBetween",
+            py::overload_cast<BodyIndex, BodyIndex, bool>(
+                &Class::SetCollisionFilteredBetween),
+            py::arg("bodyA_index"), py::arg("bodyB_index"),
+            py::arg("filter_collision"),
+            cls_doc.SetCollisionFilteredBetween
+                .doc_3args_bodyA_index_bodyB_index_filter_collision)
+        .def("SetCollisionFilteredBetween",
+            py::overload_cast<const Body<double>&, const Body<double>&, bool>(
+                &Class::SetCollisionFilteredBetween),
+            py::arg("bodyA"), py::arg("bodyB"), py::arg("filter_collision"),
+            cls_doc.SetCollisionFilteredBetween
+                .doc_3args_bodyA_bodyB_filter_collision)
+        .def("SetCollisionFilteredWithAllBodies",
+            &Class::SetCollisionFilteredWithAllBodies, py::arg("body_index"),
+            cls_doc.SetCollisionFilteredWithAllBodies.doc)
+        .def("CheckConfigCollisionFree", &Class::CheckConfigCollisionFree,
+            py::arg("q"), cls_doc.CheckConfigCollisionFree.doc)
+        .def("CheckContextConfigCollisionFree",
+            &Class::CheckContextConfigCollisionFree, py::arg("model_context"),
+            py::arg("q"), cls_doc.CheckContextConfigCollisionFree.doc)
+        .def("CheckConfigsCollisionFree", &Class::CheckConfigsCollisionFree,
+            py::arg("configs"), py::arg("parallelize") = true,
+            cls_doc.CheckConfigsCollisionFree.doc)
+        .def("SetConfigurationDistanceFunction",
+            WrapCallbacks(&Class::SetConfigurationDistanceFunction),
+            py::arg("distance_function"),
+            cls_doc.SetConfigurationDistanceFunction.doc)
+        .def("ComputeConfigurationDistance",
+            &Class::ComputeConfigurationDistance, py::arg("q1"), py::arg("q2"),
+            cls_doc.ComputeConfigurationDistance.doc)
+        .def("MakeStandaloneConfigurationDistanceFunction",
+            &Class::MakeStandaloneConfigurationDistanceFunction,
+            cls_doc.MakeStandaloneConfigurationDistanceFunction.doc)
+        .def("SetConfigurationInterpolationFunction",
+            WrapCallbacks(&Class::SetConfigurationInterpolationFunction),
+            py::arg("interpolation_function"),
+            cls_doc.SetConfigurationInterpolationFunction.doc)
+        .def("InterpolateBetweenConfigurations",
+            &Class::InterpolateBetweenConfigurations, py::arg("q1"),
+            py::arg("q2"), py::arg("ratio"),
+            cls_doc.InterpolateBetweenConfigurations.doc)
+        .def("MakeStandaloneConfigurationInterpolationFunction",
+            &Class::MakeStandaloneConfigurationInterpolationFunction,
+            cls_doc.MakeStandaloneConfigurationInterpolationFunction.doc)
+        .def("edge_step_size", &Class::edge_step_size,
+            cls_doc.edge_step_size.doc)
+        .def("set_edge_step_size", &Class::set_edge_step_size,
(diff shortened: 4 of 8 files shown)


SeanCurtis-TRI (commented):
+@SeanCurtis-TRI for feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NMo1cdm-OQoqPSafSG-:b-gpqagm)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers

SeanCurtis-TRI (commented):
:LGTM: With some questions, some superficial things, and one request about changing some particular bindings around a bit.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NMoFQov6CWkqKEp_Ept:b-qc9vqi)* status: 7 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py.h` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo3ee_7xjGCENLi93V:-NMo3ee_7xjGCENLi93W:b-iwaflg) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L15)):*
> ```objc
> namespace internal {
> 
> /* Defines bindings per planning_py_collision_checker_accoutrements.cc. */
> ```

BTW Have you considered declaring these in the same order as they must be invoked? We could go a step further and each could document which of the functions each must follow. Either way, the disparity between ordering here and the ordering of invocation in `planning_py.cc` did stand out in my mind.
___
*[`bindings/pydrake/planning/planning_py.h` line 16 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo79b7087MTSghP2BV:-NMo79b7087MTSghP2BW:b-bmpdo4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L16)):*
> ```objc
> 
> /* Defines bindings per planning_py_collision_checker_accoutrements.cc. */
> void DefinePlanningCollisionCheckerAccoutrements(py::module m);
> ```

BTW Could I persuade you to swap "accountrement" with "component"? The former has the implication of adornments -- something to decorate a pre-existing object. That implies a dependency in the wrong direction.  Components has a strong connotation of "collision checker is made up of these", communicating a better relationship.
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo5iEW6_1RnhBGPavR:-NMo5iEX0pjtzy0UsU-l:b-l1uyrw) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker.cc#L47)):*
> ```cpp
>             py::arg("body"), cls_doc.GetScopedName.doc_1args_body)
>         .def("robot_model_instances", &Class::robot_model_instances,
>             cls_doc.robot_model_instances.doc)
> ```

BTW: This returns an internal reference. Why doesn't this require `py_rvp::reference_internal`? I tried the following test, expecting a failure:

```py
    def instances_survival(self):
        dut = self._make_scene_graph_collision_checker()
        self.assertEqual(len(dut.robot_model_instances()), 1)
        return dut.robot_model_instances()

    def test_survival(self):
        instances = self.instances_survival()
        self.assertEqual(len(instances), 1)
```

But it didn't fail. What am I missing? Are we automatically creating "lists" from vectors such that the list is a unique object from the vector? And, in this case, because the contents of the vector are trivially copyable, we don't have aliasing. But if the vector contained a bunch of unique pointers, we'd get it back?

I noticed similar things with eigen vectors and strings. So, presumably, they all bind in a way that avoids the lingering dependency issue frequently associated with returning internal references, yes?
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo7wdK7ywpcrZ_Dax9:-NMo7wdK7ywpcrZ_DaxA:bh4ef21) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker.cc#L84)):*
> ```cpp
>                 const std::vector<BodyShapeDescription>&>(
>                 &Class::AddCollisionShapes),
>             py::arg("group_name"), py::arg("descriptions"),
> ```

BTW You already did some cleanup on parameter names (q_1 -> q). Is it worth tweaking this one so that there's a larger difference between `description` and `descriptions`? One trailing letter in a long word is not a lot to go on.
___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoDKhfEGjU485APE4O:-NMoDKhfEGjU485APE4P:b-gcq57o) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L49)):*
> ```cpp
> 
>   {
>     using Class = CollisionCheckerParams;
> ```

nit *Almost* alphabetical order, except `Context` should come before `Params`.
___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoBOsM-rQuygdV9wRg:-NMoBOsM-rQuygdV9wRh:b-y2z5ji) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L57)):*
> ```cpp
>             "model",
>             [](const Class& self) -> const RobotDiagram<double>* {
>               return self.model.get();
> ```

BTW This is another case of bindings where it's not clear that the `RobotDiagram` that is returned here survives destruction of the original `params`. What's the mechanism we have in place?
___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoC_sDDNq17e3FSB71:-NMoC_sE8BBucb_yPe4k:bl3d286) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L154)):*
> ```cpp
>             Class::kEnvironmentAndSelfCollision,
>             cls_doc.kEnvironmentAndSelfCollision.doc)
>         .def("SetInEnvironmentCollision", &SetInEnvironmentCollision,
> ```

This strikes me as a very odd binding. Calling `collision_type.SetInSelfCollision()` has a strong implication of mutating the object, but that's not what happens. It seems *particularly* odd to call it on `RobotCollisionType.kNoCollision`. It seems it'd be better to bind this as a class method or a free function as in C++.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py.h` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo3ee_7xjGCENLi93V:-NMoGOvt7eLb0SMxEm6J:b-1hvs7o) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L15)):*

BTW Have you considered declaring these in the same order as they must be invoked? We could go a step further and each could document which of the functions each must follow. Either way, the disparity between ordering here and the ordering of invocation in `planning_py.cc` did stand out in my mind.
</blockquote></details>

We should exactly one source of truth for what order things are loaded in, and that's the CC file.  Keeping it sync'd in >1 place is hellish.

My intent here was to alpha-order; I'll make that clearer (and swap the first two to actually be alpha-sorted).
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo5iEW6_1RnhBGPavR:-NMoFxNG7ngt47L2AWjC:b4pare3) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker.cc#L47)):*

BTW: This returns an internal reference. Why doesn't this require `py_rvp::reference_internal`? I tried the following test, expecting a failure:

```py
    def instances_survival(self):
        dut = self._make_scene_graph_collision_checker()
        self.assertEqual(len(dut.robot_model_instances()), 1)
        return dut.robot_model_instances()

    def test_survival(self):
        instances = self.instances_survival()
        self.assertEqual(len(instances), 1)
```

But it didn't fail. What am I missing? Are we automatically creating "lists" from vectors such that the list is a unique object from the vector? And, in this case, because the contents of the vector are trivially copyable, we don't have aliasing. But if the vector contained a bunch of unique pointers, we'd get it back?

I noticed similar things with eigen vectors and strings. So, presumably, they all bind in a way that avoids the lingering dependency issue frequently associated with returning internal references, yes?
</blockquote></details>

https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies

When we don't say anything here, the default of "automatic" is in effect, which will copy when the return type is a const reference.
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo7wdK7ywpcrZ_Dax9:-NMoGzYz0hNodABB9OI4:b1eq5kt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker.cc#L84)):*

BTW You already did some cleanup on parameter names (q_1 -> q). Is it worth tweaking this one so that there's a larger difference between `description` and `descriptions`? One trailing letter in a long word is not a lot to go on.
</blockquote></details>

If the user typos the name, the type-checker will fail.  That usability seems OK to me.  If I were calling these functions, I wouldn't use the kwarg names anyway, they are annoyingly long, I'd just use positional and let the overload selection grab the right one.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 5 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py.h` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo3ee_7xjGCENLi93V:-NMoHZc9Cn29dOhwRbL8:bcbnnap) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L15)):*

We should exactly one source of truth for what order things are loaded in, and that's the CC file.  Keeping it sync'd in >1 place is hellish.

My intent here was to alpha-order; I'll make that clearer (and swap the first two to actually be alpha-sorted).
</blockquote></details>

I can support alpha as well.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/test/collision_checker_test.py` line 304 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoJMT81_JAq2mKwfBy:-NMoJMT81_JAq2mKwfBz:b-e2mlue) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/test/collision_checker_test.py#L304)):*
> ```Python
>         self.assertEqual(
>             len(dut.CheckConfigsCollisionFree(
>                 configs=[q]*4, parallelize=False)),
> ```

BTW One last thought, I've recently come to the conclusion that when we create a binding with "optional" parameters, that the unit test is obliged to confirm that they truly are optional (and, ideally, default to the expected value). It seems at least as important as making sure the keyword arguments work.

Thoughts?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py.h` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo3ee_7xjGCENLi93V:-NMoPLfH8FgGTItBBa1z:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L15)):*

I can support alpha as well.
</blockquote></details>

Done
___
*[`bindings/pydrake/planning/planning_py.h` line 16 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo79b7087MTSghP2BV:-NMoSD-x2a6yCnZhln8z:b21iahz) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L16)):*

BTW Could I persuade you to swap "accountrement" with "component"? The former has the implication of adornments -- something to decorate a pre-existing object. That implies a dependency in the wrong direction.  Components has a strong connotation of "collision checker is made up of these", communicating a better relationship.
</blockquote></details>

I had a bit of the same distaste here, but I hadn't landed on anything better yet.

I'd like to find a word that most clearly denotes "tiny struct-like classes that are used by CC as argument-type and return-type sugar, and are not generally used nor useful beyond that scope".

I discarded a lot of too-generic words like "dependencies" or "helpers" or "utilities", and I think "components" is also too generic.

WDYT about "tributaries"?
___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoDKhfEGjU485APE4O:-NMoP7tjBxTYrmXyBrlv:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L49)):*

nit *Almost* alphabetical order, except `Context` should come before `Params`.
</blockquote></details>

Done
___
*[`bindings/pydrake/planning/test/collision_checker_test.py` line 304 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoJMT81_JAq2mKwfBy:-NMoPQxyBfnH5XtPfbiT:b7z7otp) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/test/collision_checker_test.py#L304)):*

BTW One last thought, I've recently come to the conclusion that when we create a binding with "optional" parameters, that the unit test is obliged to confirm that they truly are optional (and, ideally, default to the expected value). It seems at least as important as making sure the keyword arguments work.

Thoughts?
</blockquote></details>

To be precise, I understand you're talking about _defaulted_ parameters (as opposed to `std::optional` and/or `typing.Optional`).

It's somewhat borderline for me.  Is it ever likely to break and not be detected before merging to master?  Probably not.  On the other hand, it's not too difficult to add a quick call without the defaulted names, so I've done that here.

In terms of keeping the defaults in sync -- mostly that's for code review to handle.  Anytime C++ changes a default argument, during code review we should be very careful to find everywhere it was used and check for problems, and that also covers the bindings.

For cases that are more important or likely to change than the `parallelize` argument here, we should write the C++ code to not hard-code the default either.  For example, by using `std::optional` or by giving the default value a named constant and using the placeholder name in the bindings.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoBOsM-rQuygdV9wRg:-NMoTWQ4Cnrw3dUEbaAs:b5wgp15) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L57)):*
I had to go look it up to check, but I think pybind11 defaults are already correct:

> `return_value_policy::reference_internal`... is the default policy for property getters created via def_property ...
___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoC_sDDNq17e3FSB71:-NMoTpG23qVQaMtY3wD_:bt5t16) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L154)):*

This strikes me as a very odd binding. Calling `collision_type.SetInSelfCollision()` has a strong implication of mutating the object, but that's not what happens. It seems *particularly* odd to call it on `RobotCollisionType.kNoCollision`. It seems it'd be better to bind this as a class method or a free function as in C++.
</blockquote></details>

You make a good point; this is too confusing as-is.

On the other hand, my sense of Python aesthetics cannot possibly tolerate this as a module-level function.

I'll see what I can figure out a tune-up.  You idea of a staticmethod is appealing.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoBOsM-rQuygdV9wRg:-NMoUYVOCQXP_nVgxF-u:b-epvpjd) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L57)):*

I had to go look it up to check, but I think pybind11 defaults are already correct:

> `return_value_policy::reference_internal`... is the default policy for property getters created via def_property ...
</blockquote></details>

Oh, but what you're were getting at is is what happens when the Params are given to the CollisionChecker.  I'll look into that more.

jwnimmer-tri (commented):
Yes, probably a good idea.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NN2QzVI9LqHCxaxqXRZ:b-9v2dja)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py.h` line 16 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMo79b7087MTSghP2BV:-NN2R47z40T-d7eawuXd:b4ewkca) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py.h#L16)):*

"Tributaries" has the right connotation, but it is a piece of figurative language that isn't a term of art.

We could go for quite literal and say `InterfaceTypes`. It is, after all, a single function and file name and the extra characters represent a negligible cost compared the disambiguation literalness provides.
</blockquote></details>

Working

Great!  Yes, I like that proposal.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoC_sDDNq17e3FSB71:-NN2SqMQADn0cIw0e1z0:b-ws2fmg) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L154)):*

You make a good point; this is too confusing as-is.

On the other hand, my sense of Python aesthetics cannot possibly tolerate this as a module-level function.

I'll see what I can figure out a tune-up.  You idea of a staticmethod is appealing.
</blockquote></details>

Done.  See the unit test and/or generated docs for specifics.

SeanCurtis-TRI (commented):
:lgtm_cancel: 

The new approach is certainly more pythonic. I think it can be tweaked marginally and the fact that we're not doing a literal mapping of the C++ API needs to be accommodated more directly.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NN2TzLq9UTHetWS6zAl:b-ao6a3w)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/planning/test/collision_checker_test.py` line 211 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN2UcVk9IKDtgHIF4s0:-NN2UcVk9IKDtgHIF4s1:brj609v) ([raw file](https://github.com/RobotLocomotion/drake/blob/c818037f031b156b8b59cd5c0b8685326ebc0c43/bindings/pydrake/planning/test/collision_checker_test.py#L211)):*
> ```Python
>             return dut.kNoCollision
> 
>         # Brute-force test all permutations of the Update method.
> ```

nit: s/Update/Updated.
___
*[`bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc` line 155 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN2VIZqBxbwdYHJY5YI:-NN2VIZqBxbwdYHJY5YJ:beguz14) ([raw file](https://github.com/RobotLocomotion/drake/blob/c818037f031b156b8b59cd5c0b8685326ebc0c43/bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc#L155)):*
> ```cpp
>             cls_doc.kEnvironmentAndSelfCollision.doc)
>         .def(
>             "Updated",
> ```

nit: I feel this deserves a comment saying it's explicitly taking the place of the corresponding C++ free functions. It might almost be worth noting it in the documentation as well in case a someone who sees the C++ code goes looking for it in python. If the *names* appear, at least a text search on the python documentations will lead them to this alternative.

(I opted for putting it in the docstring below. If you agree with that path, there obviously isn't any need for a C++ comment.)
___
*[`bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc` line 155 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN2Vaei6JEfm-Ge0w1c:-NN2Vaei6JEfm-Ge0w1d:bwvb4y4) ([raw file](https://github.com/RobotLocomotion/drake/blob/c818037f031b156b8b59cd5c0b8685326ebc0c43/bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc#L155)):*
> ```cpp
>             cls_doc.kEnvironmentAndSelfCollision.doc)
>         .def(
>             "Updated",
> ```

nit: I still find that `Updated` has a very "mutable" flavor to it. It does borrow from the pattern of `normalize` and `normalized` to distinguish in-place vs new copy. Would it be possible to add a verb? `MakeUpdated`? `MakeTweaked`? `MakeVariant`? I think the verb `Make` would go a long way to implying something new is being created.
___
*[`bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc` line 170 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN2WBByAuL02YD0dB4e:-NN2WBByAuL02YD0dB4f:bnxqvk8) ([raw file](https://github.com/RobotLocomotion/drake/blob/c818037f031b156b8b59cd5c0b8685326ebc0c43/bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc#L170)):*
> ```cpp
>             py::kw_only(), py::arg("in_environment_collision") = py::none(),
>             py::arg("in_self_collision") = py::none(),
>             "Returns a new RobotCollisionType with a possibly-updated value; "
> ```

nit: In keeping with my `Make` suggestion above, words like "reset" still linger in the wrong domain. Furthermore, you have phrases like "the environment-collision value will be reset to the new value". But there is no such value to reset.

Here's my rough, first pass at it:

```c++
            "Returns a new RobotCollisionType derived from ``self``. The new "
            "instance defaults to encoding the same collision type as ``self``.
            " If ``in_environment_collision`` is given, the returned instance "
            "encodes the specified environment-collision state (leaving the ""
            "self-collision state the same). The converse is true with respect "
            "to defining ``in_self_collision``. If both are given, it fully "
            "specifies the new instance. This function subsumes the C++ free "
            "functions: SetInEnvironmentCollision() and SetInSelfCollision()."
```

We could also mention that it effectively creates a constructor based on the two bools. But that feels strongly implied.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NMoBOsM-rQuygdV9wRg:-NN2yKDcAnfmVKJEKnwS:bxibsdp) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f01aacc139d05f540e3b8ce13aeaa2b517cce67/bindings/pydrake/planning/planning_py_collision_checker_accoutrements.cc#L57)):*

Thanks
</blockquote></details>

Done.

For now, I've withdrawn the `Checker(params=...)` constructor.  Instead, users pass `model=...` directly, and it's manually annotated.

SeanCurtis-TRI (commented):
And back. :LGTM:

To be perfectly frank, your bindings shennigans are a bit out of my league. You might consider getting an extra pair of eyes on that constructor implementation.

> Reviewed 1 of 3 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NN3a26l6gso59ZIXhzz:b-q2xv9i)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

___
*[`bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc` line 155 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN2fcEbD7aFYFgKKL7I:-NN2fcEbD7aFYFgKKL7J:b22i9a2) ([raw file](https://github.com/RobotLocomotion/drake/blob/13e7ade81373db686e1189e675d012e8f4a935b0/bindings/pydrake/planning/planning_py_collision_checker_interface_types.cc#L155)):*
> ```cpp
>             cls_doc.kEnvironmentAndSelfCollision.doc)
>         // We don't bind the un-Pythonic SetInEnvironmentCollision and
>         // SetInSelfCollision. Instead, we'll define som Python-only sugar.
> ```

nit s/som/some

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

ggould-tri (commented):
Definitely not getting to this one today.
I can pick it up tomorrow or you can reassign.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NN3fmxdEm3A9PBL2j7e:b-pwul68)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri)

jwnimmer-tri:
I am OK to wait until tomorrow, if you are game.  I've already spilled a half dozen over to Rico today.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri)

ggould-tri (commented):
:lgtm:

> Reviewed 2 of 8 files at r1, 2 of 5 files at r3, 1 of 3 files at r4, 2 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-:-NN7kZsO-Xl5DnROH1og:bnfp4nl)* status: 2 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`planning/unimplemented_collision_checker.h` line 20 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7HqlQ7gldYevUq1qT:-NN7HqlQ7gldYevUq1qU:b-j87egs) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/planning/unimplemented_collision_checker.h#L20)):*
> ```objc
> class UnimplementedCollisionChecker : public CollisionChecker {
>  public:
>   /** Constructs a checker with the given settings. */
> ```

minor; author's discretion:  This seems to violate the spirit of the styleguide's "do not... state the completely obvious." (https://drake.mit.edu/styleguide/cppguide.html#Function_Comments)
But possibly you are doing this to make the doxygen prettier?
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 287 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7jvZf5noNhgNyJBT7:-NN7jvZf5noNhgNyJBT8:b2egqmd) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/bindings/pydrake/planning/planning_py_collision_checker.cc#L287)):*
> ```cpp
>             (std::string(cls_doc.ctor.doc) +
>                 "\n\n"
>                 "See :class:`pydrake.planning.CollisionCheckerParams` for the "
> ```

BTW:  I took a brief dive on why this link goes to half way through the relevant section, but wasn't able to figure it out in reasonable time.  Some css geometry awfulness, I guess.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 2 unresolved discussions (waiting on @ggould-tri and @jwnimmer-tri)

___
*[`planning/unimplemented_collision_checker.h` line 20 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7HqlQ7gldYevUq1qT:-NN7lp501QBBJSXEnzBS:bboaq44) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/planning/unimplemented_collision_checker.h#L20)):*

minor; author's discretion:  This seems to violate the spirit of the styleguide's "do not... state the completely obvious." (https://drake.mit.edu/styleguide/cppguide.html#Function_Comments)
But possibly you are doing this to make the doxygen prettier?
</blockquote></details>

For some reason, the mkdoc tool did not even generate an _empty_ docstring constant here (`cls_doc.ctor.doc`) so the bindings wouldn't even compile without something here.

Can you think of anything less obvious I could say here?  Maybe I should document the second argument...
___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 298 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7m2AeBhdb6oanTUYo:-NN7m2Af0tdjUTS_jKGW:bi5hbyl) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/bindings/pydrake/planning/planning_py_collision_checker.cc#L298)):*
> ```cpp
>         m, "UnimplementedCollisionChecker", cls_doc.doc);
>     cls  // BR
>         .def(py::init([](std::unique_ptr<CollisionCheckerParams> params,
> ```

Working

This needs the `model` lifetime fixup as well, I think.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`planning/unimplemented_collision_checker.h` line 20 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7HqlQ7gldYevUq1qT:-NN7mqdO-FYb4ohP9yKn:b-suyq5h) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/planning/unimplemented_collision_checker.h#L20)):*

For some reason, the mkdoc tool did not even generate an _empty_ docstring constant here (`cls_doc.ctor.doc`) so the bindings wouldn't even compile without something here.

Can you think of anything less obvious I could say here?  Maybe I should document the second argument...
</blockquote></details>

OK  There's really nothing to say about it.  It's the world's most self-explanatory ctor.  Meh.

You could always copy-paste the `supports_parallel` docs from one of the other places it appears.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @ggould-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/planning/planning_py_collision_checker.cc` line 298 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18670#-NN7m2AeBhdb6oanTUYo:-NN83hctCHnODk2fzxCz:b1yycla) ([raw file](https://github.com/RobotLocomotion/drake/blob/26d0e7a384fc31e01a0bf09ced613f6a35b6515f/bindings/pydrake/planning/planning_py_collision_checker.cc#L298)):*

Working

This needs the `model` lifetime fixup as well, I think.
</blockquote></details>

Done.

It's somewhat a lot of pasta, but the two constructors are different enough that sharing the code is a bit difficult.

I hope to have the shared_ptr thing solved in the next few weeks anyway.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18670)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

Issue #12591, opened 2020-01-16
pydrake documentation has c++ code in it.


RussTedrake:
Some of the best documentation that we have (in c++/doxygen) has nice little code snippets, but always in c++.  For a python user looking at the pydrake docs (especially one unfamiliar with c++), these are scary and even confusing.

MultibodyPlant's documentation has a few examples:
https://drake.mit.edu/pydrake/pydrake.multibody.plant.html#pydrake.multibody.plant.MultibodyPlant_

Thinking for a bit about how we could realistically/sustainably improve the situation, the doxygen code blocks are surrounded by `@code` `@endcode` blocks.  Would it be crazy to add a `@pycode` block type that c++ doxygen would know to ignore and the pydrake sphinx code would know to render?

Assigning to @sherm1 to start the discussion.  cc @EricCousineau-TRI , @jamiesnape .

sherm1:
This seems like a great idea, Russ. This (and many other things) suggests to me that proper support of PyDrake can't be fully automated -- I wonder if it is a big enough job to warrant a full time position for someone?

jamiesnape:
We are in the process of automating, though it will be months rather than weeks to implement.

RussTedrake:
@jamiesnape — my understanding is that you are automating the creation of bindings.  I don’t know how you would automate this step (touching up the code samples in the docs?).  Is that fair?

jamiesnape:
We probably wouldn't ever want to automate those. The original idea of inserting `@pycode` seems the best idea. I was more responding to "proper support of PyDrake can't be fully automated", which I do not think is a true statement.

SeanCurtis-TRI:
"proper support of PyDrake" comes down to semantics. Ultimately, if that includes converting C++ code examples in doxygen to a python version, then the statement can easily be considered true.

One thing @sherm1 and I vaguely discussed in this context is that a full-time role responsible for improving all aspects of the pydrake experience -- including creating/coordinating an appropriate set of tutorials, responsibility for the final look of python documentation (even if that means going in and adding `@pycode` tags by hand), etc. 

If I were to take a liberty and define "proper support of PyDrake" to that highest level -- maximizing ease of use, then that will never be fully automated. And perhaps we need to consider if the value we place on it corresponds to dedicated personnel.

SeanCurtis-TRI:
Another random thought on the distinction between libdrake and pydrake.

[...]

PR #18674, opened 2023-01-27
[pydrake] Fix incompatibility with Mypy

### LINKED ISSUE #18580: Incompatibility between pydrake and Mypy

jwnimmer-tri:
When parsing the docstrings of a function to find its set of overloads, C++ code samples that mention the function name were at risk of matching the C++ scoping operator `::` as a type annotation, leading to malformed output. For example `MyFunc(foo : int) -> bool` is a valid signature, but the sample code `MyFunc(Eigen::Lower)` was being mis-parsed. Fix this by patching stubgen to reset in case it sees a double colon.

Enable the stubgen regression for all of pydrake; no more caveats.

Revert da58ae08 and 2cf04fdc header file changes, which were work-arounds for this parsing bug and are no longer needed.

Closes #18580.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674)

### DIFF
--- a/bindings/pydrake/test/stubgen_test.py
+++ b/bindings/pydrake/test/stubgen_test.py
@@ -32,10 +32,6 @@ class TestStubgen(unittest.TestCase):
         stubs_dir, paths = self._expected_pyi_files()
         self.assertGreater(len(paths), 0)
         for path in paths:
-            # TODO(jwnimmer-tri) This file still has several syntax errors.
-            # We should work to fix those, but for now we'll skip over it.
-            if path == "pydrake/solvers/__init__.pyi":
-                continue
             with self.subTest(pyi=path):
                 self._check_one_valid_syntax(stubs_dir, path)
 
--- a/common/symbolic/expression/formula.h
+++ b/common/symbolic/expression/formula.h
@@ -341,9 +341,9 @@ Formula isfinite(const Expression& e);
  *
  * @note This method checks if @p m is symmetric, which can be costly. If you
  * want to avoid it, please consider using
- * `positive_semidefinite(m.triangularView<...>())` instead of
- * `positive_semidefinite(m)`, where `...` is either `Eigen::Lower` or
- * `Eigen::Upper`.
+ * `positive_semidefinite(m.triangularView<Eigen::Lower>())` or
+ * `positive_semidefinite(m.triangularView<Eigen::Upper>())` instead of
+ * `positive_semidefinite(m)`.
  *
  * @pydrake_mkdoc_identifier{1args_m}
  */
--- a/multibody/tree/geometry_spatial_inertia.h
+++ b/multibody/tree/geometry_spatial_inertia.h
@@ -18,8 +18,9 @@ namespace multibody {
  (and, therefore, Bo). These are the shapes that have symmetry across So along
  each of the axes Sx, Sy, Sz (e.g., geometry::Box, geometry::Sphere, etc.) For
  meshes, it depends on how the mesh is defined. For more discussion on the
- nuances of geometry::Mesh and geometry::Convex calculations refer to the
- `mesh` overload, below.
+ nuances of geometry::Mesh and geometry::Convex calculations
+ @ref CalcSpatialInertia(const geometry::TriangleSurfaceMesh<double>&,double)
+ "see below".
 
  Note: Spatial inertia calculations for the geometry::Convex type do not
  currently require that the underlying mesh actually be convex. Although certain
@@ -28,7 +29,8 @@ namespace multibody {
 
  @retval M_BBo_B The spatial inertia of the hypothetical body implied by the
                  given `shape`.
- @throws std::exception if `shape` is not supported (e.g., a half space)
+ @throws std::exception if `shape` is an instance of geometry::HalfSpace or
+                        geometry::MeshcatCone.
  @pydrake_mkdoc_identifier{shape} */
 SpatialInertia<double> CalcSpatialInertia(const geometry::Shape& shape,
                                           double density);
--- a/tools/workspace/mypy_internal/patches/reject_double_colon.patch
+++ b/tools/workspace/mypy_internal/patches/reject_double_colon.patch
@@ -1,0 +1,26 @@
+Fix mis-parsing of double colon ("::")
+
+When parsing the docstrings of a function to find its set of overloads,
+C++ code samples that mention the function name were at risk of matching
+the C++ scoping operator "::" as a type annotation, leading to malformed
+output. For example "MyFunc(foo: int) -> bool" is a valid signature, but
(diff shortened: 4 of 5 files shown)


jwnimmer-tri:

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NMoQakY6LzLrT-5y0-A:-NMoQakY6LzLrT-5y0-B:b2vbwgu) (no related file):*
Working

This patch changed a bit more of the pyi output than I expected:
https://gist.github.com/jwnimmer-tri/b391107de9ae4484a3950abc97d1ae39

I'm going to double-check to see if the changes are all improvements.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NMoQakY6LzLrT-5y0-A:-NMoshvx0BcJfb7r-oB_:b-5cnt2r) (no related file):*

Working

This patch changed a bit more of the pyi output than I expected:
https://gist.github.com/jwnimmer-tri/b391107de9ae4484a3950abc97d1ae39

I'm going to double-check to see if the changes are all improvements.
</blockquote></details>

OK I am happy with these changes.  All of them are either:

(1) Removing untyped argument names from signatures where the type is a C++ type -- because our bindings are not correctly topologically ordered and the C++ name is leaking into the docs (~90%).

(2) Removing totally buggy and wrong signatures like an argument named `std` which is really a mis-understood `std::` (~10%).

___

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-:-NN2RyO47x4OrSQZ6TVU:b-gf9ydc)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NMoQakY6LzLrT-5y0-A:-NN2RjduBaeDLXk23MzD:b6zf3zc) (no related file):*

OK I am happy with these changes.  All of them are either:

(1) Removing untyped argument names from signatures where the type is a C++ type -- because our bindings are not correctly topologically ordered and the C++ name is leaking into the docs (~90%).

(2) Removing totally buggy and wrong signatures like an argument named `std` which is really a mis-understood `std::` (~10%).
</blockquote></details>

I see. So, the apparent degradation (substitution of parameter names with `*args, **kwargs` and a return type of `Any` simply points to defects in the organization of binding. So, fixing *those* defects will restore the .pyi files to a better state?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NMoQakY6LzLrT-5y0-A:-NN2UB3xDLbm2F5j27k6:b66pysu) (no related file):*

I see. So, the apparent degradation (substitution of parameter names with `*args, **kwargs` and a return type of `Any` simply points to defects in the organization of binding. So, fixing *those* defects will restore the .pyi files to a better state?
</blockquote></details>

Yes, exactly.  I have a hope that down the road we'll have a build-system-automated interlock that all Drake C++ types used in Drake function signatures must resolve correctly in the docs.

___

ggould-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-:-NN2WTt1DnBsHXyylFXg:bg7igk9)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`tools/workspace/mypy_internal/patches/reject_double_colon.patch` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NN2Wgqz668qZ50SMb9D:-NN2Wgqz668qZ50SMb9E:bll4w4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/c6de1e9e417c44924d01f248444be33eb6cd29ec/tools/workspace/mypy_internal/patches/reject_double_colon.patch#L9)):*
> ```Diff
> the sample code "MyFunc(Eigen::Lower)" was being mis-parsed. Fix this by
> patching stubgen to reset in case it sees a double colon.
> ```

minor:  Has this been communicated upstream; can this reference an upstream issue number?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18674)* status: 1 unresolved discussion (waiting on @ggould-tri)

___
*[`tools/workspace/mypy_internal/patches/reject_double_colon.patch` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18674#-NN2Wgqz668qZ50SMb9D:-NN2XqAV6tseutZNjgSm:b-nw6cys) ([raw file](https://github.com/RobotLocomotion/drake/blob/c6de1e9e417c44924d01f248444be33eb6cd29ec/tools/workspace/mypy_internal/patches/reject_double_colon.patch#L9)):*

minor:  Has this been communicated upstream; can this reference an upstream issue number?
</blockquote></details>

My hypothesis is that they would not be interested in a bug report.

The premise of [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) is that it outputs a starting point, which a human is supposed to review and amend by hand and then commit into source control.  Given our scale, we've chosen not to do that.

It's also very unusual (and wrong) to have C++ sample code in Python docstrings (#12591).  I don't think most users of stubgen would hit this problem, either.

You could imagine that upstream might accept a pull request (not issue) from us with this patch, but generally my thought is to wait until all of our pydrake documentation epic is done, and then circle back and submit anything upstream.  It's plausible that we might change how this all works, so this patch is in some sense "in beta" for a while before we might submit it.

PR #18714, opened 2023-01-31
[py planning] Remove RobotBuilder const vs non-const distinction

jwnimmer-tri:
In Python, only the mutable getters make any sense when using the builder pattern.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18714)

### DIFF
--- a/bindings/pydrake/planning/planning_py_robot_diagram.cc
+++ b/bindings/pydrake/planning/planning_py_robot_diagram.cc
@@ -54,33 +54,33 @@ void DefinePlanningRobotDiagram(py::module m) {
     {
       using Class = RobotDiagramBuilder<T>;
       constexpr auto& cls_doc = doc.RobotDiagramBuilder;
-      auto cls = DefineTemplateClassWithDefault<Class>(
-          m, "RobotDiagramBuilder", GetPyParam<T>(), cls_doc.doc);
+      auto cls = DefineTemplateClassWithDefault<Class>(m, "RobotDiagramBuilder",
+          GetPyParam<T>(),
+          (std::string(cls_doc.doc) +
+              "\n\n"
+              "Note:\n"
+              "    The C++ class defines pairs of sibling methods, e.g.,\n"
+              "    parser() vs mutable_parser(). Since Python does not use\n"
+              "    const-ness, we only provide one of the two siblings here,\n"
+              "    using the shorter method name but still allowing mutation,\n"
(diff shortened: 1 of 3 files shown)


rpoyner-tri (commented):
+@rpoyner-tri for some review.

:lgtm: 

Is there some new latent style rule afoot here?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18714#-:-NN8gax_5nADzLDzjjBq:b-u6k1c7)* status: needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri:
See https://drake.mit.edu/doxygen_cxx/group__python__bindings.html at "Mutable vs. const Method Overloads":

> When a function is overloaded only by its const-ness, choose to bind the mutable overload, not the const overload.

Strictly speaking this is slightly different -- we are not literally overloaded, because we've adorned the function names to indicate const-ness.

In some cases like a `Context`, we very much care about `context.get_state()` vs `context.get_mutable_state()` both overloads, since they have very different implications.  In the case of a builder, though, I don't think it makes sense to care about constness in Python, since a "const Builder" would be an odd duck indeed.

jwnimmer-tri:
I wonder if we should also change C++ to just use plain overloading for these, so C++ users don't need to type out `mutable_` everywhere either.  I looked at the git history in Anzu and the `mutable_` in the function name was there since 2018 when the class first appeared.

rpoyner-tri (commented):
It's probably worth another look, but needn't block this patch.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18714#-:-NN95O3lCMtUeY2P-gO4:b-bfwr0g)* status: needs at least two assigned reviewers (waiting on @jwnimmer-tri)

sherm1 (commented):
Is this all new enough that no deprecation is required for the lost `mutable_` bindings?
Otherwise makes sense to me, platform :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18714#-:-NNCmIvT1q38dbl9XxHb:b42zju)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri)

jwnimmer-tri:
> Is this all new enough that no deprecation is required for the lost mutable_ bindings?

Yes, exactly (#18650).  I should have said that in the PR overview; apologies.

PR #18718, opened 2023-02-01
[planning] Improve welded body check in CollisionChecker

calderpg-tri:
Currently, `CollisionChecker::GenerateFilteredCollisionMatrix` calls `MbP::GetBodiesWeldedTo(body)` `num_bodies^2` times to check if pairs of bodies are welded together. Replaces this by calling `MbP::GetBodiesWeldedTo(body)` once per body and saving the results in an `unordered_set` for checking later.

+@jwnimmer-tri for review, thanks!

cc @sadraddini

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718)

### DIFF
--- a/planning/collision_checker.cc
+++ b/planning/collision_checker.cc
@@ -945,6 +945,13 @@ Eigen::MatrixXi CollisionChecker::GenerateFilteredCollisionMatrix() const {
     const bool i_is_robot = IsPartOfRobot(BodyIndex(i));
 
     const Body<double>& body_i = get_body(BodyIndex(i));
+
+    // Cache the welded bodies for each body, for use in the inner loop below.
+    std::unordered_set<int> bodies_welded_to_body_i;
+    for (const auto* welded_body : plant().GetBodiesWeldedTo(body_i)) {
+      bodies_welded_to_body_i.insert(welded_body->index());
+    }
+
     const std::vector<GeometryId>& geometries_i =
         plant().GetCollisionGeometriesForBody(body_i);
 
@@ -995,13 +1002,7 @@ Eigen::MatrixXi CollisionChecker::GenerateFilteredCollisionMatrix() const {
       //  and be done.
 
       // If the body pair has a welded path between them, it should be filtered.
-      bool bodies_welded_together = false;
-      for (const auto* welded_body : plant().GetBodiesWeldedTo(body_j)) {
-        if (welded_body->index() == i) {
-          bodies_welded_together = true;
-          break;
-        }
-      }
+      const bool bodies_welded_together = bodies_welded_to_body_i.count(j) > 0;
 
       // Add the filter accordingly.
       if (collisions_filtered) {
--- a/planning/collision_checker.cc
+++ b/planning/collision_checker.cc
@@ -938,6 +938,9 @@ Eigen::MatrixXi CollisionChecker::GenerateFilteredCollisionMatrix() const {
 
   const auto& inspector = model().scene_graph().model_inspector();
 
+  // Cache the welded bodies for each body, for use in the inner loop below.
+  std::unordered_set<int> bodies_welded_to_body_cache;
+
   // For consistency, (B, B) is always filtered.
   // Loop variables below use `int` for Eigen indexing compatibility.
   for (int i = 0; i < num_bodies; ++i) {
@@ -946,10 +949,14 @@ Eigen::MatrixXi CollisionChecker::GenerateFilteredCollisionMatrix() const {
 
     const Body<double>& body_i = get_body(BodyIndex(i));
 
-    // Cache the welded bodies for each body, for use in the inner loop below.
-    std::unordered_set<int> bodies_welded_to_body_i;
-    for (const auto* welded_body : plant().GetBodiesWeldedTo(body_i)) {
-      bodies_welded_to_body_i.insert(welded_body->index());
+    // Update the cache of welded bodies if necessary.
+    // Note: if body i is welded to body i-1, we don't need to update the cache.
+    if (bodies_welded_to_body_cache.count(i) == 0) {
+      bodies_welded_to_body_cache.clear();
+
+      for (const auto* welded_body : plant().GetBodiesWeldedTo(body_i)) {
+        bodies_welded_to_body_cache.insert(welded_body->index());
+      }
     }
 
     const std::vector<GeometryId>& geometries_i =
@@ -965,9 +972,6 @@ Eigen::MatrixXi CollisionChecker::GenerateFilteredCollisionMatrix() const {
       }
 
       const Body<double>& body_j = get_body(BodyIndex(j));
-      log()->debug(
-          "Checking if collision is filtered between {} [{}] and {} [{}]",
-          GetScopedName(body_i), i, GetScopedName(body_j), j);
 
       // Check if collisions between the geometries_i are already filtered.
       bool collisions_filtered = false;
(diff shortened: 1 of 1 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @calderpg-tri)

___
*[`planning/collision_checker.cc` line 949 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-NNCcqh44u4kyiXmLUIN:-NNCcqh5CP5zaWc_Iq41:b-cuj7ke) ([raw file](https://github.com/RobotLocomotion/drake/blob/7512f897db7face305807f0efce0ca323f2b23c9/planning/collision_checker.cc#L949)):*
> ```cpp
>     const Body<double>& body_i = get_body(BodyIndex(i));
> 
>     // Cache the welded bodies for each body, for use in the inner loop below.
> ```

I have two suggestions for improvements here.  I think both are probably worth doing.

(1) The low-level one:

If we hoist `bodies_welded_to_body_i` storage out of the loop, then we can reuse it each time instead of repeatedly allocating and deallocating it.  Instead of using the default constructor here, we'd call `.clear()` instead.

(2) The high-level one:

To me, it seems probable that a set of welded bodies would frequently have a lot of numerically adjacent body indices.

Also, the relation "welded to" is transitive and symmetric.

Therefore, each time we outer-loop from one `i` body to the next, we can optimistically check whether we can avoid recalculating the (possibly huge) set of welded bodies:

* Check to see if `i` is already in the `bodies_welded_to_body_i` set that was leftover from the prior loop iteration.
* If yes, then there is no need to recalculate the set.
* If no, then clear the set and recalculate it from the plant.

WDYT?
___
*[`planning/collision_checker.cc` line 969 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-NNCbAK_-Pc93dX_sHce:-NNCbAKa1A2PZ64_dkn0:b-od0uem) ([raw file](https://github.com/RobotLocomotion/drake/blob/7512f897db7face305807f0efce0ca323f2b23c9/planning/collision_checker.cc#L969)):*
> ```cpp
>       const Body<double>& body_j = get_body(BodyIndex(j));
>       log()->debug(
>           "Checking if collision is filtered between {} [{}] and {} [{}]",
> ```

BTW I wonder if this log is buying us anything.  In all cases, the answer is already logged separately, so I'm not sure what purpose an "about to check" message serves.

calderpg-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

___
*[`planning/collision_checker.cc` line 949 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-NNCcqh44u4kyiXmLUIN:-NNCwMsH-_cOYJgnTE-E:b-7wpybw) ([raw file](https://github.com/RobotLocomotion/drake/blob/7512f897db7face305807f0efce0ca323f2b23c9/planning/collision_checker.cc#L949)):*

FWIW, there is MultibodyGraph::FindSubgraphOfWeldedBodies() already. See if that helps.

Mbp uses it to filter collisions: https://github.com/RobotLocomotion/drake/blob/v1.12.0/multibody/plant/multibody_plant.cc#L1132-L1147
</blockquote></details>

@jwnimmer-tri Done

@rpoyner-tri I suspect the best way to leverage `FindSubgraphOfWeldedBodies` would be to create a body->subgraph mapping, and then check if two bodies share the same subgraph:
```
// Flatten subgraph into mapping.
std::vector<int> body_subgraph_mapping(num_bodies);
const std::vector<std::set<BodyIndex>> subgraphs = FindSubgraphOfWeldedBodies();
for (int subgraph_id = 0; subgraph_id < subgraphs.size(); ++subgraph_id) {
  const auto& subgraph = subgraphs.at(subgraph_id);
  for (const BodyIndex& body_id : subgraph) {
    body_subgraph_mapping.at(body_id) = subgraph_id;
  }
}

...

{
  ...
  const int body_i_subgraph_id = body_subgraph_mapping.at(i);
  ...
  {
    ...
    const int body_j_subgraph_id = body_subgraph_mapping.at(j);

    const bool bodies_welded_together = body_i_subgraph_id == body_j_subgraph_id;
    ...
```
assuming it's easy to get at `FindSubgraphOfWeldedBodies`, I can try this approach in r3
___
*[`planning/collision_checker.cc` line 969 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-NNCbAK_-Pc93dX_sHce:-NND2TNbCXWGaUc1fw3S:bss4a2b) ([raw file](https://github.com/RobotLocomotion/drake/blob/7512f897db7face305807f0efce0ca323f2b23c9/planning/collision_checker.cc#L969)):*

BTW I wonder if this log is buying us anything.  In all cases, the answer is already logged separately, so I'm not sure what purpose an "about to check" message serves.
</blockquote></details>

Removed.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

___
*[`planning/collision_checker.cc` line 949 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-NNCcqh44u4kyiXmLUIN:-NNDA5Z12B0QyVmp_22a:b3m7omi) ([raw file](https://github.com/RobotLocomotion/drake/blob/7512f897db7face305807f0efce0ca323f2b23c9/planning/collision_checker.cc#L949)):*

@jwnimmer-tri Done

@rpoyner-tri I suspect the best way to leverage `FindSubgraphOfWeldedBodies` would be to create a body->subgraph mapping, and then check if two bodies share the same subgraph:
```
// Flatten subgraph into mapping.
std::vector<int> body_subgraph_mapping(num_bodies);
const std::vector<std::set<BodyIndex>> subgraphs = FindSubgraphOfWeldedBodies();
for (int subgraph_id = 0; subgraph_id < subgraphs.size(); ++subgraph_id) {
  const auto& subgraph = subgraphs.at(subgraph_id);
  for (const BodyIndex& body_id : subgraph) {
    body_subgraph_mapping.at(body_id) = subgraph_id;
  }
}

...

{
  ...
  const int body_i_subgraph_id = body_subgraph_mapping.at(i);
  ...
  {
    ...
    const int body_j_subgraph_id = body_subgraph_mapping.at(j);

    const bool bodies_welded_together = body_i_subgraph_id == body_j_subgraph_id;
    ...
```
assuming it's easy to get at `FindSubgraphOfWeldedBodies`, I can try this approach in r3
</blockquote></details>

The MultibodyGraph stuff is `internal::`, but now this module is within drake, that should be fine.

calderpg-tri (commented):
See #18723 for an alternative using `FindSubgraphsOfWeldedBodies`

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-:-NNDTLc3EReOpQ8ADfMD:b-n5ledv)* status: LGTM missing from assignee sherm1(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @sherm1)

calderpg-tri:
@drake-jenkins-bot linux-focal-clang-bazel-experimental-everything-release please

sherm1 (commented):
Platform :lgtm: in case you decide to merge this one rather than the alternative.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-:-NNE7S8l8GG5fPyYy9YY:b-d8ue9o)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @calderpg-tri)

jwnimmer-tri:
My gut feeling is that this is an improvement worth merging.   If we do #18723 afterwards to make it even faster that's great, but as-is I am content with this increment to merge.

calderpg-tri (commented):
I agree on merging this now, and doing #18723 afterwards

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18718#-:-NNEGFU20JPmvX1zU6z4:bdcsj13)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @calderpg-tri)

PR #18720, opened 2023-02-01
Check browser supports gamepads in meshcat

pathammer:
Meshcat was sometimes throwing an error in Firefox:  `Uncaught TypeError: navigator.getGamepads is not a function`

This interrupted the animation loop and it would not draw any geometry from Drake.

I think root cause is the gamepad API is moving to ["secure context only"](https://caniuse.com/?search=navigator.getgamepads). This workaround will allow the visualizer to keep working in these circumstances.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)

### DIFF
--- a/geometry/meshcat.html
+++ b/geometry/meshcat.html
@@ -38,10 +38,10 @@
 
     const clamp = (num, min, max) => Math.min(Math.max(num, min), max);
 
-    // Gamepad support - first check if the gamepad feature is available
+    // Gamepad support - first check if the gamepad feature is available.
     const gamepads_supported = !!navigator.getGamepads;
     if (!gamepads_supported) {
-      if(!window.isSecureContext) {
+      if (!window.isSecureContext) {
         console.warn("Gamepads are not supported outside of a secure context. "
         + "See https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts"
         + " for details. Some browsers support localhost and allowlists.");
--- a/geometry/meshcat.html
+++ b/geometry/meshcat.html
@@ -38,6 +38,17 @@
 
     const clamp = (num, min, max) => Math.min(Math.max(num, min), max);
 
+    // Gamepad support - first check if the gamepad feature is available
+    const gamepads_supported = !!navigator.getGamepads;
+    if (!gamepads_supported) {
+      if(!window.isSecureContext) {
+        console.warn("Gamepads are not supported outside of a secure context. "
+        + "See https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts"
+        + " for details. Some browsers support localhost and allowlists.");
+      } else {
+        console.warn("Gamepads are not supported in this browser session.");
+      }
+    }
     // See https://beej.us/blog/data/javascript-gamepad/ for a tutorial.
     var last_gamepad = {};
     function handle_gamepads() {
@@ -54,15 +65,15 @@
         // user. We truncate the floating point values to two significant
         // digits to reject this noise.
         gamepad = {
-          'index': gamepads[i].index, 
+          'index': gamepads[i].index,
           'button_values': gamepads[i].buttons.map(
-            a => clamp(Math.round(a.value * 100) / 100, 0, 1)), 
+            a => clamp(Math.round(a.value * 100) / 100, 0, 1)),
           'axes': gamepads[i].axes.map(
             a => clamp(Math.round(a * 100) / 100, -1, 1)),
         };
         break;  // Just take the first connected gamepad.
       }
-      if (this.connection && this.connection.readyState == WebSocket.OPEN && 
+      if (this.connection && this.connection.readyState == WebSocket.OPEN &&
         JSON.stringify(gamepad) !== JSON.stringify(last_gamepad)) {
         this.connection.send(MeshCat.msgpack.encode(
           { 'type': 'gamepad', 'name': '', 'gamepad': gamepad }));
@@ -76,7 +87,9 @@
       realtimeRatePanel.update(latestRealtimeRate*100, 100);
       viewer.animate()
       stats.end();
-      handle_gamepads();
+      if (gamepads_supported) {
+        handle_gamepads();
+      }
       requestAnimationFrame(animate);
     }
 


SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @pathammer and @RussTedrake)

___
*[`geometry/meshcat.html` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNDAqDy7wiDA7lXeTOb:-NNDAqDy7wiDA7lXeTOc:boesoy8) ([raw file](https://github.com/RobotLocomotion/drake/blob/88bd9cb2dbd96a09d1977f87530d6eda9d75ebdf/geometry/meshcat.html#L45)):*
> ```HTML
>     function handle_gamepads() {
>       if (navigator.getGamepads === undefined) {
>         console.warn("Gamepads not supported in this browser");
> ```

Do you anticipate this would lead to console spam?  I get the sense that gamepad availability isn't transient. Either it is, or it isn't. In that case, it would be better if it were presented once (per page load or scene or some such thing) instead of once per animation frame.

pathammer (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @SeanCurtis-TRI)

___
*[`geometry/meshcat.html` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNDAqDy7wiDA7lXeTOb:-NNDJxNl4iAIGVE6Dmay:bboz63j) ([raw file](https://github.com/RobotLocomotion/drake/blob/88bd9cb2dbd96a09d1977f87530d6eda9d75ebdf/geometry/meshcat.html#L45)):*

Do you anticipate this would lead to console spam?  I get the sense that gamepad availability isn't transient. Either it is, or it isn't. In that case, it would be better if it were presented once (per page load or scene or some such thing) instead of once per animation frame.
</blockquote></details>

That's a great point. Will fix.

sherm1:
Assigned to @SeanCurtis-TRI for review or other disposition

SeanCurtis-TRI (commented):
@pathammer Thanks for this PR. It's given me the opportunity to educate myself in directions I've otherwise ignored. I've made some requests down below (hard to believe on such a small PR, I know). But I wanted to make sure that: a) we're solving the right problem and b) we spare future users the pain that you went through in the most helpful way.

Going forward we have two options:

1. We can iterate on this based on the discussion down below.
2. If you're busy, you can foist this off on me and I'll finish getting it over the finish line.


-----------------

For other reviewers:

One of the challenges is assessing the impact of this code. Here's what worked for me.

1. Run `bazel run //geometry:meshcat_manual_test`.
   - It will provide a url like: `http://localhost:7000/`
2. Open that url in a firefox browser.
3. Obtain the machine's IP address (e.g., via `ifconfig`)
4. Open `http://10.x.x.x:7000/` in another browser tab.

At least for me, firefox deemed `localhost` as secure but not the ip-address-based host. From this, I was able to observe the hanging behavior due to the unavailable API without the fix and the console message/correction *with* the fix.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-:-NNHpNib7NKFJkWyHrrJ:bqbx09q)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @pathammer and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHv8L2C203NMtoidAf:-NNHv8L31IRtcj9GLG-v:b-jbbixa) (no related file):*
Question: Did you hit this problem with meshcat server running on `localhost` or some other host? In the name of trying to expand the console error message to provide some insight, I read this [discussion on stackoverflow](https://stackoverflow.com/questions/61867895/can-firefox-76-be-forced-to-consider-localhost-subdomains-as-secure-context-wit). Based on this (and some other light reading), depending on Firefox version and server URL, this should definitely not be a problem. But it seems the discussion provides some guidance for possible steps for re-enabling gamepads. 

Do you have any particular insight into this topic?

___
___
*[`geometry/meshcat.html` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHpjvMAmplO6z6_euc:-NNHpjvMAmplO6z6_eud:b-vl6hwc) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L43)):*
> ```HTML
>     // See https://beej.us/blog/data/javascript-gamepad/ for a tutorial.
>     var last_gamepad = {};
>     var gamepads_supported = true;
> ```

Your problem statement suggests that in a single session of running meshcat.html, if `navigator.getGamepads` won't be available, it won't be available for the entire lifespan of the meshcat session (inference based on my ignorant assumption that the context of the meshcat session won't change from secure to insecure while running).

This implementation suggests that one can *start* with an accessible `navigator.getGamepads` but lose it later.

The code also admits a third possible interpretation. Perhaps the information about whether gamepads are supported or not isn't available at the time of the defining the callback (I'm guessing such is not the case).

We should be clear about the nature of the problem we're attacking and address it directly.

If it's based purely on whether you're in a secure context, I'd suggest that we should set `gamepads_supported` at the time of declaration and be done.
```js
var gamepads_supported = navigator.getGamepads !== undefined;
```

(Or, possibly as advocated by [this site](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection):

```js
var gamepads_supported = !!navigator.getGamepads;
```
)

If we *truly* need to detect support in the callback, we should provide a comment as to why.
___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHwoxA9hg9SEN8FFTw:-NNHrG9h6pvKUVSdeOh0:b6mg7av) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*
> ```HTML
>         if (gamepads_supported) {
>           gamepads_supported = false;
>           console.warn("Gamepads not supported in this browser");
> ```

Related to the discussion above, it would be good if we could provide helpful guidance for resolving the problem. (If we can't, that's life, and what you have here is definitely better than the previous state. But we can strive for even more.)

Possibly give some direction as indicated by the link I gave above (`localhost` or adding the host to the whitelist/allowlist) would likewise be helpful. Or mentioning secure contexts so they can google it for themselves. Some small breadcrumb trail they might be able to follow to a solution.
___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNI3DUe8t6zYdo3BSvi:-NNI3DUe8t6zYdo3BSvj:b-t1af9l) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*
If we end up unable to provide further guidance in this error message and stick with a simple statement of fact, let's add the missing verb:

> Gamepads are not supported in this browser session.

(Note: I also added "session" because it may just be a URL problem more than a definitive browser problem.)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @pathammer and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNI5FB8AqxccnsoZ5IP:-NNI5FB8AqxccnsoZ5IQ:b-9usp1u) (no related file):*
@jwnimmer-tri It seems we would want to do something in the future to help catch these kinds of issues. They will come up sufficiently rarely (i.e., new API added) that I don't think it'll pop into people's minds, especially since browsers seem so inclined to just do what we want when given `localhost`. Preferences?

___

jwnimmer-tri:

@jwnimmer-tri It seems we would want to do something in the future to help catch these kinds of issues. They will come up sufficiently rarely (i.e., new API added) that I don't think it'll pop into people's minds, especially since browsers seem so inclined to just do what we want when given `localhost`. Preferences?
</blockquote></details>

I don't think there's much we can procedurally do about "Newer versions of $BROWSER change things in ways that break us".  Having users file bugs and then fixing them after the fact seems OK to me, if unsatisfying.  In this case the user actually submitted a fix, which is even more awesome than a bug report.

One option would be automated browser regression testing.  I don't think that has a positive ROI for stuff like this, though.

We could imagine having a "someone does manual testing once a month" instead, but that really shouldn't be necessary -- we should expected that our developers and users are _already_ using features regularly, and they can report things they care about.  If it's broken and nobody is using it, we are not really any worse off.

The one plausible lever we do have is to narrow our footprint of browser API use, i.e., delete our gamepad support.  If it continues to cause headaches then we can consider that, but so far it hasn't crossed that line for me.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @pathammer, @RussTedrake, and @SeanCurtis-TRI)

___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHwoxA9hg9SEN8FFTw:-NNI9bAN8SNiYgp4bI49:b-a9wvik) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*

Related to the discussion above, it would be good if we could provide helpful guidance for resolving the problem. (If we can't, that's life, and what you have here is definitely better than the previous state. But we can strive for even more.)

Possibly give some direction as indicated by the link I gave above (`localhost` or adding the host to the whitelist/allowlist) would likewise be helpful. Or mentioning secure contexts so they can google it for themselves. Some small breadcrumb trail they might be able to follow to a solution.
</blockquote></details>

I don't think lacking a suggestion message here is a defect.  The most important thing we need is that gamepad support doesn't crash the 99% of users who never use it. We should prioritize landing a resolution to that quicklyk, without trying to gold-plate the code for the 1% of users to need this feature but whose browser is defeating them.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @pathammer and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNI5FB8AqxccnsoZ5IP:-NNIGdE591KZBxwRyeVA:b-tdjqzl) (no related file):*

I don't think there's much we can procedurally do about "Newer versions of $BROWSER change things in ways that break us".  Having users file bugs and then fixing them after the fact seems OK to me, if unsatisfying.  In this case the user actually submitted a fix, which is even more awesome than a bug report.

One option would be automated browser regression testing.  I don't think that has a positive ROI for stuff like this, though.

We could imagine having a "someone does manual testing once a month" instead, but that really shouldn't be necessary -- we should expected that our developers and users are _already_ using features regularly, and they can report things they care about.  If it's broken and nobody is using it, we are not really any worse off.

The one plausible lever we do have is to narrow our footprint of browser API use, i.e., delete our gamepad support.  If it continues to cause headaches then we can consider that, but so far it hasn't crossed that line for me.
</blockquote></details>

I was thinking of the most minor effort to add a single step to `meshcat_manual_test.cc` where we provide a url using the ip address (as I did above to reproduce the error). My hesitation is that it's dependent, as you say, on all sorts of things. Which browser, which version. It would be easy to get a string of false positives and so the effort would be wasted.

However, in the absence of a *reasonable* preventative measure, addressing them as they come is the best we can do.

___
___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHwoxA9hg9SEN8FFTw:-NNIH3cq5MyvRT2tM8-l:b-hil1dq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*

I don't think lacking a suggestion message here is a defect.  The most important thing we need is that gamepad support doesn't crash the 99% of users who never use it. We should prioritize landing a resolution to that quicklyk, without trying to gold-plate the code for the 1% of users to need this feature but whose browser is defeating them.
</blockquote></details>

I agree it's not a defect. It was more intended to be a discussion for consideration. If there's something useful we can say, great. If not, c'est la vie. But it seemed now was the right time to ask and answer that question.

pathammer (commented):
NP! Your steps to reproduce look perfect. Lookout for next rev Friday.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-:-NNKeehnD19v-sNWM2Ec:bq7l69v)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHv8L2C203NMtoidAf:-NNK__0N7kyXexmdttnV:b7lsd95) (no related file):*
I first discovered it when connecting to a remote host.

I found this mdn page very informative: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

Also so much for [this *plan:*](https://hacks.mozilla.org/2020/07/securing-gamepad-api/)
> From Firefox 81, we plan to require secure context for .getGamepads() by default. To avoid significant code breakage, calling .getGamepads() will return an empty array. We will display this console warning indefinitely:

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNI5FB8AqxccnsoZ5IP:-NNKf4bTDvmtGg4FKFC9:b-2tbz9x) (no related file):*

I was thinking of the most minor effort to add a single step to `meshcat_manual_test.cc` where we provide a url using the ip address (as I did above to reproduce the error). My hesitation is that it's dependent, as you say, on all sorts of things. Which browser, which version. It would be easy to get a string of false positives and so the effort would be wasted.

However, in the absence of a *reasonable* preventative measure, addressing them as they come is the best we can do.
</blockquote></details>

I was unaware of the firefox setting `dom.securecontext.allowlist` but confirmed it works to solve this also.

We can add a little more information in the error message by checking `window.isSecureContext` and adding that clue for context.

___
___
*[`geometry/meshcat.html` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHpjvMAmplO6z6_euc:-NNKhb8E3e2NmZCKmO2o:b-e8rg2o) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L43)):*

Your problem statement suggests that in a single session of running meshcat.html, if `navigator.getGamepads` won't be available, it won't be available for the entire lifespan of the meshcat session (inference based on my ignorant assumption that the context of the meshcat session won't change from secure to insecure while running).

This implementation suggests that one can *start* with an accessible `navigator.getGamepads` but lose it later.

The code also admits a third possible interpretation. Perhaps the information about whether gamepads are supported or not isn't available at the time of the defining the callback (I'm guessing such is not the case).

We should be clear about the nature of the problem we're attacking and address it directly.

If it's based purely on whether you're in a secure context, I'd suggest that we should set `gamepads_supported` at the time of declaration and be done.
```js
var gamepads_supported = navigator.getGamepads !== undefined;
```

(Or, possibly as advocated by [this site](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection):

```js
var gamepads_supported = !!navigator.getGamepads;
```
)

If we *truly* need to detect support in the callback, we should provide a comment as to why.
</blockquote></details>

I wrote it this way to encapsulate all the gampad logic in `handle_gamepads` but I agree that basic feature detection can happen at time of declaration.
___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNI3DUe8t6zYdo3BSvi:-NNKhJMVC0q15xlj-X1P:bgb0t4n) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*

If we end up unable to provide further guidance in this error message and stick with a simple statement of fact, let's add the missing verb:

> Gamepads are not supported in this browser session.

(Note: I also added "session" because it may just be a URL problem more than a definitive browser problem.)
</blockquote></details>

Will enhance. 
___
*[`geometry/meshcat.html` line 49 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHwoxA9hg9SEN8FFTw:-NNKhYvo9HMEFQIwlGJa:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4651129598a0d3dbbddd50fc9b16451ed8a86475/geometry/meshcat.html#L49)):*

I agree it's not a defect. It was more intended to be a discussion for consideration. If there's something useful we can say, great. If not, c'est la vie. But it seemed now was the right time to ask and answer that question.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
:LGTM: Looks great!

+@jwnimmer-tri for platform review (as I'm the next on-call reviewer).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-:-NNNcn4J1ukkICEbDgsE:b-o4p6ie)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-NNHv8L2C203NMtoidAf:-NNNcSi1EKjm90oSei5S:b-e1iddt) (no related file):*
<details><summary><i>Previously, pathammer (Patrick Hammer) wrote…</i></summary><blockquote>

I first discovered it when connecting to a remote host.

I found this mdn page very informative: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

Also so much for [this *plan:*](https://hacks.mozilla.org/2020/07/securing-gamepad-api/)
> From Firefox 81, we plan to require secure context for .getGamepads() by default. To avoid significant code breakage, calling .getGamepads() will return an empty array. We will display this console warning indefinitely:
</blockquote></details>

I read that same page.  :)

___

jwnimmer-tri (commented):
+(release notes: fix) :lgtm: platform.

> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18720#-:-NNP77bM-BmpAN-YyHfb:b-d8ql2x)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake)

Issue #17663, opened 2022-08-01
Implement parsing of coupler constraints


amcastro-tri:
#17639 introduces coupler constraints between two single-dof joints. This is particularly useful to model things like your parallel jaw gripper where you'd like both finger to move in opposition. 

As pointed out by @EricCousineau-TRI, we might want to have parsing for this in a not very far future.

EricCousineau-TRI:
\cc @scpeters @azeey regarding today's discussion about SDFormat's `mimic` feature

RussTedrake:
FYI -- I received another feature request for parsing the `mimic` tag today.

RussTedrake:
A few relevant links:
- [mimic tag in urdf](http://wiki.ros.org/urdf/XML/joint)
- [mimic in sdformat](https://github.com/gazebosim/sdformat/pull/1166) -- still just a PR
- [AddCouplerConstraint](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_multibody_plant.html#a71b7c4d7b4af1c97bb6f7dae906d1088) in Drake

amcastro-tri:
@RussTedrake implemented URDF parsing in #18728.
@joemasterjohn implemented SDF parsing in #20503.

scpeters:
cross-posting from https://github.com/RobotLocomotion/drake/pull/20503#issuecomment-1858314968

sorry I didn't chime in sooner (I just noticed https://github.com/RobotLocomotion/drake/pull/20503 after it was merged), but you might consider using the `//joint/axis/mimic` / `//joint/axis2/mimic` syntax that has recently been added to SDFormat 1.11 in libsdformat14:

* http://sdformat.org/spec?ver=1.11&elem=joint#axis_mimic
* http://sdformat.org/spec?ver=1.11&elem=joint#axis2_mimic
* http://sdformat.org/tutorials?tut=mimic_proposal&cat=pose_semantics_docs&
* https://github.com/gazebosim/sdformat/pull/1166
* https://github.com/gazebosim/sdformat/pull/1333

PR #18728, opened 2023-02-02
[parsing] Add support for URDF mimic element.

RussTedrake:
Implements parsing support for the `<mimic>` element in URDF. Towards #17663.

Users could have been using existing URDF models in Drake with mimic tags that were ignored. Now those models will be simulated differently if they are being simulated with DiscreteContactSolver::kSap.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)

### DIFF
--- a/multibody/parsing/detail_urdf_parser.cc
+++ b/multibody/parsing/detail_urdf_parser.cc
@@ -463,7 +463,6 @@ void UrdfParser::ParseJoint(
     return;
   }
   WarnUnsupportedElement(*node, "calibration");
-  WarnUnsupportedElement(*node, "mimic");
   WarnUnsupportedElement(*node, "safety_controller");
 
   // Parses the parent and child link names.
@@ -533,30 +532,31 @@ void UrdfParser::ParseJoint(
   };
 
   auto plant = w_.plant;
+  std::optional<JointIndex> index{};
   if (type.compare("revolute") == 0 || type.compare("continuous") == 0) {
     throw_on_custom_joint(false);
     ParseJointLimits(node, &lower, &upper, &velocity, &acceleration, &effort);
     ParseJointDynamics(node, &damping);
-    const JointIndex index = plant->AddJoint<RevoluteJoint>(
+    index = plant->AddJoint<RevoluteJoint>(
         name, *parent_body, X_PJ,
         *child_body, std::nullopt, axis, lower, upper, damping).index();
-    Joint<double>& joint = plant->get_mutable_joint(index);
+    Joint<double>& joint = plant->get_mutable_joint(*index);
     joint.set_velocity_limits(Vector1d(-velocity), Vector1d(velocity));
     joint.set_acceleration_limits(
         Vector1d(-acceleration), Vector1d(acceleration));
   } else if (type.compare("fixed") == 0) {
     throw_on_custom_joint(false);
-    plant->AddJoint<WeldJoint>(name, *parent_body, X_PJ,
-                               *child_body, std::nullopt,
-                               RigidTransformd::Identity());
+    index = plant->AddJoint<WeldJoint>(
+        name, *parent_body, X_PJ, *child_body, std::nullopt,
+        RigidTransformd::Identity()).index();
   } else if (type.compare("prismatic") == 0) {
     throw_on_custom_joint(false);
     ParseJointLimits(node, &lower, &upper, &velocity, &acceleration, &effort);
     ParseJointDynamics(node, &damping);
-    const JointIndex index = plant->AddJoint<PrismaticJoint>(
-        name, *parent_body, X_PJ,
-        *child_body, std::nullopt, axis, lower, upper, damping).index();
-    Joint<double>& joint = plant->get_mutable_joint(index);
+    index = plant->AddJoint<PrismaticJoint>(
+        name, *parent_body, X_PJ, *child_body, std::nullopt, axis, lower,
+        upper, damping).index();
+    Joint<double>& joint = plant->get_mutable_joint(*index);
     joint.set_velocity_limits(Vector1d(-velocity), Vector1d(velocity));
     joint.set_acceleration_limits(
         Vector1d(-acceleration), Vector1d(acceleration));
@@ -568,8 +568,8 @@ void UrdfParser::ParseJoint(
   } else if (type.compare("ball") == 0) {
     throw_on_custom_joint(true);
     ParseJointDynamics(node, &damping);
-    plant->AddJoint<BallRpyJoint>(name, *parent_body, X_PJ,
-                                  *child_body, std::nullopt, damping);
+    index = plant->AddJoint<BallRpyJoint>(
+      name, *parent_body, X_PJ, *child_body, std::nullopt, damping).index();
   } else if (type.compare("planar") == 0) {
     throw_on_custom_joint(true);
     Vector3d damping_vec(0, 0, 0);
@@ -577,21 +577,21 @@ void UrdfParser::ParseJoint(
     if (dynamics_node) {
       ParseVectorAttribute(dynamics_node, "damping", &damping_vec);
     }
-    plant->AddJoint<PlanarJoint>(name, *parent_body, X_PJ,
-                                 *child_body, std::nullopt, damping_vec);
+    index = plant->AddJoint<PlanarJoint>(
+      name, *parent_body, X_PJ, *child_body, std::nullopt, damping_vec).index();
   } else if (type.compare("screw") == 0) {
     throw_on_custom_joint(true);
     ParseJointDynamics(node, &damping);
     double screw_thread_pitch;
     ParseScrewJointThreadPitch(node, &screw_thread_pitch);
-    plant->AddJoint<ScrewJoint>(name, *parent_body, X_PJ, *child_body,
-                                std::nullopt, axis, screw_thread_pitch,
-                                damping);
+    index = plant->AddJoint<ScrewJoint>(
+      name, *parent_body, X_PJ, *child_body, std::nullopt, axis,
+      screw_thread_pitch, damping).index();
   } else if (type.compare("universal") == 0) {
     throw_on_custom_joint(true);
     ParseJointDynamics(node, &damping);
-    plant->AddJoint<UniversalJoint>(name, *parent_body, X_PJ,
-                                    *child_body, std::nullopt, damping);
+    index = plant->AddJoint<UniversalJoint>(
+      name, *parent_body, X_PJ, *child_body, std::nullopt, damping).index();
   } else {
     Error(*node, fmt::format("Joint '{}' has unrecognized type: '{}'",
                              name, type));
@@ -599,6 +599,76 @@ void UrdfParser::ParseJoint(
   }
 
   joint_effort_limits->emplace(name, effort);
+
+  XMLElement* mimic_node = node->FirstChildElement("mimic");
+  if (mimic_node) {
+    if (!plant->is_discrete() ||
+        plant->get_discrete_contact_solver() != DiscreteContactSolver::kSap) {
+      Warning(
+          *mimic_node,
+          fmt::format("Joint '{}' specifies a mimic element that will be "
+                      "ignored. Mimic elements are currently only supported by "
+                      "MultibodyPlant with a discrete time step and using "
+                      "DiscreteContactSolver::kSap.",
+                      name));
+    } else {
+      std::string joint_to_mimic;
+      double gear_ratio{1.0};
+      double offset{0.0};
+      if (!ParseStringAttribute(mimic_node, "joint", &joint_to_mimic)) {
+        Error(*mimic_node,
+              fmt::format("Joint '{}' mimic element is missing the "
+                          "required 'joint' attribute.",
+                          name));
+        return;
+      }
+      if (!plant->HasJointNamed(joint_to_mimic, model_instance_)) {
+        Error(*mimic_node,
+              fmt::format("Joint '{}' mimic element specifies joint '{}' which"
+                          " does not exist.",
+                          name, joint_to_mimic));
+        return;
+      }
+      ParseScalarAttribute(mimic_node, "multiplier", &gear_ratio);
+      ParseScalarAttribute(mimic_node, "offset", &offset);
+
+      if (!index) {
+        // This can currently happen if we have a "floating" joint, which does
+        // not produce the actual QuaternionFloatingJoint above.
+        Warning(*mimic_node,
+                fmt::format("Drake only supports the mimic element for "
+                            "single-dof joints. The mimic element in joint "
+                            "'{}' will be ignored.",
+                            name));
+      } else {
+        const Joint<double>& joint0 = plant->get_joint(*index);
+        const Joint<double>& joint1 =
+            plant->GetJointByName(joint_to_mimic, model_instance_);
+        if (joint1.num_velocities() != joint0.num_velocities()) {
+          Error(*mimic_node,
+                fmt::format("Joint '{}' which has {} DOF cannot mimic "
+                            "joint '{}' which has {} DOF.",
+                            name, joint0.num_velocities(), joint_to_mimic,
+                            joint1.num_velocities()));
+          return;
+        }
+        if (joint0.num_velocities() != 1) {
+          // The URDF documentation is ambiguous as to whether multi-dof joints
+          // are supported by the mimic tag. So we only raise a warning, not an
+          // error.
+          Warning(*mimic_node,
+                  fmt::format("Drake only supports the mimic element for "
+                              "single-dof joints. The joint '{}' (with {} "
+                              "dofs) is attempting to mimic joint '{}' (with "
+                              "{} dofs). The mimic element will be ignored.",
+                              name, joint0.num_velocities(), joint_to_mimic,
+                              joint1.num_velocities()));
+        } else {
+          plant->AddCouplerConstraint(joint0, joint1, gear_ratio, offset);
+        }
+      }
+    }
+  }
 }
 
 void UrdfParser::ParseMechanicalReduction(const XMLElement& node) {
--- a/multibody/parsing/parsing_doxygen.h
+++ b/multibody/parsing/parsing_doxygen.h
@@ -106,7 +106,6 @@ one of several treaments:
 
 - `/robot/@version`
 - `/robot/joint/calibration`
-- `/robot/joint/mimic`
 - `/robot/joint/safety_controller`
 - `/robot/link/@type`
 - `/robot/link/collision/verbose`
--- a/multibody/parsing/test/detail_urdf_parser_test.cc
+++ b/multibody/parsing/test/detail_urdf_parser_test.cc
@@ -79,7 +79,9 @@ class UrdfParserTest : public test::DiagnosticPolicyTestBase {
 
  protected:
   PackageMap package_map_;
-  MultibodyPlant<double> plant_{0.0};
+  // Note: We currently use a discrete plant here to be able to test
+  // Sap-specific features like the joint 'mimic' element.
+  MultibodyPlant<double> plant_{0.1};
   SceneGraph<double> scene_graph_;
 };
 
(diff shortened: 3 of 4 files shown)


RussTedrake (commented):
@rpoyner-tri for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNHF3n42ZgXOa2MyNWD:b8qhinh)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHF6l8-JodxjgiVkXG):*
Q: does this constitute a breaking change?
___
*[`multibody/parsing/detail_urdf_parser.cc` line 606 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFBbS-sYudVuPG5X_:-NNHFBbS-sYudVuPG5Xa:b-mt6r5e) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/detail_urdf_parser.cc#L606)):*
> ```cpp
>   if (mimic_node) {
>     if (!plant->is_discrete() ||
>         plant->get_discrete_contact_solver() != DiscreteContactSolver::kSap) {
> ```

q: i didn't want to check the contact solver in the parser (because parsing first and then setting the contact solver after should be acceptable behavior).  But currently the AddCouplerConstraint method throws if the solver is not kSap.  Should we change that?

I feel that this behavior is a bit different than for other features.  For instance, joint limits aren't respected by all multibody modes, but the parser does not warn nor throw if we parse joint limits and aren't using kSap.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFjcsDBEVRsRkY2xh:-NNHFjcsDBEVRsRkY2xi:by7nin7) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L48)):*
> ```cpp
>   UrdfParserTest() {
>     plant_.RegisterAsSourceForSceneGraph(&scene_graph_);
>     plant_.set_discrete_contact_solver(DiscreteContactSolver::kSap);
> ```

Note: this is related to the question above.  I don't like making the parser behavior contact-solver dependent.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 849 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFtFIEVhjL9bgiS-5:-NNHFtFIEVhjL9bgiS-6:bscc6ss) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L849)):*
> ```cpp
>   DRAKE_EXPECT_NO_THROW(plant_.GetJointByName("revolute_joint_with_mimic"));
>   // TODO(russt): Test coupler constraint properties once constraint getters are
>   // provided by MultibodyPlant (currently a TODO in multibody_plant.h).
> ```

This makes me sad. I don't actually have a way to confirm that my gear ratios and offsets are getting set properly, short of adding simulator as a dependency to this test and running a simulation here.  Perhaps it's ok for the short term?  WDYT?

rpoyner-tri (commented):
checkpoint. Generally looks good so far. pending changes around constraints and modes and error handling.

I want to look at test coverage and some closer reading still.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNI9YWw4G63SPt-gAKK:b7j10p5)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHF6l8-JodxjgiVkXG):*

Q: does this constitute a breaking change?
</blockquote></details>

I suppose it's conservative to say "yes". I don't see any "deprecation" path that makes sense, so "breaking change" will at least get us prominent text in the release notes.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 606 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFBbS-sYudVuPG5X_:-NNI4C9dA_uHTY6eVL-L:b-a48jvx) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/detail_urdf_parser.cc#L606)):*

q: i didn't want to check the contact solver in the parser (because parsing first and then setting the contact solver after should be acceptable behavior).  But currently the AddCouplerConstraint method throws if the solver is not kSap.  Should we change that?

I feel that this behavior is a bit different than for other features.  For instance, joint limits aren't respected by all multibody modes, but the parser does not warn nor throw if we parse joint limits and aren't using kSap.
</blockquote></details>

IIRC joint limits are set/represented a bit differently than solver constraints. Is this the first time we've had a connection between explicit solver constraints and the parser? If so, I think we are justified in modifying the error reporting strategy.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFjcsDBEVRsRkY2xh:-NNI4_zX6v4klTtPMTmZ:b-romn3t) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L48)):*

Note: this is related to the question above.  I don't like making the parser behavior contact-solver dependent.
</blockquote></details>

Agreed. In any case, it may make sense to have unit tests for Sap and non-Sap. Presumably the non-Sap case would be an execution-time no-op, but maybe would issue a warning at parse time?
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 849 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFtFIEVhjL9bgiS-5:-NNI3f6-C11R01CoULDy:btpxexq) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L849)):*

This makes me sad. I don't actually have a way to confirm that my gear ratios and offsets are getting set properly, short of adding simulator as a dependency to this test and running a simulation here.  Perhaps it's ok for the short term?  WDYT?
</blockquote></details>

I think the "run a simulation" route is unnecessary heroics, and would get deleted when the getters arrive. Is there a an issue filed for "add constraint getters"? In addition to the TODO here, we could file or boost the issue.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNIqXbU7q56ctcPHX-C:-NNIqXbU7q56ctcPHX-D:bqqiz1m) (no related file):*
patch to delay error checking: https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1675370291647319?thread_ts=1657230165.865699&cid=C2WBPQDB7

___

rpoyner-tri (commented):
+(release notes: breaking change)  Changes simulation semantics of URDF `<mimic>` from being ignored to being implemented, for SAP only.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNIqmyH3vmhtk37IC_U:b-u6lk2i)* status: 5 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNIwvMz6Fp3_gso9bmG:bnfp4nl)* status: 5 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNIqXbU7q56ctcPHX-C:-NNIw4l73uUK3CQVTouJ:b-hqjoe0) (no related file):*

patch to delay error checking: https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1675370291647319?thread_ts=1657230165.865699&cid=C2WBPQDB7
</blockquote></details>

See also more discussion in that slack thread: https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1657230165865699

Per nimmer's observation there, I think the patch you have`r1` is the best we can do quickly, and we're still discussing the final disposition of error handling and how to get there. TL;DR my patch will break files that "work" now, so let's go with your patch.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 606 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFBbS-sYudVuPG5X_:-NNIzd5H-TyMZafsZRKU:b-ge3ps1) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/detail_urdf_parser.cc#L606)):*

IIRC joint limits are set/represented a bit differently than solver constraints. Is this the first time we've had a connection between explicit solver constraints and the parser? If so, I think we are justified in modifying the error reporting strategy.
</blockquote></details>

See discussion thread above and linked slack discussion. I'm disappointed but satisfied with this behavior.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 849 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFtFIEVhjL9bgiS-5:-NNJ25aTD5NFRiM80FWK:b-p794wr) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L849)):*

I think the "run a simulation" route is unnecessary heroics, and would get deleted when the getters arrive. Is there a an issue filed for "add constraint getters"? In addition to the TODO here, we could file or boost the issue.
</blockquote></details>

Filed #18732.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNIqXbU7q56ctcPHX-C:-NNK4wQI6dwsVJ45769l:b7cq6bx) (no related file):*

See also more discussion in that slack thread: https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1657230165865699

Per nimmer's observation there, I think the patch you have`r1` is the best we can do quickly, and we're still discussing the final disposition of error handling and how to get there. TL;DR my patch will break files that "work" now, so let's go with your patch.
</blockquote></details>

Sounds good (I've read the slack now, too)

___
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNHFjcsDBEVRsRkY2xh:-NNK6jBE7EfoN2jzDkTF:b8lbz5g) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b093b6225df29c7da33d003de591f929798e876/multibody/parsing/test/detail_urdf_parser_test.cc#L48)):*

Agreed. In any case, it may make sense to have unit tests for Sap and non-Sap. Presumably the non-Sap case would be an execution-time no-op, but maybe would issue a warning at parse time?
</blockquote></details>

Done.  (I've narrowed the use of kSap to the mimic-specific tests, and added a test with a mimic element that is not using Sap).

SeanCurtis-TRI (commented):
+@SeanCurtis-TRI for platform

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNMnNkt5w88XhWBH29Q:b-kwmbor)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
First pass done. See comments below.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNNArrY63D-Wketpsbq:b-nay6q)* status: 8 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 535 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN0DfT9Gq4M9F50icF:-NNN0DfT9Gq4M9F50icG:b-8gdg9c) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L535)):*
> ```cpp
> 
>   auto plant = w_.plant;
>   JointIndex index{};
> ```

nit: By Hoisting `index` outside of the chain of if/else blocks, we giving the implicit promise that it will carry a *valid* value by the time we're done evaluating the block. However, the `type == "floating"` block does not set it. It *might* throw, but we have no guarantees that it will throw.

Therefore, this should be turned into an `optional<JointIndex>` and it should be validated in the `if(mimic_node)` block before using. 

Alternatively, we can just set the index now that floating joints have actual `Joint`s.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 549 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN148j1CeXyKIn5iCc:-NNN148j1CeXyKIn5iCd:b-1s299r) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L549)):*
> ```cpp
>   } else if (type.compare("fixed") == 0) {
>     throw_on_custom_joint(false);
>     index = plant->AddJoint<WeldJoint>(name, *parent_body, X_PJ,
> ```

nit: Formatting has gotten skewed; this format typically has all parameters aligned in a block. Although the block formatting doesn't work well here and won't survive clang-fomratting.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 619 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN1m9L1pZlKwlUo7nx:-NNN1m9L1pZlKwlUo7ny:bu7aljm) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L619)):*
> ```cpp
>       double offset{0.0};
>       if (!ParseStringAttribute(mimic_node, "joint", &joint_to_mimic)) {
>         Error(*mimic_node,
> ```

BTW For my own edification, is this parsing sensitive to the ordering of the XML tags? And is that normal for URDF and/or Drake parsing? I.e., what if the joint exists *in the file* but later than the joint that mimics it?

I don't have the sense that URDF particularly has this requirement. If this is our requirement, then we should be clear about this requirement in the error message.

If this is just understood by the community at large, please ignore this.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 638 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN2Ue57IQ4RHpQZQ0H:-NNN2Ue57IQ4RHpQZQ0I:b-tgj0pk) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L638)):*
> ```cpp
>       const Joint<double>& joint1 =
>           plant->GetJointByName(joint_to_mimic, model_instance_);
>       if (joint1.num_velocities() != joint0.num_velocities()) {
> ```

BTW So, we allow a prismatic joint to mimic a revolute joint? (And vice versa.)
___
*[`multibody/parsing/detail_urdf_parser.cc` line 647 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN2i780-1owMjqLubP:-NNN2i780-1owMjqLubQ:bincaku) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L647)):*
BTW While ambiguity in the URDF documentation is one issue, the more acute issue is that Drake's couple constraint only supports 1-dof joints. Perhaps it would be more honest/precise if this comment and the message indicated such.

E.g.,, warning message along the lines of:

> Drake only supports the mimic element for single-dof joints. The joint '{}' (with {} dofs) is attempting to mimic joint '{}' (with {} dofs). The mimic element will be ignored.

This means we don't have to lawyer up vis a vis URDF, we're just stating what Drake does or does not do.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 82 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3N2z36fFvYv-EwSj:-NNN3N3-7YU6qK5b9fzd:b-bo96hf) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L82)):*
> ```cpp
>  protected:
>   PackageMap package_map_;
>   MultibodyPlant<double> plant_{0.1};
> ```

nit: Prior to this PR whether the plant was discrete or continuous was irrelevant. Now it matters. Putting a note here will be helpful for the future.

Although, that will probably be undone based on my next note (about testing against a continuous plant). But, for now, the comment would help clarify the current state of the code and its dependencies.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 331 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3m2VBLh7pWdBLRcz:-NNN3m2VBLh7pWdBLRd-:b-c6lmk9) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L331)):*
> ```cpp
> }
> 
> TEST_F(UrdfParserTest, MimicNoSap) {
> ```

nit: You're missing the test: `MimicContinuous`.

I recognize that the infrastructure doesn't lend itself well to tests which vary in this dimension. At the very least, I think a TODO (possibly with @rpoyner-tri's name?) indicating that the test is missing would be good.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 332 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3shM3dAMZQdEUVIC:-NNN3shM3dAMZQdEUVID:bgmt6jb) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L332)):*
> ```cpp
> 
> TEST_F(UrdfParserTest, MimicNoSap) {
>   EXPECT_NE(AddModelFromUrdfString(R"""(
> ```

btw: This test will break when SAP becomes the default solver. Perhaps it would be better to explicitly declare TAMSI?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 535 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN0DfT9Gq4M9F50icF:-NNNRIHs1rdFJXXo1nR5:b-938mft) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L535)):*

nit: By Hoisting `index` outside of the chain of if/else blocks, we giving the implicit promise that it will carry a *valid* value by the time we're done evaluating the block. However, the `type == "floating"` block does not set it. It *might* throw, but we have no guarantees that it will throw.

Therefore, this should be turned into an `optional<JointIndex>` and it should be validated in the `if(mimic_node)` block before using. 

Alternatively, we can just set the index now that floating joints have actual `Joint`s.
</blockquote></details>

Done.

"now that floating joints have actual `Joint`s"... yes, but not until Finalize.  Add that joint here is more of a change that I want to make/test in this PR.

I've handled the floating case.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 619 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN1m9L1pZlKwlUo7nx:-NNNSQ6p80XnZv0VazJA:bdfsh4k) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L619)):*

BTW For my own edification, is this parsing sensitive to the ordering of the XML tags? And is that normal for URDF and/or Drake parsing? I.e., what if the joint exists *in the file* but later than the joint that mimics it?

I don't have the sense that URDF particularly has this requirement. If this is our requirement, then we should be clear about this requirement in the error message.

If this is just understood by the community at large, please ignore this.
</blockquote></details>

It's true that the joint referenced here must have been declared earlier in the file. That is common in our parsers (if not the xml standard).  For instance, we reference the body by name earlier in the joint; and don't handle the case where the `link` was declared after the `joint`.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 638 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN2Ue57IQ4RHpQZQ0H:-NNNSx43E__lpEHaBRui:bb77ufg) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L638)):*

BTW So, we allow a prismatic joint to mimic a revolute joint? (And vice versa.)
</blockquote></details>

Yes.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 647 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN2i780-1owMjqLubP:-NNNV17SCIHwQhebutmC:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/detail_urdf_parser.cc#L647)):*

BTW While ambiguity in the URDF documentation is one issue, the more acute issue is that Drake's couple constraint only supports 1-dof joints. Perhaps it would be more honest/precise if this comment and the message indicated such.

E.g.,, warning message along the lines of:

> Drake only supports the mimic element for single-dof joints. The joint '{}' (with {} dofs) is attempting to mimic joint '{}' (with {} dofs). The mimic element will be ignored.

This means we don't have to lawyer up vis a vis URDF, we're just stating what Drake does or does not do.
</blockquote></details>

Done.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 82 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3N2z36fFvYv-EwSj:-NNNX7w84_o4zwrF2DUs:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L82)):*

nit: Prior to this PR whether the plant was discrete or continuous was irrelevant. Now it matters. Putting a note here will be helpful for the future.

Although, that will probably be undone based on my next note (about testing against a continuous plant). But, for now, the comment would help clarify the current state of the code and its dependencies.
</blockquote></details>

Done.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 331 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3m2VBLh7pWdBLRcz:-NNNX8yL0-_xmY1kS9E3:bu9grsw) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L331)):*

nit: You're missing the test: `MimicContinuous`.

I recognize that the infrastructure doesn't lend itself well to tests which vary in this dimension. At the very least, I think a TODO (possibly with @rpoyner-tri's name?) indicating that the test is missing would be good.
</blockquote></details>

Done. Added the TODO.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 332 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3shM3dAMZQdEUVIC:-NNNXF5mDR9kb_LDnKpv:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L332)):*

btw: This test will break when SAP becomes the default solver. Perhaps it would be better to explicitly declare TAMSI?
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
:LGTM: With one style choice. But I defer to you.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-:-NNNbwKBDMRS1yeYluvY:b8o13yh)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 635 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNNbQ098VJtkXgtsEuN:-NNNbQ098VJtkXgtsEuO:bdmiw30) ([raw file](https://github.com/RobotLocomotion/drake/blob/508883f197f7f556c9d40b9137d810b680c328d3/multibody/parsing/detail_urdf_parser.cc#L635)):*
> ```cpp
>       ParseScalarAttribute(mimic_node, "offset", &offset);
> 
>       if (!index) {
> ```

BTW While correct and succinct, I get the sense that Drake tends towards `!index.has_value()` over implicit bool conversion. (Akin to preferring to test a pointer explicitly against `nullptr`.)

RussTedrake:
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 331 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18728#-NNN3m2VBLh7pWdBLRcz:-NNNslAyBUc8-YfD642P:bq0k31v) ([raw file](https://github.com/RobotLocomotion/drake/blob/364fde89b3e4ee34bf2458c0ea004ad9d9701b68/multibody/parsing/test/detail_urdf_parser_test.cc#L331)):*

Done. Added the TODO.
</blockquote></details>

@SeanCurtis-TRI -- waiting for you to clear this one.

RussTedrake:
+@rpoyner-tri for review
+@seancurtis-tri for review

(not sure why reviewable thinks i need more reviewers)

Issue #7559, opened 2017-12-04
Drake docs appear searchable but really aren't


ggould-tri:
Liang helpfully pointed me to Drake's graphviz support today; this was especially helpful because it is nearly impossible to discover that Drake has graphviz support.

For instance, if you search Drake doxygen for "graphviz" it finds nothing.
![no-gv](https://user-images.githubusercontent.com/17596413/33578718-7d076bf6-d915-11e7-881e-01e77b04d07a.png)
If you search Drake's top-level docs it finds nothing
![no-gv2](https://user-images.githubusercontent.com/17596413/33578729-899c648e-d915-11e7-86bd-01e007f96368.png)
In fact if you search google it finds nothing
![no-gv3](https://user-images.githubusercontent.com/17596413/33578740-91ae8eea-d915-11e7-9942-d625844dc8b6.png)

A similar situation applies to other Drake concepts that are not exact class names; for example "algebraic loop" or "inverse kinematics".

The underlying issue is probably in part that (1) the search at drake.mit.edu does not reach the doxygen, and (2) the doxygen search only matches exact class and method names, not doxygen comment content or words appear in class or method names.

jamiesnape:
Googling `graphviz site:drake.mit.edu` gives me around 100 results.

ggould-tri:
Good point.  We should perhaps put a google custom search on the drake.mit.edu page.  cse.google.com generates custom search boxes for searches of that type, although iirc they have to be owned by a particular google account so we should choose carefully who creates it.

RussTedrake:
if it makes sense for me to volunteer my google account for this, i'm happy to do it.

SeanCurtis-TRI:
Can't we simply make an account for drake? And then we can use LastPass to share the drake password as necessary.

SeanCurtis-TRI:
@jwnimmer-tri Given the team and component, shall we reassign this?

jwnimmer-tri:
If you're no longer interested in working on it, then I'm content for you to assign it to me.

RussTedrake:
I’d like to upvote the priority for this.  The search is really a limitation, and gives the impression to new users that we don’t have good documentation (even though we do!).
  
Before rediscovering this thread, I was looking at Doxygen’s search offerings: https://www.doxygen.nl/manual/searching.html
[...]

PR #18737, opened 2023-02-03
Include underactuated and manipulation in "Search all of Drake"

RussTedrake:
Adds underactuated.csail.mit.edu and manipulation.csail.mit.edu to the search provided by duckduckgo. According to online discussions (like https://www.reddit.com/r/duckduckgo/comments/gcabk5/multiple_site_search_not_working/) and my testing, I believe using the OR is the recommended syntax.

+@jwnimmer-tri for feature review, please?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737)

### DIFF
--- a/doc/_includes/header.html
+++ b/doc/_includes/header.html
@@ -30,10 +30,14 @@
             </svg>
           </div>
           <div class="search-bar">
-            <form id="ddg-search-form" action="https://duckduckgo.com/" method="get">
-              <input type="text" id="search-input" name="q" placeholder="Search all of Drake…" />
-              <input type="hidden" name="sites" value="drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu" />
+            <form id="search_form" action="https://google.com/search" method="get">
+              <input type="text" id="box" name="q" placeholder="Search all of Drake…" />
             </form>
+            <script>
+              search_form.addEventListener('submit', function() {
+                box.value = box.value + ' site:drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu';
+              })
+            </script>
             <div class="search-close">
               <!-- This is an inline SVG image of an "X". -->
               <svg height="20" width="20" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 373.61 373.57">
--- a/doc/_includes/header.html
+++ b/doc/_includes/header.html
@@ -32,7 +32,7 @@
           <div class="search-bar">
             <form id="ddg-search-form" action="https://duckduckgo.com/" method="get">
               <input type="text" id="search-input" name="q" placeholder="Search all of Drake…" />
-              <input type="hidden" name="sites" value="drake.mit.edu" />
+              <input type="hidden" name="sites" value="drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu" />
             </form>
             <div class="search-close">
               <!-- This is an inline SVG image of an "X". -->
--- a/doc/_includes/header.html
+++ b/doc/_includes/header.html
@@ -31,13 +31,9 @@
           </div>
           <div class="search-bar">
             <form id="search_form" action="https://google.com/search" method="get">
-              <input type="text" id="box" name="q" placeholder="Search all of Drake…" />
+              <input type="text" name="q" placeholder="Search all of Drake…" />
+              <input type="hidden" name="q" value="site:drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu" />
             </form>
-            <script>
-              search_form.addEventListener('submit', function() {
-                box.value = box.value + ' site:drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu';
-              })
-            </script>
             <div class="search-close">
               <!-- This is an inline SVG image of an "X". -->
               <svg height="20" width="20" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 373.61 373.57">
--- a/doc/doxygen_cxx/doxygen_extra.css
+++ b/doc/doxygen_cxx/doxygen_extra.css
@@ -1132,7 +1132,7 @@ footer.site-footer>div.contain {
   display: none;
 }
 
-/* Place the duckduckgo.com search box in the upper right. */
+/* Place the google.com search box in the upper right. */
 
 #titlearea {
   border-bottom: 0;
--- a/doc/doxygen_cxx/header.html.patch
+++ b/doc/doxygen_cxx/header.html.patch
@@ -69,7 +69,7 @@
 +  </form>
 +  <form id="ddg-search-form" class="wy_form" action="https://duckduckgo.com/" method="get">
 +   <input type="text" name="q" placeholder="Search all of Drake…">
-+   <input type="hidden" name="sites" value="drake.mit.edu">
++   <input type="hidden" name="sites" value="drake.mit.edu OR site:underactuated.csail.mit.edu OR site:manipulation.csail.mit.edu" />
 +  </form>
 + </div>
 + <!--BEGIN SEARCHENGINE-->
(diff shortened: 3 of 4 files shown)


jwnimmer-tri:
Do you have some example search terms where the additional search scope helps the results?

I liked this idea in theory, but in practice it doesn't seem great for some things I tried:

I tried searching for these, and the results with the new search are much worse than the old search:
- context
- inverse kinematics
- scene graph

These were possibly slightly improved, but kind of a wash:
- motion planning
- trajectory optimization

I wonder if we'd do better by adding more links from our existing website docs into the courses -- maybe in some class overviews (to cite examples of how to use it), or even in the https://drake.mit.edu/doxygen_cxx/modules.html overview, where we link directly to the course book materials that are relevant for that specific topic.

RussTedrake:
i agree with you.  google does much better with the same search

https://www.google.com/search?q=spline+site:drake.mit.edu+OR+site:underactuated.csail.mit.edu+OR+site:manipulation.csail.mit.edu

https://www.google.com/search?q=inverse+kinematics+site%3Adrake.mit.edu+OR+site%3Aunderactuated.csail.mit.edu+OR+site%3Amanipulation.csail.mit.edu

I'm afraid to ask, but is there a reason to not use google over duckduckgo?

jwnimmer-tri:
The main reason was to avoid the administrative headache of creating and maintaining the custom search auth token.  The ddg search was OK without any special auth, but as I understand it for google we need to do extra steps to enable it, and document & track those steps in our CI bookkeeping.  All do-able, but so far that hadn't been any reason to switch.

RussTedrake:
This appears to work (without any additional infrastructure? perhaps I am missing something:
```

  <form id="searchform" action="https://google.com/search" method="get">
    <input type="text" name="q" id="box" placeholder="Search these notes..." />
  </form>
  <script>
    searchform.addEventListener('submit', function() {
      box.value = box.value + ' site:manipulation.csail.mit.edu';
    })
  </script>
  ```

RussTedrake:
I put it here: https://manipulation.csail.mit.edu/index.html

RussTedrake:
I've pushed a version using google as a second commit, for your consideration

jwnimmer-tri:
> This appears to work (without any additional infrastructure? perhaps I am missing something ...

The discussions at https://github.com/RobotLocomotion/drake/issues/7559#issuecomment-351731534 led me to think site search _required_ a cse, but reading again now that was never directly stated.  So probably this does work OK.

With my schedule this week I won't be able to review until next week.  If you're interesting in landing it sooner, maybe @sherm1 or someone would like to take over review.  The review here doesn't require any special expertise, just a willingness to rebuild the new website locally and take it for a spin.

sherm1 (commented):
I'll take it for a spin. +@sherm1

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNbuZItC_jm_LXXsz53:b-vyl8pu)* status: LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

sherm1 (commented):
So far the search looks OK but the search box has some odd behavior:


Part of the search code is showing up in the search box (that's after I've searched for something and then used the back arrow to return to the main page. In order to do another search I have to select that code to replace it with my search terms.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNcHCIpAyW90hKXQ_6c:b-99xrqy)* status: LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

RussTedrake (commented):
Yes, that's expected (but I thought tolerable) behavior?  I didn't find a better way to send to google's search on my initial inspection.

I do think that the quality of the search results, w/ or w/o underactuated and manipulation is much better than with duckduckgo.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNceyxr36burLNCs40R:b-xhlq30)* status: LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri and @sherm1)

jwnimmer-tri:
I don't think the stray text in the box is acceptable.  Searching multiple times with the back button seems like a common operation, and the stray text remaining there is confusing.  At worst, we should add a javascript to manually strip it out when the user returns to the page.

On another tack, this answer https://stackoverflow.com/a/42926071 suggests that `<input name="sitesearch" type="hidden" value="example.com">` might also work, without changing `q`.

sherm1 (commented):
I agree the search is much better. The duckduckgo version does manage to avoid leaving junk around in the search box though. Would be good to fix that if possible. Otherwise :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNcgqoh6WYV0jnhZbiN:b-d150tk)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
I found a syntax that works; I just pushed it for the main site for now.  I can cycle back later and do the two API sites if nobody beats me to it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNhs4Ew-c7fbAgdoSCa:b7v0znw)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @sherm1)

RussTedrake (commented):
Thanks Jeremy.  Should be all set now.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNiaF4h0qfLgo6GmZhS:b6hdmqg)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri and @sherm1)

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 1 of 4 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-:-NNid18X9w-TMwKhmMX9:b-vtj4vo)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake)

___
*[`doc/doxygen_cxx/header.html.patch` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-NNicqRYAb2lpnkEybn4:-NNicqRZ5L2e8CuJAsMH:bs76f10) ([raw file](https://github.com/RobotLocomotion/drake/blob/14b74842c7e81a4ead1fceb8ac60c33504eb5e10/doc/doxygen_cxx/header.html.patch#L70)):*
> ```Diff
> +  </form>
> +  <script>
> +    search_form_cpp.addEventListener('submit', function() {
> ```

nit It will be a bit before I can test this locally, but to check -- should this still be javascript?  I can't quite tell.

Feel free to dismiss if its working well for you.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`doc/doxygen_cxx/header.html.patch` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18737#-NNicqRYAb2lpnkEybn4:-NNijval45FmD0T_kahT:b-46lkrp) ([raw file](https://github.com/RobotLocomotion/drake/blob/14b74842c7e81a4ead1fceb8ac60c33504eb5e10/doc/doxygen_cxx/header.html.patch#L70)):*

nit It will be a bit before I can test this locally, but to check -- should this still be javascript?  I can't quite tell.

Feel free to dismiss if its working well for you.
</blockquote></details>

Done.  You're right, I missed that one.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18737)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake)

PR #18738, opened 2023-02-03
[multibody] Adding time-varying constraints to two Toppra APIs

### LINKED ISSUE #17131: Toppra with time-varying limits
The current implementations of [AddFrameAccelerationLimit](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_toppra.html#a4439ae2c06e737634078434d897cceaf) and [AddFrameTranslationalSpeedLimit](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_toppra.html#a5cee3da4435b14492b675df2a473f611) add Cartesian acceleration/velocity limits on a specific frame of the robot throughout the whole trajectory. These constraints are often not related to the physical capabilities of the robot (even though Franka [does have Cartesian limits](https://frankaemika.github.io/docs/control_parameters.html) - apparently for user safety) but are specific to user application. What we wish is to make Toppra accept such constraints in a non-uniform way throughout the trajectory. For instance, we need the robot end effector to slow down in a portion of the trajectory while not having such a constraint on the rest. Of course, breaking the problem into segments and solving Toppra independently is not the ideal solution as the optimal solution requires the whole trajectory to run through Toppra once.

Per discussions with @jwnimmer-tri, I think the first step is to agree on the following forms of header functions. I think of overloading the current ones in this way: 

for velocity:

```c++
  /**
   * Adds a limit on the magnitude of the translational velocity of the given
   * frame, measured and expressed  in the world frame, as a function of path
   * parameter.
   * @param constraint_frame The frame to limit the translational speed of.
   * @param upper_limit The upper translational speed limit trajectory for
   *                    constraint_frame.
   * @return The bounding box constraint that will enforce the frame
   *         translational speed limit during the backward pass.
   */
  // Note: the constraints are only added between upper_limit.start_time() and
  //       upper_limit.end_time(). The rest of the time is not constrained.
  //       A warning is displayed if upper_limit.start_time() <
  //       path.start_time() or upper_limit.end_time() > path.end_time().
  Binding<BoundingBoxConstraint> AddFrameTranslationalSpeedLimit(
      const Frame<double>& constraint_frame,
      const Trajectory<double>& upper_limit);
```

for acceleration:

```c++
  /**
   * A version of accleration limit that uses a trajectory for the upper and
   * lower limits.
   * @param constraint_frame The frame to limit the acceleration of.
   * @param lower_limit The lower acceleration limit trajectory for
   * constraint_frame.
   * @param upper_limit The upper acceleration limit trajectory for
   * constraint_frame.
   * @param discretization The discretization scheme to use for this linear
   *                       constraint. See ToppraDiscretization for details.
   * @return A pair containing the linear constraints that will enforce the
   * frame acceleration limit on the backward pass and forward pass
   * respectively.
   */
  // Note: the constraints are only added between upper_limit.start_time() and
  //       upper_limit.end_time(). The rest of the time is not constrained.
  //       A warning is displayed if upper_limit.start_time() <
  //       path.start_time() or upper_limit.end_time() > path.end_time().
  std::pair<Binding<LinearConstraint>, Binding<LinearConstraint>>
  AddFrameAccelerationLimit(const Frame<double>& constraint_frame,
                            const Trajectory<double>& lower_limit,
                            const Trajectory<double>& upper_limit,
                            ToppraDiscretization discretization =
                                ToppraDiscretization::kInterpolation);
```

SeanCurtis-TRI:
This allows setting the cartesian frame speed and acceleration limits with trajectories.

Resolves #17131 

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)

### DIFF
--- a/bindings/pydrake/multibody/optimization_py.cc
+++ b/bindings/pydrake/multibody/optimization_py.cc
@@ -200,14 +200,32 @@ PYBIND11_MODULE(optimization, m) {
             py::arg("constraint_frame"), py::arg("lower_limit"),
             py::arg("upper_limit"), cls_doc.AddFrameVelocityLimit.doc)
         .def("AddFrameTranslationalSpeedLimit",
-            &Class::AddFrameTranslationalSpeedLimit,
+            py::overload_cast<const Frame<double>&, const double&>(
+                &Class::AddFrameTranslationalSpeedLimit),
             py::arg("constraint_frame"), py::arg("upper_limit"),
-            cls_doc.AddFrameTranslationalSpeedLimit.doc)
-        .def("AddFrameAccelerationLimit", &Class::AddFrameAccelerationLimit,
+            cls_doc.AddFrameTranslationalSpeedLimit.doc_const)
+        .def("AddFrameTranslationalSpeedLimit",
+            py::overload_cast<const Frame<double>&, const Trajectory<double>&>(
+                &Class::AddFrameTranslationalSpeedLimit),
+            py::arg("constraint_frame"), py::arg("upper_limit"),
+            cls_doc.AddFrameTranslationalSpeedLimit.doc_trajectory)
+        .def("AddFrameAccelerationLimit",
+            py::overload_cast<const Frame<double>&,
+                const Eigen::Ref<const Vector6d>&,
+                const Eigen::Ref<const Vector6d>&, ToppraDiscretization>(
+                &Class::AddFrameAccelerationLimit),
             py::arg("constraint_frame"), py::arg("lower_limit"),
             py::arg("upper_limit"),
             py::arg("discretization") = ToppraDiscretization::kInterpolation,
-            cls_doc.AddFrameAccelerationLimit.doc);
+            cls_doc.AddFrameAccelerationLimit.doc_const)
+        .def("AddFrameAccelerationLimit",
+            py::overload_cast<const Frame<double>&, const Trajectory<double>&,
+                const Trajectory<double>&, ToppraDiscretization>(
+                &Class::AddFrameAccelerationLimit),
+            py::arg("constraint_frame"), py::arg("lower_limit"),
+            py::arg("upper_limit"),
+            py::arg("discretization") = ToppraDiscretization::kInterpolation,
+            cls_doc.AddFrameAccelerationLimit.doc_trajectory);
   }
 }
 }  // namespace
--- a/bindings/pydrake/multibody/test/optimization_test.py
+++ b/bindings/pydrake/multibody/test/optimization_test.py
@@ -329,15 +329,35 @@ class TestToppra(unittest.TestCase):
         constraint = toppra.AddFrameTranslationalSpeedLimit(
             constraint_frame=frame, upper_limit=1.)
         self.assertIsInstance(constraint, mp.Binding[mp.BoundingBoxConstraint])
-        constraint = toppra.AddFrameAccelerationLimit(
+
+        breaks = [path.start_time(), path.end_time()]
+
+        speed_limit_traj = PiecewisePolynomial.FirstOrderHold(
+            breaks, np.array(([[1, 1]])))
+        constraint = toppra.AddFrameTranslationalSpeedLimit(
+            constraint_frame=frame, upper_limit=speed_limit_traj)
+        self.assertIsInstance(constraint, mp.Binding[mp.BoundingBoxConstraint])
+
+        backward_con, forward_con = toppra.AddFrameAccelerationLimit(
             constraint_frame=frame, lower_limit=lower_limit,
             upper_limit=upper_limit,
             discretization=ToppraDiscretization.kCollocation)
         self.assertIsInstance(backward_con, mp.Binding[mp.LinearConstraint])
         self.assertIsInstance(forward_con, mp.Binding[mp.LinearConstraint])
-        constraint = toppra.AddFrameAccelerationLimit(
+        backward_con, forward_con = toppra.AddFrameAccelerationLimit(
             constraint_frame=frame, lower_limit=lower_limit,
             upper_limit=upper_limit,
+            discretization=ToppraDiscretization.kInterpolation)
+        self.assertIsInstance(backward_con, mp.Binding[mp.LinearConstraint])
+        self.assertIsInstance(forward_con, mp.Binding[mp.LinearConstraint])
+
+        upper_traj = PiecewisePolynomial.FirstOrderHold(breaks,
+                                                        np.ones((6, 2)))
+        lower_traj = PiecewisePolynomial.FirstOrderHold(breaks,
+                                                        -np.ones((6, 2)))
+        backward_con, forward_con = toppra.AddFrameAccelerationLimit(
+            constraint_frame=frame, lower_limit=lower_traj,
+            upper_limit=upper_traj,
             discretization=ToppraDiscretization.kInterpolation)
         self.assertIsInstance(backward_con, mp.Binding[mp.LinearConstraint])
         self.assertIsInstance(forward_con, mp.Binding[mp.LinearConstraint])
--- a/multibody/optimization/BUILD.bazel
+++ b/multibody/optimization/BUILD.bazel
@@ -308,6 +308,7 @@ drake_cc_googletest(
     deps = [
         ":toppra",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//multibody/parsing",
         "//multibody/tree",
         "//solvers:gurobi_solver",
--- a/multibody/optimization/test/toppra_test.cc
+++ b/multibody/optimization/test/toppra_test.cc
@@ -1,9 +1,11 @@
 #include "drake/multibody/optimization/toppra.h"
 
+#include <utility>
+
 #include <gtest/gtest.h>
 
-#include "drake/common/find_resource.h"
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/multibody/parsing/parser.h"
 #include "drake/multibody/tree/prismatic_joint.h"
 #include "drake/solvers/gurobi_solver.h"
@@ -18,10 +20,10 @@ class IiwaToppraTest : public ::testing::Test {
   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(IiwaToppraTest)
 
   IiwaToppraTest() : iiwa_plant_(std::make_unique<MultibodyPlant<double>>(0)) {
-    std::string file_path = FindResourceOrThrow(
-        "drake/manipulation/models/iiwa_description/iiwa7/"
-        "iiwa7_no_collision.sdf");
-    multibody::Parser(iiwa_plant_.get()).AddModels(file_path);
+    multibody::Parser(iiwa_plant_.get())
+        .AddModelsFromUrl(
+            "package://drake/manipulation/models/iiwa_description/iiwa7/"
+            "iiwa7_no_collision.sdf");
     iiwa_plant_->WeldFrames(
         iiwa_plant_->world_frame(), iiwa_plant_->GetFrameByName("iiwa_link_0"));
     iiwa_plant_->Finalize();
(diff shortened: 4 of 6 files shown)


SeanCurtis-TRI (commented):
+@mpetersen94 for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-:-NNNavf7DShyA-Nc6cAW:b-5rpxvv)* status: 1 unresolved discussion, LGTM missing from assignee mpetersen94, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @mpetersen94)

___
*[`multibody/optimization/test/toppra_test.cc` line 332 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNNazIPEFRTAqPW_jmd:-NNNazIPEFRTAqPW_jme:b-ejb3m1) ([raw file](https://github.com/RobotLocomotion/drake/blob/5de896a84fab92f5bc25f97d648015f0d302088b/multibody/optimization/test/toppra_test.cc#L332)):*
> ```cpp
> }
> 
> // TODO(SeanCurtis-TRI): Create a test to validate the time-varying impact
> ```

@mpetersen94 On this TODO and the next, I'd appreciate your thoughts on two questions:

1. How important is it to eliminate this TODO now? (or relatively soon?)
2. Would you be willing to help devise an appropriate test?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)* status: 3 unresolved discussions, LGTM missing from assignee mpetersen94, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @mpetersen94 and @SeanCurtis-TRI)

___
*[`multibody/optimization/toppra.h` line 332 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNSrjac50zZ6DNiIRQq:-NNSrjac50zZ6DNiIRQr:b-it84r4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5de896a84fab92f5bc25f97d648015f0d302088b/multibody/optimization/toppra.h#L332)):*
> ```objc
>    * or a trajectory.
>    */
>   class LimitSource {
> ```

nit I not sure why we need the LimitSource machinery.

On first glance, I would have thought that the simplest way to add this feature would be for the existing method (adding constant limits) to create a Trajectory object that always outputs a constant value, and then pass that to the new method.  That way, limits are always treated as trajectories and we don't so many helper classes and functions.  It would mean adding (for each of velocity, etc.) one new overload, moving most of the code there, and inserting a 2-line stub into the old method to call the new one.
___
*[`multibody/optimization/toppra.h` line 358 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNSrW6U91Gny68kXGw0:-NNSrW6U91Gny68kXGw1:by2z67w) ([raw file](https://github.com/RobotLocomotion/drake/blob/5de896a84fab92f5bc25f97d648015f0d302088b/multibody/optimization/toppra.h#L358)):*
> ```objc
> 
>     /* For a value in the path domain, `t`, evaluate the limit. */
>     MatrixX<double> value(double t) const {
> ```

nit This is too much code to inline into a header file (GSG "Inline Functions").

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)* status: 3 unresolved discussions, LGTM missing from assignee mpetersen94, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @mpetersen94)

___
*[`multibody/optimization/toppra.h` line 332 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNSrjac50zZ6DNiIRQq:-NNbSEaw0S3NESPrEmvE:bl58o3q) ([raw file](https://github.com/RobotLocomotion/drake/blob/5de896a84fab92f5bc25f97d648015f0d302088b/multibody/optimization/toppra.h#L332)):*

nit I not sure why we need the LimitSource machinery.

On first glance, I would have thought that the simplest way to add this feature would be for the existing method (adding constant limits) to create a Trajectory object that always outputs a constant value, and then pass that to the new method.  That way, limits are always treated as trajectories and we don't so many helper classes and functions.  It would mean adding (for each of velocity, etc.) one new overload, moving most of the code there, and inserting a 2-line stub into the old method to call the new one.
</blockquote></details>

You're right.  I went too generic.
___
*[`bindings/pydrake/multibody/optimization_py.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNbY2SzBEt7D7hEDocG:-NNbY2SzBEt7D7hEDocH:bf9bm3v) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfd60a93d5848ad055a469800149a3bbaa3bce50/bindings/pydrake/multibody/optimization_py.cc#L202)):*
> ```cpp
>             py::arg("constraint_frame"), py::arg("lower_limit"),
>             py::arg("upper_limit"), cls_doc.AddFrameVelocityLimit.doc)
>         .def("AddFrameTranslationalSpeedLimit",
> ```

@jwnimmer-tri I'd like your opinion on the bindings. My gut says that the keywords for the trajectory variant should be *different* from those of the "single-value" kind. Thoughts?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)* status: 2 unresolved discussions, LGTM missing from assignee mpetersen94, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @mpetersen94)

___
*[`bindings/pydrake/multibody/optimization_py.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNbY2SzBEt7D7hEDocG:-NNboAiA0fPDAG68LH3u:b-vuzbym) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfd60a93d5848ad055a469800149a3bbaa3bce50/bindings/pydrake/multibody/optimization_py.cc#L202)):*

@jwnimmer-tri I'd like your opinion on the bindings. My gut says that the keywords for the trajectory variant should be *different* from those of the "single-value" kind. Thoughts?
</blockquote></details>

The type system should be easily able to disambiguate, so it's not a runtime problem, just a question of clarity.

It seems to me like the semantics of the two functions are basically identical, so having the kwarg name be the same is no problem.  The only reason to differ would be if we were worried about the range-no-overlap throwing, but in that case I don't think a different name would help anyway, in practice.

I think it's basically the same a C++ GSG rule on overloading -- if the effect of a function is too confusing / different, then don't use overloading, use a different function name instead.  Since we're OK with overloading on C++ (and I think rightly so), I don't see any reason for python to be different.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)* status: 3 unresolved discussions, LGTM missing from assignee mpetersen94, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @mpetersen94)

___
*[`bindings/pydrake/multibody/optimization_py.cc` line 203 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNclcNN4vwyea9-vMxM:-NNclcNN4vwyea9-vMxN:b-lzot6s) ([raw file](https://github.com/RobotLocomotion/drake/blob/d50a04c71bab48590987d98848618345621431bc/bindings/pydrake/multibody/optimization_py.cc#L203)):*
> ```cpp
>             py::arg("upper_limit"), cls_doc.AddFrameVelocityLimit.doc)
>         .def("AddFrameTranslationalSpeedLimit",
>             py::overload_cast<const Frame<double>&, const double&>(
> ```

Blocking!

Add overloads and tests for same.

SeanCurtis-TRI (commented):
-@mpetersen94 +@hongkai-dai for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-:-NNm5drM7u7GHwpbJ3bF:b-82g5t8)* status: 2 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
Sorry for the delay, PTAL.

> Reviewed 2 of 3 files at r2, 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-:-NNyfvHm02rVnpAdt2OZ:bgvovpy)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`multibody/optimization/toppra.cc` line 321 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNxR3Ey-62oFMWfJloH:-NNxR3Ey-62oFMWfJloI:bz5dgk8) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/toppra.cc#L321)):*
> ```cpp
>   }
> 
>   const MatrixX<double> kInf =
> ```

I think it is very common that we create `kInf` as a const double
```
const double kInf = std::numeric_limits<double>::infinity();
```
If later people want to create `kInf` as a double, it will collide with this `kInf` defined here as a Matrix.

Could you rename it as `kInf_matrix`?
___
*[`multibody/optimization/toppra.cc` line 387 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNyFD3G7yv1tJI9I7aL:-NNyFD3G7yv1tJI9I7aM:bs4q9op) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/toppra.cc#L387)):*
> ```cpp
>                     path_.start_time(), path_.end_time()));
>   }
>   const Vector6d kInf =
> ```

Ditto.
___
*[`multibody/optimization/test/toppra_test.cc` line 51 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNyGeX_62bUE3ihQooX:-NNyGeX_62bUE3ihQooY:b-akjfug) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/test/toppra_test.cc#L51)):*
> ```cpp
>   // trajectory as defined by s_path (mapping time t to path parameter s). This
>   // writes to the test's allocated MbP Context.
>   SpatialVelocity<double> CalcSpatialVelocity(
> ```

I wonder if we should add this functionality to toppra, seems very useful. Maybe put a TODO here?

SeanCurtis-TRI (commented):
Comments addressed. PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-:-NOAV8nvAY0IEWkwZNns:baz6201)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/optimization/toppra.cc` line 321 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNxR3Ey-62oFMWfJloH:-NOATKP6BJxbD7CKtqJU:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/toppra.cc#L321)):*

I think it is very common that we create `kInf` as a const double
```
const double kInf = std::numeric_limits<double>::infinity();
```
If later people want to create `kInf` as a double, it will collide with this `kInf` defined here as a Matrix.

Could you rename it as `kInf_matrix`?
</blockquote></details>

Done.
___
*[`multibody/optimization/toppra.cc` line 387 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNyFD3G7yv1tJI9I7aL:-NOATP5b3lPD1yTGjNFN:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/toppra.cc#L387)):*

Ditto.
</blockquote></details>

Done.
___
*[`multibody/optimization/test/toppra_test.cc` line 51 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NNyGeX_62bUE3ihQooX:-NOATbYJ4BxLvfYsckiV:bo4fu1p) ([raw file](https://github.com/RobotLocomotion/drake/blob/f9330482683f508749c81c2be66eb7096c1a2d03/multibody/optimization/test/toppra_test.cc#L51)):*

I wonder if we should add this functionality to toppra, seems very useful. Maybe put a TODO here?
</blockquote></details>

I've added a TODO. I've taken the liberty of assigning your name to it.

ggould-tri (commented):
+(release notes: feature) 
:lgtm: with many very minor notes.

> Reviewed 1 of 4 files at r1, 1 of 3 files at r2, 2 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-:-NOB9LCs4sLfHBD3hR0d:bmkcmts)* status: 7 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`multibody/optimization/toppra.h` line 195 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOB9TTF-j62pV30zcw0:-NOB9TTF-j62pV30zcw1:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.h#L195)):*
> ```objc
>   /**
>    * Adds a limit on the magnitude of the translational velocity of the given
>    * frame, measured and expressed  in the world frame, as a function of path
> ```

typo



_Suggestion:_
```objc
expressed in
```
___
*[`multibody/optimization/toppra.h` line 196 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBAXoBB9Nf45bmX2_Q:-NOBAXoBB9Nf45bmX2_R:b-frk7ly) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.h#L196)):*
> ```objc
>    * Adds a limit on the magnitude of the translational velocity of the given
>    * frame, measured and expressed  in the world frame, as a function of path
>    * parameter.
> ```

typo?  The grammar here seems wrong.



_Suggestion:_
```objc
   * frame, measured and expressed  in the world frame, as a function of the
   * path parameters.
```
___
*[`multibody/optimization/toppra.h` line 206 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBAnPrBXF8x3n5VLmm:-NOBAnPrBXF8x3n5VLmn:bsxo7cr) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.h#L206)):*
> ```objc
>    *         doesn't overlap with [path.start_time(), path.end_time()].
>    * @note The constraints are only added between upper_limit.start_time() and
>    *       upper_limit.end_time(). The rest of the time is _not_ constrained.
> ```

nit:  I think it is the path rather than the time that is constrained?



_Suggestion:_
```objc
The rest of the path is _not_ constrained
```
___
*[`multibody/optimization/toppra.h` line 236 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBBiGXA2R6vmziupUj:-NOBBiGXA2R6vmziupUk:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.h#L236)):*
> ```objc
>           ToppraDiscretization::kInterpolation);
>   /**
>    * A version of accleration limit that uses a trajectory for the upper and
> ```

typo



_Suggestion:_
```objc
 acceleration 
```
___
*[`multibody/optimization/toppra.cc` line 283 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBEemY7XQf4rzEPXfT:-NOBEemZ4MXdrADgm5vo:b-xgferh) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.cc#L283)):*
> ```cpp
>                                                 const Eigen::MatrixXd& value) {
>   const std::vector<double> breaks{path.start_time(), path.end_time()};
>   const std::vector samples{value, value};
> ```

minor:  This is a relatively complex and performance-intensive type to hide behind inference.  Better to make it explicit.



_Suggestion:_
```cpp
std::vector<Eigen::MatrixXd> 
```
___
*[`multibody/optimization/toppra.cc` line 648 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBDOwO3Yj58XDzzZG4:-NOBDOwO3Yj58XDzzZG5:bhhfbyk) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/toppra.cc#L648)):*
> ```cpp
> 
>   return PiecewisePolynomial<double>(polynomials, t_knots);
> }
> ```

nit:  Excess vertical whitespace.



_Suggestion:_
```cpp
}

}  // namespace multibody
```
___
*[`multibody/optimization/test/toppra_test.cc` line 359 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18738#-NOBDzDeDkV3t9H5qu7y:-NOBDzDeDkV3t9H5qu7z:bs5hkwa) ([raw file](https://github.com/RobotLocomotion/drake/blob/120d349c2be6faa9840933a492077e5e89b2ce15/multibody/optimization/test/toppra_test.cc#L359)):*
> ```cpp
>                     breaks[0], breaks[1], path_start, path_end));
>   }
>   {
> ```

nit:  Inconsistent vertical whitespace.



_Suggestion:_
```cpp
  }
  
  {
```

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18738)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),hongkai-dai (waiting on @SeanCurtis-TRI)

PR #18752, opened 2023-02-07
Substitute and expand polynomial

AlexandreAmice:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)

### DIFF
--- a/common/symbolic/polynomial.cc
+++ b/common/symbolic/polynomial.cc
@@ -825,22 +825,41 @@ Polynomial Polynomial::SubstituteAndExpand(
     const std::unordered_map<Variable, Polynomial>& linear_substitutions,
     std::map<Monomial, Polynomial, internal::CompareMonomial>* substitutions)
     const {
+  for (const auto& var : indeterminates_) {
+    DRAKE_DEMAND(linear_substitutions.find(var) != linear_substitutions.cend());
+  }
+
   MapType new_polynomial_coeff_map;
   if (substitutions->empty()) {
-    substitutions->emplace(Monomial(), Polynomial());
+    substitutions->emplace(Monomial(), Polynomial(1));
   }
+
+  auto find_nearest_cached_monomial =
+      [&substitutions](const Monomial& monomial) {
+        Monomial nearest_cached_monomial =
+            (*std::prev(substitutions->lower_bound(monomial))).first;
+        while (!nearest_cached_monomial.GetVariables().IsSubsetOf(
+            monomial.GetVariables())) {
+          nearest_cached_monomial =
+              (*std::prev(substitutions->lower_bound(nearest_cached_monomial)))
+                  .first;
+        }
+        return nearest_cached_monomial;
+      };
+
   auto compute_substituted_monomial_expansion =
-      [&linear_substitutions, &substitutions](
+      [&linear_substitutions, &substitutions, &find_nearest_cached_monomial](
           const Monomial& monomial,
-          auto&& compute_substituted_monomial_expansion) {
+          auto&& compute_substituted_monomial_expansion_recursion_handle) {
         // Base case. The map substitutions is non-empty and contains the
         // monomial 1.
         if (substitutions->find(monomial) != substitutions->cend()) {
           return substitutions->at(monomial);
         }
 
-        auto [nearest_cached_monomial, ret] =
-            *std::prev(substitutions->lower_bound(monomial));
+        Monomial nearest_cached_monomial =
+            find_nearest_cached_monomial(monomial);
+        Polynomial ret = substitutions->at(nearest_cached_monomial);
 
         std::map<Variable, int> remaining_powers;
         std::map<Variable, int> cached_powers{
@@ -848,17 +867,15 @@ Polynomial Polynomial::SubstituteAndExpand(
         for (const auto& [var, power] : monomial.get_powers()) {
           if (cached_powers.find(var) != cached_powers.cend()) {
             remaining_powers.emplace(var, power - cached_powers[var]);
-          }
-        }
-        for (const auto& [var, power] : remaining_powers) {
-          if (power == 1) {
+          } else {
+            remaining_powers.emplace(var, power - 1);
             ret *= linear_substitutions.at(var);
-            remaining_powers.at(var) -= 1;
           }
         }
-        return ret * compute_substituted_monomial_expansion(
-                         Monomial(remaining_powers),
-                         compute_substituted_monomial_expansion);
+        return ret *
+               compute_substituted_monomial_expansion_recursion_handle(
+                   Monomial(remaining_powers),
+                   compute_substituted_monomial_expansion_recursion_handle);
       };
   for (const auto& [old_monomial, old_coeff] : monomial_to_coefficient_map_) {
     // If substitutions doesn't contain the current substitution create it now.
@@ -877,7 +894,8 @@ Polynomial Polynomial::SubstituteAndExpand(
           new_polynomial_coeff_map.cend()) {
         new_polynomial_coeff_map.insert({new_monomial, Expression()});
       }
-      new_polynomial_coeff_map.at(new_monomial) += new_coeff * old_coeff;
+      new_polynomial_coeff_map.at(new_monomial) +=
+          (new_coeff * old_coeff).Expand();
     }
   }
   return Polynomial{new_polynomial_coeff_map};
--- a/common/symbolic/polynomial.cc
+++ b/common/symbolic/polynomial.cc
@@ -406,9 +406,7 @@ Polynomial::Polynomial(const Expression& e, Variables indeterminates)
   DRAKE_ASSERT_VOID(CheckInvariant());
 }
 
-const Variables& Polynomial::indeterminates() const {
-  return indeterminates_;
-}
+const Variables& Polynomial::indeterminates() const { return indeterminates_; }
 
 void Polynomial::SetIndeterminates(const Variables& new_indeterminates) {
   if (new_indeterminates.IsSupersetOf(indeterminates_) &&
@@ -836,18 +834,17 @@ Polynomial Polynomial::SubstituteAndExpand(
     substitutions->emplace(Monomial(), Polynomial(1));
   }
 
-  auto find_nearest_cached_monomial =
-      [&substitutions](const Monomial& monomial) {
-        Monomial nearest_cached_monomial =
-            (*std::prev(substitutions->lower_bound(monomial))).first;
-        while (!nearest_cached_monomial.GetVariables().IsSubsetOf(
-            monomial.GetVariables())) {
-          nearest_cached_monomial =
-              (*std::prev(substitutions->lower_bound(nearest_cached_monomial)))
-                  .first;
-        }
-        return nearest_cached_monomial;
-      };
+  auto find_nearest_cached_monomial = [&substitutions](
+                                          const Monomial& monomial) {
+    auto nearest_cached_monomial_iter =
+        std::prev(substitutions->lower_bound(monomial));
+    while (!((*nearest_cached_monomial_iter).first)
+                .GetVariables()
+                .IsSubsetOf(monomial.GetVariables())) {
+      nearest_cached_monomial_iter = std::prev(nearest_cached_monomial_iter);
+    }
+    return (*nearest_cached_monomial_iter).first;
+  };
 
   auto compute_substituted_monomial_expansion =
       [&linear_substitutions, &substitutions, &find_nearest_cached_monomial](
@@ -864,27 +861,46 @@ Polynomial Polynomial::SubstituteAndExpand(
         Polynomial ret = substitutions->at(nearest_cached_monomial);
 
         std::map<Variable, int> remaining_powers;
+        std::map<Variable, int> halved_powers;
+        std::map<Variable, int> mod_2_powers;
+        bool same_monomial{true};
         const std::map<Variable, int>& cached_powers{
             nearest_cached_monomial.get_powers()};
         for (const auto& [var, power] : monomial.get_powers()) {
           if (cached_powers.find(var) != cached_powers.cend()) {
             remaining_powers.emplace(var, power - cached_powers.at(var));
+            same_monomial = false;
           } else {
-            remaining_powers.emplace(var, power - 1);
-            ret *= linear_substitutions.at(var);
+            remaining_powers.emplace(var, power);
+            halved_powers.emplace(var, static_cast<int>(std::floor(power / 2)));
+            mod_2_powers.emplace(var, power % 2);
           }
         }
-        const Monomial remaining_monomial{remaining_powers};
-        const Polynomial& remaining_substitution{
+
+        const Monomial remaining_monomials = same_monomial
+                                                 ? Monomial{halved_powers}
+                                                 : Monomial{remaining_powers};
+        const Polynomial remaining_substitution{
             compute_substituted_monomial_expansion_recursion_handle(
-                Monomial(remaining_powers),
+                remaining_monomials,
                 compute_substituted_monomial_expansion_recursion_handle)};
-        // Don't store any of the linear substitutions
-        if (remaining_monomial.total_degree() != 1) {
-          substitutions->insert_or_assign(remaining_monomial,
+        if (remaining_monomials.total_degree() != 1) {
+          substitutions->insert_or_assign(remaining_monomials,
                                           remaining_substitution);
         }
-        return ret * remaining_substitution;
+
+        if (same_monomial) {
+          ret *= pow(remaining_substitution, 2);
+          for (const auto& [var, is_odd] : mod_2_powers) {
+            if (is_odd) {
+              ret *= linear_substitutions.at(var);
+            }
+          }
+
+        } else {
+          ret *= remaining_substitution;
+        }
+        return ret;
       };
   for (const auto& [old_monomial, old_coeff] : monomial_to_coefficient_map_) {
     // If substitutions doesn't contain the current substitution create it now.
@@ -991,98 +1007,54 @@ void Polynomial::CheckInvariant() const {
   }
 }
 
-Polynomial operator-(const Polynomial& p) {
-  return -1 * p;
-}
-Polynomial operator+(Polynomial p1, const Polynomial& p2) {
-  return p1 += p2;
-}
-Polynomial operator+(Polynomial p, const Monomial& m) {
-  return p += m;
-}
-Polynomial operator+(const Monomial& m, Polynomial p) {
-  return p += m;
-}
+Polynomial operator-(const Polynomial& p) { return -1 * p; }
+Polynomial operator+(Polynomial p1, const Polynomial& p2) { return p1 += p2; }
+Polynomial operator+(Polynomial p, const Monomial& m) { return p += m; }
+Polynomial operator+(const Monomial& m, Polynomial p) { return p += m; }
 Polynomial operator+(const Monomial& m1, const Monomial& m2) {
   return Polynomial(m1) + m2;
 }
-Polynomial operator+(Polynomial p, const double c) {
-  return p += c;
-}
-Polynomial operator+(const double c, Polynomial p) {
-  return p += c;
-}
+Polynomial operator+(Polynomial p, const double c) { return p += c; }
+Polynomial operator+(const double c, Polynomial p) { return p += c; }
 Polynomial operator+(const Monomial& m, const double c) {
   return Polynomial(m) + c;
 }
 Polynomial operator+(const double c, const Monomial& m) {
   return c + Polynomial(m);
 }
-Polynomial operator+(Polynomial p, const Variable& v) {
-  return p += v;
-}
-Polynomial operator+(const Variable& v, Polynomial p) {
-  return p += v;
-}
+Polynomial operator+(Polynomial p, const Variable& v) { return p += v; }
+Polynomial operator+(const Variable& v, Polynomial p) { return p += v; }
 
-Polynomial operator-(Polynomial p1, const Polynomial& p2) {
-  return p1 -= p2;
-}
-Polynomial operator-(Polynomial p, const Monomial& m) {
-  return p -= m;
-}
+Polynomial operator-(Polynomial p1, const Polynomial& p2) { return p1 -= p2; }
+Polynomial operator-(Polynomial p, const Monomial& m) { return p -= m; }
 Polynomial operator-(const Monomial& m, Polynomial p) {
   return p = -1 * p + m;  // p' = m - p = -1 * p + m.
 }
 Polynomial operator-(const Monomial& m1, const Monomial& m2) {
   return Polynomial(m1) - m2;
 }
-Polynomial operator-(Polynomial p, const double c) {
-  return p -= c;
-}
-Polynomial operator-(const double c, Polynomial p) {
-  return p = -p + c;
-}
+Polynomial operator-(Polynomial p, const double c) { return p -= c; }
+Polynomial operator-(const double c, Polynomial p) { return p = -p + c; }
 Polynomial operator-(const Monomial& m, const double c) {
   return Polynomial(m) - c;
 }
 Polynomial operator-(const double c, const Monomial& m) {
   return c - Polynomial(m);
 }
-Polynomial operator-(Polynomial p, const Variable& v) {
-  return p -= v;
-}
+Polynomial operator-(Polynomial p, const Variable& v) { return p -= v; }
 Polynomial operator-(const Variable& v, const Polynomial& p) {
   return Polynomial(v, p.indeterminates()) - p;
 }
 
-Polynomial operator*(Polynomial p1, const Polynomial& p2) {
-  return p1 *= p2;
-}
-Polynomial operator*(Polynomial p, const Monomial& m) {
-  return p *= m;
-}
-Polynomial operator*(const Monomial& m, Polynomial p) {
-  return p *= m;
-}
-Polynomial operator*(const double c, Polynomial p) {
-  return p *= c;
-}
-Polynomial operator*(Polynomial p, const double c) {
-  return p *= c;
-}
-Polynomial operator*(const Monomial& m, double c) {
-  return Polynomial(m) * c;
-}
-Polynomial operator*(double c, const Monomial& m) {
-  return c * Polynomial(m);
-}
-Polynomial operator*(Polynomial p, const Variable& v) {
-  return p *= v;
-}
-Polynomial operator*(const Variable& v, Polynomial p) {
-  return p *= v;
-}
+Polynomial operator*(Polynomial p1, const Polynomial& p2) { return p1 *= p2; }
+Polynomial operator*(Polynomial p, const Monomial& m) { return p *= m; }
+Polynomial operator*(const Monomial& m, Polynomial p) { return p *= m; }
+Polynomial operator*(const double c, Polynomial p) { return p *= c; }
+Polynomial operator*(Polynomial p, const double c) { return p *= c; }
+Polynomial operator*(const Monomial& m, double c) { return Polynomial(m) * c; }
+Polynomial operator*(double c, const Monomial& m) { return c * Polynomial(m); }
+Polynomial operator*(Polynomial p, const Variable& v) { return p *= v; }
+Polynomial operator*(const Variable& v, Polynomial p) { return p *= v; }
 
 Polynomial operator/(Polynomial p, const double v) {
   for (auto& item : p.monomial_to_coefficient_map_) {
@@ -1130,8 +1102,9 @@ ostream& operator<<(ostream& os, const Polynomial& p) {
 namespace Eigen {
 namespace numext {
 template <>
-bool equal_strict(const drake::symbolic::Polynomial& x,
-                  const drake::symbolic::Polynomial& y) {
+bool equal_strict(
+    const drake::symbolic::Polynomial& x,
+    const drake::symbolic::Polynomial& y) {
   return static_cast<bool>(x == y);
 }
 }  // namespace numext
--- a/common/symbolic/polynomial.cc
+++ b/common/symbolic/polynomial.cc
@@ -406,7 +406,9 @@ Polynomial::Polynomial(const Expression& e, Variables indeterminates)
   DRAKE_ASSERT_VOID(CheckInvariant());
 }
 
-const Variables& Polynomial::indeterminates() const { return indeterminates_; }
+const Variables& Polynomial::indeterminates() const {
+  return indeterminates_;
+}
 
 void Polynomial::SetIndeterminates(const Variables& new_indeterminates) {
   if (new_indeterminates.IsSupersetOf(indeterminates_) &&
@@ -851,8 +853,8 @@ Polynomial Polynomial::SubstituteAndExpand(
       [&linear_substitutions, &substitutions, &find_nearest_cached_monomial](
           const Monomial& monomial,
           auto&& compute_substituted_monomial_expansion_recursion_handle) {
-        // Base case. The map substitutions is non-empty and contains the
-        // monomial 1.
+        // Base case. Since the map substitutions is non-empty and contains the
+        // monomial 1 this will always be hit.
         if (substitutions->find(monomial) != substitutions->cend()) {
           return substitutions->at(monomial);
         }
@@ -862,20 +864,27 @@ Polynomial Polynomial::SubstituteAndExpand(
         Polynomial ret = substitutions->at(nearest_cached_monomial);
 
         std::map<Variable, int> remaining_powers;
-        std::map<Variable, int> cached_powers{
+        const std::map<Variable, int>& cached_powers{
             nearest_cached_monomial.get_powers()};
         for (const auto& [var, power] : monomial.get_powers()) {
           if (cached_powers.find(var) != cached_powers.cend()) {
-            remaining_powers.emplace(var, power - cached_powers[var]);
+            remaining_powers.emplace(var, power - cached_powers.at(var));
           } else {
             remaining_powers.emplace(var, power - 1);
             ret *= linear_substitutions.at(var);
           }
         }
-        return ret *
-               compute_substituted_monomial_expansion_recursion_handle(
-                   Monomial(remaining_powers),
-                   compute_substituted_monomial_expansion_recursion_handle);
+        const Monomial remaining_monomial{remaining_powers};
+        const Polynomial& remaining_substitution{
+            compute_substituted_monomial_expansion_recursion_handle(
+                Monomial(remaining_powers),
+                compute_substituted_monomial_expansion_recursion_handle)};
+        // Don't store any of the linear substitutions
+        if (remaining_monomial.total_degree() != 1) {
+          substitutions->insert_or_assign(remaining_monomial,
+                                          remaining_substitution);
+        }
+        return ret * remaining_substitution;
       };
   for (const auto& [old_monomial, old_coeff] : monomial_to_coefficient_map_) {
     // If substitutions doesn't contain the current substitution create it now.
@@ -982,54 +991,98 @@ void Polynomial::CheckInvariant() const {
   }
 }
 
-Polynomial operator-(const Polynomial& p) { return -1 * p; }
-Polynomial operator+(Polynomial p1, const Polynomial& p2) { return p1 += p2; }
-Polynomial operator+(Polynomial p, const Monomial& m) { return p += m; }
-Polynomial operator+(const Monomial& m, Polynomial p) { return p += m; }
+Polynomial operator-(const Polynomial& p) {
+  return -1 * p;
+}
+Polynomial operator+(Polynomial p1, const Polynomial& p2) {
+  return p1 += p2;
+}
+Polynomial operator+(Polynomial p, const Monomial& m) {
+  return p += m;
+}
+Polynomial operator+(const Monomial& m, Polynomial p) {
+  return p += m;
+}
 Polynomial operator+(const Monomial& m1, const Monomial& m2) {
   return Polynomial(m1) + m2;
 }
-Polynomial operator+(Polynomial p, const double c) { return p += c; }
-Polynomial operator+(const double c, Polynomial p) { return p += c; }
+Polynomial operator+(Polynomial p, const double c) {
+  return p += c;
+}
+Polynomial operator+(const double c, Polynomial p) {
+  return p += c;
+}
 Polynomial operator+(const Monomial& m, const double c) {
   return Polynomial(m) + c;
 }
 Polynomial operator+(const double c, const Monomial& m) {
   return c + Polynomial(m);
 }
-Polynomial operator+(Polynomial p, const Variable& v) { return p += v; }
-Polynomial operator+(const Variable& v, Polynomial p) { return p += v; }
+Polynomial operator+(Polynomial p, const Variable& v) {
+  return p += v;
+}
+Polynomial operator+(const Variable& v, Polynomial p) {
+  return p += v;
+}
 
-Polynomial operator-(Polynomial p1, const Polynomial& p2) { return p1 -= p2; }
-Polynomial operator-(Polynomial p, const Monomial& m) { return p -= m; }
+Polynomial operator-(Polynomial p1, const Polynomial& p2) {
+  return p1 -= p2;
+}
+Polynomial operator-(Polynomial p, const Monomial& m) {
+  return p -= m;
+}
 Polynomial operator-(const Monomial& m, Polynomial p) {
   return p = -1 * p + m;  // p' = m - p = -1 * p + m.
 }
 Polynomial operator-(const Monomial& m1, const Monomial& m2) {
   return Polynomial(m1) - m2;
 }
-Polynomial operator-(Polynomial p, const double c) { return p -= c; }
-Polynomial operator-(const double c, Polynomial p) { return p = -p + c; }
+Polynomial operator-(Polynomial p, const double c) {
+  return p -= c;
+}
+Polynomial operator-(const double c, Polynomial p) {
+  return p = -p + c;
+}
 Polynomial operator-(const Monomial& m, const double c) {
   return Polynomial(m) - c;
 }
 Polynomial operator-(const double c, const Monomial& m) {
   return c - Polynomial(m);
 }
-Polynomial operator-(Polynomial p, const Variable& v) { return p -= v; }
+Polynomial operator-(Polynomial p, const Variable& v) {
+  return p -= v;
+}
 Polynomial operator-(const Variable& v, const Polynomial& p) {
   return Polynomial(v, p.indeterminates()) - p;
 }
 
-Polynomial operator*(Polynomial p1, const Polynomial& p2) { return p1 *= p2; }
-Polynomial operator*(Polynomial p, const Monomial& m) { return p *= m; }
-Polynomial operator*(const Monomial& m, Polynomial p) { return p *= m; }
-Polynomial operator*(const double c, Polynomial p) { return p *= c; }
-Polynomial operator*(Polynomial p, const double c) { return p *= c; }
-Polynomial operator*(const Monomial& m, double c) { return Polynomial(m) * c; }
-Polynomial operator*(double c, const Monomial& m) { return c * Polynomial(m); }
-Polynomial operator*(Polynomial p, const Variable& v) { return p *= v; }
-Polynomial operator*(const Variable& v, Polynomial p) { return p *= v; }
+Polynomial operator*(Polynomial p1, const Polynomial& p2) {
+  return p1 *= p2;
+}
+Polynomial operator*(Polynomial p, const Monomial& m) {
+  return p *= m;
+}
+Polynomial operator*(const Monomial& m, Polynomial p) {
+  return p *= m;
+}
+Polynomial operator*(const double c, Polynomial p) {
+  return p *= c;
+}
+Polynomial operator*(Polynomial p, const double c) {
+  return p *= c;
+}
+Polynomial operator*(const Monomial& m, double c) {
+  return Polynomial(m) * c;
+}
+Polynomial operator*(double c, const Monomial& m) {
+  return c * Polynomial(m);
+}
+Polynomial operator*(Polynomial p, const Variable& v) {
+  return p *= v;
+}
+Polynomial operator*(const Variable& v, Polynomial p) {
+  return p *= v;
+}
 
 Polynomial operator/(Polynomial p, const double v) {
   for (auto& item : p.monomial_to_coefficient_map_) {
@@ -1077,9 +1130,8 @@ ostream& operator<<(ostream& os, const Polynomial& p) {
 namespace Eigen {
 namespace numext {
 template <>
-bool equal_strict(
-    const drake::symbolic::Polynomial& x,
-    const drake::symbolic::Polynomial& y) {
+bool equal_strict(const drake::symbolic::Polynomial& x,
+                  const drake::symbolic::Polynomial& y) {
   return static_cast<bool>(x == y);
 }
 }  // namespace numext
--- a/common/symbolic/polynomial.cc
+++ b/common/symbolic/polynomial.cc
@@ -406,7 +406,9 @@ Polynomial::Polynomial(const Expression& e, Variables indeterminates)
   DRAKE_ASSERT_VOID(CheckInvariant());
 }
 
-const Variables& Polynomial::indeterminates() const { return indeterminates_; }
+const Variables& Polynomial::indeterminates() const {
+  return indeterminates_;
+}
 
 void Polynomial::SetIndeterminates(const Variables& new_indeterminates) {
   if (new_indeterminates.IsSupersetOf(indeterminates_) &&
@@ -822,9 +824,30 @@ Polynomial& Polynomial::AddProduct(const Expression& coeff, const Monomial& m) {
 }
 
 Polynomial Polynomial::SubstituteAndExpand(
-      std::unordered_map<Monomial, Polynomial>* substitution) {
+    const std::unordered_map<Variable, Polynomial>& linear_substitutions,
+    std::map<Monomial, Polynomial, internal::CompareMonomial>* substitutions)
+    const {
   MapType new_polynomial_coeff_map;
+  auto compute_monomial_expansion = [&linear_substitutions, &substitutions](
+                                        auto&& compute_monomial_expansion,
+                                        const Monomial& monomial) {
+    if (substitutions->empty()) {
+      substitutions->emplace(Monomial(), Polynomial());
+    }
+    if (substitutions-> find(monomial) != substitutions -> cend()) {
+      return substitutions->at(monomial);
+    }
+    if (monomial.total_degree() == 1) {
+      return linear_substitutions.at(*monomial.GetVariables().cbegin());
+    }
 
+    auto nearest_cached_monomial = std::prev(substitutions->lower_bound(monomial));
+  };
+  for (const auto& [cur_monomial, cur_coeff] : monomial_to_coefficient_map_) {
+    // If substitutions doesn't contain the current substitution create it now.
+    if (substitutions->find(cur_monomial) == substitutions->end()) {
+    }
+  }
 }
 
 Polynomial Polynomial::Expand() const {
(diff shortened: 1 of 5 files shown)


AlexandreAmice (commented):
+@hongkai-dai for feature review. This method is useful when you need to perform the same substitutions into several similar polynomials. I use it to create the rationals in the trajectory certifier.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NNhfmtW1-RTTTVxpJa5:b-umb3y2)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
Checkpoint

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NNiVSrRAn6u6Ap96MFf:b13s2ul)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.h` line 264 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiO60jEKZmkva0kPIx:-NNiO60kD2OkP5j3QABZ:bm0nz5f) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.h#L264)):*
> ```objc
>   /// Substitute the monomials of this polynomial with new polynomial
>   /// expressions and expand the polynomial to the monomial basis. For example,
>   /// consider the substitution x = a(1-y) + by into the polynomial x¹⁴ +
> ```

"+ by " is redundant?
___
*[`common/symbolic/polynomial.h` line 269 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiOYKmDheciYKCjUnA:-NNiOYKmDheciYKCjUnB:b-a9e8my) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.h#L269)):*
> ```objc
>   /// map to avoid recomputing very high powers.
>   /// @param linear_substitutions The substitutions of every indeterminate with the
>   /// new desired expression. This map must containt this -> indeterminates.
> ```

s/containt/contain
___
*[`common/symbolic/polynomial.cc` line 844 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiRsH6C6Yo8eEJJtwO:-NNiRsH6C6Yo8eEJJtwP:bw48jek) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L844)):*
> ```cpp
>             monomial.GetVariables())) {
>           nearest_cached_monomial =
>               (*std::prev(substitutions->lower_bound(nearest_cached_monomial)))
> ```

Now `nearest_cached_monomial` is already a key in `substitutions`, why continue to call `lower_bound`? I would think that you can just find the iterator just before `nearest_cached_monomial`. Namely if you store the iterator of nearest_cached_monomial, you then only need to call the previous iterator.
___
*[`common/symbolic/polynomial.cc` line 855 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiSyX58kcJGYWftmL6:-NNiSyX58kcJGYWftmL7:b-fzliw2) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L855)):*
> ```cpp
>           auto&& compute_substituted_monomial_expansion_recursion_handle) {
>         // Base case. The map substitutions is non-empty and contains the
>         // monomial 1.
> ```

I think `substitutions` contains `monomial`, not "contains the monomial 1"?
___
*[`common/symbolic/polynomial.cc` line 865 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiTK-J23G81rkfAd3x:-NNiTK-J23G81rkfAd3y:b-lbehnt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L865)):*
> ```cpp
> 
>         std::map<Variable, int> remaining_powers;
>         std::map<Variable, int> cached_powers{
> ```

`const std::map<Variable, int>& cached_powers`, use const reference here to avoid unnecessary copy.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 6 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.cc` line 824 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNjLZ18D3fBy-6zeKgx:-NNjLZ18D3fBy-6zeKgy:bezw26u) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L824)):*
> ```cpp
> }
> 
> Polynomial Polynomial::SubstituteAndExpand(
> ```

I like that this function caches the intermediate result. I wonder if we caches unnecessary intermediate result in the current implementation.

As far as I can see, to substitute the monomial pow(x, 4), it first starts with pow(x, 0), and then go to pow(x, 1), pow(x, 2), pow(x, 3), pow(x, 4). Namely the degree is incremented by 1 in every recursion

I think it would be faster if we multiplies the degree by 2 in every recursion, namely we go to pow(x, 1), pow(x, 2), pow(x, 4). Every time we find the nearest monomial to the power(var, degree / 2), and then take the square. 

Using pow(x, 3) * pow(y, 4) as an example, we can compute it as

square( pow(x, 1) * pow(y, 2))   *   pow(x, 1)
We can then compute pow(x, 1) * pow(y, 2) as
square(pow(y, 1))     *   pow(x, 1)

I think this allows us to take logarithmic number of recursions, instead of linear number of recursions.

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 6 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`common/symbolic/polynomial.h` line 264 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiO60jEKZmkva0kPIx:-NNmC9SwALMl0hWWw76y:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.h#L264)):*

"+ by " is redundant?
</blockquote></details>

Done.
___
*[`common/symbolic/polynomial.h` line 269 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiOYKmDheciYKCjUnA:-NNmC9Gu1lUhJMeuPLAm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.h#L269)):*

s/containt/contain
</blockquote></details>

Done.
___
*[`common/symbolic/polynomial.cc` line 824 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNjLZ18D3fBy-6zeKgx:-NNmBddm1u5h9QT0gWnI:bg7jzq3) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L824)):*

I like that this function caches the intermediate result. I wonder if we caches unnecessary intermediate result in the current implementation.

As far as I can see, to substitute the monomial pow(x, 4), it first starts with pow(x, 0), and then go to pow(x, 1), pow(x, 2), pow(x, 3), pow(x, 4). Namely the degree is incremented by 1 in every recursion

I think it would be faster if we multiplies the degree by 2 in every recursion, namely we go to pow(x, 1), pow(x, 2), pow(x, 4). Every time we find the nearest monomial to the power(var, degree / 2), and then take the square. 

Using pow(x, 3) * pow(y, 4) as an example, we can compute it as

square( pow(x, 1) * pow(y, 2))   *   pow(x, 1)
We can then compute pow(x, 1) * pow(y, 2) as
square(pow(y, 1))     *   pow(x, 1)

I think this allows us to take logarithmic number of recursions, instead of linear number of recursions.
</blockquote></details>

I thought of that as well. While it takes fewer recursions, it is not obvious to me that it saves time in actually computing the expansion since we have to compute the intermediate expansions for the squares anyway. Do you want me to implement the change?
___
*[`common/symbolic/polynomial.cc` line 844 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiRsH6C6Yo8eEJJtwO:-NNmCJpX8ml1peS1XBVw:bfqfwnt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L844)):*

Now `nearest_cached_monomial` is already a key in `substitutions`, why continue to call `lower_bound`? I would think that you can just find the iterator just before `nearest_cached_monomial`. Namely if you store the iterator of nearest_cached_monomial, you then only need to call the previous iterator.
</blockquote></details>

Consider the case where `substitutions` has as keys `[1, x x*x*y]` and we call `lower_bound` for the key `x*x*y`. The return would be `x*x*y` which actually does not help us expand the `x*x*z`. 

And alternative implementation would be to enforce the invariant on `substitutions` that requires that for any multivariate monomial, the component univariate monomials must be contained. I worry that maintaining this invariant could be expensive.
___
*[`common/symbolic/polynomial.cc` line 855 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiSyX58kcJGYWftmL6:-NNmDXIZ1ZE-Pnz6vKNW:bjiknj3) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L855)):*

I think `substitutions` contains `monomial`, not "contains the monomial 1"?
</blockquote></details>

Yes you are correct. What I intended was for the comment to convey that this base case is guaranteed to be hit eventually since we will continue reducing the powers of `monomial` at most until the monomial 1 is hit.
___
*[`common/symbolic/polynomial.cc` line 865 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiTK-J23G81rkfAd3x:-NNmDz36EzmrNVbb-g7Y:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L865)):*

`const std::map<Variable, int>& cached_powers`, use const reference here to avoid unnecessary copy.
</blockquote></details>

Done.

hongkai-dai (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 2 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.cc` line 824 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNjLZ18D3fBy-6zeKgx:-NNp8qvy2DcAFzBT9in4:b-d1748u) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L824)):*
>  since we have to compute the intermediate expansions for the squares anyway

Not sure if I get this. Let's consider a toy problem, to replace `x` in pow(x, 4) with `x = 1+y`.  The current implementation will compute (1+y), pow(1+y, 2), pow(1+y, 3), pow(1+y, 4), and `substitution` will store the mapping
```
1 -> 1
x -> 1+y
x² -> 1+ 2y + y²
x³ -> 1+ 3y + 3y² + y³
x⁴ -> 1 + 4y + 6y² + 4y³ + y⁴
```
I think if we use the square idea, it has two benefits
1. We don't need to compute x³ -> 1+ 3y + 3y² + y³. Hence we only need logarithmic number of elements.
2. I think in the current computation, when it computes x⁴ = x * x³ = (1+y) * (1+y)³, here (1+y)³ is stored as an expanded polynomial (1 + 3y + 3y² + y³). Then the code will compute x * x³ as (1+y) * (1 + 3y + 3y² + y³) = 1 + 3y + 3y² + y³ + y + 3y² + 3y³ + y⁴, notice that there are 8 terms before we collect the coefficient for the same monomial (and hence we need to do computation to collect the coefficients). On the other hand if we take the square approach, it does x⁴ = pow(x², 2) = (1+2y+y²)² = 1 + 4y² + y⁴ + 4y + 2y² + 4y³, notice that there are only 6 terms before we collect the coefficient for the same monomial, hence I think by using square operation rather than multiplying polynomials, we also save some computation.

I am completely OK if you think the square idea is unnecessary, but just want to make sure we both agree on the pros and cons for the two approaches.
___
*[`common/symbolic/polynomial.cc` line 844 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiRsH6C6Yo8eEJJtwO:-NNp9Tj1Ax6Bu6tb6SHX:b1i7ctk) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L844)):*

Consider the case where `substitutions` has as keys `[1, x x*x*y]` and we call `lower_bound` for the key `x*x*y`. The return would be `x*x*y` which actually does not help us expand the `x*x*z`. 

And alternative implementation would be to enforce the invariant on `substitutions` that requires that for any multivariate monomial, the component univariate monomials must be contained. I worry that maintaining this invariant could be expensive.
</blockquote></details>

I am not sure if we are talking about the same thing.

I am suggesting that if `nearest_cached_monomial `is already in `substitutions.key()`, then
```
substitutions->lower_bound(nearest_cached_monomial)
```
is the same as
```
nearest_cached_monomial_iterator
```
(assuming nearest_cached_monomial_iterator is the iterator for nearest_cached_monomial), hence calling `lower_bound` is redundant here. Did I miss something?
___
*[`common/symbolic/polynomial.cc` line 855 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiSyX58kcJGYWftmL6:-NNpBWhx0r_ChecANl0N:b4izmc8) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L855)):*

Yes you are correct. What I intended was for the comment to convey that this base case is guaranteed to be hit eventually since we will continue reducing the powers of `monomial` at most until the monomial 1 is hit.
</blockquote></details>

I see what you mean, thanks for the explanation!

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 2 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`common/symbolic/polynomial.cc` line 824 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNjLZ18D3fBy-6zeKgx:-NNtQmtv7KkR--grpay1:b9budqe) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L824)):*

>  since we have to compute the intermediate expansions for the squares anyway

Not sure if I get this. Let's consider a toy problem, to replace `x` in pow(x, 4) with `x = 1+y`.  The current implementation will compute (1+y), pow(1+y, 2), pow(1+y, 3), pow(1+y, 4), and `substitution` will store the mapping
```
1 -> 1
x -> 1+y
x² -> 1+ 2y + y²
x³ -> 1+ 3y + 3y² + y³
x⁴ -> 1 + 4y + 6y² + 4y³ + y⁴
```
I think if we use the square idea, it has two benefits
1. We don't need to compute x³ -> 1+ 3y + 3y² + y³. Hence we only need logarithmic number of elements.
2. I think in the current computation, when it computes x⁴ = x * x³ = (1+y) * (1+y)³, here (1+y)³ is stored as an expanded polynomial (1 + 3y + 3y² + y³). Then the code will compute x * x³ as (1+y) * (1 + 3y + 3y² + y³) = 1 + 3y + 3y² + y³ + y + 3y² + 3y³ + y⁴, notice that there are 8 terms before we collect the coefficient for the same monomial (and hence we need to do computation to collect the coefficients). On the other hand if we take the square approach, it does x⁴ = pow(x², 2) = (1+2y+y²)² = 1 + 4y² + y⁴ + 4y + 2y² + 4y³, notice that there are only 6 terms before we collect the coefficient for the same monomial, hence I think by using square operation rather than multiplying polynomials, we also save some computation.

I am completely OK if you think the square idea is unnecessary, but just want to make sure we both agree on the pros and cons for the two approaches.
</blockquote></details>

Yes that makes sense actually. Thanks for pushing back! Hopefully this will speed things up.
___
*[`common/symbolic/polynomial.cc` line 844 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNiRsH6C6Yo8eEJJtwO:-NNt6Hms5auB34qkc1Kh:b-eijqnx) ([raw file](https://github.com/RobotLocomotion/drake/blob/5945d3cfb80390ccd5a12e577b77849c9eb74d4c/common/symbolic/polynomial.cc#L844)):*

I am not sure if we are talking about the same thing.

I am suggesting that if `nearest_cached_monomial `is already in `substitutions.key()`, then
```
substitutions->lower_bound(nearest_cached_monomial)
```
is the same as
```
nearest_cached_monomial_iterator
```
(assuming nearest_cached_monomial_iterator is the iterator for nearest_cached_monomial), hence calling `lower_bound` is redundant here. Did I miss something?
</blockquote></details>

I see what you mean. I changed it to just the iterator to make stepping backward more efficient.

hongkai-dai (commented):
Checkpoint.

> Reviewed 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NNuHUmg6kcApEmrML9L:by96gdd)* status: 6 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.h` line 279 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNu5apk5i6843qqr1ok:-NNu5apk5i6843qqr1ol:bbzh8rh) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.h#L279)):*
> ```objc
>   /// passed substitution map is consistent.
>   [[nodiscard]] Polynomial SubstituteAndExpand(
>       const std::unordered_map<Variable, Polynomial>& linear_substitutions,
> ```

I am not sure why this variable is called "linear_substitutions". What is the "linear" part? Since this substitutes each indeterminate, what do you think if we call it "indeterminate_substitutions"?
___
*[`common/symbolic/polynomial.h` line 281 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNu659_-5aLNK7AKvfa:-NNu659_-5aLNK7AKvfb:bhc300l) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.h#L281)):*
> ```objc
>       const std::unordered_map<Variable, Polynomial>& linear_substitutions,
>       std::map<Monomial, Polynomial, internal::CompareMonomial>* substitutions)
>       const;
> ```

Add a empty line below so that this "SubstituteAndExpand" is separated from "Expand" function with an empty line?
___
*[`common/symbolic/polynomial.cc` line 834 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuFwo17vDd7wKexChn:-NNuFwo17vDd7wKexCho:b-vcgdl6) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L834)):*
> ```cpp
>   MapType new_polynomial_coeff_map;
>   if (substitutions->empty()) {
>     substitutions->emplace(Monomial(), Polynomial(1));
> ```

I think we can also insert `linear_substitutions` into `substitutions` here? Namely
```cc
for (const auto&[var, poly] : linear_substitutions) {
  substitutions->emplace(Monomial(var), poly);
}
```
___
*[`common/symbolic/polynomial.cc` line 849 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNu7BDk3TNRsPYZdoBs:-NNu7BDk3TNRsPYZdoBt:bwtr096) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L849)):*
> ```cpp
>   };
> 
>   auto compute_substituted_monomial_expansion =
> ```

Sorry for being pedantic, but I think it is better to document the algorithm in this function.
___
*[`common/symbolic/polynomial.cc` line 866 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuDVGA3vgeVzbkfukV:-NNuDVGA3vgeVzbkfukW:bfamgxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L866)):*
> ```cpp
>         std::map<Variable, int> halved_powers;
>         std::map<Variable, int> mod_2_powers;
>         bool same_monomial{true};
> ```

Document this variable? same_monomial is true if `nearest_cached_monomial` contains the same set of variables as `monomial`.

BTW I don't think `same_monomial` is a proper name here. When I first saw this variable name I thought `same_monomial` means if nearest_cached_monomial == monomial. How about renaming it as`nearest_shares_same_variables`?
___
*[`common/symbolic/polynomial.cc` line 870 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuGwJa5ersKJ5CRo8z:-NNuGwJa5ersKJ5CRo9-:b-4y8cvy) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L870)):*
> ```cpp
>             nearest_cached_monomial.get_powers()};
>         for (const auto& [var, power] : monomial.get_powers()) {
>           if (cached_powers.find(var) != cached_powers.cend()) {
> ```

Sorry for being slow, I don't understand why we have two cases based on `same_monomial`. I think we can always write `monomial` as `monomial` = neares_cached_monomial * (power(halved), 2) * mod_2, where both halved and mod_2 are monomials, and then substitute halved and mod_2 recursively.

hongkai-dai (commented):
Sorry for the delay the delay. Almost there. PTAL.

> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NO2kYlJBCtUH26ec2J6:b-u743og)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.h` line 279 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2j4shCoSb90Nf3FdZ:-NO2j4shCoSb90Nf3Fd_:bki5xa9) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.h#L279)):*
> ```objc
>   /// substitutions = {x: y, x²: 2y}. To ensure correct results, ensure that the
>   /// passed substitution map is consistent.
>   [[nodiscard]] Polynomial SubstituteAndExpand(
> ```

BTW, do you think it makes sense to support the substitution without expansion? It seems to me that we only call `Expand()` in the end of the function `SubstituteAndExpand`. I imagine we could also do
```cc
[[nodiscard]] Polynomial Substitute(const std::unordered_map<Variable, Polynomial>& indeterminate_substitution, bool expand, std::ma<Monomial, Polynomial, internal::CompareMonomial>* substitutions);
```
___
*[`common/symbolic/polynomial.cc` line 834 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuFwo17vDd7wKexChn:-NNygISR7LDARaRbtUvN:b-8gxb24) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L834)):*

I see two options. Either this line goes through and puts all the `linear_substitutions` into `substitutions` and then we always perform our substitutions by looking in the map `substitutions` or this line simply ensures that the base case of `substitutions.at(Monomial()) = Polynomial(1)` is reachable. 

I prefer the latter as if we go with the former choice, we should simply require the user to pass `substitutions` with the linear monomials already in the map. I want to avoid this as this feels like a foot-gun where a user defines inconsistent substitutions such as `{x: a, x*x: 2a}`.
</blockquote></details>

From reading the code, it seems that you want to keep `substitutions` to never contain the degree-1 monomial in `indeterminate_substitutions`, even after calling this SubstituteAndExpand function. 

The reason I propose to include the `indeterminate_substitutions` in `substitutions`, is that `SubstituteAndExpand` can then check if repeated calls of this function uses the same `indeterminate_substitution`. What I want to avoid is that the user calls `SubstituteAndExpand` for one substitution of indeterminates, and then call the same function again for a different set of substitutions, then `substitutions` would contain inconsistent result, here is a wrong use case
```cc
// substitute x with 1 + y
std::map<Monomial, Polynomial> substitutions;
// After this line, substitutions contains 
// x^2 -> 1 + 2y + y*y
// x^3 -> 1 + 3y + 3*y*y+y^3
(x * x * x).SubstituteAndExpand({x : 1 + y}, &substitutions);
// After this line, we substitute x with 1-y, substitutions will update its key of x^2, but leave x^3 unchanged. Now it contains inconsistent result.
// x^2 -> 1 - 2y + y*y
// x^3 -> 1 + 3y + 3y*y + y^3
(x*x).SubstituteAndExpand({x: 1-y}, &substitutions);
```

I know you said that you don't want to check the inconsistency in `substitutions`, but it seems to me that checking inconsistency is easy in this case. If we first called
```
SubstituteAndExpand(indeterminate_substitution1, &substitions);
```
 with `indeterminate_substitution1` and store `indeterminate_substitution1` in `substitutions`, then next time if we call this same function with a another `indeterminate_substitution2`, as
```
SubstituteAndExpand(indeterminate_substitution2, &substitutions);
```
we can check if all the 1-degree monomial in `substitutions` (which are copied from `indeterminate_substitution1` is the same as `indeterminate_substitution2`, and throw an error if they are different. We can do so with the following pseudo-code
```cc
void CheckSubstituteAndExpand(const std::unordered_map<Variable, Polynomial>& indeterminate_substitution, const std::map<MOnomial, Polynomial, internal::CompareMonomial>& substitutions) {
if (!substitutions.empty()) {
// First make sure that substitions contains the same number of indeterminates as indeterminate_substitution
int num_degree_1_monomials = 0;
for (const auto& [monomial, poly] : substitutions) {
  if (monomial.total_degree() == 1) {
    num_degree_1_monomials++;
  }
  if (num_degree_1_monomial != indeterminate_substitution) {
    throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions contain {} indeterminates, but indeterminate_substitution contain {} indeterminates, they should be equal), num_degree_1_monomial, indeterminate_substitution.size()));
  }
  for (const auto& [indeterminate, poly] : indeterminate_substitution) {
    auto it = substitutions.find(symbolic::Monomial(indeterminate));
    if (it == substitutions.end()) {
      throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions doesn't contain the monomial {}", indeterminate));
    }
    if (it->second.EqualTo(poly.Expand())) {
      throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions map {} to {}, but indeterminate_substitution maps the same monomial to {}", indeterminate, it->second, poly));
    }
  }
}
return true;
}
```
We can then call `CheckSubstitutionAndExpand` in `DRAKE_DEMAND`, then in the release mode this code function is ignored.

What do you think? If you find this unnecessary, feel free to ignore it.
___
*[`common/symbolic/polynomial.cc` line 870 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuGwJa5ersKJ5CRo8z:-NO2ide_CWZZBeJiNI9q:b-cwbvvw) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L870)):*

I added some documentation on why we would not want to neccessarily do that expansion. Essentially, we want to use the cached monomials as much as possible, so we don't want to do the successive squaring method until we have fully exhausted the cached monomial expansion. Let me know if I should make this clearer.
</blockquote></details>

Thanks for the documentation, it is very clear.
___
*[`common/symbolic/polynomial.cc` line 867 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNyh28m05kV_VmDJzr8:-NNyh28m05kV_VmDJzr9:b-5ap0ny) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L867)):*
> ```cpp
>             find_nearest_cached_monomial(monomial);
> 
>         // If the nearest cached monomial is 1, then we do not have a cached
> ```

Thanks a lot for the documentation, it is very clear.
___
*[`common/symbolic/polynomial.cc` line 904 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2kFtxCzdWhym39gXI:-NO2kFtxCzdWhym39gXJ:b-nl8tq7) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L904)):*
> ```cpp
>           }
>           substitutions->insert_or_assign(halved_monomials_squared,
>                                           pow(halved_power_substitution, 2));
> ```

Does `pow` Expand the polynomial also? I would think you will need to call 
```cc
pow(halved_power_substitution, 2).Expand()
```
___
*[`common/symbolic/test/polynomial_test.cc` line 1089 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2ikWY9ZpHbR1Qwkas:-NO2ikWY9ZpHbR1Qwkat:b5t8esj) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/test/polynomial_test.cc#L1089)):*
> ```cpp
> TEST_F(SymbolicPolynomialTest, SubstituteAndExpandTest) {
>   std::map<Monomial, Polynomial, internal::CompareMonomial> substitutions;
>   std::unordered_map<Variable, Polynomial> linear_substitutions;
> ```

BTW, could you rename this `linear_substitutions` to `indeterminate_substitution`?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.cc` line 905 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO3Q6n05zKOxogSwiMl:-NO3Q6n05zKOxogSwiMm:b-3cuhks) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L905)):*
> ```cpp
>           substitutions->insert_or_assign(halved_monomials_squared,
>                                           pow(halved_power_substitution, 2));
>           return ret * substitutions->at(halved_monomials_squared);
> ```

I actually expect this lambda function `compute_substituted_monomial_expansion` to  always add the substituted polynomial into `substitutions` before calling the return argument (which guarantees that `substitutions` will contain the cached result for this `monomial`). Also to avoid unnecessary copying the returned polynomial, I would suggest to add the return type (const Polynomial&) to this lambda function.
```cc
auto compute_substituted_monomial_expansion =
      [&indeterminate_substitution, &substitutions,
       &find_nearest_cached_monomial](
          const Monomial& monomial,
          auto&& compute_substituted_monomial_expansion_recursion_handle) -> const Polynomial& {

...
auto emplace_result =  substitutions->emplace(monomial, ret * substitutions->at(halved_monomials_squared);
return emplace_result.first->second;
}
```
which returns a const reference, and hence avoid copying the symbolic polynomial.
___
*[`common/symbolic/polynomial.cc` line 928 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO3PTkX9bHVfTlOfnbA:-NO3PTkX9bHVfTlOfnbB:b-ld93du) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L928)):*
> ```cpp
>           }
>           return substitutions->at(nearest_cached_monomial) *
>                  remaining_substitution;
> ```

I think we need to call Expand() here?
```cc
return (substitutions->at(nearest_cached_monomial) * remaining_substitution).Expand()
```

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`common/symbolic/polynomial.h` line 279 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2j4shCoSb90Nf3FdZ:-NO60Ro-0QmY-zzBOiT9:b9mb5td) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.h#L279)):*

BTW, do you think it makes sense to support the substitution without expansion? It seems to me that we only call `Expand()` in the end of the function `SubstituteAndExpand`. I imagine we could also do
```cc
[[nodiscard]] Polynomial Substitute(const std::unordered_map<Variable, Polynomial>& indeterminate_substitution, bool expand, std::ma<Monomial, Polynomial, internal::CompareMonomial>* substitutions);
```
</blockquote></details>

I believe it only really makes sense to have this function if we do the expansion. If you just perform the substitution without the expansion there would be no difference between this method and just converting the polynomial to an expression and calling `Substitute`.

Where this method becomes relevant is when we want to efficiently perform the expansion after the substitution.
___
*[`common/symbolic/polynomial.cc` line 834 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NNuFwo17vDd7wKexChn:-NO61V121zOZxBLqMa4n:bjlr856) ([raw file](https://github.com/RobotLocomotion/drake/blob/0acd89edb9c8ce53b8465eb5a7ea0563a392c334/common/symbolic/polynomial.cc#L834)):*

From reading the code, it seems that you want to keep `substitutions` to never contain the degree-1 monomial in `indeterminate_substitutions`, even after calling this SubstituteAndExpand function. 

The reason I propose to include the `indeterminate_substitutions` in `substitutions`, is that `SubstituteAndExpand` can then check if repeated calls of this function uses the same `indeterminate_substitution`. What I want to avoid is that the user calls `SubstituteAndExpand` for one substitution of indeterminates, and then call the same function again for a different set of substitutions, then `substitutions` would contain inconsistent result, here is a wrong use case
```cc
// substitute x with 1 + y
std::map<Monomial, Polynomial> substitutions;
// After this line, substitutions contains 
// x^2 -> 1 + 2y + y*y
// x^3 -> 1 + 3y + 3*y*y+y^3
(x * x * x).SubstituteAndExpand({x : 1 + y}, &substitutions);
// After this line, we substitute x with 1-y, substitutions will update its key of x^2, but leave x^3 unchanged. Now it contains inconsistent result.
// x^2 -> 1 - 2y + y*y
// x^3 -> 1 + 3y + 3y*y + y^3
(x*x).SubstituteAndExpand({x: 1-y}, &substitutions);
```

I know you said that you don't want to check the inconsistency in `substitutions`, but it seems to me that checking inconsistency is easy in this case. If we first called
```
SubstituteAndExpand(indeterminate_substitution1, &substitions);
```
 with `indeterminate_substitution1` and store `indeterminate_substitution1` in `substitutions`, then next time if we call this same function with a another `indeterminate_substitution2`, as
```
SubstituteAndExpand(indeterminate_substitution2, &substitutions);
```
we can check if all the 1-degree monomial in `substitutions` (which are copied from `indeterminate_substitution1` is the same as `indeterminate_substitution2`, and throw an error if they are different. We can do so with the following pseudo-code
```cc
void CheckSubstituteAndExpand(const std::unordered_map<Variable, Polynomial>& indeterminate_substitution, const std::map<MOnomial, Polynomial, internal::CompareMonomial>& substitutions) {
if (!substitutions.empty()) {
// First make sure that substitions contains the same number of indeterminates as indeterminate_substitution
int num_degree_1_monomials = 0;
for (const auto& [monomial, poly] : substitutions) {
  if (monomial.total_degree() == 1) {
    num_degree_1_monomials++;
  }
  if (num_degree_1_monomial != indeterminate_substitution) {
    throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions contain {} indeterminates, but indeterminate_substitution contain {} indeterminates, they should be equal), num_degree_1_monomial, indeterminate_substitution.size()));
  }
  for (const auto& [indeterminate, poly] : indeterminate_substitution) {
    auto it = substitutions.find(symbolic::Monomial(indeterminate));
    if (it == substitutions.end()) {
      throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions doesn't contain the monomial {}", indeterminate));
    }
    if (it->second.EqualTo(poly.Expand())) {
      throw std::runtime_error(fmt::format("CheckSubstituteAndExpand(): substitutions map {} to {}, but indeterminate_substitution maps the same monomial to {}", indeterminate, it->second, poly));
    }
  }
}
return true;
}
```
We can then call `CheckSubstitutionAndExpand` in `DRAKE_DEMAND`, then in the release mode this code function is ignored.

What do you think? If you find this unnecessary, feel free to ignore it.
</blockquote></details>

This feels a little bit like a foot-gun. To me, the method should either assume full responsibility for checking the consistency of `substitutions` or assume no responsibility at all for `substitutions`. Implementing this check may give the user a false sense of security. 

To me, the intended use of this method is when a user wishes to expand many polynomials with the same `indeterminates_substitution`, the user should first pass an empty `substitutions` map and then keep passing the same map for subsequent calls. 
___
*[`common/symbolic/polynomial.cc` line 904 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2kFtxCzdWhym39gXI:-NO62jV213XgeLZ0vpsv:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L904)):*

Does `pow` Expand the polynomial also? I would think you will need to call 
```cc
pow(halved_power_substitution, 2).Expand()
```
</blockquote></details>

Done.
___
*[`common/symbolic/polynomial.cc` line 905 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO3Q6n05zKOxogSwiMl:-NO64H-s0JwUn0DyXyD0:bhe3sjk) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L905)):*

I actually expect this lambda function `compute_substituted_monomial_expansion` to  always add the substituted polynomial into `substitutions` before calling the return argument (which guarantees that `substitutions` will contain the cached result for this `monomial`). Also to avoid unnecessary copying the returned polynomial, I would suggest to add the return type (const Polynomial&) to this lambda function.
```cc
auto compute_substituted_monomial_expansion =
      [&indeterminate_substitution, &substitutions,
       &find_nearest_cached_monomial](
          const Monomial& monomial,
          auto&& compute_substituted_monomial_expansion_recursion_handle) -> const Polynomial& {

...
auto emplace_result =  substitutions->emplace(monomial, ret * substitutions->at(halved_monomials_squared);
return emplace_result.first->second;
}
```
which returns a const reference, and hence avoid copying the symbolic polynomial.

</blockquote></details>

Done. I now ensure that this lambda always places the expansion into `substitution` (if appropriate) before returning. I believe this avoids a copy the way I did it but please let me know if I am wrong.
___
*[`common/symbolic/polynomial.cc` line 928 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO3PTkX9bHVfTlOfnbA:-NO64YYr5Uqcha0wOZKc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.cc#L928)):*

I think we need to call Expand() here?
```cc
return (substitutions->at(nearest_cached_monomial) * remaining_substitution).Expand()
```
</blockquote></details>

Done.
___
*[`common/symbolic/test/polynomial_test.cc` line 1089 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2ikWY9ZpHbR1Qwkas:-NO6Ftpp8_kOcbI4Gx2_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/test/polynomial_test.cc#L1089)):*

BTW, could you rename this `linear_substitutions` to `indeterminate_substitution`?
</blockquote></details>

Done.

hongkai-dai (commented):
:lgtm: +@ggould-tri for platform review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NOApmhY-hJZF-uLzAqo:b-svovrs)* status: 3 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @ggould-tri)

___
*[`common/symbolic/polynomial.cc` line 862 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOAn-4iBqdHeN64YK7d:-NOAn-4iBqdHeN64YK7e:b-kpw7a7) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.cc#L862)):*
> ```cpp
>     // just find the substitution.
>     if (monomial.total_degree() == 1) {
>       return indeterminate_substitution.at(*monomial.GetVariables().begin());
> ```

Does `indeterminate_substitution.at(var)` always return an <i>expanded</i> polynomial? From the lambda function name `compute_substituted_monomial_expansion` I would expect an expanded polynomial.

It doesn't affect the final result from `SubstituteAndExpand`.
___
*[`common/symbolic/test/polynomial_test.cc` line 1089 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOApZDdDyMpjdD_e8Z_:-NOApZDdDyMpjdD_e8Za:b-w9x9s4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/test/polynomial_test.cc#L1089)):*
> ```cpp
> TEST_F(SymbolicPolynomialTest, SubstituteAndExpandTest) {
>   std::map<Monomial, Polynomial, internal::CompareMonomial> substitutions;
>   std::unordered_map<Variable, Polynomial> indeterminate_substitution;
> ```

Could you add a test case, where `indeterminate_substitution` contains un-expanded polynomials, and check that the return from `SubstituteAndExpand` is expanded?

ggould-tri (commented):
+(release notes: feature) 

> Reviewed 1 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NOAxfpd1XHRBdDPytBh:b-jeoipj)* status: 11 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOAzbME2iRuOxunAthA:-NOAzbME2iRuOxunAthB:b6p5wdz) (no related file):*
FYI, for reference in my comments below, here is what the doxygen block looks like right now (r6) for this change:


___
___
*[`common/symbolic/polynomial.h` line 269 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB--Q60mmInUpPgKbT:-NOB--Q60mmInUpPgKbU:b-tf9s36) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L269)):*
> ```objc
>   /// map to avoid recomputing very high powers.
>   /// @param indeterminate_substitution The substitutions of every indeterminate
>   /// with the new desired expression. This map must contain this ->
> ```

minor:  This doesn't render very clearly in doxygen.  Perhaps try something like this:



_Suggestion:_
```objc
  /// with the new desired expression. This map must contain each element
  /// of `indeterminates()`.
```
___
*[`common/symbolic/polynomial.h` line 271 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB-VAQBv1V4q5afa7a:-NOB-VAQBv1V4q5afa7b:b-nplhyk) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L271)):*
> ```objc
>   /// with the new desired expression. This map must contain this ->
>   /// indeterminates.
>   /// @param substitutions A map caching the higher order expansions of the
> ```

minor:  Doxygen supports marking that a parameter is used for both input and output, like this:



_Suggestion:_
```objc
@param[in,out]
```
___
*[`common/symbolic/polynomial.h` line 271 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB4Q-09FcqBovB-N0f:-NOB4Q-09FcqBovB-N0g:bvwkrsq) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L271)):*
> ```objc
>   /// with the new desired expression. This map must contain this ->
>   /// indeterminates.
>   /// @param substitutions A map caching the higher order expansions of the
> ```

It is not clear to me what I should pass in for this value if I am calling this function. I assume that it should be the empty map in most cases.  If so, perhaps it should be defaulted, either directly or by making this parameter a `std::optional` defaulted to `std::nullopt`?

It is also not clear what valid values are here.  I gather from reading the tests that the same `substitutions` object can be validly reused across different `Polynomial`s despite its mutation, but not across different `indeterminate_substition`s?  If so, that must be documented (or checked).

Alternatively, the documentation might be clearer if you included an example of how the `substitutions` parameter is meant to be used/reused.
___
*[`common/symbolic/polynomial.h` line 273 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOAzpC-Ck2nEajgVHWZ:-NOAzpC-Ck2nEajgVHW_:b-3y9cdb) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L273)):*
> ```objc
>   /// @param substitutions A map caching the higher order expansions of the
>   /// substitutions.
>   /// @out substitution will change as the substituted monomials are expanded.
> ```

minor:  `@out` is not a doxygen keyword; `substitution` is not the name of a parameter.



_Suggestion:_
```objc
  /// 
  /// `substitutions` will change as the substituted monomials are expanded.
```
___
*[`common/symbolic/polynomial.h` line 276 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB0csN3P2mga2Vaugx:-NOB0csN3P2mga2Vaugy:bdpxggj) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L276)):*
> ```objc
>   ///
>   /// Note that this function is NOT responsible for ensuring that @param
>   /// substitution is consistent i.e. this method will not throw an error if
> ```

typo:  There is no parameter called `substitution`.  Do you mean `substitutions` or `indeterminate_substitution`?



_Code quote:_
```objc
  /// Note that this function is NOT responsible for ensuring that @param
  /// substitution is consistent i.e. this method will not throw an error if
```
___
*[`common/symbolic/polynomial.cc` line 837 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB3x1E-m_Aurvli5Ca:-NOB3x1E-m_Aurvli5Cb:b-ooinuk) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.cc#L837)):*
> ```cpp
>   substitutions->insert_or_assign(Monomial(), Polynomial(1));
> 
>   auto find_nearest_cached_monomial = [&substitutions](
> ```

minor:  The meaning of "nearest" here is unclear.  Consider putting a comment on this lambda explaining what it returns.
___
*[`common/symbolic/polynomial.cc` line 838 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB293j4pRj-RmIlWIc:-NOB293j4pRj-RmIlWId:bukvmdp) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.cc#L838)):*
> ```cpp
> 
>   auto find_nearest_cached_monomial = [&substitutions](
>                                           const Monomial& monomial) {
> ```

nit:  This might be slightly more readable if you put the linebreak before the lambda?  Author's discretion.



_Suggestion:_
```cpp
  auto find_nearest_cached_monomial =
    [&substitutions](const Monomial& monomial) -> const Monomial& {
```
___
*[`common/symbolic/polynomial.cc` line 866 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB2s3q5cS9vsXY23nR:-NOB2s3q5cS9vsXY23nS:b-lei73c) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.cc#L866)):*
> ```cpp
>     // Base case. Since the map substitutions is non-empty and contains the
>     // monomial 1 this will always be hit.
>     if (substitutions->find(monomial) != substitutions->cend()) {
> ```

If this were always being hit, the rest of this function would be unreachable.  Can you clarify?



_Code quote:_
```cpp
this will always be hit
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 10 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`common/symbolic/polynomial.h` line 279 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NO2j4shCoSb90Nf3FdZ:-NOC6qzr7GKWzlepkFD2:bx40ouy) ([raw file](https://github.com/RobotLocomotion/drake/blob/d84a7e03a213c193ae1c2e51b7b93391bde26424/common/symbolic/polynomial.h#L279)):*

I believe it only really makes sense to have this function if we do the expansion. If you just perform the substitution without the expansion there would be no difference between this method and just converting the polynomial to an expression and calling `Substitute`.

Where this method becomes relevant is when we want to efficiently perform the expansion after the substitution.
</blockquote></details>

Got it, thanks for the explanation.

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 3 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri and @hongkai-dai)

___
*[`common/symbolic/polynomial.h` line 271 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOB4Q-09FcqBovB-N0f:-NOBgrybCWqCmQfiCLtz:bulggv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.h#L271)):*

It is not clear to me what I should pass in for this value if I am calling this function. I assume that it should be the empty map in most cases.  If so, perhaps it should be defaulted, either directly or by making this parameter a `std::optional` defaulted to `std::nullopt`?

It is also not clear what valid values are here.  I gather from reading the tests that the same `substitutions` object can be validly reused across different `Polynomial`s despite its mutation, but not across different `indeterminate_substition`s?  If so, that must be documented (or checked).

Alternatively, the documentation might be clearer if you included an example of how the `substitutions` parameter is meant to be used/reused.
</blockquote></details>

I gave substitutions the default value of `nullptr` so that users who do not wish to reuse the map do not need to pass the argument. I also added some documentation.
___
*[`common/symbolic/polynomial.cc` line 862 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOAn-4iBqdHeN64YK7d:-NOB8UTX1ps5zJoJglJj:bz1khaz) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/polynomial.cc#L862)):*

Does `indeterminate_substitution.at(var)` always return an <i>expanded</i> polynomial? From the lambda function name `compute_substituted_monomial_expansion` I would expect an expanded polynomial.

It doesn't affect the final result from `SubstituteAndExpand`.
</blockquote></details>

This is a very good point and actually suddenly makes me inclined to store the expanded `indeterminate_substitution` in the map `substitutions`
___
*[`common/symbolic/test/polynomial_test.cc` line 1089 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOApZDdDyMpjdD_e8Z_:-NOFkfjM5hNk8LK7NfJU:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5e4fbdbb9fdf36abbc8f769a3c0b726151a0b4e5/common/symbolic/test/polynomial_test.cc#L1089)):*

Could you add a test case, where `indeterminate_substitution` contains un-expanded polynomials, and check that the return from `SubstituteAndExpand` is expanded?
</blockquote></details>

Done.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @hongkai-dai)

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @hongkai-dai)

___
*[`common/symbolic/polynomial.cc` line 829 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOFuhFH0-a1ROYMNMof:-NOH6jBi2dBPN25G8Wsa:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8045ae62405cd80008bbf9b3b10be3763d59c490/common/symbolic/polynomial.cc#L829)):*

`substitutions_obj` is created within this function `SubstituteAndExpand`, when we exit this function, `substitutions_obj` will be destroyed, and `substitutions` will be a dangling pointer.

I would suggest to just add a check `DRAKE_DEMAND(substitutions != nullptr)`.
</blockquote></details>

Done.
___
*[`common/symbolic/polynomial.cc` line 992 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-NOFv6F59MY-FDIOEOe6:-NOGPLqC-xhl133G-bXk:b-nto0dc) ([raw file](https://github.com/RobotLocomotion/drake/blob/8045ae62405cd80008bbf9b3b10be3763d59c490/common/symbolic/polynomial.cc#L992)):*

This is redundant? The whole file is in the symbolic namespace anyway.
</blockquote></details>

Not sure how that snuck in. Thanks.

ggould-tri (commented):
:lgtm: ; I am going to be out of town and some communications for a while, but this PR seems to be moving steadily along.  Feel free to dismiss me from any items blocking on me.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752#-:-NOHewhkE_ZkyE0bc_qm:b-8untvs)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @hongkai-dai)

hongkai-dai (commented):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

hongkai-dai (commented):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18752)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

PR #18289, opened 2022-11-09
Revert "[workspace] Upgrade usockets and uwebsockets"

jwnimmer-tri:
Reverts RobotLocomotion/drake#18259.  Re-opens #17910.

We have some evidence that a Docker image using this commit will not work with the Nginx proxy server we use for our Deepnote tutorials, yielding a "Bad Gateway" error message instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18289)

PR #18755, opened 2023-02-08
Downgrade usockets/uwebsockets

RussTedrake:
The upgrade in #18516 broke meshcat on Deepnote again.  See #18289.

+@jwnimmer-tri for both reviews, please.

Marking "emergency", because I'm trying to release pset 1 tonight, and it is MUCH more complicated if I can't use the recent Drake nightlies. (we've had a lot of deprecations / changes since the last functional nightly).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18755)


jwnimmer-tri:
:lgtm:

Issue #17249, opened 2022-05-23
[tutorials] Debugging a dynamical system


hongkai-dai:
Currently we have a tutorial https://github.com/RobotLocomotion/drake/blob/master/tutorials/dynamical_systems.ipynb on the concept of  dynamical systems. It would be good to add a tutorial to talk about practical guides on debugging dynamical systems (similar to in mathematical optimization we have tutorials like [this](https://github.com/RobotLocomotion/drake/blob/master/tutorials/mathematical_program.ipynb) on the concept, and a [debugging tutorial](https://github.com/RobotLocomotion/drake/blob/master/tutorials/debug_mathematical_program.ipynb) )

I would imagine the debugging tutorial can contain these tips

1. Generating the visual connection diagram in graphviz file, that shows the overall diagram with each input/output port and connections. This looks similar to the simulink block.
2. Using Simulator.monitor or VectorLogSink to inspect the simulation data.
3. Choosing the right simulator. For example in continuous-time simulation, if the simulation step is very small, then it indicates the system being stiff. The user should look into why the system has stiff dynamics, and possibly switch to implicit integrator and discrete-time system.

cc @sherm1 @jwnimmer-tri

sherm1:
Assigning to @jwnimmer-tri for disposition as component lead.

PR #18769, opened 2023-02-12
Adds  `Authoring Leaf Systems` tutorial

RussTedrake:
+@jwnimmer-tri for one review, please (feature?)
+@sherm1 for another review, please (platform?)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)


SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri, @RussTedrake, and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA-Yj4AeoUiFHWWzBy:-NOA-Yj4AeoUiFHWWzBz:bnboyhm) (no related file):*
BTW The tutorial is clear and easy to follow and covers the important topics, but I wonder about the treatment of caching. You've got two links (one in the text and one in the bottom), but I wonder if it isn't worth tackling it directly. You've got systems where the outputs are strictly dependent on their inputs, so showing an example where you exclude it from being dependent on time or parameters might be nice. Seeing the leaf system in context might help illustrate the significance -- i.e., an adder attached to a zero-order hold with a long period evaluated in a discrete system with a small discrete time step.

That said, I can also easily imagine that it's worth getting into that in its own tutorial.

___

jwnimmer-tri (commented):
First pass complete.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOAa9MKEcIeXMqufs5N:bvb1a6z)* status: 24 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri, @RussTedrake, and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZcOF9jEU8XHgYAXy:-NOAZcOG3Kr0OY30V3PG:b3g5e37) (no related file):*
nit Consider adding a `Simulator.set_monitor` callback to some of the simulated examples, to print out the current state.  It would help show the semantics of each system, and anyway is a good tip for debugging.

___
___
*[`tutorials/authoring_leaf_systems.ipynb` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOASeHAEOXEGIABOBPX:-NOASeHAEOXEGIABOBPY:btuqxnf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L30)):*
> ```
>    "metadata": {},
>    "source": [
>     "The [Modeling Dynamical Systems](./dynamical_systems.ipynb) tutorial gave a\n",
> ```

nit To me at least, this (important!) text gets slightly visually buried under the imports.  I'm not sure how to make it better (being _above_ the imports might not necessarily help? maybe add a section header for this text?), but for your consideration in case you find something.

Somehow we should be very clear to the audience here that we _assume_ familiarity with that material; this tutorial should not be the first one they step through.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATBr_-t1eH0ZLxxI_:-NOATBr_-t1eH0ZLxxIa:b-ouqzr6) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*
> ```
>     "## Input and Output Ports\n",
>     "\n",
>     "Leaf systems can have any number of input and output ports.  There are two basic types of ports: **vector-valued** ports and \"**abstract-valued**\" ports. Let's start with vector-valued ports, which are the easiest to work.  The following system declares two vector input ports, and two vector output ports."
> ```

nit In consistent quoting vs earlier in the sentence.



_Suggestion:_
```
**abstract-valued**
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATKDQ0pqH0acjZjZI:-NOATKDQ0pqH0acjZjZJ:b98hwfv) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*
> ```
>     "## Input and Output Ports\n",
>     "\n",
>     "Leaf systems can have any number of input and output ports.  There are two basic types of ports: **vector-valued** ports and \"**abstract-valued**\" ports. Let's start with vector-valued ports, which are the easiest to work.  The following system declares two vector input ports, and two vector output ports."
> ```

nit grammar



_Suggestion:_
```
easiest to work with
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATWc4DYrGp5iZOzSW:-NOATWc5DPtoUUOmDdOO:b9m3ls9) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L64)):*
> ```
>     "\n",
>     "    def CalcSum(self, context, output):\n",
>     "        # Evaluate the input ports to obtain the 2x1 vectors\n",
> ```

nit Throughout the tutorial, please check punc/grammar for all code comments.



_Suggestion:_
```
# Evaluate the input ports to obtain the 2x1 vectors.
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATjRD-mtzPaTAvXqA:-NOATjRD-mtzPaTAvXqB:bvgz4je) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L84)):*
> ```
>     "\n",
>     "# Fix the input ports to some constant values\n",
>     "system.get_input_port(0).FixValue(context, [3, 4])\n",
> ```

~As a best practice, I'm under the impression that (at least for calling code like this) we should refer to ports by their name instead of by index.  (The exception being ports where the index is more meaningful that a name, e.g., a multiplexer.) Probably the two Calc methods above should also use the names.~

Ah, the next section fixes that by showing how to use names.

Could you clarify for me why we show it the wrong way first?  I don't grok the benefit.  I would buy introducing the right way, and then in the text below it saying "You can also access ports by index but we discourage that", but a fully-worked example of the wrong way seems to do more harm than good.

If we do want to show code for the wrong way, I'd like to see it commented like the [styleguide](https://drake.mit.edu/styleguide/cppguide.html#Local_Variables) where the wrong code is marked with inline comments saying "# This wrong; use a name instead" or similar.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAUO0Q66HCFe34ya-b:-NOAUO0Q66HCFe34ya-c:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L100)):*
> ```
>     "- You can declare as many inputs or outputs as you like (and they need not be the same size).\n",
>     "- All of the inputs can be evaluated from any of the system methods that accept a `Context`.\n",
>     "- For each output port, you define a difference callback function that implements the output.\n",
> ```

typo



_Suggestion:_
```
different
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 117 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXL1n-7eCqwZiMDFD:-NOAXL1o0pNa56x7NeQl:b-bw8axp) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L117)):*
> ```
>     "class MyAdder(LeafSystem):\n",
>     "    def __init__(self):\n",
>     "        LeafSystem.__init__(self)  # Don't forget to initialize the base class.\n",
> ```

Working

I think this is a legacy Python spelling of init.  I'll check if we can replace it with something more typical.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAVprC72oYYFREj7ml:-NOAVprC72oYYFREj7mm:b-d7rc7h) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L118)):*
> ```
>     "    def __init__(self):\n",
>     "        LeafSystem.__init__(self)  # Don't forget to initialize the base class.\n",
>     "        self._a_index = self.DeclareVectorInputPort(name=\"a\",\n",
> ```

Moving forward, I don't think we should encourage people to use indices, even within a leaf system implementation.

A better option is to just keep a handle to the port:

```py
# ctor
self._a_port = self.DeclareVectorInputPort(name="a", size=2)
self._b_port = self.DeclareVectorInputPort(name="b", size=2)

# calc
a = self._a_port.Eval(context)
b = self._b_port.Eval(context)
```

Alternatively, calling `self.GetInputPort("a")` within a calc function is also fine.



_Code quote:_
```
self._a_index = 
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAWykW5s8_3y47Oj6N:-NOAWykW5s8_3y47Oj6O:b-l69qdv) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L129)):*
> ```
>     "    def CalcSum(self, context, output):\n",
>     "        # Evaluate the input ports to obtain the 2x1 vectors\n",
>     "        a = system.get_input_port(self._a_index).Eval(context)\n",
> ```

The fact that `system` happens to work here is a Jupyter anti-feature.  We should always use `self` -- throughout the PR.



_Suggestion:_
```
self
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXAs59AQEgP9lEyK7:-NOAXAs59AQEgP9lEyK8:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*
> ```
>    "metadata": {},
>    "source": [
>     "Not all inputs and outputs are best represented as a vector. Drake's system framework also supports passing structure types across ports, which is implemented in C++ using a technique called \"type erasure\". From the perspective of the systems framework, classes work with an \"abstract\" data type, and only the system implementation itself needs to know how to work with the structured type. In practice, if you can overlook our small amount of boilerplate, the usage is straightforward. Let's give an example of a `LeafSystem` that uses a `RigidTransform` in its input and output ports:"
> ```

typo



_Suggestion:_
```
structured
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 182 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXejDCPTe6aU9tYag:-NOAXejDCPTe6aU9tYah:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L182)):*
> ```
>     "\n",
>     "    def CalcOutput(self, context, output):\n",
>     "        # Evaluate the input ports to obtain the RigidTransform\n",
> ```

typo



_Suggestion:_
```
port
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 183 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXqfTAM9f2iO-5tKg:-NOAXqfTAM9f2iO-5tKh:biuaei4) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L183)):*
> ```
>     "    def CalcOutput(self, context, output):\n",
>     "        # Evaluate the input ports to obtain the RigidTransform\n",
>     "        X_1 = system.get_input_port(0).Eval(context)\n",
> ```

nit Avoid indices.  Ditto below in the same cell (for both the input and output port).



_Suggestion:_
```
get_input_port()
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 208 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAY1qs7R93SULliPZl:-NOAY1qs7R93SULliPZm:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L208)):*
> ```
>     "## State Variables\n",
>     "\n",
>     "Systems store their state in the `Context`, and work with state in a very similar way to input and output ports. Again, we have have vector-valued state or abstract-valued state. Abstract state can only be updated in a discrete-time fashion, but vector-valued state can be declared to be either **discrete** or **continuous**.  \n",
> ```

typo



_Suggestion:_
```
we can have
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 234 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAYJ3X0-7yvTxEgJMG:-NOAYJ3X0-7yvTxEgJMH:b-wcr3zs) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L234)):*
> ```
>     "    def Update(self, context, discrete_state):\n",
>     "        x = context.get_discrete_state_vector().GetAtIndex(0)\n",
>     "        xnext = x**3\n",
> ```

nit Consider adding an underscore for readability.



_Suggestion:_
```
x_next
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 286 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAYvSF7qLvuoYyYxRy:-NOAYvSF7qLvuoYyYxRz:bipu7l4) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L286)):*
> ```
>    "metadata": {},
>    "source": [
>     "For abstract-valued state, we have a similar workflow. We declare the state in the constructor, and define update events to update that state.  Discrete update events are restricted to updating *only* the discrete states, so we use \"unrestricted\" events to update the abstract state. Here is an example which stores a simple `PiecewisePolynomial` trajectory as abstract state:"
> ```

nit clarity



_Suggestion:_
```
the abstract state
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 305 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZ7bR4svjCw0SuKsf:-NOAZ7bR4svjCw0SuKsg:buh905k) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L305)):*
> ```
>     "                                                    update=self.Update)\n",
>     "\n",
>     "    # x[n+1] = x^3[n]\n",
> ```

nit This comment is stale (incorrect).
___
*[`tutorials/authoring_leaf_systems.ipynb` line 308 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZIXv81B0_gSbgjjM:-NOAZIXv81B0_gSbgjjN:b-j0bhj6) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L308)):*
> ```
>     "    def Update(self, context, discrete_state):\n",
>     "        t = context.get_time()\n",
>     "        # Update the state\n",
> ```

nit It would probably be more clear to split the update into two statements -- one to compute the new value, and one to set it back into the context -- like we did for  the prior two examples.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 403 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_xzz4p_DEcruHlzK:-NOA_xzz4p_DEcruHlzL:bx9h6h9) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L403)):*
> ```
>     "\n",
>     "        def _construct_copy(self, other, converter=None):\n",
>     "            Impl._construct(self, converter=converter)\n",
> ```

Should we mention here that any member fields (e.g., a config value) need to be transferred form `other` to `self`?
___
*[`tutorials/authoring_leaf_systems.ipynb` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_OAv8dKS4L0voQmG:-NOA_OAv8dKS4L0voQmH:b14j2d9) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L406)):*
> ```
>     "\n",
>     "        def CostOutput(self, context, output):\n",
>     "            x = self.get_input_port(0).Eval(context)\n",
> ```

nit A reminder to remove port-by-magic-number throughout the tutorial.



_Code quote:_
```
get_input_port(0)
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 447 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_lij5RwLd6vgEOtI:-NOA_lij5RwLd6vgEOtJ:bxsntby) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L447)):*
> ```
>     "# either by declaring them directly:\n",
>     "system_ad = RunningCost_[AutoDiffXd]\n",
>     "system_symbolic = RunningCost_[Expression]\n",
> ```

nit Call the constructors



_Suggestion:_
```
    "# either by declaring them directly:\n",
    "system_ad = RunningCost_[AutoDiffXd]()\n",
    "system_symbolic = RunningCost_[Expression]()\n",
```
___
*[`tutorials/dynamical_systems.ipynb` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAQeaAAV1Qo5GpT65l:-NOAQeaB0pI4NpokfY83:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L34)):*
> ```
>     "systems](https://drake.mit.edu/doxygen_cxx/group__systems.html) for you to\n",
>     "build on.  These include primitive systems like integrators and multiplexers,\n",
>     "as well as more advanced systems for that implement the physics engine, or more\n",
> ```

typo



_Suggestion:_
```
systems that implement
```
___
*[`tutorials/dynamical_systems.ipynb` line 37 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAQmaG9pc6kYMEnj8N:-NOAQmaG9pc6kYMEnj8O:b2ni8l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L37)):*
> ```
>     "advanced algorithms for estimation and control.  As we will see below, these\n",
>     "building blocks can be combined into more complicated systems.  However, Drake\n",
>     "also makes it easy to write your own systems (in C++ or directly in python).  \n",
> ```

nit



_Suggestion:_
```
Python
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)* status: 26 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @RussTedrake and @sherm1)

___
*[`tutorials/authoring_leaf_systems.ipynb` line 117 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXL1n-7eCqwZiMDFD:-NOAeU_V19CC9gxaGzLo:b-ugicdk) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L117)):*

Working

I think this is a legacy Python spelling of init.  I'll check if we can replace it with something more typical.
</blockquote></details>

Yes.

Throughout the PR, we should use the preferred (i.e., less brittle) spelling:

```py
super().__init__()  # Don't forget to initialize the base class.
```

The `LeafSystem` becomes `super()`, and we drop the `self`.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 375 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAcCp58BAmDE4jxhgd:-NOAcCp58BAmDE4jxhge:b-o1s7zf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L375)):*
> ```
>    "metadata": {},
>    "source": [
>     "## Supporting Scalar Type Conversion (double, AutoDiff, and Symbolic)\n",
> ```

nit Somewhere in this section, we should probably link to here:

https://drake.mit.edu/pydrake/pydrake.systems.scalar_conversion.html#pydrake.systems.scalar_conversion.TemplateSystem
___
*[`tutorials/dynamical_systems.ipynb` line 190 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAcbiV8iSPSs1pQUU9:-NOAcbiV8iSPSs1pQUUA:bbl27xg) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L190)):*
> ```
>     "We also supply a [number of other helper classes and methods](https://drake.mit.edu/doxygen_cxx/group__systems.html) that derive from or construct a `LeafSystem`, such as the [`pydrake.systems.primitives.LinearSystem`](https://drake.mit.edu/pydrake/pydrake.systems.primitives.html?highlight=linearsystem#pydrake.systems.primitives.LinearSystem) class or the [`pydrake.systems.primitives.Linearize()`](https://drake.mit.edu/pydrake/pydrake.systems.primitives.html?highlight=linearize#pydrake.systems.primitives.Linearize) method.  And in many cases, like simulating the dynamics and actuators/sensors of robots, most of the classes that you need have already been implemented.\n",
>     "\n",
>     "For more examples of using the more advanced features of `LeafSystem`, the current best place to look is the [`custom_test.py`](https://github.com/RobotLocomotion/drake/blob/master/bindings/pydrake/systems/test/custom_test.py) file that provides test coverage for the python bindings. \n",
> ```

nit Here we should probably link to the new tutorial.

sherm1 (commented):
This looks very useful! Platform :lgtm: pending comments from reviewers to be addressed.

BTW I'm not sure whether it will be obvious to readers what the various features demonstrated here are used for. Consider starting each section with a possible problem to be solved, like "Suppose you need to know gradients of the system dynamics ..." or "You are doing a design study and have parameters you'd like to be able to vary ..." or (hopefully) some better examples.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOBwy1lDlsm0rM3zZor:bdkkhti)* status: 33 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @RussTedrake)

___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBi42UBiTl6z_aR00I:-NOBi42UBiTl6z_aR00J:b-esjprf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*
> ```
>     "## Input and Output Ports\n",
>     "\n",
>     "Leaf systems can have any number of input and output ports.  There are two basic types of ports: **vector-valued** ports and \"**abstract-valued**\" ports. Let's start with vector-valued ports, which are the easiest to work.  The following system declares two vector input ports, and two vector output ports."
> ```

BTW two thoughts:
- add "It forms the sum and difference of the two 2-element input vectors and places the results on the two output ports."
- It would be very nice to have one of those nice system diagrams here to provide a visual hook on which to hang the code.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATjRD-mtzPaTAvXqA:-NOBnZpYAE7FtT9IwfOE:b-kao3gw) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L84)):*

~As a best practice, I'm under the impression that (at least for calling code like this) we should refer to ports by their name instead of by index.  (The exception being ports where the index is more meaningful that a name, e.g., a multiplexer.) Probably the two Calc methods above should also use the names.~

Ah, the next section fixes that by showing how to use names.

Could you clarify for me why we show it the wrong way first?  I don't grok the benefit.  I would buy introducing the right way, and then in the text below it saying "You can also access ports by index but we discourage that", but a fully-worked example of the wrong way seems to do more harm than good.

If we do want to show code for the wrong way, I'd like to see it commented like the [styleguide](https://drake.mit.edu/styleguide/cppguide.html#Local_Variables) where the wrong code is marked with inline comments saying "# This wrong; use a name instead" or similar.
</blockquote></details>

BTW my preference would be only to show code that accesses the ports by name, with just a text comment below noting that ports can also be accessed by index or ordinal. Otherwise I think the dispreferred code will propagate via cut&paste.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBrYAA0ON0wDrifvNy:-NOBrYAA0ON0wDrifvNz:b2nhbfk) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*
> ```
>    "metadata": {},
>    "source": [
>     "Not all inputs and outputs are best represented as a vector. Drake's system framework also supports passing structure types across ports, which is implemented in C++ using a technique called \"type erasure\". From the perspective of the systems framework, classes work with an \"abstract\" data type, and only the system implementation itself needs to know how to work with the structured type. In practice, if you can overlook our small amount of boilerplate, the usage is straightforward. Let's give an example of a `LeafSystem` that uses a `RigidTransform` in its input and output ports:"
> ```

BTW



_Suggestion:_
```
a small amount
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 210 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBsTUC09oYAPZ9AV9-:-NOBsTUC09oYAPZ9AV90:b-jxhxd0) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L210)):*
> ```
>     "Systems store their state in the `Context`, and work with state in a very similar way to input and output ports. Again, we have have vector-valued state or abstract-valued state. Abstract state can only be updated in a discrete-time fashion, but vector-valued state can be declared to be either **discrete** or **continuous**.  \n",
>     "\n",
>     "A discrete state $x_d$ will evolved based on a update events, the most common of which would be a simple periodic event to define a difference equation.  We saw an example of this in the [introductory notebook](./dynamical_systems.ipynb)."
> ```

typo s



_Suggestion:_
```
evolve based on update events
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 252 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBtFp7-4dsgK6m8yIg:-NOBtFp7-4dsgK6m8yIh:b-t9yneb) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L252)):*
> ```
>     "Note that the states of a system are not immediately accessible to other systems in a `Diagram`. For a system to share its state, it must do that using an output port. We provide the `DeclareStateOutputPort()` method to make this common case easy.\n",
>     "\n",
>     "The evolution of continuous vector-valued state variables is governed by a differential equation. While it is possible and convenient to declare many different groups of discrete state variables (they may even be updated at different rates by different events), we only define zero or one continuous state vector for a system, and define its dynamics by overloading the `LeafSystem::DoCalcTimeDerivatives()` method. Here is the simple continuous-time example in the introductory notebook again: "
> ```

BTW consider a link to the intro notebook here



_Code quote:_
```
introductory notebook
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 400 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBuax_Cd1Jra_yQQGN:-NOBuax_Cd1Jra_yQQGO:b-tlxehp) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L400)):*
> ```
>     "            self.DeclareVectorInputPort(\"state\", 2)\n",
>     "            self.DeclareVectorInputPort(\"command\", 1)\n",
>     "            self.DeclareVectorOutputPort(\"cost\", 1,self.CostOutput)\n",
> ```

nit: missing space after comma
___
*[`tutorials/authoring_leaf_systems.ipynb` line 458 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBvmQuE7X6dzmuLQPt:-NOBvmQuE7X6dzmuLQPu:b-1picrd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L458)):*
> ```
>     "context_symbolic.SetTimeStateAndParametersFrom(context)\n",
>     "system_symbolic.FixInputPortsFrom(system, context, context_symbolic)\n",
>     "print(system_symbolic.get_output_port().Eval(context_symbolic))"
> ```

BTW consider also showing what you can do with the AutoDiff version of the system
___
*[`tutorials/dynamical_systems.ipynb` line 31 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBwAQKEXoCbgkFKYyG:-NOBwAQKEXoCbgkFKYyH:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L31)):*
> ```
>     "## A library of dynamical systems\n",
>     "\n",
>     "Drake provide a [rich library of dynamical\n",
> ```

typo



_Suggestion:_
```
provides
```

zachfang (commented):
Some small editorial suggestions, but it looks good to me otherwise. Very informative!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOChbLo4V1AAINUE3Ws:b-mjrxxl)* status: 40 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @RussTedrake)

___
*[`tutorials/authoring_leaf_systems.ipynb` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCguV_DWSKrIk33HrA:-NOCguV_DWSKrIk33HrB:bpderbj) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L99)):*
> ```
>     "There are a few things to notice here:\n",
>     "- You can declare as many inputs or outputs as you like (and they need not be the same size).\n",
>     "- All of the inputs can be evaluated from any of the system methods that accept a `Context`.\n",
> ```

BTW. This isn't clear to me as a systems framework newbie. I assume it means you can do `system.get_input_port(1)` to get/set values?



_Code quote:_
```
All of the inputs can be evaluated from any of the system methods that accept a `Context`.
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCYdOECX6Th6beyMx5:-NOCYdOECX6Th6beyMx6:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L106)):*
> ```
>     "- You can also make inputs optional, by checking [InputPort::HasValue()](https://drake.mit.edu/doxygen_cxx/classdrake_1_1systems_1_1_input_port.html#a5536b94a4642fa4cf47164437dc66ae8) in the system method implementations.\n",
>     "\n",
>     "Keeping track of the order of the ports can be a bit brittle. So we can improve the code e.g. by using their index or referencing them by name.  Here is a modified version of that example which uses both approaches:"
> ```

nit.



_Suggestion:_
```
, e.g.,
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOC_RjJDCBi49lwjAYk:-NOC_RjJDCBi49lwjAYl:bldrbym) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*
> ```
>    "metadata": {},
>    "source": [
>     "Not all inputs and outputs are best represented as a vector. Drake's system framework also supports passing structure types across ports, which is implemented in C++ using a technique called \"type erasure\". From the perspective of the systems framework, classes work with an \"abstract\" data type, and only the system implementation itself needs to know how to work with the structured type. In practice, if you can overlook our small amount of boilerplate, the usage is straightforward. Let's give an example of a `LeafSystem` that uses a `RigidTransform` in its input and output ports:"
> ```

nit: we use `systems framework` for the entire tutorial, and I think this is a typo? Also, there is one place at the very beginning where the `s` is capitalized, i.e., Systems framework, if we want to fix that as well.



_Suggestion:_
```
systems framework
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCaGyn3lO7m2Ff5MDJ:-NOCaGyn3lO7m2Ff5MDK:bdsh2j7) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*
> ```
>    "metadata": {},
>    "source": [
>     "Not all inputs and outputs are best represented as a vector. Drake's system framework also supports passing structure types across ports, which is implemented in C++ using a technique called \"type erasure\". From the perspective of the systems framework, classes work with an \"abstract\" data type, and only the system implementation itself needs to know how to work with the structured type. In practice, if you can overlook our small amount of boilerplate, the usage is straightforward. Let's give an example of a `LeafSystem` that uses a `RigidTransform` in its input and output ports:"
> ```

BTW. I wonder if using a `###Abstract-valued Ports` title would help readability here to emphasize `vector-valued` vs. `abstract-valued` ports, i.e., the reader would know immediately we are switching to a different type of port.



_Code quote:_
```
Not all inputs
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 327 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCeQNIBghNp8gb2fQD:-NOCeQNIBghNp8gb2fQE:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L327)):*
> ```
>     "## Parameters\n",
>     "\n",
>     "Parameters are like states, except that they are constant through the lifetime of a simulation. Parameters in the systems framework are declared and accessed almost identically to state, but they are never updated. Once again, we have vector-valued (declared with `DeclareNumericParameter`) and abstract-valued (via `DeclareAbstractParameter`) parameters."
> ```

nit.



_Suggestion:_
```
states
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 422 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCfQwj8209k9J2k9uK:-NOCfQwj8209k9J2k9uL:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L422)):*
> ```
>    "source": [
>     "The important steps are:\n",
>     "- Add the @TemplateSystem decorator\n",
> ```

nit.



_Suggestion:_
```
Add the `@TemplateSystem` decorator.
```
___
*[`tutorials/authoring_leaf_systems.ipynb` line 426 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCfoDRCqMJeBjtiA-f:-NOCfoDRCqMJeBjtiA-g:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L426)):*
> ```
>     "- Implement the `_construct` method *instead* of the typical `__init__` method.\n",
>     "- Implement the `_construct_copy` method.\n",
>     "- Add the default instantiation, so that you can continue to refer to the system as e.g. `RunningCost` in addition to using `RunningCost_[float]`.\n",
> ```

nit.



_Suggestion:_
```
, e.g.,
```

RussTedrake (commented):
Done.  Thanks for all of the comments.  Apologies for the incredible number of typos/grammatical errors!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOEqMI_DEN8qj__R_N0:b-xhhcr2)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri, @sherm1, and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA-Yj4AeoUiFHWWzBy:-NOEpBhT4Pi7Lf4X22wS:b-fmf9fp) (no related file):*

BTW The tutorial is clear and easy to follow and covers the important topics, but I wonder about the treatment of caching. You've got two links (one in the text and one in the bottom), but I wonder if it isn't worth tackling it directly. You've got systems where the outputs are strictly dependent on their inputs, so showing an example where you exclude it from being dependent on time or parameters might be nice. Seeing the leaf system in context might help illustrate the significance -- i.e., an adder attached to a zero-order hold with a long period evaluated in a discrete system with a small discrete time step.

That said, I can also easily imagine that it's worth getting into that in its own tutorial.
</blockquote></details>

Done. I added a bullet to introduce the potential importance of using `prerequisites_of_calc`, and feel like I don't want to add much more here.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZcOF9jEU8XHgYAXy:-NOEnGcJ8oW58brQw4y7:b-ygr69m) (no related file):*

nit Consider adding a `Simulator.set_monitor` callback to some of the simulated examples, to print out the current state.  It would help show the semantics of each system, and anyway is a good tip for debugging.
</blockquote></details>

I think we should definitely have a "debugging in the systems framework" tutorial in addition to this one (#17249); I feel that's a better place for it?

___
___
*[`tutorials/authoring_leaf_systems.ipynb` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOASeHAEOXEGIABOBPX:-NOCxXN0AIaxQxnKBKE5:b5a6tfz) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L30)):*

nit To me at least, this (important!) text gets slightly visually buried under the imports.  I'm not sure how to make it better (being _above_ the imports might not necessarily help? maybe add a section header for this text?), but for your consideration in case you find something.

Somehow we should be very clear to the audience here that we _assume_ familiarity with that material; this tutorial should not be the first one they step through.
</blockquote></details>

done. added a subsection heading.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATBr_-t1eH0ZLxxI_:-NOD3_ge6zLcg9Sbx4YO:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*

nit In consistent quoting vs earlier in the sentence.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATKDQ0pqH0acjZjZI:-NOD3_7H8wcVC2Knkm5H:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*

nit grammar
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBi42UBiTl6z_aR00I:-NOD3ZQJ4mrMAf5aonUg:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L44)):*

BTW two thoughts:
- add "It forms the sum and difference of the two 2-element input vectors and places the results on the two output ports."
- It would be very nice to have one of those nice system diagrams here to provide a visual hook on which to hang the code.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATWc4DYrGp5iZOzSW:-NOEVyRy15Z7vj8N3lJw:b-o6wqvu) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L64)):*

nit Throughout the tutorial, please check punc/grammar for all code comments.
</blockquote></details>

done.  (mostly added a lot of periods throughout)
___
*[`tutorials/authoring_leaf_systems.ipynb` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOATjRD-mtzPaTAvXqA:-NOEWWU4BzR7OKKzHo58:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L84)):*

BTW my preference would be only to show code that accesses the ports by name, with just a text comment below noting that ports can also be accessed by index or ordinal. Otherwise I think the dispreferred code will propagate via cut&paste.
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCguV_DWSKrIk33HrA:-NOEWo0T-PpSiEYNZF65:b66ocx4) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L99)):*

BTW. This isn't clear to me as a systems framework newbie. I assume it means you can do `system.get_input_port(1)` to get/set values?
</blockquote></details>

Done.  Tried to make it more precise/specific.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAUO0Q66HCFe34ya-b:-NOEWt5y1G1IULy0Pc6I:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L100)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCYdOECX6Th6beyMx5:-NOEXDyM8145rYbfKAbJ:bb7f2mf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L106)):*

nit.
</blockquote></details>

done
___
*[`tutorials/authoring_leaf_systems.ipynb` line 117 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXL1n-7eCqwZiMDFD:-NOEXV4NBKBkIEzD19su:bso2kcb) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L117)):*

Yes.

Throughout the PR, we should use the preferred (i.e., less brittle) spelling:

```py
super().__init__()  # Don't forget to initialize the base class.
```

The `LeafSystem` becomes `super()`, and we drop the `self`.
</blockquote></details>

done.  i hadn't actually realized that!
___
*[`tutorials/authoring_leaf_systems.ipynb` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAVprC72oYYFREj7ml:-NOEY51P7w_miZ2g0jhp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L118)):*

Moving forward, I don't think we should encourage people to use indices, even within a leaf system implementation.

A better option is to just keep a handle to the port:

```py
# ctor
self._a_port = self.DeclareVectorInputPort(name="a", size=2)
self._b_port = self.DeclareVectorInputPort(name="b", size=2)

# calc
a = self._a_port.Eval(context)
b = self._b_port.Eval(context)
```

Alternatively, calling `self.GetInputPort("a")` within a calc function is also fine.
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAWykW5s8_3y47Oj6N:-NOEYAQ12VgeAGqoJpja:b5z05cy) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L129)):*

The fact that `system` happens to work here is a Jupyter anti-feature.  We should always use `self` -- throughout the PR.
</blockquote></details>

Done (that was just a typo, but it's moot now anyhow)
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXAs59AQEgP9lEyK7:-NOEYMEfCFW3jGnWJonM:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBrYAA0ON0wDrifvNy:-NOEYSiH0Xg7FIYmjv9O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*

BTW
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOC_RjJDCBi49lwjAYk:-NOEYky370k6rixHQgYZ:bcj4xes) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*

nit: we use `systems framework` for the entire tutorial, and I think this is a typo? Also, there is one place at the very beginning where the `s` is capitalized, i.e., Systems framework, if we want to fix that as well.
</blockquote></details>

I've always used "systems framework" https://medium.com/toyotaresearch/drake-model-based-design-in-the-age-of-robotics-and-machine-learning-59938c985515 
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCaGyn3lO7m2Ff5MDJ:-NOEZ1IL04FlNVRFIZF3:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*

BTW. I wonder if using a `###Abstract-valued Ports` title would help readability here to emphasize `vector-valued` vs. `abstract-valued` ports, i.e., the reader would know immediately we are switching to a different type of port.
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 182 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXejDCPTe6aU9tYag:-NOEZ5Cr3DqWoTvmvrMT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L182)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 183 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAXqfTAM9f2iO-5tKg:-NOEZ5wgBRLe3sGJSJxl:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L183)):*

nit Avoid indices.  Ditto below in the same cell (for both the input and output port).
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 208 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAY1qs7R93SULliPZl:-NOEZClh4xMT9YPvqUb1:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L208)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 210 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBsTUC09oYAPZ9AV9-:-NOEZHjz3tt5Df2VncEG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L210)):*

typo s
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 234 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAYJ3X0-7yvTxEgJMG:-NOEZS4e8rHQ6oeXWlvi:bj3cq1x) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L234)):*

nit Consider adding an underscore for readability.
</blockquote></details>

Done.  here and in the original.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 252 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBtFp7-4dsgK6m8yIg:-NOEZcoW1ZuNyXWUuukB:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L252)):*

BTW consider a link to the intro notebook here
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 286 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAYvSF7qLvuoYyYxRy:-NOEZggVE1iRX4fuogCZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L286)):*

nit clarity
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 305 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZ7bR4svjCw0SuKsf:-NOEZj6c0Um8jTS5wGrm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L305)):*

nit This comment is stale (incorrect).
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 308 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZIXv81B0_gSbgjjM:-NOE_8OU5w4gM3Hd53rU:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L308)):*

nit It would probably be more clear to split the update into two statements -- one to compute the new value, and one to set it back into the context -- like we did for  the prior two examples.
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 327 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCeQNIBghNp8gb2fQD:-NOE_DlM9qp8bkntroUj:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L327)):*

nit.
</blockquote></details>

Done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 375 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAcCp58BAmDE4jxhgd:-NOEhYmx4WxNZ9UgHyJW:bb7f2mf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L375)):*

nit Somewhere in this section, we should probably link to here:

https://drake.mit.edu/pydrake/pydrake.systems.scalar_conversion.html#pydrake.systems.scalar_conversion.TemplateSystem


</blockquote></details>

done
___
*[`tutorials/authoring_leaf_systems.ipynb` line 400 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBuax_Cd1Jra_yQQGN:-NOEhatcCFhzKPS2hO59:bb7f2mf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L400)):*

nit: missing space after comma
</blockquote></details>

done
___
*[`tutorials/authoring_leaf_systems.ipynb` line 403 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_xzz4p_DEcruHlzK:-NOEj1yK5I0yN5nK3ISu:b-nu9rf) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L403)):*

Should we mention here that any member fields (e.g., a config value) need to be transferred form `other` to `self`?
</blockquote></details>

done.  made Q a config value to make the point.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_OAv8dKS4L0voQmG:-NOEjLbg-ntQvty3B1nh:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L406)):*

nit A reminder to remove port-by-magic-number throughout the tutorial.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 422 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCfQwj8209k9J2k9uK:-NOEjfTS8Goqsh2fDSey:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L422)):*

nit.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 426 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOCfoDRCqMJeBjtiA-f:-NOEjiscEssh3yb12Xl2:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L426)):*

nit.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 447 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOA_lij5RwLd6vgEOtI:-NOEjo1vEpTdjHmNe-Sj:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L447)):*

nit Call the constructors
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 458 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBvmQuE7X6dzmuLQPt:-NOEkVUY0Cn1fhuyBapf:bq6g6hc) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L458)):*

BTW consider also showing what you can do with the AutoDiff version of the system
</blockquote></details>

I intentionally dodged writing an autodiff tutorial here, though it would certainly be valuable to have.
___
*[`tutorials/dynamical_systems.ipynb` line 31 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOBwAQKEXoCbgkFKYyG:-NOEpl_U09QH_pb-qTab:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L31)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/dynamical_systems.ipynb` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAQeaAAV1Qo5GpT65l:-NOEps1N-Eu-JtGf6GF-:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L34)):*

typo
</blockquote></details>

Done.
___
*[`tutorials/dynamical_systems.ipynb` line 37 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAQmaG9pc6kYMEnj8N:-NOEq1Xf-1_dp8gGv_Ts:b-ynyoa5) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L37)):*

nit
</blockquote></details>

Done.  here and throughout.
___
*[`tutorials/dynamical_systems.ipynb` line 190 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAcbiV8iSPSs1pQUU9:-NOEqKP-4GT4k5Iq8tWi:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/dynamical_systems.ipynb#L190)):*

nit Here we should probably link to the new tutorial.
</blockquote></details>

Done.

jwnimmer-tri (commented):
I'll circle back and re-run this locally soon.

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOF_J87-VLh_RlawoBM:b9c84mh)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @RussTedrake and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZcOF9jEU8XHgYAXy:-NOFZ5Si7KSuoQOTFEAL:bpg62sw) (no related file):*

I think we should definitely have a "debugging in the systems framework" tutorial in addition to this one (#17249); I feel that's a better place for it?
</blockquote></details>

For sure a real treatment deserves its own tutorial.

I mentioned it here only because when I ran a lot of examples, the only output from the cell was `<SimulatorStatus at 0xdeadbeef>`, which didn't really communicate anything about the dynamics that we just implemented.

Another option would be to put some `print()` statements in the `CalcUpdate()` handlers.

(I am also OK without any logging/prints.)

___

jwnimmer-tri (commented):
:lgtm:

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-:-NOFfhgV583ljHX1pqs6:bnfp4nl)* status: 3 unresolved discussions (waiting on @RussTedrake and @sherm1)

___
*[`tutorials/authoring_leaf_systems.ipynb` line 394 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOFfnjeA_TeIoKnmf1y:-NOFfnjf0EHrEu6aR0z1:b-tqaczs) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b0a53b543cc23cad7318defee0afba59cf1a81a/tutorials/authoring_leaf_systems.ipynb#L394)):*
> ```
>     "- Derive from `LeafSystem_[T]` instead of simply `LeafSystem`.\n",
>     "- Implement the `_construct` method *instead* of the typical `__init__` method.\n",
>     "- Implement the `_construct_copy` method.\n",
> ```

nit Consider highlighting here that _construct_copy needs to populate the same member fields as _construct (ala self.Q).

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)* status: 2 unresolved discussions (waiting on @RussTedrake)

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZcOF9jEU8XHgYAXy:-NOGCIX30Gu-LbbLeq7o:b-tatjc1) (no related file):*

For sure a real treatment deserves its own tutorial.

I mentioned it here only because when I ran a lot of examples, the only output from the cell was `<SimulatorStatus at 0xdeadbeef>`, which didn't really communicate anything about the dynamics that we just implemented.

Another option would be to put some `print()` statements in the `CalcUpdate()` handlers.

(I am also OK without any logging/prints.)
</blockquote></details>

I've added some print statements, which ended up including some demonstrations of how to set / access the different state components, and caught a bug.  :grimacing: 

___
___
*[`tutorials/authoring_leaf_systems.ipynb` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOC_RjJDCBi49lwjAYk:-NOGCYhPBoolUq0LHBH8:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db695e7f0fa9426ec6ad5724ea5d96e4566bc3a7/tutorials/authoring_leaf_systems.ipynb#L161)):*

Got it. Then we have `Drake's system framework` here that needs fixing.
</blockquote></details>

done.
___
*[`tutorials/authoring_leaf_systems.ipynb` line 394 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOFfnjeA_TeIoKnmf1y:-NOGCm0y3MiXwPIZqatP:b-7rl0jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b0a53b543cc23cad7318defee0afba59cf1a81a/tutorials/authoring_leaf_systems.ipynb#L394)):*

nit Consider highlighting here that _construct_copy needs to populate the same member fields as _construct (ala self.Q).
</blockquote></details>

done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18769)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18769#-NOAZcOF9jEU8XHgYAXy:-NOGHrFwBXIKM9OaR0rj:bu841ps) (no related file):*

I've added some print statements, which ended up including some demonstrations of how to set / access the different state components, and caught a bug.  :grimacing: 
</blockquote></details>

Nice, I think those better than a monitor.

___

PR #18770, opened 2023-02-12
Rename RobotDiagram::BuildDiagram() => Build()

RussTedrake:
This is less verbose and more consistent with `DiagramBuilder()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18770)

PR #18772, opened 2023-02-13
[planning] Simplify RobotDiagramBuilder function names

jwnimmer-tri:
The typical use case of a builder is that the object is mutable and users are mutating it. In that case, the clarity benefit of using "mutable_" prefix on is small compared to the cost. (In contrast with, e.g., a Context where mutation access causes cache entry invalidation so using special names make sense.)  This brings C++ up to par with what we did in Python.

Also deprecate the IsPlantFinalized and FinalizePlant sugar. The sugar isn't encapsulating or abbreviating anything versus calling the MbP functions directly on the plant() accessor, where the MbP functions have full documentation directly under your auto-complete IDE fingertips.

---

Follow-up from #18770.

\CC @RussTedrake @calderpg-tri FYI

Confirmed `Anzu-master-Drake-PR/181` passes vs these deprecations.

Confirmed `anzu/pull/10107` passes with this PR + Anzu fixes for the new spellings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18772)

### DIFF
--- a/bindings/pydrake/planning/planning_py_robot_diagram.cc
+++ b/bindings/pydrake/planning/planning_py_robot_diagram.cc
@@ -55,37 +55,30 @@ void DefinePlanningRobotDiagram(py::module m) {
     {
       using Class = RobotDiagramBuilder<T>;
       constexpr auto& cls_doc = doc.RobotDiagramBuilder;
-      auto cls = DefineTemplateClassWithDefault<Class>(m, "RobotDiagramBuilder",
-          GetPyParam<T>(),
-          (std::string(cls_doc.doc) +
-              "\n\n"
-              "Note:\n"
-              "    The C++ class defines pairs of sibling methods, e.g.,\n"
-              "    parser() vs mutable_parser(). Since Python does not use\n"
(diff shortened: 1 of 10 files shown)


RussTedrake (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18772#-:-NO9FM9V2kDrm9W7IyIU:b-wnpj7i)* status: needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

jwnimmer-tri:
+@ggould-tri for platform review per schedule, please.

ggould-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18772#-:-NOBGTG66hIOGEvSZxHF:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),RussTedrake(platform) (waiting on @jwnimmer-tri)

PR #18774, opened 2023-02-13
[visualization] Subscribe to point cloud LCM messages in Meldis

zachfang:


zachfang (commented):
@jwnimmer-tri Can I ask for an initial review to ensure the changes are on track?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NOAc4Fs6UbI6_cJYNPn:bpp77kg)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOAbR6L9lQVo5UaZi9L:b-spit2h) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*
> ```Python
>                         message_type=lcmt_viewer_link_data,
>                         handler=default_viewer.on_viewer_draw_deformable)
>         self._subscribe(channel="DRAKE_POINT_CLOUD",
> ```

FYI. I wonder if the LCM channel name should be `DRAKE_POINT_CLOUD` or a regex `DRAKE_POINT_CLOUD_*`.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbkxh1ZlWVbE8HN2I:-NOAbkxh1ZlWVbE8HN2J:bqsdy1c) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/test/meldis_test.py#L356)):*
> ```Python
>         context.SetTime(1.0)
>         cloud_to_lcm.get_input_port().FixValue(context, xyz_cloud)
>         lcm_point_cloud = cloud_to_lcm.get_output_port().Eval(context)
> ```

Working.

Currently, I get an error message when running the test.
```
RuntimeError: This derived class of `AbstractValue`, `drake::Value<drake::lcmt_point_cloud>`, is not exposed to pybind11, so `get_value` cannot be called. See `AddValueInstantiation` for how to bind it.
```

I assume I need to bind that LCM message type as suggested.

zachfang:
This is the initial visualization result of the point cloud published through `mock_station_simulation`. (Need to check why it's all black.)

Screenshot from 2023-02-13 11-14-06

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLerUCE1RxL59uYKpt):*
nit For users, I think the important take-away message would go more like this:



_Suggestion:_
```
[visualization] Meldis displays point cloud data
```
___
*[`bindings/pydrake/visualization/_meldis.py` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLf3Ej1yjzD3zJzSqb:-NOLf3Ek5CbbfpLTwkA5:b-6dxyuz) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L45)):*
> ```Python
>     ContactVisualizerParams,
> )
> from pydrake.perception import BaseField, Fields, PointCloud
> ```

nit The import convention in this file is one-class-per-line (see above).
___
*[`bindings/pydrake/visualization/_meldis.py` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLfkGz8lqnmrpkUDTY:-NOLfkGz8lqnmrpkUDTZ:b47kcn7) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L81)):*
> ```Python
> 
> class _ViewerApplet:
>     """Displays lcmt_viewer_load_robot and lcmt_viewer_draw into MeshCat."""
> ```

See comment here.  This applet was designed only for "viewer" messages.

The point cloud message infra is different, so should be its own separate `class _PointCloudApplet`.
___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOLfCWK1EKJ8-ANqB27:b-mxmt8p) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

FYI. I wonder if the LCM channel name should be `DRAKE_POINT_CLOUD` or a regex `DRAKE_POINT_CLOUD_*`.
</blockquote></details>

What did `drake_visualizer` do?  We should aim to do that (at minimum).
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 341 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLfzFe28xGlt2j2UX7:-NOLfzFf5E6EwVv2ijHb:bqcnwmg) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/test/meldis_test.py#L341)):*
> ```Python
>         num_points = 10
>         xyzs = np.random.uniform(-0.1, 0.1, (3, num_points))
>         rgbs = np.random.uniform(0.0, 255.0, (3, num_points))
> ```

nit This should probably be dtype=uint8?  Possibly numpy converts it for us, but we might as well not borrow trouble.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbkxh1ZlWVbE8HN2I:-NOLgPUt2AizuVImnN_n:b-iab1xq) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/test/meldis_test.py#L356)):*

Working.

Currently, I get an error message when running the test.
```
RuntimeError: This derived class of `AbstractValue`, `drake::Value<drake::lcmt_point_cloud>`, is not exposed to pybind11, so `get_value` cannot be called. See `AddValueInstantiation` for how to bind it.
```

I assume I need to bind that LCM message type as suggested.
</blockquote></details>

My suggestion is to use an LcmPublisherSystem so all of the data-handling is in C++.

* Make a Diagram.
* Construct the LcmPublisyerSystem pointing its `lcm=` argument to the `dut._lcm`, with `use_cpp_serializer=True`.
* Add the PointCloudToLcm and LcmPublisyerSystem the diagram.
* Connect the PointCloudToLcm output port to the LcmPublisyerSystem input port.
* Build the diagram and create a context.
* Call Publish() on the diagram.  That will send the message (in C++).

zachfang (commented):
Addressed the comments with two remaining issues/items.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NOQ1N6c1qynFI1EM6aJ:b-y056x6)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLf3Ej1yjzD3zJzSqb:-NOPzfcT40gX06lT3mKB:bsgc4f7) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L45)):*

nit The import convention in this file is one-class-per-line (see above).
</blockquote></details>

Done. Made changes to another file as well.
___
*[`bindings/pydrake/visualization/_meldis.py` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOLfkGz8lqnmrpkUDTY:-NOPzn8b6v4pJS_GCjVV:b-38zryc) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L81)):*

See comment here.  This applet was designed only for "viewer" messages.

The point cloud message infra is different, so should be its own separate `class _PointCloudApplet`.
</blockquote></details>

Done. Moved the point cloud pieces to a `__PointCloudApplet` class.
___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOQ-HjwA1g_mTFUl7Ln:b-fqc1ev) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

What did `drake_visualizer` do?  We should aim to do that (at minimum).
</blockquote></details>

Working.  darke_visualizer subscribed to channels with the `DRAKE_POINT_CLOUD_*` prefix, which makes more sense to me. Will look into this next.

```
        self._subscriber = lcmUtils.addSubscriber(
            'DRAKE_POINT_CLOUD_.*',
            .... )
```
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbkxh1ZlWVbE8HN2I:-NOQ-pWC65eCcj3_asgh:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/test/meldis_test.py#L356)):*

My suggestion is to use an LcmPublisherSystem so all of the data-handling is in C++.

* Make a Diagram.
* Construct the LcmPublisyerSystem pointing its `lcm=` argument to the `dut._lcm`, with `use_cpp_serializer=True`.
* Add the PointCloudToLcm and LcmPublisyerSystem the diagram.
* Connect the PointCloudToLcm output port to the LcmPublisyerSystem input port.
* Build the diagram and create a context.
* Call Publish() on the diagram.  That will send the message (in C++).

</blockquote></details>

Done.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOQ-sQSCgdEdb4sHotJ:-NOQ-sQSCgdEdb4sHotK:b1w0g4y) ([raw file](https://github.com/RobotLocomotion/drake/blob/5574799e1295b5abb82d23b1f3e0429970416724/bindings/pydrake/visualization/test/meldis_test.py#L356)):*
> ```Python
> 
>         """
>         cloud = PointCloud(
> ```

Working. My intent is to publish one `XyzRgbNormal` cloud to increase the coverage. However, I think some information is missing during the PointCloudToLcm conversion since it complains about a field mismatch. From what I read, setting proper mutable fields should be enough, and the same PointCloud can be displayed in Meldis when bypassing LCM de/serialization.

```
RuntimeError: PointCloud does not have the exact expected fields.
Expected (kXYZs | kNormals | kRGBs), got (kXYZs)

```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri and @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOfTfrY4sUjjR9M1j0y:b-be7h7j) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

Working.  darke_visualizer subscribed to channels with the `DRAKE_POINT_CLOUD_*` prefix, which makes more sense to me. Will look into this next.

```
        self._subscriber = lcmUtils.addSubscriber(
            'DRAKE_POINT_CLOUD_.*',
            .... )
```
</blockquote></details>

Another discussion point-

From my understanding, LCM, by default, can take a regex pattern. There are two ways to subscribe to arbitrary channels with the prefix "DRAKE_POINT_CLOUD_*".

(1) Use an LCM object directly and add some of the lower-level LCM operations.
(2) Extend DrakeLcm. First of all, we only bind `Subscribe()` to Python, but not `SubscribeAllChannels()`. SubscribeAllChannels() does mostly the job needed here, but it currently consumes a suffix, not a prefix. I know we discussed that when adding the suffix infrastructure.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOQ-sQSCgdEdb4sHotJ:-NOfQL4t7w-jsx0L7o5e:bx5f50x) ([raw file](https://github.com/RobotLocomotion/drake/blob/5574799e1295b5abb82d23b1f3e0429970416724/bindings/pydrake/visualization/test/meldis_test.py#L356)):*

Working. My intent is to publish one `XyzRgbNormal` cloud to increase the coverage. However, I think some information is missing during the PointCloudToLcm conversion since it complains about a field mismatch. From what I read, setting proper mutable fields should be enough, and the same PointCloud can be displayed in Meldis when bypassing LCM de/serialization.

```
RuntimeError: PointCloud does not have the exact expected fields.
Expected (kXYZs | kNormals | kRGBs), got (kXYZs)

```
</blockquote></details>

After some digging, the problem arises when the PointCloud in Python is somehow passed through an assignment operation to the PointCloudToLcm system while (I believe) the same code in C++ goes through the copy-constructor.

The [copy-constructor](https://github.com/RobotLocomotion/drake/blob/33147ae5c4c7ba5820eefa02544c6f9f15db20b9/perception/point_cloud.cc#L156-L159) will create a point cloud that has the same size and fields. However, if a point cloud is generated first and then assigned to another point cloud (we created in the test in Python), we require they have [the same fields](https://github.com/RobotLocomotion/drake/blob/33147ae5c4c7ba5820eefa02544c6f9f15db20b9/perception/point_cloud.cc#L232-L233), which is by default only XYZ.

I have added a few lines to force update the size and fields in the assignment operation for illustration purposes. This is my current assessment, if that makes sense. :thinking: 

```
cloud_to_lcm.get_input_port().FixValue(cloud_context, cloud)
```

jwnimmer-tri:
*[`bindings/pydrake/visualization/test/meldis_test.py` line 356 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOQ-sQSCgdEdb4sHotJ:-NOjPmAl0rfUdrJUtGNn:bdi0jlc) ([raw file](https://github.com/RobotLocomotion/drake/blob/5574799e1295b5abb82d23b1f3e0429970416724/bindings/pydrake/visualization/test/meldis_test.py#L356)):*

After some digging, the problem arises when the PointCloud in Python is somehow passed through an assignment operation to the PointCloudToLcm system while (I believe) the same code in C++ goes through the copy-constructor.

The [copy-constructor](https://github.com/RobotLocomotion/drake/blob/33147ae5c4c7ba5820eefa02544c6f9f15db20b9/perception/point_cloud.cc#L156-L159) will create a point cloud that has the same size and fields. However, if a point cloud is generated first and then assigned to another point cloud (we created in the test in Python), we require they have [the same fields](https://github.com/RobotLocomotion/drake/blob/33147ae5c4c7ba5820eefa02544c6f9f15db20b9/perception/point_cloud.cc#L232-L233), which is by default only XYZ.

I have added a few lines to force update the size and fields in the assignment operation for illustration purposes. This is my current assessment, if that makes sense. :thinking: 

```
cloud_to_lcm.get_input_port().FixValue(cloud_context, cloud)
```
</blockquote></details>

Yes, your diagnosis is correct.

Here's[1] where the Python setting happens.  It creates an empty value, and then populates it via `set_value` which will use the assignment operator for copyable types.

[1]
https://github.com/RobotLocomotion/drake/blob/239f0a4016bfcdcdefa8dd5ef1759a2a4fabe391/bindings/pydrake/systems/framework_py_semantics.cc#L760-L770

In contrast, C++ calls to FixValue will typically have the static type of `cloud` available, so can clone it directly instead of assigning it.

Note though the comment on L760.  If we scroll up a tiny bit, we can see that arguments that are already type-erased won't get a call to `set_value`.

Therefore, until we fix PointCloud we can work around the issue here by type-erasing the cloud before setting it:

```patch
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -24,6 +24,9 @@ from drake import (
 from pydrake.common import (
     FindResourceOrThrow,
 )
+from pydrake.common.value import (
+    Value,
+)
 from pydrake.geometry import (
     DrakeVisualizer,
     DrakeVisualizerParams,
@@ -373,7 +376,8 @@ class TestMeldis(unittest.TestCase):
         # Set input and publish the point cloud.
         context = diagram.CreateDefaultContext()
         cloud_context = cloud_to_lcm.GetMyContextFromRoot(context)
-        cloud_to_lcm.get_input_port().FixValue(cloud_context, cloud)
+        cloud_to_lcm.get_input_port().FixValue(
+            cloud_context, Value[PointCloud](cloud))
         diagram.ForcedPublish(context)
 
         # Check the Meshcat path before and after the message handling.
```

jwnimmer-tri:
This PR has unearthed a weird facet of PointCloud -- that the `operator=` functions might throw or generally fail when the fields types differ.  We should fix that (in a different PR).  Would you like do to that @zachfang?  I'm happy to do it myself as well, if you like.

The goal would be to change (and unit test) both `PointCloud::operator=` functions to allow them to change fields of `this`.  Since we already have a `PointCloud::SetFrom` function that allows for morphing fields, the promise in the class overview docs that the fields never change is already incorrect, and having a class provide an `operator=` that can't actually assign in some cases is just too weird of a condition to impose on users. Beyond fixing the code, we'd also update the class overview docs to explain the new reality.  In the code, we should be sure to preserve efficiency of assignments when the two clouds _were_ the same size and fields-type (i.e., only re-allocate when necessary).

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOjTCT-9G0sf6Opl98C:-NOjZ0Lv-phZpZe_Yw6Z:b-tvri2h) (no related file):*

This PR has unearthed a weird facet of PointCloud -- that the `operator=` functions might throw or generally fail when the fields types differ.  We should fix that (in a different PR).  Would you like do to that @zachfang?  I'm happy to do it myself as well, if you like.

The goal would be to change (and unit test) both `PointCloud::operator=` functions to allow them to change fields of `this`.  Since we already have a `PointCloud::SetFrom` function that allows for morphing fields, the promise in the class overview docs that the fields never change is already incorrect, and having a class provide an `operator=` that can't actually assign in some cases is just too weird of a condition to impose on users. Beyond fixing the code, we'd also update the class overview docs to explain the new reality.  In the code, we should be sure to preserve efficiency of assignments when the two clouds _were_ the same size and fields-type (i.e., only re-allocate when necessary).
</blockquote></details>

Sure. I can have a following PR to fix PointCloud `operator=` as I have read through the code quite extensively this week.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOjag8481Br99SXwqHF:b-guf5u0) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

Another discussion point-

From my understanding, LCM, by default, can take a regex pattern. There are two ways to subscribe to arbitrary channels with the prefix "DRAKE_POINT_CLOUD_*".

(1) Use an LCM object directly and add some of the lower-level LCM operations.
(2) Extend DrakeLcm. First of all, we only bind `Subscribe()` to Python, but not `SubscribeAllChannels()`. SubscribeAllChannels() does mostly the job needed here, but it currently consumes a suffix, not a prefix. I know we discussed that when adding the suffix infrastructure.

</blockquote></details>

I'll work on adding this function to DrakeLcm.  The implementation will be somewhat tricky.

```c++
  /**
   * Subscribes to all channels with the given prefix.
   *
   * Only channels whose name begins with `prefix` will be subscribed. The
   * prefix will _not_ be stripped off when calling the handler function.
   */
  virtual std::shared_ptr<DrakeSubscriptionInterface> SubscribeMultichannel(
      std::string_view prefix, MultichannelHandlerFunction) = 0;
```

We could imagine binding `SubscribeAllChannels` in the meantime so that you could do the filtering by hand, but I suspect it's just worth waiting for my PR.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOjpgpO91YFb4RECR-k:bizdf79) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

This PR currently only subscribes to `DRAKE_POINT_CLOUD` channel, but I think the functionality is in place. I can either have another PR to update the multi-channel subscription or wait for your changes and pause for now. Let me know. 
</blockquote></details>

The problem is that there is (to my understanding) no way to run an acceptance test of the UX, since we do not have any examples that publish on that topic.

Unless there's a good way around that, I think we need to wait for the subscription features.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOkBIZk7NhzL7DXR-ns:b-jms1b7) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

The problem is that there is (to my understanding) no way to run an acceptance test of the UX, since we do not have any examples that publish on that topic.

Unless there's a good way around that, I think we need to wait for the subscription features.
</blockquote></details>

=> https://github.com/RobotLocomotion/drake/pull/18826

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 451 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOAbR6L9lQVo5UaZi9K:-NOlaSk38nnHLUViGe_h:bdfs5a5) ([raw file](https://github.com/RobotLocomotion/drake/blob/fd6a588a7c88066332d8300f7da33aae3d512ecc/bindings/pydrake/visualization/_meldis.py#L451)):*

=> https://github.com/RobotLocomotion/drake/pull/18826
</blockquote></details>

Done. Add a separate but similar function to process the channel name into the same workflow.

zachfang:
A quick visualization through the newly added test with 1000 XYZRGBNormal points.

Screenshot from 2023-02-20 20-27-05

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 414 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOlpDCQ03NFzQM2Uiv7:-NOlpDCRCBK3Zwyk4VZ9:bjzdfw5) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L414)):*
> ```Python
> 
>         self._meshcat.SetObject(
>             path="/POINT_CLOUD", cloud=cloud, point_size=0.01
> ```

The meshcat path needs to reflect the serial number part of the channel name. If there are multiple point cloud publishers (each on a different channel), we need to update each one independently so that we always show all of the data.

I'd suggest mapping channel names like so:
```
DRAKE_POINT_CLOUD => /POINT_CLOUD/default
DRAKE_POINT_CLOUD_serial => /POINT_CLOUD/{serial}
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOougXoAvOsDLHa4c8_:b-k3k7nx) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*
> ```Python
>         assert message.num_fields in [3, 4, 7]
>         # Transform the raw data into an N x num_fields array.
>         raw_data = np.frombuffer(message.data, dtype=np.float32).reshape(
> ```

This decoding logic is brittle.  It happens to pass with Drake's current publisher, but other people's publishers might not map the fields the same way.

We need to obey all of the message fields, starting with point_step, row_step,  and byte_offset to obtain the view of any particular channel.  Use field.name to know which channel it is.  Also check endian-ness.

In case we don't want to support certain obscure message options, we could raise an exception (or logger warning) in that case -- in other words, fail-fast.  What we can't do is display data incorrectly, which is what we'd do now.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 363 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOos7BZBwM1sNO2UMtg:-NOos7BZBwM1sNO2UMth:b-spz8tk) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/test/meldis_test.py#L363)):*
nit It would probably substantially help the next person reason this code if we explained what's going on here.  Sample text:

> Given the complexity of the lcmt_point_cloud message format, the easiest way to feed the DUT with a sample point cloud is to use the PointCloudToLcmSystem to publish a PointCloud object onto the DUT's LCM bus directly.

With that comment, to me line 363-382 are all one paragraph of code that implements that idea, so I would nuke the comment on L377 and smoosh it all together under that one umbrella.

Or even better, extract it all into a test class helper function:

```
def _publish_lcm_point_cloud(self, cloud, lcm):
    """Publishes the given PointCloud object as an LCM message on the
    given DrakeLcmInterface."""
```

That way, the actual test case logic is a lot less cluttered.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOpK8yx-XhJy4KlZwNn:b-xyw8cq) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

This decoding logic is brittle.  It happens to pass with Drake's current publisher, but other people's publishers might not map the fields the same way.

We need to obey all of the message fields, starting with point_step, row_step,  and byte_offset to obtain the view of any particular channel.  Use field.name to know which channel it is.  Also check endian-ness.

In case we don't want to support certain obscure message options, we could raise an exception (or logger warning) in that case -- in other words, fail-fast.  What we can't do is display data incorrectly, which is what we'd do now.
</blockquote></details>

In fact, `show_point_cloud.py` at `drake/tools/workspace/drake_visualizer/_drake_visualizer_builtin_scripts` already had satisfactory message decoding logic.  You can just copy that.  It rejects unknown formats.

zachfang (commented):
Will get back to this later today.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NOpPEspBWBT69mIajtK:b-zicjvg)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOpMu-99Wxhzhx_YWDR:b-u2ul6m) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

In fact, `show_point_cloud.py` at `drake/tools/workspace/drake_visualizer/_drake_visualizer_builtin_scripts` already had satisfactory message decoding logic.  You can just copy that.  It rejects unknown formats.
</blockquote></details>

Working.
Yes, I reference some of the processing steps there but didn't know whether we want to only accept a certain type of point cloud.
___
*[`bindings/pydrake/visualization/_meldis.py` line 414 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOlpDCQ03NFzQM2Uiv7:-NOpNHl4EncJ5F-OrWlq:b-wrhh36) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L414)):*

The meshcat path needs to reflect the serial number part of the channel name. If there are multiple point cloud publishers (each on a different channel), we need to update each one independently so that we always show all of the data.

I'd suggest mapping channel names like so:
```
DRAKE_POINT_CLOUD => /POINT_CLOUD/default
DRAKE_POINT_CLOUD_serial => /POINT_CLOUD/{serial}
```
</blockquote></details>

Working.

I did think about that when adding the test. I looked through the code again. I think we don't have a way to access the channel name in this callback function (at least the current infrastructure), do we? In some other functions, they grab identifier-like information from the message, but `lcmt_point_cloud` is relatively generic.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOjTCT-9G0sf6Opl98C:-NOpQXPYADppr5bHL_wD:b-ucfkn4) (no related file):*

Sure. I can have a following PR to fix PointCloud `operator=` as I have read through the code quite extensively this week.
</blockquote></details>

Thanks.

___
___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOpQ3R5454vgeDxVPiD:b-qakbdb) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

Working.
Yes, I reference some of the processing steps there but didn't know whether we want to only accept a certain type of point cloud.
</blockquote></details>

I leave it up to you what variety of message contents we choose to support.

My only invariant is that for any given message, we either need to display it correctly, or not display it at all.
___
*[`bindings/pydrake/visualization/_meldis.py` line 414 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOlpDCQ03NFzQM2Uiv7:-NOpPnIf98FBQES62T3H:b-7xpeo5) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L414)):*

Working.

I did think about that when adding the test. I looked through the code again. I think we don't have a way to access the channel name in this callback function (at least the current infrastructure), do we? In some other functions, they grab identifier-like information from the message, but `lcmt_point_cloud` is relatively generic. 
</blockquote></details>

The `self._message_pending_data` has the channel names (as the dictionary key).  You just need to wire it from there into the applet.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOuj82v9xmCnt33a2S3:b-tof4ly) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

I leave it up to you what variety of message contents we choose to support.

My only invariant is that for any given message, we either need to display it correctly, or not display it at all.
</blockquote></details>

I think three types of clouds, i.e.,  XYZ, XYZGRB, and XYZRGBNormal, make sense to me. I separated the logic into a function.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 363 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOos7BZBwM1sNO2UMtg:-NOujZ3bEroWqoU9iWo5:bcxnmqe) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/test/meldis_test.py#L363)):*

nit It would probably substantially help the next person reason this code if we explained what's going on here.  Sample text:

> Given the complexity of the lcmt_point_cloud message format, the easiest way to feed the DUT with a sample point cloud is to use the PointCloudToLcmSystem to publish a PointCloud object onto the DUT's LCM bus directly.

With that comment, to me line 363-382 are all one paragraph of code that implements that idea, so I would nuke the comment on L377 and smoosh it all together under that one umbrella.

Or even better, extract it all into a test class helper function:

```
def _publish_lcm_point_cloud(self, cloud, lcm):
    """Publishes the given PointCloud object as an LCM message on the
    given DrakeLcmInterface."""
```

That way, the actual test case logic is a lot less cluttered.
</blockquote></details>

Done. Moved them into a helper function. If you think the L377 comment is not needed, I can remove that too.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOukTTTEuDkboiwEaT3:bwn4eji) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

I think three types of clouds, i.e.,  XYZ, XYZGRB, and XYZRGBNormal, make sense to me. I separated the logic into a function.
</blockquote></details>

It's fine to cover all three, but then we need more unit tests that exercise each allowed option in turn.
___
*[`bindings/pydrake/visualization/_meldis.py` line 377 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOukgqvCNZuS8eeR3x0:-NOukgqvCNZuS8eeR3x1:b6ypqnb) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L377)):*
> ```Python
>         self._meshcat = meshcat
> 
>     _POINT_CLOUD_FIELDS = (
> ```

nit I think static data usually precedes the constructor?  I can't find a Python style rule that says one or the other ...
___
*[`bindings/pydrake/visualization/_meldis.py` line 407 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOulLfg15zEHY5xeoPP:-NOulLfg15zEHY5xeoPQ:b-qpwgou) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L407)):*
> ```Python
>             ):
>                 return False
>         return True
> ```

nit If this function returned the `Fields` value, it would make the call site somewhat less brittle, e.g., `Fields(BaseField.kXYZs | BaseField.kRGBs)`.
___
*[`bindings/pydrake/visualization/_meldis.py` line 416 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOukzkh9TemElLD6eC1:-NOukzki8mFAEWXSV7gD:bw7da9h) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L416)):*
> ```Python
>         """
>         if not self._validate_point_cloud(message):
>             _logger.warn("Unsupported point cloud data type.")
> ```

nit This will spam the user at 10 Hz .. 30 Hz or something, in case they are using an unsupported format.  We need to throttle the warning rate somehow -- maybe just once per process, or once per channel name, or once every 10 seconds, or ...

zachfang (commented):
All the comments addressed. PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NOwH-h06zx67_J5gArA:bmve7w3)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 385 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOougXoAvOsDLHa4c8Z:-NOwFvhNBEIklNKzEFd4:blbvxj) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L385)):*

It's fine to cover all three, but then we need more unit tests that exercise each allowed option in turn.
</blockquote></details>

Done. Modified the unit test to iterate through three types of point clouds.
___
*[`bindings/pydrake/visualization/_meldis.py` line 414 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOlpDCQ03NFzQM2Uiv7:-NOwG8Vn-bYXh5ZHMo2v:b2in7ve) ([raw file](https://github.com/RobotLocomotion/drake/blob/07facbc5ad9e926c95727c5f6b8b597c67fe0279/bindings/pydrake/visualization/_meldis.py#L414)):*

The `self._message_pending_data` has the channel names (as the dictionary key).  You just need to wire it from there into the applet.
</blockquote></details>

Done. I have been thinking what's the best way to achieve that, and I think `_invoke_subscription()` is the place we have the message-channel-name association.
___
*[`bindings/pydrake/visualization/_meldis.py` line 407 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOulLfg15zEHY5xeoPP:-NOwF_zc29YLZHHPV2SD:b-bbdlds) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L407)):*

nit If this function returned the `Fields` value, it would make the call site somewhat less brittle, e.g., `Fields(BaseField.kXYZs | BaseField.kRGBs)`.
</blockquote></details>

Done. Changed the function name to reflect that.
___
*[`bindings/pydrake/visualization/_meldis.py` line 416 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOukzkh9TemElLD6eC1:-NOwFmuI8hPmv_PQpmKy:b-nt4o72) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L416)):*

nit This will spam the user at 10 Hz .. 30 Hz or something, in case they are using an unsupported format.  We need to throttle the warning rate somehow -- maybe just once per process, or once per channel name, or once every 10 seconds, or ...
</blockquote></details>

Done. Only print once for each channel.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-PX4BntpNuVChAPf:-NOz-PX4BntpNuVChAPg:bqqi3yr) (no related file):*
Working

Once the code is finished, I'll test this locally -- in particular, to confirm the RGB decoding.

___
___
*[`bindings/pydrake/visualization/_meldis.py` line 416 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOukzkh9TemElLD6eC1:-NOyzBA50szMfUr7uMi1:btl6gj1) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L416)):*

Done. Only print once for each channel.
</blockquote></details>

Are you sure?  I don't see how `is_new_channel` ever goes True for unsupported formats.

For muting messages, if we're going to do it manually instead of using a built-in warning function, then do the simple thing: have a `set` of `self._already_warned_channel_names`, and check/add vs that set in the warning branch (cloud_fields is None) only.

Relatedly, the `channel_to_meshcat_path` is too awkward.  I like that it's a separate function, but it's obscure to launder its result via a dict in `self` instead of just a nice and boring return value.  We should just have `path = self._channel_to_meshcat_path(channel)` and call it a day, no dictionary.
___
*[`bindings/pydrake/visualization/_meldis.py` line 390 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-7uO3YldEmO8gYsz:-NOz-7uP23Of_oqjRIKI:b-md4kck) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/_meldis.py#L390)):*
> ```Python
>         self._channel_to_meshcat_path = {}
> 
>     def _validate_and_get_fields(self, message):
> ```

nit This needs to check the message's endianness.
___
*[`bindings/pydrake/visualization/_meldis.py` line 606 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOyxoPABC8RR9u32XmY:-NOyxoPB033T7SFYbCwU:b-7u9hxb) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/_meldis.py#L606)):*
> ```Python
>         for channel, data in self._message_pending_data.items():
>             message = self._message_types[channel].decode(data)
>             for function in self._message_handlers[channel]:
> ```

We should just to the simple thing here, no special-case.  All handlers will receive the channel name.

The trick is then in `def _subscribe` we need to wrap the user's handler to discard the channel name.  Something like this:

```
def multi_handler(*, channel, message):
    handler(message)
self._message_handlers.setdefault(channel, []).append(multi_handler)
```

In the alternative, we could add "channel" argument to the existing handlers, and just have them ignore it as unnecessary, but it keeps the callback signature the same everywhere.



_Suggestion:_
```Python
            for function in self._message_handlers[channel]:
                function(channel=channel, message=message)
```
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 78 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOyw8ZsCFimZdntFLzW:-NOyw8ZsCFimZdntFLzX:b7tunqk) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/test/meldis_test.py#L78)):*
> ```Python
>     def _create_point_cloud(self, num_fields):
>         point_cloud_fields = None
>         if num_fields == 3:
> ```

nit Unused `point_cloud_fields`



_Suggestion:_
```Python
    def _create_point_cloud(self, num_fields):
        if num_fields == 3:
```
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 82 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOywFcW073KTvNhbnps:-NOywFcX2VAF1yd00Sqj:bt4x1d2) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/test/meldis_test.py#L82)):*
> ```Python
>         elif num_fields == 4:
>             cloud_fields = Fields(BaseField.kXYZs | BaseField.kRGBs)
>         else:
> ```

nit Better to fail-fast when a future programmer makes a mistake.



_Suggestion:_
```Python
        else:
            assert num_fields == 7
            cloud_fields = Fields(
                BaseField.kXYZs | BaseField.kRGBs | BaseField.kNormals
            )
```
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 405 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOywngP1kbh6g2Hkj3U:-NOywngQ9y_ch9ttzfLK:b-kh43fw) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/test/meldis_test.py#L405)):*
> ```Python
>            (7, "DRAKE_POINT_CLOUD_12345", "/POINT_CLOUD/12345"),
>         )
>         for num_fields, channel, meshcat_path in test_tuples:
> ```

BTW If you haven't seen it yet, take a look at the `unittest.subTest` context manager and consider using it here within the loop for improved error messages.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 416 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOukzkh9TemElLD6eC1:-NOzH3JA5icLrp9IIBhz:bgsxegk) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff94cb356a019d6a08d9a6245b0f3b8a67a9b11/bindings/pydrake/visualization/_meldis.py#L416)):*
Done. My original thought was to use one data structure to guard the warnings and server as a lookup table.

> if we're going to do it manually instead of using a built-in warning function

Is there a built-in way to throttle messages? I looked through `Logger` class yesterday, and I didn't find relevant pieces.
___
*[`bindings/pydrake/visualization/_meldis.py` line 390 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-7uO3YldEmO8gYsz:-NOzGWYcAMS3E23x5iEg:b-6w2hq0) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/_meldis.py#L390)):*

nit This needs to check the message's endianness.
</blockquote></details>

Does that translate to this?
```
if message.flags != lcmt_point_cloud.IS_STRICTLY_FINITE and message.flags != lcmt_point_cloud.IS_BIGENDIAN:
``` 
___
*[`bindings/pydrake/visualization/_meldis.py` line 606 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOyxoPABC8RR9u32XmY:-NOzHN4L4gtWn-UsdAKm:b-26ohlm) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/_meldis.py#L606)):*

We should just to the simple thing here, no special-case.  All handlers will receive the channel name.

The trick is then in `def _subscribe` we need to wrap the user's handler to discard the channel name.  Something like this:

```
def multi_handler(*, channel, message):
    handler(message)
self._message_handlers.setdefault(channel, []).append(multi_handler)
```

In the alternative, we could add "channel" argument to the existing handlers, and just have them ignore it as unnecessary, but it keeps the callback signature the same everywhere.
</blockquote></details>

Done. I picked the former one to only provide the necessary info to the rest of the callbacks.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 78 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOyw8ZsCFimZdntFLzW:-NOzHZMpD5Evf5bH9Y8g:bxt6ylz) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/test/meldis_test.py#L78)):*

nit Unused `point_cloud_fields`
</blockquote></details>

Done. It was meant to be `cloud_field=None`.
___
*[`bindings/pydrake/visualization/test/meldis_test.py` line 405 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOywngP1kbh6g2Hkj3U:-NOzHf6J6r3ctUKW_EMj:b-ok3u11) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/test/meldis_test.py#L405)):*

BTW If you haven't seen it yet, take a look at the `unittest.subTest` context manager and consider using it here within the loop for improved error messages.
</blockquote></details>

Done. Yes, this crossed my mind, and I looked through the Gltf integration test before pushing the commit.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 390 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-7uO3YldEmO8gYsz:-NOzSoaC2gsEwB4qW-fR:bboj32l) ([raw file](https://github.com/RobotLocomotion/drake/blob/bea3bc5b5fe15b9ab18fd4004e8e982dadbf3e6b/bindings/pydrake/visualization/_meldis.py#L390)):*

Does that translate to this?
```
if message.flags != lcmt_point_cloud.IS_STRICTLY_FINITE and message.flags != lcmt_point_cloud.IS_BIGENDIAN:
``` 
</blockquote></details>

My mistake.  I forgot flags was already bitfield, so we're _already_ check that it's little endian.  Withdrawn.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-PX4BntpNuVChAPf:-NOzm9uZ8YwJu9l5aalR:b-ufzw5w) (no related file):*

Working

Once the code is finished, I'll test this locally -- in particular, to confirm the RGB decoding.
</blockquote></details>

It looks pretty good, but I don't understand why the 3 clouds have a different opinion of the color of the floor.  I'm going to cross-check versus the drake-visualizer clouds on Focal.

___

jwnimmer-tri (commented):
:lgtm:

+@EricCousineau-TRI for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NP-6_4H-TLzCqL-UZ2d:b-nbay8p)* status: LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NOz-PX4BntpNuVChAPf:-NP-6TTt3gfcYMyYWGzD:bjxljaq) (no related file):*

It looks pretty good, but I don't understand why the 3 clouds have a different opinion of the color of the floor.  I'm going to cross-check versus the drake-visualizer clouds on Focal.
</blockquote></details>

OK It's the same on Focal with drake_visualizer.  I'm not sure why the clouds are colored that way, but _shrug_ it's not any worse in from this PR.

___

jwnimmer-tri:
-@EricCousineau-TRI +@ggould-tri for platform review per schedule, please.

jwnimmer-tri:
We recently changed our CI config, so we need to:

jwnimmer-tri:

ggould-tri (commented):
:lgtm:; one discretionary magic-numbers quibble.

> Reviewed 1 of 2 files at r5, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-:-NPIeLTC5Oa9fPLkoDXA:bmqlioa)* status: 1 unresolved discussion (waiting on @zachfang)

___
*[`bindings/pydrake/visualization/_meldis.py` line 398 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NPIf-BJAuap0Q3Q49dy:-NPIf-BJAuap0Q3Q49dz:b-2i1pf) ([raw file](https://github.com/RobotLocomotion/drake/blob/9f6233cce9120e2d41bc51fb32dd3a2c1e361ea4/bindings/pydrake/visualization/_meldis.py#L398)):*
> ```Python
>             return None
> 
>         if message.num_fields not in (3, 4, 7):
> ```

minor:  `3` is the clearest way to write the cardinality of xyz, but `4` and `7` seem a bit magic-number-ish.  Consider moving them to named constants -- `_NORMALS_FIELD_INDEX` and `_MAX_NUM_FIELDS` or something like that?

If it's too annoying with all the off-by-one-ing, maybe just comments pointing to `_POINT_CLOUD_FIELDS` and explaining the layout semantics there?

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),jwnimmer-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_meldis.py` line 398 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/18774#-NPIf-BJAuap0Q3Q49dy:-NPJ24oY89eGqg5U0z83:bph3efn) ([raw file](https://github.com/RobotLocomotion/drake/blob/9f6233cce9120e2d41bc51fb32dd3a2c1e361ea4/bindings/pydrake/visualization/_meldis.py#L398)):*

minor:  `3` is the clearest way to write the cardinality of xyz, but `4` and `7` seem a bit magic-number-ish.  Consider moving them to named constants -- `_NORMALS_FIELD_INDEX` and `_MAX_NUM_FIELDS` or something like that?

If it's too annoying with all the off-by-one-ing, maybe just comments pointing to `_POINT_CLOUD_FIELDS` and explaining the layout semantics there?
</blockquote></details>

Done. A comment is added for `_POINT_CLOUD_FIELDS`.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18774)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),jwnimmer-tri(platform) (waiting on @zachfang)

svenevs:
Changing release notes to `none` as this is being reverted in #18885 [per discussion on slack](https://drakedevelopers.slack.com/archives/C270MN28G/p1677530370736159?thread_ts=1677523550.314859&cid=C270MN28G).

PR #18562, opened 2023-01-10
Add bounding box or rotated lorentz cone constraint for small psd matrices

hongkai-dai:
When calling NewSosPolynomial, if the gram matrix has size 1x1, add a bounding box constraint. if the gram matrix has size 2x2, add a rotated lorentz cone constraint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18562)

PR #18789, opened 2023-02-15
Add bounding box or rotated lorentz cone constraint for small psd matrices

hongkai-dai:
When calling NewSosPolynomial, if the gram matrix has size 1x1, add a bounding box constraint. if the gram matrix has size 2x2, add a rotated lorentz cone constraint.

This PR was submitted in #18562, but was reverted later due to failure on mac. Now I fixed the Mac failure and submit it again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18789)

### DIFF
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -730,33 +730,36 @@ void BindMathematicalProgram(py::module m) {
               MathematicalProgram::*)(
               const Eigen::Ref<const VectorX<Monomial>>&,
               MathematicalProgram::NonnegativePolynomial,
-              const std::string& gram_name)>(
+              const std::string& gram_name, bool small_gram_as_lorentz_cone)>(
               &MathematicalProgram::NewSosPolynomial),
           py::arg("monomial_basis"),
           py::arg("type") = MathematicalProgram::NonnegativePolynomial::kSos,
           py::arg("gram_name") = "S",
+          py::arg("small_gram_as_lorentz_cone") = true,
           doc.MathematicalProgram.NewSosPolynomial
-              .doc_3args_monomial_basis_type_gram_name)
+              .doc_4args_monomial_basis_type_gram_name_small_gram_as_lorentz_cone)
       .def("NewSosPolynomial",
           static_cast<Polynomial (MathematicalProgram::*)(
               const Eigen::Ref<const MatrixX<symbolic::Variable>>&,
               const Eigen::Ref<const VectorX<Monomial>>&,
-              MathematicalProgram::NonnegativePolynomial)>(
+              MathematicalProgram::NonnegativePolynomial, bool)>(
               &MathematicalProgram::NewSosPolynomial),
           py::arg("gramian"), py::arg("monomial_basis"),
           py::arg("type") = MathematicalProgram::NonnegativePolynomial::kSos,
+          py::arg("small_gram_as_lorentz_cone") = true,
           doc.MathematicalProgram.NewSosPolynomial
-              .doc_3args_gramian_monomial_basis_type)
+              .doc_4args_gramian_monomial_basis_type_small_gram_as_lorentz_cone)
       .def("NewSosPolynomial",
           static_cast<std::pair<Polynomial, MatrixXDecisionVariable> (
               MathematicalProgram::*)(const Variables&, int,
-              MathematicalProgram::NonnegativePolynomial, const std::string&)>(
-              &MathematicalProgram::NewSosPolynomial),
+              MathematicalProgram::NonnegativePolynomial, const std::string&,
+              bool)>(&MathematicalProgram::NewSosPolynomial),
           py::arg("indeterminates"), py::arg("degree"),
           py::arg("type") = MathematicalProgram::NonnegativePolynomial::kSos,
           py::arg("gram_name") = "S",
+          py::arg("small_gram_as_lorentz_cone") = true,
           doc.MathematicalProgram.NewSosPolynomial
-              .doc_4args_indeterminates_degree_type_gram_name)
+              .doc_5args_indeterminates_degree_type_gram_name_small_gram_as_lorentz_cone)
       .def("NewEvenDegreeNonnegativePolynomial",
           &MathematicalProgram::NewEvenDegreeNonnegativePolynomial,
           py::arg("indeterminates"), py::arg("degree"), py::arg("type"),
--- a/bindings/pydrake/solvers/test/mathematicalprogram_test.py
+++ b/bindings/pydrake/solvers/test/mathematicalprogram_test.py
@@ -642,7 +642,7 @@ class TestMathematicalProgram(unittest.TestCase):
         (poly1, gramian1) = prog.NewSosPolynomial(
             indeterminates=sym.Variables(x), degree=4,
             type=mp.MathematicalProgram.NonnegativePolynomial.kSdsos,
-            gram_name="M0")
+            gram_name="M0", small_gram_as_lorentz_cone=True)
         self.assertIsInstance(poly1, sym.Polynomial)
         self.assertIsInstance(gramian1, np.ndarray)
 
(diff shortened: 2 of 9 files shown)


hongkai-dai (commented):
+@SeanCurtis-TRI for feature review please because you reverted the original PR in #18648.

To reviewers: this PR has two commits. The first commit has been reviewed in #18562. You can choose to only review the second commit which attempts to pass the Mac CI.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-:-NOHyiL23PuzRO1rZozl:b8haqf7)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

hongkai-dai:
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-address-sanitizer please
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-debug please
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-address-sanitizer please
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-debug please
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-everything-address-sanitizer please
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-everything-release please
@drake-jenkins-bot mac-x86-monterey-unprovisioned-clang-bazel-experimental-release please

SeanCurtis-TRI (commented):
+(release notes: feature) 

Presumably, it is your intent to merge these and the current separation into commits is intended to be an aid (and a very appreciated aid) to review, right?

----------------------------------------------------------------

After careful rumination, I don't think that the solution in the second commit is quite the right way to go about addressing the CSDP issue.

__Assumption__: we don't trust csdp on mac to do the right thing for small SOS problems with these new constraints. (Ignore everything else if this assumption is wrong.)

If that's the case, this patch makes *our* unit tests ignore the problem and leaves the trap open for users to fall into. If it's not trustworthy for our unit tests, would it be trustworthy for a user?

__Conclusion__: Whatever mechanism we use to get away from the "bad" area, we need to make available to the user.

__Proposal__: Expose the "use SDP constraint for small SOS" choice to the user through the API.

Looking at the `NewSosPolynomial()` family of functions, it doesn't seem that adding an argument there would be very ergonomic. Furthermore, it would be nice to leave those APIs untouched. Instead, I'd add a member to `MathematicalProgram` (with the appropriate getter/setter). If the user knows they'll be solving it via CSDP, they should turn off the SOS optimization before defining their problem and the "correct" thing should simply happen under the hood by calling `NewSosPolynomial()`.

The behavior and API would be carefully documented giving the context for why they should ever consider calling it letting them know that the API will go away in the future when the CSDP issues are resolved. We don't even have to get into the nitty gritty. We can simply say, "If you're doing SOS and CSDP, call this particular method after constructing the program, but before defining the math." A very simply rubric that shouldn't hurt them.

If and when we're able to figure out why CSDP isn't happy and can make it trustworthy again, we can simply ignore the flag and deprecate the API because it won't be necessary any longer.

What do you think?

(FTR, it would be *very* cool if the mathematical program could detect CSDP and modify itself automatically....but that feels *way* too much for such a weird corner case.)

> Reviewed 4 of 4 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-:-NOKesn986brg4q6SFbH:bbxksfv)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @hongkai-dai)

___
*[`solvers/test/csdp_solver_test.cc` line 438 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-NOKgpMl07ysJTnv6mFR:-NOKgpMl07ysJTnv6mFS:bry4t09) ([raw file](https://github.com/RobotLocomotion/drake/blob/a8a2202b161c037933eabe2c5be0514e090320a5/solvers/test/csdp_solver_test.cc#L438)):*
> ```cpp
> 
> GTEST_TEST(TestSOS, UnivariateNonnegative1) {
>   UnivariateNonnegative1 dut{true /*small_sos_as_sdp */};
> ```

This test is opting out of the new constraint spelling that this PR adds. So, the behavior that this *test* represents is not the behavior that someone *using* csdp to do a SOS problem would see.

Seems to me that's a bit tenuous and requires some clear justification for why this is a good thing.

As I consider this, it almost seems that the true solution is to modify the API and allow someone adding the SOS constraint to opt out of the optimization. Then it's part of the public API and we can explicitly document to the end user why they might want to do that. Then, we're explicitly testing what the API recommends.
___
*[`solvers/test/sos_examples.h` line 135 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-NOKfXKi6TbGLL2u4hFM:-NOKfXKi6TbGLL2u4hFN:b-8knmln) ([raw file](https://github.com/RobotLocomotion/drake/blob/a8a2202b161c037933eabe2c5be0514e090320a5/solvers/test/sos_examples.h#L135)):*
> ```objc
>   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(UnivariateNonnegative1)
> 
>   // @param small_sos_as_sdp If the sos Gram matrix is small (like 1 x 1 or 2 x
> ```

BTW As I read this documentation, I'm not 100% sure what to make of it. The keypoints I'm inferring are:

1. the new constraints mysteriously fail under mac -- and we don't know why.
2. This parameter allows particular unit tests to opt out of the new constraints and rely on the old, seemingly more universally successful constraint.
3. This opt out is not particularly desirable, but it is currently necessary. Therefore, it shouldn't be interpreted as a positive example of "goodness".

This test class is used across multiple solver tests, but this parameter is only used for the csdp test.

------

I'd written more, but after consideration, I'm going to double down on my previous note about *where* this "hack" goes.

hongkai-dai (commented):
@SeanCurtis-TRI that is excellent suggestion. Now I add an option in `NewSosPolynomial` which would opt in/out adding rotated Lorentz cone constraint for small size Gram matrix. 

One simple explanation for why CSDP doesn't like the new formulation, is that CSDP doesn't have a good routine for rotated Lorentz cone constraint. Instead it will adds a large matrix positive semidefinite constraint, larger than the original size of the gram matrix. This larger-size formulation is more expensive than the formulation if we set `small_gram_as_lorentz_cone` to false.

+@RussTedrake for feature review also since you reviewed the original PR #18562. Only the second commit of this PR is new so you might want to review the second commit only.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-:-NOLMxpR98cwLAiqKVQF:bjyojht)* status: 2 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),RussTedrake(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake and @SeanCurtis-TRI)

___
*[`solvers/test/csdp_solver_test.cc` line 438 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-NOKgpMl07ysJTnv6mFR:-NOLO12R6Yq1k-UeoZJV:b-9e4t0i) ([raw file](https://github.com/RobotLocomotion/drake/blob/a8a2202b161c037933eabe2c5be0514e090320a5/solvers/test/csdp_solver_test.cc#L438)):*

This test is opting out of the new constraint spelling that this PR adds. So, the behavior that this *test* represents is not the behavior that someone *using* csdp to do a SOS problem would see.

Seems to me that's a bit tenuous and requires some clear justification for why this is a good thing.

As I consider this, it almost seems that the true solution is to modify the API and allow someone adding the SOS constraint to opt out of the optimization. Then it's part of the public API and we can explicitly document to the end user why they might want to do that. Then, we're explicitly testing what the API recommends.
</blockquote></details>

Done. Thanks for the excellent suggestion. I modified the API of NewSosPolynomial to allow the user to opt in/out Lorentz cone formulation.

jwnimmer-tri:
I haven't read the code at all here -- but my mental model for MathematicalProgram is that the user shouldn't need to care about this (option).  They call the function with their _intent_, and then in the solver back-ends we _automatically_ transform it to the best encoding for that particular solver.

hongkai-dai (commented):
OK, I will mark this PR as "Do not review". I will file a separate PR to fix the problem inside CsdpSolver, so that it can automatically transform the MathematicalProgram to the good form that CSDP likes.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18789#-:-NOLWXTl8u6Tf5QJg6gb:b5zckdh)* status: 2 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),RussTedrake(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake and @SeanCurtis-TRI)

jwnimmer-tri:
What's the plan for proceeding with this PR?

jwnimmer-tri:
The plan is to open a new PR.

Issue #18803, opened 2023-02-16
Parse closed kinematic chains from SDF (and perhaps also URDF)


RussTedrake:
SAP now supports constraints, but currently users can only define those constraints programmatically.

Closed kinematic chains are a common feature in robots (e.g. Cassie, or in many underactuated hands). We should support parsing a robot with a closed kinematic chain directly from SDF and/or URDF.  The specification in URDF is such that one cannot define a closed chain without additional tags, but in SDF it is possible to define a closed chain just by adding the already defined links and joints.  Presumably we through an error in this case, but I have not verified that.

We used to have a Drake-specific "loop_joint" tag.  You can still find vestiges of it in our codebase; the URDF parser even reports `    Error(*node, "loop joints are not supported in MultibodyPlant");` if it encounters one.  I believe @sherm1 feels that it's better for the user to be able to decide where the mechanism should be split and closed with a constraint (vs the true joints), and that it's better to split in the middle of a link and then have a constraint which welds the link back together (vs a lower-DOF constraint implemented at a joint).  I am ok with requiring a custom tag.

A victory condition here would be to have the `four_bar.sdf` and even `strandebeest` use constraints without having to add them programmatically.  They should be specified in the robot description file.

sherm1:
FYI I opened a related draft PR #18804 that can turn these into spanning tree + constraint models.

rpoyner-tri:
There are potentially two pieces of work here. (1) make the parser map closed chains onto appropriate MbP features; (2) make MbP support closed chains at all. Do we believe enough of (2) is done to undertake (1)? Based on Sherm's comment above and other discussion, I'm not sure.

jwnimmer-tri:
My intuition is that we could write a `four_bar.sdf` or `four_bar.urdf` file that uses a to-be-added custom tag to define the constraint, at which point MbP already offers everything we need.  It's probably worth trying to write down that exemplar input file as the next step here.

In that design, the user would still need to point their joints in right direction by hand in the xml, and annotate the constraint by hand in xml, but it would meet the victory condition of not calling MbP API to set up the model.  (Other than to put it into SAP mode, of course.)

sherm1:
There are definitely two different problems, greatly differing in difficulty:
1. Allow users _some_ way to construct tree+constraint models, or
2. Read unmodified sdf files containing only links and joints and automatically turn those into internal tree+constraint systems

[...]

Issue #17429, opened 2022-06-21
Add the ability to add WeldJoint() to a body that is not a "base"


JoseBarreiros-TRI:
Welding a base body to the world_frame can be done by adding a WeldJoint (see figure below), but this doesn't work for any other body that is not the base of the tree (e.g. waist). A method that rejjiters the kinematic tree to make any body the base, so a WeldJoint can be added, will improve usability.  

Drake suggests using a constraint: `RuntimeError: This mobilizer is creating a closed loop since the outboard body already has an inboard mobilizer connected to it. If a physical loop is really needed, consider using a constraint instead.` ; however, a [PositionConstraint](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_position_constraint.html) is not enforced as a weld.

![image](https://user-images.githubusercontent.com/103549637/174842183-724dfb3f-b389-42ff-b13b-f22aa436e8eb.png)

sherm1:
Related: #13040

jwnimmer-tri:
@sherm1 In what way is this not a duplicate of #13040?  I can't seem to discern the difference between the two postings.

sherm1:
I read this one as wanting to create a loop, while #13040 is asking just to make a tree with a different base body. However, I see I may have misread this one -- @JoseBarreiros-TRI do you want to have to system welded to World in two places (forming a loop) or do you just want a tree with the welded body somewhere in the middle (like #13040)?

JoseBarreiros-TRI:
I want to add only one weld to a body that is not a base, like in [13040](https://github.com/RobotLocomotion/drake/issues/13040)

sherm1:
OK, let's keep this issue open since it is current. I'll close #13040 and #9939 (they will remain available for reference).

xuchenhan-tri:
IIUC, the current workaround is the same as that proposed in #13040. Since there's no immediate plan to add this feature, I'll put this one in the backlog for now.

sherm1:
@JoseBarreiros-TRI do you by any chance have the input file that caused this trouble? Or maybe a similar one? I'm close to having this fixed and would like to be able to try the fix against the original issue.

PR #18804, opened 2023-02-16
[multibody] Imported MultibodyGraphMaker from Simbody; modified for Drake

sherm1:
[WIP] Just testing, don't review

This is the ultimate in parsing arbitrary systems with kinematic loops, massless bodies, and whatnot and computing an optimal spanning tree + constraints model suitable for implementation with MultibodyTree. Could be overkill, OTOH it already works for every perverted case I could think of.

Related to #17429 (need for reversed joints) and #18803 (need for kinematic loops)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18804)

### DIFF
--- a/common/type_safe_index.h
+++ b/common/type_safe_index.h
@@ -193,8 +193,13 @@ class TypeSafeIndex {
   /// @name     Utility methods
   ///@{
 
-  /// Implicit conversion-to-int operator.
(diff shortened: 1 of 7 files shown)


sherm1:
Closing in favor of more recent draft #20225

PR #18822, opened 2023-02-19
Improve defaults for AffineSystem and LinearSystem

RussTedrake:
It's always been a pain to produce simple linear/affine systems, because we needed to pass in empty matrices of the correct size (e.g. even if we have no states, the B matrix must have zero rows and number of input columns).  This PR makes the constructor for AffineSystem more robust, setting any empty matrix to an appropriate size. This makes it reasonable to have default values.

This motivation for this was that I was writing a tutorial where the simplest way to produce a system that simple adds an offset to a signal (e.g. to change coordinates), took many lines and lots of explaining. Now we can just have e.g.

```
system = AffineSystem(D=np.eye(2), y0=[1, 2])
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)

### DIFF
--- a/bindings/pydrake/systems/primitives_py.cc
+++ b/bindings/pydrake/systems/primitives_py.cc
@@ -82,9 +82,10 @@ PYBIND11_MODULE(primitives, m) {
                  const Eigen::Ref<const MatrixXd>&,
                  const Eigen::Ref<const MatrixXd>&,
                  const Eigen::Ref<const VectorXd>&, double>(),
-            py::arg("A"), py::arg("B"), py::arg("f0"), py::arg("C"),
-            py::arg("D"), py::arg("y0"), py::arg("time_period") = 0.0,
-            doc.AffineSystem.ctor.doc_7args)
+            py::arg("A") = Eigen::MatrixXd(), py::arg("B") = Eigen::MatrixXd(),
+            py::arg("f0") = Eigen::VectorXd(), py::arg("C") = Eigen::MatrixXd(),
+            py::arg("D") = Eigen::MatrixXd(), py::arg("y0") = Eigen::VectorXd(),
+            py::arg("time_period") = 0.0, doc.AffineSystem.ctor.doc_7args)
         // TODO(eric.cousineau): Fix these to return references instead of
         // copies.
         .def("A", overload_cast_explicit<const MatrixXd&>(&AffineSystem<T>::A),
@@ -106,6 +107,12 @@ PYBIND11_MODULE(primitives, m) {
         // wrapped.
         .def("time_period", &AffineSystem<T>::time_period,
             doc.TimeVaryingAffineSystem.time_period.doc)
+        .def("num_states", &TrajectoryAffineSystem<T>::num_states,
+            doc.TimeVaryingAffineSystem.num_states.doc)
+        .def("num_inputs", &TrajectoryAffineSystem<T>::num_inputs,
+            doc.TimeVaryingAffineSystem.num_inputs.doc)
+        .def("num_outputs", &TrajectoryAffineSystem<T>::num_outputs,
+            doc.TimeVaryingAffineSystem.num_outputs.doc)
         .def("configure_default_state",
             &TimeVaryingAffineSystem<T>::configure_default_state, py::arg("x0"),
             doc.TimeVaryingAffineSystem.configure_default_state.doc)
@@ -214,7 +221,8 @@ PYBIND11_MODULE(primitives, m) {
                  const Eigen::Ref<const MatrixXd>&,
                  const Eigen::Ref<const MatrixXd>&,
                  const Eigen::Ref<const MatrixXd>&, double>(),
-            py::arg("A"), py::arg("B"), py::arg("C"), py::arg("D"),
+            py::arg("A") = Eigen::MatrixXd(), py::arg("B") = Eigen::MatrixXd(),
+            py::arg("C") = Eigen::MatrixXd(), py::arg("D") = Eigen::MatrixXd(),
             py::arg("time_period") = 0.0, doc.LinearSystem.ctor.doc_5args);
 
     DefineTemplateClassWithDefault<MatrixGain<T>, LinearSystem<T>>(
--- a/bindings/pydrake/systems/test/primitives_test.py
+++ b/bindings/pydrake/systems/test/primitives_test.py
@@ -225,6 +225,32 @@ class TestGeneral(unittest.TestCase):
         system.configure_default_state(x0=np.array([1, 2]))
         system.configure_random_state(covariance=np.eye(2))
 
+    def test_linear_affine_system_empty_matrices(self):
+        # Confirm the default values for the system matrices in the
+        # constructor.
+        def CheckSizes(system, num_states, num_inputs, num_outputs):
+            self.assertEqual(system.num_continuous_states(), num_states)
+            self.assertEqual(system.num_inputs(), num_inputs)
+            self.assertEqual(system.num_outputs(), num_outputs)
+
+        # A constant vector system.
+        system = AffineSystem(y0=[2, 1])
+        CheckSizes(system, num_states=0, num_inputs=0, num_outputs=2)
+
+        # A matrix gain.
+        system = AffineSystem(D=np.eye(2))
+        CheckSizes(system, num_states=0, num_inputs=2, num_outputs=2)
+        system = LinearSystem(D=np.eye(2))
+        CheckSizes(system, num_states=0, num_inputs=2, num_outputs=2)
+
+        # Add an offset.
+        system = AffineSystem(D=np.eye(2), y0=[1, 2])
+        CheckSizes(system, num_states=0, num_inputs=2, num_outputs=2)
+
+        # An integrator.
+        system = LinearSystem(B=np.eye(2))
+        CheckSizes(system, num_states=2, num_inputs=2, num_outputs=0)
+
     def test_linear_system_zero_size(self):
         # Explicitly test #12633.
         num_x = 0
--- a/systems/primitives/affine_system.cc
+++ b/systems/primitives/affine_system.cc
@@ -232,25 +232,106 @@ void TimeVaryingAffineSystem<T>::SetRandomState(
 // Our public constructor declares that our most specific subclass is
 // AffineSystem, and then delegates to our protected constructor.
 template <typename T>
-AffineSystem<T>::AffineSystem(const Eigen::Ref<const Eigen::MatrixXd>& A,
-                              const Eigen::Ref<const Eigen::MatrixXd>& B,
-                              const Eigen::Ref<const Eigen::VectorXd>& f0,
-                              const Eigen::Ref<const Eigen::MatrixXd>& C,
-                              const Eigen::Ref<const Eigen::MatrixXd>& D,
-                              const Eigen::Ref<const Eigen::VectorXd>& y0,
-                              double time_period)
-    : AffineSystem<T>(
-          SystemTypeTag<AffineSystem>{},
-          A, B, f0, C, D, y0, time_period) {}
+AffineSystem<T>::AffineSystem(
+    const Eigen::Ref<const Eigen::MatrixXd>& A,
+    const Eigen::Ref<const Eigen::MatrixXd>& B,
+    const Eigen::Ref<const Eigen::VectorXd>& f0,
+    const Eigen::Ref<const Eigen::MatrixXd>& C,
+    const Eigen::Ref<const Eigen::MatrixXd>& D,
+    const Eigen::Ref<const Eigen::VectorXd>& y0,
+
+    double time_period)
+    : AffineSystem<T>(SystemTypeTag<AffineSystem>{}, A, B, f0, C, D, y0,
+                      time_period) {}
 
 namespace {
 
+// Returns the number of states, where any empty matrix is assumed to have the
+// correct size.
+int CalcNumStates(const Eigen::Ref<const Eigen::MatrixXd>& A,
+                  const Eigen::Ref<const Eigen::MatrixXd>& B,
+                  const Eigen::Ref<const Eigen::VectorXd>& f0,
+                  const Eigen::Ref<const Eigen::MatrixXd>& C) {
+  int num_states = 0;
+  if (A.size() > 0) {
+    num_states = A.rows();
+    DRAKE_DEMAND(A.rows() == A.cols());
+  }
+  if (B.size() > 0) {
+    if (num_states) {
+      // This error message is more informative than comparing with num_states.
+      DRAKE_DEMAND(B.rows() == A.rows());
+    } else {
+      num_states = B.rows();
+    }
+  }
+  if (f0.size() > 0) {
+    if (num_states) {
+      DRAKE_DEMAND(f0.size() == num_states);
+    } else {
+      num_states = f0.size();
+    }
+  }
+  if (C.size() > 0) {
+    if (num_states) {
+      DRAKE_DEMAND(C.cols() == num_states);
+    } else {
+      num_states = C.cols();
+    }
+  }
+  return num_states;
+}
+
+// Returns the number of inputs, where any empty matrix is assumed to have the
+// correct size.
+int CalcNumInputs(const Eigen::Ref<const Eigen::MatrixXd>& B,
+                  const Eigen::Ref<const Eigen::MatrixXd>& D) {
+  int num_inputs = 0;
+  if (B.size() > 0) {
+    num_inputs = B.cols();
+  }
+  if (D.size() > 0) {
+    if (num_inputs) {
+      DRAKE_DEMAND(D.cols() == B.cols());
+    } else {
+      num_inputs = D.cols();
+    }
+  }
+  return num_inputs;
+}
+
+// Returns the number of outputs, where any empty matrix is assumed to have the
+// correct size.
+int CalcNumOutputs(const Eigen::Ref<const Eigen::MatrixXd>& C,
+                   const Eigen::Ref<const Eigen::MatrixXd>& D,
+                   const Eigen::Ref<const Eigen::VectorXd>& y0) {
+  int num_outputs = 0;
+  if (C.size() > 0) {
+    num_outputs = C.rows();
+  }
+  if (D.size() > 0) {
+    if (num_outputs) {
+      DRAKE_DEMAND(D.rows() == C.rows());
+    } else {
+      num_outputs = D.rows();
+    }
+  }
+  if (y0.size() > 0) {
+    if (num_outputs) {
+      DRAKE_DEMAND(y0.size() == num_outputs);
+    } else {
+      num_outputs = y0.size();
+    }
+  }
+  return num_outputs;
+}
+
 // Returns whether a matrix is "meaningful" when pre-multiplying a vector.
-bool IsMeaningful(const Eigen::MatrixXd& m) {
+bool IsMeaningful(const Eigen::Ref<const Eigen::MatrixXd>& m) {
   return m.size() > 0 && (m.array() != 0).any();
 }
 
-}  // namespace
+}   // namespace
 
 // Our protected constructor does all of the real work -- everything else
 // delegates to here.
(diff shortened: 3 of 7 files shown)


hongkai-dai (commented):
PTAL.

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOfthIe-AJysBs7OMg3:b-83330m)* status: 5 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`systems/primitives/affine_system.h` line 214 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOfqyBP92UdtMm2Z3QC:-NOfqyBP92UdtMm2Z3QD:b3m0uwa) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/affine_system.h#L214)):*
> ```objc
>   /// Subclasses must use the protected constructor, not this one.
>   AffineSystem(
>       const Eigen::Ref<const Eigen::MatrixXd>& A =
> ```

BTW, what would you think if we use `const std::optional<const Eigen::Ref<const Eigen::MatrixXd>>&` as the type of `A, B, C, D`? Then the user only needs to type `std::nullopt` instead of the longer `Eigen::Matrix<double, 0, 0>()` if he/she means the empty matrix.

The caveat is that I don't know how `std::optional` works with `const Eigen::Ref<const Eigen::MatrixXd>`.
___
*[`systems/primitives/affine_system.cc` line 266 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOfryZ_7IB5P_Z3g_vh:-NOfryZ_7IB5P_Z3g_vi:b-5jaye1) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/affine_system.cc#L266)):*
> ```cpp
>   if (B.size() > 0) {
>     if (num_states) {
>       DRAKE_DEMAND(B.rows() == A.rows());
> ```

BTW. To stay consistent with the code below (f0 and C case), maybe use B.rows() == num_states here?
___
*[`systems/primitives/test/affine_system_test.cc` line 583 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftDSLCc7ms9QdNFsN:-NOftDSLCc7ms9QdNFsO:b-jc2yz8) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L583)):*
> ```cpp
>   CheckSizes(std::make_unique<AffineSystem<double>>(
>                  A, Eigen::Matrix<double, kNumStates, 0>(), f0, C, D, y0),
>              kNumStates, kNumInputs, kNumOutputs);
> ```

Also check the case that B has zero rows and cols?
___
*[`systems/primitives/test/affine_system_test.cc` line 595 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftIDP7BZeTMUNaVVN:-NOftIDP7BZeTMUNaVVO:b-yaerlr) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L595)):*
> ```cpp
>   CheckSizes(std::make_unique<AffineSystem<double>>(
>                  A, B, f0, Eigen::Matrix<double, kNumOutputs, 0>(), D, y0),
>              kNumStates, kNumInputs, kNumOutputs);
> ```

Ditto, also check the case that C has zero rows and cols?
___
*[`systems/primitives/test/affine_system_test.cc` line 603 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftLnp3SpCh0RhrMEE:-NOftLnp3SpCh0RhrMEF:b-89a4l7) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L603)):*
> ```cpp
>   CheckSizes(std::make_unique<AffineSystem<double>>(
>                  A, B, f0, C, Eigen::Matrix<double, kNumOutputs, 0>(), y0),
>              kNumStates, kNumInputs, kNumOutputs);
> ```

Ditto

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @hongkai-dai)

___
*[`systems/primitives/affine_system.h` line 214 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOfqyBP92UdtMm2Z3QC:-NOg-Cu1-6uKfpE-GLuU:b-pv9mmi) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/affine_system.h#L214)):*

BTW, what would you think if we use `const std::optional<const Eigen::Ref<const Eigen::MatrixXd>>&` as the type of `A, B, C, D`? Then the user only needs to type `std::nullopt` instead of the longer `Eigen::Matrix<double, 0, 0>()` if he/she means the empty matrix.

The caveat is that I don't know how `std::optional` works with `const Eigen::Ref<const Eigen::MatrixXd>`.
</blockquote></details>

Done.  You're right.  I had thought of doing this at one point, but somehow went down this worse path.  This version is much better!  Thanks!
___
*[`systems/primitives/affine_system.cc` line 266 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOfryZ_7IB5P_Z3g_vh:-NOg-YClCPgEQd4zRLPY:b6bq3yk) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/affine_system.cc#L266)):*

BTW. To stay consistent with the code below (f0 and C case), maybe use B.rows() == num_states here?
</blockquote></details>

I did this intentionally.  My journey started today when the c++ error message said that something didn't match num_states, which was completely unhelpful (i couldn't tell which other matrix caused num_states to be defined). 

In the code below, I need to be more ambiguous.  But at this point in the code, this version of the error message is correct and more informative.  I've added a comment.
___
*[`systems/primitives/test/affine_system_test.cc` line 583 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftDSLCc7ms9QdNFsN:-NOg-yZ8A549NSmD-NEg:bribuob) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L583)):*

Also check the case that B has zero rows and cols?
</blockquote></details>

Done.  (not relevant any more)
___
*[`systems/primitives/test/affine_system_test.cc` line 595 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftIDP7BZeTMUNaVVN:-NOg0-y-4C-1XIJd9pfc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L595)):*

Ditto, also check the case that C has zero rows and cols?
</blockquote></details>

Done.
___
*[`systems/primitives/test/affine_system_test.cc` line 603 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOftLnp3SpCh0RhrMEE:-NOg00h7-Tm4xkNpRztQ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/test/affine_system_test.cc#L603)):*

Ditto
</blockquote></details>

Done.

hongkai-dai (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOgFJjkEzP9vTpL9E6p:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`systems/primitives/affine_system.cc` line 266 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOfryZ_7IB5P_Z3g_vh:-NOgF4lm37i6ZsY0XpFP:bw53tpu) ([raw file](https://github.com/RobotLocomotion/drake/blob/34e8990c80ca245dfb35bfc68f8f186161727d38/systems/primitives/affine_system.cc#L266)):*

I did this intentionally.  My journey started today when the c++ error message said that something didn't match num_states, which was completely unhelpful (i couldn't tell which other matrix caused num_states to be defined). 

In the code below, I need to be more ambiguous.  But at this point in the code, this version of the error message is correct and more informative.  I've added a comment.
</blockquote></details>

Got it, that makes sense

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOgcHDZ48AgWEfUns69:b-6cxuoj)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @hongkai-dai and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
First pass done -- a request of another change. If you're swamped, let me know and I can push a commit to your PR.

> Reviewed 1 of 7 files at r1, 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOkGreo6cyy7MK435G9:bf4pnrd)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`systems/primitives/affine_system.h` line 214 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjnhu88J_GVKkkYPNI:-NOjnhu88J_GVKkkYPNJ:b-gljt15) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L214)):*
> ```objc
>   ///
>   /// Subclasses must use the protected constructor, not this one.
>   AffineSystem(
> ```

nit: This now needs to be `explicit` because it *could* be invoked with a single parameter leading to ambiguity or implicit construction.
___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOjvjQU-lInkGj0ahYN:bf137q8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*
Ignoring all of the Eigen stuff for a moment, if I had an api that took a `const Foo&` and I wanted to make it optional, it would become `Foo const *`. The name for "optional reference" is "pointer".

We certainly have the infrastructure to support such an Eigen-flavored transformation: `EigenPtr`. That class's documentation includes 

> It can also be used to pass optional Eigen objects since %EigenPtr, unlike `Eigen::Ref`, can be null.

However, that does constitute a change in the API. Matrices would have to be passed by pointer instead of value which precludes passing temporary values - a potentially significant problem when talking about Eigen types.

Here's my recommendation/request:

1. Don't *change* the current constructor.
2. Reimplement *this* constructor as an *additional* constructor using `EigenPtr`.
3. In the future we can consider if we want to deprecate the "everything-required" constructor.

This should give you the full benefit as observed via the python bindings and satisfy the C++ lawyers.
___
*[`systems/primitives/affine_system.h` line 286 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjt1Ur4xnb51B10PXj:-NOjt1Ur4xnb51B10PXk:b-hcnk2l) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L286)):*
> ```objc
>       const Context<T>& context, DiscreteValues<T>* updates) const final;
> 
>   Eigen::MatrixXd A_;
> ```

nit You didn't have to lose the `const` on these; You're initializing them all in the constructor's initializer list.
___
*[`systems/primitives/test/affine_system_test.cc` line 548 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOkCVnX3mOoYPb4_GKD:-NOkCVnX3mOoYPb4_GKE:b3delo0) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/test/affine_system_test.cc#L548)):*
> ```cpp
> }
> 
> // Confirm that empty matrices passed to the constructor are treated as zeros
> ```

nit: You haven't actually confirmed *zeros*, only "correct size".
___
*[`systems/primitives/test/affine_system_test.cc` line 554 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOkCP5p2VJ0WI8eZESq:-NOkCP5qBlgeZ5zSvfY-:bndvx7j) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/test/affine_system_test.cc#L554)):*
> ```cpp
>   const int kNumInputs = 2;
>   const int kNumOutputs = 1;
>   // Default matrices (uninitialized; only their size matters).
> ```

nit: Apparently, gcc under jammy cares.

jwnimmer-tri:
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOkIZAi-kh6FBvtoSo6:b-9i9jmu) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

Ignoring all of the Eigen stuff for a moment, if I had an api that took a `const Foo&` and I wanted to make it optional, it would become `Foo const *`. The name for "optional reference" is "pointer".

We certainly have the infrastructure to support such an Eigen-flavored transformation: `EigenPtr`. That class's documentation includes 

> It can also be used to pass optional Eigen objects since %EigenPtr, unlike `Eigen::Ref`, can be null.

However, that does constitute a change in the API. Matrices would have to be passed by pointer instead of value which precludes passing temporary values - a potentially significant problem when talking about Eigen types.

Here's my recommendation/request:

1. Don't *change* the current constructor.
2. Reimplement *this* constructor as an *additional* constructor using `EigenPtr`.
3. In the future we can consider if we want to deprecate the "everything-required" constructor.

This should give you the full benefit as observed via the python bindings and satisfy the C++ lawyers.

</blockquote></details>

There's another escape hatch here, and one that I would advocate for: decide that the brevity of passing `std::nullopt` vs `Eigen::MatrixXd()` at the call sites isn't actually worth trying to thread this needle.

The original `Eigen::Matrix<double, 0, 0>()` spelling was overkill -- the default constructor `Eigen::MatrixXd()` does the right thing without any extra fixed-size template arguments.  I don't think it's too much to ask for users to write that out.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @RussTedrake)

___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOkKU-oE9DeqG9C9u_5:b-k7fnrl) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

There's another escape hatch here, and one that I would advocate for: decide that the brevity of passing `std::nullopt` vs `Eigen::MatrixXd()` at the call sites isn't actually worth trying to thread this needle.

The original `Eigen::Matrix<double, 0, 0>()` spelling was overkill -- the default constructor `Eigen::MatrixXd()` does the right thing without any extra fixed-size template arguments.  I don't think it's too much to ask for users to write that out.
</blockquote></details>

I'd definitely be on board with that.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @RussTedrake and @SeanCurtis-TRI)

___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOkTsxRCGxAakKfS0xb:bdmvm3p) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

I'd definitely be on board with that.
</blockquote></details>

I am on board with that also.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOkwODH6w3DU1906fPZ:bk02qgv) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

I am on board with that also.
</blockquote></details>

i didn't think brevity was Hongkai's concern... it was ambiguity?

but if i'm hearing it right, I should basically revert to my previous commit and then make that small change?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @RussTedrake and @SeanCurtis-TRI)

___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOl45sw-iO5ZszpvMPQ:b-6ax3e1) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

i didn't think brevity was Hongkai's concern... it was ambiguity?

but if i'm hearing it right, I should basically revert to my previous commit and then make that small change?
</blockquote></details>

My interpretation of his "instead of the longer" was that it was fundamentally about brevity.

Yes, my proposal is to to use this idiom throughout these two headers and ditto in the pydrake defaults:

```c++
const Eigen::Ref<const Eigen::MatrixXd>& A = Eigen::MatrixXd(),
```

hongkai-dai (commented):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @RussTedrake and @SeanCurtis-TRI)

___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOl47WUA3mJq4VAskoR:bjvsa1g) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

My interpretation of his "instead of the longer" was that it was fundamentally about brevity.

Yes, my proposal is to to use this idiom throughout these two headers and ditto in the pydrake defaults:

```c++
const Eigen::Ref<const Eigen::MatrixXd>& A = Eigen::MatrixXd(),
```
</blockquote></details>

Yeah I think Jeremy's suggestion(to use Eigen::MatrixXd() as the default value) resolves brevity, which was my concern. Sorry I made the suggestion to use std::optional.

RussTedrake (commented):
Done.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOlV8f259Cjuhp470Yv:b-19xu5k)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @hongkai-dai and @SeanCurtis-TRI)

___
*[`systems/primitives/affine_system.h` line 214 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjnhu88J_GVKkkYPNI:-NOlKVhw3mNqGVJCakWl:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L214)):*

nit: This now needs to be `explicit` because it *could* be invoked with a single parameter leading to ambiguity or implicit construction.
</blockquote></details>

Done.
___
*[`systems/primitives/affine_system.h` line 215 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjvjQU-lInkGj0ahYM:-NOlKUeK4lWG2GWll2Pp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L215)):*

Yeah I think Jeremy's suggestion(to use Eigen::MatrixXd() as the default value) resolves brevity, which was my concern. Sorry I made the suggestion to use std::optional.
</blockquote></details>

Done.
___
*[`systems/primitives/affine_system.h` line 286 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOjt1Ur4xnb51B10PXj:-NOlKXo5BoU3A-iKNmXb:b-kg6k64) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/affine_system.h#L286)):*

nit You didn't have to lose the `const` on these; You're initializing them all in the constructor's initializer list.
</blockquote></details>

Done.  (i had a previous revision that required dropping the const)
___
*[`systems/primitives/test/affine_system_test.cc` line 548 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOkCVnX3mOoYPb4_GKD:-NOlMBM--LaYfmOY782s:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/test/affine_system_test.cc#L548)):*

nit: You haven't actually confirmed *zeros*, only "correct size".
</blockquote></details>

Done.
___
*[`systems/primitives/test/affine_system_test.cc` line 554 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-NOkCP5p2VJ0WI8eZESq:-NOlSUDG1W4nrGzasVqc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dab602b62a6706dc014d635875fdac3cdf225fb/systems/primitives/test/affine_system_test.cc#L554)):*

nit: Apparently, gcc under jammy cares.  
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18822#-:-NOoVmwz2rpoEa6NV9JX:bmxilld)* status: needs at least two assigned reviewers (waiting on @RussTedrake)

PR #18824, opened 2023-02-20
[C-IRIS] move c_iris_collision_geometry out of dev folder.

hongkai-dai:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)


hongkai-dai (commented):
@SeanCurtis-TRI is it OK if I ask you to platform-review this big PR? It moves the code out of the dev folder. The code has been feature reviewed already.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOhSamy1mmFvsFI65lF:bnaxbes)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
+@SeanCurtis-TRI 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOkS1ZwEw9ezSvrmkDN:b-a74z6d)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
Breakpoint

> Reviewed 4 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOl5x0lB6ABz4QnKFGD:bv8fj6e)* status: 21 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 1 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl5s97CtMg15ATMzmz:-NOl5s97CtMg15ATMzn-:b-bwhyhx) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L1)):*
> ```cpp
> #include "drake/geometry/optimization/test/c_iris_test_utilities.h"
> ```

BTW Per our conversation, let's see if we can't get git to understand that this file moved.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 3 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkUby6AT8pMTvYZUKr:-NOkUby6AT8pMTvYZUKs:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L3)):*
> ```objc
> #pragma once
> 
> #include <optional>
> ```

nit: unused
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 41 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkgZwS788wy80qWAEq:-NOkgZwS788wy80qWAEr:b-mt6p1h) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L41)):*
> ```objc
> 
>   /**
>    @param geometry The actual geometry object.
> ```

Two things:

1. Does it *have* to be an aliased shape? Can't it simply be a locally owned copy?
2. If it does need to be aliased, the documentation should be clear that it *is* aliased and should stay alive for the duration of `this` instance.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 54 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkgu48D5SJKCi9uhA9:-NOkgu48D5SJKCi9uhAA:bo28eet) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L54)):*
> ```objc
>   multibody::BodyIndex body_index() const { return body_index_; }
> 
>   GeometryId id() const { return id_; }
> ```

nit missing namespace



_Suggestion:_
```objc
geometry::GeometryId id() 
```
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkh5Uh99AleNnqmbEc:-NOkh5Uh99AleNnqmbEd:bamffrd) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L59)):*
> ```objc
> 
>   /** Impose the constraint that the geometry is on a given side of the plane {x
>    | aᵀx+b=0}.
> ```

BTW You'd do your C++ reader a favor by putting the entire mathematical statement on a singe line.

```c++
  /** Impose the constraint that the geometry is on a given side of the plane
   {x | aᵀx+b=0}.
  */
```
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl4Shs0e4V4eCe32f9:-NOl4Shs0e4V4eCe32fA:b-ia7xfb) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L81)):*
> ```objc
>    is positive.
> 
>    @param a The normal vector in the separating plane. a is expressed in frame
> ```

nit: I assumed "the normal vector" was unit length. Probably worth noting that it isn't required. (Possibly partnered with a `@pre` that it's length be greater than zero?)
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 92 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkjWUy9JMCbIyXsbuN:-NOkjWUy9JMCbIyXsbuO:bprlefw) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L92)):*
> ```objc
>    @param plane_side Whether the geometry is on the positive or negative side of
>    the plane.
>    @param y_slack The slack variable y in the documentation above, used for
> ```

nit: Probably worth noting the size of `y`.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOktXvS2pJY39dc-jsS:-NOktXvS2pJY39dc-jsT:bla35y6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L94)):*
> ```objc
>    @param y_slack The slack variable y in the documentation above, used for
>    non-polytopic geometries.
>    @param[out] rationals The rational functions that need to be
> ```

nit: It seems that the rational functions representing the half space constraint are *added* to `rationals`. This suggests that they are the only thing in the vector upon completion. The documentation should be clear about that distinction.

(We probably need `@pre rationals != nullptr.`.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 114 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkid4G7AjZVKdMOeMQ:-NOkid4G7AjZVKdMOeMR:b-qhu975) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L114)):*
> ```objc
> 
>  private:
>   const Shape* geometry_;
> ```

nit: requires a moustache (here and on `body_index_`).
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkjBUIAbq6Wv0qqmVh:-NOkjBUIAbq6Wv0qqmVi:b-53d1vd) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L120)):*
> ```objc
> };
> 
> /** Computes the signed distance from `collision_geometry` to the halfspace ℋ,
> ```

BTW geometry typically treats "half space" as a compound noun of two words.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 121 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkjK4PDSEYPBMTvm8x:-NOkjK4PDSEYPBMTvm8y:bt4aeff) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L121)):*
> ```objc
> 
> /** Computes the signed distance from `collision_geometry` to the halfspace ℋ,
>  where ℋ ={ x | aᵀx+b >= 0} if plane_side=PlaneSide::kPositive, and ℋ ={ x |
> ```

nit Weird whitespace (X2)



_Suggestion:_
```objc
ℋ = {x | aᵀx+b >= 0}
```
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 123 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOklZHx1hcfyl5TOefd:-NOklZHx1hcfyl5TOefe:b9zeyp2) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L123)):*
> ```objc
>  where ℋ ={ x | aᵀx+b >= 0} if plane_side=PlaneSide::kPositive, and ℋ ={ x |
>  aᵀx+b <= 0} if plane_side=PlaneSide::kNegative.
>  The halfspace is expressed in the expressed_body's body frame.
> ```

nit: Half spaces are not just expressed, they are measured and expressed.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 23 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkpGXsBdP4DOELdU66:-NOkpGXt6wpqLyyb0csI:b833opo) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L23)):*
> ```cpp
> struct ReifyData {
>   ReifyData(
>       const Vector3<symbolic::Polynomial>& m_a, const symbolic::Polynomial& m_b,
> ```

nit This constructor doesn't get you much. It kills the ability to do `{}` initialization (including designated initializers). It also masks the aliasing (i.e., passing in `const&` and storing `const*`.

You also need `{}` on all of the pointer-valued members. Here's what it would look like.

```c++
struct ReifyData {
  // Temporary container for the quantities necessary to process a shape.
  const Vector3<symbolic::Polynomial>* a{};
  const symbolic::Polynomial* b{};
  const multibody::RationalForwardKinematics::Pose<symbolic::Polynomial>*
      X_AB_multilinear{};
  const multibody::RationalForwardKinematics* rational_forward_kin{};
  const PlaneSide plane_side;
  const GeometryId geometry_id;
  const VectorX<symbolic::Variable>* y_slack{};
  std::vector<symbolic::RationalFunction>* rationals{};
};
```
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkt0_N0eBOUz-YDM-N:-NOkt0_N0eBOUz-YDM-O:b5y6ey9) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L85)):*
> ```cpp
> 
> void ImplementPointRationals(const Eigen::Ref<const Eigen::Matrix3Xd>& p_GV,
>                              const math::RigidTransformd& X_BG, void* data) {
> ```

nit: It'd be nicer if the casting to `ReifyData` happened in the reifier and the properly typed data were were passed here. The goal would be to isolate the footprint of `ReifyData` and reduce the amount of `void*` sprinkled around the code.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkyTdaEWO4O20-AfiB:-NOkyTdaEWO4O20-AfiC:bl8wznx) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L100)):*
> ```cpp
>   // The position of the vertices V in the geometry frame G.
>   for (int i = 0; i < p_GV.cols(); ++i) {
>     reify_data->rationals->push_back(ComputePointOnPlaneSideRational(
> ```

nit: This would be better as `emplace_back`.



_Suggestion:_
```cpp
reify_data->rationals->emplace_back(
```
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 369 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2T7ZCGYS38-dOSUE:-NOl2T7ZCGYS38-dOSUF:b-65b6cq) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L369)):*
> ```cpp
> };
> 
> class CIrisGeometryTypeReifier : public ShapeReifier {
> ```

BTW Here's one that would benefit from simplification:

```c++
class CIrisGeometryTypeReifier : public ShapeReifier {
 public:
  explicit CIrisGeometryTypeReifier(const geometry::Shape& shape) {
    shape.Reify(this);
  }

  CIrisGeometryType type() const { return type_; }

 private:
  using ShapeReifier::ImplementGeometry;
  void ImplementGeometry(const Box&, void*) {
    type_ = CIrisGeometryType::kPolytope;
  }

  void ImplementGeometry(const Convex&, void*) {
    type_ = CIrisGeometryType::kPolytope;
  }

  void ImplementGeometry(const Sphere&, void*) {
    type_ = CIrisGeometryType::kSphere;
  }

  void ImplementGeometry(const Capsule&, void*) {
    type_ = CIrisGeometryType::kCapsule;
  }

  void ImplementGeometry(const Cylinder&, void*) {
    type_ = CIrisGeometryType::kCylinder;
  }

  CIrisGeometryType type_;
};
```

The updated call site is flagged down below.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 410 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2uvc3cqXOjOazzbx:-NOl2uvc3cqXOjOazzby:bkmlyyu) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L410)):*
> ```cpp
> };
> 
> class NumRationalsReifier : public ShapeReifier {
> ```

BTW This gets the same benefit. Instead of storing the shape, have a member `int count_{};`. And `ProcessData` simply becomes:

```c++
int count() const { return count_; }
```

And calls become:

```c++
const int num_rationals = NumRationalsReififer(*geometry_).count();
```
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 467 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl51d522_qwu-ka4sA:-NOl51d522_qwu-ka4sB:b-urb66d) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L467)):*
> ```cpp
> 
> /**
>  Compute the signed distance from a collision geometry to the halfspace { p_GQ |
> ```

Does this class have a non-penetrating assumption in play?
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 470 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl3R9ZAULdlD57srd7:-NOl3R9ZAULdlD57srd8:bxoknbp) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L470)):*
> ```cpp
>  a_Gᵀ*p_GQ+b_G >= 0}, where a_G is in the collision geometry frame.
>  */
> class DistanceToHalfspaceReifier : public ShapeReifier {
> ```

BTW Repeat note here.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 497 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl4KvK5axW4aOK8O44:-NOl4KvK5axW4aOK8O45:b-g8puax) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L497)):*
> ```cpp
>     p_GV.row(1) *= box.depth() / 2;
>     p_GV.row(2) *= box.height() / 2;
>     *distance = ((a_G_.transpose() * p_GV).minCoeff() + b_G_) / (a_G_.norm());
> ```

BTW This is the first time I realized that you didn't assume "The normal vector" is not unit length.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 554 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2gulCbQt6Gv2W6D-:-NOl2gulCbQt6Gv2W6D0:b-1tvtdc) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L554)):*
> ```cpp
> 
> CIrisGeometryType CIrisCollisionGeometry::type() const {
>   CIrisGeometryTypeReifier reifier(geometry_);
> ```

BTW If you take the simplification of `CIrisGeometryTypeReifier`, this function likewise simplifies.



_Suggestion:_
```cpp
CIrisGeometryType CIrisCollisionGeometry::type() const {
  return CIrisGeometryTypeReifier(*geometry_).type();
}
```

jwnimmer-tri:
This PR exceeds 1500 lines.  It requires >1 platform reviewer consensus.

https://drake.mit.edu/developers.html#review-process

```
jwnimmer@call-cps:~/jwnimmer-tri/drake/geometry/optimization/dev$ wc -l c_iris_collision_geometry*.* test/c_iris_collision_geometry_test.cc test/c_iris_test_utilities.*
  603 c_iris_collision_geometry.cc
  127 c_iris_collision_geometry.h
  695 test/c_iris_collision_geometry_test.cc
  233 test/c_iris_test_utilities.cc
  223 test/c_iris_test_utilities.h
 1881 total
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 22 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl6eOu1ZXx-7wm27T7:-NOlJ8d_5qARxb1iBjXx:b-e8b0xu) (no related file):*

This PR exceeds 1500 lines.  It requires >1 platform reviewer consensus.

https://drake.mit.edu/developers.html#review-process

```
jwnimmer@call-cps:~/jwnimmer-tri/drake/geometry/optimization/dev$ wc -l c_iris_collision_geometry*.* test/c_iris_collision_geometry_test.cc test/c_iris_test_utilities.*
  603 c_iris_collision_geometry.cc
  127 c_iris_collision_geometry.h
  695 test/c_iris_collision_geometry_test.cc
  233 test/c_iris_test_utilities.cc
  223 test/c_iris_test_utilities.h
 1881 total
```
</blockquote></details>

I am sorry to introduce this giant PR. I am happy to break this down to smaller PRs. For example I can first submit a PR that only includes `c_iris_test_utilities.h/cc`; and then a second PR that introduces the other files, which would bring the second PR to be less than 1500 LOC. Would that be OK?

If we want to stick to the single-PR, @jwnimmer-tri is it OK if you could do the platform review as well? Thanks a lot!

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 22 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl6eOu1ZXx-7wm27T7:-NOlLpSGCY7JvNTPnMOl:b-jwdvu8) (no related file):*

I am sorry to introduce this giant PR. I am happy to break this down to smaller PRs. For example I can first submit a PR that only includes `c_iris_test_utilities.h/cc`; and then a second PR that introduces the other files, which would bring the second PR to be less than 1500 LOC. Would that be OK?

If we want to stick to the single-PR, @jwnimmer-tri is it OK if you could do the platform review as well? Thanks a lot!
</blockquote></details>

My main aim is to ensure we take note of the process.

In the common case of coding outside of `dev`, we typically end up with multiple platform reviewers over time as the feature grows organically, a few hundreds of lines at a time.  When using `dev` and then moving out later, we won't get multiple reviewers by default unless we specifically aim for it.  Part of the reason for the review limit is to ensure that enough people look at our code.

I'll defer to @SeanCurtis-TRI for a suggestion on how to proceed here, i.e., whether it's worth more people and how to deploy them.

I'll also note in particular that most of `c_iris_test_utilities.cc` is a pile of modelling data that's inexplicably typed in as C++ code instead of a URDF file (or SDFormat).  Using model files for model data might be a good way to cut down on the lines of code, and helpful in its own right.  If the test rig were somehow parameterized maybe it would deserve to be in code, but calling the MbP API to construct a kinematic robot is awkward for everyone involved.

___

SeanCurtis-TRI (commented):
First pass done.

+@sherm1 for aid in platform review. The tests in `c_iris_collision_geometry_test.cc` would *most* benefit from the second set of eyes (700 lines by itself). But all the rest of the PR is fair game for whatever catches your eye.

> Reviewed 11 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOp1T9x7Pn-XoH3redr:bgbyqy8)* status: 30 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),sherm1(platform), missing label for release notes (waiting on @hongkai-dai, @jwnimmer-tri, and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl6eOu1ZXx-7wm27T7:-NOl9bQjAxe4zpMnZWOf:bjcpdi8) (no related file):*

My main aim is to ensure we take note of the process.

In the common case of coding outside of `dev`, we typically end up with multiple platform reviewers over time as the feature grows organically, a few hundreds of lines at a time.  When using `dev` and then moving out later, we won't get multiple reviewers by default unless we specifically aim for it.  Part of the reason for the review limit is to ensure that enough people look at our code.

I'll defer to @SeanCurtis-TRI for a suggestion on how to proceed here, i.e., whether it's worth more people and how to deploy them.

I'll also note in particular that most of `c_iris_test_utilities.cc` is a pile of modelling data that's inexplicably typed in as C++ code instead of a URDF file (or SDFormat).  Using model files for model data might be a good way to cut down on the lines of code, and helpful in its own right.  If the test rig were somehow parameterized maybe it would deserve to be in code, but calling the MbP API to construct a kinematic robot is awkward for everyone involved.
</blockquote></details>

The size is artificially inflated because 233 lines were treated as deleted and copied instead of moved.

___
___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 3 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOokUd26BzD0GKr0uva:-NOokUd26BzD0GKr0uvb:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L3)):*
> ```objc
> #pragma once
> 
> #include <array>
> ```

nit: unused
___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 5 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOokX6C30GnGdhYg9mj:-NOokX6C30GnGdhYg9mk:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L5)):*
> ```objc
> #include <array>
> #include <memory>
> #include <optional>
> ```

nit: unused
___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 6 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOokYid6VP-w_648BQ3:-NOokYid6VP-w_648BQ4:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L6)):*
> ```objc
> #include <memory>
> #include <optional>
> #include <unordered_map>
> ```

nit: unused
___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 7 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOok_er2ZxJzrLupI7P:-NOok_er2ZxJzrLupI7Q:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L7)):*
> ```objc
> #include <optional>
> #include <unordered_map>
> #include <unordered_set>
> ```

nit: unused
___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 20 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOoo_0t9pf_ZxxnJoJM:-NOoo_0t9pf_ZxxnJoJN:b-l96x53) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L20)):*
> ```objc
> namespace optimization {
> 
> // Create a toy robot with different collision geometries.
> ```

BTW Just looking here, it's not immediately clear what aspects of the toy robot are significant.

In the implementation, you've emphasized multiple times that the dynamics don't matter (so we have arbitrary contact materials and mass properties). Dimensions of geometry? How about the registration of the geometries to their bodies? Or the frames on which the joints operate?

Giving more of an overview of the important properties of the model constructed here would be nice.
___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOolo-_0crx-CFCyz1_:-NOolo-_0crx-CFCyz1a:bxoaqz8) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L21)):*
> ```cpp
>   systems::DiagramBuilder<double> builder;
>   plant_ = builder.AddSystem<multibody::MultibodyPlant<double>>(0.);
>   scene_graph_ = builder.AddSystem<geometry::SceneGraph<double>>();
> ```

nit: This can be nicely simplified:

```c++
  std::tie(plant_, scene_graph_) =
      multibody::AddMultibodyPlantSceneGraph(&builder, 0);
```



_Code quote:_
```cpp
  plant_ = builder.AddSystem<multibody::MultibodyPlant<double>>(0.);
  scene_graph_ = builder.AddSystem<geometry::SceneGraph<double>>();
  plant_->RegisterAsSourceForSceneGraph(scene_graph_);

  builder.Connect(scene_graph_->get_query_output_port(),
                  plant_->get_geometry_query_input_port());
  builder.Connect(
      plant_->get_geometry_poses_output_port(),
      scene_graph_->get_source_pose_port(plant_->get_source_id().value()));
```
___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOomo4_BwDo2PnQ9YzF:-NOomo4_BwDo2PnQ9YzG:b-ditkcr) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L50)):*
> ```cpp
>   const multibody::SpatialInertia<double> spatial_inertia(
>       1, Eigen::Vector3d::Zero(),
>       multibody::UnitInertia<double>(0.01, 0.01, 0.01, 0, 0, 0));
> ```

FYI Three values are sufficient; if you omit the off-diagonal values, they default to zero.

Another option would be to use one of the new APIs:

```c++
  const auto spatial_inertia =
      multibody::SpatialInertia<double>::SolidSphereWithDensity(1.0, 1.0);
```
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOoqby5DiMvYCyxPeJe:-NOoqby5DiMvYCyxPeJf:bv8qkmt) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_collision_geometry_test.cc#L53)):*
> ```cpp
>         diagram_context_{diagram_->CreateDefaultContext()} {
>     SetupPlane(rational_forward_kin_.s(), &a_, &b_);
>     plant_context_ = &(
> ```

FYI I think the preferred spelling is more along the lines of:

```c++
    plant_context_ =
        &plant_->GetMyMutableContextFromRoot(diagram_context_.get());
```

You don't have to bring `diagram_` into it at all.
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 296 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOp-ObM8JF-WdJkKKA4:-NOp-ObM8JF-WdJkKKA5:bp8b9fh) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_collision_geometry_test.cc#L296)):*
> ```cpp
>   // Test CIrisCollisionGeometry constructed from a sphere object.
>   const auto& model_inspector = scene_graph_->model_inspector();
>   CIrisCollisionGeometry sphere(&model_inspector.GetShape(body2_sphere_),
> ```

nit: These tests would be easier to consume if the similar parts were consistently similar. In this particular case, you don't declare the `geometry_body` variable.
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 377 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOp0TwD6Mx7Gw5Z_NFL:-NOp0TwD6Mx7Gw5Z_NFM:b5se9ff) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_collision_geometry_test.cc#L377)):*
> ```cpp
>   const symbolic::Expression b_expr = b_.EvaluatePartial(env).ToExpression();
> 
>   // Negative side
> ```

nit (meta) Generally make sure your comments end with punctuation. Lots and lots of instances with missing period.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 29 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),sherm1(platform), missing label for release notes (waiting on @hongkai-dai and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl6eOu1ZXx-7wm27T7:-NOp32l88mh81gRCuPN5:bpurv0k) (no related file):*

The size is artificially inflated because 233 lines were treated as deleted and copied instead of moved.
</blockquote></details>

I posted the line count atop this thread.  In a "coming out of dev", **moved doesn't matter** since the moved-from copy was never reviewed in the first place.

In any case, the second assigned reviewer sufficiently resolves this.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 4 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),sherm1(platform), missing label for release notes (waiting on @SeanCurtis-TRI and @sherm1)

___
*[`geometry/optimization/test/c_iris_test_utilities.h` line 20 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOoo_0t9pf_ZxxnJoJM:-NOpnuP2353LSYray4mB:b-56754c) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.h#L20)):*

BTW Just looking here, it's not immediately clear what aspects of the toy robot are significant.

In the implementation, you've emphasized multiple times that the dynamics don't matter (so we have arbitrary contact materials and mass properties). Dimensions of geometry? How about the registration of the geometries to their bodies? Or the frames on which the joints operate?

Giving more of an overview of the important properties of the model constructed here would be nice.
</blockquote></details>

Done. Good call. I added the documentation.
___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 1 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl5s97CtMg15ATMzmz:-NOpqK0y9L0LEk7A-cWd:b-g7yyzf) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L1)):*

BTW Per our conversation, let's see if we can't get git to understand that this file moved.
</blockquote></details>

I moved the function implementation in the header of `dev/c_iris_test_utilities.h` to `dev/c_iris_test_utilities.cc`. Let's see if that helps.
___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOolo-_0crx-CFCyz1_:-NOpoBCU3IbE46SCIRgw:bq6bi7h) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L21)):*

nit: This can be nicely simplified:

```c++
  std::tie(plant_, scene_graph_) =
      multibody::AddMultibodyPlantSceneGraph(&builder, 0);
```
</blockquote></details>

Done. Thanks a lot!
___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOomo4_BwDo2PnQ9YzF:-NOpoKsf-5qHH-IzwymX:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L50)):*

FYI Three values are sufficient; if you omit the off-diagonal values, they default to zero.

Another option would be to use one of the new APIs:

```c++
  const auto spatial_inertia =
      multibody::SpatialInertia<double>::SolidSphereWithDensity(1.0, 1.0);
```
</blockquote></details>

Done.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 41 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkgZwS788wy80qWAEq:-NOpEYhY26ibv4ALQtCW:bf40ro7) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L41)):*

Two things:

1. Does it *have* to be an aliased shape? Can't it simply be a locally owned copy?
2. If it does need to be aliased, the documentation should be clear that it *is* aliased and should stay alive for the duration of `this` instance.
</blockquote></details>

I think whenever we use this `CIrisCollisionGeometry` class, the original shape is always alive, hence we can always use alias to reduce the memory copy. I added the documentation to mention about the lifetime of this instance.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 54 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkgu48D5SJKCi9uhA9:-NOpHXtr4SjvHnNIz5Gp:bxknqk3) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L54)):*

nit missing namespace
</blockquote></details>

According to f2f discussion, we don't need to change the namespace here.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkh5Uh99AleNnqmbEc:-NOpHkHM-2A_ne92k8Gb:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L59)):*

BTW You'd do your C++ reader a favor by putting the entire mathematical statement on a singe line.

```c++
  /** Impose the constraint that the geometry is on a given side of the plane
   {x | aᵀx+b=0}.
  */
```
</blockquote></details>

Done. Good call.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl4Shs0e4V4eCe32f9:-NOpHuAB6Z-ChdgexZIx:bwp5kc1) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L81)):*

nit: I assumed "the normal vector" was unit length. Probably worth noting that it isn't required. (Possibly partnered with a `@pre` that it's length be greater than zero?)
</blockquote></details>

Done. Good call.

In our formulation `a` is actually a vector of rational functions, not a constant. The user doesn't need to guarantee `a != 0`. If they provide a 0-value rational function, then our mathematical program will catch that and report the program is infeasible.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOktXvS2pJY39dc-jsS:-NOpJCb90A3k21yDuleH:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L94)):*

nit: It seems that the rational functions representing the half space constraint are *added* to `rationals`. This suggests that they are the only thing in the vector upon completion. The documentation should be clear about that distinction.

(We probably need `@pre rationals != nullptr.`.
</blockquote></details>

Done.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 114 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkid4G7AjZVKdMOeMQ:-NOpJvBsAxWI_i8y2_17:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.h#L114)):*

nit: requires a moustache (here and on `body_index_`).
</blockquote></details>

Done.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 23 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkpGXsBdP4DOELdU66:-NOpS0QPBvHhLDEBVwj1:bt4ox15) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L23)):*

nit This constructor doesn't get you much. It kills the ability to do `{}` initialization (including designated initializers). It also masks the aliasing (i.e., passing in `const&` and storing `const*`.

You also need `{}` on all of the pointer-valued members. Here's what it would look like.

```c++
struct ReifyData {
  // Temporary container for the quantities necessary to process a shape.
  const Vector3<symbolic::Polynomial>* a{};
  const symbolic::Polynomial* b{};
  const multibody::RationalForwardKinematics::Pose<symbolic::Polynomial>*
      X_AB_multilinear{};
  const multibody::RationalForwardKinematics* rational_forward_kin{};
  const PlaneSide plane_side;
  const GeometryId geometry_id;
  const VectorX<symbolic::Variable>* y_slack{};
  std::vector<symbolic::RationalFunction>* rationals{};
};
```
</blockquote></details>

Done. Thanks for the suggestion!
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkt0_N0eBOUz-YDM-N:-NOpS3WK27ai7Kyxh3ik:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L85)):*

nit: It'd be nicer if the casting to `ReifyData` happened in the reifier and the properly typed data were were passed here. The goal would be to isolate the footprint of `ReifyData` and reduce the amount of `void*` sprinkled around the code.
</blockquote></details>

Done.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkyTdaEWO4O20-AfiB:-NOpSSGW6hlhgH-_Da8_:b-877mfn) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L100)):*

nit: This would be better as `emplace_back`.
</blockquote></details>

Done.

Sorry I don't quite understand how this works. I thought that `std::vector<T>.emplace_back()` pass in the constructor arguments of `T`. Here when I call `emplace_back()`, I didn't pass in the constructor argument for `RationalFunction` (which would be `emplace_back(numerator_polynomial, denominator_polynomial)`), but instead I have `emplace_back(ComputePointOnPlaneSideRational(...))`, where that `ComputePointOnPlaneSideRational()` function returns a `RationalFunction` object. Why it still works?
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 369 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2T7ZCGYS38-dOSUE:-NOpUXHT7EyC21UBV9q9:brpnvxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L369)):*

BTW Here's one that would benefit from simplification:

```c++
class CIrisGeometryTypeReifier : public ShapeReifier {
 public:
  explicit CIrisGeometryTypeReifier(const geometry::Shape& shape) {
    shape.Reify(this);
  }

  CIrisGeometryType type() const { return type_; }

 private:
  using ShapeReifier::ImplementGeometry;
  void ImplementGeometry(const Box&, void*) {
    type_ = CIrisGeometryType::kPolytope;
  }

  void ImplementGeometry(const Convex&, void*) {
    type_ = CIrisGeometryType::kPolytope;
  }

  void ImplementGeometry(const Sphere&, void*) {
    type_ = CIrisGeometryType::kSphere;
  }

  void ImplementGeometry(const Capsule&, void*) {
    type_ = CIrisGeometryType::kCapsule;
  }

  void ImplementGeometry(const Cylinder&, void*) {
    type_ = CIrisGeometryType::kCylinder;
  }

  CIrisGeometryType type_;
};
```

The updated call site is flagged down below.


</blockquote></details>

Done. The code is indeed more clear. Thanks a lot for the suggestion.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 410 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2uvc3cqXOjOazzbx:-NOpYgg74rSiuDWejgKA:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L410)):*

BTW This gets the same benefit. Instead of storing the shape, have a member `int count_{};`. And `ProcessData` simply becomes:

```c++
int count() const { return count_; }
```

And calls become:

```c++
const int num_rationals = NumRationalsReififer(*geometry_).count();
```
</blockquote></details>

Done.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 467 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl51d522_qwu-ka4sA:-NOpYlbN1wVfdBqesUf6:brc69d8) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L467)):*

Does this class have a non-penetrating assumption in play?
</blockquote></details>

No it doesn't. If it penetrates, it just returns the penetration depth. I added it in the documentation in the header file.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 497 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl4KvK5axW4aOK8O44:-NOpZvLACPSnYijiZ_3e:bvj3fzr) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L497)):*

BTW This is the first time I realized that you didn't assume "The normal vector" is not unit length.
</blockquote></details>

Done. I added a comment in the documentation in the header.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 554 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl2gulCbQt6Gv2W6D-:-NOp_JMCDuO-4xRRBoyW:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L554)):*

BTW If you take the simplification of `CIrisGeometryTypeReifier`, this function likewise simplifies.
</blockquote></details>

Done.
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOoqby5DiMvYCyxPeJe:-NOpa6tB1UJGcNk0kFVJ:bnn9j0u) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_collision_geometry_test.cc#L53)):*

FYI I think the preferred spelling is more along the lines of:

```c++
    plant_context_ =
        &plant_->GetMyMutableContextFromRoot(diagram_context_.get());
```

You don't have to bring `diagram_` into it at all.
</blockquote></details>

Done. That is neat.
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 296 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOp-ObM8JF-WdJkKKA4:-NOpmVed52t62YWT8hRV:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_collision_geometry_test.cc#L296)):*

nit: These tests would be easier to consume if the similar parts were consistently similar. In this particular case, you don't declare the `geometry_body` variable.
</blockquote></details>

Done. Good call.

sherm1 (commented):
Partial platform :lgtm:. I looked mostly at the tests and test utilities. A few nits here and there but I love these very extensive tests!

> Reviewed 3 of 15 files at r1, 5 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOq7J-Q8eEgylrsw4Qm:bpnsq5e)* status: 13 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @hongkai-dai and @SeanCurtis-TRI)

___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 18 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOq6Ywo5pl8B1iEsYuJ:-NOq6Ywo5pl8B1iEsYuK:b-q5ajr0) ([raw file](https://github.com/RobotLocomotion/drake/blob/2a8df4cf21d8898c62aeb053282ae21706e7d190/geometry/optimization/test/c_iris_test_utilities.cc#L18)):*
> ```cpp
> namespace geometry {
> namespace optimization {
> CIrisToyRobotTest::CIrisToyRobotTest() {
> ```

nit: insert a blank line here
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOp8r_eCjHvHJsKFbDE:-NOp8r_eCjHvHJsKFbDF:b-y33uro) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L49)):*
> ```cpp
>   const GeometryId geometry_id;
>   const VectorX<symbolic::Variable>* y_slack;
>   std::vector<symbolic::RationalFunction>* rationals;
> ```

nit: pointers need to be default-initialized with `{}` or nullptr
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 466 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpEXvZAK9syx1_JYRT:-NOpEXvZAK9syx1_JYRU:b-235xal) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L466)):*
> ```cpp
> };
> 
> /**
> ```

nit: shouldn't have Doxygen comment here in the .cc file, just use `/*`
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpBJXn-sHuq8FgOh_3:-NOpBJXn-sHuq8FgOh_4:b28iwur) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L570)):*
> ```cpp
>                            const systems::Context<double>& plant_context) {
>   // Transforms the halfspace from expressed frame (E) to the collision_geometry
>   // geometry frame (G). First compute the pose of expressed frame (E) to the
> ```

nit: you aren't just re-expressing the halfspace here from E to G, you're also re-measuring it from G's origin. You could just say "transform the halfspace from frame E to frame G" or "Halfspace is measured & expressed in frame E; transform it to frame G."
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 584 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpB1UDDuFyxibB7CND:-NOpB1UDDuFyxibB7CNE:b5ej10g) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L584)):*
> ```cpp
>   Eigen::Vector3d a_G;
>   double b_G;
>   a_G = X_GE.rotation() * a;
> ```

nit: combine for nice RAII



_Suggestion:_
```cpp
  Eigen::Vector3d a_G = X_GE.rotation() * a;
  double b_G = b - a_G.dot(X_GE.translation());
```
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOq1bPZAl2NBQPoCFHc:-NOq1bPZAl2NBQPoCFHd:b-2gaps5) ([raw file](https://github.com/RobotLocomotion/drake/blob/2a8df4cf21d8898c62aeb053282ae21706e7d190/geometry/optimization/test/c_iris_collision_geometry_test.cc#L21)):*
> ```cpp
> namespace drake {
> namespace geometry {
> namespace optimization {
> ```

BTW we typically add an anonymous `namespace {}` around test code to avoid potential name conflicts
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 23 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpvmc18UX0jDlps4Vq:-NOpvmc18UX0jDlps4Vr:b-umv53b) ([raw file](https://github.com/RobotLocomotion/drake/blob/2a8df4cf21d8898c62aeb053282ae21706e7d190/geometry/optimization/test/c_iris_collision_geometry_test.cc#L23)):*
> ```cpp
> namespace optimization {
> 
> void SetupPlane(const Eigen::Ref<const VectorX<symbolic::Variable>>& s,
> ```

minor: please add a short comment explaining what this does and the meaning of s, a, b parameters.
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 56 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpwLrH7XhCix0eGgZd:-NOpwLrI3ojDH9XCc1ak:bfgmbnv) ([raw file](https://github.com/RobotLocomotion/drake/blob/2a8df4cf21d8898c62aeb053282ae21706e7d190/geometry/optimization/test/c_iris_collision_geometry_test.cc#L56)):*
> ```cpp
>         &(plant_->GetMyMutableContextFromRoot(diagram_context_.get()));
>     scene_graph_context_ = &(diagram_->GetMutableSubsystemContext(
>         *scene_graph_, diagram_context_.get()));
> ```

nit: this should also use GetMyMutableContextFromRoot()
___
*[`geometry/optimization/test/c_iris_collision_geometry_test.cc` line 259 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOq3giE-f6d0G8SqNZg:-NOq3giE-f6d0G8SqNZh:b-6xoskq) ([raw file](https://github.com/RobotLocomotion/drake/blob/2a8df4cf21d8898c62aeb053282ae21706e7d190/geometry/optimization/test/c_iris_collision_geometry_test.cc#L259)):*
> ```cpp
>       plant_->get_body(expressed_body).body_frame(), &p_AV);
> 
>   // negative side.
> ```

nit: capitalize for consistency with "Positive side" below

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @SeanCurtis-TRI and @sherm1)

___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOpBJXn-sHuq8FgOh_3:-NOqIVj59777qPJ3oBD_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L570)):*

nit: you aren't just re-expressing the halfspace here from E to G, you're also re-measuring it from G's origin. You could just say "transform the halfspace from frame E to frame G" or "Halfspace is measured & expressed in frame E; transform it to frame G."
</blockquote></details>

Done.

jwnimmer-tri:
@drake-jenkins-bot retest this please

(I'd like to be sure that some of my recent commits didn't break this PR.)

sherm1 (commented):
> Reviewed 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM:

> Reviewed 4 of 7 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOuJihtBJP9VJoDegOy:bmxilld)* status: 5 unresolved discussions, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/test/c_iris_test_utilities.cc` line 1 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl5s97CtMg15ATMzmz:-NOuJMuCD84i8iGMZudS:bnbleh9) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/test/c_iris_test_utilities.cc#L1)):*

I moved the function implementation in the header of `dev/c_iris_test_utilities.h` to `dev/c_iris_test_utilities.cc`. Let's see if that helps.
</blockquote></details>

Meh. A battle not worth fighting further.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 46 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuFVJG8x2mugYS-pzW:-NOuFVJG8x2mugYS-pzX:b-goo6dg) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.h#L46)):*
> ```objc
>    @param X_BG The pose of the geometry (G) in the attached body frame (B).
>    */
>   CIrisCollisionGeometry(const geometry::Shape* geometry,
> ```

nit: Here's the usage of `geometry::` that primed my mind to think that `GeometryId` was missing. Here, the `id` parameter on this same function, and the member `id_`.  Three more in the .cc file.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 124 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuGEEz0mQgm7LQrU3p:-NOuGEEz0mQgm7LQrU3q:bcatnah) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.h#L124)):*
> ```objc
> 
> /** Computes the signed distance from `collision_geometry` to the half space ℋ,
>  where ℋ  = {x | aᵀx+b >= 0} if plane_side=PlaneSide::kPositive, and
> ```

nit: weird. You've added the space between `=` and `{`, but you've also added an extra space between `ℋ` and `=`. 
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkyTdaEWO4O20-AfiB:-NOuGbJm95lQA1aW4Qw9:b4qvum9) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L100)):*

Done.

Sorry I don't quite understand how this works. I thought that `std::vector<T>.emplace_back()` pass in the constructor arguments of `T`. Here when I call `emplace_back()`, I didn't pass in the constructor argument for `RationalFunction` (which would be `emplace_back(numerator_polynomial, denominator_polynomial)`), but instead I have `emplace_back(ComputePointOnPlaneSideRational(...))`, where that `ComputePointOnPlaneSideRational()` function returns a `RationalFunction` object. Why it still works?
</blockquote></details>

I know that this can be confusing. In this case, `ComputePointOnPlaneSideRational()` returns by value. But that value is in a *temporary* variable (r-value). Therefore, you have a *move* constructor of `RationalFunction` which can be invoked by `emplace_back()`. That's why using `emplace_back` works.

Here's where things get weird.

Calling `push_back` accomplishes the same thing -- there is a spelling of `push_back` that *also* takes an r-value and does move construction. So, in terms of code generated, the change from `push_back` to `emplace_back` is essentially non-existent. However, you're now communicating to the reader of the code that you expect construction in place and if the signature of `ComputePointOnPlaneSideRational` were to change in a way such that construction in place were no longer possible, the compiler would yell with `emplace_back` but remain silent with `push_back`. So, it's a slightly clearer indication of intent.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 467 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOl51d522_qwu-ka4sA:-NOuISYYDdHI652Kks8C:bu8tu7c) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L467)):*

No it doesn't. If it penetrates, it just returns the penetration depth. I added it in the documentation in the header file.
</blockquote></details>

I just realized my error. I saw `minCoeff` and didn't think in terms of sign, merely magnitude.  :-P  Closest to zero and not closest to negative infinity. I really have no excuse for that.  
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 357 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuHcRj2bzvqi2sKMwZ:-NOuHcRj2bzvqi2sKMw_:br1mjg7) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.cc#L357)):*
> ```cpp
> class CIrisGeometryTypeReifier : public ShapeReifier {
>  public:
>   explicit CIrisGeometryTypeReifier(const geometry::Shape* shape) {
> ```

nit: This should be a *reference* not a *pointer*. `nullptr` is not a valid input value so we should just omit the possibility.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 390 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuIEbL1rj2w08nM6KG:-NOuIEbL1rj2w08nM6KH:b-gfd6sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.cc#L390)):*
> ```cpp
> class NumRationalsReifier : public ShapeReifier {
>  public:
>   explicit NumRationalsReifier(const Shape* shape) { shape->Reify(this); }
> ```

nit swap `&` for `*`.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 439 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuIHQb6Cmtxi1LHKk4:-NOuIHQb6Cmtxi1LHKk5:b-gfd6sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.cc#L439)):*
> ```cpp
> class DistanceToHalfspaceReifier : public ShapeReifier {
>  public:
>   explicit DistanceToHalfspaceReifier(const Shape* shape, Eigen::Vector3d a_G,
> ```

nit swap `&` for `*`.

jwnimmer-tri:
*[`geometry/optimization/c_iris_collision_geometry.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkyTdaEWO4O20-AfiB:-NOuKx7Z6ZhZyKCeUEQB:b-qfkpt4) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L100)):*

I know that this can be confusing. In this case, `ComputePointOnPlaneSideRational()` returns by value. But that value is in a *temporary* variable (r-value). Therefore, you have a *move* constructor of `RationalFunction` which can be invoked by `emplace_back()`. That's why using `emplace_back` works.

Here's where things get weird.

Calling `push_back` accomplishes the same thing -- there is a spelling of `push_back` that *also* takes an r-value and does move construction. So, in terms of code generated, the change from `push_back` to `emplace_back` is essentially non-existent. However, you're now communicating to the reader of the code that you expect construction in place and if the signature of `ComputePointOnPlaneSideRational` were to change in a way such that construction in place were no longer possible, the compiler would yell with `emplace_back` but remain silent with `push_back`. So, it's a slightly clearer indication of intent.
</blockquote></details>

Are you sure the compiler would yell with emplace_back?  It seems to me like an lvalue reference passed to emplace_back would just copy-construct.

For an argument that's already the correct type, push_back and emplace_back should be equivalent, and I don't think it's at all atypical or unreadable to use push_back.  In fact -- I think push_back communicate to me that "I already constructed the thing, just append it" while emplace_back communicates "please construct the thing".

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 5 unresolved discussions, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOkyTdaEWO4O20-AfiB:-NOuMrIsBkrZ39-MJqdT:bghld9s) ([raw file](https://github.com/RobotLocomotion/drake/blob/6715c361476a31f407b91ff401919b9ba05ccc17/geometry/optimization/c_iris_collision_geometry.cc#L100)):*

Are you sure the compiler would yell with emplace_back?  It seems to me like an lvalue reference passed to emplace_back would just copy-construct.

For an argument that's already the correct type, push_back and emplace_back should be equivalent, and I don't think it's at all atypical or unreadable to use push_back.  In fact -- I think push_back communicate to me that "I already constructed the thing, just append it" while emplace_back communicates "please construct the thing".
</blockquote></details>

You're right. My justification was a bridge too far.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 2 unresolved discussions, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/c_iris_collision_geometry.h` line 46 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuFVJG8x2mugYS-pzW:-NOuPG8Z0yt3ahUagqF7:bkcpdg9) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.h#L46)):*

nit: Here's the usage of `geometry::` that primed my mind to think that `GeometryId` was missing. Here, the `id` parameter on this same function, and the member `id_`.  Three more in the .cc file.
</blockquote></details>

Got it, I will remove the `geometry::` namespace.
___
*[`geometry/optimization/c_iris_collision_geometry.h` line 124 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuGEEz0mQgm7LQrU3p:-NOuPxjC881p0ZVHwqkn:b-9ti4l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.h#L124)):*

nit: weird. You've added the space between `=` and `{`, but you've also added an extra space between `ℋ` and `=`. 
</blockquote></details>

Ah that is the rendering of ℋ on my vim editor. It is very tilted on my vim editor, which almost eats up the empty space immediately after ℋ, 


hence I thought I need another empty space. I have removed the extra space.
___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 357 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuHcRj2bzvqi2sKMwZ:-NOuRaqxCukMudgk0_Jl:b-rjx77w) ([raw file](https://github.com/RobotLocomotion/drake/blob/4f7587ab315a3b8d0ae5d82b0a8a45e093783741/geometry/optimization/c_iris_collision_geometry.cc#L357)):*

nit: This should be a *reference* not a *pointer*. `nullptr` is not a valid input value so we should just omit the possibility.
</blockquote></details>

Done. I changed the input arguments to all Reifier to a reference.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: 1 unresolved discussion, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 139 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuY3a4-Zm-AX-6JPo4:-NOuY3a4-Zm-AX-6JPo5:b-6mbzdr) ([raw file](https://github.com/RobotLocomotion/drake/blob/50c33056dff716852ccc0615658d2634327acd30/geometry/optimization/c_iris_collision_geometry.cc#L139)):*
> ```cpp
> class OnPlaneSideReifier : public ShapeReifier {
>  public:
>   OnPlaneSideReifier(const Shape& geometry, math::RigidTransformd X_BG,
> ```

nit: This is the exception. Because you *alias* `geometry`, you should pass it as a pointer (by styleguide convention).

Your other reifiers don't capture the shape -- they exercise it in the constructor and then ignore it.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/c_iris_collision_geometry.cc` line 139 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-NOuY3a4-Zm-AX-6JPo4:-NOuu-Zt1O58AoKKqKtU:b-ieo89z) ([raw file](https://github.com/RobotLocomotion/drake/blob/50c33056dff716852ccc0615658d2634327acd30/geometry/optimization/c_iris_collision_geometry.cc#L139)):*

nit: This is the exception. Because you *alias* `geometry`, you should pass it as a pointer (by styleguide convention).

Your other reifiers don't capture the shape -- they exercise it in the constructor and then ignore it.
</blockquote></details>

Done. Thanks for the explanation.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824)* status: missing label for release notes (waiting on @hongkai-dai)

SeanCurtis-TRI (commented):
+(release notes: feature) Unless there's a good reason to mask this from the release notes. @hongkai-dai feel free to remove the tag if you think it shouldn't be announced yet.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18824#-:-NOv03LE7uewrs3gG4B8:b-ukl1g4)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),sherm1(platform) (waiting on @hongkai-dai)

Issue #18828, opened 2023-02-20
Enable OpenMP in monthly releases


jwnimmer-tri:
Towards #14431 and #14858, we should enable OpenMP in our packaged binaries.

jwnimmer-tri:
I have a wheel fix here: https://github.com/jwnimmer-tri/drake/commits/omp-packaging

For the packaging builds, I'm not loving that we encode the configuration into the job name.  I think we should probably switch to `--config=packaging` in the CI scripts, and then let Drake rcfiles dictate what that means for the build.

PR #18838, opened 2023-02-21
[tools] Keep bazel analysis phase intact

jwnimmer-tri:
As it turns out, varying 'test' vs 'build' for options in a bazelrc causes the analysis phase to be re-run as the user alternates between 'bazel build' and 'bazel test'.  Since all test options are valid for build as well, we should just list everything in build.

This removes "Options have changed, discarding analysis cache" output from bazel in the typical case.

(See also `anzu#4460` for the ditto change.)

Towards #18828.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18838)

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -15,37 +15,37 @@ build --strip=never
 build --strict_system_includes
 
 # Default test options.
-test --test_output=errors
-test --test_summary=terse
+build --test_output=errors
+build --test_summary=terse
 
 # By default, disable execution of tests that require proprietary software.
 # Individual targets that use proprietary software are responsible for ensuring
 # they can also build without it, typically by using a select statement.
 # config_setting rules for proprietary software are provided in //tools.
-test --test_tag_filters=-gurobi,-mosek,-snopt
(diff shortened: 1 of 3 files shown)


jwnimmer-tri:
+@rpoyner-tri for both reviews, please.

\CC @EricCousineau-TRI FYI

rpoyner-tri (commented):
:lgtm: do we need a maintainer note somewhere besides the commit log message?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18838#-:-NOot90iBOUdLUjH7eEf:b9u8hyv)* status: :shipit: complete! all discussions resolved, LGTM from assignee rpoyner-tri(platform) (waiting on @jwnimmer-tri)

jwnimmer-tri:
My hope is that people's nature to cargo cult what's already here will be sufficient to keep the invariant intact.

jwnimmer-tri:
One thing we discovered later here:

If I a developer had `build --test_tag_filters=...` in a home rcfile, we were previously ignoring it by default (because our test options took precedence).  As of this change, we _do_ obey the user's filters.

PR #18843, opened 2023-02-21
[tools] Add all models to model_visualizer's runfiles

### LINKED ISSUE #18827: `bazel run //tools:model_visualizer` should data-depend on all models
We should have `//tools:model_visualizer` have a `data = [...]` dependency on every non-test model file in Drake.  I hit this myself last week trying to preview our PR2 model, and did Sean trying to visualize the Jaco.

N.B. The pydrake target should not have those data.  Only the tools shortcut should have it. This is a "for Drake developers only" feature.

I think the way I would factor that is to define in the project top-level BUILD.bazel

```
filegroup(
    name = "all_prod_models",
    data = [
        "//examples/pr2:models",
        # etc etc
    ],
    visibility = ["//:__subpackages__"],
)
```

and then use that in the model visualizer.

See [slack thread here](https://app.slack.com/client/T0JNB2DS4/C43KX47A9/thread/C43KX47A9-1676584736.707409).

jwnimmer-tri:
Closes #18827.

Note that this is only relevant for source builds.  Install users of model_visualizer still only have access to the same set of installed models as before.

+@SeanCurtis-TRI for both reviews, please.

\CC @trowell-tri FYI

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18843)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -35,6 +35,47 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+# A manually-curated collection of most model files in Drake, so that we can
+# easily provide access to them for tools like //tools:model_visualizer.
+filegroup(
+    name = "all_models",
+    data = [
+        "//bindings/pydrake/multibody:models",
+        "//examples/acrobot:models",
+        "//examples/atlas:models",
+        "//examples/compass_gait:models",
+        "//examples/hardware_sim:demo_data",
+        "//examples/hydroelastic/ball_plate:floor_files",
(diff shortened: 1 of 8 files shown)


SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18843#-:-NOprcat4L4N2Cm6i84k:bmxilld)* status: :shipit: complete! all discussions resolved, LGTM from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

PR #18863, opened 2023-02-24
[C-IRIS] move CspaceFreePolytope constructor out of dev folder.

hongkai-dai:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -87,7 +87,10 @@ drake_cc_library(
 
 drake_cc_library(
     name = "cspace_free_polytope",
-    srcs = ["//geometry/optimization/dev:cspace_free_polytope.cc"],
+    srcs = [
+        "cspace_free_polytope.cc",
+        "//geometry/optimization/dev:cspace_free_polytope.cc",
+    ],
     hdrs = ["cspace_free_polytope.h"],
     deps = [
         "//common/symbolic:monomial_util",
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -1,0 +1,371 @@
+#include "drake/geometry/optimization/cspace_free_polytope.h"
+
+#include "drake/multibody/rational/rational_forward_kinematics.h"
+#include "drake/multibody/rational/rational_forward_kinematics_internal.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+namespace {
+// Returns true if all the bodies on the kinematics chain from `start` to `end`
+// are welded together (namely all the mobilizers in between are welded).
+[[nodiscard]] bool ChainIsWelded(const multibody::MultibodyPlant<double>& plant,
+                                 multibody::BodyIndex start,
+                                 multibody::BodyIndex end) {
+  const std::vector<multibody::internal::MobilizerIndex> mobilizers =
+      multibody::internal::FindMobilizersOnPath(plant, start, end);
+  if (mobilizers.size() == 0) {
+    return true;
+  }
+  const multibody::internal::MultibodyTree<double>& tree =
+      multibody::internal::GetInternalTree(plant);
+  for (const auto& mobilizer : mobilizers) {
+    if (tree.get_mobilizer(mobilizer).num_positions() != 0) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// For all the geometries `link1_geometries` on link1, and `link2_geometries` on
+// link2, return them as pairs if they are not filtered.
+[[nodiscard]] std::vector<
+    std::pair<const CIrisCollisionGeometry*, const CIrisCollisionGeometry*>>
+GetLinkCollisionPairs(
+    const multibody::MultibodyPlant<double>& plant,
+    const SceneGraph<double>& scene_graph, multibody::BodyIndex link1,
+    multibody::BodyIndex link2,
+    const std::vector<std::unique_ptr<CIrisCollisionGeometry>>&
+        link1_geometries,
+    const std::vector<std::unique_ptr<CIrisCollisionGeometry>>&
+        link2_geometries) {
+  std::vector<
+      std::pair<const CIrisCollisionGeometry*, const CIrisCollisionGeometry*>>
+      ret;
+  if (ChainIsWelded(plant, link1, link2)) {
+    // Two links cannot collide if there are only welded joints between them as
+    // they cannot move relative to each other. Return an empty vector.
+    return ret;
+  }
+  const auto& model_inspector = scene_graph.model_inspector();
+  for (const auto& geometry1 : link1_geometries) {
+    for (const auto& geometry2 : link2_geometries) {
+      if (!model_inspector.CollisionFiltered(geometry1->id(),
+                                             geometry2->id())) {
+        ret.emplace_back(geometry1.get(), geometry2.get());
+      }
+    }
+  }
+  return ret;
+}
+
+struct BodyPair {
+  BodyPair(multibody::BodyIndex m_body1, multibody::BodyIndex m_body2)
+      : body1{m_body1}, body2{m_body2} {}
+  bool operator==(const BodyPair& other) const {
+    return body1 == other.body1 && body2 == other.body2;
+  }
+
+  multibody::BodyIndex body1;
+  multibody::BodyIndex body2;
+};
+
+struct BodyPairHash {
+  size_t operator()(const BodyPair& p) const { return p.body1 * 100 + p.body2; }
+};
+
+// Returns the number of y_slack variables in `rational`.
+// Not all y_slack necessarily appear in `rational`.
+int GetNumYInRational(const symbolic::RationalFunction& rational,
+                      const Vector3<symbolic::Variable>& y_slack) {
+  int count = 0;
+  for (int i = 0; i < 3; ++i) {
+    if (rational.numerator().indeterminates().find(y_slack(i)) !=
+        rational.numerator().indeterminates().end()) {
+      ++count;
+    }
+  }
+  return count;
+}
+
+/**
+ The monomials in the numerator of body point position has the form ∏ᵢ pow(tᵢ,
+ dᵢ) * ∏ⱼ pow(xⱼ, cⱼ), where tᵢ is the configuration variable tᵢ = tan(Δθᵢ/2)
+ for the revolute joint, dᵢ = 0, 1, or 2, on the kinematic chain between the
+ pair of bodies; xⱼ is the prismatic joint displacement on the same kinematic
+ chain, cⱼ = 0 or 1. When we use SeparatingPlaneOrder = kAffine, the monomial
+ basis would include all the monomials in the form ∏ᵢ pow(sᵢ, nᵢ) with nᵢ = 0
+ or 1, s ∈ {tᵢ, xⱼ}. Let's denote this monomial basis as m(s).
+
+ When we have non-polytopic collision geometries, we will also impose a matrix
+ of rational functions being positive semidefinite. This requires a matrix-sos
+ constraint with slack variable y. We will need the monomial basis in the form
+ yᵢ*m(s) with i=0,1,2.
+
+ @param[in/out] map_body_to_monomial_basis_array stores all the monomials basis
+ already computed.
+ @param[out] monomial_basis_array The monomial basis array for this pair of
+ body, monomial_basis_array = [m(s), y₀*m(s), y₁*m(s), y₂*m(s)]
+ */
+void FindMonomialBasisArray(
+    const multibody::RationalForwardKinematics& rational_forward_kin,
+    const Vector3<symbolic::Variable>& y_slack,
+    const SortedPair<multibody::BodyIndex>& body_pair,
+    std::unordered_map<SortedPair<multibody::BodyIndex>,
+                       std::array<VectorX<symbolic::Monomial>, 4>>*
+        map_body_to_monomial_basis_array,
+    std::array<VectorX<symbolic::Monomial>, 4>* monomial_basis_array) {
+  auto body_pair_it = map_body_to_monomial_basis_array->find(body_pair);
+  if (body_pair_it == map_body_to_monomial_basis_array->end()) {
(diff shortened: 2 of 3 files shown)


hongkai-dai (commented):
+@ggould-tri for next Monday's platform review please, thanks!

I move the constructor CspaceFreePolytope out of the dev folder. It has been feature reviewed. Could you platform-review this file? Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-:-NP-uQz19RIWzSptM1hm:b3hjcd1)* status: LGTM missing from assignee ggould-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @ggould-tri)

ggould-tri (commented):
:lgtm: ; a few notes below.
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-:-NPIgE_-AODb4fPL86bg:b-9qn6up)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIssnc4g8_Fy873dZX:-NPIssnc4g8_Fy873dZY:bmcnmw7) (no related file):*
Is there a follow-up PR to move the unit test?  Moving the implementation without its corresponding test coverage is a bit surprising.

___
___
*[`geometry/optimization/cspace_free_polytope.cc` line 11 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIhR6ICaw1VhbhLLc2:-NPIhR6ICaw1VhbhLLc3:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L11)):*
> ```cpp
> namespace {
> // Returns true if all the bodies on the kinematics chain from `start` to `end`
> // are welded together (namely all the mobilizer in between are welded).
> ```

typo



_Suggestion:_
```cpp
mobilizers
```
___
*[`geometry/optimization/cspace_free_polytope.cc` line 12 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIhbE1Bj7SAXtMWHAq:-NPIhbE1Bj7SAXtMWHAr:b-62n41v) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L12)):*
> ```cpp
> // Returns true if all the bodies on the kinematics chain from `start` to `end`
> // are welded together (namely all the mobilizer in between are welded).
> [[nodiscard]] bool ChainIsWeld(const multibody::MultibodyPlant<double>& plant,
> ```

typo: grammar -- "is welded" seems more fluent.



_Suggestion:_
```cpp
 ChainIsWelded(c
```
___
*[`geometry/optimization/cspace_free_polytope.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIkJEl7Vr2ETc15Bgq:-NPIkJEmCrVHskFDWya4:b3jby3i) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L78)):*
> ```cpp
> 
> // Returns the number of y_slack variables in `rational`.
> int GetNumY(const symbolic::RationalFunction& rational,
> ```

minor:  The name `GetNumY` is not at all descriptive of what the function does; it's particularly confusing given the explicit `Vector3`, and the common just restates the content of the function without explaining it.
Also this function is only ever used for its zero/nonzero return, so it could easily return bool (and return instead of incrementing count, for a small performance improvement.
Finally, the actual logic of this function mostly already exists in `Variables` so it could probably be more clearly written as a one-liner.
As a sketch, you could for instance write something like
```
bool FunctionContainsVar(const symbolic::RationalFunction& rational, const Vector3<symbolic::Variable>& vars) {
  return !rational.numerator().indeterminates().intersect(symbolic::Variables(vars)).empty();
}
```
___
*[`geometry/optimization/cspace_free_polytope.cc` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIoWNA33Xy9DHjBROg:-NPIoWNA33Xy9DHjBROh:ba1xe5e) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L118)):*
> ```cpp
>     std::array<VectorX<symbolic::Monomial>, 4>* monomial_basis_array) {
>   auto it = map_body_to_monomial_basis_array->find(body_pair);
>   if (it == map_body_to_monomial_basis_array->end()) {
> ```

minor:  For readability, consider using `count` or `contains` for membership checking.

Alternatively, if you are relying on the iterator position for the hinting performance of the later `emplace_hint` call, `it` is not a sufficiently descriptive variable name.



_Suggestion:_
```cpp
  if (!map_body_to_monomial_basis_array->contains(body_pair)()) {
```
___
*[`geometry/optimization/cspace_free_polytope.cc` line 158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIq31oBERMkhnbyNNk:-NPIq31oBERMkhnbyNNl:b536l7e) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L158)):*
> ```cpp
>     }
>     it = map_body_to_monomial_basis_array->emplace_hint(it, body_pair,
>                                                         *monomial_basis_array);
> ```

minor:
(1) assigning to `it` seems unnecessary, as it is being discarded.
(2) I don't see any basis for hinting the end of the list for the emplace; if we expect the body pairs to be visited in order, that should be mentioned in the function comment, but nothing at the call site suggests that this is the case.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @ggould-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIssnc4g8_Fy873dZX:-NPJ1mc_3m_0V0GcqjBV:bm0x9qj) (no related file):*

Is there a follow-up PR to move the unit test?  Moving the implementation without its corresponding test coverage is a bit surprising.
</blockquote></details>

Yes, there is a follow-up PR to move the unit test. Doing it all in one PR can make the PR size too large. (I tried to put the test code in one PR, but easily exceed 800 LOC).

___
___
*[`geometry/optimization/cspace_free_polytope.cc` line 11 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIhR6ICaw1VhbhLLc2:-NPJ8Yp660yyOve1fEHe:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L11)):*

typo
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 12 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIhbE1Bj7SAXtMWHAq:-NPJ8gHiBD5boDYQx64c:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L12)):*

typo: grammar -- "is welded" seems more fluent.
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIkJEl7Vr2ETc15Bgq:-NPJ9LoS7lu0Df2Oohv_:b-fbdu7r) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L78)):*
> Also this function is only ever used for its zero/nonzero return, so it could easily return bool (and return instead of incrementing count, for a small performance improvement.

It is only used in its zero/nonzero return in the code that gets moved out from `dev` folder. In the code that still lives in the `dev` folder, it uses the actual integer value of the return, not just the boolean value.

I re-named the function and add some documentation. Is it better now?
___
*[`geometry/optimization/cspace_free_polytope.cc` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIoWNA33Xy9DHjBROg:-NPJAXim0CjP0fccuU-i:bk4w3g8) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L118)):*

minor:  For readability, consider using `count` or `contains` for membership checking.

Alternatively, if you are relying on the iterator position for the hinting performance of the later `emplace_hint` call, `it` is not a sufficiently descriptive variable name.
</blockquote></details>

Done. I changed the variable name to `body_pair_it`.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIq31oBERMkhnbyNNk:-NPJB5sx6YY_E0Ge92K4:b66h97p) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L158)):*

minor:
(1) assigning to `it` seems unnecessary, as it is being discarded.
(2) I don't see any basis for hinting the end of the list for the emplace; if we expect the body pairs to be visited in order, that should be mentioned in the function comment, but nothing at the call site suggests that this is the case.
</blockquote></details>

Done. I removed the return.

I use emplace_hint with the end of list because we have tried to find this key in the map but hasn't found it yet. In this case I suppose we should emplace it at the end?

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863)* status: needs at least two assigned reviewers (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIkJEl7Vr2ETc15Bgq:-NPJIV7o-PG2S_UKKAEB:bb71kmx) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L78)):*

> Also this function is only ever used for its zero/nonzero return, so it could easily return bool (and return instead of incrementing count, for a small performance improvement.

It is only used in its zero/nonzero return in the code that gets moved out from `dev` folder. In the code that still lives in the `dev` folder, it uses the actual integer value of the return, not just the boolean value.

I re-named the function and add some documentation. Is it better now?
</blockquote></details>

Okay
___
*[`geometry/optimization/cspace_free_polytope.cc` line 158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIq31oBERMkhnbyNNk:-NPJI_-g80aMhg07pzM2:b6kqquy) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L158)):*

Done. I removed the return.

I use emplace_hint with the end of list because we have tried to find this key in the map but hasn't found it yet. In this case I suppose we should emplace it at the end?
</blockquote></details>

Hinting is useful when the iterator is at the correct insertion point; in this case because of the check on line 118 you know that `it` was at the end, so that's what you were hinting.  Because `unordered_map` is implemented as a sorted list, this will be more efficient than `emplace` if you know that the item you are adding is in fact the greatest element, and (slightly) slower otherwise.
So the case elsewhere in this PR where it hints `end()` and is in fact always inserting a new greatest element (because it traverses the pairs in order) is correct.
In this case, I can't see a good argument for why the new element will be last.
Regardless it doesn't matter very much -- the savings and costs are small for a small map.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863)* status: needs at least two assigned reviewers (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIq31oBERMkhnbyNNk:-NPJMsFCDPmrVmtTpwlG:b33ljkt) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L158)):*

Sorry I think I missed something, I thought that `unordered_map` is <b> not </b> sorted, it doesn't even require the key in the map to be comparable, so how to implement `unordered_map` as a sorted list?
</blockquote></details>

Oh, you're right, it's bucketed -- I had it confused.  But I think that's actually worse -- hinting `end()` will never be right in that case, I think?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18863)* status: needs at least two assigned reviewers (waiting on @ggould-tri)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18863#-NPIq31oBERMkhnbyNNk:-NPJNFzJ9RulZ64b05cK:b-4u28rf) ([raw file](https://github.com/RobotLocomotion/drake/blob/f5919838e43d89087910259e270bd1bb2a37b7e4/geometry/optimization/cspace_free_polytope.cc#L158)):*

Oh, you're right, it's bucketed -- I had it confused.  But I think that's actually worse -- hinting `end()` will never be right in that case, I think?
</blockquote></details>

I see, after reading the answer in this stackoverflow question https://stackoverflow.com/questions/22800405/when-do-you-use-stdunordered-mapemplace-hint I think I understand it better.

I also modified the other emplace_hint to emplace in this file.

PR #18861, opened 2023-02-23
Add a benchmark on trace(A*X) where X contains symbolic variables.

hongkai-dai:
This is extremely slow for A and X being of size 200 * 200, takes 77 seconds on my Puget.

As reported from the slack discussion https://drakedevelopers.slack.com/archives/C3L92BM2Q/p1677172164932209

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18861)

PR #18869, opened 2023-02-24
[symbolic] Speed up Add and Mul cell ctors

jwnimmer-tri:
Note that `expression_cell.h` is internal use only, so this API change will not disturb any users.

This is roughly a 50% reduction in BenchmarkMatrixInnerProduct/200, and a 10% reduction in all three BenchmarkSosProgram.

Also move two tiny Variable getters to the header file. It's a little difficult to show this in a profile, but I think we'd all agree that this change is blindingly obvious and should have been like this from the beginning.

Relates to #18861.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18869)

### DIFF
--- a/common/symbolic/expression/expression.cc
+++ b/common/symbolic/expression/expression.cc
@@ -276,7 +276,7 @@ void Expression::AddImpl(const Expression& rhs) {
     }
   }
   // Extract an expression from factory
-  lhs = add_factory.GetExpression();
+  lhs = std::move(add_factory).GetExpression();
 }
 
 Expression& Expression::operator++() {
@@ -484,7 +484,7 @@ void Expression::MulImpl(const Expression& rhs) {
       mul_factory.AddExpression(rhs);
     }
   }
-  lhs = mul_factory.GetExpression();
+  lhs = std::move(mul_factory).GetExpression();
 }
 
 void Expression::DivImpl(const Expression& rhs) {
@@ -941,7 +941,7 @@ Expression Exp(const vector<Expression>& terms, const MultiIndex& alpha) {
   for (size_t i = 0; i < terms.size(); ++i) {
     factory.AddExpression(pow(terms[i], alpha[i]));
   }
-  return factory.GetExpression();
+  return std::move(factory).GetExpression();
 }
 
 // Computes ∑_{|α| = order} ∂fᵅ(a) / α! * (x - a)ᵅ.
@@ -978,7 +978,7 @@ Expression TaylorExpand(const Expression& f, const Environment& a,
   for (int i = 1; i <= order; ++i) {
     DoTaylorExpand(f, a, terms, i, num_vars, &factory);
   }
-  return factory.GetExpression();
+  return std::move(factory).GetExpression();
 }
 
 namespace {
--- a/common/symbolic/expression/expression_cell.cc
+++ b/common/symbolic/expression/expression_cell.cc
@@ -131,7 +131,7 @@ Expression ExpandMultiplication(const Expression& e1, const Expression& e2) {
     for (const pair<const Expression, double>& p : m1) {
       fac.AddExpression(ExpandMultiplication(p.second, p.first, e2));
     }
-    return fac.GetExpression();
+    return std::move(fac).GetExpression();
   }
   if (is_addition(e2)) {
     //   e1 * (c0 + c1 * e_{2,1} + ... + c_n * e_{2, n})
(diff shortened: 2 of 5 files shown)


hongkai-dai (commented):
:lgtm: 

Sorry for the delay, I had to read about r-value reference to review the PR.

My understanding of using r-value reference in this PR, is that after calling `ExpressionAddFactory::GetExpreesion()` method, the original `ExpressionAddFactory` is not useful anymore. We only need its `expr_to_coeff_map_`, moved into the returned `ExpressionAdd`, the original `expr_to_coeff_map_`, the original ExpressionAddFactory, and its `ExpressionAddFactory::expr_to_coeff_map_` can be discarded. Hence it makes sense to use r-value reference to swap these maps, instead of doing the regular copy. Is that right?

Also, in the future, should I always consider to return an r-value reference, if I find that a function returns a reference, similar to what `Negate` function does here?

+@rpoyner-tri for Tuesday's platform review please, thanks!

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-:-NPK6rljAnhQTy-bq_-D:b-d539wy)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`common/symbolic/expression/expression_cell.h` line 291 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-NPK5Tyg2h-2qCxHpjng:-NPK5Tyg2h-2qCxHpjnh:b-diweu7) ([raw file](https://github.com/RobotLocomotion/drake/blob/2423e55d67694d48039fc5c6373994f73b8a8712/common/symbolic/expression/expression_cell.h#L291)):*
> ```objc
>    * @returns *this.
>    */
>   ExpressionAddFactory&& Negate() &&;
> ```

For my own educational purpose, does this mean that this `Negate` function can only be called on an r-value reference `ExpressionAddFactory`? Namely we should always use `std::move(my_expression_add_factory).Negate()` to change `my_expression_add_factory` to an r-value reference?

jwnimmer-tri (commented):
> My understanding of using r-value reference in this PR, ... Is that right?

Previously, we were copying the map when creating `Expression` result.  Since that copy is expensive, and nobody needs the old map afterwards anyway (we just destroy it), it makes sense to `move` the map over into the Expresion, instead of copying it.

We _could_ actually move the map without using rvalue references, but by marking the functions as `&&` it helps remind us at the call site that the factory will no longer be useful, so we shouldn't keep adding any more things into it.

> Also, in the future, should I always consider to return an r-value reference, if I find that a function returns a reference, similar to what Negate function does here?

This was just an oddball case.  We don't have a lot of functions that "modify myself in place and then return myself".  Here since it's all internal code anyway, it was easiest to add the `&&` instead of changing the function signatures.  In a more public API, we would instead change the function to be something like `void NegateInPlace();` with no special markers.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-:-NPKHjrc4xIi6AXxBXEQ:bdre87u)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @hongkai-dai and @rpoyner-tri)

___
*[`common/symbolic/expression/expression_cell.h` line 291 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-NPK5Tyg2h-2qCxHpjng:-NPKIaML24G0Q0_f7BVL:b-dm9ygn) ([raw file](https://github.com/RobotLocomotion/drake/blob/2423e55d67694d48039fc5c6373994f73b8a8712/common/symbolic/expression/expression_cell.h#L291)):*
> For my own educational purpose, does this mean that this Negate function can only be called on an r-value reference ExpressionAddFactory?

Yes

> Namely we should always use std::move(my_expression_add_factory).Negate() to change my_expression_add_factory to an r-value reference?

Yes.  Though note the factory is internal to the `expression` package, so the all of the relevant places are covered in this PR.

The compiler will tell us if we ever forget the `move()` in this situation, so it's less about a rule to memorize, and more just making the code build.

hongkai-dai (commented):
Got it, thanks for the explanation!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-:-NPKO2OK-OHiHBu7Ursr:bx40oul)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18869#-:-NPNVkCzCP3faFagFrFY:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),hongkai-dai (waiting on @jwnimmer-tri)

PR #18664, opened 2023-01-26
[planning] Reorganize planning artifacts in planning (1 of 3)

SeanCurtis-TRI:
Moves the *unbound* artifacts into `planning` with deprecated breadcrumbs.

 - manipulation/planner to multibody/inverse_kinematics
   - constraint_relaxing_ik
 - manipulation/planner to manipulation/util
    - robot_plan_interpolator

Moving includes updating all references and namespaces.

relates #18640

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18664)

PR #17169, opened 2022-05-12
[geometry] Move RenderEngineVtk into its own directory

jwnimmer-tri:
Leave only the params class and factory function as non-internal.

This scheme aligns with what we agreed on for `geometry/render_gltf_client`.

Towards #16502.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/17169)

PR #18877, opened 2023-02-27
[geometry] Make namespace of all three RenderEngine consistent

zachfang:
According to [a Slack discussion](https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1646937427862689) for the RenderEngine folder structure and their namespaces, this PR attempts to address the loose ends that do not currently obey the policy.

For all three RenderEngine implementations, change the namespace of the public interfaces to `geometry` and all the internal classes to `geometry/render_*/internal`.

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_render.cc
+++ b/bindings/pydrake/geometry/geometry_py_render.cc
@@ -336,7 +336,7 @@ void DoScalarIndependentDefinitions(py::module m) {
   }
 
   {
-    using Class = geometry::RenderEngineVtkParams;
+    using Class = RenderEngineVtkParams;
     constexpr auto& cls_doc = doc_geometry.RenderEngineVtkParams;
     py::class_<Class> cls(m, "RenderEngineVtkParams", cls_doc.doc);
     cls  // BR
@@ -346,12 +346,14 @@ void DoScalarIndependentDefinitions(py::module m) {
     DefCopyAndDeepCopy(&cls);
   }
 
-  m.def("MakeRenderEngineVtk", &geometry::MakeRenderEngineVtk,
-      py::arg("params"), doc_geometry.MakeRenderEngineVtk.doc);
+  m.def("MakeRenderEngineVtk", &MakeRenderEngineVtk, py::arg("params"),
+      doc_geometry.MakeRenderEngineVtk.doc);
 
   {
-    using Class = RenderEngineGlParams;
-    constexpr auto& cls_doc = doc.RenderEngineGlParams;
+    // TODO(zachfang): During the 2023-07-01 deprecation removals also
+    // remove the spurious `geometry::` qualifier on this typename.
+    using Class = geometry::RenderEngineGlParams;
+    constexpr auto& cls_doc = doc_geometry.RenderEngineGlParams;
     py::class_<Class> cls(m, "RenderEngineGlParams", cls_doc.doc);
     cls  // BR
         .def(ParamInit<Class>());
@@ -360,11 +362,14 @@ void DoScalarIndependentDefinitions(py::module m) {
     DefCopyAndDeepCopy(&cls);
   }
 
-  m.def("MakeRenderEngineGl", &MakeRenderEngineGl,
-      py::arg("params") = RenderEngineGlParams(), doc.MakeRenderEngineGl.doc);
+  // TODO(zachfang): During the 2023-07-01 deprecation removals also
+  // remove the spurious `geometry::` qualifier on this typename.
+  m.def("MakeRenderEngineGl", &geometry::MakeRenderEngineGl,
+      py::arg("params") = geometry::RenderEngineGlParams(),
+      doc_geometry.MakeRenderEngineGl.doc);
 
   {
-    using Class = geometry::RenderEngineGltfClientParams;
+    using Class = RenderEngineGltfClientParams;
     constexpr auto& cls_doc = doc_geometry.RenderEngineGltfClientParams;
     py::class_<Class> cls(m, "RenderEngineGltfClientParams", cls_doc.doc);
     cls  // BR
@@ -374,7 +379,7 @@ void DoScalarIndependentDefinitions(py::module m) {
     DefCopyAndDeepCopy(&cls);
   }
 
-  m.def("MakeRenderEngineGltfClient", &geometry::MakeRenderEngineGltfClient,
+  m.def("MakeRenderEngineGltfClient", &MakeRenderEngineGltfClient,
       py::arg("params") = RenderEngineGltfClientParams(),
       doc_geometry.MakeRenderEngineGltfClient.doc);
 
@@ -383,7 +388,9 @@ void DoScalarIndependentDefinitions(py::module m) {
 }  // namespace
 
 void DefineGeometryRender(py::module m) {
-  m.doc() = "Local bindings for `drake::geometry::render`";
+  m.doc() =
+      "Local bindings for render artifacts found in `drake::geometry` and "
+      "`drake::geometry::render`";
   DoScalarIndependentDefinitions(m);
 }
 
--- a/geometry/benchmarking/render_benchmark.cc
+++ b/geometry/benchmarking/render_benchmark.cc
@@ -39,7 +39,6 @@ using render::DepthRange;
 using render::DepthRenderCamera;
 using render::RenderCameraCore;
 using render::RenderEngine;
-using render::RenderEngineGlParams;
 using render::RenderEngineTester;
 using render::RenderLabel;
 using systems::sensors::ImageDepth32F;
--- a/geometry/render_gl/BUILD.bazel
+++ b/geometry/render_gl/BUILD.bazel
@@ -243,6 +243,20 @@ drake_cc_googletest(
     ],
 )
 
+# TODO(zachfang): Remove this along with deprecation on 2023-07-01.
+drake_cc_googletest(
+    name = "deprecation_test",
+    copts = [
+        "-Wno-cpp",
+        "-Wno-deprecated-declarations",
+    ],
+    deps = [
+        ":factory",
+        ":render_engine_gl_params",
+        "//common/test_utilities:expect_throws_message",
+    ],
+)
+
 drake_cc_googletest_ubuntu_only(
     name = "internal_shape_meshes_test",
     data = [
--- a/geometry/render_gl/factory.cc
+++ b/geometry/render_gl/factory.cc
@@ -6,16 +6,21 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
 
 // Definition of extern bool in factory.h. When we build against *this* .cc file
 // RenderEngineGl is always available.
 const bool kHasRenderEngineGl = true;
 
-std::unique_ptr<RenderEngine> MakeRenderEngineGl(RenderEngineGlParams params) {
-  return std::make_unique<internal::RenderEngineGl>(std::move(params));
+std::unique_ptr<render::RenderEngine> MakeRenderEngineGl(
+    RenderEngineGlParams params) {
+  return std::make_unique<render_gl::internal::RenderEngineGl>(
+      std::move(params));
 }
 
+namespace render {
+  DRAKE_DEPRECATED("2023-07-01", "Use the geometry namespace instead.")
+  const bool kHasRenderEngineGl = true;
 }  // namespace render
+
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/factory.h
+++ b/geometry/render_gl/factory.h
@@ -2,12 +2,12 @@
 
 #include <memory>
 
+#include "drake/common/drake_deprecated.h"
 #include "drake/geometry/render/render_engine.h"
 #include "drake/geometry/render_gl/render_engine_gl_params.h"
 
 namespace drake {
 namespace geometry {
-namespace render {
 
 /** Reports the availability of the RenderEngineGl implementation. */
 extern const bool kHasRenderEngineGl;
@@ -30,9 +30,18 @@ extern const bool kHasRenderEngineGl;
  SceneGraph, rendering in multiple threads may exhibit issues.
 
  @throws std::exception if kHasRenderEngineGl is false. */
-std::unique_ptr<RenderEngine> MakeRenderEngineGl(
+std::unique_ptr<render::RenderEngine> MakeRenderEngineGl(
     RenderEngineGlParams params = {});
 
+namespace render {
+
+DRAKE_DEPRECATED("2023-07-01", "Use the geometry namespace instead.")
+extern const bool kHasRenderEngineGl;
+
+DRAKE_DEPRECATED("2023-07-01", "Use the geometry namespace instead.")
+constexpr auto MakeRenderEngineGl = &geometry::MakeRenderEngineGl;
+
 }  // namespace render
+
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_buffer_dim.h
+++ b/geometry/render_gl/internal_buffer_dim.h
@@ -6,7 +6,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* Simple class for recording the dimensions of a render target. Serves as a
@@ -57,7 +57,7 @@ struct RenderTarget {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
 
@@ -65,6 +65,6 @@ namespace std {
 
 /* Provides std::hash<BufferDim>.  */
 template <>
-struct hash<drake::geometry::render::internal::BufferDim>
+struct hash<drake::geometry::render_gl::internal::BufferDim>
     : public drake::DefaultHash {};
 }  // namespace std
--- a/geometry/render_gl/internal_opengl_context.cc
+++ b/geometry/render_gl/internal_opengl_context.cc
@@ -19,7 +19,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 namespace {
@@ -326,6 +326,6 @@ GLint OpenGlContext::max_allowable_texture_size() {
 }
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_opengl_context.h
+++ b/geometry/render_gl/internal_opengl_context.h
@@ -6,7 +6,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* Handle OpenGL context initialization, clean-up, and generic OpenGL queries.
@@ -66,6 +66,6 @@ class OpenGlContext {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_opengl_geometry.h
+++ b/geometry/render_gl/internal_opengl_geometry.h
@@ -11,7 +11,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 // TODO(SeanCurtis-TRI): Consider moving this up to RenderEngine; it's useful
@@ -126,6 +126,6 @@ struct OpenGlInstance {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_render_engine_gl.cc
+++ b/geometry/render_gl/internal_render_engine_gl.cc
@@ -12,7 +12,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 using Eigen::Vector2d;
@@ -25,6 +25,11 @@ using std::string;
 using std::unique_ptr;
 using std::unordered_map;
 using std::vector;
+using render::ColorRenderCamera;
+using render::DepthRenderCamera;
+using render::RenderCameraCore;
+using render::RenderEngine;
+using render::RenderLabel;
 using systems::sensors::ColorD;
 using systems::sensors::ColorI;
 using systems::sensors::ImageDepth32F;
@@ -1132,6 +1137,6 @@ ShaderProgramData RenderEngineGl::GetShaderProgram(
 }
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_render_engine_gl.h
+++ b/geometry/render_gl/internal_render_engine_gl.h
@@ -23,11 +23,11 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /** See documentation of MakeRenderEngineGl().  */
-class RenderEngineGl final : public RenderEngine {
+class RenderEngineGl final : public render::RenderEngine {
  public:
   /** @name Does not allow public copy, move, or assignment  */
   //@{
@@ -53,7 +53,7 @@ class RenderEngineGl final : public RenderEngine {
 
   /** @name    Shape reification  */
   //@{
-  using RenderEngine::ImplementGeometry;
+  using render::RenderEngine::ImplementGeometry;
   void ImplementGeometry(const Box& box, void* user_data) final;
   void ImplementGeometry(const Capsule& capsule, void* user_data) final;
   void ImplementGeometry(const Convex& convex, void* user_data) final;
@@ -92,17 +92,17 @@ class RenderEngineGl final : public RenderEngine {
 
   // @see RenderEngine::DoRenderColorImage().
   void DoRenderColorImage(
-      const ColorRenderCamera& camera,
+      const render::ColorRenderCamera& camera,
       systems::sensors::ImageRgba8U* color_image_out) const final;
 
   // @see RenderEngine::DoRenderDepthImage().
   void DoRenderDepthImage(
-      const DepthRenderCamera& render_camera,
+      const render::DepthRenderCamera& render_camera,
       systems::sensors::ImageDepth32F* depth_image_out) const final;
 
   // @see RenderEngine::DoRenderLabelImage().
   void DoRenderLabelImage(
-      const ColorRenderCamera& camera,
+      const render::ColorRenderCamera& camera,
       systems::sensors::ImageLabel16I* label_image_out) const final;
 
   // Copy constructor used for cloning.
@@ -143,7 +143,7 @@ class RenderEngineGl final : public RenderEngine {
   // called if there is not already a cached render target for the camera's
   // reported image size (w, h) in render_targets_.
   static RenderTarget CreateRenderTarget(
-      const RenderCameraCore& camera, RenderType render_type);
+      const render::RenderCameraCore& camera, RenderType render_type);
 
   // Obtains the label image rendered from a specific object pose. This is
   // slower than it has to be because it does per-pixel processing on the CPU.
@@ -154,8 +154,8 @@ class RenderEngineGl final : public RenderEngine {
   // Acquires the render target for the given camera. "Acquiring" the render
   // target guarantees that the target will be ready for receiving OpenGL
   // draw commands.
-  RenderTarget GetRenderTarget(
-      const RenderCameraCore& camera, RenderType render_type) const;
+  RenderTarget GetRenderTarget(const render::RenderCameraCore& camera,
+                               RenderType render_type) const;
 
   // Creates an OpenGlGeometry from the mesh defined by the given `mesh_data`.
   static OpenGlGeometry CreateGlGeometry(
@@ -170,8 +170,8 @@ class RenderEngineGl final : public RenderEngine {
   //  - the window is made hidden (or remains hidden).
   // @pre RenderTarget's frame buffer has the same dimensions as reported by the
   // camera.
-  void SetWindowVisibility(const RenderCameraCore& camera, bool show_window,
-                           const RenderTarget& target) const;
+  void SetWindowVisibility(const render::RenderCameraCore& camera,
+                           bool show_window, const RenderTarget& target) const;
 
   // Adds a shader program to the set of candidate shaders for the given render
   // type.
@@ -271,6 +271,6 @@ class RenderEngineGl final : public RenderEngine {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_shader_program.cc
+++ b/geometry/render_gl/internal_shader_program.cc
@@ -10,7 +10,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 using Eigen::Vector3d;
@@ -148,6 +148,6 @@ void ShaderProgram::Unuse() const {
 }
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_shader_program.h
+++ b/geometry/render_gl/internal_shader_program.h
@@ -14,7 +14,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* TODO(SeanCurtis-TRI) This implies a compile-time shader definition. It may
@@ -117,7 +117,7 @@ class ShaderProgram {
    properties. This should *not* include model -> camera -> device transforms.
    they are handled elsewhere.  */
   virtual void SetDepthCameraParameters(
-      const DepthRenderCamera& /* camera */) const {}
+      const render::DepthRenderCamera& /* camera */) const {}
 
   /* Sets the direction of the directional light (if supported).
    @pre light_dir_C.norm() == 1.0.  */
@@ -205,6 +205,6 @@ class ShaderProgram {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_shader_program_data.h
+++ b/geometry/render_gl/internal_shader_program_data.h
@@ -9,7 +9,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* Type used to identify unique shader programs in RenderEngineGl. */
@@ -42,6 +42,6 @@ class ShaderProgramData {
 };
 }  // namespace internal
 
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_shape_meshes.cc
+++ b/geometry/render_gl/internal_shape_meshes.cc
@@ -19,7 +19,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 using Eigen::Vector2d;
@@ -1077,6 +1077,6 @@ MeshData MakeCapsule(int samples, double radius, double length) {
 }
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_shape_meshes.h
+++ b/geometry/render_gl/internal_shape_meshes.h
@@ -9,7 +9,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* The data representing a mesh. The triangle mesh is defined by `indices`. Row
@@ -170,6 +170,6 @@ MeshData MakeUnitBox();
 MeshData MakeCapsule(int samples, double radius, double length);
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_texture_library.cc
+++ b/geometry/render_gl/internal_texture_library.cc
@@ -16,7 +16,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 TextureLibrary::TextureLibrary(const OpenGlContext* context)
@@ -101,6 +101,6 @@ std::optional<GLuint> TextureLibrary::GetTextureId(
 }
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/internal_texture_library.h
+++ b/geometry/render_gl/internal_texture_library.h
@@ -9,7 +9,7 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
+namespace render_gl {
 namespace internal {
 
 /* Stores a set of OpenGl textures objects, keyed by their in-filesystem name.
@@ -41,6 +41,6 @@ class TextureLibrary {
 };
 
 }  // namespace internal
-}  // namespace render
+}  // namespace render_gl
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/no_factory.cc
+++ b/geometry/render_gl/no_factory.cc
@@ -3,18 +3,21 @@
 
 namespace drake {
 namespace geometry {
-namespace render {
 
 // Definition of extern bool in factory.h. When we build against *this* .cc file
 // RenderEngineGl is not available.
 const bool kHasRenderEngineGl = false;
 
-std::unique_ptr<RenderEngine> MakeRenderEngineGl(RenderEngineGlParams) {
+std::unique_ptr<render::RenderEngine> MakeRenderEngineGl(RenderEngineGlParams) {
   throw std::runtime_error(
       "RenderEngineGl was not compiled. You'll need to use a different render "
       "engine.");
 }
 
+namespace render {
+  DRAKE_DEPRECATED("2023-07-01", "Use the geometry namespace instead.")
+  const bool kHasRenderEngineGl = false;
 }  // namespace render
+
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/render_gl/render_engine_gl_params.h
+++ b/geometry/render_gl/render_engine_gl_params.h
@@ -1,12 +1,12 @@
 #pragma once
 
+#include "drake/common/drake_deprecated.h"
 #include "drake/common/name_value.h"
 #include "drake/geometry/render/render_label.h"
 #include "drake/geometry/rgba.h"
 
 namespace drake {
 namespace geometry {
-namespace render {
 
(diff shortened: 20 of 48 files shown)


zachfang:
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-release please

SeanCurtis-TRI (commented):
+@SeanCurtis-TRI for feature review

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NPblI7Z17ZLlGazvstf:b-7evfmk)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
The only issue of significance is that the *public* API needs to be deprecated. The good news is that since the bindings are already directly in the `pydrake.geometry` module, you don't have to jump through any hoops on that side. But you do need to leave a footprint behind in C++ for:

  - `drake::geometry::render::kHasRenderEngineGl`
  - `drake::geometry::render::MakeRenderEngineGl`
  - `drake::geometry::render::RenderEngineGlParams`

Everything else seems safely ensconced in an `internal` namespace, so you are free to move those cost free. 

You can see an example on how to do this from #18664. Let me know if you'd like to talk through the strategy/mechanisms.

> Reviewed 29 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NPbm14G6PepxNiFcd9F:b-fvdvcr)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPc8H3A-au1VLE93A92:b-e52pq8) (no related file):*
In C++ the render parameters and factories have moved from `drake::geometry::render` to simply `drake::geometry`. However, in python they're still in `drake.geometry.render`. Is there a plan to reconcile them?

___
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPboFlR01pKh5XIyRpH:-NPboFlR01pKh5XIyRpI:b-95n56c) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/geometry/render_gl/internal_render_engine_gl.h#L30)):*
> ```objc
> 
> /** See documentation of MakeRenderEngineGl().  */
> class RenderEngineGl final : public geometry::render::RenderEngine {
> ```

nit I'm surprised that you're using `geometry::render::` as the namespace qualifier. I would've assumed that just `render::` would be sufficient. This applies here and in multiple cases below. But you *don't* use it on `render::RenderEngine`.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbp4ZV91aH15Sdib-A:-NPbp4ZV91aH15Sdib-B:b2qdavr) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L350)):*
> ```cpp
> 
>   m.def("MakeRenderEngineVtk", &geometry::MakeRenderEngineVtk,
>       py::arg("params") = RenderEngineVtkParams(),
> ```

BTW Thanks for cleaning these bindings up.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 354 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbpFRCA3KKf2JXD3ys:-NPbpFRCA3KKf2JXD3yt:btq2nqw) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L354)):*
> ```cpp
> 
>   {
>     using Class = geometry::RenderEngineGlParams;
> ```

nit: `geometry::` shouldn't be necessary. Up above, you've already declared `using namespace drake::geometry`. It should likewise be unnecessary for the other render engines.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 388 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbpZ2h02vrMOgZL3Rh:-NPbpZ2h02vrMOgZL3Ri:bv6bkrd) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L388)):*
nit: This hasn't been true for a while now -- ever since render things came out of the render namespace. Perhaps it would be better as:

> Local bindings for render artifacts found in `drake::geometry` and `drake::geometry::render`.

zachfang (commented):
Thanks for the review! I will keep this PR as my filler task so the turnover rate might be slower.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NPebd5sAkWcXO7ws7r1:b-xll3w0)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPebXwF8uKDf8MffJ0V:bv228no) (no related file):*
Working. Good point!

> The good news is that since the bindings are already directly in the pydrake.geometry module, you don't have to jump through any hoops on that side. 

Don't we need to do a similar deprecation process on the Python side as well? The users currently import RenderEngineGl from `drake.geometry.render`, but it will be changed to `drake.geometry`.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPebsgb4qWY4f8asv5A:-NPebsgb4qWY4f8asv5B:bcn9mo7) (no related file):*
Working. Add deprecation pieces.
```
drake::geometry::render::kHasRenderEngineGl
drake::geometry::render::MakeRenderEngineGl
drake::geometry::render::RenderEngineGlParams
```

___
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPboFlR01pKh5XIyRpH:-NPeVD3gDB-mLXz31Vy0:b-mer42i) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/geometry/render_gl/internal_render_engine_gl.h#L30)):*

nit I'm surprised that you're using `geometry::render::` as the namespace qualifier. I would've assumed that just `render::` would be sufficient. This applies here and in multiple cases below. But you *don't* use it on `render::RenderEngine`.
</blockquote></details>

Done. This is probably me in the copy-pasta mania mode.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 354 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbpFRCA3KKf2JXD3ys:-NPebLJ0EGJmKi8gGpw-:bannsoz) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L354)):*

nit: `geometry::` shouldn't be necessary. Up above, you've already declared `using namespace drake::geometry`. It should likewise be unnecessary for the other render engines.
</blockquote></details>

Done. Modified the other two as well.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPsexyT3vPvQ0ECpRo6:bd7txwu) (no related file):*

Working. Good point!

> The good news is that since the bindings are already directly in the pydrake.geometry module, you don't have to jump through any hoops on that side. 

Don't we need to do a similar deprecation process on the Python side as well? The users currently import RenderEngineGl from `drake.geometry.render`, but it will be changed to `drake.geometry`.
</blockquote></details>

That's true for *all* rendering artifacts. All the factories, the render label, etc. So, even if this PR merges (in its current state), everything will still be in `pydrake.geometry.render`. So, I infer from your response that the eventual plan will be that `pydrake` structure will more closely match the C++ namespaces in this regard? I.e., somethings (`RenderEngien` and `RenderLabael`) will stay behind in `pydrake.geometry.render`, but the render engine factories and params will come out?

___

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPwjvOgBp6DSYW9dJmi:ba34ri4) (no related file):*

That's true for *all* rendering artifacts. All the factories, the render label, etc. So, even if this PR merges (in its current state), everything will still be in `pydrake.geometry.render`. So, I infer from your response that the eventual plan will be that `pydrake` structure will more closely match the C++ namespaces in this regard? I.e., somethings (`RenderEngien` and `RenderLabael`) will stay behind in `pydrake.geometry.render`, but the render engine factories and params will come out?
</blockquote></details>

Ah, now I understand your underlying question. Should this PR fix all the namespace issues for the *three* `RenderEngine`s in *both* C++ and Python?

If you think it makes sense to fit everything in one PR, I can go for that. I was only targeting `RenderEngineGl` while I noticed other issues with `RenderEnigineVtk` and `RenderEngineGltfClient` and planned to do that in the following PRs. I didn't factor in the deprecation part, so it seems logical to me now to do it all in one PR. 

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPwqX0a34pAZsc3c_wJ:b-s7kugf) (no related file):*

Ah, now I understand your underlying question. Should this PR fix all the namespace issues for the *three* `RenderEngine`s in *both* C++ and Python?

If you think it makes sense to fit everything in one PR, I can go for that. I was only targeting `RenderEngineGl` while I noticed other issues with `RenderEnigineVtk` and `RenderEngineGltfClient` and planned to do that in the following PRs. I didn't factor in the deprecation part, so it seems logical to me now to do it all in one PR. 


</blockquote></details>

Actually, I wasn't proposing anything, *per se*. I didn't know if we had a plan for python one way or the other. The policy isn't clear to me. For example, why is `RenderEngine` in `geometry::render` but `RenderEngeinVtk` in `geometry::`, etc. So, I didn't know if there was an intent for python modules to shift to match the C++ at all or if that was going to happen soon, or whatever.

You *could* match them up in this PR. However, we've already gone a while with the `RenderEngineVtk` already occupying two different locations between C++ and python for a while. As such, I can believe that it might thrash a bit of code to make that change. So, if you want to move all of the python instances in a follow up PR, that's fine with me.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPwrIrr9H_lAQNmjBdX:-NPwrIrr9H_lAQNmjBdY:brfmris) (no related file):*
Blocking for some Anzu exploration.

___

jwnimmer-tri:
Big picture:

My general posture is that drip-feeding deprecations to users is a minor kind of torture.  It's better if we can deprecate everything all at once, so they can do all of their boilerplate fixup work all at once.

That doesn't mean you can't move things in pieces -- it just means that the shims you add for the old names to keep working _are not marked deprecated_ until all of the changes have landed.  They are marked with "TODO deprecate me along with the rest of my cadre" instead.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NPwu-E77P5M0fUaSuC5:b8t4bl6) (no related file):*

Actually, I wasn't proposing anything, *per se*. I didn't know if we had a plan for python one way or the other. The policy isn't clear to me. For example, why is `RenderEngine` in `geometry::render` but `RenderEngeinVtk` in `geometry::`, etc. So, I didn't know if there was an intent for python modules to shift to match the C++ at all or if that was going to happen soon, or whatever.

You *could* match them up in this PR. However, we've already gone a while with the `RenderEngineVtk` already occupying two different locations between C++ and python for a while. As such, I can believe that it might thrash a bit of code to make that change. So, if you want to move all of the python instances in a follow up PR, that's fine with me.
</blockquote></details>

I should provide more context in the first place. We discussed the folder structure and the namespace hierarchy in a [Slack thread](https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1646937427862689) a while back, and my plan is to update things to obey the policy.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPebsgb4qWY4f8asv5A:-NQ69FsXCMqPJaNz4HgN:b-uotp05) (no related file):*

Working. Add deprecation pieces.
```
drake::geometry::render::kHasRenderEngineGl
drake::geometry::render::MakeRenderEngineGl
drake::geometry::render::RenderEngineGlParams
```
</blockquote></details>

Just to tie this in to Jeremy's comment above.

Providing the mechanism so that it's available in both old and new namespaces is sufficient for now. And the act of deprecating those (and the python bindings and whatever else might be included) can happen in a follow up PR.

___

zachfang:
@drake-jenkins-bot mac-x86-monterey-clang-bazel-experimental-release please
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-release please

zachfang (commented):
The PR is ready for another round of review. Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRWCSa865GEK_Q0bUcK:b8wfhis)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NRWC91eDw_8hZadBmbu:be73r0n) (no related file):*

I should provide more context in the first place. We discussed the folder structure and the namespace hierarchy in a [Slack thread](https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1646937427862689) a while back, and my plan is to update things to obey the policy.
</blockquote></details>

OK. As this PR already involves many files (although non-functional), I will have a follow-up PR for the Python part instead.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPebsgb4qWY4f8asv5A:-NRWBUHy-b9wPHFW_pCP:bkd8ic5) (no related file):*

Just to tie this in to Jeremy's comment above.

Providing the mechanism so that it's available in both old and new namespaces is sufficient for now. And the act of deprecating those (and the python bindings and whatever else might be included) can happen in a follow up PR.
</blockquote></details>

I think these are the pieces for the C++ deprecation. PTAL.

___

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/factory.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRWD8Ar8K9riGUHVrkN:-NRWD8Ar8K9riGUHVrkO:b-mzjh16) ([raw file](https://github.com/RobotLocomotion/drake/blob/716bf41649e3151bd49f19720ff9a14ee6e98eac/geometry/render_gl/factory.h#L38)):*
> ```objc
> namespace render {
> 
> DRAKE_DEPRECATED("2023-08-01", "Use the geometry namespace instead.")
> ```

Working. Fix all the deprecation dates. It's chosen somehow arbitrarily now.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 355 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRYS5D8DkciCkGrxoSS:-NRYS5D9AWqzNVltZxf6:bc7gw5r) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/bindings/pydrake/geometry/geometry_py_render.cc#L355)):*
> ```cpp
>   {
>     // TODO(zachfang): Remove `geometry` ns once the deprecation is removed.
>     using Class = geometry::RenderEngineGlParams;
> ```

FYI. The `geometry` namespace is required for RenderEngineGlParams and MakeRenderEngineGl due to the deprecation process. (The CI elevates a warning to an error.)

jwnimmer-tri (commented):
:lgtm: feature.

> Reviewed 25 of 29 files at r1, 1 of 2 files at r2, 15 of 17 files at r3, 3 of 4 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRddJp96D7E4Ev71FCX:b-n37teh)* status: 12 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NRdXuujDN1qkBLIe081:b-58zqxs) (no related file):*

OK. As this PR already involves many files (although non-functional), I will have a follow-up PR for the Python part instead.
</blockquote></details>

Might be worth capturing in an issue so we can discuss the victory condition. It should be that `pydrake.geometry.render` disappears entirely and _all_ of it moves into geometry, not just the engines.  Once there's a ticket for that, I can cross-link to the other related issues that explain the reasoning.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPebsgb4qWY4f8asv5A:-NRddki92NroNs7Hw37y:bb6ywmx) (no related file):*

I think these are the pieces for the C++ deprecation. PTAL.
</blockquote></details>

LGTM

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPwrIrr9H_lAQNmjBdX:-NRdYKMC55pildEDi4NU:b-i111tf) (no related file):*

Blocking for some Anzu exploration.
</blockquote></details>

I think there is a bug in this PR that's not caught by Drake CI testing, but shows up in Anzu CI:

```
what():  Invalid camera configuration; requested renderer_name = 'default' and renderer_class = 'RenderEngineGl'. The name is already used with a different type: drake::geometry::render_gl::internal::RenderEngineGl.
```

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRd_e7MCmtBUUxxsxo7:-NRd_e7MCmtBUUxxsxo8:b6sirhs) (no related file):*
Anzu spews a lot of deprecation warnings from this PR.  Prior to merging this PR, you need to either prepare and code-review the Anzu PR that resolves the warnings, or recruit someone on the Anzu team who will do so.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRd_fWkELP7ngnPKF86:-NRd_fWkELP7ngnPKF87:b-etstbq) (no related file):*
A lot of the now-internal classes have misleading Doxygen formatting of `///` (should be `/* */` or `//`).  Please open a ticket to clean that up later in a follow-up PR.  (It's OK if you'd like to assign to me.)

___
___
*[`geometry/render_gl/BUILD.bazel` line 88 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRddTvV3CgtZ_ce84h5:-NRddTvW3KOTVuUEWPM-:b9tcn2g) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/BUILD.bazel#L88)):*
> ```Starlark
> )
> 
> # TODO(zachfang): Remove this along with deprecation on 2023-08-01.
> ```

Is this stale? I don't see anything nearby here that should be changed or removed on the deprecation removal date.
___
*[`geometry/render_gl/BUILD.bazel` line 131 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRddchpD-JxXCdrt1Zz:-NRddchpD-JxXCdrt1_-:b9tcn2g) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/BUILD.bazel#L131)):*
> ```Starlark
> )
> 
> # TODO(zachfang): Remove this along with deprecation on 2023-08-01.
> ```

Is this stale? I don't see anything nearby here that should be changed or removed on the deprecation removal date.
___
*[`geometry/render_gl/factory.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRWD8Ar8K9riGUHVrkN:-NRdbgPQ1v_U9vhAPRXT:bh52zr5) ([raw file](https://github.com/RobotLocomotion/drake/blob/716bf41649e3151bd49f19720ff9a14ee6e98eac/geometry/render_gl/factory.h#L38)):*

Working. Fix all the deprecation dates. It's chosen somehow arbitrarily now.
</blockquote></details>

The dates seem fine?  You could try to squeak in with 2023-07-01 if you prefer but that only gives you another two days runway before this has to merge or else we're back to 2023-08-01.
___
*[`geometry/render_gl/factory.h` line 39 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdbvzx2ETwrypASLlx:-NRdbvzx2ETwrypASLly:b-ql33ot) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/factory.h#L39)):*
> ```objc
> 
> DRAKE_DEPRECATED("2023-08-01", "Use the geometry namespace instead.")
> constexpr auto kHasRenderEngineGl = &geometry::kHasRenderEngineGl;
> ```

I'm not sure why this is a pointer.  Shouldn't it be a bool?

It might also need to be either `extern` or `inline`.  Add a deprecation_test case to cover it so that we can be sure.



_Suggestion:_
```objc
const bool kHasRenderEngineGl = geometry::kHasRenderEngineGl;
```
___
*[`geometry/render_gl/render_engine_gl_params.h` line 36 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdbEAd2-ukKQxdPme6:-NRdbEAd2-ukKQxdPme7:b9w54ns) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/render_engine_gl_params.h#L36)):*
> ```objc
> 
> using RenderEngineGlParams
>     DRAKE_DEPRECATED("2023-08-01", "Use the geometry namespace instead.")
> ```

nit Missing `#include` for the DRAKE_DEPRECATED macro.
___
*[`geometry/render_gl/test/deprecation_test.cc` line 13 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdcya37G0h0ocbNTtI:-NRdcya49tEYraCP6aHF:bkmrqhh) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/test/deprecation_test.cc#L13)):*
> ```cpp
> namespace geometry {
> namespace render {
> namespace {
> ```

This is confusing.  We're already in `drake::geometry::render` but below we qualify `render::` extra again?

The purpose of this test is to be a stand-in for what a user would see.  As such, we should lose these namespace scopes entirely.  Users won't have those.



_Suggestion:_
```cpp
namespace {
```
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbp4ZV91aH15Sdib-A:-NRdVL4f7Nuon-L8SxJu:b1j17w6) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L350)):*

BTW Thanks for cleaning these bindings up.
</blockquote></details>

The C++ does _not_ have a default value for this argument.  So we shouldn't be adding one for Python either.  The C++ class has a default argument (which is silly, since it's probably never used in its defaulted form) but the factory function does not have a default.

In a separate PR, you could choose to add one for both C++ and Python with appropriate unit tests, but it shouldn't pollute the re-namespacing PR.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 354 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdWIMx1ZVs59Ekpflv:-NRdWIMx1ZVs59Ekpflw:byiboev) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/bindings/pydrake/geometry/geometry_py_render.cc#L354)):*
> ```cpp
> 
>   {
>     // TODO(zachfang): Remove `geometry` ns once the deprecation is removed.
> ```

nit Deprecation TODOs must always have the deprecation date magic string in them, or else they will be missed in the monthly grep.  Also don't abbreviate ("ns" means "nanosecond", not "namespace").

Ditto below.



_Suggestion:_
```cpp
// TODO(zachfang): During the 2023-08-01 deprepcation removals also
// remove the spurious `geometry::` qualifier on this typename.
```

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 7 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/factory.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRWD8Ar8K9riGUHVrkN:-NRem1VQ24IS7kqDp9Im:bsrw4k1) ([raw file](https://github.com/RobotLocomotion/drake/blob/716bf41649e3151bd49f19720ff9a14ee6e98eac/geometry/render_gl/factory.h#L38)):*

The dates seem fine?  You could try to squeak in with 2023-07-01 if you prefer but that only gives you another two days runway before this has to merge or else we're back to 2023-08-01.
</blockquote></details>

So seems like the deprecation timeline is normally three months. Is there a page for that? I checked our Drake website, e.g., the `For Developer` page, but didn't find that (maybe I missed it).
___
*[`geometry/render_gl/factory.h` line 39 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdbvzx2ETwrypASLlx:-NRemYqo3ZnzHLtHy9Zx:b-d4ibc3) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/factory.h#L39)):*

I'm not sure why this is a pointer.  Shouldn't it be a bool?

It might also need to be either `extern` or `inline`.  Add a deprecation_test case to cover it so that we can be sure.
</blockquote></details>

Done. It was a copy-pasta. Added this to the deprecation test. 
___
*[`geometry/render_gl/test/deprecation_test.cc` line 13 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdcya37G0h0ocbNTtI:-NRemqSK3a6UhlGs4m-U:bza3i8q) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/test/deprecation_test.cc#L13)):*

This is confusing.  We're already in `drake::geometry::render` but below we qualify `render::` extra again?

The purpose of this test is to be a stand-in for what a user would see.  As such, we should lose these namespace scopes entirely.  Users won't have those.
</blockquote></details>

Done. I borrowed the test mostly from #17169.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPbp4ZV91aH15Sdib-A:-NRen10_7LYyJslURTBD:beq7uod) ([raw file](https://github.com/RobotLocomotion/drake/blob/51528e67bf9dc36e2b659e48fea74786615c4ed8/bindings/pydrake/geometry/geometry_py_render.cc#L350)):*

The C++ does _not_ have a default value for this argument.  So we shouldn't be adding one for Python either.  The C++ class has a default argument (which is silly, since it's probably never used in its defaulted form) but the factory function does not have a default.

In a separate PR, you could choose to add one for both C++ and Python with appropriate unit tests, but it shouldn't pollute the re-namespacing PR.
</blockquote></details>

Done. Resumed to the original form.

jwnimmer-tri (commented):
+@sherm1 for platform review per schedule, please.

> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRewXNZAO3IyZY9oybV:bybe5xg)* status: 5 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @sherm1 and @zachfang)

___
*[`geometry/render_gl/factory.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRWD8Ar8K9riGUHVrkN:-NReuMN452k4IILn_neV:b-ejshj0) ([raw file](https://github.com/RobotLocomotion/drake/blob/716bf41649e3151bd49f19720ff9a14ee6e98eac/geometry/render_gl/factory.h#L38)):*
Yes, three months.

It's documented, though a bit hidden:

> Absent any other particular need, we prefer to use a deprecation period of three months by default, often rounded up to the next first of the month. So for code announced as deprecated on 2018-01-22 the removal_date would nominally be set to 2018-05-01.

-- https://drake.mit.edu/doxygen_cxx/drake__deprecated_8h.html#a847d827d084f28e4ae1788a0ae75be96
___
*[`geometry/render_gl/factory.h` line 39 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdbvzx2ETwrypASLlx:-NRev-JM4KtYpUdXSeD7:b-s8pg9m) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/factory.h#L39)):*

Done. It was a copy-pasta. Added this to the deprecation test. 
</blockquote></details>

I typed a similar example into godbolt and looked at the object code: https://godbolt.org/z/TKd8f87bf

It introduces a global constructor in all code that includes this file, because we now have a non-constexpr global variable with an initializer in a header file.

We'll need to change the deprecated one to `extern` as well (and provide a definition for it in the cc files):

The header will be like this:

```c++
namespace render {
DRAKE_DEPRECATED(...)
extern const bool kHasRenderEngineGl;
}
```

The cc files will replicate what's already there for the other (non-deprecated) definition, just in another namespace.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 4 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`geometry/render_gl/factory.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRWD8Ar8K9riGUHVrkN:-NRf9c2P5lVUrPcqer0R:b-7t4sr7) ([raw file](https://github.com/RobotLocomotion/drake/blob/716bf41649e3151bd49f19720ff9a14ee6e98eac/geometry/render_gl/factory.h#L38)):*

Yes, three months.

It's documented, though a bit hidden:

> Absent any other particular need, we prefer to use a deprecation period of three months by default, often rounded up to the next first of the month. So for code announced as deprecated on 2018-01-22 the removal_date would nominally be set to 2018-05-01.

-- https://drake.mit.edu/doxygen_cxx/drake__deprecated_8h.html#a847d827d084f28e4ae1788a0ae75be96
</blockquote></details>

OK. Thanks for the reference.
___
*[`geometry/render_gl/factory.h` line 39 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRdbvzx2ETwrypASLlx:-NRf9nO24Y2NOKcsvaNG:b-ckmbbj) ([raw file](https://github.com/RobotLocomotion/drake/blob/df666ef8e920bcf223990897aa543dfd52d85ccc/geometry/render_gl/factory.h#L39)):*

I typed a similar example into godbolt and looked at the object code: https://godbolt.org/z/TKd8f87bf

It introduces a global constructor in all code that includes this file, because we now have a non-constexpr global variable with an initializer in a header file.

We'll need to change the deprecated one to `extern` as well (and provide a definition for it in the cc files):

The header will be like this:

```c++
namespace render {
DRAKE_DEPRECATED(...)
extern const bool kHasRenderEngineGl;
}
```

The cc files will replicate what's already there for the other (non-deprecated) definition, just in another namespace.
</blockquote></details>

Done. This is my understanding.

jwnimmer-tri (commented):
> Dismissed @SeanCurtis-TRI from a discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @sherm1 and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPc8H3A-au1VLE93A91:-NRfBwYx-VAR6jUrUu5u:b9y754v) (no related file):*

Might be worth capturing in an issue so we can discuss the victory condition. It should be that `pydrake.geometry.render` disappears entirely and _all_ of it moves into geometry, not just the engines.  Once there's a ticket for that, I can cross-link to the other related issues that explain the reasoning.
</blockquote></details>

=> #19088

___

zachfang (commented):
Rebased to the latest main branch to incorporate the `drake_models.json` changes for local testing. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRhhTK--GUJCegWKJMC:b-dyxyfo)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform) (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPwrIrr9H_lAQNmjBdX:-NRhxHH05WXg5i7PmABj:b-gtkdi6) (no related file):*

I think there is a bug in this PR that's not caught by Drake CI testing, but shows up in Anzu CI:

```
what():  Invalid camera configuration; requested renderer_name = 'default' and renderer_class = 'RenderEngineGl'. The name is already used with a different type: drake::geometry::render_gl::internal::RenderEngineGl.
```
</blockquote></details>

Fixed with the change in `systems/sensors/camera_config_functions.cc`.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRd_e7MCmtBUUxxsxo7:-NRhxWc0AH9j8ZFbEsEx:b54i6z8) (no related file):*

Anzu spews a lot of deprecation warnings from this PR.  Prior to merging this PR, you need to either prepare and code-review the Anzu PR that resolves the warnings, or recruit someone on the Anzu team who will do so.
</blockquote></details>

A small PR is coming up on the Anzu side.

___

sherm1 (commented):
Thanks for untangling this mess, Zach! 
We normally deprecate for 3 months -- consider making the data 2023-07-01 if this lands soon.
Platform :lgtm: -- please be sure to take care of resulting Anzu troubles either yourself or by proxy.

> Reviewed 22 of 29 files at r1, 1 of 2 files at r2, 13 of 17 files at r3, 2 of 4 files at r5, 3 of 5 files at r6, 3 of 3 files at r7, 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRhvXwh8whZ8ZdPiJih:b-f8lwc7)* status: 4 unresolved discussions (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @zachfang)

___
*[`geometry/render_gl/test/deprecation_test.cc` line 3 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRhtNzu4pAKM6c-kbLu:-NRhtNzu4pAKM6c-kbLv:bvlo3hl) ([raw file](https://github.com/RobotLocomotion/drake/blob/95fd39de32be7d91021019bc9b2a522cca708fe5/geometry/render_gl/test/deprecation_test.cc#L3)):*
> ```cpp
> // This file serves to check that our deprecation shims compile successfully,
> // aside from warning messages.  We can remove this file entirely once the
> // deprecation period ends.
> ```

BTW please consider adding the deprecation date to this comment to make searching easier later.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 353 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRhvK_73X4NTlbYJENt:-NRhvK_73X4NTlbYJENu:b-qhd0ui) ([raw file](https://github.com/RobotLocomotion/drake/blob/95fd39de32be7d91021019bc9b2a522cca708fe5/bindings/pydrake/geometry/geometry_py_render.cc#L353)):*
> ```cpp
> 
>   {
>     // TODO(zachfang): During the 2023-08-01 deprepcation removals also
> ```

typo: deprepcation -> deprecation

zachfang (commented):
Thanks. I changed the deprecation date as I think the PR can be merged today.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRi-OLS3S_j99_xw1we:b-4c6mjg)* status: 2 unresolved discussions (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @sherm1)

___
*[`geometry/render_gl/test/deprecation_test.cc` line 3 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRhtNzu4pAKM6c-kbLu:-NRi-K8G3YpcIVR7mLWg:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/95fd39de32be7d91021019bc9b2a522cca708fe5/geometry/render_gl/test/deprecation_test.cc#L3)):*

BTW please consider adding the deprecation date to this comment to make searching easier later.
</blockquote></details>

Done.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 353 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRhvK_73X4NTlbYJENt:-NRi-Mnt9SHsJsrLrHz7:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/95fd39de32be7d91021019bc9b2a522cca708fe5/bindings/pydrake/geometry/geometry_py_render.cc#L353)):*

typo: deprepcation -> deprecation
</blockquote></details>

Done.

zachfang:
Test the mac build to be on the safe side (the `no_factory.cc` part).
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-release please

jwnimmer-tri (commented):
> I think the PR can be merged today.

It’s a bold move Cotton, let’s see if it pays off.

> Reviewed 4 of 4 files at r8, 1 of 1 files at r9, 7 of 7 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-:-NRi1gOa5jPFValC8lpI:b-ji0kh8)* status: 3 unresolved discussions (waiting on @SeanCurtis-TRI and @zachfang)

___
*[`systems/sensors/camera_config_functions.cc` line 59 at r10](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRi0ghXD2lfiieYf_E9:-NRi0ghY9ewKXwPDsuYo:b-4bkecw) ([raw file](https://github.com/RobotLocomotion/drake/blob/5a70c6890ffc8618af5aa00a1b6d99ccc3e19411/systems/sensors/camera_config_functions.cc#L59)):*
> ```cpp
>            "drake::geometry::render_vtk::internal::RenderEngineVtk"},
>           {"RenderEngineGl",
>            "drake::geometry::render_gl::internal::RenderEngineGl"}});
> ```

Why didn't Drake CI catch this?  We should not be relying on Anzu for testing the correctness of Drake's code.  This needs a unit test.

jwnimmer-tri (commented):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions (waiting on @SeanCurtis-TRI)

zachfang:
@drake-jenkins-bot mac-arm-monterey-clang-bazel-experimental-release please

sherm1 (commented):
> Reviewed 2 of 3 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 2 unresolved discussions (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Dismissed @SeanCurtis-TRI from a discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18877)* status: 1 unresolved discussion (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NPwrIrr9H_lAQNmjBdX:-NRjF7vdErz3jwYc_ENj:b-yt5au3) (no related file):*

@SeanCurtis-TRI I assume the fix and the Anzu PR suffice the original blocking intent.
</blockquote></details>

Yes, I've dismissed Sean now.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18877#-NRd_e7MCmtBUUxxsxo7:-NRjFAbu5FITy3qAGOcd:bmf954f) (no related file):*

The corresponding Anzu PR passed the CI.
</blockquote></details>

It needs to pass code review too, so it's ready to merge (modulo fixing the git sha).

I'll review it now.

___

PR #18880, opened 2023-02-27
[solvers] Change SolverBase constructor to prefer SolverId directly

jwnimmer-tri:
The constructor with a `SolverId`-returning functor is now deprecated.

Fetching the `solver_id()` should be a zero-cost abstraction, which means the function must be final and inline, rather than a virtual function that makes another indirect call.

Add an API comment and runtime warning about `SolverId` pessimizations. The `SolverId` is supposed to be a integer-like flyweight that's cheap to pass by value, but when its name exceeds 15 characters (i.e., more than SSO can handle) then we were using heap when copying it (e.g., into the `SolverOptions` maps).

While we're updating comments anyway, upgrade them all to the preferred lightweight Doxygen style.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18880)

### DIFF
--- a/solvers/BUILD.bazel
+++ b/solvers/BUILD.bazel
@@ -1613,6 +1613,7 @@ drake_cc_googletest(
     name = "solver_id_test",
     deps = [
         ":solver_id",
+        "//common/test_utilities:limit_malloc",
     ],
 )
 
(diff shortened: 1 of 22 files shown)


jwnimmer-tri:
+@ggould-tri for both(?) reviews per schedule, please.

PR #18879, opened 2023-02-27
Improve error message for MbP::GetModelInstanceByName()

RussTedrake:
Now when the model instance is not found, it helpfully reports the model instances that could have been found.

+@ggould-tri for both reviews, please.  (I think you'll approve of the change!)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18879)

PR #18898, opened 2023-02-28
[multibody] Fix O(n²) string formatting in error message

jwnimmer-tri:
Sort the names in the exception message. They hail from an unordered map, so the output to the user would be non-deterministic otherwise.

---

Amends #18879.


jwnimmer-tri:
+@ggould-tri for both reviews, please.

\CC @RussTedrake FYI

ggould-tri (commented):
:lgtm:
Not sure if "nondeterministic" is quite right, but certainly _arbitrary_ :-)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18898#-:-NPNkkGP19YIFy2TnLwq:b-h3p4h1)* status: :shipit: complete! all discussions resolved, LGTM from assignee ggould-tri(platform) (waiting on @jwnimmer-tri)

PR #18910, opened 2023-03-01
ModelVisualizer: re-pose joints by name

trowell-tri:
Makes the reload feature of ModelVisualizer more robust by setting joint positions in the reloaded model to the values of joints with the same names, rather than requiring an exact match of the number of joints before and after.

Also fixes docstrings relating to accessing the visualizer's parser, and warns the user if the reload option is invoked when the parser was accessed directly. Also adds a package map accessor method so that the main script doesn't need to access the parser method in order to get to the package map.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910)

### DIFF
--- a/bindings/pydrake/visualization/_model_visualizer.py
+++ b/bindings/pydrake/visualization/_model_visualizer.py
@@ -14,6 +14,7 @@ from pydrake.math import RigidTransform, RotationMatrix
 from pydrake.multibody.meshcat import JointSliders
 from pydrake.multibody.parsing import Parser
 from pydrake.multibody.plant import AddMultibodyPlantSceneGraph
+from pydrake.multibody.tree import JointIndex
 from pydrake.systems.analysis import Simulator
 from pydrake.systems.framework import DiagramBuilder
 from pydrake.systems.planar_scenegraph_visualizer import (
@@ -45,10 +46,10 @@ class ModelVisualizer:
       visualizer.AddModels(filename)
       visualizer.Run()
 
-    The class also provides a `parser` property to allow more complex
+    The class also provides a `parser()` method to allow more complex
     Parser handling, e.g. adding a model from a string::
 
-      visualizer.parser.AddModelsFromString(buffer_containing_model, 'sdf')
+      visualizer.parser().AddModelsFromString(buffer_containing_model, 'sdf')
 
     This class may also be run as a standalone command-line tool using the
     ``pydrake.visualization.model_visualizer`` script, or via
@@ -110,6 +111,7 @@ class ModelVisualizer:
         self._plant, self._scene_graph = AddMultibodyPlantSceneGraph(
             self._builder, time_step=0.0)
         self._parser = Parser(self._plant)
+        self._parser_accessed = False
 
     @staticmethod
     def _get_constructor_defaults():
@@ -140,7 +142,17 @@ class ModelVisualizer:
         This method cannot be used after Finalize is called.
         """
         assert self._parser is not None, "Finalize has already been called."
+        self._parser_accessed = True
         return self._parser
+
+    def package_map(self):
+        """
+        Returns the package map used by this visualizer's Parser.
+
+        This method cannot be used after Finalize is called.
+        """
+        assert self._parser is not None, "Finalize has already been called."
+        return self._parser.package_map()
 
     def meshcat(self):
         """
@@ -348,6 +360,7 @@ class ModelVisualizer:
         Only models added using AddModels() will be properly reloaded.
         If the parser() was accessed to add models directly to the parser
         then those models will not be properly reloaded by this method.
+        A warning message will be printed in this case.
 
         Args:
           position: an ndarray-like list of slider positions for the model(s);
@@ -372,13 +385,29 @@ class ModelVisualizer:
             self._context,
             self._plant_context)])
 
+        # Warn the user if the parser was directly accessed; this method
+        # *may* still work which is why we don't just abort.
+        if self._parser_accessed:
+            print("Warning: the Parser object of this ModelVisualizer was "
+                  "accessed via the parser() method before RunWithReload() "
+                  "was called; doing so may prevent model reloading from "
+                  "working properly.")
+
+        prior_pose_values = None  # Maps joint name -> pose value.
+
         # We don't just test 'position' because NumPy does weird things with
         # the truth values of arrays.
         if position is not None and len(position) > 0:
             # If we have a position list to validate, finalize the outer
-            # visualizer's plant only so we can validate positions.
+            # visualizer's plant so we can validate positions, then create
+            # a position map from the passed positions.
             self._plant.Finalize()
             self._raise_if_invalid_positions(position)
+
+            pos_to_name_map = self._get_position_to_name_map()
+            prior_pose_values = dict()
+            for index, value in enumerate(position):
+                prior_pose_values[pos_to_name_map[index]] = value
 
         if self._browser_new:
             url = self.meshcat().web_url()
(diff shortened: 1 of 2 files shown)


trowell-tri (commented):
+(release notes: none)
+@SeanCurtis-TRI for feature review, please

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-:-NPPpAI391-XkQLlTH4I:btecu8y)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

trowell-tri (commented):
I tried various things for testing but the simplest was a file like this:

```
directives:
- add_model:
    name: iiwa1
    file: package://drake/manipulation/models/iiwa_description/sdf/iiwa14_polytope_collision.sdf
- add_weld:
    parent: world
    child: iiwa1::iiwa_link_0
    X_PC: { translation: [0, 0, 0] }
- add_model:
    name: iiwa2
    file: package://drake/manipulation/models/iiwa_description/sdf/iiwa14_polytope_collision.sdf
- add_weld:
    parent: world
    child: iiwa2::iiwa_link_0
    X_PC: { translation: [1, 0, 0] }
```

And then delete or add the last two stanzas between reposing and hitting the reload button.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-:-NPPpPgG1MYeh9H6eKd_:b-o0se03)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI and @trowell-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-NPTx7oW7m6lqr38RaKZ:-NPU-FFbC61faojdlOpD:bn40shr) (no related file):*

Exposed parser

In the current state, we've got the parser exposed and one of its APIs wrapped in the model visualizer. Exposing the parser leads to a message that isn't dispositive. (I.e., we know you accessed the parser, but we don't know if that's a problem or not.)

Alternate proposal:

1. Create our own internal parser type that captures all of the models loaded.
2. All models are added to the visualizer by calling `visualizer.parser().SomeParserAPI(....)`.

This would eliminate the shadowing of APIs, and allow the full expressive capacity of the parser's API.

I'd started a small effort in this regard. I don't think I've plumbed it all the way through correctly (when I run it, no models appear). My tweaks can be found in the last commit of this branch: https://github.com/SeanCurtis-TRI/drake/tree/pr_18910_tweaks. Make of it what you will.
</blockquote></details>

This has been on my list for a while.  This PR came out of nowhere so I didn't have a chance to head it off as the pass.

My intention is that if the user accesses the parser, then the reload feature / option becomes disabled, full stop.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-NPTyGgg8rkm0TAb8ejP:-NPU-VdW9T3ynIz1AMc7:b-e1fmrt) (no related file):*

Running without the `-w` flag

As we observed, if you enable reloading but *don't* have model visualizer launch the window itself, clicking reloading causes the old server URL to become invalid and a new one gets spawned. See the following console output:


```bash
bazel run //tools:model_visualizer -- --enable_reload /home/seancurtis/code/drake/dummy.dmd.yaml 
INFO: Analyzed target //tools:model_visualizer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //bindings/pydrake/visualization:model_visualizer up-to-date:
  bazel-bin/bindings/pydrake/visualization/model_visualizer
INFO: Elapsed time: 0.260s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/bindings/pydrake/visualization/model_visualizer --enable_reload /home/seancurtis/code/drake/dummy.dmd.yaml
INFO:drake:Meshcat listening for connections at http://localhost:7000
Click 'Stop Running' or press Esc to quit
INFO:drake:Meshcat listening for connections at http://localhost:7001
Click 'Stop Running' or press Esc to quit
INFO:drake:Meshcat listening for connections at http://localhost:7002
Click 'Stop Running' or press Esc to quit
```

The goal should be the following:

1. `MeshVisualizer` starts a server which can be addressed at a specific URL. That URL should be unchanging the duration of its lifespan.
2. `MeshVisualizer` offers some basic utility that will launch a browser with its URL. This should be a simple operation that has no effect on the visualizer's state. In fact, one should be able to call it over and over and keep getting more and more windows.

We want to keep the `MeshVisualizer`'s book keeping of its internal state correctly independent on whether or not it's done the courtesy of triggering a window or not.
</blockquote></details>

Reloading has been broken for a while.  It's been on my list to rewrite to not use the weird nesting thing, which would fix this.

___

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-NPTx7oW7m6lqr38RaKZ:-NPU0GUoEXy8UDxDQoi5:bc0xanp) (no related file):*

This has been on my list for a while.  This PR came out of nowhere so I didn't have a chance to head it off as the pass.

My intention is that if the user accesses the parser, then the reload feature / option becomes disabled, full stop.
</blockquote></details>

Sorry this came out of nowhere, I just really wanted it for playing with model files, and then the inertia viz draft came so quickly that I just wanted to get this improvement out.

___

SeanCurtis-TRI (commented):
@jwnimmer-tri's comments imply the simplest path forward is to make this PR strictly about the joint pose propagation and leave everything else out for future work.  (Possibly adding the "accessing the parser disables reloads" here.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-:-NPU2EOdDxO9L4yEHNTz:b-ksr9o9)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @trowell-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-NPTx7oW7m6lqr38RaKZ:-NPU1t03B5A0vkiaKGB8:b-9vns7x) (no related file):*

Sorry this came out of nowhere, I just really wanted it for playing with model files, and then the inertia viz draft came so quickly that I just wanted to get this improvement out.
</blockquote></details>

That's fine too.

There are weaknesses of the internal parser type. I was also thinking that we don't expose the parser at all and simply control how models get added. I don't have a sense for why the parser gets exposed.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI and @trowell-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18910#-NPTx7oW7m6lqr38RaKZ:-NPU3Gk59tTPUxd9RHUX:bv04h4h) (no related file):*
> I don't have a sense for why the parser gets exposed.

The common, boring use is for adding to the package map.  I'd add a direct sugar function on the visualizer object for that, though, so we can still use "you touched the parser" as the guard for "we'll allow reloading".

The less typical case for exposing the parser when the user wants to visualize a custom scene, which is created programmatically instead of from a file.  In that case, reload doesn't make sense anyway.

___

SeanCurtis-TRI (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18910)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @trowell-tri)

Issue #15774, opened 2021-09-14
pip install: Fetch platform-independent data 


BetsyMcPhail:
Working towards #1183

Follow up to #15628

jwnimmer-tri:
To make sure that we all understand this ticket -- the victory condition here is that a user of `pip install drake` is able to easily load the models & meshes for all of Drake's included robots and manipulands, e.g., the iiwa, atlas, ycb objects, etc., with the expectation that under the hood we're fetching them from urls.  (At the moment, we've had to exclude certain model files from the whl because they are too large.)

Is that accurate?

BetsyMcPhail:
The exact implementation details still need to be worked out but that is my understanding

jwnimmer-tri:
Relates to #15024 somewhat.

jwnimmer-tri:
Relates to #13942 somewhat and #11913 more directly.

Another comment (channeling Russ) -- something like https://pytorch.org/tutorials/beginner/basics/data_tutorial.html may be the best solution here.  Either that library directly, or some Drake-compatible simpler implementation.

jwnimmer-tri:
https://packages.ubuntu.com/source/bionic/ros-resource-retriever might be useful.

jwnimmer-tri:
For better issue search, here's the list of mesh file paths that are excluded from the wheel:

```
rm -rf \
    ${WHEEL_DATA_DIR}/manipulation/models/franka_description/meshes \
    ${WHEEL_DATA_DIR}/manipulation/models/tri-homecart/*.obj \
    ${WHEEL_DATA_DIR}/manipulation/models/tri-homecart/*.png \
    ${WHEEL_DATA_DIR}/manipulation/models/ur3e/*.obj \
    ${WHEEL_DATA_DIR}/manipulation/models/ur3e/*.png \
    ${WHEEL_DATA_DIR}/manipulation/models/ycb/meshes \
    ${WHEEL_DATA_DIR}/examples/atlas \
    ${WHEEL_DATA_DIR}/examples/hydroelastic/spatula_slip_control
```

https://github.com/RobotLocomotion/drake/blob/4a0ece64324368d8c1d3a212ac16234668b00b69/tools/wheel/image/build-wheel.sh#L99-L107

SwappyG:
I'm running into this issue when trying to load models (@jwnimmer-tri redirected me to here from my stackoverflow question. I'd be happy to help with making a PR for this issue if it's not actively being worked on.

[...]

PR #18923, opened 2023-03-02
Tease apart models_filegroup vs install_data

### LINKED ISSUE #11010: install_data: Make explicit, remove unused cruft
(Warning: Rant-y)

At present, `install_data` implicitly names its target `models`. I've always been confused by this, and I want that to change.

Additionally, there are *zero* usages of any customization of `prod_models_prefix` or `test_models_prefix`. Those should be explicitly defined too, rather than arbitrary prefixes. The `extra_prod_models` story also makes me sad.

This should be broken into two parts:
* Model globbing should be consolidated, to permit globbing install + test data. (Test data isn't even installed; why is it even part of `install_data`???)
* Installation for models should be consolidated (and named specifically for `drake`, as it does that with its `dest` prefix).

EDIT: More complaint: Why do we even concatenate `prod_models` + `test_models` into `models`??? We should only have `models` and `test_models`; if anything, `test_models` should depend on `models` (if we even really want that combination.)

Ran into this when considering #11009

jwnimmer-tri:
Delete unused `PendulumWithFriction.urdf`.

---

I've tested locally and the list of installed files is the same between this PR and the master branch.

Closes #11010.

Towards #15774.

FYI https://drake.mit.edu/stable.html#model-files

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18923)


jwnimmer-tri:
+@EricCousineau-TRI I think you get right of first refusal for this review.  If you're busy, we can push it to @sammy-tri instead.

jwnimmer-tri (commented):
> Reviewed 34 of 34 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18923)* status: LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
:lgtm_strong:, beautimous! Small nit on globbing, though maybe it's pre-existing.

> Reviewed 23 of 34 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18923#-:-NPXlOLk6XlTpoQqHSjQ:b-drk5qs)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`tools/skylark/drake_data.bzl` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18923#-NPXkzhcDhnCEIFGV1Jk:-NPXkzhdC7znHxgCz9cX:b-a07odz) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ece06b5372da38c26db7a5b09c05d88249d72c1/tools/skylark/drake_data.bzl#L36)):*
> ```Starlark
>     ]
>     exclude = glob_exclude + [
>         "**/test/*",
> ```

nit (pre-existing) It's not clear to me how intuitive this exclude pattern will be for `testonly = True`.

Possible options:
- TODO
- `glob_exclude` gains these defaults only when `testonly = False`
- The default values are these, explicitly, perhaps via global constant `DEFAULT_GLOB_EXCLUDE`, but users can override them. Docstring says they should?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18923)* status: 1 unresolved discussion (waiting on @EricCousineau-TRI)

___
*[`tools/skylark/drake_data.bzl` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18923#-NPXkzhcDhnCEIFGV1Jk:-NPXyEmO8_y2qdqQmTjR:bmxjmc6) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ece06b5372da38c26db7a5b09c05d88249d72c1/tools/skylark/drake_data.bzl#L36)):*

nit (pre-existing) It's not clear to me how intuitive this exclude pattern will be for `testonly = True`.

Possible options:
- TODO
- `glob_exclude` gains these defaults only when `testonly = False`
- The default values are these, explicitly, perhaps via global constant `DEFAULT_GLOB_EXCLUDE`, but users can override them. Docstring says they should?
</blockquote></details>

Aha!  Thanks, I'd meant to circle back here and totally forgot.

There are no tests that use this.  I've update the docs.

PTAL.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18923)* status: :shipit: complete! all discussions resolved, LGTM from assignee EricCousineau-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`tools/skylark/drake_data.bzl` line 36 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18923#-NPXkzhcDhnCEIFGV1Jk:-NPYdigU3iYHsu29bgJo:b-o58whx) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ece06b5372da38c26db7a5b09c05d88249d72c1/tools/skylark/drake_data.bzl#L36)):*

Aha!  Thanks, I'd meant to circle back here and totally forgot.

There are no tests that use this.  I've update the docs.

PTAL.
</blockquote></details>

OK Sweet ,thanks!

PR #18933, opened 2023-03-02
[visualization] Fix model_visualizer reloading

jwnimmer-tri:
Previously, after pushing the reload button it would vanish along with the stop button. The new control flow logic is less brittle and will be easier to maintain over time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)

### DIFF
--- a/bindings/pydrake/multibody/parsing_py.cc
+++ b/bindings/pydrake/multibody/parsing_py.cc
@@ -34,6 +34,7 @@ PYBIND11_MODULE(parsing, m) {
     py::class_<Class> cls(m, "PackageMap", cls_doc.doc);
     cls  // BR
         .def(py::init<>(), cls_doc.ctor.doc)
+        .def(py::init<const Class&>(), py::arg("other"), "Copy constructor")
         .def("Add", &Class::Add, py::arg("package_name"),
             py::arg("package_path"), cls_doc.Add.doc)
         .def("AddMap", &Class::AddMap, py::arg("other_map"), cls_doc.AddMap.doc)
@@ -62,6 +63,7 @@ PYBIND11_MODULE(parsing, m) {
         .def("PopulateFromRosPackagePath", &Class::PopulateFromRosPackagePath,
             cls_doc.PopulateFromRosPackagePath.doc)
         .def_static("MakeEmpty", &Class::MakeEmpty, cls_doc.MakeEmpty.doc);
+    DefCopyAndDeepCopy(&cls);
   }
 
   // Parser
--- a/bindings/pydrake/multibody/test/parsing_test.py
+++ b/bindings/pydrake/multibody/test/parsing_test.py
@@ -38,9 +38,11 @@ from pydrake.multibody.plant import (
 class TestParsing(unittest.TestCase):
 
     def test_package_map(self):
-        # Simple coverage test for default constructor
+        # Simple coverage test for constructors.
         dut = PackageMap()
         self.assertEqual(dut.size(), 1)
+        PackageMap(other=dut)
+        copy.copy(dut)
 
         dut = PackageMap.MakeEmpty()
         dut2 = PackageMap.MakeEmpty()
--- a/bindings/pydrake/visualization/BUILD.bazel
+++ b/bindings/pydrake/visualization/BUILD.bazel
@@ -147,10 +147,14 @@ drake_py_unittest(
     name = "model_visualizer_test",
     data = [
         ":model_visualizer",
+        "//geometry:meshcat_websocket_client",
         "//manipulation/util:test_models",
         "//multibody/benchmarks/acrobot:models",
     ],
-    deps = [":model_visualizer"],
+    deps = [
+        ":model_visualizer",
+        "//bindings/pydrake/common/test_utilities:deprecation_py",
+    ],
 )
 
 drake_py_unittest(
--- a/bindings/pydrake/visualization/_model_visualizer.py
+++ b/bindings/pydrake/visualization/_model_visualizer.py
@@ -1,9 +1,11 @@
+import copy
 from enum import Enum
 import time
-import webbrowser
+from webbrowser import open as _webbrowser_open
 
 import numpy as np
 
+from pydrake.common.deprecation import deprecated
 from pydrake.geometry import (
     Cylinder,
     GeometryInstance,
@@ -12,10 +14,8 @@ from pydrake.geometry import (
 )
 from pydrake.math import RigidTransform, RotationMatrix
 from pydrake.multibody.meshcat import JointSliders
-from pydrake.multibody.parsing import Parser
-from pydrake.multibody.plant import AddMultibodyPlantSceneGraph
+from pydrake.planning import RobotDiagramBuilder
 from pydrake.systems.analysis import Simulator
-from pydrake.systems.framework import DiagramBuilder
 from pydrake.systems.planar_scenegraph_visualizer import (
     ConnectPlanarSceneGraphVisualizer,
 )
@@ -26,7 +26,7 @@ from pydrake.visualization import (
 
 
 class RunResult(Enum):
-    """Result code indicating how Run() terminated."""
+    """This class is deprecated and will be removed on or after 2023-07-01."""
     KEEP_GOING = 0  # Note that this is never returned but is used internally.
     LOOP_ONCE = 1
     STOPPED = 2
@@ -77,7 +77,7 @@ class ModelVisualizer:
           publish_contacts: a flag for VisualizationConfig.
 
           browser_new: a flag that will open the MeshCat display in a new
-            browser window.
+            browser window during Run().
           pyplot: a flag that will open a pyplot figure for rendering using
             PlanarSceneGraphVisualizer.
 
@@ -93,23 +93,47 @@ class ModelVisualizer:
         self._pyplot = pyplot
         self._meshcat = meshcat
 
-        # This is both a button label and a flag to enable the button.
-        self._reload_button_name = None
+        # This is the list of loaded models, to enable the Reload button.
+        # If set to None, it means that we won't support reloading because
+        # the user might have added models outside of our purview.
         self._model_filenames = list()
 
-        # The following fields remain valid for this object's lifetime after
-        # Finalize() has been called.
-        self._diagram = None
+        # This is set to a non-None value iff our Meshcat has a reload button.
+        self._reload_button_name = None
+
+        # The builder is set to None during Finalize(), though during a Reload
+        # it will be temporarily resurrected.
+        self._builder = RobotDiagramBuilder()
+
+        # The following fields are set non-None during Finalize().
+        self._original_package_map = None
+        self._diagram = None  # This will be a planning.RobotDiagram.
         self._sliders = None
         self._context = None
-        self._plant_context = None
 
-        # The builder, scene_graph, and parser become invalid after Finalize().
-        # The plant remains valid for this object's lifetime.
-        self._builder = DiagramBuilder()
-        self._plant, self._scene_graph = AddMultibodyPlantSceneGraph(
-            self._builder, time_step=0.0)
-        self._parser = Parser(self._plant)
+    def _check_rep(self, *, finalized):
+        """
+        Checks that our self members are consistent with the provided expected
+        state of finalization.
+        """
+        if not finalized:
+            # The builder is alive.
+            assert self._builder is not None
+            # The meshcat might or might not exist yet.
+            # Everything else is dead.
+            assert self._original_package_map is None
+            assert self._diagram is None
+            assert self._sliders is None
+            assert self._context is None
+        else:
+            # The builder is dead.
+            assert self._builder is None
+            # Everything else is alive.
+            assert self._meshcat is not None
+            assert self._original_package_map is not None
+            assert self._diagram is not None
+            assert self._sliders is not None
+            assert self._context is not None
 
     @staticmethod
     def _get_constructor_defaults():
@@ -133,14 +157,39 @@ class ModelVisualizer:
             result[name] = value
         return result
 
-    def parser(self):
+    def package_map(self):
         """
-        Returns a Parser that will load models into this visualizer.
+        Returns the PackageMap being used during parsing. Users should add
+        entries here to be able to load models from non-Drake packages.
 
         This method cannot be used after Finalize is called.
         """
-        assert self._parser is not None, "Finalize has already been called."
-        return self._parser
+        if self._builder is None:
+            raise ValueError("Finalize has already been called.")
+        self._check_rep(finalized=False)
+        # It's safe to let the user change the package map. We'll make a copy
+        # of it during Finalize().
+        return self._builder.parser().package_map()
+
+    def parser(self):
+        """
+        (Advanced) Returns a Parser that will load models into this visualizer.
+
+        Prefer to use package_map() and AddModels() to load models, instead of
+        this method.
+
+        Calling this method will disable the "Reload" button in the visualizer,
+        because we can no longer determine the scope of what to reload.
+
+        This method cannot be used after Finalize is called.
+        """
+        if self._builder is None:
+            raise ValueError("Finalize has already been called.")
+        self._check_rep(finalized=False)
+        # We can't easily know what the user is going to do with the parser,
+        # so we need to disable model reloading once they access it.
+        self._model_filenames = None
+        return self._builder.parser()
 
     def meshcat(self):
         """
@@ -160,8 +209,12 @@ class ModelVisualizer:
         Args:
           filename: the name of a file containing one or more models.
         """
-        self._parser.AddModels(filename)
-        self._model_filenames.append(filename)
+        if self._builder is None:
+            raise ValueError("Finalize has already been called.")
+        self._check_rep(finalized=False)
+        self._builder.parser().AddModels(filename)
+        if self._model_filenames is not None:
+            self._model_filenames.append(filename)
 
     def Finalize(self, position=None):
         """
@@ -173,26 +226,19 @@ class ModelVisualizer:
             the number of positions in all model(s), including the 7 positions
             corresponding to any model bases that aren't welded to the world.
         """
+        self._check_rep(finalized=False)
         if self._builder is None:
             raise RuntimeError("Finalize has already been called.")
-        assert all([x is not None for x in (
-            self._plant,
-            self._scene_graph,
-            self._parser)])
-        assert all([x is None for x in (
-            self._diagram,
-            self._sliders,
-            self._context,
-            self._plant_context)])
 
         if self._visualize_frames:
             # Find all the frames and draw them using _add_triad().
             # The frames are drawn using the parsed length.
             # The world frame is drawn thicker than the rest.
-            inspector = self._scene_graph.model_inspector()
+            inspector = self._builder.scene_graph().model_inspector()
             for frame_id in inspector.GetAllFrameIds():
+                world_id = self._builder.scene_graph().world_frame_id()
                 radius = self._triad_radius * (
-                    3 if frame_id == self._scene_graph.world_frame_id() else 1
+                    3 if frame_id == world_id else 1
                     )
                 self._add_triad(
                     frame_id,
@@ -201,7 +247,7 @@ class ModelVisualizer:
                     opacity=self._triad_opacity,
                 )
 
-        self._plant.Finalize()
+        self._builder.plant().Finalize()
 
         # (Re-)initialize the meshcat instance, creating one if needed.
         self.meshcat()
@@ -210,7 +256,8 @@ class ModelVisualizer:
 
         # We want to place this button far away from the Stop Running button,
         # hence the work to do this here.
-        if self._reload_button_name:
+        if self._model_filenames:
+            self._reload_button_name = "Reload model files"
             self._meshcat.AddButton(self._reload_button_name)
 
         # Connect to drake_visualizer, meldis, and meshcat.
@@ -220,34 +267,35 @@ class ModelVisualizer:
             config=VisualizationConfig(
                 publish_contacts=self._publish_contacts,
                 enable_alpha_sliders=True),
-            plant=self._plant,
-            scene_graph=self._scene_graph,
-            builder=self._builder,
+            plant=self._builder.plant(),
+            scene_graph=self._builder.scene_graph(),
+            builder=self._builder.builder(),
             meshcat=self._meshcat)
 
         # Add joint sliders to meshcat.
-        self._sliders = self._builder.AddNamedSystem(
+        self._sliders = self._builder.builder().AddNamedSystem(
             "joint_sliders", JointSliders(meshcat=self._meshcat,
-                                          plant=self._plant))
+                                          plant=self._builder.plant()))
 
         # Connect to PyPlot.
         if self._pyplot:
-            ConnectPlanarSceneGraphVisualizer(self._builder, self._scene_graph)
+            ConnectPlanarSceneGraphVisualizer(
+                self._builder.builder(), self._builder.scene_graph())
 
+        self._original_package_map = copy.copy(
+            self._builder.parser().package_map())
         self._diagram = self._builder.Build()
+        self._builder = None
         self._context = self._diagram.CreateDefaultContext()
-        self._plant_context = self._plant.GetMyContextFromRoot(self._context)
 
         # We don't just test 'position' because NumPy does weird things with
         # the truth values of arrays.
         if position is not None and len(position) > 0:
             self._raise_if_invalid_positions(position)
-            self._plant.SetPositions(self._plant_context, position)
             self._sliders.SetPositions(position)
-
-        if self._browser_new:
-            url = self._meshcat.web_url()
-            webbrowser.open(url=url, new=self._browser_new)
+            self._diagram.plant().SetPositions(
+                self._diagram.plant().GetMyContextFromRoot(self._context),
+                position)
 
         # Use Simulator to dispatch initialization events.
         # TODO(eric.cousineau): Simplify as part of #13776 (was #10015).
@@ -259,9 +307,31 @@ class ModelVisualizer:
         # the checkbox in the meshcat controls.
         self._meshcat.SetProperty("proximity", "visible", False)
 
-        self._builder = None
-        self._scene_graph = None
-        self._parser = None
+        self._check_rep(finalized=True)
+
+    def _reload(self):
+        """
+        Re-creates the Diagram using the same sequence of calls to AddModels
+        as the user performed. In effect, this will refresh the visualizer to
+        show any changes the user made on disk to their models.
+        """
+        self._check_rep(finalized=True)
+        assert self._model_filenames is not None
+
+        # Clear out the old diagram.
+        self._diagram = None
+        self._sliders = None
+        self._context = None
+
+        # Populate the diagram builder again with the same packages and models.
+        self._builder = RobotDiagramBuilder()
+        self._builder.parser().package_map().AddMap(self._original_package_map)
+        self._original_package_map = None
+        for filename in self._model_filenames:
+            self._builder.parser().AddModels(filename)
+
+        # Finalize the rest of the systems and widgets.
+        self.Finalize()
 
     def Run(self, position=None, loop_once=False):
         """
@@ -277,35 +347,31 @@ class ModelVisualizer:
             7 positions corresponding to any model bases that aren't welded to
             the world.
           loop_once: a flag that exits the evaluation loop after one pass.
-
-        Returns:
-          A RunResult code indicating why the method stopped.
         """
-        if self._builder:
+        if self._builder is not None:
             self.Finalize(position=position)
-        elif position is not None and len(position) > 0:
-            self._plant.SetPositions(self._plant_context, position)
-            self._sliders.SetPositions(position)
-            self._diagram.ForcedPublish(self._context)
+        else:
+            self._check_rep(finalized=True)
+            if position is not None and len(position) > 0:
+                self._diagram.plant().SetPositions(
+                    self._diagram.plant().GetMyContextFromRoot(self._context),
+                    position)
+                self._sliders.SetPositions(position)
+                self._diagram.ForcedPublish(self._context)
 
-        assert all([x is None for x in (
-            self._builder,
-            self._scene_graph,
-            self._parser)])
-        assert all([x is not None for x in (
-            self._plant,
-            self._diagram,
-            self._sliders,
-            self._context,
-            self._plant_context,
-            self._meshcat)])
+        # Everything is finally fully configured. We can open the window now.
+        # TODO(jwnimmer-tri) The browser_new config knob would probably make
+        # more sense as an argument to Run() vs an argument to our constructor.
+        if self._browser_new:
+            self._browser_new = False
+            url = self._meshcat.web_url()
+            _webbrowser_open(url=url, new=True)
 
         # Wait for the user to cancel us.
         stop_button_name = "Stop Running"
         if not loop_once:
             print(f"Click '{stop_button_name}' or press Esc to quit")
 
-        loop_result = RunResult.KEEP_GOING
         try:
             self._meshcat.AddButton(stop_button_name, "Escape")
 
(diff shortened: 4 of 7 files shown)


jwnimmer-tri:
+@trowell-tri for feature review, please.

trowell-tri (commented):
> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: LGTM missing from assignees trowell-tri,SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

trowell-tri (commented):
A couple of issues, a typo, and a few nits which are more questions than anything, but otherwise :lgtm: 

> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-:-NP_-xLL-onMQi66A1hj:bnhfm3j)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 111 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZn7HC9THyyofVcpwO:-NPZn7HC9THyyofVcpwP:bmr6n8i) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L111)):*
> ```Python
>         # The following fields are set non-None during Finalize().
>         self._original_package_map = None
>         self._diagram = None  # This will of a planning.RobotDiagram.
> ```

nit this comment is incomplete I think … "will be an instance of a" ?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZGGfq3221aJcwsdiz:-NPZGGfq3221aJcwsdj-:bgnhb0c) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L118)):*
> ```Python
>         """
>         Checks that our self members are consistent with the provided expected
>         staate of finalization.
> ```

typo "state"



_Code quote:_
```Python
staate
```
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPZs-h77L1V2j-lZLlt:bt95cc7) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*
> ```Python
>         This method cannot be used after Finalize is called.
>         """
>         assert self._builder is not None, "Finalize has already been called."
> ```

nit it seems weird to manually check part of the rep and then call `_check_rep()` but I assume that's just to get the assert message? The message is nice but possibly could be rolled into `_check_rep`? There is one case though where we use that message in an exception so maybe not worth it.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 214 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZrC8N6wDPxycf1Bta:-NPZrC8N6wDPxycf1Btb:b9by2lo) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L214)):*
> ```Python
>         self._check_rep(finalized=False)
>         self._builder.parser().AddModels(filename)
>         self._model_filenames.append(filename)
> ```

If you call `model_viz.parser()` and then `model_viz.AddModels(foo)` then it looks like you'd get a `NoneType has no attribute` error because we don't check, and we also don't advise the user of this in the docstring. I _think_ what we want is to just check rather than die?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPZyedt1K0YXcXPctsO:b-gx2usy) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*
> ```Python
>         if self._browser_new:
>             url = self._meshcat.web_url()
>             webbrowser.open(url=url, new=self._browser_new)
> ```

This change fixes the issue where reload button would disappear if you didn't use the `--open-window` CLI flag, but now if you do use that flag you get a new window every time you hit reload, which didn't happen before. The `_Reload` method could clear the `_browser_new` attribute before calling `Finalize` to fix this since we only need to make a new window during the first pass through `Finalize`.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 353 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZvlLO3ZFu9LhnzGXc:-NPZvlLO3ZFu9LhnzGXd:b-61ysmy) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L353)):*
> ```Python
> 
>         Returns:
>           A RunResult code indicating why the method stopped.
> ```

btw returning the `RunResult` was only required before because the former outer visualizer needed to understand the end state of the inner visualizer's `Run` method. Seems like we could do away with that now? Or deprecate it?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZwUQA4C9x_z7R5GCI:-NPZwUQA4C9x_z7R5GCJ:brc4cww) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L406)):*
> ```Python
>         return loop_result
> 
>     @deprecated("Use Run() instead.", date="2023-07-01")
> ```

nit since we've deprecated this do we need to deprecate the command line flag as well? Making this a nit because I'm guessing the answer is "no".

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri, @SeanCurtis-TRI, and @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NP_WQ0_8WBeq7_9o8fv:bh3way3) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

nit it seems weird to manually check part of the rep and then call `_check_rep()` but I assume that's just to get the assert message? The message is nice but possibly could be rolled into `_check_rep`? There is one case though where we use that message in an exception so maybe not worth it.
</blockquote></details>

These checks are enforcing the precondition of the function.  When they trip, it's because the user made a mistake in their calling code.

Check rep is enforcing an invariant over the class's fields.  When it trips, it's because a drake developer made a mistake inside this class's code.

To make that more clear, I've changed the user-facing checks to use ValueError.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 214 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZrC8N6wDPxycf1Bta:-NP_ZjuM7I-0ez0ACSUG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L214)):*

If you call `model_viz.parser()` and then `model_viz.AddModels(foo)` then it looks like you'd get a `NoneType has no attribute` error because we don't check, and we also don't advise the user of this in the docstring. I _think_ what we want is to just check rather than die?
</blockquote></details>

Done.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NP_WrB--bzfkWw7wO5Z:b-v3d1nv) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

This change fixes the issue where reload button would disappear if you didn't use the `--open-window` CLI flag, but now if you do use that flag you get a new window every time you hit reload, which didn't happen before. The `_Reload` method could clear the `_browser_new` attribute before calling `Finalize` to fix this since we only need to make a new window during the first pass through `Finalize`.
</blockquote></details>

Done, but I haven't quite got regression test coveraged plumbed here yet.  Stay tuned.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 353 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZvlLO3ZFu9LhnzGXc:-NP_WKia3XiqrGQ5K5dH:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L353)):*

btw returning the `RunResult` was only required before because the former outer visualizer needed to understand the end state of the inner visualizer's `Run` method. Seems like we could do away with that now? Or deprecate it?
</blockquote></details>

Done.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZwUQA4C9x_z7R5GCI:-NP_VN5R7r5j-GzLwf8R:bbm4kb) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L406)):*

nit since we've deprecated this do we need to deprecate the command line flag as well? Making this a nit because I'm guessing the answer is "no".
</blockquote></details>

Stability policy is here:
https://drake.mit.edu/stable.html
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 395 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NP_bbYGDax5o5Np4qwo:-NP_bbYGDax5o5Np4qwp:b-7ey6tw) ([raw file](https://github.com/RobotLocomotion/drake/blob/ad925b85140c06563b2f5288480e2b35c0f07f7f/bindings/pydrake/visualization/_model_visualizer.py#L395)):*
> ```Python
>                 if has_clicks(self._reload_button_name):
>                     self._meshcat.DeleteButton(stop_button_name)
>                     self._reload()
> ```

Working

Add regression test coverage.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI and @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NP_cAQO387CaM9PkDBf:b-bp53mp) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

Done, but I haven't quite got regression test coveraged plumbed here yet.  Stay tuned.
</blockquote></details>

Done.

This one was actually easy.  Moving the browser window launching to Run() is correct by inspection.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI and @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 395 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NP_bbYGDax5o5Np4qwo:-NP_kktU2zcvWLyPhGfb:ba4k4u0) ([raw file](https://github.com/RobotLocomotion/drake/blob/ad925b85140c06563b2f5288480e2b35c0f07f7f/bindings/pydrake/visualization/_model_visualizer.py#L395)):*

Working

Add regression test coverage.
</blockquote></details>

Done.  It took me several false starts, but I like what's here now.

SeanCurtis-TRI (commented):
:LGTM: With my own collection of nits.

> Reviewed 6 of 6 files at r1, 2 of 4 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-:-NPc-DVf4YyfPbF5TsaF:b9mx87e)* status: 7 unresolved discussions (waiting on @jwnimmer-tri and @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZBMdF8_-z1UcdTnDG:-NPZBMdGC0fasK-a9Scf:b-d8v3d3) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L120)):*
> ```Python
>         staate of finalization.
>         """
>         if not finalized:
> ```

BTW Why the negative predicate? I always find them harder to read than the positive so reserve them for when it's awkward to provide a positive predicate?

Is there any harm with making it positive and reversing the blocks?
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPbx0o9Eq-atpW-viEm:bg0wpov) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

These checks are enforcing the precondition of the function.  When they trip, it's because the user made a mistake in their calling code.

Check rep is enforcing an invariant over the class's fields.  When it trips, it's because a drake developer made a mistake inside this class's code.

To make that more clear, I've changed the user-facing checks to use ValueError.
</blockquote></details>

Something smells odd about the code and the reasoning.

I buy the idea that you want to separate function preconditions from implementation errors (user vs. developer).

But the fact that `finalized=False` is 100% correlated with user-facing preconditions and `finalized=True` is 100% correlated with developer reality checks, makes the copy-and-paste nature of the prefixed `ValueError` seem like a mistake. Such a high degree of correlation suggests an opportunity for refactoring.

That said, I'm not going to block on this. The code is correct...just slightly odd.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPbu7TJ5vLa6awgyCyl:bqtm28f) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

Done.

This one was actually easy.  Moving the browser window launching to Run() is correct by inspection.
</blockquote></details>

nit For my money, that bool shouldn't be a member of the class. It should simply be a parameter of `Run()`. It should never reflect the state of the visualizer, just a courtesy operation related to starting the server. It also simplifies the unit test.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 405 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbuj1S4U_x4Bkif4b8:-NPbuj1T4uNIGEPAhq9e:b-wwwkcw) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/_model_visualizer.py#L405)):*
BTW This has a weird smell to me. A non-deprecated API returning a deprecated type. It's a weird scenario for deprecation. Most of our deprecations have the flavor of, "Don't call *this* API, but call *that* one." In this case, the deprecation isn't to encourage people to call a different API, but to encourage them to ignore the return value.

Rather than simply removing the documentation of the return type, I'd advocate putting a comment along those lines:

> The return value has been deprecated and will be removed on DATE. There is no need to consider the return value and, in the future, there will be no return value at all.
___
*[`bindings/pydrake/visualization/test/model_visualizer_test.py` line 97 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbvioB1oz-LQhBYzD1:-NPbvioB1oz-LQhBYzD2:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/test/model_visualizer_test.py#L97)):*
> ```Python
>         mut_private._webbrowser_open = self._mock_webbrowser_open
>         # When we want to allow webbrowser.open, this will be a list() that
>         # captures the kwargs it was called it.
> ```

nit: typo



_Suggestion:_
```Python
with
```
___
*[`bindings/pydrake/visualization/test/model_visualizer_test.py` line 181 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbw2LNCHbLdMgX0gsl:-NPbw2LOE2W3OPovimML:b-t5jvlx) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/test/model_visualizer_test.py#L181)):*
> ```Python
>         dut = mut.ModelVisualizer(meshcat=meshcat)
>         filename = "drake/multibody/benchmarks/acrobot/acrobot.sdf"
>         dut.AddModels(FindResourceOrThrow(filename))
> ```

BTW A real shame one can't simply pass a `package://drake` URL after all of your good work in this regard.
___
*[`bindings/pydrake/visualization/test/model_visualizer_test.py` line 189 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbwIdoD8CTCWB9cvtU:-NPbwIdoD8CTCWB9cvtV:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/test/model_visualizer_test.py#L189)):*
> ```Python
>         self.assertEqual(meshcat.GetButtonClicks(button), 0)
> 
>         # Remember the originally-created digram.
> ```

nit: typo



_Suggestion:_
```Python
diagram
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions (waiting on @SeanCurtis-TRI and @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZBMdF8_-z1UcdTnDG:-NPc6H_e75uAcq0eGBzY:b-pb6kaw) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L120)):*

BTW Why the negative predicate? I always find them harder to read than the positive so reserve them for when it's awkward to provide a positive predicate?

Is there any harm with making it positive and reversing the blocks?
</blockquote></details>

I've placed the checks in their lifecycle order.  The initial conditions first, the final conditions second.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPc71lm3k8zlDy5k6-h:b-ip17ds) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

nit For my money, that bool shouldn't be a member of the class. It should simply be a parameter of `Run()`. It should never reflect the state of the visualizer, just a courtesy operation related to starting the server. It also simplifies the unit test.
</blockquote></details>

I probably agree, but in any case that's a lot more deprecation churn, so I've done a TODO for now.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 405 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbuj1S4U_x4Bkif4b8:-NPc6SPf6uxMP4u6Rx-V:bxt4jbq) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/_model_visualizer.py#L405)):*

BTW This has a weird smell to me. A non-deprecated API returning a deprecated type. It's a weird scenario for deprecation. Most of our deprecations have the flavor of, "Don't call *this* API, but call *that* one." In this case, the deprecation isn't to encourage people to call a different API, but to encourage them to ignore the return value.

Rather than simply removing the documentation of the return type, I'd advocate putting a comment along those lines:

> The return value has been deprecated and will be removed on DATE. There is no need to consider the return value and, in the future, there will be no return value at all.
</blockquote></details>

My money is on my current way being the least confusing for users.
___
*[`bindings/pydrake/visualization/test/model_visualizer_test.py` line 181 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbw2LNCHbLdMgX0gsl:-NPc6bl6AU8td_G7fKTU:b-j4bogb) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/test/model_visualizer_test.py#L181)):*

BTW A real shame one can't simply pass a `package://drake` URL after all of your good work in this regard.
</blockquote></details>

Yes, I have that in mind for future work.  I wanted to keep this PR as a lateral move only.

trowell-tri (commented):
> Reviewed 2 of 4 files at r2, 2 of 4 files at r4, 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 2 unresolved discussions (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPc7A4Y5eDzOhKTBcTy:btx3kgl) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

Something smells odd about the code and the reasoning.

I buy the idea that you want to separate function preconditions from implementation errors (user vs. developer).

But the fact that `finalized=False` is 100% correlated with user-facing preconditions and `finalized=True` is 100% correlated with developer reality checks, makes the copy-and-paste nature of the prefixed `ValueError` seem like a mistake. Such a high degree of correlation suggests an opportunity for refactoring.

That said, I'm not going to block on this. The code is correct...just slightly odd.
</blockquote></details>

Now it looks like the `ValueError` stanza isn't reachable? Since `_check_rep` guarantees builder is not none here? Same issue in `package_map()` and `AddModels()` and `Finalize()`. I must be missing something.

I didn't see `_check_rep` as being split the way Sean describes, though. The contents there didn't surprise me at all, it seems to enforce the proper state of the visualizer's life cycle. ¯\_(ツ)_/¯
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPcBw056Awq4i2vv0L4:b27syoj) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

I probably agree, but in any case that's a lot more deprecation churn, so I've done a TODO for now.
</blockquote></details>

I agree with Sean here as well, but that ended up as a class member so we could do the "construct an object to get the defaults" trick. IIRC those defaults were initially separate values so they could be accessed by both the constructor and the outer `model_visualizer.py` (not leading underscore) script and it sounds like something like that would be better? I don't think it's critical to change though.

Moving the browser window launching to `Run` seems like the most obvious fix, wish I had thought of it.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZwUQA4C9x_z7R5GCI:-NPcDkL87r15e7zu4gyY:buytnpp) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L406)):*

Stability policy is here:
https://drake.mit.edu/stable.html
</blockquote></details>

I don't see any comments about CLI arguments, which I guess means we make no promises? I figured stability was API only when I wrote the above. Certainly if you make the mistake of passing the old flag you get a reasonable error message.
___
*[`bindings/pydrake/visualization/test/model_visualizer_test.py` line 210 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPc9E6b2aLwyyOgE-z1:-NPc9E6b2aLwyyOgE-z2:bz6tzvn) ([raw file](https://github.com/RobotLocomotion/drake/blob/0bb06287ee3588dc75c1c87c5640875e83906b31/bindings/pydrake/visualization/test/model_visualizer_test.py#L210)):*
> ```Python
>     def test_webbrowser(self):
>         """
>         Checks that the webbrower launch command is properly invoked.
> ```

typo "webbrowser"



_Code quote:_
```Python
webbrower
```

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPcHb1T3aoNxpUGTlPC:b-we5dyd) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

I agree with Sean here as well, but that ended up as a class member so we could do the "construct an object to get the defaults" trick. IIRC those defaults were initially separate values so they could be accessed by both the constructor and the outer `model_visualizer.py` (not leading underscore) script and it sounds like something like that would be better? I don't think it's critical to change though.

Moving the browser window launching to `Run` seems like the most obvious fix, wish I had thought of it.
</blockquote></details>

(I'm blocking here only to get comments on what people think we should do about the defaults situation in the future; perhaps blocking is too strict a disposition?)

trowell-tri (commented):
> Reviewed 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 3 unresolved discussions (waiting on @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPcLmnG3Gs04DYnoMDg:bxex4xf) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

Now it looks like the `ValueError` stanza isn't reachable? Since `_check_rep` guarantees builder is not none here? Same issue in `package_map()` and `AddModels()` and `Finalize()`. I must be missing something.

I didn't see `_check_rep` as being split the way Sean describes, though. The contents there didn't surprise me at all, it seems to enforce the proper state of the visualizer's life cycle. ¯\_(ツ)_/¯
</blockquote></details>

Oops.  I tried to make it a bit more clear, and broke it.  Working.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPcLuIvDFyn-kWd_4EN:bakwiqo) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

(I'm blocking here only to get comments on what people think we should do about the defaults situation in the future; perhaps blocking is too strict a disposition?)
</blockquote></details>

I don't think hard-coding the default for "open a web browser" to "False" is any great loss.  It's not like True will ever become the default.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 406 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZwUQA4C9x_z7R5GCI:-NPcM1ctDm1xG_IIUVex:b-ht0ukm) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L406)):*

I don't see any comments about CLI arguments, which I guess means we make no promises? I figured stability was API only when I wrote the above. Certainly if you make the mistake of passing the old flag you get a reasonable error message.
</blockquote></details>

That's right.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 2 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 299 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZyedt1K0YXcXPctsN:-NPcMEKC7fW7WZZD1ZDX:b-3ziuu1) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L299)):*

I don't think hard-coding the default for "open a web browser" to "False" is any great loss.  It's not like True will ever become the default.
</blockquote></details>

That's true, we did the same thing to `loop_once` and for the better. Thanks.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: 1 unresolved discussion (waiting on @trowell-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPcN2blC1CqvC3huS7q:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

Oops.  I tried to make it a bit more clear, and broke it.  Working.
</blockquote></details>

Done.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18933)* status: :shipit: complete! all discussions resolved, LGTM from assignees trowell-tri,SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 186 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPZs-h77L1V2j-lZLls:-NPcOMR13HWWL1QPRIMX:b-tt90yt) ([raw file](https://github.com/RobotLocomotion/drake/blob/7f424d6aa6bd7b5402da1654db75ed2c3d9b77c1/bindings/pydrake/visualization/_model_visualizer.py#L186)):*

Done.
</blockquote></details>

Happy with this.

jwnimmer-tri:
*[`bindings/pydrake/visualization/_model_visualizer.py` line 405 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbuj1S4U_x4Bkif4b8:-NPcSZ5G9MLIJiQigyPk:bsbh2xn) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/_model_visualizer.py#L405)):*
> ... if the normal usecase doesn't provide any deprecation warning at runtime, then why deprecate?

The RunResult feature is unwanted and we need to delete it.  We could either do it as a breaking change, or via deprecation.

A sufficient condition for deprecation is announcing it in the release notes.  A runtime warning is nice to have when it's easy, but it's not easy in this case.

Your penultimate message was not about runtime warnings anyway.  It was about the docstring of this function.  Saying "This function returns a value but you should ignore it", to me, doesn't add any value and just one more thing to clean up later.

jwnimmer-tri:
> "A laptop! Are you mad?!?!" To which I'd say...no, just working with limited resources.

When developing with your laptop as the display, you should still be using your workstation for builds.  In this case it's easy too, because the tools in question (only the model visualizer in this cast) just need 1 http port -- your laptop webbrowser can connect to the visualizer that was built and running on your workstation.

jwnimmer-tri:
> I'm pretty sure I didn't reapply the magical cache encantation after updating my OS so, I could use a pointer to those instructions.

... here's my `~/.bazelrc`:

```
build --keep_going
build --jobs=HOST_CPUS*0.8

fetch --repository_cache /home/jwnimmer/.cache/bazel-externals
fetch --disk_cache /home/jwnimmer/.cache/bazel_local_disk
build --repository_cache /home/jwnimmer/.cache/bazel-externals
build --disk_cache /home/jwnimmer/.cache/bazel_local_disk

build --progress_report_interval=3
```

jwnimmer-tri:
*[`bindings/pydrake/visualization/_model_visualizer.py` line 405 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18933#-NPbuj1S4U_x4Bkif4b8:-NPcVJjJ9jcWLP3mo-Vt:bqm0ej8) ([raw file](https://github.com/RobotLocomotion/drake/blob/abc49b617032c84d5d4dbc30f0bc357792bc9ccd/bindings/pydrake/visualization/_model_visualizer.py#L405)):*
Re-reading the whole thread again now, I think I can better clarify what I meant here:

> My money is on my current way being the least confusing for users.

By "users", I meant the normal people who just want to (programmatically) invoke a view their models.  They do not care about the return value.  They never did, and still don't.  Making _them_ read an extra page of API text is to their detriment.

I was _not_ talking about users that were actively using the return value.  My money was on that set being empty.

SeanCurtis-TRI:
FYI I can't successfully run `model_visualizer` on my puget and connect my laptop browser to it.

The most relevant information I can offer up is that I'm running on VPN and have swapped the default `localhost` address with my machine's IP address. However, attempts to connect time out. Is port 7000 open to external connections? Is that a VPN tunneling thing, perhaps?

jwnimmer-tri:
You can open a slack thread and Rico and I can help debug.  Either your workstation has a firewall, or the VPN doesn't route between those endpoints.  If all else fails, ssh with port forwarding should work.

SeanCurtis-TRI:
Moving to slack...

rpoyner-tri:
Add this line to your ~/.ssh/config for your puget:
```
 LocalForward 7000 localhost:7000
```
and restart your ssh connection. Run the visualizer. Then point your browser to `localhost:7000`.

PR #18936, opened 2023-03-04
[solvers] Throw an warning if lower bound is larger than upper bound.

hongkai-dai:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)

### DIFF
--- a/multibody/inverse_kinematics/BUILD.bazel
+++ b/multibody/inverse_kinematics/BUILD.bazel
@@ -347,6 +347,7 @@ drake_cc_googletest(
     deps = [
         ":inverse_kinematics_core",
         ":inverse_kinematics_test_utilities",
+        "//common/test_utilities:expect_throws_message",
         "//math:geometric_transform",
         "//solvers:solve",
     ],
--- a/multibody/inverse_kinematics/test/inverse_kinematics_test.cc
+++ b/multibody/inverse_kinematics/test/inverse_kinematics_test.cc
@@ -4,6 +4,7 @@
 
 #include "drake/common/find_resource.h"
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/math/rotation_matrix.h"
 #include "drake/math/wrap_to.h"
 #include "drake/multibody/inverse_kinematics/test/inverse_kinematics_test_utilities.h"
@@ -107,10 +108,12 @@ GTEST_TEST(InverseKinematicsTest, ConstructorWithJointLimits) {
     EXPECT_TRUE(check_q_without_joint_limits(q_good));
     Eigen::VectorXd q_bad = q_good;
     q_bad(i) = -0.01 * lower_limits(i) + 1.01 * upper_limits(i);
-    EXPECT_FALSE(check_q_with_joint_limits(q_bad));
+    DRAKE_EXPECT_THROWS_MESSAGE(check_q_with_joint_limits(q_bad),
+                                "AggregateBoundingBoxConstraints.*");
     EXPECT_TRUE(check_q_without_joint_limits(q_bad));
     q_bad(i) = 1.01 * lower_limits(i) - 0.01 * upper_limits(i);
-    EXPECT_FALSE(check_q_with_joint_limits(q_bad));
+    DRAKE_EXPECT_THROWS_MESSAGE(check_q_with_joint_limits(q_bad),
+                                "AggregateBoundingBoxConstraints.*");
     EXPECT_TRUE(check_q_without_joint_limits(q_bad));
   }
 }
--- a/solvers/BUILD.bazel
+++ b/solvers/BUILD.bazel
@@ -1758,6 +1758,7 @@ drake_cc_googletest(
     deps = [
         ":aggregate_costs_constraints",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//common/test_utilities:symbolic_test_util",
     ],
 )
--- a/solvers/aggregate_costs_constraints.cc
+++ b/solvers/aggregate_costs_constraints.cc
@@ -179,6 +179,14 @@ std::unordered_map<symbolic::Variable, Bound> AggregateBoundingBoxConstraints(
       }
     }
   }
+  for (const auto& [var, bound] : bounds) {
+    if (bound.lower > bound.upper) {
+      throw std::invalid_argument(
+          fmt::format("AggregateBoundingBoxConstraints(): Variable {} has "
+                      "lower bound {}, larger than the upper bound {}",
+                      var.get_name(), bound.lower, bound.upper));
+    }
+  }
   return bounds;
 }
 
@@ -197,6 +205,14 @@ void AggregateBoundingBoxConstraints(const MathematicalProgram& prog,
       if (constraint.evaluator()->upper_bound()(i) < (*upper)(var_index)) {
         (*upper)(var_index) = constraint.evaluator()->upper_bound()(i);
       }
+    }
+  }
+  for (int i = 0; i < prog.num_vars(); ++i) {
+    if ((*lower)(i) > (*upper)(i)) {
+      throw std::invalid_argument(fmt::format(
+          "AggregateBoundingBoxConstraints(): Variable {} has lower bound {}, "
+          "larger than the upper bound {}",
+          prog.decision_variable(i).get_name(), (*lower)(i), (*upper)(i)));
     }
   }
 }
--- a/solvers/constraint.h
+++ b/solvers/constraint.h
@@ -12,6 +12,7 @@
 
 #include <Eigen/Core>
 #include <Eigen/SparseCore>
+#include <fmt/format.h>
 
 #include "drake/common/drake_assert.h"
 #include "drake/common/drake_copyable.h"
@@ -68,6 +69,14 @@ class Constraint : public EvaluatorBase {
     check(num_constraints);
     DRAKE_DEMAND(!lower_bound_.array().isNaN().any());
     DRAKE_DEMAND(!upper_bound_.array().isNaN().any());
+    for (int i = 0; i < lb.rows(); ++i) {
+      if (lb(i) > ub(i)) {
+        throw std::invalid_argument(fmt::format(
+            "Row {} of constraint {} has lower bound {}, larger than the upper "
+            "bound {}",
+            i, description, lb(i), ub(i)));
+      }
+    }
   }
 
   /**
--- a/solvers/test/aggregate_costs_constraints_test.cc
+++ b/solvers/test/aggregate_costs_constraints_test.cc
@@ -7,6 +7,7 @@
 #include <gtest/gtest.h>
 
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/common/test_utilities/symbolic_test_util.h"
 
 using ::testing::HasSubstr;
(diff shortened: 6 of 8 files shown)


hongkai-dai (commented):
+@aykut-tri for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-:-NPeYRr66iRiw__l47gu:b-w629gj)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: 1 unresolved discussion, LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPe_EUbD6aaIl7nbeVl:brjx5sn) (no related file):*
Why isn't this an exception?  Users doing optimization in the cloud might never see the spdlog console.

Except in unusual circumstances, malformed inputs should produces exceptions, not text logging.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: 1 unresolved discussion, LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPfILGhBrEW11AJ9ou9:b-mfbsiy) (no related file):*

Why isn't this an exception?  Users doing optimization in the cloud might never see the spdlog console.

Except in unusual circumstances, malformed inputs should produces exceptions, not text logging.
</blockquote></details>

Done. I changed to throw an error.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPooEz4-dI9z3BzWKhk:bivf6nf) (no related file):*

Done. I changed to throw an error.
</blockquote></details>

@jwnimmer-tri after we throw an error instead a warning, now we change the behavior of many functions, as you can see in the CI failure test. 

Previously we check a lot of conditions (for example in geometry/optimization, that whether a set belongs to another set) by forming an optimization problem, solving the problem, and checking if the problem is feasible. Now we cannot do that any more, as when we solve the problem it will throw an error, complaining that the lower bound is larger than the upper bound.

So we could change these failure test by catching the exception, but I wonder if there is a better way to do that without causing the downstream code to change behaviors.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPow94C86RlnaaFShHc:b-v4v7rw) (no related file):*

@jwnimmer-tri after we throw an error instead a warning, now we change the behavior of many functions, as you can see in the CI failure test. 

Previously we check a lot of conditions (for example in geometry/optimization, that whether a set belongs to another set) by forming an optimization problem, solving the problem, and checking if the problem is feasible. Now we cannot do that any more, as when we solve the problem it will throw an error, complaining that the lower bound is larger than the upper bound.

So we could change these failure test by catching the exception, but I wonder if there is a better way to do that without causing the downstream code to change behaviors.
</blockquote></details>

So let's step back... what problem was the warning trying to solve?  If we start from a description of a problem or scenario, it will probably lead us to better conclusions.

I'll assume that a user added vacuously false constraint to a program, and was confused about why the program was always infeasible...

Had the user read the `debug_mathematical_program`? If they hadn't read it, why didn't they? How can we advertise it more prominently? Should the MPResult output messages hyperlink to it somehow?

If they had read it, did they trying printing out the MP, or printing `GetInfeasibleConstraints`, or any of the other techniques in that tutorial?  If they tried those and it didn't help, what other debugging help can we give?  Where did those tips fall short?

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPp-moy35lo6LqMZnAD:bujak7v) (no related file):*
> I'll assume that a user added vacuously false constraint to a program, and was confused about why the program was always infeasible...

Exactly. @JoseBarreiros-TRI has a trajectory optimization problem, where he first constrains the initial and final state of the trajectory to a certain value. He then imposes bounding box constraint to say that all states should be within certain bound. Unfortunately his initial/final state falls outside of these bounds, and SNOPT just says `result.is_success() = False`. He then checked `result.get_solver_details<SnoptSolver>::info = 91`, but he doesn't quite know how to parse this info. So I would like to make the debugging message more informative.

In this case, it is extremely hard to print out the MP and identify the problem (the program is large, and it is just hard to eye-ball the incorrect lower/upper bound). If they do `GetInfeasibleConstraints` that won't help either. SNOPT just fails immediately when it sees that a constraint has lower bound larger than upper bound and directly returns the initial guess of the problem as the solution. This initial guess would essentially violates all constraints in the program (it is a bad initial guess).

So my thought is that it is really easy for the program to check if the lower bound is larger than the upper bound and then warn the user. We could throw an error in this case, but then that will impact the downstream code (this is a breaking change), and I would like to seek your opinion on the breaking change.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPpDoaN6sTv6kQjyFtV:bv8vlyf) (no related file):*

> I'll assume that a user added vacuously false constraint to a program, and was confused about why the program was always infeasible...

Exactly. @JoseBarreiros-TRI has a trajectory optimization problem, where he first constrains the initial and final state of the trajectory to a certain value. He then imposes bounding box constraint to say that all states should be within certain bound. Unfortunately his initial/final state falls outside of these bounds, and SNOPT just says `result.is_success() = False`. He then checked `result.get_solver_details<SnoptSolver>::info = 91`, but he doesn't quite know how to parse this info. So I would like to make the debugging message more informative.

In this case, it is extremely hard to print out the MP and identify the problem (the program is large, and it is just hard to eye-ball the incorrect lower/upper bound). If they do `GetInfeasibleConstraints` that won't help either. SNOPT just fails immediately when it sees that a constraint has lower bound larger than upper bound and directly returns the initial guess of the problem as the solution. This initial guess would essentially violates all constraints in the program (it is a bad initial guess).

So my thought is that it is really easy for the program to check if the lower bound is larger than the upper bound and then warn the user. We could throw an error in this case, but then that will impact the downstream code (this is a breaking change), and I would like to seek your opinion on the breaking change.
</blockquote></details>

If our debugging function that reports why a program failed doesn't actually highlight why it failed, then we need to improve the debugging function.  The message is appropriate when introspecting a on a MPResult, _not_ when creating a program and constraints.  As you note, it's valid and commonplace to create a vacuously-false constraint sometimes, or perhaps to update constraints bounds after-the-fact out of violation.  We should focus on debugging the  solution _result_ not logging during construction.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18936)* status: LGTM missing from assignee aykut-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @aykut-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/18936#-NPe_EUa6PCu8fCAAGAf:-NPpK3AACvNT0GbY40W4:b-1ednih) (no related file):*

If our debugging function that reports why a program failed doesn't actually highlight why it failed, then we need to improve the debugging function.  The message is appropriate when introspecting a on a MPResult, _not_ when creating a program and constraints.  As you note, it's valid and commonplace to create a vacuously-false constraint sometimes, or perhaps to update constraints bounds after-the-fact out of violation.  We should focus on debugging the  solution _result_ not logging during construction.
</blockquote></details>

OK, I will close this PR. My revised plan is to modify each solver to report more information for debugging.

So the question is what information to report. If the solve is not successful, I can check the bounds of each variable and constraints, and report if the lower bound is larger than the upper bound. Should I throw a warning then (I don't think you like throwing warning from the comment above)? But I do want to make the user to read some text message, such as "Variable x(0) has lower bound 3, larger than its upper bound 1". How can I pass these debugging info to the user without using `drake::log()->warn(...)`?

___

jwnimmer-tri:
Maybe you could open an issue with an sample program that highlights the complexity here, the issue being "it's too hard to debug why this seemingly-reasonable program isn't working".

One answer might be trajectory-optimization-specific sugar that makes this error impossible (an exception) in the first place, instead of trying to fix it at the lower-level solvers where things are more murky.

My next question is whether SNOPT's "Print file" output contained any hints about this failure.  If it did, can we update the tutorial to better highlight how to use that info?  Why didn't the user find it on his own?

Assuming that the SNOPT log doesn't help, then it's up to us to guide the user.

Here's what the tutorial currently shows to help debug this:

```
infeasible_constraints = result.GetInfeasibleConstraints(prog)
for c in infeasible_constraints:
    print(f"infeasible constraint: {c}")
```

I think your premise is that's not good enough? Maybe because too many of the constraints were infeasible, so the lower > upper detail on one of them gets lost in the noise? Or did the user not even try it?

The general thrust of my proposal would be to expand that section of the tutorial, adding more functions if need be.

The first point would be that in `GetInfeasibleConstraints`, if we have `lower > upper`, then we should use a different description string, something like adding scream caps "ALWAYS FALSE" into the message.  We could also sort these to the front of the list of returned strings (i.e., prioritize the advice).

If we find that's still not enough, we could imagine adding more kind of help strings to that vector, or even creating a second function like `std::string HelpMeDiagnoseUnsolveableProgram() const;` that returns whole paragraphs of advice and hyperlinks, particular to the program in question.

My bottom line is that:

(1) We use a tutorial-centric approach.  If the user isn't reading the existing tutorial that already helps them, fix that first.  Don't add more features when the users aren't even using the current ones.

(2) We use a result-centric approach.  During program setup and solve, we shouldn't pay the cost of trying to anticipate debugging questions; we should run the solve quickly, and without console spam.  If there is an _interactive user_ who is struggling, then they can call functions that help guide them through debugging, given a failed `MathematicalProgramResult`.  The tutorials will show them how.  We could even imagine that `to_string(MPResult)` contains 1 hyperlink, to the tutorial or a Doxygen overview that elucidates how to debug.

PR #18860, opened 2023-02-23
[C-IRIS] move cspace_free_polytope.h out of dev folder.

hongkai-dai:
I will move the cc file out of dev folder in my next PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18860)

PR #18938, opened 2023-03-05
[C-IRIS] move the code to search for separation certificate out of dev.

hongkai-dai:
This moves the function that search for the certificate for a fixed C-space polytope.


hongkai-dai (commented):
+@SeanCurtis-TRI for platform review on Monday please, thanks!

I move some code out from the `dev` folder. It has been feature reviewed already.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-:-NPjO9mA7PD0qwpSqbls:b-4zl96j)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
I didn't flag them all, but you have a number of aggressive uses of `auto`; I tagged a couple.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-:-NPrvJBV4Yq1dFdcC0Sy:b-4ke7eb)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 323 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPruzjR4YmaedXwsZhE:-NPruzjR4YmaedXwsZhF:b-v8atcg) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L323)):*
> ```cpp
>   // Add the constraint that the polynomial represented by this Gram and
>   // monomial basis is sos.
>   // @param is_zero_poly If true, then constrain all the Gram matrices to be
> ```

nit: There is no `is_zero_poly` param.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 440 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrfJ-j4oPxRogdl4jd:-NPrfJ-j4oPxRogdl4je:b-uau9l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L440)):*
> ```cpp
> }
> 
> CspaceFreePolytope::SeparatingPlaneLagrangians
> ```

This struct definition came through in an earlier PR (#18860). It's not really a good candidate for being a *struct*.

With reference to the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Structs_vs._Classes):

"The struct must not have invariants that imply relationships between different fields". Your lower and upper fields must have the same size.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPri_zG9StIkeE5Ni2Z:-NPri_zG9StIkeE5Ni2_:b-y7p9kk) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L441)):*
> ```cpp
> 
> CspaceFreePolytope::SeparatingPlaneLagrangians
> CspaceFreePolytope::SeparatingPlaneLagrangians::GetSolution(
> ```

BTW There's something *slightly* odd about this struct to me. It serves as the basis for both defining constraints on the mathematical program as well as storing the result.

Generally, that's not a problem. We have functions that consume doubles and return doubles as a matter of course. The difference for those functions is that the semantics is clearly documented. In this case, it's not clear from the documentation of the type that there is a distinction in semantics. And maybe there *isn't* a distinction. Hypothetically, I could create a `SeparatingPlaneLangrangians` instance, use it to configure constraints on a mathematical program, run the program, extract the result into another `SPL`, and then turn right around and use that *new* instance to apply constraints to a mathematical program. (Although, to my ignorant mind, that sounds incredibly silly to write).

If there is a defect here, the primary defect may be in the documentation (this particular function is undocumented). 
___
*[`geometry/optimization/cspace_free_polytope.cc` line 484 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrmZid70Pt0Z6p_2Gk:-NPrmZid70Pt0Z6p_2Gl:bk3rcuc) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L484)):*
> ```cpp
>     const int num_gram_vars_per_sos =
>         GetGramVarSize(monomial_basis_array, this->with_cross_y_, y_size);
>     gram_var_count += num_gram_vars_per_sos * num_sos;
> ```

BTW Burying this increment in the lambda gives something that *looks* like a calculation a side effect. If instead, you returned `num_gram_vars_per_sos * num_sos` then the call sites would change from:

```c++
count_gram(rational, monomial_basis_array_positive_side);
```

to

```c++
gram_var_count += count_gram(rational, monomial_basis_array_positive_side);
```

It would be much easier to recognize where in the function you are writing to `gram_var_count`.

Alternatively, passing it in as an output variable, would make it apparent that you're changing it. That would also allow this lambda and `add_rational_nonnegative` work the same vis a vis this variable.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrnYPP5Ys-9etjyGNe:-NPrnYPP5Ys-9etjyGNf:b-7v3gqe) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L488)):*
nit: According to the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Type_deduction):

> use type deduction [auto keyword] only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type.

I know the actual type of this is huge and obnoxious (which gives rise to the use of `auto`), but it is certainly not consistent with the guidance.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 748 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrsCZj5EhUNRBcP9iW:-NPrsCZj5EhUNRBcP9iX:b4ozl9c) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L748)):*
> ```cpp
>     if (options.verbose) {
>       std::string bad_pairs;
>       const auto& inspector = scene_graph_.model_inspector();
> ```

BTW Presumably, your geometry configuration will not change in the actual `SceneGraph` context?
___
*[`geometry/optimization/cspace_free_polytope.cc` line 790 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrsr6R58-DhYkLNxqS:-NPrsr6R58-DhYkLNxqT:b-s1tlms) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L790)):*
> ```cpp
>       &s_lower_redundant_indices, &s_upper_redundant_indices);
> 
>   const auto certificates_vec = this->FindSeparationCertificateGivenPolytope(
> ```

BTW: Another case where the use of `auto` is too aggressive.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 440 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrfJ-j4oPxRogdl4jd:-NPudw8O72FZRIJtGG_-:b-ldkorw) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L440)):*

This struct definition came through in an earlier PR (#18860). It's not really a good candidate for being a *struct*.

With reference to the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Structs_vs._Classes):

"The struct must not have invariants that imply relationships between different fields". Your lower and upper fields must have the same size.
</blockquote></details>

Done. I changed this to a class.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPri_zG9StIkeE5Ni2Z:-NPudy2k3yVwCAbsqX32:b-auhxs7) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L441)):*

BTW There's something *slightly* odd about this struct to me. It serves as the basis for both defining constraints on the mathematical program as well as storing the result.

Generally, that's not a problem. We have functions that consume doubles and return doubles as a matter of course. The difference for those functions is that the semantics is clearly documented. In this case, it's not clear from the documentation of the type that there is a distinction in semantics. And maybe there *isn't* a distinction. Hypothetically, I could create a `SeparatingPlaneLangrangians` instance, use it to configure constraints on a mathematical program, run the program, extract the result into another `SPL`, and then turn right around and use that *new* instance to apply constraints to a mathematical program. (Although, to my ignorant mind, that sounds incredibly silly to write).

If there is a defect here, the primary defect may be in the documentation (this particular function is undocumented). 
</blockquote></details>

I added the documentation. Is it better now?
___
*[`geometry/optimization/cspace_free_polytope.cc` line 484 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrmZid70Pt0Z6p_2Gk:-NPuea1f6AH8mMI65fHN:b-ebg8b) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L484)):*

BTW Burying this increment in the lambda gives something that *looks* like a calculation a side effect. If instead, you returned `num_gram_vars_per_sos * num_sos` then the call sites would change from:

```c++
count_gram(rational, monomial_basis_array_positive_side);
```

to

```c++
gram_var_count += count_gram(rational, monomial_basis_array_positive_side);
```

It would be much easier to recognize where in the function you are writing to `gram_var_count`.

Alternatively, passing it in as an output variable, would make it apparent that you're changing it. That would also allow this lambda and `add_rational_nonnegative` work the same vis a vis this variable.
</blockquote></details>

Done. Good call, it does look better now.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 748 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrsCZj5EhUNRBcP9iW:-NPusZwXABw7Mi11KQ2J:bywvthz) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L748)):*

BTW Presumably, your geometry configuration will not change in the actual `SceneGraph` context?
</blockquote></details>

Sorry I don't understand the question, what is "geometry configuration"? Do you mean the pose of each geometry::Shape? Or do you mean the parameter (radius, length) of each geometry::Shape?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.h` line 172 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPwkHqA1xfzI65hZx4M:-NPwkHqA1xfzI65hZx4N:b-4t1sh9) ([raw file](https://github.com/RobotLocomotion/drake/blob/37f9f4cebf98ade51137438437e165b01da654bd/geometry/optimization/cspace_free_polytope.h#L172)):*
> ```objc
> 
>     /// The Lagrangians for d - C*s >= 0.
>     VectorX<symbolic::Polynomial>& get_mutable_polytope() { return polytope_; }
> ```

BTW If you wanted, you could cut `get_` from the name just to make things shorter.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPri_zG9StIkeE5Ni2Z:-NPwjbmQ7F1Q3Wt-mCBY:bd8w6x7) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L441)):*

I added the documentation. Is it better now?
</blockquote></details>

I'm not seeing documentation.

In contrast, I notice you have `SeparationCertificateResult` and `SeparationCertificate`.  Admittedly, those are different of necessity (completely different quantities).
___
*[`geometry/optimization/cspace_free_polytope.cc` line 748 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrsCZj5EhUNRBcP9iW:-NPwl8268yti8W28RPiU:b-t9bqai) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L748)):*
So, geometry can be registered and configured on `SceneGraph`. Once a context is created, the context's version of the universe can be tweaked. This code ties this function ties `CspaceFreePolytope` to `SceneGraph`'s model and must ignore any per-context tweaks.

If that's the design choice, it's probably worth documenting that in the constructor. Something along the lines of:

> CspaceFreePolytope knows nothing about Contexts. The plant and scene_graph must be fully configured before instantiating. Any changes made in contexts subsequently will have no effect.

(I believe this mostly concerns `SceneGraph` as `MbP` can't change its toplogy after finalization and these polytopes don't care about the dynamics properties.)
___
*[`geometry/optimization/cspace_free_polytope.cc` line 474 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPwkdCA-1_ExKNMDIZ3:-NPwkdCA-1_ExKNMDIZ4:b43l824) ([raw file](https://github.com/RobotLocomotion/drake/blob/37f9f4cebf98ade51137438437e165b01da654bd/geometry/optimization/cspace_free_polytope.cc#L474)):*
> ```cpp
>                         const symbolic::RationalFunction& rational,
>                         const std::array<VectorX<symbolic::Monomial>, 4>&
>                             monomial_basis_array) -> int {
> ```

BTW I shouldn't think you would need this. There's a single return statement which is unambiguously the product of two ints. C++ should be able to 100% infer the correct return value. Nothing necessarily wrong with being explicit, I just wanted to make sure you were actively making that decision.  :)
___
*[`geometry/optimization/cspace_free_polytope.cc` line 536 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPwir8y8VbrsYgRZ98o:-NPwir8y8VbrsYgRZ98p:b-94dlun) ([raw file](https://github.com/RobotLocomotion/drake/blob/37f9f4cebf98ade51137438437e165b01da654bd/geometry/optimization/cspace_free_polytope.cc#L536)):*
> ```cpp
>         gram_and_monomial_basis.AddSos(
>             prog, gram_vars.segment(gram_var_count, num_gram_vars_per_sos),
>             &(lagrangians.get_mutable_s_lower()(j)));
> ```

nit: This new spelling has confused our linter. It now parses this as a cast. You don't *need* the parentheses.

```c++
&lagrangians.get_mutable_polytope()(j)
```

is fine. [Operator precedence](https://en.cppreference.com/w/cpp/language/operator_precedence) takes care of this for you. Your expression consists of function calls and member access, both of which have precedence over `&`. So, I'd simply throw out the parentheses and walk away.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPri_zG9StIkeE5Ni2Z:-NPzb0YyAnRqN0OyxJPi:b4ynwzl) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L441)):*

I'm not seeing documentation.

In contrast, I notice you have `SeparationCertificateResult` and `SeparationCertificate`.  Admittedly, those are different of necessity (completely different quantities).
</blockquote></details>

Done. I added the documentation before and somehow reset the modification in my last commit.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 748 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPrsCZj5EhUNRBcP9iW:-NPzbiwDBEgg2hDNvjUx:bsjaxwv) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7ec752aee3eb6bea1e7741cdf03c46955b6e944/geometry/optimization/cspace_free_polytope.cc#L748)):*

So, geometry can be registered and configured on `SceneGraph`. Once a context is created, the context's version of the universe can be tweaked. This code ties this function ties `CspaceFreePolytope` to `SceneGraph`'s model and must ignore any per-context tweaks.

If that's the design choice, it's probably worth documenting that in the constructor. Something along the lines of:

> CspaceFreePolytope knows nothing about Contexts. The plant and scene_graph must be fully configured before instantiating. Any changes made in contexts subsequently will have no effect.

(I believe this mostly concerns `SceneGraph` as `MbP` can't change its toplogy after finalization and these polytopes don't care about the dynamics properties.)
</blockquote></details>

Done. I added the documentation to the constructor of CspaceFreePolytope.
___
*[`geometry/optimization/cspace_free_polytope.cc` line 536 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPwir8y8VbrsYgRZ98o:-NPzbCJzBWceORfa-uMj:bor38ld) ([raw file](https://github.com/RobotLocomotion/drake/blob/37f9f4cebf98ade51137438437e165b01da654bd/geometry/optimization/cspace_free_polytope.cc#L536)):*

nit: This new spelling has confused our linter. It now parses this as a cast. You don't *need* the parentheses.

```c++
&lagrangians.get_mutable_polytope()(j)
```

is fine. [Operator precedence](https://en.cppreference.com/w/cpp/language/operator_precedence) takes care of this for you. Your expression consists of function calls and member access, both of which have precedence over `&`. So, I'd simply throw out the parentheses and walk away.
</blockquote></details>

Done. Thanks for the explanation. I always forgot the operator precedence and then just put the parentheses.

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-:-NQ1RMcP6eO6p7JcWPCu:bmxilld)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope.cc` line 536 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NPwir8y8VbrsYgRZ98o:-NQ1RBQNBj0MUoxdHSjd:bxpx2ck) ([raw file](https://github.com/RobotLocomotion/drake/blob/37f9f4cebf98ade51137438437e165b01da654bd/geometry/optimization/cspace_free_polytope.cc#L536)):*

Done. Thanks for the explanation. I always forgot the operator precedence and then just put the parentheses.
</blockquote></details>

And that would *generally* be fine...but for our linter that isn't 100% up on C++ :)

SeanCurtis-TRI (commented):
+(release notes: none) 
+(status: single reviewer ok)  (only platform required).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-:-NQ1RcAr1_izpxxYmx2U:bnuxen3)* status: :shipit: complete! all discussions resolved, LGTM from assignee SeanCurtis-TRI(platform) (waiting on @hongkai-dai)

jwnimmer-tri:
*[`geometry/optimization/cspace_free_polytope.h` line 40 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18938#-NUTGOLi6I74KC_kp36A:-NUTGOLi6I74KC_kp36B:bsei5cw) ([raw file](https://github.com/RobotLocomotion/drake/blob/85d862d26792834aa65596347944d99beeac2a0f/geometry/optimization/cspace_free_polytope.h#L40)):*
> ```objc
>   const std::vector<symbolic::RationalFunction> positive_side_rationals;
>   const std::vector<symbolic::RationalFunction> negative_side_rationals;
>   int plane_index;
> ```

nit Missing required GSG initialization



_Suggestion:_
```objc
int plane_index{};
```

PR #18941, opened 2023-03-05
[yaml] Opt-in to clang-format linting

jwnimmer-tri:
Towards #15774.

+@rpoyner-tri for both reviews, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18941)

PR #18943, opened 2023-03-05
[yaml] Add json output

jwnimmer-tri:
Requires #18941.

Towards #15774.  When shelling out to the python downloader, it will be run using the system Python which might not have any libraries, so to pass data back and forth we need to use JSON, not YAML.

### DIFF
--- a/common/yaml/BUILD.bazel
+++ b/common/yaml/BUILD.bazel
@@ -166,4 +166,4 @@ drake_cc_googletest(
     ],
 )
 
-add_lint_tests()
+add_lint_tests(enable_clang_format_lint = True)
--- a/common/yaml/BUILD.bazel
+++ b/common/yaml/BUILD.bazel
@@ -94,6 +94,15 @@ drake_cc_library(
 )
 
 drake_cc_googletest(
+    name = "json_test",
+    deps = [
+        ":example_structs",
+        ":yaml_io",
+        "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
     name = "yaml_doxygen_test",
     deps = [
         ":yaml_io",
--- a/common/yaml/test/example_structs.h
+++ b/common/yaml/test/example_structs.h
@@ -87,9 +87,7 @@ struct ArrayStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  ArrayStruct() {
-    value.fill(kNominalDouble);
-  }
+  ArrayStruct() { value.fill(kNominalDouble); }
 
   explicit ArrayStruct(const std::array<double, 3>& value_in)
       : value(value_in) {}
@@ -103,9 +101,7 @@ struct VectorStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  VectorStruct() {
-    value.resize(1, kNominalDouble);
-  }
+  VectorStruct() { value.resize(1, kNominalDouble); }
 
   explicit VectorStruct(const std::vector<double>& value_in)
       : value(value_in) {}
@@ -119,9 +115,7 @@ struct NonPodVectorStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  NonPodVectorStruct() {
-    value.resize(1, {"kNominalDouble"});
-  }
+  NonPodVectorStruct() { value.resize(1, {"kNominalDouble"}); }
 
   std::vector<StringStruct> value;
 };
@@ -132,9 +126,7 @@ struct MapStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  MapStruct() {
-    value["kNominalDouble"] = kNominalDouble;
-  }
+  MapStruct() { value["kNominalDouble"] = kNominalDouble; }
 
   explicit MapStruct(const std::map<std::string, double>& value_in)
       : value(value_in) {}
@@ -148,9 +140,7 @@ struct UnorderedMapStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  UnorderedMapStruct() {
-    value["kNominalDouble"] = kNominalDouble;
-  }
+  UnorderedMapStruct() { value["kNominalDouble"] = kNominalDouble; }
 
   explicit UnorderedMapStruct(
       const std::unordered_map<std::string, double>& value_in)
@@ -165,9 +155,7 @@ struct OptionalStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  OptionalStruct() {
-    value = kNominalDouble;
-  }
+  OptionalStruct() { value = kNominalDouble; }
 
   explicit OptionalStruct(const double value_in)
       : OptionalStruct(std::optional<double>(value_in)) {}
@@ -184,8 +172,7 @@ struct OptionalStructNoDefault {
     a->Visit(DRAKE_NVP(value));
   }
 
-  OptionalStructNoDefault()
-      : value(std::nullopt) {}
+  OptionalStructNoDefault() : value(std::nullopt) {}
 
   explicit OptionalStructNoDefault(const double value_in)
       : OptionalStructNoDefault(std::optional<double>(value_in)) {}
@@ -239,8 +226,8 @@ using EigenMatrix00Struct = EigenStruct<0, 0>;
 using EigenMatrixUpTo6Struct =
     EigenStruct<Eigen::Dynamic, Eigen::Dynamic, 6, 6>;
 
-using Variant4 = std::variant<
-    std::string, double, DoubleStruct, EigenVecStruct>;
+using Variant4 =
+    std::variant<std::string, double, DoubleStruct, EigenVecStruct>;
 
 std::ostream& operator<<(std::ostream& os, const Variant4& value) {
   if (value.index() == 0) {
@@ -261,12 +248,9 @@ struct VariantStruct {
     a->Visit(DRAKE_NVP(value));
   }
 
-  VariantStruct() {
-    value = kNominalDouble;
-  }
+  VariantStruct() { value = kNominalDouble; }
 
-  explicit VariantStruct(const Variant4& value_in)
-      : value(value_in) {}
+  explicit VariantStruct(const Variant4& value_in) : value(value_in) {}
 
   Variant4 value;
 };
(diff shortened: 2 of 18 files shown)


jwnimmer-tri:
Possibly @rpoyner-tri could feature-review this, but I've also been peppering him with a lot of reviews lately.  Other people (@xuchenhan-tri @sammy-tri @ggould-tri) might be interested in taking a feature review.

Whoever self-assigns first can take it.

ggould-tri (commented):
+@ggould-tri I've a bit of time now, I'll look in to this.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-:-NPshhqD9OveRxETFAWU:b-ymqc0u)* status: LGTM missing from assignee ggould-tri(platform), needs at least two assigned reviewers, missing label for release notes

ggould-tri (commented):
:lgtm:; modest notes below.

> Reviewed 3 of 9 files at r2, 11 of 12 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-:-NPszGhUEPTty8xKGaFz:b-hl6axq)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPsi2po5LHQsBd9WTGh):*
minor: Assuming +(release notes: feature), this could be a little bit more explicit for the release engineer.



_Suggestion:_
```
[yaml] Yaml-serializable objects can also be output as JSON
```
___
*[`common/yaml/yaml_io.h` line 124 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPswVD66TQLT9hHoW2n:-NPswVD66TQLT9hHoW2o:bbgc7ti) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_io.h#L124)):*
> ```objc
> 
> Note that there is no LoadJsonFile() function, because LoadYamlString() can
> already load JSON data.
> ```

This statement is not reflected in the test cases.

I believe it in general, but with respect to the non-finite values I think there may be some format shear.  Can you provide a test showing that json non-finites load as yaml non-finites?  This could be a trivial round-trip test, assert `x == LoadYamlString(SaveJsonString(x))` for a well-constructed x.

(There are also some other stupid and pedantic bits of shear around field names, but I don't see any reason to mention those in doxygen or to test them; reference https://metacpan.org/pod/JSON::XS#JSON-and-YAML if you wish to purchase a ticket to that particular flamewar)



_Code quote:_
```objc
Note that there is no LoadJsonFile() function, because LoadYamlString() can
already load JSON data.
```
___
*[`common/yaml/yaml_read_archive.h` line 285 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPskPP26GcvfZ8F8I-r:-NPskPP26GcvfZ8F8I-s:bgkn5bh) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_read_archive.h#L285)):*
> ```objc
>     }
>     // Figure out which variant<...> type we have based on the node's tag.
>     const std::string_view tag = sub_node->GetTag();
> ```

nit:  You burned away the `const` in every other case but this one.
Arguably this conforms to our style wrt const value args, so author's discretion.



_Code quote:_
```objc
const 
```

jwnimmer-tri (commented):
+@xuchenhan-tri for platform review (Wednesday schedule), please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-:-NPt7hDx8w3bMD_J7O-I:b-1xty0k)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform) (waiting on @ggould-tri and @xuchenhan-tri)

___
*[`common/yaml/yaml_io.h` line 124 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPswVD66TQLT9hHoW2n:-NPt7YfL7XyLI4Wk-fhY:b-gopv3c) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_io.h#L124)):*

This statement is not reflected in the test cases.

I believe it in general, but with respect to the non-finite values I think there may be some format shear.  Can you provide a test showing that json non-finites load as yaml non-finites?  This could be a trivial round-trip test, assert `x == LoadYamlString(SaveJsonString(x))` for a well-constructed x.

(There are also some other stupid and pedantic bits of shear around field names, but I don't see any reason to mention those in doxygen or to test them; reference https://metacpan.org/pod/JSON::XS#JSON-and-YAML if you wish to purchase a ticket to that particular flamewar)
</blockquote></details>

This is a good point.  I took the easy way out.  I'll leave this thread open for you to resolve if you're OK with it.
___
*[`common/yaml/yaml_read_archive.h` line 285 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPskPP26GcvfZ8F8I-r:-NPt6B8ME3BqR-vV6W0I:b-6tkxls) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_read_archive.h#L285)):*

nit:  You burned away the `const` in every other case but this one.
Arguably this conforms to our style wrt const value args, so author's discretion.
</blockquote></details>

OK I'm content with "const all locals" but "don't const arguments" as a convention for these related stanzas.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform) (waiting on @ggould-tri and @xuchenhan-tri)

___
*[`common/yaml/yaml_io.h` line 124 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPswVD66TQLT9hHoW2n:-NPtMCQQ1jAdb33iqloC:bid39j1) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_io.h#L124)):*

This is a good point.  I took the easy way out.  I'll leave this thread open for you to resolve if you're OK with it.
</blockquote></details>

(I think that JSON rant is a little bit out of date now.  The unicode string encoding and "/" escaping have since been fixed, at least as of yaml 1.2.2.  To my first reading, the only remaining incompatibility is that 1024-char upper bound on map key strings.)

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`common/yaml/yaml_io.h` line 124 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPswVD66TQLT9hHoW2n:-NPwIalKD8MQZcYy4EIz:bxfz9pg) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d8551f1dd4932b4a1c62b2f98fa8b8af2629b8/common/yaml/yaml_io.h#L124)):*

(I think that JSON rant is a little bit out of date now.  The unicode string encoding and "/" escaping have since been fixed, at least as of yaml 1.2.2.  To my first reading, the only remaining incompatibility is that 1024-char upper bound on map key strings.)
</blockquote></details>

:+1:  Onward to Norway!

xuchenhan-tri (commented):
Platform :lgtm: 

> Reviewed 3 of 9 files at r2, 10 of 12 files at r3, 4 of 4 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-:-NPxf_aW0-BOw6wMcH6i:b-cxwtpw)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`common/yaml/yaml_write_archive.cc` line 207 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPxfXKq6NdowQrdKbrq:-NPxfXKq6NdowQrdKbrr:b-9h4656) ([raw file](https://github.com/RobotLocomotion/drake/blob/72c9c1f1ba851a07684d7be922f8fd7a1cf82b49/common/yaml/yaml_write_archive.cc#L207)):*
> ```cpp
>         // a sequence. We'll still fail-fast in case it ever does.
>         throw std::logic_error(fmt::format(
>             "SaveJsonString: Cannot save a YAML sequence with tag '{}' to JSON",
> ```

nit, this is covered by unit tests.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943)* status: 1 unresolved discussion (waiting on @xuchenhan-tri)

___
*[`common/yaml/yaml_write_archive.cc` line 207 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPxfXKq6NdowQrdKbrq:-NPxgmEhDYhk0kGhZwAt:b-au7bvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/72c9c1f1ba851a07684d7be922f8fd7a1cf82b49/common/yaml/yaml_write_archive.cc#L207)):*

nit, this is covered by unit tests.
</blockquote></details>

That's right.  The comment immediately above was my attempt to caution the read about what.  WDYT?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18943)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri)

___
*[`common/yaml/yaml_write_archive.cc` line 207 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18943#-NPxfXKq6NdowQrdKbrq:-NPxweswA61dmLZuv8vL:bb1fbuz) ([raw file](https://github.com/RobotLocomotion/drake/blob/72c9c1f1ba851a07684d7be922f8fd7a1cf82b49/common/yaml/yaml_write_archive.cc#L207)):*

That's right.  The comment immediately above was my attempt to caution the read about what.  WDYT?
</blockquote></details>

I see. I misinterpreted "can never" as "should never". I think this is fine.

Retracted.

Issue #18951, opened 2023-03-07
Ball joint should report the correct error when there is an axis


marcoag:
### What happened?

As dicussed [here](https://github.com/RobotLocomotion/drake/pull/18462#pullrequestreview-1320367196), when parsing ball joints, there is no warning recorded if there is an axis. Instead of that warning, the parser goes ahead and throws a [limits related warning] that doesn't seem very relevant for a ball joint that shouldn't have an axis (see [here](https://github.com/RobotLocomotion/drake/blob/153f3fb18b8c2f5203ffb0531e96af75610c5040/multibody/parsing/detail_sdf_parser.cc#L464)).

We should correct this and add a warning similar to the one that is there for [axis2 in ball joints.](https://github.com/RobotLocomotion/drake/blob/153f3fb18b8c2f5203ffb0531e96af75610c5040/multibody/parsing/detail_sdf_parser.cc#L472)

### What operating system are you using?

Ubuntu 22.04

### What installation option are you using?

compiled from source code using Bazel

rpoyner-tri:
@marcoag you appear to have the context. Are you willing to take this?

marcoag:
Please do. I will follow up with a PR on this.

scpeters:
Here's my thoughts:

Every physical mechanism has friction of some kind, including ball joints. In SDFormat we currently store this information in the `//joint/axis` and `//joint/axis2` elements, which allows different friction parameters for each axis of a multi-axis joint. It is a little awkward for ball joints, to use the `//axis` element because `//axis/xyz` is required but doesn't have meaning for a ball joint. I think the simplest thing would be to just store the information in `//joint/axis/dynamics/friction` for a ball joint and ignore the `//joint/axis/xyz` element, or explicitly set it to `0 0 0`.

With this approach, the `//joint/axis/dynamics` tag would be relevant for ball joints, and the `//joint/axis/xyz` will just be there since it's a required element, but it could be ignored or set to `0 0 0`.

In terms of interpreting these values, it's important to be clear about it. I would interpret the friction and damping values in `//axis/dynamics` as opposing any relative angular motion. I think I would ignore the spring stiffness and reference parameters.

sherm1:
Thanks, Steve! 

That seems like a workable hack, but is clearly a bug in the sdf definition ("axis" makes NO sense for a ball joint!). Is there a way to get this fixed in a future sdf spec update?

jwnimmer-tri:
[...]

PR #18966, opened 2023-03-09
Fix warning when ball joint has an axis

marcoag:
Closes https://github.com/RobotLocomotion/drake/issues/18951.

Ball joints should have no axis and if there is one the sdf parser will ignore it.

This PR corrects the warning the user gets when the sdf parser finds an axis in a ball joint. It also adds fixes to the relevant tests and removes irrelevant code related to parameters of a ball joint axis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966)

### DIFF
--- a/multibody/parsing/detail_sdf_parser.cc
+++ b/multibody/parsing/detail_sdf_parser.cc
@@ -406,15 +406,22 @@ void AddJointActuatorFromSpecification(
                joint_spec.Type() == sdf::JointType::REVOLUTE ||
                joint_spec.Type() == sdf::JointType::CONTINUOUS);
 
-  // Ball joints cannot specify an axis (nor actuation) per SDFormat. However,
-  // Drake still permits the first axis in order to specify damping, but it
-  // should not have actuation, nor should there be a second axis.
+  // Ball joints do not have an axis (nor actuation). However, Drake still
+  // permits the declaration of a first axis in order to specify damping, but
+  // it should not have actuation, nor should there be a second axis.
   if (joint_spec.Type() == sdf::JointType::BALL) {
     if (joint_spec.Axis(0) != nullptr) {
       std::string message = fmt::format(
-          "An axis may not be specified for ball joint '{}' and will be "
-          "ignored", joint_spec.Name());
+          "A ball joint axis will be ignored. Only the dynamic parameters"
+          " and limits will be considered.", joint_spec.Name());
       diagnostic.Warning(joint_spec.Element(), std::move(message));
+      if (GetEffortLimit(diagnostic, joint_spec, 0) != 0) {
+        std::string effort_message = fmt::format(
+            "Actuation (via non-zero effort limits) for ball joint '{}' is"
+            " not implemented yet and will be ignored.",
+            joint_spec.Name());
+        diagnostic.Warning(joint_spec.Element(), std::move(effort_message));
+      }
     }
     if (joint_spec.Axis(1) != nullptr) {
       std::string message = fmt::format(
--- a/multibody/parsing/detail_sdf_parser.cc
+++ b/multibody/parsing/detail_sdf_parser.cc
@@ -411,13 +411,10 @@ void AddJointActuatorFromSpecification(
   // should not have actuation, nor should there be a second axis.
   if (joint_spec.Type() == sdf::JointType::BALL) {
     if (joint_spec.Axis(0) != nullptr) {
-      if (GetEffortLimit(diagnostic, joint_spec, 0) != 0) {
-        std::string message = fmt::format(
-            "Actuation (via non-zero effort limits) for ball joint '{}' is"
-            " not implemented yet and will be ignored.",
-            joint_spec.Name());
-        diagnostic.Warning(joint_spec.Element(), std::move(message));
-      }
+      std::string message = fmt::format(
+          "An axis may not be specified for ball joint '{}' and will be "
+          "ignored", joint_spec.Name());
+      diagnostic.Warning(joint_spec.Element(), std::move(message));
     }
     if (joint_spec.Axis(1) != nullptr) {
       std::string message = fmt::format(
--- a/multibody/parsing/test/detail_sdf_parser_test.cc
+++ b/multibody/parsing/test/detail_sdf_parser_test.cc
@@ -754,18 +754,11 @@ TEST_F(SdfParserTest, BallJointWithAxis2Error) {
   <joint name='should_not_have_axis' type='ball'>
     <parent>a</parent>
     <child>b</child>
-    <axis>
-      <xyz>0 0 1</xyz>
-    </axis>
     <axis2>
       <xyz>0 0 1</xyz>
     </axis2>
   </joint>
 </model>)""");
-  EXPECT_THAT(TakeWarning(), ::testing::MatchesRegex(
-      R"(.*Actuation \(via non-zero effort limits\) for ball joint )"
-      R"('should_not_have_axis' is not implemented yet and will be )"
-      R"(ignored.*)"));
   EXPECT_THAT(TakeWarning(), ::testing::MatchesRegex(
       ".*An axis2 may not be specified for ball joint 'should_not_have_axis' "
       "and will be ignored.*"));
@@ -1130,7 +1123,6 @@ TEST_F(SdfParserTest, JointParsingTest) {
   EXPECT_EQ(ball_joint.name(), "ball_joint");
   EXPECT_EQ(ball_joint.parent_body().name(), "link4");
   EXPECT_EQ(ball_joint.child_body().name(), "link5");
-  EXPECT_EQ(ball_joint.damping(), 0.1);
   const Vector3d inf3(std::numeric_limits<double>::infinity(),
                       std::numeric_limits<double>::infinity(),
                       std::numeric_limits<double>::infinity());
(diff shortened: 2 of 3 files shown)


EricCousineau-TRI (commented):
+@EricCousineau-TRI :lgtm: feature, thanks!
+@rpoyner-tri for platform review, please!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NQ6lBXw1pvcngepKObz:b-cyb6sm)* status: LGTM missing from assignee rpoyner-tri(platform), missing label for release notes (waiting on @rpoyner-tri)

marcoag:
Sorry for the confusion here but it seems that axis are still needed for ball joints. They are indeed ignored but some of the parameters specified through them are still relevant, i.e. [damping](https://github.com/RobotLocomotion/drake/blob/b3aea181f2a7bb9775d0aa42b3250ecea792d762/multibody/parsing/detail_sdf_parser.cc#L410). 

I removed all the previous changes and added a warning that clarifies the axis will be ignored but the parameters will be used. I also clarified a bit the comment that explain axis on ball joint.

Because this seems to be a confusing thing from the SDFormat library I opened an issue [here](https://github.com/gazebosim/sdformat/issues/1250) in order to discuss some further clarifications over there.

EricCousineau-TRI (commented):
Sweet, thanks! I have a question about when this warning will get issues
@rpoyner-tri Can I ask if my question makes sense (and if you have any inklings on the answer)?

> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NV6oTKh558bu4cXdDaA:b9u8c7j)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

___
*[`multibody/parsing/detail_sdf_parser.cc` line 413 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6o83W-j7eMba0g5Ga:-NV6o83W-j7eMba0g5Gb:bv8upbn) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/detail_sdf_parser.cc#L413)):*
> ```cpp
>   // it should not have actuation, nor should there be a second axis.
>   if (joint_spec.Type() == sdf::JointType::BALL) {
>     if (joint_spec.Axis(0) != nullptr) {
> ```

I'm not sure if I understand when this warning will trigger, and if it will be superfluous.

If a user specifies damping / dynamic parameters, but does not specify `<xyz>`, does the warning get triggered?
If so, is there any way we could ignore it?

e.g., only emit a warning if the axis value itself is explicitly specified? If we don't have access to that, can we at least suppress the warning if it's a default value?
___
*[`multibody/parsing/test/detail_sdf_parser_test.cc` line 1153 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6nry6-LNhWOiQ2Ghq:-NV6nry6-LNhWOiQ2Ghr:b-5pqy05) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/test/detail_sdf_parser_test.cc#L1153)):*
> ```cpp
>       " parameters and limits will be considered.*"));
>   FlushDiagnostics();
> ```

BTW Extra newline?

rpoyner-tri (commented):
@EricCousineau-TRI I'm not sure exactly what you are asking. The unit test constructs cases where the warnings are issued. Like any other parser warnings, they don't cause the parse to fail, and could be ignored if you need to ignore them. If the concern is console spam, I don't have a good sense of how many ball joints with damping are out there in the world.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NVBD-S39t4VBYlNpXZb:b92mz9b)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

jwnimmer-tri:
I took Eric's comment to ask whether or not this SDFormat would have any warnings.  I didn't immediately find any test case on point for it:

```xml
  <joint name="jerry" type="ball">
    <parent>world</parent>
    <child>larry</child>
    <axis>
      <limit>
        <effort>0</effort>
      </limit>
    </axis>
  </joint>
```

Note that it's unactuated, with no xyz values given for any axes.  I think what we want is no warnings in this case.

rpoyner-tri (commented):
Ok then. So that looks like a unit test that needs to get written: given the model shown above, demonstrate no ball axis warnings are issued.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NVBQug415vd-51Ea3G3:bz1g3ul)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

EricCousineau-TRI (commented):
Sorry for not being clearer, but exactly! Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NVBZpMU9pY4MAqqwY03:b-mhifpv)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

___
*[`multibody/parsing/detail_sdf_parser.cc` line 413 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6o83W-j7eMba0g5Ga:-NVBZtGmDVYdZnUGmEi2:bwro6ga) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/detail_sdf_parser.cc#L413)):*

I'm not sure if I understand when this warning will trigger, and if it will be superfluous.

If a user specifies damping / dynamic parameters, but does not specify `<xyz>`, does the warning get triggered?
If so, is there any way we could ignore it?

e.g., only emit a warning if the axis value itself is explicitly specified? If we don't have access to that, can we at least suppress the warning if it's a default value?
</blockquote></details>

Macro - would you be able to add in the test that Jeremy mentioned?

marcoag (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

___
*[`multibody/parsing/detail_sdf_parser.cc` line 413 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6o83W-j7eMba0g5Ga:-NVdAEjSAGONALuVil2u:b7blb24) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/detail_sdf_parser.cc#L413)):*

Macro - would you be able to add in the test that Jeremy mentioned?
</blockquote></details>

The thing is that according to the SDFormat specification`<xyz>` is required for any `<joint>` `<axis>`: http://sdformat.org/spec?ver=1.9&elem=joint#axis_xyz. So technically you would not be able to specify damping /dynamic parameters without an`<xyz>`. Anytime you have an `<axis>` the warning would be there.
___
*[`multibody/parsing/test/detail_sdf_parser_test.cc` line 1153 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6nry6-LNhWOiQ2Ghq:-NVEUdpFC2Vcwn8K76ED:bss4a2b) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/test/detail_sdf_parser_test.cc#L1153)):*

BTW Extra newline?
</blockquote></details>

Removed.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

___
*[`multibody/parsing/detail_sdf_parser.cc` line 413 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6o83W-j7eMba0g5Ga:-NVduWYq1cUdyjmE5ydr:b-z6z3gq) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/detail_sdf_parser.cc#L413)):*
<details><summary><i>Previously, marcoag (Marco A. Gutierrez) wrote…</i></summary><blockquote>

The thing is that according to the SDFormat specification`<xyz>` is required for any `<joint>` `<axis>`: http://sdformat.org/spec?ver=1.9&elem=joint#axis_xyz. So technically you would not be able to specify damping /dynamic parameters without an`<xyz>`. Anytime you have an `<axis>` the warning would be there.
</blockquote></details>

For the record, I agree it's a moot point and the code is OK as-is.

EricCousineau-TRI (commented):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @marcoag)

___
*[`multibody/parsing/detail_sdf_parser.cc` line 413 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-NV6o83W-j7eMba0g5Ga:-NVjPsoFE4IwuygVLa0_:b-r1ktqm) ([raw file](https://github.com/RobotLocomotion/drake/blob/973d14b8e3f97a6110150f240851f2ac3f52e5cd/multibody/parsing/detail_sdf_parser.cc#L413)):*

For the record, I agree it's a moot point and the code is OK as-is.
</blockquote></details>

OK Ah, that makes sense - thanks for clarifying!

rpoyner-tri (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18966#-:-NWDJhPN-a6yn8CWqChc:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),EricCousineau-TRI(platform)

PR #18967, opened 2023-03-09
[parsing] Update processing of AddWeld to try registering both (parent, child) and (child, parent)

EricCousineau-TRI:
One candidate path towards discussion in  https://github.com/robotlocomotion/drake/pull/17223#pullrequestreview-1333038327

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18967)


EricCousineau-TRI:
Per VC this week and two weeks ago, this doesn't make sense as part of #17223 because it is out of scope (we should only try to convert what is presently supported)

PR #18987, opened 2023-03-13
[perception] PointCloud can morph its fields after construction

zachfang:
As discussed in https://github.com/RobotLocomotion/drake/pull/18774#issuecomment-1437232982, this PR modified the code and relevant documentation to allow morphing fields after the construction of a PointCloud.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)

### DIFF
--- a/bindings/pydrake/perception_py.cc
+++ b/bindings/pydrake/perception_py.cc
@@ -119,6 +119,8 @@ void init_perception(py::module m) {
               self->SetFrom(other);
             },
             py::arg("other"), cls_doc.SetFrom.doc)
+        .def("SetFields", &Class::SetFields, py::arg("new_fields"),
+            py::arg("skip_initialize") = false, cls_doc.SetFields.doc)
         .def("Crop", &Class::Crop, py::arg("lower_xyz"), py::arg("upper_xyz"),
             cls_doc.Crop.doc)
         .def("FlipNormalsTowardPoint", &Class::FlipNormalsTowardPoint,
--- a/bindings/pydrake/test/perception_test.py
+++ b/bindings/pydrake/test/perception_test.py
@@ -77,6 +77,22 @@ class TestPerception(unittest.TestCase):
         pc.rgbs()
         pc.mutable_rgb(i=0)
         pc.rgb(i=0)
+        # Test morphing fields after PointCloud creation.
+        rgb_pc = mut.PointCloud(new_size=count,
+                                fields=mut.Fields(mut.BaseField.kRGBs))
+        all_fields_pc = mut.PointCloud(new_size=count, fields=all_fields)
+        test_rgbs = [[1, 2, 3]]
+        all_fields_pc.mutable_rgbs().T[:] = test_rgbs
+        self.assertFalse((rgb_pc.rgbs().T == test_rgbs).all())
+        self.assertFalse(rgb_pc.has_xyzs())
+        rgb_pc = all_fields_pc
+        self.assertTrue((rgb_pc.rgbs().T == test_rgbs).all())
+        self.assertTrue(rgb_pc.has_xyzs())
+        self.assertTrue(rgb_pc.has_normals())
+        rgb_pc.SetFields(new_fields=mut.Fields(mut.BaseField.kNormals),
+                         skip_initialize=False)
+        self.assertFalse(rgb_pc.has_rgbs())
+        self.assertTrue(rgb_pc.has_normals())
         # - Check for none.
         with self.assertRaises(RuntimeError) as ex:
             mut.PointCloud(new_size=0, fields=mut.Fields(mut.BaseField.kNone))
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -25,9 +25,6 @@ from drake import (
 from pydrake.common import (
     FindResourceOrThrow,
 )
-from pydrake.common.value import (
-    Value,
-)
 from pydrake.geometry import (
     DrakeVisualizer,
     DrakeVisualizerParams,
@@ -119,8 +116,7 @@ class TestMeldis(unittest.TestCase):
         # Set input and publish the point cloud.
         context = diagram.CreateDefaultContext()
         cloud_context = cloud_to_lcm.GetMyContextFromRoot(context)
-        cloud_to_lcm.get_input_port().FixValue(
-            cloud_context, Value[PointCloud](cloud))
+        cloud_to_lcm.get_input_port().FixValue(cloud_context, cloud)
         diagram.ForcedPublish(context)
 
     def test_viewer_applet(self):
--- a/perception/point_cloud.cc
+++ b/perception/point_cloud.cc
@@ -83,7 +83,10 @@ class PointCloud::Storage {
     normals_.conservativeResize(NoChange,
                                 f.contains(pc_flags::kNormals) ? size_ : 0);
     rgbs_.conservativeResize(NoChange, f.contains(pc_flags::kRGBs) ? size_ : 0);
-    descriptors_.conservativeResize(NoChange, f.has_descriptor() ? size_ : 0);
+    // Note: The row size can change depends on whether 'f' contains a
+    // descriptor field and the type of the descriptor.
+    descriptors_.conservativeResize(f.descriptor_type().size(),
+                                    f.has_descriptor() ? size_ : 0);
     fields_ = f;
     CheckInvariants();
   }
@@ -95,21 +98,32 @@ class PointCloud::Storage {
 
  private:
   void CheckInvariants() const {
+    const int xyz_size = xyzs_.cols();
     if (fields_.contains(pc_flags::kXYZs)) {
-      const int xyz_size = xyzs_.cols();
       DRAKE_DEMAND(xyz_size == size());
+    } else {
+      DRAKE_DEMAND(xyz_size == 0);
     }
+    const int normals_size = normals_.cols();
     if (fields_.contains(pc_flags::kNormals)) {
-      const int normals_size = normals_.cols();
       DRAKE_DEMAND(normals_size == size());
+    } else {
+      DRAKE_DEMAND(normals_size == 0);
     }
+    const int rgbs_size = rgbs_.cols();
     if (fields_.contains(pc_flags::kRGBs)) {
-      const int rgbs_size = rgbs_.cols();
       DRAKE_DEMAND(rgbs_size == size());
+    } else {
+      DRAKE_DEMAND(rgbs_size == 0);
     }
+    const int descriptor_cols = descriptors_.cols();
+    const int descriptor_rows = descriptors_.rows();
     if (fields_.has_descriptor()) {
-      const int descriptor_size = descriptors_.cols();
-      DRAKE_DEMAND(descriptor_size == size());
+      DRAKE_DEMAND(descriptor_cols == size());
+      DRAKE_DEMAND(descriptor_rows == fields_.descriptor_type().size());
+    } else {
+      DRAKE_DEMAND(descriptor_cols == 0);
+      DRAKE_DEMAND(descriptor_rows == 0);
     }
   }
 
@@ -132,23 +146,21 @@ pc_flags::Fields ResolveFields(
   }
 }
 
-// Resolves the fields from a pair of point clouds and desired fields.
-// Implements the resolution rules in `SetFrom`.
-// @pre Valid point clouds `a` and `b`.
-// @returns Fields that both point clouds have.
-pc_flags::Fields ResolvePairFields(
-    const PointCloud& a,
-    const PointCloud& b,
-    pc_flags::Fields fields) {
-  if (fields == pc_flags::kInherit) {
-    // If we do not permit a subset, expect the exact same fields.
-    a.RequireExactFields(b.fields());
-    return a.fields();
-  } else {
-    a.RequireFields(fields);
-    b.RequireFields(fields);
-    return fields;
+// Finds the added fields from `new_fields` w.r.t. `old_fields`.
+pc_flags::Fields FindAddedFields(pc_flags::Fields old_fields,
+                                 pc_flags::Fields new_fields) {
+  pc_flags::Fields added_fields(pc_flags::kNone);
+  for (const auto field :
+       {pc_flags::kXYZs, pc_flags::kNormals, pc_flags::kRGBs}) {
+    if (!old_fields.contains(field) && new_fields.contains(field))
+      added_fields |= field;
   }
+
+  if (new_fields.has_descriptor() &&
+      old_fields.descriptor_type() != new_fields.descriptor_type()) {
+    added_fields |= new_fields.descriptor_type();
+  }
+  return added_fields;
 }
 
 }  // namespace
@@ -172,7 +184,7 @@ PointCloud::PointCloud(const PointCloud& other,
 }
 
 PointCloud::PointCloud(PointCloud&& other)
-    : PointCloud(0, other.fields(), true) {
+    : PointCloud(0, other.storage_->fields(), true) {
   // This has zero size. Directly swap storages.
   storage_.swap(other.storage_);
 }
@@ -183,8 +195,6 @@ PointCloud& PointCloud::operator=(const PointCloud& other) {
 }
 
 PointCloud& PointCloud::operator=(PointCloud&& other) {
-  // We may only take rvalue references if the fields match exactly.
-  RequireExactFields(other.fields());
   // Swap storages.
   storage_.swap(other.storage_);
   // Empty out the other cloud, but let it remain being a valid point cloud
@@ -207,11 +217,34 @@ int PointCloud::size() const {
 void PointCloud::resize(int new_size, bool skip_initialization) {
   DRAKE_DEMAND(new_size >= 0);
   const int old_size = size();
+  if (old_size == new_size)
+    return;
   storage_->resize(new_size);
   DRAKE_DEMAND(storage_->size() == new_size);
   if (new_size > old_size && !skip_initialization) {
     const int size_diff = new_size - old_size;
     SetDefault(old_size, size_diff);
+  }
+}
+
+void PointCloud::SetFields(pc_flags::Fields new_fields, bool skip_initialize) {
+  const pc_flags::Fields old_fields = storage_->fields();
+  if (old_fields == new_fields)
+    return;
+  storage_->UpdateFields(new_fields);
+
+  if (!skip_initialize) {
+    // Default-initialize containers for newly added fields.
+    const pc_flags::Fields added_fields =
+        FindAddedFields(old_fields, new_fields);
+    if (added_fields.contains(pc_flags::kXYZs))
+      mutable_xyzs().setConstant(kDefaultValue);
+    if (added_fields.contains(pc_flags::kNormals))
+      mutable_normals().setConstant(kDefaultValue);
+    if (added_fields.contains(pc_flags::kRGBs))
+      mutable_rgbs().setConstant(kDefaultColor);
+    if (added_fields.has_descriptor())
+      mutable_descriptors().setConstant(kDefaultValue);
   }
 }
 
@@ -236,6 +269,7 @@ void PointCloud::SetDefault(int start, int num) {
 void PointCloud::SetFrom(const PointCloud& other,
                          pc_flags::Fields fields_in,
                          bool allow_resize) {
+  // Update the size of this point cloud if necessary.
   int old_size = size();
   int new_size = other.size();
   if (allow_resize) {
@@ -244,18 +278,28 @@ void PointCloud::SetFrom(const PointCloud& other,
     throw std::runtime_error(
         fmt::format("SetFrom: {} != {}", new_size, old_size));
   }
-  pc_flags::Fields fields_resolved =
-      ResolvePairFields(*this, other, fields_in);
-  if (fields_resolved.contains(pc_flags::kXYZs)) {
+
+  // Update or check the fields of the point cloud(s) if necessary.
+  pc_flags::Fields fields_to_copy = fields_in;
+  if (fields_in == pc_flags::kInherit) {
+    fields_to_copy = other.storage_->fields();
+    SetFields(other.storage_->fields(), true);
+  } else {
+    this->RequireFields(fields_to_copy);
+    other.RequireFields(fields_to_copy);
+  }
+
+  // Populate data from `other` to this point cloud.
+  if (fields_to_copy.contains(pc_flags::kXYZs)) {
     mutable_xyzs() = other.xyzs();
   }
-  if (fields_resolved.contains(pc_flags::kNormals)) {
+  if (fields_to_copy.contains(pc_flags::kNormals)) {
     mutable_normals() = other.normals();
   }
-  if (fields_resolved.contains(pc_flags::kRGBs)) {
+  if (fields_to_copy.contains(pc_flags::kRGBs)) {
     mutable_rgbs() = other.rgbs();
   }
-  if (fields_resolved.has_descriptor()) {
+  if (fields_to_copy.has_descriptor()) {
     mutable_descriptors() = other.descriptors();
   }
 }
--- a/perception/point_cloud.h
+++ b/perception/point_cloud.h
@@ -22,10 +22,6 @@ namespace perception {
 /// This is a mix between the philosophy of PCL (templated interface to
 /// provide a compile-time open set, run-time closed set) and VTK (non-templated
 /// interface to provide a very free form run-time open set).
-/// You may construct one PointCloud which will contain different sets of
-/// data, but you cannot change the contained data types after construction.
-/// However, you can mutate the data contained within the structure and resize
-/// the cloud.
 ///
 /// Definitions:
 ///
@@ -42,7 +38,7 @@ namespace perception {
 ///
 /// @note "contiguous" here means contiguous in memory. This was chosen to
 /// avoid ambiguity between PCL and Eigen, where in PCL "dense" implies that
-/// the point cloud corresponds to a cloud with invalid values, and in Eigen
+/// the point cloud corresponds to a cloud with only valid values, and in Eigen
 /// "dense" implies contiguous storage.
 ///
 /// @note The accessors / mutators for the point fields of this class returns
@@ -90,7 +86,7 @@ class PointCloud final {
   /// @param fields
   ///   Fields that the point cloud contains.
   /// @param skip_initialize
-  ///    Do not default-initialize new values.
+  ///   Do not default-initialize new values.
   explicit PointCloud(int new_size = 0,
                       pc_flags::Fields fields = pc_flags::kXYZs,
                       bool skip_initialize = false);
@@ -253,9 +249,9 @@ class PointCloud final {
   /// @param other
   ///    Other point cloud.
   /// @param fields_in
-  ///    Fields to copy. If this is `kInherit`, then both clouds must have the
-  ///    exact same fields. Otherwise, both clouds must support the fields
-  ///    indicated this parameter.
+  ///    Fields to copy. If this is `kInherit`, then `other`s fields will be
+  ///    copied. Otherwise, both clouds must support the fields indicated this
+  ///    parameter.
   /// @param allow_resize
   ///    Permit resizing to the other cloud's size.
   void SetFrom(
@@ -276,6 +272,16 @@ class PointCloud final {
 
   /// @name Fields
   /// @{
+
+  /// Updates the point cloud to a given set of fields. In the case of
+  /// introducing a new field, its container will be allocated with the current
+  /// size and default initialized. The data for all retained fields will remain
+  /// unchanged.
+  /// @param new_fields
+  ///    New fields to set to.
+  /// @param skip_initialize
+  ///    Do not default-initialize new values.
+  void SetFields(pc_flags::Fields new_fields, bool skip_initialize = false);
 
   /// Returns if a point cloud has a given set of fields.
   bool HasFields(pc_flags::Fields fields_in) const;
--- a/perception/test/point_cloud_test.cc
+++ b/perception/test/point_cloud_test.cc
@@ -11,7 +11,7 @@
 #include "drake/common/test_utilities/expect_no_throw.h"
 
 using Eigen::Matrix3Xf;
-using Eigen::Matrix4Xf;
+using Eigen::RowVectorXf;
 using Eigen::Vector3f;
 
 using ::testing::AssertionResult;
@@ -55,6 +55,24 @@ struct check_helper<uint8_t> {
   }
 };
 
+void CompareClouds(const PointCloud& cloud_1, const PointCloud& cloud_2) {
+  EXPECT_TRUE(cloud_1.fields() == cloud_2.fields());
+
+  pc_flags::Fields fields_to_compare = cloud_1.fields();
+  if (fields_to_compare.contains(pc_flags::kXYZs)) {
+    EXPECT_TRUE(CompareMatrices(cloud_1.xyzs(), cloud_2.xyzs()));
+  }
+  if (fields_to_compare.contains(pc_flags::kRGBs)) {
+    EXPECT_TRUE((cloud_1.rgbs().array() == cloud_2.rgbs().array()).all());
+  }
+  if (fields_to_compare.contains(pc_flags::kNormals)) {
+    EXPECT_TRUE(CompareMatrices(cloud_1.normals(), cloud_2.normals()));
+  }
+  if (fields_to_compare.has_descriptor()) {
+    EXPECT_TRUE(CompareMatrices(cloud_1.descriptors(), cloud_2.descriptors()));
+  }
+}
+
 GTEST_TEST(PointCloudTest, TestExpectedNumThreads) {
 #if defined(_OPENMP)
   constexpr bool has_openmp = true;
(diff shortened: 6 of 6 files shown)


zachfang (commented):
+@jwnimmer-tri for the feature review, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-:-NQRh7iq-aPEJvprUziW:b4t3349)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri:
Remind me -- is there / was there a plan to change meldis or meldis tests to take advantage of this new capability now?  And if so, should that come in this PR or else can you link to another patch where it's going to happen later?

jwnimmer-tri (commented):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRsXSK8Z_YC1n-UeXg:-NQRsXSK8Z_YC1n-UeXh:b-ooxwmd) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*
> ```objc
> 
>   /// Updates fields for the point cloud and its underlying storage.
>   void UpdateFields(pc_flags::Fields new_fields);
> ```

The verb "update" does not communicate any semantics, other than the basic premise of something changing.  We should choose a specific verb that communicates what we're actually doing here.

If it's simply setting the fields to a new value, we should call it `SetFields`.

If we're adding new fields but leaving existing fields alone, we should call it `AddFields`.

Either way, we need to specify what happens to the data when new fields are introduced -- are they set to zeros? or ones?  or left as uninitialized memory?
___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRtLzx6N2gz15xDTFz:-NQRtLzx6N2gz15xDTG-:b-qbv5nr) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*
> ```objc
> 
>   /// Updates fields for the point cloud and its underlying storage.
>   void UpdateFields(pc_flags::Fields new_fields);
> ```

Could you explain the reasoning for this being in the Doxygen group for "Fields" vs the group for "Container Manipulation"?
___
*[`perception/point_cloud.cc` line 200 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRuLQ88zAja9_4nzqT:-NQRuLQ88zAja9_4nzqU:b-ekwk3s) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.cc#L200)):*
> ```cpp
>   fields_ = new_fields;
>   storage_->UpdateFields(fields_);
>   DRAKE_DEMAND(storage_->fields() == fields_);
> ```

I am not buying the complexity of storing both `size_` and `fields_` in two places.

Is there a reason why `size_` and `fields_` need to live on the outer class?  My intuition is that they would be just fine as fields of the Storage only, and the other class could ask the storage for its details.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.cc` line 200 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRuLQ88zAja9_4nzqT:-NQS5fLM9460o0scKqf0:bynjidq) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.cc#L200)):*

I am not buying the complexity of storing both `size_` and `fields_` in two places.

Is there a reason why `size_` and `fields_` need to live on the outer class?  My intuition is that they would be just fine as fields of the Storage only, and the other class could ask the storage for its details.
</blockquote></details>

My proposal is that we land this first:

https://github.com/RobotLocomotion/drake/pull/18989

zachfang (commented):
I was using that as a local test myself, and I have brought the meldis change into this PR.

I also rebased this PR on top of your deduplication PR (I don't think there will be significant changes) to get it to a working state. Will rebase again once it's merged.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-:-NQStZ117BtAuh-UwVrg:bgt1dou)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRtLzx6N2gz15xDTFz:-NQSrqtd6f0OtcLfPIAE:b68m2bn) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*

Could you explain the reasoning for this being in the Doxygen group for "Fields" vs the group for "Container Manipulation"?
</blockquote></details>

My motivation was that the function is updating PoinCloud's fields, and it would be easier to see what operations the class provides regarding the fields in one place.
___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRsXSK8Z_YC1n-UeXg:-NQSsmMKEksTD9SBRFa-:b-rdf24i) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*

The verb "update" does not communicate any semantics, other than the basic premise of something changing.  We should choose a specific verb that communicates what we're actually doing here.

If it's simply setting the fields to a new value, we should call it `SetFields`.

If we're adding new fields but leaving existing fields alone, we should call it `AddFields`.

Either way, we need to specify what happens to the data when new fields are introduced -- are they set to zeros? or ones?  or left as uninitialized memory?
</blockquote></details>

Done. I think `SetFields` makes the most sense. Updated the comment to clarify how the underlying container(s) is handled.

Along the line, should I change `Storage::UpdateFields()` as well? The PointCloud function just wraps around it so that people can morph the fields of the PointCloud.
___
*[`perception/point_cloud.cc` line 200 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRuLQ88zAja9_4nzqT:-NQSrXtR4_caEdELJG4n:bbamka2) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.cc#L200)):*

My proposal is that we land this first:

https://github.com/RobotLocomotion/drake/pull/18989
</blockquote></details>

Done. I did find it harder to follow with two sets of metadata.

jwnimmer-tri (commented):
> Reviewed 2 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.h` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRuqRPA4bcI1QBqerI:-NQRuqRQ66_Esctbw4ai:b-10hlf7) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L27)):*
> ```objc
> /// You may construct one PointCloud with different sets of data, mutate the
> /// data contained within the structure, resize the cloud, and change the
> /// contained data types after the construction.
> ```

nit This incremental evolution of the overview text has led is into a word salad that only makes sense in retrospect.  I don't think a first-time reader of this last sentence would understand what we're trying to say.  (I don't really, either.)

My vote is to just purge the last sentence entirely.

If you think it's valuable, that's fine but you'll need to rewrite it to communicate whatever it's trying to say more effectively.
___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRtLzx6N2gz15xDTFz:-NQVfbUU6Bna_ChPnOki:bgme3d8) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*

My motivation was that the function is updating PoinCloud's fields, and it would be easier to see what operations the class provides regarding the fields in one place.
</blockquote></details>

Yeah, it's kind of a mess already and this doesn't really make it that much worse.

(In "Container Manipulation" we have `Expand` but `resize` if off in no-man's land on its own.  It probably makes sense to rethink the grouping, but we can defer that if you like.)
___
*[`perception/point_cloud.h` line 282 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRsXSK8Z_YC1n-UeXg:-NQVhPrS6qFzCfJU7jPt:bq8q1x1) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L282)):*
> Along the line, should I change Storage::UpdateFields() as well? The PointCloud function just wraps around it so that people can morph the fields of the PointCloud.

If it does the same thing then it should use the same name.

Hypothetically if UpdateFields left new channels uninitialize and then SetFields zeros them out, that would be a reason for them to have different names.
___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVf11C8sBSDLg1NEeP:-NQVf11C8sBSDLg1NEeQ:b-t2qfif) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*
> ```objc
>   /// Updates the point cloud to a given set of fields. In the case of
>   /// introducing a new field, its container will be allocated with the current
>   /// size but not initialized.
> ```

Giving the user garbage memory is a too-dangerous default.

If we need that as an option for performance, then we need to provide it as opt-in, e.g., like we did in the constructor:

```c++
void SetFields(
    pc_flags::Fields new_fields,
    bool skip_initialize = false);
```

If we don't need the extra performance, than we can omit the extra flag and just zero the new data unconditionally instead.
___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVj5csEQUUamshOThG:-NQVj5csEQUUamshOThH:b-q3zg6d) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*
nit WDYT about being extra clear for the user in this doc, with an extra sentence at the end?

> Updates the point cloud to a given set of fields. In the case of introducing a new field, its container will be allocated with the current size but not initialized.  **The data for all retained fields will remain unchanged.**
___
*[`perception/point_cloud.h` line 282 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVhyzcEoRwt-NuwDSs:-NQVhyzdBHbkqkKK4fRG:b-fqwcfn) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L282)):*
> ```objc
>   /// introducing a new field, its container will be allocated with the current
>   /// size but not initialized.
>   void SetFields(pc_flags::Fields new_fields);
> ```

nit Could you add pydrake bindings for the new function?  I bet we'll forget unless we do it now.  (If you'd like to do it in another PR that's fine too, just make sure you have a reminder set somewhere.)
___
*[`perception/point_cloud.cc` line 227 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVkmWD0vRU8elcnL5t:-NQVkmWD0vRU8elcnL5u:b-pek41f) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.cc#L227)):*
> ```cpp
>   // Update or check the fields of the point cloud(s) if necessary.
>   if (fields_in == pc_flags::kInherit) {
>     if (storage_->fields() != other.storage_->fields()) {
> ```

nit Guarding a call to "SetFields" with a no-op check at _every call site_ will lead to bloat.

The `SetFields` function itself should be smart enough to bail out out in case the command is a no-op (assuming that doing so is important for runtime performance).
___
*[`perception/point_cloud.cc` line 249 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVk648DB8rBaUlwnT0:-NQVk649AYtwQYMZU6cA:bgrxn07) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.cc#L249)):*
> ```cpp
>   // Populate data from `other` to this point cloud.
>   pc_flags::Fields fields_resolved =
>       (fields_in == pc_flags::kInherit) ? other.storage_->fields() : fields_in;
> ```

nit For my money, `fields_resolved` looks like a consequence of the first if-else block atop this function, and for cohesiveness should be reflected there instead of separated down here.  They two need to agree.

Suggestion for atop:

```c++
  // Update or check the fields of the point cloud(s) if necessary.
  pc_flags::Fields fields_to_copy = fields_in;
  if (fields_in == pc_flags::kInherit) {
    fields_to_copy = other.storage_->fields();
    SetFields(fields_to_copy);
  } else {
    this->RequireFields(fields_to_copy);
    other.RequireFields(fields_to_copy);
  }
```

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @zachfang)

___
*[`perception/point_cloud.h` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQRuqRPA4bcI1QBqerI:-NQahkxGDKYQOaIggF9O:b-gehoa0) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fff1ecfe5c4195b5c710e228f1cd52e515d3655/perception/point_cloud.h#L27)):*

nit This incremental evolution of the overview text has led is into a word salad that only makes sense in retrospect.  I don't think a first-time reader of this last sentence would understand what we're trying to say.  (I don't really, either.)

My vote is to just purge the last sentence entirely.

If you think it's valuable, that's fine but you'll need to rewrite it to communicate whatever it's trying to say more effectively.
</blockquote></details>

Done. I think the paragraph was stressing the immutable property of a PointCloud which is no longer true. I agree with deleting it.
___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVf11C8sBSDLg1NEeP:-NQai4HD1WnsdK0kJsPQ:bhjv2qb) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*

Giving the user garbage memory is a too-dangerous default.

If we need that as an option for performance, then we need to provide it as opt-in, e.g., like we did in the constructor:

```c++
void SetFields(
    pc_flags::Fields new_fields,
    bool skip_initialize = false);
```

If we don't need the extra performance, than we can omit the extra flag and just zero the new data unconditionally instead.
</blockquote></details>

Working. Setting default values when resizing is easier because we expect all the existing containers will undergo the same operation. I am thinking a way to detect newly added fields (by bit-operation rather than many `if-else`s) so that we can default-initialize only the subset.  
___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVj5csEQUUamshOThG:-NQaiy3y8uUSWui5547f:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*

nit WDYT about being extra clear for the user in this doc, with an extra sentence at the end?

> Updates the point cloud to a given set of fields. In the case of introducing a new field, its container will be allocated with the current size but not initialized.  **The data for all retained fields will remain unchanged.**

</blockquote></details>

Done.
___
*[`perception/point_cloud.h` line 282 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVhyzcEoRwt-NuwDSs:-NQaizHEEWOdqjDkjfn6:b5w8abb) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L282)):*

nit Could you add pydrake bindings for the new function?  I bet we'll forget unless we do it now.  (If you'd like to do it in another PR that's fine too, just make sure you have a reminder set somewhere.)
</blockquote></details>

Done. Added `SetFields()` with some basic testing.
___
*[`perception/point_cloud.cc` line 227 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVkmWD0vRU8elcnL5t:-NQaj5Q2AOviddBagpX6:b-tehk3z) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.cc#L227)):*

nit Guarding a call to "SetFields" with a no-op check at _every call site_ will lead to bloat.

The `SetFields` function itself should be smart enough to bail out out in case the command is a no-op (assuming that doing so is important for runtime performance).
</blockquote></details>

Done. I did the same for `resize()` if it makes sense.
___
*[`perception/point_cloud.cc` line 249 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVk648DB8rBaUlwnT0:-NQajA1qAq3Jw3uVM4ZC:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.cc#L249)):*

nit For my money, `fields_resolved` looks like a consequence of the first if-else block atop this function, and for cohesiveness should be reflected there instead of separated down here.  They two need to agree.

Suggestion for atop:

```c++
  // Update or check the fields of the point cloud(s) if necessary.
  pc_flags::Fields fields_to_copy = fields_in;
  if (fields_in == pc_flags::kInherit) {
    fields_to_copy = other.storage_->fields();
    SetFields(fields_to_copy);
  } else {
    this->RequireFields(fields_to_copy);
    other.RequireFields(fields_to_copy);
  }
```

</blockquote></details>

Done.

jwnimmer-tri (commented):
> Reviewed 3 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVf11C8sBSDLg1NEeP:-NQaqXAq-wPqPqL8-k52:b-tevbx8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*

Working. Setting default values when resizing is easier because we expect all the existing containers will undergo the same operation. I am thinking a way to detect newly added fields (by bit-operation rather than many `if-else`s) so that we can default-initialize only the subset.  
</blockquote></details>

I think doing the work in the inner class would be easy?

Instead of calling, `xyzs_.conservativeResize(NoChange, f.contains(pc_flags::kXYZs) ? size_ : 0);`, call a little templated helper function that handles all three cases:

(1) this flag is unchanged -- noop
(2) this flag is gone -- resize to 0
(3) this flag is new -- resize to and then setConstant.

?
___
*[`bindings/pydrake/test/perception_test.py` line 80 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQaq-u66zYWfJC9ff52:-NQaq-u66zYWfJC9ff53:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5474bd21b395088be647a6136c31305d14818119/bindings/pydrake/test/perception_test.py#L80)):*
> ```Python
>         pc.mutable_rgb(i=0)
>         pc.rgb(i=0)
>         # Test morphying fields after PointCloud creation.
> ```

typo



_Suggestion:_
```Python
morphing
```
___
*[`bindings/pydrake/test/perception_test.py` line 92 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQaq2KL5Sg9JNM75M68:-NQaq2KM9J0Sr4PiKbtN:by45eqe) ([raw file](https://github.com/RobotLocomotion/drake/blob/5474bd21b395088be647a6136c31305d14818119/bindings/pydrake/test/perception_test.py#L92)):*
> ```Python
>         self.assertTrue(rgb_pc.has_xyzs())
>         self.assertTrue(rgb_pc.has_normals())
>         rgb_pc.SetFields(mut.Fields(mut.BaseField.kNormals))
> ```

nit When testing a binding, always use the kwargs (`new_fields=...`) spelling, and always pass all kwargs (including optional ones) to prove that the `py::arg()` details in the binding are well-formed.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

___
*[`perception/point_cloud.h` line 281 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQVf11C8sBSDLg1NEeP:-NQfgY0T3qjrOM0uhCXu:b-lvf9c9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b9850291360d73c7331331e4f23f1d4185271f0/perception/point_cloud.h#L281)):*

I think doing the work in the inner class would be easy?

Instead of calling, `xyzs_.conservativeResize(NoChange, f.contains(pc_flags::kXYZs) ? size_ : 0);`, call a little templated helper function that handles all three cases:

(1) this flag is unchanged -- noop
(2) this flag is gone -- resize to 0
(3) this flag is new -- resize to and then setConstant.

?
</blockquote></details>

So this is what I have at the moment after thinking hard about how to add the initialization piece.

I ended up creating a helper function to detect the newly-introduced field and just initialize containers based on that. My reasons are:
(1) We seem to do conservativeResize() operations in the inner class and all initialization operations in PointCloud. Thus, in both `resize()` and `SetFields()`, we follow the same pattern to first allocate and then initialize if desired.
(2) I was thinking about the templated helper, as suggested. What baffled me is the different function calls to check whether a `Fields` has a field, i.e., `f.contains()` vs. `f.has_descriptor()`.

This leads to a helper function to do the diff and the initialization after. Let me know what you think. 
___
*[`bindings/pydrake/test/perception_test.py` line 92 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQaq2KL5Sg9JNM75M68:-NQfgVil1yM87ILMg83T:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5474bd21b395088be647a6136c31305d14818119/bindings/pydrake/test/perception_test.py#L92)):*

nit When testing a binding, always use the kwargs (`new_fields=...`) spelling, and always pass all kwargs (including optional ones) to prove that the `py::arg()` details in the binding are well-formed.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> Reviewed 1 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.cc` line 138 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQfl_gQ0Y86imQtYVr4:-NQfl_gQ0Y86imQtYVr5:b-k3wja2) ([raw file](https://github.com/RobotLocomotion/drake/blob/815b224c3049e0ff22e52502ef84b8724adb1b52/perception/point_cloud.cc#L138)):*
> ```cpp
> }
> 
> // Finds different fields introduced by `new_fields` w.r.t. `old_fields`.
> ```

"Different" vs "Added" are two distinct things.  I think we do want "added" as our predicate (like we have within the function body here), so it's the function name and comment which are wrong.
___
*[`perception/point_cloud.cc` line 148 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQfmYfBD7dhRozup1AT:-NQfmYfCDUy3s-SPC3jY:b-xq3dft) ([raw file](https://github.com/RobotLocomotion/drake/blob/815b224c3049e0ff22e52502ef84b8724adb1b52/perception/point_cloud.cc#L148)):*
> ```cpp
>   }
> 
>   if (old_fields.descriptor_type() != new_fields.descriptor_type()) {
> ```

Wow, this check is correct but completely confusing to read.  Calling `!=` typically means "differs" not "difference", but given how `kDescriptorNone` works the code is correct.

To me ...

```
(new_fields.has_descriptor() && 
 (old_fields.descriptor_type() != new_fields.descriptor_type()))
```

... would be a ton more readable.  Then, the reader doesn't need to go figure out what `|` does here.

(Overloading `|` for `DescriptorType` was a GSG violation in the first place.  It does not have bitset semantics!  It has assignment semantics!  But that ship has sailed.)

jwnimmer-tri (commented):
:lgtm: feature.

> Reviewed 1 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-:-NQk1shR3qoc2txbOfRK:b-n37teh)* status: 6 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.cc` line 86 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk0wMAEVCImGAGUB6q:-NQk0wMB6_uCPafdABpC:b-322mek) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L86)):*
> ```cpp
>                                 f.contains(pc_flags::kNormals) ? size_ : 0);
>     rgbs_.conservativeResize(NoChange, f.contains(pc_flags::kRGBs) ? size_ : 0);
>     // Note: The row number can change depends on whether 'f' contains a
> ```

nit "row number" isn't precisely correct.  (A "row number" is an index into the array, not the size of the array.) 

Either "number of rows" or "row size" would be fine as an alternative.
___
*[`perception/point_cloud.cc` line 104 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk1USu3-kAzXS-3JR_:-NQk1USu3-kAzXS-3JRa:b-737nt) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L104)):*
> ```cpp
>       const int xyz_size = xyzs_.cols();
>       DRAKE_DEMAND(xyz_size == size());
>     }
> ```

BTW It's strange to me that the invariant doesn't check for the zero side of things, too.



_Suggestion:_
```cpp
    const int xyz_size = xyzs_.cols();
    if (fields_.contains(pc_flags::kXYZs)) {
      DRAKE_DEMAND(xyz_size == size());
    } else {
      DRAKE_DEMAND(xyz_size == 0);
    }
```
___
*[`perception/point_cloud.cc` line 114 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk1mSc0aYI4g7e7VKZ:-NQk1mSc0aYI4g7e7VK_:b-bsjsye) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L114)):*
> ```cpp
>     }
>     if (fields_.has_descriptor()) {
>       const int descriptor_size = descriptors_.cols();
> ```

BTW The descriptors have variable-size `rows()` as well, which should match the descriptor.  Might also be worth double-checking that here.
___
*[`perception/point_cloud.cc` line 229 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk-j1f6t9wVlp14sxu:-NQk-j1f6t9wVlp14sxv:bbeowvy) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L229)):*
> ```cpp
>     pc_flags::Fields added_fields = FindAddedFields(old_fields, new_fields);
>     if (added_fields.contains(pc_flags::kXYZs))
>       mutable_xyzs().middleCols(0, size()).setConstant(kDefaultValue);
> ```

nit I'm not sure what the `middleCols` is accomplishing doing here. On first look, it seems like we're assigning the entire array anyway. Would `mutable_xyzs().setConstant(kDefaultValue);` work just as well?

(Likewise for all channels.)
___
*[`perception/point_cloud.cc` line 264 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQjzdEX0Cq_Gil59Luf:-NQjzdEX0Cq_Gil59Lug:b-ssdahb) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L264)):*
> ```cpp
>   if (fields_in == pc_flags::kInherit) {
>     fields_to_copy = other.storage_->fields();
>     SetFields(other.storage_->fields());
> ```

nit Do you want to enable "skip initialization" here?  I think we'll always overwrite the zero'd data down below, so zeroing it here is unnecessary?
___
*[`perception/point_cloud.cc` line 270 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQjzsqAEPoIH0tV29xL:-NQjzsqB0m03MpHfFId7:b-ipx9j1) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L270)):*
> ```cpp
>   }
> 
>   // Update the size of this point cloud if necessary.
> ```

minor It's not strictly required, but for transactional integrity I think we should update the size first and then the fields second.

If we hit the `throw` for the case of `allow_resize == false`, then we will have only _partially_ copied the other cloud -- `this` will have different fields, but nothing else.  Ideally, it would have either all of the other data, or else no changes.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

___
*[`perception/point_cloud.cc` line 86 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk0wMAEVCImGAGUB6q:-NQksxmd5MVyweSEcQU9:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L86)):*

nit "row number" isn't precisely correct.  (A "row number" is an index into the array, not the size of the array.) 

Either "number of rows" or "row size" would be fine as an alternative.
</blockquote></details>

Done.
___
*[`perception/point_cloud.cc` line 104 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk1USu3-kAzXS-3JR_:-NQkszRmDo5OgLQNAgAU:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L104)):*

BTW It's strange to me that the invariant doesn't check for the zero side of things, too.
</blockquote></details>

Done.
___
*[`perception/point_cloud.cc` line 114 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk1mSc0aYI4g7e7VKZ:-NQkt08u1mcqqv8RCkec:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L114)):*

BTW The descriptors have variable-size `rows()` as well, which should match the descriptor.  Might also be worth double-checking that here.
</blockquote></details>

Done.
___
*[`perception/point_cloud.cc` line 229 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQk-j1f6t9wVlp14sxu:-NQkt1BP7CGohx0XbvFU:bmsf30e) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L229)):*

nit I'm not sure what the `middleCols` is accomplishing doing here. On first look, it seems like we're assigning the entire array anyway. Would `mutable_xyzs().setConstant(kDefaultValue);` work just as well?

(Likewise for all channels.)
</blockquote></details>

Done. Right!
___
*[`perception/point_cloud.cc` line 270 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQjzsqAEPoIH0tV29xL:-NQkoeaADNpPpSSLnRxv:b5jfqax) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L270)):*

minor It's not strictly required, but for transactional integrity I think we should update the size first and then the fields second.

If we hit the `throw` for the case of `allow_resize == false`, then we will have only _partially_ copied the other cloud -- `this` will have different fields, but nothing else.  Ideally, it would have either all of the other data, or else no changes.
</blockquote></details>

Done. To make sure I follow-
(1) It's not strictly required because we don't catch exceptions. The program will terminate anyway, but it's a good practice to preserve transactional integrity.
(2) The order of `resize() and then Setfields()` is better than the other way because we currently don't throw during SetFields() operations.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`perception/point_cloud.cc` line 270 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-NQjzsqAEPoIH0tV29xL:-NQkyNHsArHR-ghLmWvn:b1l7pz) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e60c0cee719ea31505892ef81f6b02165cd7d64/perception/point_cloud.cc#L270)):*

Done. To make sure I follow-
(1) It's not strictly required because we don't catch exceptions. The program will terminate anyway, but it's a good practice to preserve transactional integrity.
(2) The order of `resize() and then Setfields()` is better than the other way because we currently don't throw during SetFields() operations.
</blockquote></details>

Yes, mostly.

I should have linked to https://en.cppreference.com/w/cpp/language/exceptions#Exception_safety -- what I had in mind was the "2. Strong exception safety guarantee".

The thing to remember is that in Python we _do_ catch exceptions (and our C++ users might decide to catch as well, even if we don't in our own code).  So when it's easy to reach the level 2 guarantee, it's nicer.

zachfang (commented):
+@SeanCurtis-TRI for the platform review per schedule. Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-:-NQpnW8O00a6ct_G6ZYT:b-1t1vc0)* status: LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987)* status: LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM:

> Reviewed 1 of 4 files at r3, 2 of 4 files at r4, 1 of 3 files at r5, 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18987#-:-NR-gHHc2-FTQk_-45SV:bmxilld)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform) (waiting on @zachfang)

PR #19000, opened 2023-03-14
[multibody] Add Block3x3SparseMatrix

xuchenhan-tri:
Block3x3SparseMatrix is a highly-specialized sparse matrix data structure that stores data as 3x3 dense blocks. It provides a minimal set of APIs to support the matrix operations required for constraint Jacobians in SAP. It is particularly useful for constraints involving deformable bodies where the non-zero entries usually assume a 3x3 block structure (i.e. each constraint has 3 constraint equations and dofs involved in the constraint also come in groups of 3 as the x,y,z coordinates of the deformable mesh vertex involved in the constraint).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -14,6 +14,7 @@ drake_cc_package_library(
     name = "contact_solvers",
     visibility = ["//visibility:public"],
     deps = [
+        ":block_3x3_sparse_matrix",
         ":block_sparse_matrix",
         ":contact_solver",
         ":contact_solver_results",
@@ -25,6 +26,16 @@ drake_cc_package_library(
         ":sparse_linear_operator",
         ":supernodal_solver",
         ":system_dynamics_data",
+    ],
+)
+
+drake_cc_library(
+    name = "block_3x3_sparse_matrix",
+    srcs = ["block_3x3_sparse_matrix.cc"],
+    hdrs = ["block_3x3_sparse_matrix.h"],
+    deps = [
+        "//common:default_scalars",
+        "//common:essential",
     ],
 )
 
@@ -160,6 +171,14 @@ drake_cc_library(
     deps = [
         ":linear_operator",
         "//common:default_scalars",
+    ],
+)
+
+drake_cc_googletest(
+    name = "block_3x3_sparse_matrix_test",
+    deps = [
+        ":block_3x3_sparse_matrix",
+        "//common/test_utilities:eigen_matrix_compare",
     ],
 )
 
--- a/multibody/contact_solvers/block_3x3_sparse_matrix.cc
+++ b/multibody/contact_solvers/block_3x3_sparse_matrix.cc
@@ -1,0 +1,228 @@
+#include "drake/multibody/contact_solvers/block_3x3_sparse_matrix.h"
+
+#include <algorithm>
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+namespace {
+
+template <typename T>
+using Triplet = typename Block3x3SparseMatrix<T>::Triplet;
+/* Given a vector of Triplets with the same block row index that is sorted
+ in increasing block column index, merge entries that have the same block
+ column index in place by summing them up. After this function, the triplets in
+ `data` is still sorted in increasing block column index but does not contain
+ duplicate block column indices, and the size of the vector is equal to the
+ number of unique block column indices. */
+template <typename T>
+void MergeDuplicates(std::vector<Triplet<T>>* data) {
+  DRAKE_DEMAND(data != nullptr);
+  if (data->empty()) return;
+  const int block_row = std::get<0>((*data)[0]);
+  /* Traverse the vector with slow/fast pointer where the fast pointer is the
+   source and the slow pointer is the destination. */
+  int slow = 0;
+  int fast = 0;
+  int slow_col = std::get<1>((*data)[slow]);
+  Matrix3<T> value = Matrix3<T>::Zero();
+  while (fast < static_cast<int>(data->size())) {
+    const Triplet<T>& t = (*data)[fast];
+    const int fast_col = std::get<1>(t);
+    if (fast_col == slow_col) {
+      value += std::get<2>(t);
+    } else {
+      (*data)[slow++] = {block_row, slow_col, value};
+      slow_col = fast_col;
+      value = std::get<2>(t);
+    }
+    ++fast;
+  }
+  /* Don't forget to write down the last batch of triplets. */
+  (*data)[slow++] = {block_row, slow_col, value};
+  data->resize(slow);
+}
+
+}  // namespace
+
+template <class T>
+void Block3x3SparseMatrix<T>::SetFromTriplets(
+    const std::vector<Triplet>& triplets) {
+  /* Clear all existing data. */
+  for (std::vector<Triplet>& row : row_data_) {
+    row.clear();
+  }
+  for (std::vector<Index>& indices : col_to_indices_) {
+    indices.clear();
+  }
+
+  /* Populate `row_data_` and `col_to_indices_`. */
+  for (const Triplet& t : triplets) {
+    const int block_row = std::get<0>(t);
+    const int block_col = std::get<1>(t);
+    DRAKE_DEMAND(0 <= block_row && block_row < block_rows_);
+    DRAKE_DEMAND(0 <= block_col && block_col < block_cols_);
+    row_data_[block_row].push_back(t);
+  }
+
+  /* Maintain the ordering invariance in `row_data_` and compute the total
+   number of non-zero blocks. */
+  num_blocks_ = 0;
+  for (int r = 0; r < block_rows_; ++r) {
+    std::sort(row_data_[r].begin(), row_data_[r].end(),
+              [](const Triplet& t1, const Triplet& t2) {
+                const int col_1 = std::get<1>(t1);
+                const int col_2 = std::get<1>(t2);
+                return col_1 < col_2;
+              });
+    MergeDuplicates<T>(&(row_data_[r]));
+    num_blocks_ += row_data_[r].size();
+  }
+
+  for (int block_row = 0; block_row < static_cast<int>(row_data_.size());
+       ++block_row) {
+    for (int flat_index = 0;
+         flat_index < static_cast<int>(row_data_[block_row].size());
+         ++flat_index) {
+      const int block_col = std::get<1>(row_data_[block_row][flat_index]);
+      col_to_indices_[block_col].push_back({block_row, flat_index});
+    }
+  }
+}
+
+template <class T>
+void Block3x3SparseMatrix<T>::MultiplyAndAddTo(
+    const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(A.rows() == cols());
+  DRAKE_DEMAND(y->rows() == rows());
+  for (const std::vector<Triplet>& row : row_data_) {
+    for (const Triplet& triplet : row) {
+      const int block_row = std::get<0>(triplet);
+      const int block_col = std::get<1>(triplet);
+      const Matrix3<T>& m = std::get<2>(triplet);
+      y->template middleRows<3>(3 * block_row) +=
+          m * A.template middleRows<3>(3 * block_col);
+    }
+  }
+}
+
+template <class T>
+void Block3x3SparseMatrix<T>::LeftMultiplyAndAddTo(
+    const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(A.cols() == rows());
+  DRAKE_DEMAND(y->rows() == A.rows());
+  for (const std::vector<Triplet>& row : row_data_) {
+    for (const Triplet& triplet : row) {
+      const int block_row = std::get<0>(triplet);
+      const int block_col = std::get<1>(triplet);
+      const Matrix3<T>& m = std::get<2>(triplet);
+      y->template middleCols<3>(3 * block_col) +=
+          A.template middleCols<3>(3 * block_row) * m;
+    }
+  }
+}
+
+template <class T>
+void Block3x3SparseMatrix<T>::TransposeAndMultiplyAndAddTo(
+    const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(rows() == A.rows());
+  DRAKE_DEMAND(y->rows() == cols());
+  for (const std::vector<Triplet>& row : row_data_) {
+    for (const Triplet& triplet : row) {
+      const int block_row = std::get<0>(triplet);
+      const int block_col = std::get<1>(triplet);
+      const Matrix3<T>& m = std::get<2>(triplet);
+      y->template middleRows<3>(3 * block_col) +=
+          m.transpose() * A.template middleRows<3>(3 * block_row);
+    }
+  }
+}
+
+template <class T>
+void Block3x3SparseMatrix<T>::TransposeAndMultiplyAndAddTo(
+    const Block3x3SparseMatrix<T>& A, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(rows() == A.rows());
+  DRAKE_DEMAND(y->rows() == this->cols());
+  DRAKE_DEMAND(y->cols() == A.cols());
+
+  if (A.row_data_.empty() || this->row_data_.empty()) {
+    return;
+  }
+
+  /* We are performing yᵢⱼ += ∑ₖ Mₖᵢᵀ * Aₖⱼ. For each ij block in y, we need to
+   sum over block row indices, k, where Mₖᵢ and Aₖⱼ both have non-zero blocks.
+   We do this by looping over block row indices. For each block row k, we find
+   all blocks of A and M that have k as block row index. Then we loop over their
+   block column indices (i for M and j for A), perform the dense multiplication,
+   and add to the corresponding block in y. */
+  for (int k = 0; k < block_rows_; ++k) {
+    for (const Triplet& m : row_data_[k]) {
+      const int i = std::get<1>(m);  // block column index of M block
+      const Matrix3<T>& M_ki = std::get<2>(m);
+      for (const Triplet& a : A.row_data_[k]) {
+        const int j = std::get<1>(a);  // block column index of A block
+        const Matrix3<T>& A_kj = std::get<2>(a);
+        y->template block<3, 3>(3 * i, 3 * j) += M_ki.transpose() * A_kj;
+      }
+    }
+  }
+}
+
+template <class T>
+void Block3x3SparseMatrix<T>::MultiplyWithScaledTransposeAndAddTo(
+    const VectorX<T>& scale, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(cols() == scale.size());
+  DRAKE_DEMAND(rows() == y->rows());
+  DRAKE_DEMAND(rows() == y->cols());
+  /* Let y be the result, then we are computing
+   yᵢⱼ += ∑ₖ Mᵢₖ * scaleₖ * Mⱼₖ. */
+  for (int k = 0; k < block_cols_; ++k) {
+    const std::vector<Index>& indices = col_to_indices_[k];
+    const auto scale_block = scale.template segment<3>(3 * k);
+    for (int a = 0; a < static_cast<int>(indices.size()); ++a) {
+      const Triplet& t1 = row_data_[indices[a].row][indices[a].flat];
+      const int i = std::get<0>(t1);
+      const Matrix3<T>& M_ik = std::get<2>(t1);
+      for (int b = a; b < static_cast<int>(indices.size()); ++b) {
+        const Triplet& t2 = row_data_[indices[b].row][indices[b].flat];
+        const int j = std::get<0>(t2);
+        const Matrix3<T>& M_kj = std::get<2>(t2);
+        const Matrix3<T> y_ij =
+            M_ik * scale_block.asDiagonal() * M_kj.transpose();
+        y->template block<3, 3>(3 * i, 3 * j) += y_ij;
+        /* Exploit symmetry. */
+        if (a != b) {
+          y->template block<3, 3>(3 * j, 3 * i) += y_ij.transpose();
+        }
+      }
+    }
+  }
+}
+
+template <class T>
+MatrixX<T> Block3x3SparseMatrix<T>::MakeDenseMatrix() const {
+  MatrixX<T> result = MatrixX<T>::Zero(rows(), cols());
+  for (const std::vector<Triplet>& row : row_data_) {
+    for (const Triplet& triplet : row) {
+      const int block_row = std::get<0>(triplet);
+      const int block_col = std::get<1>(triplet);
+      const Matrix3<T>& m = std::get<2>(triplet);
+      result.template block<3, 3>(3 * block_row, 3 * block_col) = m;
+    }
+  }
+  return result;
+}
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
+
+DRAKE_DEFINE_CLASS_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_SCALARS(
+    class ::drake::multibody::contact_solvers::internal::Block3x3SparseMatrix)
--- a/multibody/contact_solvers/block_3x3_sparse_matrix.h
+++ b/multibody/contact_solvers/block_3x3_sparse_matrix.h
@@ -1,0 +1,131 @@
+#pragma once
+
+#include <tuple>
+#include <vector>
+
+#include "drake/common/default_scalars.h"
+#include "drake/common/eigen_types.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+/* A sparse matrix data structure composed of m-by-n block submatrices, where
+ each submatrix is of size 3-by-3. Only the non-zero submatrices are stored for
+ efficiency. We use `M` to denote `this` Block3x3SparseMatrix throughout this
+ class.
+ @tparam_default_scalar */
+template <class T>
+class Block3x3SparseMatrix {
+ public:
+  DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Block3x3SparseMatrix);
+
+  /* A triplet is a tuple (i, j, value) defining a non-zero submatrix where i is
+   the block row index, j is the block column index, and value is a 3x3 dense
+   matrix. */
+  using Triplet = std::tuple<int, int, Matrix3<T>>;
+
+  /* Creates a Block3x3SparseMatrix with capacity for
+   `block_rows`-by-`block_cols` block submatrices. The matrix is initialized to
+   be empty, i.e. no non-zero blocks. */
+  Block3x3SparseMatrix(int block_rows, int block_cols)
+      : row_data_(block_rows),
+        block_rows_(block_rows),
+        block_cols_(block_cols),
+        col_to_indices_(block_cols) {
+    DRAKE_DEMAND(block_rows >= 0);
+    DRAKE_DEMAND(block_cols >= 0);
+  }
+
+  /* Fills `this` matrix with the given vector of triplets. The input vector of
+   triplets does not have to be sorted. Duplicated items (i.e. triplet with the
+   same block row and column index) are summed. Therefore, num_blocks() returns
+   the number of non-repeated non-zero blocks in the resulting matrix. The
+   contents in `this` matrix prior to the call to this function is destroyed.
+   @pre The block row and column indices for each triplet in the given vector is
+   within [0, block_rows) and [0, column_blocks) prescribed in the constructor
+   of this class. */
+  void SetFromTriplets(const std::vector<Triplet>& triplets);
+
+  int rows() const { return block_rows_ * 3; }
+  int cols() const { return block_cols_ * 3; }
+  int block_rows() const { return block_rows_; }
+  int block_cols() const { return block_cols_; }
+
+  /* Performs y += M * A.
+   @pre y != nullptr and the sizes of A and y are compatible with this matrix.
+  */
+  void MultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
+                        EigenPtr<MatrixX<T>> y) const;
+
+  /* Performs y += A * M.
+   @pre y != nullptr and the sizes of A and y are compatible with this matrix.
+  */
+  void LeftMultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
+                            EigenPtr<MatrixX<T>> y) const;
+
+  /* Performs y += Mᵀ * A, where A is dense.
+   @pre y != nullptr and the sizes of A and y are compatible with this matrix.
+  */
+  void TransposeAndMultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
+                                    EigenPtr<MatrixX<T>> y) const;
+
+  /* Performs y += Mᵀ * A, where A is also 3x3 block sparse.
+   @pre y != nullptr and the sizes of A and y are compatible with this matrix.
+  */
+  void TransposeAndMultiplyAndAddTo(const Block3x3SparseMatrix<T>& A,
+                                    EigenPtr<MatrixX<T>> y) const;
+
+  /* Performs y += M * scale.asDiagonal() * M.transpose().
+   @pre y != nullptr and the sizes of scale and y are compatible with this
+   matrix. */
+  void MultiplyWithScaledTransposeAndAddTo(const VectorX<T>& scale,
+                                           EigenPtr<MatrixX<T>> y) const;
+
+  /* Returns the non-zero blocks in the matrix in a row-major fashion. Within
+   each block row (result[row]), the blocks are sorted in increasing column
+   indices. The returned std::vector always has size `block_rows()`. If
+   result[row] is empty, then it means that there's no non-zero block in that
+   block row. */
+  const std::vector<std::vector<Triplet>>& get_triplets() const {
+    return row_data_;
+  }
+
+  /* Returns the number of non-zero blocks.
+   @note the result might not be equal to the number of triplets passed into
+   `SetFromTriplets` as duplicated blocks are summed. */
+  int num_blocks() const { return num_blocks_; }
+
+  /* Returns the matrix as an Eigen dense matrix. Useful for debugging. */
+  MatrixX<T> MakeDenseMatrix() const;
+
+ private:
+  /* We store the non-zero blocks in the matrix in a row-major fashion. Within
+   each row, the blocks are sorted in increasing column indices.
+   row_data_.size() is always equal to block_rows() and row_data_ never stores
+   duplicated blocks. If row_data_[i] is empty, then it means that there's no
+   non-zero block in that block row. */
+  std::vector<std::vector<Triplet>> row_data_;
+  int block_rows_{};
+  int block_cols_{};
+  int num_blocks_{};
+
+  /* Index into `row_data_`. For a given `index`,
+   row_data_[index.row][index.flat] retrieves the corresponding triplet. */
+  struct Index {
+    int row;
+    int flat;
+  };
+  /* For each block column j, col_to_indices_[j] contains all the indices
+   corresponding to non-zero entries in block column j in increasing row index.
+   That is, col_to_indices_[j][i1].row < col_to_indices_[j][i2].row if i1 < i2.
+   The same information can be computed from `row_data_`, but we store this
+   redundant, precomputed information for convenience. */
+  std::vector<std::vector<Index>> col_to_indices_;
+};
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
--- a/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc
+++ b/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc
@@ -1,0 +1,151 @@
+#include "drake/multibody/contact_solvers/block_3x3_sparse_matrix.h"
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+namespace {
+
+using Eigen::Matrix3d;
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
+
+/* Returns an arbitrary non-zero matrix of size m-by-n.*/
+MatrixXd MakeArbitraryMatrix(int m, int n) {
+  MatrixXd A(m, n);
+  for (int i = 0; i < m; ++i) {
+    for (int j = 0; j < n; ++j) {
+      A(i, j) = 3 * i + 4 * j;
+    }
+  }
+  return A;
+}
+
+/* Returns a dummy 3x3 matrix with all entries in the matrix equal to the given
+ value. */
+Matrix3d MakeMatrix(double value) {
+  return value * Matrix3d::Ones();
+}
+
+/* Returns an arbitrary Block3x3SparseMatrix with size 12-by-9. */
+Block3x3SparseMatrix<double> MakeBlockSparseMatrix() {
+  Block3x3SparseMatrix<double> sparse_matrix(4, 3);
+  std::vector<Block3x3SparseMatrix<double>::Triplet> triplets;
+  triplets.emplace_back(0, 0, MakeMatrix(1.0));
+  triplets.emplace_back(0, 0, MakeMatrix(2.0));
+  triplets.emplace_back(0, 1, MakeMatrix(3.0));
+  triplets.emplace_back(0, 1, MakeMatrix(4.0));
+  triplets.emplace_back(2, 1, MakeMatrix(5.0));
+  triplets.emplace_back(3, 2, MakeMatrix(6.0));
+  sparse_matrix.SetFromTriplets(triplets);
+  EXPECT_EQ(sparse_matrix.num_blocks(), 4);
+  return sparse_matrix;
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, Size) {
+  const Block3x3SparseMatrix<double> sparse_matrix(4, 3);
+  EXPECT_EQ(sparse_matrix.rows(), 12);
+  EXPECT_EQ(sparse_matrix.cols(), 9);
+  EXPECT_EQ(sparse_matrix.block_rows(), 4);
+  EXPECT_EQ(sparse_matrix.block_cols(), 3);
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, SetFromTriplets) {
+  Block3x3SparseMatrix<double> sparse_matrix = MakeBlockSparseMatrix();
+  MatrixXd expected_matrix(12, 9);
+  expected_matrix.topLeftCorner<3, 3>() = MakeMatrix(3.0);
+  expected_matrix.block<3, 3>(0, 3) = MakeMatrix(7.0);
+  expected_matrix.block<3, 3>(6, 3) = MakeMatrix(5.0);
+  expected_matrix.bottomRightCorner<3, 3>() = MakeMatrix(6.0);
+  EXPECT_EQ(sparse_matrix.MakeDenseMatrix(), expected_matrix);
+
+  /* Verify that setting the sparse matrix with a new vector of triplets resets
+   the matrix. */
+  std::vector<Block3x3SparseMatrix<double>::Triplet> triplets;
+  triplets.emplace_back(2, 2, MakeMatrix(4.0));
+  sparse_matrix.SetFromTriplets(triplets);
+  expected_matrix.setZero();
+  expected_matrix.block<3, 3>(6, 6) = MakeMatrix(4.0);
+  EXPECT_EQ(sparse_matrix.MakeDenseMatrix(), expected_matrix);
+  EXPECT_EQ(sparse_matrix.num_blocks(), 1);
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, MultiplyAndAddTo) {
+  const MatrixXd A = MakeArbitraryMatrix(9, 7);
+  const Block3x3SparseMatrix<double> sparse_matrix = MakeBlockSparseMatrix();
+  const MatrixXd dense_matrix = sparse_matrix.MakeDenseMatrix();
+
+  /* Set the destinations to compatible-sized non-zero matrices. */
+  MatrixXd y1 = MakeArbitraryMatrix(12, 7);
+  MatrixXd y2 = y1;
+
+  sparse_matrix.MultiplyAndAddTo(A, &y1);
+  y2 += dense_matrix * A;
+  EXPECT_TRUE(CompareMatrices(y1, y2));
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, LeftMultiplyAndAddTo) {
+  const Block3x3SparseMatrix<double> sparse_matrix = MakeBlockSparseMatrix();
+  const MatrixXd dense_matrix = sparse_matrix.MakeDenseMatrix();
+
+  const MatrixXd A = MakeArbitraryMatrix(7, 12);
+
+  /* Set the destinations to compatible-sized non-zero matrices. */
+  MatrixXd y1 = MakeArbitraryMatrix(7, 9);
+  MatrixXd y2 = y1;
+
+  sparse_matrix.LeftMultiplyAndAddTo(A, &y1);
+  y2 += A * dense_matrix;
+  EXPECT_TRUE(CompareMatrices(y1, y2));
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, TransposeAndMultiplyAndAddTo) {
+  const Block3x3SparseMatrix<double> sparse_matrix = MakeBlockSparseMatrix();
+  const MatrixXd dense_matrix = sparse_matrix.MakeDenseMatrix();
+
+  const MatrixXd A = MakeArbitraryMatrix(12, 6);
+
+  /* Test the sparse-dense multiplication. */
+  /* Set the destinations to compatible-sized non-zero matrices. */
+  MatrixXd y1 = MakeArbitraryMatrix(9, 6);
+  MatrixXd y2 = y1;
+
+  sparse_matrix.TransposeAndMultiplyAndAddTo(A, &y1);
+  y2 += dense_matrix.transpose() * A;
+  EXPECT_TRUE(CompareMatrices(y1, y2));
+
+  /* Test the sparse-sparse multiplication. */
+  Block3x3SparseMatrix<double> A_sparse(4, 2);
+  std::vector<Block3x3SparseMatrix<double>::Triplet> triplets;
+  triplets.emplace_back(0, 1, MakeMatrix(4.0));
+  triplets.emplace_back(2, 1, MakeMatrix(5.0));
+  A_sparse.SetFromTriplets(triplets);
+
+  sparse_matrix.TransposeAndMultiplyAndAddTo(A_sparse, &y1);
+  y2 += dense_matrix.transpose() * A_sparse.MakeDenseMatrix();
+  EXPECT_TRUE(CompareMatrices(y1, y2));
+}
+
+GTEST_TEST(Block3x3SparseMatrixTest, MultiplyWithScaledTransposeAndAddTo) {
+  const Block3x3SparseMatrix<double> sparse_matrix = MakeBlockSparseMatrix();
+  const MatrixXd dense_matrix = sparse_matrix.MakeDenseMatrix();
+  const VectorXd scale = VectorXd::LinSpaced(9, 0.0, 1.0);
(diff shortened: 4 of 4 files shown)


xuchenhan-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-:-NQXYijd7nDzAlmwsesM:b-57wv7n)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri)

amcastro-tri (commented):
working...

> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-:-NQ_soU69FThm1Nmwd0l:b-s297dq)* status: 15 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_jIAKAafwW5trILEe:-NQ_jIAKAafwW5trILEf:b-b4cfb5) (no related file):*
Have you considered to consolidate this class with `BlockSparseMatrix`? besides having fixed 3x3 blocks, I assume you found this class to be more performant and therefore the reason to write from scratch? If so, I'd at least add a few developer notes somewhere in this class that explains the future of BlockSparseMatrix. Should we remove it altogether? 

Considering we were exploring to make this class more general, to store arbitrary sized blocks, wouldn't that make it pretty similar to BlockSparseMatrix?

I think a few notes explaining the whys and design choices would help understand the future of these clases.

___
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_kmJu1YaC7KbEWhpY:-NQ_kmJu1YaC7KbEWhpZ:b-yozkle) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L49)):*
> ```objc
>    of this class.
>    @warn `triplets` must not contain items with duplicated block row and column
>    indices. */
> ```

what happens if the `@pre` and `@warn` conditions are not met?
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_ktW-1-eyBw4kuEad:-NQ_ktW-1-eyBw4kuEae:b-5689j) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L57)):*
> ```objc
>   int block_cols() const { return block_cols_; }
> 
>   /* Performs *y += M * A.
> ```

nit, for math, I would not write `*y`, but simply `y`. Same below.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_rRQa7kL3IcfqB6Aw:-NQ_rRQa7kL3IcfqB6Ax:bffreys) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L60)):*
> ```objc
>    @pre y != nullptr and the sizes of A and y are compatible with M. */
>   void MultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
>                         EigenPtr<VectorX<T>> y) const;
> ```

I believe this should be an `EigenPtr<Matrix<T>>`
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 63 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_l4beBtBksaXh9tS9:-NQ_l4beBtBksaXh9tSA:b-e1hsbm) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L63)):*
> ```objc
> 
>   /* Performs *y += A * M.
>    @pre y != nullptr and the sizes of A and y are compatible with M. */
> ```

nit, I know you "warned" far away in the class's docs. But really changing `M` for `this matrix` saves you only a couple characters and there is no need for a symbol out of context.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NQ_lhg7BsYhlOIBEiVk:bwr5wrb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*
> ```objc
>   /* Performs *y += Mᵀ * A, where A is dense.
>    @pre y != nullptr and the sizes of A and y are compatible with M. */
>   void TransposeAndMultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
> ```

This name in my mind seems to imply two operations; first transpose, then multiply.
Bottom line, it sounds weird to me. 
I'd suggest `MultiplyByTransposeAndAddTo()`, which reads more like the way you'd describe this operation.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_kREJ2yyErzjeuhaO:-NQ_kREJ2yyErzjeuhaP:bkdoprt) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L85)):*
> ```objc
>    each block row (result[row]), the blocks are sorted in increasing column
>    indices. If result[row] is empty, then it means that there's no non-zero
>    block in that block row. */
> ```

nit, state that the returned vector always has size `block_rows().`
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_hRjG-Ny9E0urSc6-:-NQ_hRjG-Ny9E0urSc60:b-8x16c9) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L100)):*
> ```objc
>   /* We store the non-zero blocks in the matrix in a row-major fashion. Within
>    each row, the blocks are sorted in increasing column indices. If data_[i] is
>    empty, then it means that there's no non-zero block in that block row. */
> ```

minor, this seems to imply `data_.size()` equals the number of block rows. Maybe worth stating explicitly here.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 101 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_iFyM6ERWMlKnQl6e:-NQ_iFyM6ERWMlKnQl6f:b-xbzdna) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L101)):*
> ```objc
>    each row, the blocks are sorted in increasing column indices. If data_[i] is
>    empty, then it means that there's no non-zero block in that block row. */
>   std::vector<RowData> data_;
> ```

nit, here I'd advocate for a different naming. I'd call this member `row_data_` and the type of the data stored per row simply `VectorOfTriplets` (since there's nothing specific in std::vector<Triplet> for rows only). 
Then, I believe that locally far away from these docs, it'll be clear that there is something row-ish about this member called `row_data_`.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_o__q83H3tqspGbKQ:-NQ_o__q83H3tqspGbKR:b1g3gw5) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L53)):*
> ```cpp
> void Block3x3SparseMatrix<T>::MultiplyAndAddTo(
>     const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<VectorX<T>> y) const {
>   DRAKE_DEMAND(A.rows() == cols());
> ```

DRAKE_DEMAND(y != nullptr)
DRAKE_DEMAND(y->rows() == rows())
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_rfWrBxNaevcGNy1h:-NQ_rfWrBxNaevcGNy1i:b-f9qxt9) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L59)):*
> ```cpp
>       const int block_col = std::get<1>(triplet);
>       const Matrix3<T>& m = std::get<2>(triplet);
>       y->template segment<3>(3 * block_row) +=
> ```

shouldn't this one be `middleRows` as well?
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_oikrB7mN7r5uVyqZ:-NQ_oikrB7mN7r5uVyq_:bkfhjb0) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L68)):*
> ```cpp
> void Block3x3SparseMatrix<T>::LeftMultiplyAndAddTo(
>     const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<MatrixX<T>> y) const {
>   DRAKE_DEMAND(A.cols() == rows());
> ```

similarly here and below.

DRAKE_DEMAND(y != nullptr)
DRAKE_DEMAND(y->rows() == A.rows())
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_ph0n3RRhJ2_Q7lDM:-NQ_ph0n3RRhJ2_Q7lDN:blsqapb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L99)):*
> ```cpp
>     const Block3x3SparseMatrix<T>& A, EigenPtr<MatrixX<T>> y) const {
>   DRAKE_DEMAND(rows() == A.rows());
>   DRAKE_DEMAND(y->rows() == this->cols());
> ```

minor, here you use `this` while you don't in other places.



_Code quote:_
```cpp
MultiplyAndAddTo(
```
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_plPs2dSdS3EhY56X:-NQ_plPs2dSdS3EhY56Y:b955vrs) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L99)):*
> ```cpp
>     const Block3x3SparseMatrix<T>& A, EigenPtr<MatrixX<T>> y) const {
>   DRAKE_DEMAND(rows() == A.rows());
>   DRAKE_DEMAND(y->rows() == this->cols());
> ```

nit, DEMAND y is not nullptr
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_sci29LzQFnT14jkd:-NQ_sci29LzQFnT14jke:bbfewwt) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L165)):*
> ```cpp
>       const int block_col = std::get<1>(triplet);
>       const Matrix3<T>& m = std::get<2>(triplet);
>       result.template block<3, 3>(3 * block_row, 3 * block_col) += m;
> ```

do we allow for repeated having several values for the same i,j block? with the semantics that they'll get added?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 15 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_jIAKAafwW5trILEe:-NQa364r51yiNKPLcs8n:bsjvs3e) (no related file):*
I forgot to share the branch where I have all the commits working together for a bigger picture. Here it is: https://github.com/xuchenhan-tri/drake/tree/sparse-constraint-jacobian

The design is that `Block3x3SparseMatrix` along with Eigen dense matrix and potential other matrix storage types are all represented as a `MatrixBlock` which is a wrapper around `std::variant<Block3x3SparseMatrix, MatrixXd, YourNewMatrixType>`
Then `BlockSparseMatrix` will contain `MatrixBlock` as its blocks (instead of Eigen dense matrices as it does now).

> Have you considered to consolidate this class with BlockSparseMatrix? 

Yes, one can imagine a world where each individual block in `BlockSparseMatrix` is itself a `BlockSparseMatrix`. I played with the idea but it turned out cumbersome. It's nice to have `BlockSparseMatrix` hold the entire Jacobian and `MatrixBlock` hold individual Jacobian blocks. That gives me the freedom to exploit whatever structure of a specific Jacobian (e.g. 3x3 blocks for deformable) without worrying about breaking invariant for other types of Jacobians.

___
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NQa5Sd3CtC2FjxWHmxZ:bh15blb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*
I initially did name the function as you suggested. The nice thing about `MultiplyByTransposeAndAddTo()` is that it aligns nicely with the naming convention that you have in `BlockSparseMatrix`. However, the name is ambiguous about whether the operation is M^T * A or M * A^T. 

> This name in my mind seems to imply two operations; first transpose, then multiply.

This concern is mitigated by the fact that this function is const and can't modify the matrix. Any transpose can only be temporary, which is exactly what the function does.

> Bottom line, it sounds weird to me.

Fair point.  It sounds weird to me too. After much debate with myself, I have chosen weird over ambiguity. 

I discussed this with @mitiguy yesterday. He has a few more provoking ideas :)

I struggled with the naming for `LeftMultiplyAndAddTo` above as well, but you don't seem to oppose it.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 15 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NQb6syj3CISD17_DyKv:bg44dc9) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*

I initially did name the function as you suggested. The nice thing about `MultiplyByTransposeAndAddTo()` is that it aligns nicely with the naming convention that you have in `BlockSparseMatrix`. However, the name is ambiguous about whether the operation is M^T * A or M * A^T. 

> This name in my mind seems to imply two operations; first transpose, then multiply.

This concern is mitigated by the fact that this function is const and can't modify the matrix. Any transpose can only be temporary, which is exactly what the function does.

> Bottom line, it sounds weird to me.

Fair point.  It sounds weird to me too. After much debate with myself, I have chosen weird over ambiguity. 

I discussed this with @mitiguy yesterday. He has a few more provoking ideas :)

I struggled with the naming for `LeftMultiplyAndAddTo` above as well, but you don't seem to oppose it.
</blockquote></details>

FYI:  From our whiteboard discussion and from the language suggested by Steffi:
MultiplyArgByThis()  as contrasted with
MultiplyThisByArg()

Note:  I am heading to teach now.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 10 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @pre)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_kmJu1YaC7KbEWhpY:-NQb_3JpDJxGh2bPleLq:b-mmrtwt) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L49)):*

what happens if the `@pre` and `@warn` conditions are not met?
</blockquote></details>

I don't think we usually are obliged to explain what happens if the precondition is not met, but we usually try to be nice about it. 

 I added a couple of DRAKE_DEMAND for the first @pre and removed the requirement of no duplicates. The signature is close to the Eigen::SparseMatrix one and it's easy to assume that the semantic is the same.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 57 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_ktW-1-eyBw4kuEad:-NQbCXCV3ZWF6bzTz4R-:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L57)):*

nit, for math, I would not write `*y`, but simply `y`. Same below.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_rRQa7kL3IcfqB6Aw:-NQbChwMCsX6_g3Wsy28:b-dtkdrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L60)):*

I believe this should be an `EigenPtr<Matrix<T>>`
</blockquote></details>

Done, thanks.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 63 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_l4beBtBksaXh9tS9:-NQbCngDEkzPVxrp6Re_:b-fiswgm) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L63)):*

nit, I know you "warned" far away in the class's docs. But really changing `M` for `this matrix` saves you only a couple characters and there is no need for a symbol out of context.
</blockquote></details>

Are you proposing changing to "y += A * this matrix" too or just this line?
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_kREJ2yyErzjeuhaO:-NQbEeBE2xXVcoHNtAKt:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L85)):*

nit, state that the returned vector always has size `block_rows().`
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_hRjG-Ny9E0urSc6-:-NQbEoC1-sjCL6fruKAR:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L100)):*

minor, this seems to imply `data_.size()` equals the number of block rows. Maybe worth stating explicitly here.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 101 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_iFyM6ERWMlKnQl6e:-NQbFJM96SPT94KcTVTs:b8vtiex) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L101)):*

nit, here I'd advocate for a different naming. I'd call this member `row_data_` and the type of the data stored per row simply `VectorOfTriplets` (since there's nothing specific in std::vector<Triplet> for rows only). 
Then, I believe that locally far away from these docs, it'll be clear that there is something row-ish about this member called `row_data_`.
</blockquote></details>

Done.

I removed the alias because `VectorOfTriplets` isn't really more concise.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_o__q83H3tqspGbKQ:-NQbFVCI3cApBrUsM8wL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L53)):*

DRAKE_DEMAND(y != nullptr)
DRAKE_DEMAND(y->rows() == rows())
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_rfWrBxNaevcGNy1h:-NQbFWCw0JZUPrJ7uGKm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L59)):*

shouldn't this one be `middleRows` as well?
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_oikrB7mN7r5uVyqZ:-NQbFuUB5kZ7NO8sMKte:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L68)):*

similarly here and below.

DRAKE_DEMAND(y != nullptr)
DRAKE_DEMAND(y->rows() == A.rows())
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_ph0n3RRhJ2_Q7lDM:-NQbG6Um4U0jiJlBTHKN:b-1tx8x3) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L99)):*

minor, here you use `this` while you don't in other places.
</blockquote></details>

I usually do that when another object of the same type is involved in the computation.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_sci29LzQFnT14jkd:-NQbGOP-1D1UThfFi4cp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L165)):*

do we allow for repeated having several values for the same i,j block? with the semantics that they'll get added?
</blockquote></details>

Done.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 7 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_jIAKAafwW5trILEe:-NQemncy70Jly1046cmo:b-8cg2om) (no related file):*

I forgot to share the branch where I have all the commits working together for a bigger picture. Here it is: https://github.com/xuchenhan-tri/drake/tree/sparse-constraint-jacobian

The design is that `Block3x3SparseMatrix` along with Eigen dense matrix and potential other matrix storage types are all represented as a `MatrixBlock` which is a wrapper around `std::variant<Block3x3SparseMatrix, MatrixXd, YourNewMatrixType>`
Then `BlockSparseMatrix` will contain `MatrixBlock` as its blocks (instead of Eigen dense matrices as it does now).

> Have you considered to consolidate this class with BlockSparseMatrix? 

Yes, one can imagine a world where each individual block in `BlockSparseMatrix` is itself a `BlockSparseMatrix`. I played with the idea but it turned out cumbersome. It's nice to have `BlockSparseMatrix` hold the entire Jacobian and `MatrixBlock` hold individual Jacobian blocks. That gives me the freedom to exploit whatever structure of a specific Jacobian (e.g. 3x3 blocks for deformable) without worrying about breaking invariant for other types of Jacobians.
</blockquote></details>

I see, thank you so much for the detailed explanation. Now I understand how these classes talk to each other. 

___
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_kmJu1YaC7KbEWhpY:-NQenIVG9M8EGxgqIpvi:bzi1zjy) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L49)):*

I don't think we usually are obliged to explain what happens if the precondition is not met, but we usually try to be nice about it. 

 I added a couple of DRAKE_DEMAND for the first @pre and removed the requirement of no duplicates. The signature is close to the Eigen::SparseMatrix one and it's easy to assume that the semantic is the same.
</blockquote></details>

I agree. However here I usually understand the `@pre` as "at least there is a DEMAND". I was just wondering if the constructed object could be left in an invalid state, which would be quite bad for the user of this class. I like how you resolved the `@warn` now. Thanks.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 63 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_l4beBtBksaXh9tS9:-NQenpbGCM_HwhNNDARK:b-7oc0nw) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L63)):*

Are you proposing changing to "y += A * this matrix" too or just this line?
</blockquote></details>

ah no! what you did now is great. Thanks
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NQeoUsG6DcSrsZIH_lN:b8dexro) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*
> However, the name is ambiguous about whether the operation is M^T * A or M * A^T.

That's actually a good point. I guess in that case we could go with Paul's more verbose version. Thus far however, we don't even provide `M*A^T` so there is no ambiguity, yet.

I think I like Paul's suggestion since there would be no ambiguity:
`MultiplyThisTransposeByArgument()` vs. `MultiplyThisByArgumentTranspose()`

Marking myself satisfied though.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 101 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_iFyM6ERWMlKnQl6e:-NQeqNfXE5mG4WcCA4S0:bf0oexb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L101)):*

Done.

I removed the alias because `VectorOfTriplets` isn't really more concise.
</blockquote></details>

SGTM, thanks
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_ph0n3RRhJ2_Q7lDM:-NQessru6G7QgWbDi1eb:b-80ikcm) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L99)):*

I usually do that when another object of the same type is involved in the computation.
</blockquote></details>

interesting convention. 
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_sci29LzQFnT14jkd:-NQet2A46WakOp5rntKr:b-s8g4ff) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L165)):*

Done.
</blockquote></details>

Per my comment above. Maybe you can get rid of `MergeDuplicates()` and write this as you did before with `+=`. We just have to make sure the addition of duplicates assumptions holds everywhere.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 79 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQerJ8MBIOIUu2w-fh-:-NQerJ8MBIOIUu2w-fh0:b-893n6) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L79)):*
> ```cpp
>                 return col_1 < col_2;
>               });
>     MergeDuplicates<T>(&(row_data_[r]));
> ```

question, is this step necessary? that is, if we slightly modify our invariant on `row_data` to be that column blocks are ordered in (not strictly) increasing order, shouldn't all the methods below work just fine?
Similarly for `col_to_indices_`, wouldn't it be just ok if it had indexes with the same row?

If yes, I'd probably say a word or two here on why the strict requiremnt is needed. 
If no, I'd just simply remove and update the invariants docs for simpler code.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQetfvn2nIPprCb1fOD:-NQetfvn2nIPprCb1fOE:b-dq4b29) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*
> ```cpp
>   }
> 
>   /* In Einstein notation, We are performing *yᵢⱼ += Mₖᵢ * Aₖⱼ. For each ij
> ```

for math, no need to have `*` in front of `yij`.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NQeuBwKAq2UGflhESwC:bg3kstb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*
> ```cpp
>   }
> 
>   /* In Einstein notation, We are performing *yᵢⱼ += Mₖᵢ * Aₖⱼ. For each ij
> ```

Transpose missing in Mki, i.e. `yᵢⱼ += Mₖᵢᵀ * Aₖⱼ`
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 167 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQev-3q4cjZiaF2I-ZJ:-NQev-3q4cjZiaF2I-ZK:bnpelqw) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L167)):*
> ```cpp
>     for (const Triplet& m : row_data_[k]) {
>       const int i = std::get<1>(m);
>       const Matrix3<T>& M_ki = std::get<2>(m).transpose();
> ```

I believe this is creating a new object so either:
  1. Remove `&`.
  2. use `const auto` and let it resolve to whatever Eigen type comes back from the return of `.transpose()`.
  3. Do not use `.transpose()` here but later when you multiply with A_kj below
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 196 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQexr5e04ntSYPLSv0r:-NQexr5fDCBhs6wSV17h:b-5tpvmb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L196)):*
> ```cpp
>         const Triplet& t2 = row_data_[indices[b].row][indices[b].flat];
>         const int j = std::get<0>(t2);
>         const Matrix3<T>& M_jk = std::get<2>(t2).transpose();
> ```

ditto on reference here. Maybe even call `M_jk_transpose` for clarity.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 197 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeyidT4LEwLLMhrc9r:-NQeyGkGDEgdGXNuInYg:b-axdfyd) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L197)):*
> ```cpp
>         const int j = std::get<0>(t2);
>         const Matrix3<T>& M_jk = std::get<2>(t2).transpose();
>         Matrix3<T> y_ij = M_ik * scale_block.asDiagonal() * M_jk;
> ```

nit, const

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf-q68840TZ6A5Z-Ve:-NQf-q68840TZ6A5Z-Vf:b-l15lwv) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.h#L43)):*
> ```objc
>   /* Fills `this` matrix with the given vector of triplets. The input vector of
>    triplets does not have to be sorted. Duplicated items (i.e. triplet with the
>    same block row and column index) are summed. The contents in `this` matrix
> ```

re duplicates. I like what you did here. 
I'd simply add here "Therefore num_blocks() returns the number of non-repeated non-zero blocks in the resulting matrix".
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 104 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf0Hgu9UjDZ0j1oZ3g:-NQf0Hgu9UjDZ0j1oZ3h:b-y8jbnj) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.h#L104)):*
> ```objc
>    each row, the blocks are sorted in increasing column indices.
>    row_data_.size() is always equal to block_rows(). If row_data_[i] is empty,
>    then it means that there's no non-zero block in that block row. */
> ```

Add here that per invariants enforced at construction, `row_data_` will never store duplicated blocks.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 79 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQerJ8MBIOIUu2w-fh-:-NQf07be3_-KTUZ1o4o8:bgpyjgq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L79)):*

question, is this step necessary? that is, if we slightly modify our invariant on `row_data` to be that column blocks are ordered in (not strictly) increasing order, shouldn't all the methods below work just fine?
Similarly for `col_to_indices_`, wouldn't it be just ok if it had indexes with the same row?

If yes, I'd probably say a word or two here on why the strict requiremnt is needed. 
If no, I'd just simply remove and update the invariants docs for simpler code.
</blockquote></details>

Ignore this comment, if you want.
I actually like the semantics of the resulting class. 
___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 34 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQezNnN5dzB3ceZkypD:-NQezNnN5dzB3ceZkypE:b-pwxnc1) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L34)):*
> ```cpp
>   triplets.emplace_back(3, 2, MakeMatrix(6.0));
>   sparse_matrix.SetFromTriplets(triplets);
>   EXPECT_EQ(sparse_matrix.num_blocks(), 4);
> ```

ah, this wasn't clear to me. If you go with this invariant, make sure the docs explicitly say this might not match the number of triplets provided at construction.
___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 44 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQezgUmDROaSX1A-d_J:-NQezgUmDROaSX1A-d_K:bbuuz91) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L44)):*
> ```cpp
>   EXPECT_EQ(sparse_matrix.block_rows(), 4);
>   EXPECT_EQ(sparse_matrix.block_cols(), 3);
>   EXPECT_EQ(sparse_matrix.block_cols(), 3);
> ```

repeated line

amcastro-tri (commented):
Feature review done.
This is excellent Xuchen, great work.

:lgtm_strong: 

> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-:-NQf2DoX6JbenvuZtn6j:b-eonvwc)* status: 12 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 71 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf0tNq01ricXRVBs6v:-NQf0tNq01ricXRVBs6w:b5m8gv4) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L71)):*
> ```cpp
>   for (int i = 0; i < 9; ++i) {
>     for (int j = 0; j < 7; ++j) {
>       A(i, j) = 3 * i + 4 * j;
> ```

since this snippet repeats several times, I'd advocate for something like `MakeArbitraryMatrix(int n, int m)`

xuchenhan-tri (commented):
+@sammy-tri for platform review per rotation please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-:-NQfPc2O32vfRDHwcAKM:b-o7bgk8)* status: 11 unresolved discussions, LGTM missing from assignee sammy-tri(platform) (waiting on @amcastro-tri and @sammy-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NQfMS8G6xCQ7-Irc2XD:bhogu9f) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*

> However, the name is ambiguous about whether the operation is M^T * A or M * A^T.

That's actually a good point. I guess in that case we could go with Paul's more verbose version. Thus far however, we don't even provide `M*A^T` so there is no ambiguity, yet.

I think I like Paul's suggestion since there would be no ambiguity:
`MultiplyThisTransposeByArgument()` vs. `MultiplyThisByArgumentTranspose()`

Marking myself satisfied though.



</blockquote></details>

Ok, I'll proceed with the current name for this PR. We can discuss this more at the callsite in the followup PR.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf-q68840TZ6A5Z-Ve:-NQfMNHj6e9YyFzDrE1N:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.h#L43)):*

re duplicates. I like what you did here. 
I'd simply add here "Therefore num_blocks() returns the number of non-repeated non-zero blocks in the resulting matrix".
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 104 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf0Hgu9UjDZ0j1oZ3g:-NQfMyHL34uduGAeff6V:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.h#L104)):*

Add here that per invariants enforced at construction, `row_data_` will never store duplicated blocks.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_sci29LzQFnT14jkd:-NQfNvB1DilSyEikzLQD:bq1d16w) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L165)):*

Per my comment above. Maybe you can get rid of `MergeDuplicates()` and write this as you did before with `+=`. We just have to make sure the addition of duplicates assumptions holds everywhere.
</blockquote></details>

Resolving per thread above.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 79 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQerJ8MBIOIUu2w-fh-:-NQfN-bP73rvUrbnsi2U:b553nu2) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L79)):*

Ignore this comment, if you want.
I actually like the semantics of the resulting class. 
</blockquote></details>

ok, marking this as resolved.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQetfvn2nIPprCb1fOD:-NQfNDi0-kmsCpMdQY9x:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

for math, no need to have `*` in front of `yij`.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NQfNE7DCfJby7yTTga6:b-b7qsyb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

Transpose missing in Mki, i.e. `yᵢⱼ += Mₖᵢᵀ * Aₖⱼ`
</blockquote></details>

The transpose is implied in the indices. For non-transposed multiplication, it would be Mᵢₖ * Aₖⱼ instead.
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 167 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQev-3q4cjZiaF2I-ZJ:-NQfNfBUEbuvSTKwKEcb:b-dtkds3) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L167)):*

I believe this is creating a new object so either:
  1. Remove `&`.
  2. use `const auto` and let it resolve to whatever Eigen type comes back from the return of `.transpose()`.
  3. Do not use `.transpose()` here but later when you multiply with A_kj below
</blockquote></details>

Done, thanks!
___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 196 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQexr5e04ntSYPLSv0r:-NQfNn637yYwMpRfRdyj:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L196)):*

ditto on reference here. Maybe even call `M_jk_transpose` for clarity.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 34 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQezNnN5dzB3ceZkypD:-NQfOGQ84SZUCHerDJ2K:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L34)):*

ah, this wasn't clear to me. If you go with this invariant, make sure the docs explicitly say this might not match the number of triplets provided at construction.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 44 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQezgUmDROaSX1A-d_J:-NQfOM2V5Jz8_rMUIHru:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L44)):*

repeated line
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc` line 71 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQf0tNq01ricXRVBs6v:-NQfPX-p5CyspoGrnX5-:b-dtkdrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/test/block_3x3_sparse_matrix_test.cc#L71)):*

since this snippet repeats several times, I'd advocate for something like `MakeArbitraryMatrix(int n, int m)`
</blockquote></details>

Done, thanks.

sammy-tri (commented):
:lgtm:

> Reviewed 1 of 4 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-:-NQg6et-8qo0aTXg4FTs:bnfp4nl)* status: 11 unresolved discussions (waiting on @amcastro-tri)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 1 unresolved discussion (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NR-OuYeBxjJ6YxF0zUH:b-4v90me) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

The transpose is implied in the indices. For non-transposed multiplication, it would be Mᵢₖ * Aₖⱼ instead.
</blockquote></details>

I still believe the transpose must be there. In your code, I think `M_ik` should actually be named `M_ki`

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 1 unresolved discussion (waiting on @amcastro-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NRYyQfE4ksKmmZjsKiu:b-ctnwaf) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

I still believe the transpose must be there. In your code, I think `M_ik` should actually be named `M_ki`
</blockquote></details>

The transpose notation `ᵀ` is not used in index notation. See, for example, [here](https://physics.stackexchange.com/questions/76640/how-do-you-show-from-the-index-notation-that-the-change-of-frame-formula-for-a-m).

I removed the subscript in individual blocks involved in the summation in an attempt to reduce confusion. PTAL.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: 1 unresolved discussion (waiting on @amcastro-tri and @sammy-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQ_lhg7BsYhlOIBEiVj:-NR_nXdDAcMmYxfqmZO3:b-agvkxv) ([raw file](https://github.com/RobotLocomotion/drake/blob/3716709369e046444cd003d18bf9f448065b3c97/multibody/contact_solvers/block_3x3_sparse_matrix.h#L69)):*

Ok, I'll proceed with the current name for this PR. We can discuss this more at the callsite in the followup PR.
</blockquote></details>

FYI:  My preference (in general) is unambiguous function names, notation,etc., even at the cost of longer names. 
However, let's proceed as you deem best.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),amcastro-tri (waiting on @sammy-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NRd2NsH9iSHrZ0KFPAC:b-q7onwx) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

Ok, then I misunderstood your notation and I wonder where things failed.
The whole time I thought you were indexing matrix blocks, not scalars (unlike the basic introductory text you referenced).
For matrix blocks, shouldn't the T stay? (I usually draw these in my head in case that clarifies my point)
</blockquote></details>

resolved per f2f. Comment above, even when it uses the same indexes names, referes to scalar entries. The code below refers to block entries. Thus my confusion.
Xuchen will fix notation to make them match. Thanks!

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19000)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),amcastro-tri (waiting on @amcastro-tri)

___
*[`multibody/contact_solvers/block_3x3_sparse_matrix.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19000#-NQeuBwKAq2UGflhESwB:-NRdR5XP6GT15IRxvRYy:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4625ecbc92d13e4a0f465a7f6ed6bf6effd953ff/multibody/contact_solvers/block_3x3_sparse_matrix.cc#L157)):*

resolved per f2f. Comment above, even when it uses the same indexes names, referes to scalar entries. The code below refers to block entries. Thus my confusion.
Xuchen will fix notation to make them match. Thanks!
</blockquote></details>

Done.

PR #18948, opened 2023-03-06
[parsing] Add package_downloader tool

jwnimmer-tri:
For now this is only unit tested; not installed nor used in earnest.

Towards #15774 and #9498.  You can see it used in action in #18955.

+@rpoyner-tri for feature review, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18948)

PR #19010, opened 2023-03-16
[parsing] Fix race conditions in the downloader

jwnimmer-tri:
Amends #18948.

Towards #15774 and #9498.

(This program wasn't used yet (outside of a little unit testing), so it's non-notable.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010)

### DIFF
--- a/multibody/parsing/BUILD.bazel
+++ b/multibody/parsing/BUILD.bazel
@@ -600,6 +600,12 @@ drake_py_unittest(
     ],
 )
 
+drake_py_unittest(
+    name = "package_downloader_stress_test",
+    data = ["package_downloader.py"],
+    tags = ["cpu:8"],
+)
+
 drake_cc_googletest(
     name = "package_map_test",
     data = [
--- a/multibody/parsing/package_downloader.py
+++ b/multibody/parsing/package_downloader.py
@@ -2,14 +2,16 @@
 program by hand; it is used by drake/multibody/parsing/package_map.cc to fetch
 remote packages on demand into a per-user local cache (e.g., ~/.cache/...).
 
-WARNING: This tool will recursively delete (and recreate) the `output_dir` as
-well as a few files and directories with a similar name but added extension.
-For example, with an output_dir of `~/.cache/drake/package_map/foo` the tool
-will delete and/or overwrite these files and/or directories:
+For example, with an output_dir of `~/.cache/drake/package_map/foo` this
+program will create this directory and an adjacent file:
 
- ~/.cache/drake/package_map/foo
- ~/.cache/drake/package_map/foo.tmp
+ ~/.cache/drake/package_map/foo/
  ~/.cache/drake/package_map/foo.README
+
+It is not an error for the output_dir to already exist when this program
+begins, nor for it to come into being at any point while this program is
+running. In other words, it is safe for two downloaders to be running at the
+same time with the same arguments.
 """
 
 # N.B. The ONLY packages we're allowed to use here are the Python standard
@@ -18,12 +20,14 @@ will delete and/or overwrite these files and/or directories:
 import hashlib
 import json
 import logging
-import os.path
+from pathlib import Path
+import os
 import urllib.parse
 import urllib.request as request
 import shutil
 import sys
 import tempfile
+from typing import List
 
 
 def _fail(message):
@@ -31,8 +35,8 @@ def _fail(message):
     sys.exit(1)
 
 
-def _run(*, temp_dir, package_name, urls, sha256, output_dir,
-         archive_type=None, strip_prefix=None):
+def _run(*, temp_dir: Path, package_name: str, urls: List[str], sha256: str,
+         output_dir: Path, archive_type: str = None, strip_prefix: str = None):
     """Runs the download and extract logic, assuming already-validated args.
 
     Args:
@@ -44,9 +48,9 @@ def _run(*, temp_dir, package_name, urls, sha256, output_dir,
         sha256: The cryptographic checksum of the file to be downloaded, as a
             64-character hexadecimal string.
         output_dir: The root directory to extract into. Its parent must exist,
-            but the output_dir itself must not already exist. This script will
-            create the output_dir only if the download and extraction was
-            successful, i.e., as a single atomic transaction.
+            but the output_dir itself generally will not already exist. This
+            script will create the output_dir if and only if the download and
+            extraction was successful, i.e., as a single atomic transaction.
         archive_type: The archive type of the downloaded file. When not given,
             the archive type is determined from the file extension of the URL.
         strip_prefix: A directory prefix to strip from the extracted files. If
@@ -54,24 +58,29 @@ def _run(*, temp_dir, package_name, urls, sha256, output_dir,
             and inaccessible.
     """
     # Add a README nearby to help explain the fetched data.
-    with open(f"{output_dir}.README", "wt", encoding="utf-8") as f:
-        f.write(f"Our sibling directory ./{os.path.basename(output_dir)} "
-                f"contains package://{package_name}.\n\n")
-        f.write("It was downloaded by Drake from one of these URLs:\n")
+    # To avoid stomping on concurrently-running download of the same package
+    # from another process, we need to write into a tempfile and then rename
+    # it into place as a final step. (Renames are atomic; writes are not.)
+    readme = output_dir.with_name(output_dir.name + ".README")
+    fd, temp_readme = tempfile.mkstemp(
+        dir=readme.parent,
+        prefix=f"temp.{readme.name}.")
+    temp_readme = Path(temp_readme)
+    with os.fdopen(fd, "w") as f:
+        f.write(f"Our sibling directory ./{output_dir.name} contains "
+                f"package://{package_name}.\n\n"
+                f"It was downloaded by Drake from one of these URLs:\n")
         for url in urls:
             f.write(f" {url}\n")
-
-    # Erase any spurious lingering data from a prior attempt at extraction.
-    output_dir_tmp = f"{output_dir}.tmp"
-    if os.path.exists(output_dir_tmp):
-        shutil.rmtree(output_dir_tmp)
+    # Fix the README's access permissions (to 'rw-r--r--' from 'rw-------').
+    temp_readme.chmod(0o644)
 
     # Try each url in turn.
     success = False
     errors = []
     for url in urls:
         basename = urllib.parse.urlparse(url).path.split("/")[-1] or "empty"
-        temp_filename = os.path.join(temp_dir, basename)
+        temp_filename = temp_dir / basename
         hasher = hashlib.sha256()
         with open(temp_filename, "wb") as f:
             try:
@@ -89,7 +98,6 @@ def _run(*, temp_dir, package_name, urls, sha256, output_dir,
         if download_sha256 == sha256:
             success = True
             break
-        os.remove(temp_filename)
         errors.append(
             f"Candidate {url} failed:\n"
             f"Checksum mismatch; was {download_sha256} but wanted {sha256}.")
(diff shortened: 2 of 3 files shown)


jwnimmer-tri:
+@xuchenhan-tri could I interest you in feature review?

jwnimmer-tri:
*[`multibody/parsing/package_downloader.py` line 137 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQfWeKPCeoa4JUVekbT:-NQfWeKPCeoa4JUVekbU:bg21pyk) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f48feb8ea4ed376667c9b0c3853c9effe80d55/multibody/parsing/package_downloader.py#L137)):*
> ```Python
>     finally:
>         try:
>             shutil.rmtree(output_dir_temp)
> ```

Hmm.  I can see now that I should have used `with tempfile.TemporaryDirectory` to own the `output_dir_temp`, to use the simpler auto-cleanup of the context manager, instead of my own try-finally.  I'll refactor it that way shortly.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/parsing/package_downloader.py` line 137 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQfWeKPCeoa4JUVekbT:-NQfYrFB2xf2Vr2cV_SJ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f48feb8ea4ed376667c9b0c3853c9effe80d55/multibody/parsing/package_downloader.py#L137)):*

Hmm.  I can see now that I should have used `with tempfile.TemporaryDirectory` to own the `output_dir_temp`, to use the simpler auto-cleanup of the context manager, instead of my own try-finally.  I'll refactor it that way shortly.
</blockquote></details>

Done.

xuchenhan-tri (commented):
Feature review done. The big picture is clear to me, but I was tripped on a few details. PTAL, thanks.

> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-:-NQgm8d47KSW-6bFT5MP:b-62q2yb)* status: 4 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`multibody/parsing/package_downloader.py` line 137 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQfWeKPCeoa4JUVekbT:-NQgjRme-25Bi9cvdsDu:bknupct) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f48feb8ea4ed376667c9b0c3853c9effe80d55/multibody/parsing/package_downloader.py#L137)):*

Done.
</blockquote></details>

I don't understand how we recover to a valid state if we ever get to this assertion. 

IIUC, if an exception occurs during `move_destination.rename(output_dir)`, then the rename doesn't happen and the downloaded content is still in that temp dir? How do we recover from that?
___
*[`multibody/parsing/package_downloader.py` line 73 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgeGHb513en2Nqucq9:-NQgeGHb513en2NqucqA:b-6irj4q) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/package_downloader.py#L73)):*
> ```Python
>         for url in urls:
>             f.write(f" {url}\n")
>     temp.chmod(0o644)
> ```

This is cryptic to me. Could you explain?
___
*[`multibody/parsing/package_downloader.py` line 135 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgl0Nb58BNfpYI39bD:-NQgl0Nb58BNfpYI39bE:blsw242) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/package_downloader.py#L135)):*
> ```Python
> 
>     # All done.
>     assert output_dir.is_dir()
> ```

What does this assert achieve? Isn't this the directory passed in by the caller?
___
*[`multibody/parsing/test/package_downloader_stress_test.py` line 52 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgiUDn8Cf0XWhAVObJ:-NQgiUDn8Cf0XWhAVObK:b-bh06e1) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/test/package_downloader_stress_test.py#L52)):*
> ```Python
>             # Launch child processes until we are running at max parallelism.
>             while (len(children) + 1 < cpus) and (launch_count < N):
>                 children.append(subprocess.Popen([
> ```

Are we guaranteed here that there are more than one process running at the same time? If the downloader finishes fast enough, could this still be serial?

jwnimmer-tri (commented):
+@SeanCurtis-TRI for platform review per schedule, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-:-NQgvh748sVzLBozL4t5:bbnvyd5)* status: 4 unresolved discussions, LGTM missing from assignees xuchenhan-tri(platform),SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI and @xuchenhan-tri)

___
*[`multibody/parsing/package_downloader.py` line 137 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQfWeKPCeoa4JUVekbT:-NQguv6oAaKbL3UGF9Ys:bno152w) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f48feb8ea4ed376667c9b0c3853c9effe80d55/multibody/parsing/package_downloader.py#L137)):*

I don't understand how we recover to a valid state if we ever get to this assertion. 

IIUC, if an exception occurs during `move_destination.rename(output_dir)`, then the rename doesn't happen and the downloaded content is still in that temp dir? How do we recover from that?
</blockquote></details>

Hopefully the new comments help?
___
*[`multibody/parsing/package_downloader.py` line 73 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgeGHb513en2Nqucq9:-NQgvNdF948SI25cjt1Y:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/package_downloader.py#L73)):*

This is cryptic to me. Could you explain?
</blockquote></details>

Done.
___
*[`multibody/parsing/package_downloader.py` line 135 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgl0Nb58BNfpYI39bD:-NQguBOY-a7mVysK4q_Y:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/package_downloader.py#L135)):*

What does this assert achieve? Isn't this the directory passed in by the caller?
</blockquote></details>

Done.
___
*[`multibody/parsing/test/package_downloader_stress_test.py` line 52 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQgiUDn8Cf0XWhAVObJ:-NQgvgYHArVxSxqCXKni:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf320c5ce5340cf0761c20618db823e3737b2652/multibody/parsing/test/package_downloader_stress_test.py#L52)):*

Are we guaranteed here that there are more than one process running at the same time? If the downloader finishes fast enough, could this still be serial?
</blockquote></details>

Done.

xuchenhan-tri (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-:-NQgwxCt-gu_PlTO4Q8e:bg7igkv)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/parsing/package_downloader.py` line 137 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQfWeKPCeoa4JUVekbT:-NQgwfRO5khn9awGdW1P:b7six7g) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f48feb8ea4ed376667c9b0c3853c9effe80d55/multibody/parsing/package_downloader.py#L137)):*

Hopefully the new comments help?
</blockquote></details>

Ah I see, thanks!

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI and @xuchenhan-tri)

___
*[`multibody/parsing/package_downloader.py` line 12 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQzQ5H96hpLD9dDrFud:-NQziJmB64xFJLQJoINC:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/12d5d0c053d0233a08f4e337e7831402b79a7354/multibody/parsing/package_downloader.py#L12)):*

BTW While this sentence is strictly true, it has some misleading implications. I expect the common inference would be to interpret this sentence as "no op" (or at least a reasonable approximation thereof). In reality:

1. The readme will definitely be overwritten/created.
2. The download logic (network activity, sha computation, and extraction) will be engaged. So, it's a lot of work to not change the contents of the `output_dir`.
3. The "silent exit" presumes success in the discarded download logic. If, for whatever reason, the download logic fails, we'll get an error logged.
</blockquote></details>

Done
___
*[`multibody/parsing/package_downloader.py` line 74 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQzNin5DxPg-W6P0pAT:-NQzjbWN6MKzidZ6jd6d:b-3st36u) ([raw file](https://github.com/RobotLocomotion/drake/blob/12d5d0c053d0233a08f4e337e7831402b79a7354/multibody/parsing/package_downloader.py#L74)):*

BTW What are the odds of two different attempts to download `foo` with different sets of urls (or in alternate ordering)? That would be the only case where a race condition on the creation of this file would be meaningfully apparent. (Also, I'm kind of surprised that the renaming isn't deferred until the successful download/extraction).
</blockquote></details>

It's not _supposed_ to happen, but I think you're right that we need to defend ourselves against confused users who call PackageMap with bogus arguments.
___
*[`multibody/parsing/package_downloader.py` line 118 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQzR59z-P1DsUJlQB9v:-NQziWTL7MXgSRi_VmlP:bw70yzr) ([raw file](https://github.com/RobotLocomotion/drake/blob/12d5d0c053d0233a08f4e337e7831402b79a7354/multibody/parsing/package_downloader.py#L118)):*

BTW Might be worth noting that the choice has been made to blindly move (instead of testing to see if our original temp directory was already on the common filesystem) as producing simpler code with no appreciable performance impact.

Otherwise, I fear that future readers of this code will think, "Why the additional move if we've already satisfied the condition?"
</blockquote></details>

In general, it's impossible to query whether two paths are on the same filesystem.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/package_downloader.py` line 118 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19010#-NQzR59z-P1DsUJlQB9v:-NQzqXkD7GQ1wwW2u2HF:b-qc8bck) ([raw file](https://github.com/RobotLocomotion/drake/blob/12d5d0c053d0233a08f4e337e7831402b79a7354/multibody/parsing/package_downloader.py#L118)):*

In general, it's impossible to query whether two paths are on the same filesystem.
</blockquote></details>

I hadn't been aware of that. That *particular* insight might be helpful so future readers don't waste their time thinking otherwise. How about:

```py
    # Now comes the really tricky part. Moving a directory is atomic but only
    # on the same filesystem (generally untestable)! Explicitly moving it from
    # the temporary directory into the output directory (under a temporary
    # name), satisfies the requirement for an atomic rename to the real name.
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19010)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri)

PR #19029, opened 2023-03-19
[parsing] Allow duplicate geometry names in URDF input

### LINKED ISSUE #19028: URDFs with duplicate geometry names
See http://sdformat.org/spec?ver=1.9&elem=visual for the SDFormat contract on geometry names.  The specification of `name` says:

>  Description: Unique name for the visual element within the scope of the parent link.

... i.e., that we're supposed to reject duplicate names.

To date, Drake has assumed that the same property must be true for URDF files as well.

However, if we compare the URDF spec for [joint names](http://wiki.ros.org/urdf/XML/joint) versus [geometry names](http://wiki.ros.org/urdf/XML/link#Elements) we see that joint names are specified as "a unique name" and geometry names are "a name for a part of a link's geometry" with no mention of uniqueness.  From that, I conclude that URDFs with duplicate geometry names are valid.

Apparently some URDF exporting tools generate files that look like this:
```xml
<?xml version="1.0" ?>
<robot name="...">
  <link name="link_0">
    <visual name="link_0">
      <origin xyz="..."/>
      <geometry>
        <mesh filename="link_0_0.obj"/>
      </geometry>
    </visual>
    <visual name="link_0">
      <origin xyz="..."/>
      <geometry>
        <mesh filename="link_0_1.obj"/>
      </geometry>
    </visual>
    ...
```

Note that the visual name is `link_0` for all meshes on that link.  Currently, Drake flags this as an error.

Given that the URDF specification does not mandate that geometry names must be unique, I think flagging an error here is too stern.

Instead, for URDF files we should demote this condition to a warning instead of error (by automatically salting geometry names in case there are duplicates).  Then users who turn on "strict" parsing mode would still get errors (because Drake's can't actually implement the file as-requested, due to the salting), but normal users would be able to proceed (modulo a warning message on their console).

jwnimmer-tri:
Closes #19028

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)

### DIFF
--- a/multibody/parsing/detail_urdf_geometry.cc
+++ b/multibody/parsing/detail_urdf_geometry.cc
@@ -354,32 +354,51 @@ std::unique_ptr<geometry::Shape> ParseGeometry(
   return {};
 }
 
-// The goal here is to invent a name that will be unique within the enclosing
-// body (i.e., link). To be as useful as possible, we'll use the shape name as
-// the geometry name, and then tack on a number if necessary to be unique.
-//
-// TODO(jwnimmer-tri) The "tack on a number" heuristic might fail if the user
-// mixes named and unnamed geometry within the same link and they happen to
-// use the shape name. We should probably try to handle this in the outer
-// parsing loop during ParseBody (by detecting the name collision). For now,
-// though, that doesn't seem worth dealing with yet.
-std::string MakeDefaultGeometryName(
-    const drake::internal::DiagnosticPolicy& policy,
-    const geometry::Shape& shape,
-    const std::unordered_set<std::string>& geometry_names,
-    int numeric_suffix_limit) {
-  const std::string shape_name = geometry::ShapeName(shape).name();
-  if (geometry_names.count(shape_name) == 0) {
-    return shape_name;
+// The goal here is to choose a name that will be unique within the enclosing
+// body (i.e., link). If the name is already unique, we return it unchanged.
+// Otherwise, we'll try to tack on a number until it is unique (or we run out
+// of numbers). If no name was given, we'll choose a default based on the
+// shape name with a uniqueifying integer tacked on.
+std::optional<std::string> ParseGeometryName(
+    const char* visual_or_collision, const TinyXml2Diagnostic& diagnostic,
+    const tinyxml2::XMLElement* node, const geometry::Shape& shape,
+    std::unordered_set<std::string>* geometry_names,
+    int numeric_name_suffix_limit) {
+  auto policy = diagnostic.MakePolicyForNode(node);
+
+  // Start with either the given xml name, or else the name of the shape type.
+  bool explicitly_named = false;
+  std::string result;
+  if (ParseStringAttribute(node, "name", &result)) {
+    explicitly_named = true;
+  } else {
+    result = geometry::ShapeName(shape).name();
   }
-  for (int i = 1; i < numeric_suffix_limit; ++i) {
-    std::string guess = fmt::format("{}{}", shape_name, i);
-    if (geometry_names.count(guess) == 0) {
-      return guess;
+
+  // Check if we need to salt it.
+  if (geometry_names->count(result) > 0) {
+    for (int i = 1;; ++i) {
+      if (i >= numeric_name_suffix_limit) {
+        policy.Error(fmt::format("Too many geometries with identical name '{}'",
+                                 result));
+        return {};
+      }
+      std::string guess = fmt::format("{}_{}", result, i);
+      if (geometry_names->count(guess) == 0) {
+        if (explicitly_named) {
+          policy.Warning(fmt::format(
+              "{} name '{}' has already been used, renaming to '{}' instead",
+              visual_or_collision, result, guess));
+        }
+        result = guess;
+        break;
+      }
     }
   }
-  policy.Error("Too many identical geometries with default names.");
-  return {};
+
+  // Success.
+  geometry_names->insert(result);
+  return result;
 }
 
 }  // namespace
@@ -465,20 +484,15 @@ std::optional<geometry::GeometryInstance> ParseVisual(
     properties.AddProperty("renderer", "accepting", std::move(accepting_names));
   }
 
-  std::string geometry_name;
-  if (!ParseStringAttribute(node, "name", &geometry_name)) {
-    geometry_name = MakeDefaultGeometryName(diagnostic.MakePolicyForNode(node),
-                                            *shape, *geometry_names,
-                                            numeric_name_suffix_limit);
-    if (geometry_name.empty()) {
-      // Error should already have been emitted.
-      return {};
-    }
+  std::optional<std::string> geometry_name =
+      ParseGeometryName("visual", diagnostic, node, *shape, geometry_names,
+                        numeric_name_suffix_limit);
+  if (!geometry_name) {
+    return {};
   }
-  geometry_names->insert(geometry_name);
 
   auto instance = geometry::GeometryInstance(T_element_to_link,
-                                             std::move(shape), geometry_name);
+                                             std::move(shape), *geometry_name);
   instance.set_illustration_properties(properties);
   return instance;
 }
@@ -701,20 +715,15 @@ std::optional<geometry::GeometryInstance> ParseCollision(
     }
   }
 
-  std::string geometry_name;
-  if (!ParseStringAttribute(node, "name", &geometry_name)) {
-    geometry_name = MakeDefaultGeometryName(diagnostic.MakePolicyForNode(node),
-                                            *shape, *geometry_names,
-                                            numeric_name_suffix_limit);
-    if (geometry_name.empty()) {
-      // Error should already have been emitted.
-      return {};
-    }
+  std::optional<std::string> geometry_name =
+      ParseGeometryName("collision", diagnostic, node, *shape, geometry_names,
+                        numeric_name_suffix_limit);
+  if (!geometry_name) {
+    return {};
   }
-  geometry_names->insert(geometry_name);
 
   geometry::GeometryInstance instance(T_element_to_link, std::move(shape),
-                                      geometry_name);
+                                      *geometry_name);
   instance.set_proximity_properties(std::move(props));
   return instance;
 }
--- a/multibody/parsing/detail_urdf_geometry.h
+++ b/multibody/parsing/detail_urdf_geometry.h
@@ -139,9 +139,10 @@ constexpr int kDefaultNumericSuffixLimit = 10000;
  which are specified by both color and name will be added to the map and
  can be used by later visual elements.  Material definitions may be
  repeated if the material properties are identical.
- @param[in,out] geometry_names The list of geometry names already used within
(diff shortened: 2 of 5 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/test/detail_urdf_geometry_test.cc` line 782 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQvOATP8yKoRNHP3kO1:-NQvOATP8yKoRNHP3kO2:btubjuk) ([raw file](https://github.com/RobotLocomotion/drake/blob/d9ee19ee60ef51c7d1950f2b6da280d495ff0ded/multibody/parsing/test/detail_urdf_geometry_test.cc#L782)):*
> ```cpp
>   ASSERT_EQ(collision_instances_.size(), 1);
>   EXPECT_EQ(visual_instances_[0].name(), "box");
>   EXPECT_EQ(collision_instances_[0].name(), "box2");
> ```

Should visual vs collision share a uniquefication domain, or should each one be its own domain?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/parsing/test/detail_urdf_geometry_test.cc` line 782 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQvOATP8yKoRNHP3kO1:-NQvu6n6556nNi6bgJVA:b-3fa3dt) ([raw file](https://github.com/RobotLocomotion/drake/blob/d9ee19ee60ef51c7d1950f2b6da280d495ff0ded/multibody/parsing/test/detail_urdf_geometry_test.cc#L782)):*

Should visual vs collision share a uniquefication domain, or should each one be its own domain?
</blockquote></details>

Done.  Name uniqueness should be within a single role, so this is tested at the URDF layer now, not the geometry utilities.

SeanCurtis-TRI (commented):
:LGTM: with some small clarity quibbles.

> Reviewed 1 of 2 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-:-NQz_2Kp59maUVF-h5g2:bn02y37)* status: 3 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_geometry.h` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzV5yZE8M-FlBYB4iz:-NQzV5yZE8M-FlBYB4j-:b4atlqp) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/detail_urdf_geometry.h#L143)):*
> ```objc
>  repeated if the material properties are identical.
>  @param[in,out] geometry_names The list of geometry names already used within
>  the current MbP body (i.e., link) with the illustration role, so that this
> ```

nit: "...with the illustration and perception roles".

We currently have no facility to distinguish the two roles during parsing, so both get set. Admittedly, it happens via a tortuous mechanism (in which MbP promotes illustration to illustration + perception). So, although this code doesn't reference PerceptionProperties, it is still currently written with that end in mind.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzY6fRB9jc-JyY70GY:-NQzY6fRB9jc-JyY70GZ:b-h2t8t6) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/detail_urdf_parser.cc#L265)):*
BTW Worth putting a note here about *why* this clear is here? 

> SceneGraph supports overlapping names for different roles; this segregates geometry names implicitly along those lines.

(Or some such thing.)
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzZCTnEvLAVwBk5uTj:-NQzZCTnEvLAVwBk5uTk:b8iixdn) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/test/detail_urdf_parser_test.cc#L1102)):*
> ```cpp
> }
> 
> TEST_F(UrdfParserTest, TestDuplicateVisualName) {
> ```

nit: Two thoughts:

1. What about the same test for collision?
2. What about the case where a collision and visual geometry share the same name? Both should be instantiated verbatim. That is dependent on the `geometry_names.clear();`(whose significance doesn't get its full weight without such a supporting test). In fact, these tests wouldn't fail if you removed the line.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 3 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzZCTnEvLAVwBk5uTj:-NQza3DqA_o_ZAMlr2vr:b-pbl2el) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/test/detail_urdf_parser_test.cc#L1102)):*

nit: Two thoughts:

1. What about the same test for collision?
2. What about the case where a collision and visual geometry share the same name? Both should be instantiated verbatim. That is dependent on the `geometry_names.clear();`(whose significance doesn't get its full weight without such a supporting test). In fact, these tests wouldn't fail if you removed the line.
</blockquote></details>

Oops, I somehow used the wrong test case name here, which I think confused you.

This is supposed to be the "visual vs collision" test with no warnings, not "duplicate visual".

(1) That's already covered in the other test program.

(2) That's exactly what this test is.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_urdf_geometry.h` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzV5yZE8M-FlBYB4iz:-NQzfgo27jWgpBINz3bV:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/detail_urdf_geometry.h#L143)):*

nit: "...with the illustration and perception roles".

We currently have no facility to distinguish the two roles during parsing, so both get set. Admittedly, it happens via a tortuous mechanism (in which MbP promotes illustration to illustration + perception). So, although this code doesn't reference PerceptionProperties, it is still currently written with that end in mind.
</blockquote></details>

Done.
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzZCTnEvLAVwBk5uTj:-NQzfKtB8KKxEeLvcz-O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/test/detail_urdf_parser_test.cc#L1102)):*

Oops, I somehow used the wrong test case name here, which I think confused you.

This is supposed to be the "visual vs collision" test with no warnings, not "duplicate visual".

(1) That's already covered in the other test program.

(2) That's exactly what this test is.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzZCTnEvLAVwBk5uTj:-NQzprAM6Y2Xa4jbrsUq:b-t7ayc7) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc0d547de00b763f5fa6228cafcdda03837ece2/multibody/parsing/test/detail_urdf_parser_test.cc#L1102)):*

Done.
</blockquote></details>

You are correct. THe test named lulled my attention. Still, it led to an improvement, even if by accident.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @jwnimmer-tri)

___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzr6S18VBGfXWAppaC):*
This numbering scheme does not match 153f3fb18b8c2f5203ffb0531e96af75610c5040, nor does it match the scheme discussed [in slack](https://drakedevelopers.slack.com/archives/C2NN588UA/p1677769925341499). Did something change?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-NQzr6S18VBGfXWAppaC):*

This numbering scheme does not match 153f3fb18b8c2f5203ffb0531e96af75610c5040, nor does it match the scheme discussed [in slack](https://drakedevelopers.slack.com/archives/C2NN588UA/p1677769925341499). Did something change?
</blockquote></details>

Nope.  I just completely misremembered what the consensus was.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
:lgtm:

> Reviewed 3 of 3 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19029#-:-NR-L9jPAqbq52QsvZJz:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

PR #19040, opened 2023-03-21
[math] Move code from rotation_matrix.h to cc

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)

### DIFF
--- a/math/rotation_matrix.cc
+++ b/math/rotation_matrix.cc
@@ -55,7 +55,7 @@ void ThrowIfAnyElementInQuaternionIsInfinityOrNaN(
 template <typename T>
 RotationMatrix<T> RotationMatrix<T>::MakeXRotation(const T& theta) {
   using std::isfinite;
-  DRAKE_DEMAND(isfinite(theta));
+  DRAKE_THROW_UNLESS(isfinite(theta));
   RotationMatrix<T> R(internal::DoNotInitializeMemberFields{});
   using std::cos;
   using std::sin;
@@ -71,7 +71,7 @@ RotationMatrix<T> RotationMatrix<T>::MakeXRotation(const T& theta) {
 template <typename T>
 RotationMatrix<T> RotationMatrix<T>::MakeYRotation(const T& theta) {
   using std::isfinite;
-  DRAKE_DEMAND(isfinite(theta));
+  DRAKE_THROW_UNLESS(isfinite(theta));
   RotationMatrix<T> R(internal::DoNotInitializeMemberFields{});
   using std::cos;
   using std::sin;
@@ -87,7 +87,7 @@ RotationMatrix<T> RotationMatrix<T>::MakeYRotation(const T& theta) {
 template <typename T>
 RotationMatrix<T> RotationMatrix<T>::MakeZRotation(const T& theta) {
   using std::isfinite;
-  DRAKE_DEMAND(isfinite(theta));
+  DRAKE_THROW_UNLESS(isfinite(theta));
   RotationMatrix<T> R(internal::DoNotInitializeMemberFields{});
   using std::cos;
   using std::sin;
--- a/math/rotation_matrix.cc
+++ b/math/rotation_matrix.cc
@@ -397,6 +397,40 @@ Vector3<T> RotationMatrix<T>::NormalizeOrThrow(const Vector3<T>& v,
   return u;
 }
 
+template <typename T>
+RotationMatrix<T> RotationMatrix<T>::InvertAndCompose(
+    const RotationMatrix<T>& other) const {
+  const RotationMatrix<T>& R_AC = other;  // Nicer name.
+  RotationMatrix<T> R_BC(internal::DoNotInitializeMemberFields{});
+  if constexpr (std::is_same_v<T, double>) {
+    internal::ComposeRinvR(*this, R_AC, &R_BC);
+  } else {
+    const RotationMatrix<T> R_BA = inverse();
+    R_BC = R_BA * R_AC;
+  }
+  return R_BC;
+}
+
+template <typename T>
+Eigen::Quaternion<T> RotationMatrix<T>::ToQuaternion(
+    const Eigen::Ref<const Matrix3<T>>& M) {
+  Eigen::Quaternion<T> q = RotationMatrixToUnnormalizedQuaternion(M);
+
+  // Since the quaternions q and -q correspond to the same rotation matrix,
+  // choose to return a canonical quaternion, i.e., with q(0) >= 0.
+  const T canonical_factor = if_then_else(q.w() < 0, T(-1), T(1));
+
+  // The quantity q calculated thus far in this algorithm is not a quaternion
+  // with magnitude 1.  It differs from a quaternion in that all elements of
+  // q are scaled by the same factor. To return a valid quaternion, q must be
+  // normalized so q(0)^2 + q(1)^2 + q(2)^2 + q(3)^2 = 1.
+  const T scale = canonical_factor / q.norm();
+  q.coeffs() *= scale;
+
+  DRAKE_ASSERT_VOID(ThrowIfNotValid(QuaternionToRotationMatrix(q, T(2))));
+  return q;
+}
+
 }  // namespace math
 }  // namespace drake
 
--- a/math/rotation_matrix.cc
+++ b/math/rotation_matrix.cc
@@ -53,6 +53,54 @@ void ThrowIfAnyElementInQuaternionIsInfinityOrNaN(
 }  // namespace
 
 template <typename T>
+RotationMatrix<T> RotationMatrix<T>::MakeXRotation(const T& theta) {
+  using std::isfinite;
+  DRAKE_DEMAND(isfinite(theta));
+  RotationMatrix<T> R = RotationMatrix<T>::MakeUninitialized();
+  using std::cos;
+  using std::sin;
+  const T c = cos(theta), s = sin(theta);
+  // clang-format off
+  R.R_AB_ << 1,  0,  0,
+             0,  c, -s,
+             0,  s,  c;
+  // clang-format on
+  return R;
+}
+
+template <typename T>
+RotationMatrix<T> RotationMatrix<T>::MakeYRotation(const T& theta) {
+  using std::isfinite;
+  DRAKE_DEMAND(isfinite(theta));
+  RotationMatrix<T> R = RotationMatrix<T>::MakeUninitialized();
+  using std::cos;
+  using std::sin;
+  const T c = cos(theta), s = sin(theta);
+  // clang-format off
+  R.R_AB_ <<  c,  0,  s,
+              0,  1,  0,
+             -s,  0,  c;
+  // clang-format on
+  return R;
+}
+
+template <typename T>
+RotationMatrix<T> RotationMatrix<T>::MakeZRotation(const T& theta) {
+  using std::isfinite;
+  DRAKE_DEMAND(isfinite(theta));
+  RotationMatrix<T> R = RotationMatrix<T>::MakeUninitialized();
+  using std::cos;
+  using std::sin;
+  const T c = cos(theta), s = sin(theta);
+  // clang-format off
+  R.R_AB_ << c, -s,  0,
+             s,  c,  0,
+             0,  0,  1;
+  // clang-format on
+  return RotationMatrix(R);
+}
+
+template <typename T>
 RotationMatrix<T>::RotationMatrix(const RollPitchYaw<T>& rpy) {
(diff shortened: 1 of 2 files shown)


mitiguy (commented):
Feature review +@jwnimmer-tri 
Perhaps you can also platform review too? 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-:-NRAv8c_4m0xGjDSRTQH:bo8v0f8)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`math/rotation_matrix.h` line 636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAwJf11zOH0fmW1EGl:-NRAwJf2-MRYAQ5i9ROj:buivzcd) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.h#L636)):*
> ```objc
>   /// (Internal use only) Constructs a RotationMatrix without initializing the
>   /// underlying 3x3 matrix. Here for use by RigidTransform but no one else.
>   explicit RotationMatrix(internal::DoNotInitializeMemberFields) {}
> ```

We already have a "do not init" constructor.  Why do we need the new "bool" one 20 lines below?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`math/rotation_matrix.cc` line 454 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAybJp7olufRMhBbxE:-NRAybJp7olufRMhBbxF:b-2htwdf) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.cc#L454)):*
> ```cpp
>   RotationMatrix<T> R_BC(internal::DoNotInitializeMemberFields{});
>   if constexpr (std::is_same_v<T, double>) {
>     internal::ComposeRinvR(*this, R_AC, &R_BC);
> ```

@sherm1 please check this.

My understanding of your work is that this ComposeRintR really needs to remain in the header file.

I'd certainly buy (and prefer) having the `else` body live in the cc file, but I'm not sure about the `if` body.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri and @mitiguy)

___
*[`math/rotation_matrix.h` line 413 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NREPHxwCrgnGG3LbI7G:-NREPHxwCrgnGG3LbI7H:b-5rqa31) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.h#L413)):*
> ```objc
>   /// @note It is possible (albeit improbable) to create an invalid rotation
>   /// matrix by accumulating round-off error with a large number of multiplies.
>   RotationMatrix<T> InvertAndCompose(const RotationMatrix<T>& other) const;
> ```

minor: this method should remain inline
___
*[`math/rotation_matrix.cc` line 454 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAybJp7olufRMhBbxE:-NREODX07VZ5lll3XVQM:bn6brny) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.cc#L454)):*

@sherm1 please check this.

My understanding of your work is that this ComposeRintR really needs to remain in the header file.

I'd certainly buy (and prefer) having the `else` body live in the cc file, but I'm not sure about the `if` body.
</blockquote></details>

Thanks for the heads-up! Yes, these should remain inline. It's not that they wouldn't function correctly like this, but that the whole point of them is to speed up these measured hot-spot operations. Adding unnecessary linkage overhead partially defeats that purpose.

All of these methods meet our standard for inline functions -- consider the important case where T==double. This method is literally a single executable line, just the call to ComposeRinvR(), with return value optimization eliminating the temporary variable.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`math/rotation_matrix.cc` line 454 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAybJp7olufRMhBbxE:-NREQNRW5Vjwhh3KJTKK:b-48puhy) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.cc#L454)):*

Thanks for the heads-up! Yes, these should remain inline. It's not that they wouldn't function correctly like this, but that the whole point of them is to speed up these measured hot-spot operations. Adding unnecessary linkage overhead partially defeats that purpose.

All of these methods meet our standard for inline functions -- consider the important case where T==double. This method is literally a single executable line, just the call to ComposeRinvR(), with return value optimization eliminating the temporary variable.
</blockquote></details>

When you say "all of" do you mean

"all of the things we call within InvertAndCompose like inverse and operator-mul"

or do you mean

"all of the changes in this PR"

?

Because to me, `ToQuaternion` does _not_ meet the definition of okay-to-inline, for example.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri and @mitiguy)

___
*[`math/rotation_matrix.cc` line 454 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAybJp7olufRMhBbxE:-NREYlVY1pgE1yQVWh6I:b-3wj06u) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.cc#L454)):*

When you say "all of" do you mean

"all of the things we call within InvertAndCompose like inverse and operator-mul"

or do you mean

"all of the changes in this PR"

?

Because to me, `ToQuaternion` does _not_ meet the definition of okay-to-inline, for example.
</blockquote></details>

Apologies for ambiguity. By "all of" I meant the four methods that use bespoke AVX functions for speed (two of those are in RotationMatrix and two in RigidTransform). Only one of those four appears to be relocated in this PR and that's the only thing I would like to see moved back to inline. I agree that the other relocations here are good.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`math/rotation_matrix.cc` line 454 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRAybJp7olufRMhBbxE:-NRE_0M049oFnxOmvf-H:b2xlyuj) ([raw file](https://github.com/RobotLocomotion/drake/blob/6a188a1816eec5d27d6944a3aab818c6208e8f1b/math/rotation_matrix.cc#L454)):*

Apologies for ambiguity. By "all of" I meant the four methods that use bespoke AVX functions for speed (two of those are in RotationMatrix and two in RigidTransform). Only one of those four appears to be relocated in this PR and that's the only thing I would like to see moved back to inline. I agree that the other relocations here are good.
</blockquote></details>

Thanks!  Fully agree.

jwnimmer-tri (commented):
:lgtm: both reviews.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-:-NRF7jiO9v80dWbzIVnh:b9krqcj)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRF7m-00JmaHIQpk8Lk):*
I'll plan to fix the commit message when I squash, but in case you beat me to it:

This PR does more than just move some code.  It introduces sanity-checks on `theta`.  We should mention that in the commit message, since that's the actual user-visible effect of this change.
___
*[`math/rotation_matrix.cc` line 58 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRF7OTw3Fm-Z86cs9bV:-NRF7OTw3Fm-Z86cs9bW:bfdu7n7) ([raw file](https://github.com/RobotLocomotion/drake/blob/96ed675cff895c90be489e19220eddfd71e47891/math/rotation_matrix.cc#L58)):*
> ```cpp
> RotationMatrix<T> RotationMatrix<T>::MakeXRotation(const T& theta) {
>   using std::isfinite;
>   DRAKE_DEMAND(isfinite(theta));
> ```

nit Nowhere else in this class aborts the problem if the user provides wrong input.  We should maintain that convention in the new code here as well.

Likewise the other two MakeRotation factory functions.



_Suggestion:_
```cpp
DRAKE_THROW_UNLESS(isfinite(theta));
```

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRF7m-00JmaHIQpk8Lk):*

I'll plan to fix the commit message when I squash, but in case you beat me to it:

This PR does more than just move some code.  It introduces sanity-checks on `theta`.  We should mention that in the commit message, since that's the actual user-visible effect of this change.
</blockquote></details>

Possible commit message, something like:
Move code from rotation_matrix.h to rotation_matrix.cc. Sanity-check theta arg to MakeRotationMatrixZ(), etc.
___
*[`math/rotation_matrix.cc` line 58 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRF7OTw3Fm-Z86cs9bV:-NRFpEfIA9j3vMdW4D-y:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/96ed675cff895c90be489e19220eddfd71e47891/math/rotation_matrix.cc#L58)):*

nit Nowhere else in this class aborts the problem if the user provides wrong input.  We should maintain that convention in the new code here as well.

Likewise the other two MakeRotation factory functions.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19040)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19040#-NRF7m-00JmaHIQpk8Lk):*

Possible commit message, something like:
Move code from rotation_matrix.h to rotation_matrix.cc. Sanity-check theta arg to MakeRotationMatrixZ(), etc.
</blockquote></details>

OK, I'll use that as inspiration -- but with a subject line less than the 63-character upper limit.

EricCousineau-TRI:
Setting to `release notes: none` as this may not be relevant to users.

EricCousineau-TRI:
(but still a great change ;) )

jwnimmer-tri:
The release note should be the second sentence: "Sanity-check the theta arg to MakeRotationMatrixZ(), etc."

PR #19046, opened 2023-03-21
[parsing] Switch drake_models to a remote package

jwnimmer-tri:
Closes #15774.

---

Local testing of a CMake install:

```console
$ bazel run //:install -- /home/jwnimmer/tmp/install
$ cd ~/tmp
$ env PYTHONPATH=install/lib/python3.10/site-packages python3
>>> from pydrake.all import PackageMap
>>> PackageMap().GetPath("drake_models")
INFO:drake:PackageMap: Downloading https://github.com/RobotLocomotion/models/archive/6751ba850fd46fee789a095edba7ad9fc99cd188.tar.gz
'/home/jwnimmer/.cache/drake/package_map/dc38658f7acb4d07442e6dae4d18818d31ac6cc6d66bf16913ef1b208692627c'
```

---

Local testing of an Ubuntu Wheel install:

```console
$ cd tmp
$ python3 -m venv env && env/bin/pip install 'https://drake-packages.csail.mit.edu/drake/experimental/drake-0.0.2023.3.24.16.25.24%2Bgit0cd06a59-cp310-cp310-manylinux_2_31_x86_64.whl'
$ env/bin/python3
>>> from pydrake.all import *
>>> path = PackageMap().GetPath("drake_models")
INFO:drake:PackageMap: Downloading https://github.com/RobotLocomotion/models/archive/6751ba850fd46fee789a095edba7ad9fc99cd188.tar.gz
>>> with open(f"{path}/README.md") as f: f.read()
'# models\n\nShareable large model files ...'
```

---

Local testing of a macOS Wheel install:

```console
% cd tmp
% python3 -m venv env && env/bin/pip install 'https://drake-packages.csail.mit.edu/drake/experimental/drake-0.0.2023.3.24.12.26.16%2Bgit0cd06a59-cp311-cp311-macosx_12_0_arm64.whl'
% env/bin/python3
>>> from pydrake.all import *
>>> path = PackageMap().GetPath("drake_models")
INFO:drake:PackageMap: Downloading https://github.com/RobotLocomotion/models/archive/6751ba850fd46fee789a095edba7ad9fc99cd188.tar.gz
>>> with open(f"{path}/README.md") as f: f.read()
... 
'# models\n\nShareable large model files ...'
```

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19046)

PR #19050, opened 2023-03-22
Add DRAKE_ALLOW_NETWORK governor

jwnimmer-tri:
Towards #15774 and #19046.  I think it's important to have our test-fixtures fail fast when PackageMap is going to hit the internet, before we can land #19046.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)

### DIFF
--- a/bindings/pydrake/geometry/BUILD.bazel
+++ b/bindings/pydrake/geometry/BUILD.bazel
@@ -110,6 +110,7 @@ drake_py_unittest(
 
 drake_py_unittest(
     name = "render_test",
+    allow_network = ["render_gltf_client"],
     deps = [
         ":geometry",
         "//bindings/pydrake/common/test_utilities",
--- a/common/BUILD.bazel
+++ b/common/BUILD.bazel
@@ -42,6 +42,7 @@ drake_cc_package_library(
         ":is_cloneable",
         ":is_less_than_comparable",
         ":name_value",
+        ":network_policy",
         ":nice_type_name",
         ":pointer_cast",
         ":polynomial",
@@ -349,6 +350,15 @@ drake_cc_library(
 drake_cc_library(
     name = "name_value",
     hdrs = ["name_value.h"],
+    deps = [
+        ":essential",
+    ],
+)
+
+drake_cc_library(
+    name = "network_policy",
+    srcs = ["network_policy.cc"],
+    hdrs = ["network_policy.h"],
     deps = [
         ":essential",
     ],
@@ -955,6 +965,13 @@ drake_cc_googletest(
     name = "name_value_test",
     deps = [
         ":name_value",
+    ],
+)
+
+drake_cc_googletest(
+    name = "network_policy_test",
+    deps = [
+        ":network_policy",
     ],
 )
 
--- a/common/network_policy.cc
+++ b/common/network_policy.cc
@@ -1,0 +1,75 @@
+#include "drake/common/network_policy.h"
+
+#include <algorithm>
+#include <cstdlib>
+
+#include "drake/common/drake_throw.h"
+#include "drake/common/text_logging.h"
+
+namespace drake {
+namespace internal {
+namespace {
+
+bool IsAsciiLowercaseAlphaNumeric(std::string_view word) {
+  return std::all_of(word.begin(), word.end(), [](char ch) {
+    return ('a' <= ch && ch <= 'z') || ('0' <= ch && ch <= '9') || (ch == '_');
+  });
+}
+
+}  // namespace
+
+bool IsNetworkingAllowed(std::string_view component) {
+  DRAKE_THROW_UNLESS(component.length() > 0);
+  DRAKE_THROW_UNLESS(component != "none");
+  DRAKE_THROW_UNLESS(IsAsciiLowercaseAlphaNumeric(component));
+
+  // Unset or empty means allow-all.
+  const char* const env_cstr = std::getenv("DRAKE_ALLOW_NETWORK");
+  if (env_cstr == nullptr) {
+    return true;
+  }
+  if (*env_cstr == '\0') {
+    return true;
+  }
+  const std::string_view env_view{env_cstr};
+
+  // Set to "none" means deny-all.
+  if (env_view == "none") {
+    return false;
+  }
+
+  // Iterate over the colon-delimited tokens.
+  // N.B. We purposefully do not warn for unknown tokens because they may evolve
+  // over time and we don't want to force users to churn their policy variables
+  // to be congruent with their Drake version pin.
+  // TODO(jwnimmer-tri) As of C++20, use std::ranges::lazy_split_view.
+  bool match = false;
+  std::string_view worklist = env_view;
+  while (!worklist.empty()) {
+    std::string_view token;
+    auto delim = worklist.find(':');
+    if (delim == std::string_view::npos) {
+      token = worklist;
+      worklist = {};
+    } else {
+      token = worklist.substr(0, delim);
+      worklist.remove_prefix(delim + 1);
+    }
+    if (token == "none") {
+      static const logging::Warn log_once(
+          "Setting DRAKE_ALLOW_NETWORK={} combines 'none' with non-none "
+          "values; this is probably not what you wanted! The effect is "
+          "the same as just saying 'none' on its own; nothing is allowed!",
+          env_view);
+      return false;
+    }
+    if (token == component) {
+      match = true;
+    }
+  }
+
+  return match;
+}
+
+}  // namespace internal
+}  // namespace drake
--- a/common/network_policy.h
+++ b/common/network_policy.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <string_view>
+
+namespace drake {
+
+/** @addtogroup environment_variables
+@{
+@defgroup allow_network DRAKE_ALLOW_NETWORK
+
+Users can set the environment variable `DRAKE_ALLOW_NETWORK` to a
+colon-separated list to limit which components are permitted network access.
+
+If `DRAKE_ALLOW_NETWORK` is unset or set to the empty string, then all
+networking is allowed.
+
+The component values supported by Drake are:
+- <b>lcm</b> (see drake::lcm::DrakeLcm)
+- <b>meshcat</b> (see drake::geometry::Meshcat)
+- <b>package_map</b> (see drake::multibody::PackageMap)
+- <b>render_gltf_client</b> (see drake::geometry::MakeRenderEngineVtk())
+
+... as well as the special value <b>none</b>.
+
+The environment variable only governs the *network* access of the component.
+If the component also supports other mechanisms (e.g., files or memory buffers)
+those are unaffected.
+
+---
+
+For example, to allow only lcm and meshcat, run this command in your terminal:
+```sh
+export DRAKE_ALLOW_NETWORK=lcm:meshcat
+```
+
+To disable all networking, run this command in your terminal:
+```sh
+export DRAKE_ALLOW_NETWORK=none
+```
+@} */
+
+namespace internal {
+
+/* Returns true iff the given component is allowed to access the network based
+on the current value of the DRAKE_ALLOW_NETWORK environment variable.
+@param component is the name to query, which must be alphanumeric lowercase
+(underscores allowed) and must not be "none" nor the empty string. */
+bool IsNetworkingAllowed(std::string_view component);
+
+}  // namespace internal
+}  // namespace drake
--- a/common/test/network_policy_test.cc
+++ b/common/test/network_policy_test.cc
@@ -1,0 +1,64 @@
+#include "drake/common/network_policy.h"
+
+#include <stdlib.h>
+
+#include <gtest/gtest.h>
+
+namespace drake {
+namespace internal {
+namespace {
+
+constexpr const char kAllow[] = "DRAKE_ALLOW_NETWORK";
+
+GTEST_TEST(NetworkPolicyTest, Missing) {
+  ASSERT_EQ(::unsetenv(kAllow), 0);
+  EXPECT_TRUE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, Empty) {
+  ASSERT_EQ(::setenv(kAllow, "", 1), 0);
+  EXPECT_TRUE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, None) {
+  ASSERT_EQ(::setenv(kAllow, "none", 1), 0);
+  EXPECT_FALSE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, MixedNone) {
+  // N.B. This prints a spdlog warning.
+  ASSERT_EQ(::setenv(kAllow, "none:foo", 1), 0);
+  EXPECT_FALSE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, YesMatch) {
+  ASSERT_EQ(::setenv(kAllow, "foo", 1), 0);
+  EXPECT_TRUE(IsNetworkingAllowed("foo"));
+
+  ASSERT_EQ(::setenv(kAllow, "foo:bar", 1), 0);
+  EXPECT_TRUE(IsNetworkingAllowed("foo"));
+
+  ASSERT_EQ(::setenv(kAllow, "bar:foo:baz", 1), 0);
+  EXPECT_TRUE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, NoMatch) {
+  ASSERT_EQ(::setenv(kAllow, "bar", 1), 0);
+  EXPECT_FALSE(IsNetworkingAllowed("foo"));
+
+  ASSERT_EQ(::setenv(kAllow, "bar:baz", 1), 0);
+  EXPECT_FALSE(IsNetworkingAllowed("foo"));
+
+  ASSERT_EQ(::setenv(kAllow, ":bar:baz:", 1), 0);
+  EXPECT_FALSE(IsNetworkingAllowed("foo"));
+}
+
+GTEST_TEST(NetworkPolicyTest, BadName) {
+  EXPECT_THROW(IsNetworkingAllowed(""), std::exception);
+  EXPECT_THROW(IsNetworkingAllowed("none"), std::exception);
+  EXPECT_THROW(IsNetworkingAllowed("LCM"), std::exception);
+}
+
+}  // namespace
+}  // namespace internal
+}  // namespace drake
--- a/doc/doxygen_cxx/doxygen.h
+++ b/doc/doxygen_cxx/doxygen.h
@@ -216,6 +216,7 @@ namespace solvers {
 
  This section provides an inventory of environment variables relevant to Drake.
 
+ - <b>\ref allow_network "DRAKE_ALLOW_NETWORK"</b>
  - <b>\ref drake::common::FindResource() "DRAKE_RESOURCE_ROOT"</b>
  - <b>\ref drake::solvers::GurobiSolver "GRB_LICENSE_FILE"</b> (see also 
    <a href="https://support.gurobi.com/hc/en-us/articles/360013417211-Where-do-I-place-the-Gurobi-license-file-gurobi-lic-">upstream documentation</a>)
--- a/examples/allegro_hand/joint_control/BUILD.bazel
+++ b/examples/allegro_hand/joint_control/BUILD.bazel
@@ -54,6 +54,7 @@ drake_cc_binary(
 drake_py_unittest(
     name = "run_twisting_mug_test",
     timeout = "moderate",
+    allow_network = ["lcm"],
     args = select({
         "//tools/cc_toolchain:debug": ["--compilation_mode=dbg"],
         "//conditions:default": ["--compilation_mode=opt"],
--- a/geometry/BUILD.bazel
+++ b/geometry/BUILD.bazel
@@ -467,6 +467,7 @@ drake_cc_library(
     deps = [
         "//common:drake_export",
         "//common:find_resource",
+        "//common:network_policy",
         "//common:scope_exit",
         "//common:unused",
         "@common_robotics_utilities",
@@ -521,6 +522,15 @@ drake_cc_googletest(
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
         "@msgpack_internal//:msgpack",
+    ],
+)
+
+drake_cc_googletest(
+    name = "meshcat_denied_test",
+    allow_network = ["none"],
+    deps = [
+        ":meshcat",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/geometry/meshcat.cc
+++ b/geometry/meshcat.cc
@@ -27,6 +27,7 @@
 #include "drake/common/drake_export.h"
 #include "drake/common/drake_throw.h"
 #include "drake/common/find_resource.h"
+#include "drake/common/network_policy.h"
 #include "drake/common/never_destroyed.h"
 #include "drake/common/scope_exit.h"
 #include "drake/common/text_logging.h"
@@ -546,6 +547,11 @@ class Meshcat::Impl {
         main_thread_id_(std::this_thread::get_id()),
         params_(params) {
     DRAKE_THROW_UNLESS(params.port.value_or(7000) >= 1024);
+    if (!drake::internal::IsNetworkingAllowed("meshcat")) {
+      throw std::runtime_error(
+          "Meshcat has been disabled via the DRAKE_ALLOW_NETWORK environment "
+          "variable");
+    }
 
     // Sanity-check the pattern, by passing it (along with dummy host and port
     // values) through to fmt to allow any fmt-specific exception to percolate.
--- a/geometry/meshcat.h
+++ b/geometry/meshcat.h
@@ -126,7 +126,13 @@ in the visualizer.
 - All user objects can easily be cleared by a single, parameter-free call to
 Delete(). You are welcome to use absolute paths to organize your data, but the
 burden on tracking and cleaning them up lie on you.
-*/
+
+@section network_access Network access
+
+See MeshcatParams for options to control the hostname and port to bind to.
+
+See \ref allow_network "DRAKE_ALLOW_NETWORK" for an environment variable
+option to deny %Meshcat entirely. */
 class Meshcat {
  public:
   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(Meshcat)
--- a/geometry/render_gltf_client/BUILD.bazel
+++ b/geometry/render_gltf_client/BUILD.bazel
@@ -31,10 +31,11 @@ drake_cc_library(
     visibility = ["//visibility:public"],
     interface_deps = [
         ":render_engine_gltf_client_params",
-        "//geometry/render:render_camera",
+        "//geometry/render:render_engine",
     ],
     deps = [
         ":internal_render_engine_gltf_client",
+        "//common:network_policy",
     ],
 )
 
@@ -125,11 +126,21 @@ drake_cc_library(
 
 drake_cc_googletest(
     name = "factory_test",
+    allow_network = ["render_gltf_client"],
     deps = [
         ":factory",
         ":internal_render_engine_gltf_client",
         ":render_engine_gltf_client_params",
         "//geometry/render:render_camera",
+    ],
+)
+
+drake_cc_googletest(
+    name = "factory_denied_test",
+    allow_network = ["none"],
+    deps = [
+        ":factory",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
@@ -282,6 +293,7 @@ drake_py_unittest(
 drake_py_unittest(
     name = "integration_test",
     size = "medium",
+    allow_network = ["render_gltf_client"],
     data = [
         ":client_demo",
         ":server_demo",
--- a/geometry/render_gltf_client/factory.cc
+++ b/geometry/render_gltf_client/factory.cc
@@ -1,5 +1,6 @@
 #include "drake/geometry/render_gltf_client/factory.h"
 
+#include "drake/common/network_policy.h"
 #include "drake/geometry/render_gltf_client/internal_render_engine_gltf_client.h"
 
 namespace drake {
@@ -7,6 +8,12 @@ namespace geometry {
 
 std::unique_ptr<render::RenderEngine> MakeRenderEngineGltfClient(
     const RenderEngineGltfClientParams& params) {
+  if (!drake::internal::IsNetworkingAllowed("render_gltf_client")) {
+    throw std::runtime_error(
+        "RenderEngineGltfClient has been disabled via the DRAKE_ALLOW_NETWORK "
+        "environment variable");
+  }
+
   return std::make_unique<render_gltf_client::internal::RenderEngineGltfClient>(
       params);
 }
--- a/geometry/render_gltf_client/factory.h
+++ b/geometry/render_gltf_client/factory.h
@@ -26,7 +26,9 @@ namespace geometry {
    [`curl_global_init(CURL_GLOBAL_ALL | CURL_GLOBAL_ACK_EINTR)`][libcurl_init],
    the implication for consuming applications being:
  @note
-   1. If you intend to have your rendering take place in a threaded context,
+   1. See \ref allow_network "DRAKE_ALLOW_NETWORK" for an environment variable
+      option to deny remote rendering entirely.
+   2. If you intend to have your rendering take place in a threaded context,
       you **must** instantiate this RenderEngine via %MakeRenderEngineGltfClient
       from the main thread **before** spawning your threaded workers.  As soon
       as one of these RenderEngine instances has been constructed, libcurl will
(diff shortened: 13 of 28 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/factory.cc` line 11 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NR8e6DL6ZafkGYdP3oG:-NR8e6DL6ZafkGYdP3oH:b-4q28l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9e91fdde2f2e5881402a3b3ae086faaac215aa4/geometry/render_gltf_client/factory.cc#L11)):*
> ```cpp
> std::unique_ptr<render::RenderEngine> MakeRenderEngineGltfClient(
>     const RenderEngineGltfClientParams& params) {
>   if (!IsNetworkingAllowed("render_gltf_client")) {
> ```

Here, even _constructing_ a render client will throw even if nobody ever asks for an image output.

We could imagine deferring the failure to evaluation-time instead of construction-time.

Should we allow people to add this render engine to their diagram, but then if no cameras use it, that's okay?

ggould-tri (commented):
LGTM as to LCM.

> Reviewed 4 of 19 files at r1, 7 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-:-NR9PEBp11QlC1bjTmVA:b-i0lu22)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`common/network_policy.h` line 34 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NR9MqHX1HJq63nd7KAP:-NR9MqHX1HJq63nd7KAQ:b74e5n6) ([raw file](https://github.com/RobotLocomotion/drake/blob/0e3df99eb66bf9b8a222f974fe24c923d0e50b39/common/network_policy.h#L34)):*
> ```objc
> - If `DRAKE_ALLOW_NETWORK` is set to "none" or a list that contains "none" then
>   all networking is denied. This holds true even if other components are listed,
>   e.g., "none:lcm" still denies lcm.
> ```

These semantics are confusing; unless this is working around some specific case of programatically constructed lists, getting an error from `none:lcm` is more likely to help the user than trying to makes sense of that nonsensical setting.



_Code quote:_
```objc
  all networking is denied. This holds true even if other components are listed,
  e.g., "none:lcm" still denies lcm.
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @ggould-tri)

___
*[`common/network_policy.h` line 34 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NR9MqHX1HJq63nd7KAP:-NRALPh42Di1KuJyG1Lh:br58jpb) ([raw file](https://github.com/RobotLocomotion/drake/blob/0e3df99eb66bf9b8a222f974fe24c923d0e50b39/common/network_policy.h#L34)):*

These semantics are confusing; unless this is working around some specific case of programatically constructed lists, getting an error from `none:lcm` is more likely to help the user than trying to makes sense of that nonsensical setting.
</blockquote></details>

Good point.  Removed from the public documentation.

For a diagnostic, I used warning instead an exception.  Since we fail-closed already, hopefully the user will dig through the logs to figure out why their networking was denied.

I am hesitant to be too demanding about environment variables (i.e., to throw exceptions when parsing them), since they are not always subject to good administrative controls and can have far-reaching effects outside of the subsystem they were aimed at.

jwnimmer-tri:
I hate to pull you away from your other stuff but I need to move this forward, so +@SeanCurtis-TRI for feature review please.

SeanCurtis-TRI (commented):
Thoughts on stamping out hidden effects.

> Reviewed 7 of 19 files at r1, 11 of 16 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-:-NRDzyOT2fqf7iW0j9ZF:bpvrf2n)* status: 8 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDtb8PCl32c_yq6FbP:-NRDtb8PCl32c_yq6FbQ:b-240wq3) (no related file):*
On the subject of throwing:

I've removed my comments below to distill them here. It boils down to what happens to existing code when the network is "disabled". The user should be able to anticipate what will happen to their program if they independently tweak environment variables.

Disabling the network affects three systems and one "operation" (initialization):

  - `Meshcat` and `RenderEngineGltfClient` simply throw upon construction.
  - `DrakeLcm` conditionally throws based on the URL
  - `PackageMap` throws if we ask for the path to a remote package which has no recourse to file URLs.
     - It's worth noting that remote packages URLs are inherently prioritized. It could be that a user has been using a network resource all for a long time, forgotten they have a file backup and they allow that to go out of sync. Now, simply by disabling network they could end up with an out-of-date model with the only hint is a trace message in the log.
       - Waiting for actually calling `GetPath()` is analogous to waiting for asking for an image from a renderer.      
       - In passing, I'd make the argument that instead of silently skipping, this should be treated like `DrakeLcm`; the user has specified something that Drake simply isn't configured even to *try*. It should simply throw.

If the goal is to facilitate users in predicting behavior, I'd suggest:

1. We need the simplest rules possible.
    - As we already have two classes that simply throw upon construction, any Diagram that would contain such a thing must immediately throw. It would be consistent with making everything else strict to a comparable degree.
    - Alternatively, we can go with the policy of "defer throwing as late as possible" (as implemented in `PackageMap` and possibly `RenderEngineGltfClient`). However, that might be a slippery slope; first, it invites aggressive refactoring because "as late as possible' might invite efforts to look at current implementations and argue, "Well, it *could* be later...", and, second, it can lead to arbitrary distances between the initial seed for the throw and the user's act that finally collects is (e.g., `RenderEngineGltfClient` throws because you evaluated a `RenderCamera` output port.
2. We need to document functions/classes that have network dependency and what their relationship to the network being disabled is.

___
___
*[`common/network_policy.h` line 8 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRBRPWwALtUFTikq21V:-NRBRPWwALtUFTikq21W:bl5duwz) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L8)):*
> ```objc
> namespace internal {
> 
> /** @addtogroup environment_variables
> ```

TIL I was surprised that this ended up in doxygen. I guess I didn't understand the relationship between `internal` namespaces and how our doxygen works. So, the internal *function* is omitted, but this module documentation is included.
___
*[`common/network_policy.h` line 37 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDneGF7auwkWLW1MoX:-NRDneGF7auwkWLW1MoY:bti6bc6) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L37)):*
> ```objc
> @} */
> 
> /* Returns true iff the given component is allowed to access the network,
> ```

nit Erroneous comma (which is ironic for me to call out because I'm the king of extensive commas).



_Code quote:_
```objc
,
```
___
*[`common/network_policy.h` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDnqI00lmJH5YbQ_kh:-NRDnqI1DrZIlc_FB7JT:b7eas1p) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L41)):*
> ```objc
> @param component is the name to query, which must be alphanumeric lowercase
> (underscores allowed) and must not be "none" nor the empty string. */
> bool IsNetworkingAllowed(std::string_view component);
> ```

BTW In principle, this API is an invitation for typo mistakes.

I presume the primary arguments here are:

1) It's internal so users won't trip over it (and if they do, they asked for it).
2) Invocations of this method are rare -- reviewers should catch if a call site is providing a bad value by cross referencing with the documentation in this file.
3) If the *environment* has a typo, we expect any user that would attempt to set it would relatively easily be able to compare their environment variable with the documented glyph *and* recognize the need to do so based on this returning false (typically triggered by the exception message).
___
*[`geometry/render_gltf_client/factory.cc` line 11 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NR8e6DL6ZafkGYdP3oG:-NRDq5Ip3YzfCiRgTqKY:bwgb47r) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9e91fdde2f2e5881402a3b3ae086faaac215aa4/geometry/render_gltf_client/factory.cc#L11)):*

Here, even _constructing_ a render client will throw even if nobody ever asks for an image output.

We could imagine deferring the failure to evaluation-time instead of construction-time.

Should we allow people to add this render engine to their diagram, but then if no cameras use it, that's okay?
</blockquote></details>

Rather than tackling this class alone, I've started a conversation above attempting to cover the question more broadly.
___
*[`multibody/parsing/package_map.cc` line 278 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDsNxNCmdZ4CA5uU8z:-NRDsNxNCmdZ4CA5uU9-:bci4srx) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L278)):*
> ```cpp
>     auto should_deny = [](const auto& url) {
>       const bool is_ok = url.substr(0, 7) == "file://";
>       if (!is_ok) {
> ```

BTW I'm not a huge fan of negative logic. Had you considered this alternate spelling? I find it reads easier.



_Suggestion:_
```cpp
    auto should_deny = [](const auto& url) {
      const bool needs_network = url.substr(0, 7) != "file://";
      if (needs_network) {
        log()->trace("PackageMap ignoring '{}' due to DRAKE_ALLOW_NETWORK",
                     url);
      }
      return needs_network;
    };
```
___
*[`multibody/parsing/package_map.cc` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDrgQsBsMM-Sv_gwUd:-NRDrgQsBsMM-Sv_gwUe:b3jltqq) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L288)):*
> ```cpp
>                            safe_params.urls.end());
>   }
>   if (safe_params.urls.empty()) {
> ```

BTW Why are we attempting to create a cache when we are going to throw anyways? I'd argue the logic here is inverted.

We have an implicit expectation entering into this that for a remote package, `urls` is not empty. The previous version of the code relied on that. However, you are now using an environmental condition to blur that precondition and possibly invalidate it. I.e., if this logic were applied earlier, we'd have never gotten to this point; it would've been considered an invalid set of `RemoteParams` because there were no valid URLs.
___
*[`tools/skylark/README.md` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDtKfV4xfOzaScjtVA:-NRDtKfV4xfOzaScjtVB:b-x4cxwn) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/tools/skylark/README.md#L24)):*
> ```Markdown
> A list of components allowed to use the network, per IsNetworkingAllowed()
> in drake/common/network_policy.h. When empty or not set, defaults to allowing
> ["meshcat"] but nothing else. You can set it to ["none"] to deny everything.
> ```

nit: I believe this would benefit for an explanation as to *why* only `meshcat` is allowed by default. Capturing the justification would allow the decision to be revisited by developers in the future.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDtb8PCl32c_yq6FbP:-NRE1jf557yUOwamAXHz:b-81lkxj) (no related file):*
> ... they have a file backup and they allow that to go out of sync. Now, simply by disabling network they could end up with an out-of-date model with the only hint is a trace message in the log.

That's not what would happen.  Model archives have a required checksum.  If the file URL points to an old copy, the package map will throw because the checksum doesn't match.

Re: semantics -- I'll give a spin to having PackageMap throw on Add instead of Get.  I _think_ it'll probably be better.

___
___
*[`common/network_policy.h` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDnqI00lmJH5YbQ_kh:-NRE0OPBBlrSRO7_QsyH:b-e87o4p) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L41)):*

BTW In principle, this API is an invitation for typo mistakes.

I presume the primary arguments here are:

1) It's internal so users won't trip over it (and if they do, they asked for it).
2) Invocations of this method are rare -- reviewers should catch if a call site is providing a bad value by cross referencing with the documentation in this file.
3) If the *environment* has a typo, we expect any user that would attempt to set it would relatively easily be able to compare their environment variable with the documented glyph *and* recognize the need to do so based on this returning false (typically triggered by the exception message).

</blockquote></details>

I don't necessarily disagree with any of that, but I'd posit a different mitigation:

Everything here is unit tested (e.g. "drake_lcm_disabled_test).  It's up to reviewers to cross-check that the unit test is correct.
___
*[`multibody/parsing/package_map.cc` line 278 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDsNxNCmdZ4CA5uU8z:-NRE0juvDpUcq2OLV-DZ:bo67h5s) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L278)):*

BTW I'm not a huge fan of negative logic. Had you considered this alternate spelling? I find it reads easier.
</blockquote></details>

To me, the goal is to fail-closed. As such an _allow list_ is the clearest phrasing.  We say what's allowed, and everything else is denied.

I'll rename to `is_allowed` to make that a tiny bit more clear.
___
*[`multibody/parsing/package_map.cc` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDrgQsBsMM-Sv_gwUd:-NRE13qnCOfVuyjbDrXk:blba6y3) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L288)):*

BTW Why are we attempting to create a cache when we are going to throw anyways? I'd argue the logic here is inverted.

We have an implicit expectation entering into this that for a remote package, `urls` is not empty. The previous version of the code relied on that. However, you are now using an environmental condition to blur that precondition and possibly invalidate it. I.e., if this logic were applied earlier, we'd have never gotten to this point; it would've been considered an invalid set of `RemoteParams` because there were no valid URLs.
</blockquote></details>

Ah.

To be clear -- you're arguing for moving up the exception from `package_map.GetPath()` to `package_map.AddRemote()`.  Is that right?

I agree.  I'll give it a spin.

The implication will be that later (once `package://drake_models` uses `AddRemote`) that the default constructor will only add `package://drake` and not the models (since they are remote-only).

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDtb8PCl32c_yq6FbP:-NRE5_GdD6_aX0svvTiH:bsr66rs) (no related file):*

> ... they have a file backup and they allow that to go out of sync. Now, simply by disabling network they could end up with an out-of-date model with the only hint is a trace message in the log.

That's not what would happen.  Model archives have a required checksum.  If the file URL points to an old copy, the package map will throw because the checksum doesn't match.

Re: semantics -- I'll give a spin to having PackageMap throw on Add instead of Get.  I _think_ it'll probably be better.
</blockquote></details>

Excellent point, re: sha. So, all that would *really* happen if they allowed one of their URLs to go out of date is that failure of the prioritized URL (for whatever reason), would merely lead to them being made aware that it didn't match sha.

___
___
*[`common/network_policy.h` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDnqI00lmJH5YbQ_kh:-NRE6iB-C8sn0QkdabpR:b-mq7yr0) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L41)):*

I don't necessarily disagree with any of that, but I'd posit a different mitigation:

Everything here is unit tested (e.g. "drake_lcm_disabled_test).  It's up to reviewers to cross-check that the unit test is correct.
</blockquote></details>

To be perfectly frank, I'd have less concern for (1) and (2), and the most for (3). We have no way to tell them that they've specified nonsense in their environment variable.

(Although a static never_destroyed could could evaluate and log defects in the environment variable including "this glyph doesn't mean anything to drake".)
___
*[`multibody/parsing/package_map.cc` line 278 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDsNxNCmdZ4CA5uU8z:-NRE7KN5CCz91OfPFTpR:b6mec1k) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L278)):*

To me, the goal is to fail-closed. As such an _allow list_ is the clearest phrasing.  We say what's allowed, and everything else is denied.

I'll rename to `is_allowed` to make that a tiny bit more clear.
</blockquote></details>

Don't bother changing to the synonym. I'm content to accept your justification.  For you, the reversal obfuscates the allow list. For me, the allow list obfuscates the subsequent logic. There is no "right" or "wrong" so I happily defer to author's taste having proposed an alternative consistent with reviewer's taste.
___
*[`multibody/parsing/package_map.cc` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDrgQsBsMM-Sv_gwUd:-NRE8-Bf48CLR0ABPwy7:b-je1qft) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L288)):*

Ah.

To be clear -- you're arguing for moving up the exception from `package_map.GetPath()` to `package_map.AddRemote()`.  Is that right?

I agree.  I'll give it a spin.

The implication will be that later (once `package://drake_models` uses `AddRemote`) that the default constructor will only add `package://drake` and not the models (since they are remote-only).
</blockquote></details>

You got my point exactly right.

As to the implications, I don't think that those implications are due to the re-ordering here. I think that is a consequence of changing `PackageMap`'s throw from `Get` to `Add`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @ggould-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDtb8PCl32c_yq6FbP:-NRENb0IEAM7rmJpaElm:b-lnidzk) (no related file):*

Excellent point, re: sha. So, all that would *really* happen if they allowed one of their URLs to go out of date is that failure of the prioritized URL (for whatever reason), would merely lead to them being made aware that it didn't match sha.
</blockquote></details>

Done (documentation added).

___
___
*[`common/network_policy.h` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDnqI00lmJH5YbQ_kh:-NRE9Mex3Dnjb2_08y1F:b-o5x5wc) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L41)):*

To be perfectly frank, I'd have less concern for (1) and (2), and the most for (3). We have no way to tell them that they've specified nonsense in their environment variable.

(Although a static never_destroyed could could evaluate and log defects in the environment variable including "this glyph doesn't mean anything to drake".)
</blockquote></details>

Ya, the way they find typos in their environment is that the subsystem throws ("not allowed") and then they need to dig in.

I don't want to add a warning because _the set of nouns will change over time_.  So if the user sets the allowlist environment variable for Drake v1.18.0 in their dev folder and then runs the Drake v1.17.0 pin from prod that doesn't know about it yet, it will yell at them for no reason.  As with the other similar discussions -- environment variables are very very loosely coupled, so we should be extremely generous with their interpretation (IMO).
___
*[`multibody/parsing/package_map.cc` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDrgQsBsMM-Sv_gwUd:-NRE9za5DVcOQg5eUlGn:b-wsm83s) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L288)):*

You got my point exactly right.

As to the implications, I don't think that those implications are due to the re-ordering here. I think that is a consequence of changing `PackageMap`'s throw from `Get` to `Add`.
</blockquote></details>

Done.

It's a bit tricky too -- it's only an exception at Add time when there are no file:// URLS *and* the cache doesn't already have it.

jwnimmer-tri:
+@EricCousineau-TRI for platform review per schedule, please.

EricCousineau-TRI (commented):
:lgtm: platform, nice!

> Reviewed 7 of 19 files at r1, 8 of 16 files at r2, 4 of 5 files at r3, 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-:-NRETAPcDqr8bBDJsGWM:bb368ni)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`common/network_policy.h` line 8 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRES2o41Bm4u2esghVK:-NRES2o41Bm4u2esghVL:b-oajpbs) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/common/network_policy.h#L8)):*
> ```objc
> namespace internal {
> 
> /** @addtogroup environment_variables
> ```

nit To perhaps clarify intent, can this user-facing documentation block be hoisted outside of `namespace internal`?

I assume there would be no functional change, but may make it easier for future readers.
___
*[`common/network_policy.h` line 43 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NREQYAzAJfYJxltYhgQ:-NREQYB-9Z5GjgYzch46:b67t258) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/common/network_policy.h#L43)):*
> ```objc
> @} */
> 
> /* Returns true iff the given component is allowed to access the network based
> ```

BTW From my reading of this, nothing is cached and there is no other internal representation users can toggle (e.g. a global parameter struct).
This means users can change the environment variable during execution and see effects - which IMO sounds good and reasonable.

Is it worth documenting this? (I see that it's effectively tested in the unittest)
And is there any value to configuring via other means?
Naively, I would assume no, aside from perhaps multithreading (I have no idea how env + threading works), but just want to check.
___
*[`common/network_policy.cc` line 45 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRERggX6DwrjTT5W8xa:-NRERggY95vCLxsXW7nQ:b-cccnem) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/common/network_policy.cc#L45)):*
> ```cpp
>   bool match = false;
>   std::string_view worklist = env_view;
>   while (!worklist.empty()) {
> ```

nit Is it worth documenting y'all's discussion about not checking for valid identifiers? e.g.

```
// N.B. We purposefully do not check for identifier validity because they may evolve over time,
// and it may be hard for users to have full curated control over their environment.
```
___
*[`tools/skylark/README.md` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRESU0N0ZyrkGtP3kOL:-NRESU0N0ZyrkGtP3kOM:b-mmwwyh) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/tools/skylark/README.md#L13)):*
> ```Markdown
> 
> 
> # Common options
> ```

BTW These all seem intended for tests. Consider qualifying section? (unless I'm wrong and it's used in benchmarking)
___
*[`tools/skylark/README.md` line 26 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRESbDJ2DRJ0tWjX51T:-NRESbDJ2DRJ0tWjX51U:b-5wc2st) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/tools/skylark/README.md#L26)):*
> ```Markdown
> ["meshcat"] but nothing else. You can set it to ["none"] to deny everything.
> 
> We allow meshcat by default because it's mostly harmless, a bajillion tests use
> ```

BTW Thanks for clarifying!

SeanCurtis-TRI (commented):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`common/network_policy.h` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDnqI00lmJH5YbQ_kh:-NREPVkg8gXunkair8rO:b-s3i77g) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/common/network_policy.h#L41)):*

Ya, the way they find typos in their environment is that the subsystem throws ("not allowed") and then they need to dig in.

I don't want to add a warning because _the set of nouns will change over time_.  So if the user sets the allowlist environment variable for Drake v1.18.0 in their dev folder and then runs the Drake v1.17.0 pin from prod that doesn't know about it yet, it will yell at them for no reason.  As with the other similar discussions -- environment variables are very very loosely coupled, so we should be extremely generous with their interpretation (IMO).
</blockquote></details>

Yeah. I figured that was basically the argument. I'm just happy to hear you say it.
___
*[`geometry/meshcat.h` line 135 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NREQYMM6VkiATL2j9TS:-NREQYMM6VkiATL2j9TT:bom8ubj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/geometry/meshcat.h#L135)):*
> ```objc
> 
> See \ref allow_network "DRAKE_ALLOW_NETWORK" for an environment variable
> option to disable %Meshcat entirely. */
> ```

nit: It doesn't *disable* `Meshcat`, it turns it into a bomb. To me "disable" has more of a no-op connotation. In my most generous interpretation, this comment still undersells the actual implications.

Same applies to the `RenderEngineGltfClient`.

FTR I'd *considered* giving these classes a "disabled" state but felt that it was too much for this initial PR. I can imagine if this causes pain in the future, opting to truly disable them (while keeping them runtime inoffensive) can always be done.
___
*[`multibody/parsing/package_map.h` line 133 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRERR3N8oBPsTslFNXc:-NRERR3N8oBPsTslFNXd:b-3ssqah) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/multibody/parsing/package_map.h#L133)):*
> ```objc
>   See \ref allow_network "DRAKE_ALLOW_NETWORK" for an environment variable
>   option to disable network fetching. %AddRemote may still be used even with
>   network fetching disabled -- in that case, the urls must contain a `file:`
> ```

nit: You typically refer to these url prefixes with a trailing `//` (i.e., `file://`).
___
*[`multibody/parsing/package_map.cc` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRDrgQsBsMM-Sv_gwUd:-NREPbOfE6hagp1aYuXX:b-ygvl4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/d58ce623fc5500bee505b4755556df5f6d65c308/multibody/parsing/package_map.cc#L288)):*

Done.

It's a bit tricky too -- it's only an exception at Add time when there are no file:// URLS *and* the cache doesn't already have it.
</blockquote></details>

Hmmm....there's a subtlety that I didn't share earlier, I'm going to share it now.

Before the change, the only reason drake would fire through alternate URLs was become something *external* to Drake prevented the highest priority URL to be used. If network is disabled, it is Drake skipping URLs of its own accord. Those feel quite different. In the former, Drake is foiled in its effort to act on the configuration as given by the user. In the latter, Drake has skipped it with a muted voice.

That said, if I draw an analogy where the user provides an SDF which references unsupported tags, it is sufficient that we log a comment saying, "Hey! I'm skipping things." In that analogy, the fact that we're filtering and announcing it in the log is clearly sufficient.

And, as you point out, we also have the nicety of allowing previously cached remote resources. So, for all of that, I'm marking myself satisified.
___
*[`multibody/parsing/package_map.cc` line 96 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRERf2331C4gS0kr1w5:-NRERf2331C4gS0kr1w6:b-bx3irn) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/multibody/parsing/package_map.cc#L96)):*
> ```cpp
>   @pre needs_fetch() is true.
>   @pre either the lock() is currently held or the caller has exclusive
>   ownership of `this` */
> ```

nit "Exclusive ownership" feels like the wrong term of art. We can have exclusive ownership (e.g., `unique_ptr`) but spray raw pointers everywhere. The sticking word for me is "ownership". My gut tells me you're going for something stricter than ownership, something more akin to access.
___
*[`multibody/parsing/package_map.cc` line 232 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRESe3x1IfG30_37QUg:-NRES-5N5z5f4whPnWXA:bt527e2) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/multibody/parsing/package_map.cc#L232)):*
btw Pass by value?

Quoting cppreference vis a vis when named return value optimization can be applied:

> In a return statement, when the operand is the name of a non-volatile object with automatic storage duration, which __*isn't a function parameter*__ or a...

Therefore, I can't infer that it was used to enable return value optimization.

Alternatively, you're doing pass-by-value not to facilitate return value optimization, but to optimize the invocation with the expectation that it is generally called with r-values? That doesn't generally seem true either.

What am I missing?

SeanCurtis-TRI (commented):
:LGTM: With a couple of small things.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-:-NREVLTEBtVE2_rbS4MA:bti4v7h)* status: 6 unresolved discussions (waiting on @jwnimmer-tri)

EricCousineau-TRI (commented):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform) (waiting on @jwnimmer-tri)

SeanCurtis-TRI:
*[`multibody/parsing/package_map.h` line 133 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19050#-NRERR3N8oBPsTslFNXc:-NREfjB-4khUTh3rwMfO:by3giay) ([raw file](https://github.com/RobotLocomotion/drake/blob/9119735ab20e86da1b4133d822fd5b8b7f2c174d/multibody/parsing/package_map.h#L133)):*

Possibly because it's the end of the line? It works just fine in the [PackageMap documentation](https://drake.mit.edu/doxygen_cxx/structdrake_1_1multibody_1_1_package_map_1_1_remote_params.html#a306ada97a24c8198c1a0c9c3c5b0bb25).

I've generally seen that lines that end with slashes make doxygen persnickety. As this is public API, it'd be nice to have it consistent.
</blockquote></details>

Or, as in that case, surrounding it with double quotes instead.  :)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19050)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform) (waiting on @jwnimmer-tri)

Issue #10320, opened 2018-12-27
Some constraint is not thread safe


hongkai-dai:
`ExpressionConstraint` is not thread safe, as it contains *mutable* `symbolic::Environment`
https://github.com/RobotLocomotion/drake/blob/f4a0b01b5f16308676529aa540e8cffcf88c0ca5/solvers/constraint.h#L902-L903

We can have two solutions

1. Add `bool is_thread_safe` to `EvaluatorBase` class. This flag is default to false, and every constraint/cost has to set it to true if the user wants to call them in multiple threads.
2. Fix `ExpressionConstraint`, we can pass a `symbolic::Environment` pointer to `ExpressionConstraint` like
```C++
class ExpressionConstraint {
 public:
  ExpressionConstraint(const symbolic::Expression& e, symbolic::Environment* const env);

  template <typename T>
  void DoEvalGeneric(const Eigen::Ref<const VectorX<T>>& x, VectorX<T>* y) const {
    UpdateEnvironment(x, env_);
    // Then evaluate the expression with this env.
    *y =  env_.Evaluate(e_);
  }

 private:
  symbolic::Expression e_;
  symbolic::Environment* const env_;
}
```
This is similar to how we pass `Context` pointer to InverseKinematic constraint, and evaluate these constraints by first updating the context. This would make `ExpressionConstraint` to be thread safe.

@RussTedrake @avalenzu @jwnimmer-tri do you have some suggestions? Which approach would you prefer?

RussTedrake:
I prefer -- both!  There are other constraints out there (e.g. the direct collocation constraint) that are also not thread safe.  I think we should systematically clean them up and set the flag to "safe" every time we bless an implementation.  But new constraints should have to opt-in to being declared thread safe.  

The only alternative I see is to fix every constraint we know about, and then document the assumption that EVERY constraint must be implemented as thread safe.

jwnimmer-tri:
I would do this:
(1) Run a benchmark, to get a baseline performance measurement.
(2) Change `environment_` to be a local variable `env` in DoEval, not a member field.
(3) Re-run the benchmark.  If it's fast enough, you're done.
(4) Profile what's slow in DoEval.  Is it rehashing, or is is malloc?
(5) If its rehashing, trying adding method `Environment::reserve`, since we know up-front how many variables we're going to need.
[...]

PR #19052, opened 2023-03-22
Add Methods to solve mathematical programs in parallel

AlexandreAmice:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)

### DIFF
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -2067,10 +2067,13 @@ void BindFreeFunctions(py::module m) {
           py::arg("solver_options") = py::none(), doc.Solve.doc_3args)
       .def("SolveInParallel",
           py::overload_cast<const std::vector<MathematicalProgram>&,
-              const int,
-              const bool>(&solvers::SolveInParallel),
-          py::arg("prog_list"), py::arg("num_threads") = -1,
-          py::arg("verbose") = true)
+              const std::optional<std::vector<std::optional<Eigen::VectorXd>>>&,
+              const std::optional<SolverOptions>&, const int, const bool>(
+              &solvers::SolveInParallel),
+          py::arg("prog_list"), py::arg("initial_guess") = py::none(),
+          py::arg("solver_options") = py::none(), py::arg("num_threads") = -1,
+          py::arg("terminate_at_first_infeasible") = false,
+          doc.SolveInParallel.doc)
       .def("GetProgramType", &solvers::GetProgramType, doc.GetProgramType.doc);
 }
 
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -2065,10 +2065,12 @@ void BindFreeFunctions(py::module m) {
               const std::optional<SolverOptions>&>(&solvers::Solve),
           py::arg("prog"), py::arg("initial_guess") = py::none(),
           py::arg("solver_options") = py::none(), doc.Solve.doc_3args)
-//      .def("SolveInParallel",
-//          &solvers::SolveInParallel,
-//          py::arg("prog_list"), py::arg("num_threads") = -1,
-//          py::arg("verbose") = true)
+      .def("SolveInParallel",
+          py::overload_cast<const std::vector<MathematicalProgram>&,
+              const int,
+              const bool>(&solvers::SolveInParallel),
+          py::arg("prog_list"), py::arg("num_threads") = -1,
+          py::arg("verbose") = true)
       .def("GetProgramType", &solvers::GetProgramType, doc.GetProgramType.doc);
 }
 
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -2065,15 +2065,6 @@ void BindFreeFunctions(py::module m) {
               const std::optional<SolverOptions>&>(&solvers::Solve),
           py::arg("prog"), py::arg("initial_guess") = py::none(),
           py::arg("solver_options") = py::none(), doc.Solve.doc_3args)
-      .def("SolveInParallel",
-          py::overload_cast<const std::vector<MathematicalProgram>&,
-              const std::optional<std::vector<std::optional<Eigen::VectorXd>>>&,
-              const std::optional<SolverOptions>&, const int, const bool>(
-              &solvers::SolveInParallel),
-          py::arg("prog_list"), py::arg("initial_guess") = py::none(),
-          py::arg("solver_options") = py::none(), py::arg("num_threads") = -1,
-          py::arg("terminate_at_first_infeasible") = false,
-          doc.SolveInParallel.doc)
       .def("GetProgramType", &solvers::GetProgramType, doc.GetProgramType.doc);
 }
 
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -2065,6 +2065,10 @@ void BindFreeFunctions(py::module m) {
               const std::optional<SolverOptions>&>(&solvers::Solve),
           py::arg("prog"), py::arg("initial_guess") = py::none(),
           py::arg("solver_options") = py::none(), doc.Solve.doc_3args)
+//      .def("SolveInParallel",
+//          &solvers::SolveInParallel,
+//          py::arg("prog_list"), py::arg("num_threads") = -1,
+//          py::arg("verbose") = true)
       .def("GetProgramType", &solvers::GetProgramType, doc.GetProgramType.doc);
 }
 
--- a/bindings/pydrake/solvers/test/mathematicalprogram_test.py
+++ b/bindings/pydrake/solvers/test/mathematicalprogram_test.py
@@ -195,6 +195,20 @@ class TestMathematicalProgram(unittest.TestCase):
             prog_type=mp.ProgramType.kLP)
         self.assertGreater(len(linear_solvers), 0)
 
+    def test_solve_in_parallel(self):
+        prog = mp.MathematicalProgram()
+        x = prog.NewContinuousVariables(2, "x")
+
+        # Add linear equality constraints; make sure the solver works.
+        prog.AddLinearConstraint(x[0] + x[1] == 0)
+        prog.AddLinearConstraint(2 * x[0] - x[1] == 1)
+        solver_id = mp.ChooseBestSolver(prog)
+        self.assertEqual(solver_id.name(), "Linear system")
+        solver = mp.MakeSolver(solver_id)
+
+        prog_list = [prog.Clone() for _ in range(25)]
+        # result_list = mp.SolveInParallel(prog_list)
+
     def test_module_level_solve_function_and_result_accessors(self):
         qp = TestQP()
         x_expected = np.array([1, 1])
--- a/bindings/pydrake/solvers/test/mathematicalprogram_test.py
+++ b/bindings/pydrake/solvers/test/mathematicalprogram_test.py
@@ -195,20 +195,6 @@ class TestMathematicalProgram(unittest.TestCase):
             prog_type=mp.ProgramType.kLP)
         self.assertGreater(len(linear_solvers), 0)
 
-    def test_solve_in_parallel(self):
-        prog = mp.MathematicalProgram()
-        x = prog.NewContinuousVariables(2, "x")
-
-        # Add linear equality constraints; make sure the solver works.
-        prog.AddLinearConstraint(x[0] + x[1] == 0)
-        prog.AddLinearConstraint(2 * x[0] - x[1] == 1)
-        solver_id = mp.ChooseBestSolver(prog)
-        self.assertEqual(solver_id.name(), "Linear system")
-        solver = mp.MakeSolver(solver_id)
-
-        prog_list = [prog.Clone() for _ in range(25)]
-        # result_list = mp.SolveInParallel(prog_list)
-
     def test_module_level_solve_function_and_result_accessors(self):
         qp = TestQP()
         x_expected = np.array([1, 1])
--- a/solvers/solve.cc
+++ b/solvers/solve.cc
@@ -7,6 +7,11 @@
 #include "drake/solvers/choose_best_solver.h"
 #include "drake/solvers/solver_interface.h"
 
+#include <thread>
+#include <future>
+
+
+
 namespace drake {
 namespace solvers {
 MathematicalProgramResult Solve(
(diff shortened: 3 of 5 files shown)


AlexandreAmice (commented):
+@hongkai-dai for feature review please! Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-:-NR8s_nl5yUQaNWpVWZR:b-atfp80)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri:
The code you copied this idiom from was written prior to OpenMP support in Drake.  I think it's probable that we should not land this implementation but instead rewrite using OpenMP instead.  Or possibly a veneer over OpenMP so that it also works on macOS.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

___
*[`solvers/solve.h` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NR9Akl-AaMho8RncIgN:-NR9JIOB-QQ1kMOimsqZ:b-8sg80j) ([raw file](https://github.com/RobotLocomotion/drake/blob/859e1d6329538b4a6b7e1723a77d140118868756/solvers/solve.h#L85)):*

We want to leave the optional-eigen so that the user can specify initial guesses for some, but not all the optimization programs.
</blockquote></details>

Right.  My proposal is to lose the outermost optional, but keep the inner one.
___
*[`solvers/solve.h` line 89 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NR9BxTi5iz61kZkKe1y:-NR9JNHiAJ7ZlVj3CHHG:b1o3vb6) ([raw file](https://github.com/RobotLocomotion/drake/blob/859e1d6329538b4a6b7e1723a77d140118868756/solvers/solve.h#L89)):*

Yes we can. I was following the pattern already established for `Solve` in this file. Do you know why the `Solve` method doesn't do this optional argument as well.
</blockquote></details>

I think a long time ago we had more overloads, and the defaulted arguments were confusing the compiler. (Possibly it was even long enough ago that we didn't have std::optional yet either!)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

___
*[`solvers/solve.h` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NR9AGLHCeJbaQ0EyYcn:-NR9JhmZ818jTB-FlxJ7:bba04xr) ([raw file](https://github.com/RobotLocomotion/drake/blob/859e1d6329538b4a6b7e1723a77d140118868756/solvers/solve.h#L83)):*

That is likely why I was having trouble binding these methods in pydrake. I don't think that passing a vector of `MathematicalProgram&` would fix this problem for pydrake correct?
</blockquote></details>

I'm not 100% sure about the right fix.

Passing `vector<const Program*>` might be one way, with some kind of sugar for C++ users who already have `vector<Program>` already created.

Really, it should be a `std::ranges::range` but we haven't quite tried to bite that off yet.

I can try a few spellings locally and get back to you.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 4 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

___
*[`solvers/solve.h` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NR9AGLHCeJbaQ0EyYcn:-NRJ-0aK2jIhhL5ARad-:b-nlftkz) ([raw file](https://github.com/RobotLocomotion/drake/blob/859e1d6329538b4a6b7e1723a77d140118868756/solvers/solve.h#L83)):*

I'm not 100% sure about the right fix.

Passing `vector<const Program*>` might be one way, with some kind of sugar for C++ users who already have `vector<Program>` already created.

Really, it should be a `std::ranges::range` but we haven't quite tried to bite that off yet.

I can try a few spellings locally and get back to you.
</blockquote></details>

Have a look at the top commit here and see what you think:

https://github.com/jwnimmer-tri/drake/commits/parallel_mp_solve

That's my best guess for the API of the core function (that actually performs the solving). I think it will be easier to pybind (and call from Python) than the current API. If it seems like its on the right track and you like it, you could merge it into this branch and we keep going from there.

What it lacks is nice C++ sugar.  I wasn't exactly sure what the typical C++ call site would look like.  Is the common case to provide:
* N programs, with
* either 0 or N initial guesses?
* either 0 or 1 solver options?

I am actually thinking that for C++ instead of using a pile of arguments, we would use a helper struct or class, so that there is less declaration syntax overhead at the call sites.  Stay tuned.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRKgTDb82po09wEWSm7:b-cyjimv) (no related file):*
Also can you clarify the bigger-picture goal here?

If we only need to solve-in-parallel within some Drake algorithm (like a GCS thing), then there is probably a much, much simpler way to do this.

Or if the goal is to provide this feature to users, especially Python users, then we need to do even more that my WIP so far, for the same reason as we did for MonteCarlo:

https://github.com/RobotLocomotion/drake/blob/41f0cbcae7b8ee0a4d499ca6f4ec6a591c11edd8/bindings/pydrake/systems/analysis_py.cc#L349-L351

___

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRKiuAb5lwU6GjpRNCb:b-g9vfw8) (no related file):*

Also can you clarify the bigger-picture goal here?

If we only need to solve-in-parallel within some Drake algorithm (like a GCS thing), then there is probably a much, much simpler way to do this.

Or if the goal is to provide this feature to users, especially Python users, then we need to do even more that my WIP so far, for the same reason as we did for MonteCarlo:

https://github.com/RobotLocomotion/drake/blob/41f0cbcae7b8ee0a4d499ca6f4ec6a591c11edd8/bindings/pydrake/systems/analysis_py.cc#L349-L351


</blockquote></details>

The primary objective is to enable parallel solving of MathematicalPrograms for Python users. Several members of my lab have asked for this and it seems easier to provide this functionality on the C++ side than make MathematicalProgram pickleable so it can be used in Python multiprocessing code.

I hope to look at your push soon.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRKk87p7_ow75yTh2I4:b-j9o3ih) (no related file):*

The primary objective is to enable parallel solving of MathematicalPrograms for Python users. Several members of my lab have asked for this and it seems easier to provide this functionality on the C++ side than make MathematicalProgram pickleable so it can be used in Python multiprocessing code.

I hope to look at your push soon.
</blockquote></details>

Ah, okay.  I will need to give it more thought in that case.  In general, a Python call into this function with `num_threads_actual > 1` risks crashing.

See also https://github.com/RobotLocomotion/drake/issues/10320 where certain programs (certain constraints) cannot be solved in parallel either.

___

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRPrMP_4EtAImHWyD0g:bfi4qj5) (no related file):*

Ah, okay.  I will need to give it more thought in that case.  In general, a Python call into this function with `num_threads_actual > 1` risks crashing.

See also https://github.com/RobotLocomotion/drake/issues/10320 where certain programs (certain constraints) cannot be solved in parallel either.
</blockquote></details>

This is showing my lack of understanding of how parallelism works. Why would a Python call into C++ function that has multiple threads risk crashing? Also, why would the thread safety of the constraints matter in this situation? Here each program is being solved in its own thread and so should not be cross talking to any other programs.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRQyLUQ02XOSO_Ca5M-:bu4dx1h) (no related file):*
> Why would a Python call into C++ function that has multiple threads risk crashing?

Sometimes, costs or constraints are Python functions:
https://drake.mit.edu/pydrake/pydrake.solvers.html#pydrake.solvers.PyFunctionConstraint

When Python code (e.g., a constraint eval) is running, it needs to hold the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) or else things go very badly.  Ensuring that the GIL is locked exactly when needed (but not too much, or else we have a deadlock), is difficult.

> Also, why would the thread safety of the constraints matter in this situation?

The same Constraint object might have been added to multiple MathematicalPrograms.

___

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRi1sOLEUmO34rScIiS:b-mo1w8n) (no related file):*

> Why would a Python call into C++ function that has multiple threads risk crashing?

Sometimes, costs or constraints are Python functions:
https://drake.mit.edu/pydrake/pydrake.solvers.html#pydrake.solvers.PyFunctionConstraint

When Python code (e.g., a constraint eval) is running, it needs to hold the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) or else things go very badly.  Ensuring that the GIL is locked exactly when needed (but not too much, or else we have a deadlock), is difficult.

> Also, why would the thread safety of the constraints matter in this situation?

The same Constraint object might have been added to multiple MathematicalPrograms.
</blockquote></details>

I see. Will this also be a problem when we move forward with binding the method in `cspace_free_polytope` such as this method? https://github.com/RobotLocomotion/drake/blob/0300d225f3fc201101f5bf4f7b2db09b29eaaa57/geometry/optimization/cspace_free_polytope.cc#L658

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRi7WDe2BgseJ57ziZq:bbchtj4) (no related file):*

I see. Will this also be a problem when we move forward with binding the method in `cspace_free_polytope` such as this method? https://github.com/RobotLocomotion/drake/blob/0300d225f3fc201101f5bf4f7b2db09b29eaaa57/geometry/optimization/cspace_free_polytope.cc#L658
</blockquote></details>

Since the users cannot add PyFunctionConstraint to the generated MathematicalProgram (the MathematicalProgram is buried inside this FindSeparationCertificateGivenPolytope function that the user cannot access), I think it should be fine?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NRk4xD84RnEEGHRZbL9:btjvnhu) (no related file):*
> Since the users cannot add PyFunctionConstraint to the generated MathematicalProgram (the MathematicalProgram is buried inside this FindSeparationCertificateGivenPolytope function that the user cannot access), I think it should be fine?

Correct.  However, also note that you are very nearby something that would be a hazard.  The `CspaceFreePolytope` has a member field `const geometry::SceneGraph<double>& scene_graph_` which is used within `FindSeparationCertificateGivenPolytope`.  If the use were inside of the worker thread, it would be invalid (since the SceneGraph could be Python-involved).  However, it's only use is after the worker threads are finished.

> The same Constraint object might have been added to multiple MathematicalPrograms.

See https://github.com/RobotLocomotion/drake/issues/10320 where we've talked about tagging evaluators with whether or not its safe to call a `const` method from multiple threads.  That feature would be a prerequisite of landing this PR.

___

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19052)* status: 5 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),hongkai-dai, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19052#-NRKgTDb82po09wEWSm6:-NS1YcPCCyrQwN_K9Qb-:b-i6lzac) (no related file):*

> Since the users cannot add PyFunctionConstraint to the generated MathematicalProgram (the MathematicalProgram is buried inside this FindSeparationCertificateGivenPolytope function that the user cannot access), I think it should be fine?

Correct.  However, also note that you are very nearby something that would be a hazard.  The `CspaceFreePolytope` has a member field `const geometry::SceneGraph<double>& scene_graph_` which is used within `FindSeparationCertificateGivenPolytope`.  If the use were inside of the worker thread, it would be invalid (since the SceneGraph could be Python-involved).  However, it's only use is after the worker threads are finished.

> The same Constraint object might have been added to multiple MathematicalPrograms.

See https://github.com/RobotLocomotion/drake/issues/10320 where we've talked about tagging evaluators with whether or not its safe to call a `const` method from multiple threads.  That feature would be a prerequisite of landing this PR.
</blockquote></details>

I think I will move this to a feature request on Github and close this PR. I think there is a lot of value to being able to solve Mathematical Programs in parallel, especially for Python users who are limited by the GIL, but there appears to be much more work to do before we actually are able to support this.

___

Issue #18844, opened 2023-02-21
Texturing for Drake supported RenderEngine and visualizers


zachfang:
The goal of this epic issue is to **summarize** and **prioritize** existing issues (usability, consistency, missing feature) regarding `primitive` and `mesh` texturing in either Drake's supported renderers, e.g., `RenderEngine{Vtk, Gl, GltfClient}`, or visualizers, e.g., `MeshcatVisualizer` and `Meldis`.

The ultimate goal is to **(1) support common texture-loading scenarios** and to **(2) have consistent texturing behavior** across renderers and visualizers in Drake. Along the way to achieve the goal, we should improve the documentation and code (warning or throw) to inform users of the current status, i.e., what is supported and what is not.

Backlog:
- [ ] RenderEngine(s) should support loading mtl file (#11949).

zachfang:
To break down the problem, these are the existing or common ways to specify a texture (IMO).
1. A single phong-like diffuse color.
2. `OBJ-to-PNG` association in the same folder.
3. Specify a texture through `<drake:diffuse_map>` tag via parsers.
4. Specify texture(s) through an MTL file.

- Drake's visualizers (only tested with Meldis for now) support (1) and (4).
  - For (3), `diffuse_map` is only used in `perception` roles only.
  - Note that primitives are not textured in the visualizers.

- Drake's renderers (only tested with RenderEngineVtk for now) support (1), (2), and (3)*.
  - `<drake:diffuse_map>` is [only parsed via SDFormats but not URDFs](https://github.com/RobotLocomotion/drake/blob/1397c24defca001f169dfddaa7700b8f2deb6abb/multibody/parsing/parsing_doxygen.h#L414-L418).

zachfang:
For demonstration purposes, five meshes (base, shoulder, upperarm, forearm, and wrist1 of an UR3e) are loaded through an SDFormat and a URDF. The visualization results from a visualizer (Meldis) and a renderer (RenderEngineVtk) are compared side-by-side.

Five meshes all use different texture specifications.
- Base: Textured with the referenced MTL file with the same name in the OBJ.
- Shoulder: Textured with the referenced MTL file with a different name in the OBJ.
- Upperarm: No MTL file. Textured with a simple diffuse color, i.e., RED.
- Forearm: No MTL file. Textured via `<drake:diffuse_map>` tag in the SDFormat or URDF file.
- Wrist1: No MTL file. Textured via a PNG file with the same name.

The result from the URDF:
[...]

Issue #16997, opened 2022-04-19
URDF parser: emit diagnostics for typos, malformed documents, etc.


rpoyner-tri:
This is a reduced-scope version of #12610. For URDF, we do not yet have a mechanism that can reasonably complain about completely unknown wrong input data. The current implementation relies on tinyxml2 to build a "DOM" tree, which the "parser" then effectively performs bespoke queries against. Since tinyxml2 does not even guarantee well-formedness, it is possible to have files full of stray garbage (#16516), near-miss formulations of known tags, etc.

The victory condition of this ticket is to produce a parser that can complain about unknown/malformed elements in a controllable, comprehensible, actionable way.

It may be a lot of work.

Possible paths include:
* add a schema checking pass to the URDF parse; requires schema maintenance and possibly tricky diagnostic plumbing.
* rewrite the existing tinyxml2 parser to visit every element/attribute and make some decision (implement-on-plant, diagnostic, ignore).
** requires disciplined coding and/or a switch to some visitor pattern (tinyxml2 SAX API??)
* use some other library/technology that helps with these problems.

rpoyner-tri:
See #16229 for an overview of parser improvements.

jwnimmer-tri:
See also https://github.com/RobotLocomotion/drake/issues/16785#issuecomment-1098445866 for an implementation idea:

For URDF, separate from any other parsing, we could walk the entire tinyxml2 tree from the root on down and check every element name and attribute name against a position-agnostic allow-list (i.e., irrespective of its position within the tree), producing an error for any completely-unknown name.  That would catch typos where the name was wrong no matter it might live in the tree, but it would not detect elements or attributes that were located in the wrong position within the tree.  So, it is not very precise, but on the other hand it's only ~100 lines of code to implement.

Or on another tack, the libsdformat syntax checker and low-level DOM is actually quite good (https://github.com/ignitionrobotics/sdformat/tree/sdf12/sdf/1.9).  With maybe a few days of work, we could encode (Drake's subset of) the URDF specification into that format, and then use libsdformat to do the syntax checking.  Possibly even OpenRobotics would be interested to collaborate on that.

jwnimmer-tri:
> Or on another tack, the libsdformat syntax checker and low-level DOM is actually quite good. With maybe a few days of work, we could encode (Drake's subset of) the URDF specification into that format, and then use libsdformat to do the syntax checking.

[...]

PR #19066, opened 2023-03-27
[parsing] Error out if <drake:diffuse_map> is specified in a URDF

zachfang:
Toward #18844 

This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066)

### DIFF
--- a/multibody/parsing/detail_urdf_geometry.cc
+++ b/multibody/parsing/detail_urdf_geometry.cc
@@ -132,6 +132,18 @@ UrdfMaterial ParseMaterial(const TinyXml2Diagnostic& diagnostic,
     return {};
   }
 
+  const XMLElement* drake_diffuse_map_node =
+      node->FirstChildElement("drake:diffuse_map");
+  if (drake_diffuse_map_node) {
+    // Error condition: #2: an SDFormat-specific Drake extension,
+    // <drake:diffuse_map>, specified in a URDF.
+    diagnostic.Error(*node,
+                     "<drake:diffuse_map> is not supported in URDF. See "
+                     "https://drake.mit.edu/doxygen_cxx/group__multibody__parsing.html#tag_drake_diffuse_map"  // NOLINT
+                     " for more details.");
+    return {};
+  }
+
   // Test for texture information.
   std::optional<std::string> texture_path;
   const XMLElement* texture_node = node->FirstChildElement("texture");
@@ -164,7 +176,7 @@ UrdfMaterial ParseMaterial(const TinyXml2Diagnostic& diagnostic,
 
   if (!rgba && !texture_path) {
     if (!name.empty() && materials->find(name) == materials->end()) {
-      // Error condition: #2: name with no properties has not been previously
+      // Error condition: #3: name with no properties has not been previously
       // defined.
       diagnostic.Error(
           *node, fmt::format("Material '{}' not previously defined, but has no"
@@ -176,7 +188,7 @@ UrdfMaterial ParseMaterial(const TinyXml2Diagnostic& diagnostic,
   UrdfMaterial material{rgba, texture_path};
 
   if (!name.empty()) {
-    // Error condition: #3.
+    // Error condition: #4.
     // If a name is *required*, then simply matching names should lead to an
     // error.
     material = AddMaterialToMaterialMap(
--- a/multibody/parsing/parsing_doxygen.h
+++ b/multibody/parsing/parsing_doxygen.h
@@ -414,14 +414,17 @@ be drake::geometry::Convex, rather than drake::geometry::Mesh.
 @subsection tag_drake_diffuse_map drake:diffuse_map
 
 - SDFormat path: `//model/link/visual/material/drake:diffuse_map`
-- URDF path: N/A
+- URDF path: N/A (see a note below).
 - Syntax: URI.
 
 @subsubsection tag_drake_diffuse_map_semantics Semantics
 
 If present, this element indicates (by filename or `package:` URI) a PNG file
 to use as a diffuse texture map. The resolved path name is stored in a
-PerceptionProperties object under `(phong, diffuse_map)`.
+PerceptionProperties object under `(phong, diffuse_map)`. URDF provides a
+built-in tag, i.e., `//robot/material/texture` or
+`//robot/link/visual/material/texture`, to specify a diffuse texture map for a
+link.
 
 @see drake::geometry::PerceptionProperties, drake::multibody::PackageMap,
 @ref render_engine_vtk_properties "Geometry perception properties"
--- a/multibody/parsing/test/detail_urdf_geometry_test.cc
+++ b/multibody/parsing/test/detail_urdf_geometry_test.cc
@@ -740,6 +740,24 @@ TEST_F(UrdfGeometryTest, TestDuplicateVisualName) {
   EXPECT_THAT(TakeWarning(), ContainsRegex("visual.*hello.*renam.*hello_1"));
 }
 
+TEST_F(UrdfGeometryTest, TestDrakeDiffuseMapError) {
+  // If <drake:diffuse_map> is specified, the parser should report an error. The
+  // URDF snippet models a user that copies <drake:diffuse> from an SDFormat
+  // file verbatim (so it maintains the SDFormat style instead of URDF).
+  std::string robot = R"""(
+    <robot name='a'>
+      <link name='b'>
+        <visual name='hello'>
+          <geometry><box size='1 2 3'/></geometry>
+          <material><drake:diffuse_map>abc.png</drake:diffuse_map></material>
+        </visual>
+      </link>
+    </robot>)""";
+  DRAKE_EXPECT_NO_THROW(ParseUrdfGeometryString(robot));
+  EXPECT_THAT(TakeError(),
+              ContainsRegex("<drake:diffuse_map> is not supported.*"));
+}
+
 TEST_F(UrdfGeometryTest, TestDuplicateCollisionName) {
   // Duplicate visual names are allowed by URDF, but not by SceneGraph.
   // They should produce a warning (with automatic renaming).


zachfang:
@SeanCurtis-TRI @jwnimmer-tri This is my initial attempt to add some clarity when specifying a texture map in URDF. Also, if either of you would like to be the feature reviewer assuming the changes make sense to you.

SeanCurtis-TRI:
+@SeanCurtis-TRI  I'll take feature

SeanCurtis-TRI (commented):
I'm not persuaded that *all* of the changes in this PR are necessary or even desirable. The changes to *code* seem to be solving a non-existent problem.

1. The new code guards against the usage of a non-URDF tag in a URDF file. That's generally a problem we don't tackle.
    - Of course, if we're generally validating the XML against a schema, then I take that back. But I assume we're not, or this tag would already cause an error/warning to be dispatched.
2. The *particular* tag you're guarding against is [explicitly defined](https://drake.mit.edu/doxygen_cxx/group__multibody__parsing.html#tag_drake_diffuse_map) as applying to `SDFormat` but *not* `URDF`. So, I wouldn't predict that a user would have the expectation that this tag works in a URDF. I'm open to being persuaded otherwise.

The portion of this PR where you clarify that the offending tag only applies to SDF has the greatest value.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NR_729zCQwRs-FuE_A2:bgx3xy1)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`multibody/parsing/parsing_doxygen.h` line 425 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NR_7wSA66arBfsSsqnf:-NR_7wSA66arBfsSsqng:b-l9n27f) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b08c035c5cce225e3a923b30650264658f74edd/multibody/parsing/parsing_doxygen.h#L425)):*
> ```objc
> to use as a diffuse texture map. The resolved path name is stored in a
> PerceptionProperties object under `(phong, diffuse_map)`. URDF provides a
> built-in tag, i.e., <texture>, to specify a diffuse texture map for a link.
> ```

nit: Instead of just "<texture>", I'd provide the full path(s) as is done for the SDFormat path above:

`//robot/material/texture` or `//robot/link/visual/material/texture`

jwnimmer-tri:
> ... if we're generally validating the XML against a schema ... I assume we're not, or this tag would already cause an error/warning to be dispatched.

That's right; in URDF we don't validate yet.  See #16997.  \CC @rpoyner-tri FYI

> The new code guards against the usage of a non-URDF tag in a URDF file. That's generally a problem we don't tackle.

Given the lack of schema validation, in principle it's OK for us to add specific warnings to mitigate user confusion.  The only question is: do we think this is a mistake a user could plausibly make?  If yes, we should warn.  If no, you're right that we could skip it here with the hope of #16997 down the line.

> The particular tag you're guarding against is [explicitly defined](https://drake.mit.edu/doxygen_cxx/group__multibody__parsing.html#tag_drake_diffuse_map) as applying to SDFormat but not URDF. So, I wouldn't predict that a user would have the expectation that this tag works in a URDF. I'm open to being persuaded otherwise.

That page is really difficult to use.  I don't think we can rely on users consulting or understanding it.

Do we think someone might take some SDFormat example text and paste it into the URDF file?  Seems plausible enough to me, and the "fix" here is pretty cheap.

SeanCurtis-TRI (commented):
If we think there is a credible possibility of someone injecting that tag into a URDF file, then I say a response is good. Do we have any evidence that this has ever happened?

If we decide to respond, I'd suggest that we should raise an error and not a warning. The warning makes sense for a valid URDF with tags Drake doesn't handle yet -- "Hey! Nice URDF. Sorry, I can't do that right now."  In this case, they've gone out of their way to create a generally *invalid* URDF for the express purpose of making Drake happy. There's no reason we should let them just continue on their merry way in that case. They should simply fix their URDF and limit the potential for future copypasta.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NR_EQeL6a2BbIBlSqJz:b-99sw9v)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

zachfang (commented):
> So, I wouldn't predict that a user would have the expectation that this tag works in a URDF. I'm open to being persuaded otherwise.

I am probably a bit biased on that as I did try to add that tag after viewing some SDFormat examples when I was rendering a URDF/xacro file and found out the documentation later on.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NRZ0qjCCVXv64p7wj54:b-h6fwm9)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

SeanCurtis-TRI (commented):
Then I'd suggest that's a slam dunk. We do have evidence that this has happened. And if it can happen to you, it can happen to anyone.

So, back to my suggestion about changing the nature of the response.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NR_Gkom1_e21YVtC6qJ:b-tegd)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

jwnimmer-tri:
> If we decide to respond, I'd suggest that we should raise an error and not a warning.

I'll defer to @rpoyner-tri on the policy here.  I can live with it either way.

rpoyner-tri (commented):
Having a few specific cases covered until #16997 is done is not a terrible outcome. Such code has happened before. ( /me blesses without actually reviewing. )

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NRcpT3DCdexm3KG7tyy:b1km2ss)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

zachfang:
OK. I will summarize the above discussions as follows:
(1) Continue this PR as an intermedia "fix".
(2) Throw a parser **error** instead of a warning.
(3) Might remove the code if 16997 lands in the future.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`multibody/parsing/parsing_doxygen.h` line 425 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NR_7wSA66arBfsSsqnf:-NRdLAncCr5nb3ZV587J:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b08c035c5cce225e3a923b30650264658f74edd/multibody/parsing/parsing_doxygen.h#L425)):*

nit: Instead of just "<texture>", I'd provide the full path(s) as is done for the SDFormat path above:

`//robot/material/texture` or `//robot/link/visual/material/texture`
</blockquote></details>

Done.
___
*[`multibody/parsing/test/detail_urdf_geometry_test.cc` line 750 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NRdLDLmEMvqt815FewR:-NRdLDLmEMvqt815FewS:b-8jh64o) ([raw file](https://github.com/RobotLocomotion/drake/blob/f8f3f479b1c0507969f90c1e7aa0f0ea17ba85a6/multibody/parsing/test/detail_urdf_geometry_test.cc#L750)):*
> ```cpp
>         <visual name='hello'>
>           <geometry><box size='1 2 3'/></geometry>
>           <material><drake:diffuse_map>abc.png</drake:diffuse_map></material>
> ```

FYI. I changed to the actual SDFormat syntax to specify the diffuse map, which was exactly what I did before. :sweat_smile: 



_Code quote:_
```cpp
<material><drake:diffuse_map>abc.png</drake:diffuse_map></material>
```

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NRdTiLv4khZwBnomR9b:bmxilld)* status: 1 unresolved discussion, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`multibody/parsing/detail_urdf_geometry.cc` line 142 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NRdSVoeCg0FWG4hybax:-NRdSVoeCg0FWG4hybay:bckrs7o) ([raw file](https://github.com/RobotLocomotion/drake/blob/f8f3f479b1c0507969f90c1e7aa0f0ea17ba85a6/multibody/parsing/detail_urdf_geometry.cc#L142)):*
nit: Instead of explicitly proposing the built-in tag, I'd include an html to the tag information.


> <drake:diffuse_map> is not supported in URDF. See https://drake.mit.edu/doxygen_cxx/group__multibody__parsing.html#tag_drake_diffuse_map for more details.

This means that whatever we want to say about it, we only have to say it one place. Plus, they might become aware of documentation from which they could learn other stuff.
___
*[`multibody/parsing/parsing_doxygen.h` line 425 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NR_7wSA66arBfsSsqnf:-NRdSxln8_dfLuFURqfC:b-r3pqe7) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b08c035c5cce225e3a923b30650264658f74edd/multibody/parsing/parsing_doxygen.h#L425)):*

Done.
</blockquote></details>

For a URDF both paths I provided above are valid definitions of `<texture>`. I think you probably want to include both.
___
*[`multibody/parsing/test/detail_urdf_geometry_test.cc` line 750 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NRdLDLmEMvqt815FewR:-NRdT8X6CpvFg8JYzZeb:b87cew1) ([raw file](https://github.com/RobotLocomotion/drake/blob/f8f3f479b1c0507969f90c1e7aa0f0ea17ba85a6/multibody/parsing/test/detail_urdf_geometry_test.cc#L750)):*
An interesting choice.

It might be worth making of note of what you're modeling. Something like:

> This models a user that copies <drake:diffuse> from an sdf file verbatim (so it maintains SDFormat style instead of URDF).

Otherwise, it's possible that someone reading that in the future would say, "Why does this URDF look like SDF?"

SeanCurtis-TRI (commented):
+@rpoyner-tri for platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NRdY5rR3EON7hAOESzj:bqljb9l)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform), missing label for release notes (waiting on @rpoyner-tri and @zachfang)

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066)* status: LGTM missing from assignee rpoyner-tri(platform), missing label for release notes (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_urdf_geometry.cc` line 142 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NRdSVoeCg0FWG4hybax:-NRdnZ353bzMVRIlwZxY:bt0owyg) ([raw file](https://github.com/RobotLocomotion/drake/blob/f8f3f479b1c0507969f90c1e7aa0f0ea17ba85a6/multibody/parsing/detail_urdf_geometry.cc#L142)):*

nit: Instead of explicitly proposing the built-in tag, I'd include an html to the tag information.


> <drake:diffuse_map> is not supported in URDF. See https://drake.mit.edu/doxygen_cxx/group__multibody__parsing.html#tag_drake_diffuse_map for more details.

This means that whatever we want to say about it, we only have to say it one place. Plus, they might become aware of documentation from which they could learn other stuff.
</blockquote></details>

Done. I can merge the second and third lines together if the formatting looks prettier.
___
*[`multibody/parsing/parsing_doxygen.h` line 425 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NR_7wSA66arBfsSsqnf:-NRdnSMn12Ex_VIaF0Xi:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b08c035c5cce225e3a923b30650264658f74edd/multibody/parsing/parsing_doxygen.h#L425)):*

For a URDF both paths I provided above are valid definitions of `<texture>`. I think you probably want to include both.
</blockquote></details>

Done.
___
*[`multibody/parsing/test/detail_urdf_geometry_test.cc` line 750 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-NRdLDLmEMvqt815FewR:-NRdnVqT6eT3dMeJilja:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f8f3f479b1c0507969f90c1e7aa0f0ea17ba85a6/multibody/parsing/test/detail_urdf_geometry_test.cc#L750)):*

An interesting choice.

It might be worth making of note of what you're modeling. Something like:

> This models a user that copies <drake:diffuse> from an sdf file verbatim (so it maintains SDFormat style instead of URDF).

Otherwise, it's possible that someone reading that in the future would say, "Why does this URDF look like SDF?"
</blockquote></details>

Done.

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NReDnILCEJqHtqUelOk:bnfp4nl)* status: missing label for release notes (waiting on @zachfang)

rpoyner-tri (commented):
-(release notes: feature) +(release notes: fix) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19066#-:-NReKOMf03PnOOsyc82A:b-rtv606)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),SeanCurtis-TRI(platform) (waiting on @zachfang)

PR #19082, opened 2023-03-28
Fix the ordering of the Delassus operator approximation

amcastro-tri:
Both `SapConstraintBundle`'s constructor and `SapModel::CalcDelassusDiagonalApproximation()` document that the Delassus operator diagonal approximation is stored in the order constraints are enumerated in `SapProblem`.

`SapConstraintBundle` expects this order. However, `SapModel::CalcDelassusDiagonalApproximation()` was inconsistently ordering them by cluster. This can lead to very bad scaling of the constraints, specially so in problems with large mass ratios.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082)

### DIFF
--- a/multibody/contact_solvers/sap/sap_model.cc
+++ b/multibody/contact_solvers/sap/sap_model.cc
@@ -430,13 +430,8 @@ void SapModel<T>::CalcDelassusDiagonalApproximation(
   // Compute Delassus_diagonal as the rms norm of the diagonal block for the
   // i-th constraint.
   delassus_diagonal->resize(num_constraints);
-  int permuted_constraint_index = 0;
-  for (const ContactProblemGraph::ConstraintCluster& e :
-       problem().graph().clusters()) {
-    for (int i : e.constraint_index()) {
-      (*delassus_diagonal)[permuted_constraint_index++] =
-          W[i].norm() / W[i].rows();
-    }
+  for (int i = 0; i < num_constraints; ++i) {
+    (*delassus_diagonal)[i] = W[i].norm() / W[i].rows();
   }
 }
 
--- a/multibody/contact_solvers/sap/test/sap_model_test.cc
+++ b/multibody/contact_solvers/sap/test/sap_model_test.cc
@@ -572,6 +572,8 @@ class DummyModelTest : public ::testing::Test {
 
     // Now we compute a diagonal scaling for each constraints by taking the RMS
     // norm of the diagonal block for that constraint.
+    // The i-th entry in W_diagonal_approximation must contain the approximation
+    // corresponding to the i-th constraint.
     VectorXd W_diagonal_approximation =
         VectorXd::Zero(sap_problem_->num_constraints());
     for (int i = 0; i < sap_problem_->num_constraints(); ++i) {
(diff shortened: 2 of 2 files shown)


amcastro-tri (commented):
+@xuchenhan-tri for feature review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082#-:-NRdliufDWbrchtEjFZ6:bzgdhpr)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
:lgtm_strong: This seems very hard to find, thanks!

One minor comment on test.

+@rpoyner-tri for platform review if you have time to squeeze this in for today.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082#-:-NRe33ov-vAb0bZclC6Y:bv49p7j)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @amcastro-tri and @rpoyner-tri)

___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 595 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19082#-NRe-0Vh3s8iTcmd5or4:-NRe-0Vh3s8iTcmd5or5:b-8qing8) ([raw file](https://github.com/RobotLocomotion/drake/blob/57cfc8185d51a29ec89f80502753298ca2759654/multibody/contact_solvers/sap/test/sap_model_test.cc#L595)):*
BTW, this seems unnecessary to me. Having shuffled cluster indexes only prevents the false negative that delassus approximation is (incorrectly) ordered by clusters, but the test (incorrectly) passes due to the dumb luck that clusters and constraints are ordered the same way. However, without the context of Russ's bug, it's hard to see why that might be an important thing to check.

If you think this check deserves its place. I'd tweak the justification a little bit. For example,

> According to the documentation of `SapModel::CalcDelassusDiagonalApproximation`, `delassus_diagonal` should be ordered by constraints, not constraint clusters. Here we make sure that clusters are not ordered so that the test wouldn't accidentally pass because `delassus_diagonal` happen to be ordered by constraints because constraint clusters are ordered.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @xuchenhan-tri)

___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 595 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19082#-NRe-0Vh3s8iTcmd5or4:-NRe6ArcAu9g9pZBbjgV:buqpxpl) ([raw file](https://github.com/RobotLocomotion/drake/blob/57cfc8185d51a29ec89f80502753298ca2759654/multibody/contact_solvers/sap/test/sap_model_test.cc#L595)):*

BTW, this seems unnecessary to me. Having shuffled cluster indexes only prevents the false negative that delassus approximation is (incorrectly) ordered by clusters, but the test (incorrectly) passes due to the dumb luck that clusters and constraints are ordered the same way. However, without the context of Russ's bug, it's hard to see why that might be an important thing to check.

If you think this check deserves its place. I'd tweak the justification a little bit. For example,

> According to the documentation of `SapModel::CalcDelassusDiagonalApproximation`, `delassus_diagonal` should be ordered by constraints, not constraint clusters. Here we make sure that clusters are not ordered so that the test wouldn't accidentally pass because `delassus_diagonal` happen to be ordered by constraints because constraint clusters are ordered.
</blockquote></details>

Very good point. I updated the comment as suggested.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082)* status: LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19082#-:-NReLT-9DIWjDK0ybhSA:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),xuchenhan-tri(platform) (waiting on @amcastro-tri)

PR #19083, opened 2023-03-28
[multibody] Add MatrixBlock class

xuchenhan-tri:
MatrixBlock wraps around std::variant<MatrixX<T>, Block3x3SparseMatrix<T>> and provides a common set of API for a set of matrix operations that leverages sparsity when possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -20,6 +20,7 @@ drake_cc_package_library(
         ":contact_solver_results",
         ":contact_solver_utils",
         ":linear_operator",
+        ":matrix_block",
         ":newton_with_bisection",
         ":pgs_solver",
         ":point_contact_data",
@@ -104,6 +105,17 @@ drake_cc_library(
         "//common:default_scalars",
         "//common:essential",
         "//common:nice_type_name",
+    ],
+)
+
+drake_cc_library(
+    name = "matrix_block",
+    srcs = ["matrix_block.cc"],
+    hdrs = ["matrix_block.h"],
+    deps = [
+        ":block_3x3_sparse_matrix",
+        "//common:default_scalars",
+        "//common:essential",
     ],
 )
 
@@ -197,6 +209,14 @@ drake_cc_googletest(
         ":block_sparse_matrix",
         ":linear_operator",
         "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
+    name = "matrix_block_test",
+    deps = [
+        ":matrix_block",
+        "//common/test_utilities:eigen_matrix_compare",
     ],
 )
 
--- a/multibody/contact_solvers/matrix_block.cc
+++ b/multibody/contact_solvers/matrix_block.cc
@@ -1,0 +1,249 @@
+#include "drake/multibody/contact_solvers/matrix_block.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+template <class T>
+MatrixBlock<T>::MatrixBlock(Block3x3SparseMatrix<T> data)
+    : data_(std::move(data)), is_dense_(false) {}
+
+template <class T>
+MatrixBlock<T>::MatrixBlock(MatrixX<T> data)
+    : data_(std::move(data)), is_dense_(true) {}
+
+template <class T>
+int MatrixBlock<T>::rows() const {
+  return std::visit(
+      [](auto&& arg) {
+        /* We need the static_cast here because Eigen's rows() is `long`. */
+        return static_cast<int>(arg.rows());
+      },
+      data_);
+}
+
+template <class T>
+int MatrixBlock<T>::cols() const {
+  return std::visit(
+      [](auto&& arg) {
+        /* We need the static_cast here because Eigen's cols() is `long`. */
+        return static_cast<int>(arg.cols());
+      },
+      data_);
+}
+
+template <class T>
+void MatrixBlock<T>::MultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
+                                      EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(cols() == A.rows());
+  DRAKE_DEMAND(rows() == y->rows());
+  DRAKE_DEMAND(A.cols() == y->cols());
+
+  if (is_dense_) {
+    const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
+    *y += M_dense * A;
+    return;
+  }
+  const Block3x3SparseMatrix<T>& M_sparse =
+      std::get<Block3x3SparseMatrix<T>>(data_);
+  M_sparse.MultiplyAndAddTo(A, y);
+}
+
+template <class T>
+void MatrixBlock<T>::TransposeAndMultiplyAndAddTo(
+    const Eigen::Ref<const MatrixX<T>>& A, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(cols() == y->rows());
+  DRAKE_DEMAND(rows() == A.rows());
+  DRAKE_DEMAND(A.cols() == y->cols());
+
+  if (is_dense_) {
+    const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
+    *y += M_dense.transpose() * A;
+    return;
+  }
+  const Block3x3SparseMatrix<T>& M_sparse =
+      std::get<Block3x3SparseMatrix<T>>(data_);
+  M_sparse.TransposeAndMultiplyAndAddTo(A, y);
+}
+
+// TODO(xuchenhan-tri): consider a double dispatch strategy where each block
+// type provides an API to operate on every other block type.
+template <class T>
+void MatrixBlock<T>::TransposeAndMultiplyAndAddTo(const MatrixBlock<T>& A,
+                                                 EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(cols() == y->rows());
+  DRAKE_DEMAND(rows() == A.rows());
+  DRAKE_DEMAND(A.cols() == y->cols());
+  if (A.is_dense_) {
+    const MatrixX<T>& A_dense = std::get<MatrixX<T>>(A.data_);
+    this->TransposeAndMultiplyAndAddTo(A_dense, y);
+    return;
+  }
+
+  /* A is sparse. */
+  const Block3x3SparseMatrix<T>& A_sparse =
+      std::get<Block3x3SparseMatrix<T>>(A.data_);
+  if (this->is_dense_) {
+    const MatrixX<T>& M_dense = std::get<MatrixX<T>>(this->data_);
+    A_sparse.LeftMultiplyAndAddTo(M_dense.transpose(), y);
+    return;
+  }
+
+  /* A and M are both sparse. */
+  const Block3x3SparseMatrix<T>& M_sparse =
+      std::get<Block3x3SparseMatrix<T>>(this->data_);
+  M_sparse.TransposeAndMultiplyAndAddTo(A_sparse, y);
+}
+
+// TODO(xuchenhan-tri): consider a double dispatch strategy where each block
+// type provides an API to operate on every other block type.
+template <class T>
+MatrixBlock<T> MatrixBlock<T>::LeftMultiplyByBlockDiagonal(
+    const std::vector<MatrixX<T>>& Gs, int start, int end) const {
+  DRAKE_DEMAND(start >= 0);
+  DRAKE_DEMAND(end >= start);
+  DRAKE_DEMAND(static_cast<int>(Gs.size()) > end);
+  /* Verify that the sizes of G and M is compatible. */
+  int G_rows = 0;
+  for (int i = start; i <= end; ++i) {
+    DRAKE_DEMAND(Gs[i].rows() == Gs[i].cols());
+    G_rows += Gs[i].rows();
+    if (!is_dense_) {
+      DRAKE_DEMAND(Gs[i].rows() == 3);
+    }
+  }
+  DRAKE_DEMAND(G_rows == rows());
+
+  if (is_dense_) {
+    const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
+    MatrixX<T> GM(rows(), cols());
+    int row_offset = 0;
+    for (int i = start; i <= end; ++i) {
+      const int rows = Gs[i].rows();
+      GM.middleRows(row_offset, rows).noalias() =
+          Gs[i] * M_dense.middleRows(row_offset, rows);
+      row_offset += rows;
+    }
+    return MatrixBlock<T>(std::move(GM));
+  }
+
+  const Block3x3SparseMatrix<T>& M_sparse =
+      std::get<Block3x3SparseMatrix<T>>(data_);
+  Block3x3SparseMatrix<T> GM(M_sparse.block_rows(), M_sparse.block_cols());
+  using Triplet = typename Block3x3SparseMatrix<T>::Triplet;
+  const std::vector<std::vector<Triplet>>& M_triplets = M_sparse.get_triplets();
+  std::vector<Triplet> GM_triplets;
+  GM_triplets.reserve(M_sparse.num_blocks());
+  for (const auto& row_data : M_triplets) {
+    for (const Triplet& t : row_data) {
+      const int block_row = std::get<0>(t);
+      const int block_col = std::get<1>(t);
+      const Matrix3<T>& M_block = std::get<2>(t);
+      GM_triplets.emplace_back(block_row, block_col,
+                               Gs[start + block_row] * M_block);
+    }
+  }
+  GM.SetFromTriplets(GM_triplets);
+  return MatrixBlock<T>(std::move(GM));
+}
+
+template <class T>
+void MatrixBlock<T>::MultiplyWithScaledTransposeAndAddTo(
+    const VectorX<T>& scale, EigenPtr<MatrixX<T>> y) const {
+  DRAKE_DEMAND(y != nullptr);
+  DRAKE_DEMAND(cols() == scale.size());
+  DRAKE_DEMAND(rows() == y->rows());
+  DRAKE_DEMAND(rows() == y->cols());
+
+  if (is_dense_) {
+    const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
+    *y += M_dense * scale.asDiagonal() * M_dense.transpose();
+    return;
+  }
+  const Block3x3SparseMatrix<T>& M_sparse =
+      std::get<Block3x3SparseMatrix<T>>(data_);
+  M_sparse.MultiplyWithScaledTransposeAndAddTo(scale, y);
+}
+
+template <class T>
+MatrixX<T> MatrixBlock<T>::MakeDenseMatrix() const {
+  if (is_dense_) {
+    return std::get<MatrixX<T>>(data_);
+  }
+  return std::get<Block3x3SparseMatrix<T>>(data_).MakeDenseMatrix();
+}
+
+template <typename T>
+MatrixBlock<T> StackMatrixBlocks(const std::vector<MatrixBlock<T>>& blocks) {
+  if (blocks.empty()) {
+    return {};
+  }
+
+  const bool is_dense = blocks[0].is_dense_;
+  const int cols = blocks[0].cols();
+  int rows = 0;
+  for (const auto& b : blocks) {
+    DRAKE_DEMAND(is_dense == b.is_dense_);
+    DRAKE_DEMAND(cols == b.cols());
+    rows += b.rows();
+  }
+
+  if (is_dense) {
+    MatrixX<T> result(rows, cols);
+    int row_offset = 0;
+    for (const auto& b : blocks) {
+      result.middleRows(row_offset, b.rows()) = std::get<MatrixX<T>>(b.data_);
+      row_offset += b.rows();
+    }
+    return MatrixBlock<T>(std::move(result));
+  }
+
+  /* Each entry in `blocks` is Block3x3SparseMatrix. */
+  DRAKE_DEMAND(rows % 3 == 0);
+  DRAKE_DEMAND(cols % 3 == 0);
+  const int block_rows = rows / 3;
+  const int block_cols = cols / 3;
+  int block_row_offset = 0;
+  Block3x3SparseMatrix<T> result(block_rows, block_cols);
+  using Triplet = typename Block3x3SparseMatrix<T>::Triplet;
+  std::vector<Triplet> result_triplets;
+  int nonzero_blocks = 0;
+  for (const auto& b : blocks) {
+    const Block3x3SparseMatrix<T>& entry =
+        std::get<Block3x3SparseMatrix<T>>(b.data_);
+    nonzero_blocks += entry.num_blocks();
+  }
+  result_triplets.reserve(nonzero_blocks);
+
+  for (const auto& b : blocks) {
+    const Block3x3SparseMatrix<T>& entry =
+        std::get<Block3x3SparseMatrix<T>>(b.data_);
+    const std::vector<std::vector<Triplet>>& b_triplets = entry.get_triplets();
+    for (const auto& row_data : b_triplets) {
+      for (const Triplet& t : row_data) {
+        const int block_row = std::get<0>(t) + block_row_offset;
+        const int block_col = std::get<1>(t);
+        const Matrix3<T>& m = std::get<2>(t);
+        result_triplets.emplace_back(block_row, block_col, m);
+      }
+    }
+    block_row_offset += entry.block_rows();
+  }
+  result.SetFromTriplets(result_triplets);
+  return MatrixBlock<T>(std::move(result));
+}
+
+DRAKE_DEFINE_FUNCTION_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_SCALARS(
+    (&StackMatrixBlocks<T>))
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
+
+DRAKE_DEFINE_CLASS_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_SCALARS(
+    class ::drake::multibody::contact_solvers::internal::MatrixBlock)
--- a/multibody/contact_solvers/matrix_block.h
+++ b/multibody/contact_solvers/matrix_block.h
@@ -1,0 +1,118 @@
+#pragma once
+
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "drake/common/eigen_types.h"
+#include "drake/multibody/contact_solvers/block_3x3_sparse_matrix.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+template <typename T>
+class MatrixBlock;
+
+/* Row-wise concatenates the given vector of MatrixBlocks into a single
+ MatrixBlock. In other words, given blocks [b₁, b₂, ..., bₙ], the resulting
+ stacked matrix M is
+         _    _
+        |  b₁  |
+        |  b₂  |
+    M = |  ... |
+        |  ... |
+        |_ bₙ _|
+
+ @pre all entries in `blocks` have the same `cols()`.
+ @pre either all entries in `blocks` are dense, or all entries in `blocks` are
+ sparse.
+ @returns A MatrixBlock concatenating the given blocks row-wise that's dense if
+ all the given blocks are dense and sparse otherwise.
+ @tparam_default_scalar */
+template <typename T>
+MatrixBlock<T> StackMatrixBlocks(const std::vector<MatrixBlock<T>>& blocks);
+
+/* Data structure to store individual non-zero blocks of BlockSparseMatrix.
+ The MatrixBlock can either be dense or sparse, which is determined by how it is
+ constructed. Right now, the only sparsity pattern supported is block 3-by-3
+ (see Block3x3SparseMatrix). Once constructed, the MatrixBlock cannot change
+ from a dense representation to a sparse one or vice versa. This class provides
+ a series of matrix computation functions that exploits the sparsity if the
+ MatrixBlock is sparse and falls back to dense algebra otherwise. We use `M` to
+ denote `this` MatrixBlock throughout this class.
+ @tparam_default_scalar */
+template <class T>
+class MatrixBlock {
+ public:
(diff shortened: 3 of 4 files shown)


sherm1:
Is this ready for feature review? If not please consider making it a Draft or [WIP].

xuchenhan-tri (commented):
+@amcastro-tri for feature review please. Feel free re-delegate if you don't have the bandwidth.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NRe6ccu-BEzjnjFHbij:barrkvi)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri)

amcastro-tri (commented):
working...

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NS83yNuDJ5YS35GwdF5:b-s297dq)* status: 7 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 30 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7iYjy51DF725E1M-G:-NS7iYjy51DF725E1M-H:b-52ypqt) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L30)):*
> ```objc
>  @pre all entries in `blocks` have the same `cols()`.
>  @pre either all entries in `blocks` are dense, or all entries in `blocks` are
>  sparse.
> ```

Looking at this requirement, I imagine this determines the underlyaing structure of the resulting block. That is, if we pass all dense blocks then the result is a single dense block and if we pass all 3x3 bocks, then the result is sparse. 
Maybe say something in the docs about this?
___
*[`multibody/contact_solvers/matrix_block.h` line 67 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7tiFbBfqZpAnQOOYM:-NS7tiFbBfqZpAnQOOYN:bppi0k8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L67)):*
> ```objc
>    @pre y != nullptr and the sizes of A and y are compatible with M. */
>   void MultiplyAndAddTo(const Eigen::Ref<const MatrixX<T>>& A,
>                         EigenPtr<VectorX<T>> y) const;
> ```

should this one be EigenPtr to MatrixX?
___
*[`multibody/contact_solvers/matrix_block.h` line 77 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7u3KQ1N04hhxMB6r_:-NS7u3KQ1N04hhxMB6ra:b-my34ck) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L77)):*
> ```objc
>    @pre y != nullptr and the sizes of A and y are compatible with M. */
>   void TransposeAndMultiplyAndAddTo(const MatrixBlock<T>& A,
>                                    EigenPtr<MatrixX<T>> y) const;
> ```

Wondering if y should be a MatrixBlock here instead. If not, maybe consider mentioning that user might need to consider an alternative API here that can write into a MatrixBlock to take additional advantage of sparsity
___
*[`multibody/contact_solvers/matrix_block.h` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7x1Hu1PHiEibsuCQv:-NS7x1Hv25wZGUhrZi9u:bgbzo98) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L88)):*
> ```objc
>    @pre If M is sparse, all matrices in Gs are 3x3. */
>   MatrixBlock<T> LeftMultiplyByBlockDiagonal(const std::vector<MatrixX<T>>& Gs,
>                                              int start, int end) const;
> ```

minor, wondering why this API did not follow the `DoAndAddTo()` pattern above, passing a mutable pointer to a `MatrixBlock`.
And just double checking, do you really need this API with start/end as inputs? will they be difreent from start=0 and Gs.size() in your case?
___
*[`multibody/contact_solvers/matrix_block.h` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7jN92Anzl5ADMJiqN:-NS7jN93EbiDSA0Hm_1y:b8befl9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L101)):*
> ```objc
>  private:
>   friend MatrixBlock<T> StackMatrixBlocks<T>(
>       const std::vector<MatrixBlock<T>>& blocks);
> ```

does this method need to live outside the MatrixBlock class? why not to make a private function?
Otherwise, why not to expose public methods `is_dense()` and/or 
___
*[`multibody/contact_solvers/matrix_block.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7yqoNB2oJZAbSESEX:-NS7yqoNB2oJZAbSESEY:b-z81jc7) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L42)):*
> ```cpp
>   DRAKE_DEMAND(cols() == A.rows());
>   DRAKE_DEMAND(rows() == y->size());
> ```

demand `y.cols() == A.cols()`?
___
*[`multibody/contact_solvers/matrix_block.cc` line 97 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS80EW24TOt3DvpJs5F:-NS80EW24TOt3DvpJs5G:bumqq3c) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L97)):*
> ```cpp
>   const Block3x3SparseMatrix<T>& M_sparse =
>       std::get<Block3x3SparseMatrix<T>>(this->data_);
>   M_sparse.TransposeAndMultiplyAndAddTo(A_sparse, y);
> ```

nit, there is a bit of a combinatorial explosion going on here with these ifs. Not yet, but it'd explode if we add a third variant.

I wonder if you considered a double-dispatch strategy. For that to work, probably we'd need a class `DenseBlock` that wraps a MatrixX so that it has the same APIs as `Block3x3SparseMatrix`.
Then a method like this one would look like:
```c++
void MatrixBlock<T>::TransposeAndMultiplyAndAddTo(const MatrixBlock<T>& A,
                                                 EigenPtr<MatrixX<T>> y) const {

  

  return std::visit(
      [this](auto&& Avariant) {
        if(is_dense) {
           const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
           Avariant.LeftMultiplyByTransposeAndAddTo(M_dense);
       } else
             const Block3x3SparseMatrix<T>& M_sparse =
             std::get<Block3x3SparseMatrix<T>>(data_);
             Avariant.LeftMultiplyByTransposeAndAddTo(M_sparse);             
       }
      },
      A.data_);
```

I think in this way each block type would need to know how to operate on every other block type, but the implementation would live locally within the class that knows how to operate on its own internals. 

I don't think this makes things easier for the two variants case, only if we happen to have more variants. Maybe at TODO is ok.

amcastro-tri (commented):
Feature review done.
:lgtm_strong: 

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NSBWHe2-dZ5QSIw6QHG:b4pmuxx)* status: 10 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/matrix_block.cc` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS85_wADJoBuUpv7poZ:-NS85_wB1KgjjOPDJ6Aw:b-uzoh20) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L143)):*
> ```cpp
>       const Matrix3<T>& M_block = std::get<2>(t);
>       GM_triplets.emplace_back(block_row, block_col,
>                                Gs[start + block_row] * M_block);
> ```

this is where I think this "double dispatch" idea would help clean up things also. This code could read:
```c++
if (is_dense) {
  const DenseBlock<T>& M_dense = std::get<DenseBlock<T>>(data_);
  return MatrixBlock<T>(M_dense.LeftMultiplyByBlockDiagonal(Gs, start, end));  // Dense block knows how to perform this operation.  
} else {
  const Block3x3SparseMatrix<T>& M_sparse = std::get<Block3x3SparseMatrix<T>>(data_);
  return MatrixBlock<T>(M_sparse.LeftMultiplyByBlockDiagonal(Gs, start, end));  // Sparse block knows how to perform this operation.    
}

```
with separate unit testing contained to specific test files.
___
*[`multibody/contact_solvers/test/matrix_block_test.cc` line 30 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSBO3rU1Jn7on4t61Oc:-NSBO3rU1Jn7on4t61Od:b-qqtwk6) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/test/matrix_block_test.cc#L30)):*
> ```cpp
> /* Returns a dummy 3x3 matrix with all entries in the matrix equal to the given
>  value. */
> Matrix3d Make3x3Matrix(double value) {
> ```

We have `Matrix3::Constant(value)` so no sure how much this adds. If it does to you, maybe rename to `MakeConstantMatrix3()`?
___
*[`multibody/contact_solvers/test/matrix_block_test.cc` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSBPLRP1wAWY2SsgF-w:-NSBPLRP1wAWY2SsgF-x:bu7l6ps) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/test/matrix_block_test.cc#L73)):*
> ```cpp
> 
>   /* Set the destinations to compatible-sized non-zero vectors. */
>   VectorXd y1 = MakeArbitraryMatrix(12, 1);
> ```

nit, if only the size matters and not the values (even when arbitrary), I'd probably just use Eigen's constructors here:
```c++
VectorXd y1(12);
VectorXd y2(12);
VectorXd expected_y(12);
```

xuchenhan-tri (commented):
Comments addressed. PTAL, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NSCiYJr56vpCL83EmjF:b-pr0h6x)* status: 10 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 30 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7iYjy51DF725E1M-G:-NSCSkN73WHrDR56XOMh:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L30)):*

Looking at this requirement, I imagine this determines the underlyaing structure of the resulting block. That is, if we pass all dense blocks then the result is a single dense block and if we pass all 3x3 bocks, then the result is sparse. 
Maybe say something in the docs about this?
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/matrix_block.h` line 67 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7tiFbBfqZpAnQOOYM:-NSCTVSx96QN68ojIksf:b-hqzll7) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L67)):*

should this one be EigenPtr to MatrixX?
</blockquote></details>

Done.

Yes, nice catch!
___
*[`multibody/contact_solvers/matrix_block.h` line 77 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7u3KQ1N04hhxMB6r_:-NSCcsmR04_B-9ZLk5px:bwk8hc1) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L77)):*

Wondering if y should be a MatrixBlock here instead. If not, maybe consider mentioning that user might need to consider an alternative API here that can write into a MatrixBlock to take additional advantage of sparsity
</blockquote></details>

I added a TODO here.

Currently, the only time that the sparsity structure is preserved through the multiplication when we call this function is in the supernodal solver. However, the callsite immediately discards the sparsity because conex works with dense blocks. We may be able to leverage the sparsity if we later adopt a different supernodal solver.
___
*[`multibody/contact_solvers/matrix_block.h` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7x1Hu1PHiEibsuCQv:-NSCU0qxDcMERRByPBlD:bdtc15n) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L88)):*

> And just double checking, do you really need this API with start/end as inputs? will they be difreent from start=0 and Gs.size() in your case?

Yes, the SuperNodalSolver class needs this. This is the (intended future) [callsite](https://github.com/RobotLocomotion/drake/blob/0e80c9d129e053f9670658e2aa95be7e39567140/multibody/contact_solvers/supernodal_solver.cc#L46).
___
*[`multibody/contact_solvers/matrix_block.h` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7jN92Anzl5ADMJiqN:-NSCUkMe87w3kqGsg5ic:b-ke3nz6) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L101)):*
> does this method need to live outside the MatrixBlock class? why not to make a private function?

This function is useful for to stack up constraint jacobians for the constraints in the same cluster. See (intended future) [callsite](https://github.com/RobotLocomotion/drake/blob/0e80c9d129e053f9670658e2aa95be7e39567140/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L107).

> Otherwise, why not to expose public methods `is_dense()`.

Done.
___
*[`multibody/contact_solvers/matrix_block.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7yqoNB2oJZAbSESEX:-NSCbk6eCH_K6o4d6dTT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L42)):*

demand `y.cols() == A.cols()`?
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/matrix_block.cc` line 97 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS80EW24TOt3DvpJs5F:-NSCerEH6bPP9FCLoNWJ:bn2umz1) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L97)):*

nit, there is a bit of a combinatorial explosion going on here with these ifs. Not yet, but it'd explode if we add a third variant.

I wonder if you considered a double-dispatch strategy. For that to work, probably we'd need a class `DenseBlock` that wraps a MatrixX so that it has the same APIs as `Block3x3SparseMatrix`.
Then a method like this one would look like:
```c++
void MatrixBlock<T>::TransposeAndMultiplyAndAddTo(const MatrixBlock<T>& A,
                                                 EigenPtr<MatrixX<T>> y) const {

  

  return std::visit(
      [this](auto&& Avariant) {
        if(is_dense) {
           const MatrixX<T>& M_dense = std::get<MatrixX<T>>(data_);
           Avariant.LeftMultiplyByTransposeAndAddTo(M_dense);
       } else
             const Block3x3SparseMatrix<T>& M_sparse =
             std::get<Block3x3SparseMatrix<T>>(data_);
             Avariant.LeftMultiplyByTransposeAndAddTo(M_sparse);             
       }
      },
      A.data_);
```

I think in this way each block type would need to know how to operate on every other block type, but the implementation would live locally within the class that knows how to operate on its own internals. 

I don't think this makes things easier for the two variants case, only if we happen to have more variants. Maybe at TODO is ok.

</blockquote></details>

I added a TODO.
___
*[`multibody/contact_solvers/matrix_block.cc` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS85_wADJoBuUpv7poZ:-NSCf02X63_75z7Fpoap:b-hukn99) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L143)):*

this is where I think this "double dispatch" idea would help clean up things also. This code could read:
```c++
if (is_dense) {
  const DenseBlock<T>& M_dense = std::get<DenseBlock<T>>(data_);
  return MatrixBlock<T>(M_dense.LeftMultiplyByBlockDiagonal(Gs, start, end));  // Dense block knows how to perform this operation.  
} else {
  const Block3x3SparseMatrix<T>& M_sparse = std::get<Block3x3SparseMatrix<T>>(data_);
  return MatrixBlock<T>(M_sparse.LeftMultiplyByBlockDiagonal(Gs, start, end));  // Sparse block knows how to perform this operation.    
}

```
with separate unit testing contained to specific test files.
</blockquote></details>

Added TODO.
___
*[`multibody/contact_solvers/test/matrix_block_test.cc` line 30 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSBO3rU1Jn7on4t61Oc:-NSCgS810HbjFw9fxGOT:bs2vhuc) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/test/matrix_block_test.cc#L30)):*

We have `Matrix3::Constant(value)` so no sure how much this adds. If it does to you, maybe rename to `MakeConstantMatrix3()`?
</blockquote></details>

Good point, removed.
___
*[`multibody/contact_solvers/test/matrix_block_test.cc` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSBPLRP1wAWY2SsgF-w:-NSChdiC1O_uFasj2E49:b-24iou9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/test/matrix_block_test.cc#L73)):*

nit, if only the size matters and not the values (even when arbitrary), I'd probably just use Eigen's constructors here:
```c++
VectorXd y1(12);
VectorXd y2(12);
VectorXd expected_y(12);
```
</blockquote></details>

I don't think using uninitialized values for testing is a good idea.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 77 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7u3KQ1N04hhxMB6r_:-NSGPSuk122TCm34hhfS:bird4b0) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L77)):*

I added a TODO here.

Currently, the only time that the sparsity structure is preserved through the multiplication when we call this function is in the supernodal solver. However, the callsite immediately discards the sparsity because conex works with dense blocks. We may be able to leverage the sparsity if we later adopt a different supernodal solver.
</blockquote></details>

ah, thanks. 
Just so that I am aware of what's coming. What about the new solver you are writing? Would this capability help to build the sparse Hessian?
___
*[`multibody/contact_solvers/matrix_block.h` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7x1Hu1PHiEibsuCQv:-NSGPyeM7lmUuqSQeWpS:b-luo5qc) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L88)):*


> And just double checking, do you really need this API with start/end as inputs? will they be difreent from start=0 and Gs.size() in your case?

Yes, the SuperNodalSolver class needs this. This is the (intended future) [callsite](https://github.com/RobotLocomotion/drake/blob/0e80c9d129e053f9670658e2aa95be7e39567140/multibody/contact_solvers/supernodal_solver.cc#L46).

</blockquote></details>

Thank you for the reference. That callsite already passes a, presumably already allocated, `temp` matrix. You might consider this method to also take a mutable pointer to the result as the others do? non-blocking
___
*[`multibody/contact_solvers/matrix_block.h` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7jN92Anzl5ADMJiqN:-NSGQb1N3dcnJ7E1IINe:b-9g7spu) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L101)):*

> does this method need to live outside the MatrixBlock class? why not to make a private function?

This function is useful for to stack up constraint jacobians for the constraints in the same cluster. See (intended future) [callsite](https://github.com/RobotLocomotion/drake/blob/0e80c9d129e053f9670658e2aa95be7e39567140/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L107).

> Otherwise, why not to expose public methods `is_dense()`.

Done.
</blockquote></details>

I was trying to avoid the friendship, or justify it. 
If the MatrixBlock exposes enough APIs maybe this method does not need to be friend?
or you might consider making it static?
___
*[`multibody/contact_solvers/matrix_block.cc` line 97 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS80EW24TOt3DvpJs5F:-NSGQt8jAeTkwwYasyDg:b-5936v8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.cc#L97)):*

I added a TODO.
</blockquote></details>

SGTM. Right now  we only have two axes so not much of an explosion yet.
___
*[`multibody/contact_solvers/test/matrix_block_test.cc` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSBPLRP1wAWY2SsgF-w:-NSGR8hT9F08k91t6iTr:bgkuyee) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/test/matrix_block_test.cc#L73)):*

I don't think using uninitialized values for testing is a good idea.
</blockquote></details>

ah, you are right, these methods "add to".

amcastro-tri (commented):
+@EricCousineau-TRI for platform per rotation (for Thursday. Today Wednesday Xuchen is platform).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NSGRsYk6v5yPPT0c0uM:b-le195h)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI and @xuchenhan-tri)

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @EricCousineau-TRI)

___
*[`multibody/contact_solvers/matrix_block.h` line 77 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7u3KQ1N04hhxMB6r_:-NSHIBMX9hyeeqcNuMzM:b-1r1cfv) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L77)):*

ah, thanks. 
Just so that I am aware of what's coming. What about the new solver you are writing? Would this capability help to build the sparse Hessian?
</blockquote></details>

Not particularly. I'm using the `MatrixBlock` class to store jacobian blocks that corresponds to a single patch-clique pair. The sparsity of the Hessian exists at a higher level and treats each non-zero block as dense.
___
*[`multibody/contact_solvers/matrix_block.h` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7x1Hu1PHiEibsuCQv:-NSHJRn0EzV5p8XLRzsu:b-dwnjdw) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L88)):*

Thank you for the reference. That callsite already passes a, presumably already allocated, `temp` matrix. You might consider this method to also take a mutable pointer to the result as the others do? non-blocking
</blockquote></details>

In this case, the allocation needs to happen one way or another. One either has to create a `temp` matrix and pass it in as a mutable pointer, or allocate by returning by value. So passing in a mutable pointer doesn't save anything. OTOH, with returning by value, we know whether the result is sparse or dense as well as the sparsity pattern if the result is indeed sparse based on the object that invokes the function. You don't have that information when creating the `temp` matrix.
___
*[`multibody/contact_solvers/matrix_block.h` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7jN92Anzl5ADMJiqN:-NSHN4w_2zkDKv5wHG_K:b-h0cczc) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L101)):*
I don't think friending the StackMatrixBlocks function breaks encapsulation in anyway. See https://drake.mit.edu/styleguide/cppguide.html#Friends for justification. The usage is similar to the `FooBuilder` example in the styleguide.

> or you might consider making it static?

Could you explain the advantage of making it static?

xuchenhan-tri (commented):
-@EricCousineau-TRI +@rpoyner-tri for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NSIOhbL3NXTz51uH26c:b-t5590p)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @rpoyner-tri)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083)* status: LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @rpoyner-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7x1Hu1PHiEibsuCQv:-NSLr3GO2MSEBlHmplYH:bieqch5) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L88)):*
> You don't have that information when creating the temp matrix.

Ah, that's right, thanks.
___
*[`multibody/contact_solvers/matrix_block.h` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NS7jN92Anzl5ADMJiqN:-NSLrDiV0uVuPo3v5BQU:b-uwckm5) ([raw file](https://github.com/RobotLocomotion/drake/blob/1dabf20b5f20ff0ded2c864a5a2feb5c896ccb17/multibody/contact_solvers/matrix_block.h#L101)):*
The FooBuilder is a good example, I think you are right this fits in that category. 

> the advantage of making it static?

TBH, here we'd be in the realm of c++ bike shedding and/or author preferences. That's why I made this comment non-blocking.
I can see both ways being equally valid.

rpoyner-tri (commented):
:lgtm: notation question

> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-:-NSMcOM30tSRXsD4Xpuj:boq2uut)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 82 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSMbrUB7ZWRqg9Z7NdK:-NSMbrUB7ZWRqg9Z7NdL:b-fs97fw) ([raw file](https://github.com/RobotLocomotion/drake/blob/913e2b64e74abfc8205fb74554b0819328b36188/multibody/contact_solvers/matrix_block.h#L82)):*
> ```objc
>   // TODO(xuchenhan-tri): Consider writing the output to a MatrixBlock to
>   // preserve the sparsity structure if it exists.
>   /* Performs *y += Mᵀ * A.
> ```

nit I'm curious why the `*` is here and not in the other overloaded method documentation. This the `*` some math symbol?



_Code quote:_
```objc
*
```

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19083)* status: 1 unresolved discussion (waiting on @amcastro-tri and @rpoyner-tri)

___
*[`multibody/contact_solvers/matrix_block.h` line 82 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19083#-NSMbrUB7ZWRqg9Z7NdK:-NSMd3X00eqn2ST9dY76:b1wrlrf) ([raw file](https://github.com/RobotLocomotion/drake/blob/913e2b64e74abfc8205fb74554b0819328b36188/multibody/contact_solvers/matrix_block.h#L82)):*

nit I'm curious why the `*` is here and not in the other overloaded method documentation. This the `*` some math symbol?
</blockquote></details>

No, it's a vestige of an incomplete cleanup.

PR #19112, opened 2023-03-31
[visualization] Hide near-zero hydroelastic force or moment.

DamrongGuoy:
Before this fix, MeshCat used the arrows from the last sizable force or moment in the previous time step.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112)

### DIFF
--- a/multibody/meshcat/BUILD.bazel
+++ b/multibody/meshcat/BUILD.bazel
@@ -127,6 +127,7 @@ drake_cc_library(
         ":contact_visualizer_params",
         "//geometry:meshcat",
         "//math:geometric_transform",
+        "@msgpack_internal//:msgpack",
     ],
 )
 
--- a/multibody/meshcat/BUILD.bazel
+++ b/multibody/meshcat/BUILD.bazel
@@ -65,6 +65,7 @@ drake_cc_googletest(
     deps = [
         ":hydroelastic_contact_visualizer",
         "//common/test_utilities:expect_no_throw",
+        "@msgpack_internal//:msgpack",
     ],
 )
 
@@ -127,7 +128,6 @@ drake_cc_library(
         ":contact_visualizer_params",
         "//geometry:meshcat",
         "//math:geometric_transform",
-        "@msgpack_internal//:msgpack",
     ],
 )
 
--- a/multibody/meshcat/hydroelastic_contact_visualizer.cc
+++ b/multibody/meshcat/hydroelastic_contact_visualizer.cc
@@ -70,7 +70,8 @@ void HydroelasticContactVisualizer::Update(
     if (force_norm >= params_.force_threshold) {
       meshcat_->SetTransform(path + "/force_C_W",
                              RigidTransformd(RotationMatrixd::MakeFromOneVector(
-                                 item.force_C_W, 2)), time);
+                                 item.force_C_W, 2)),
+                             time);
 
       // Stretch the cylinder in z.
       const double height = force_norm / params_.newtons_per_meter;
@@ -88,6 +89,9 @@ void HydroelasticContactVisualizer::Update(
           RigidTransformd(RotationMatrixd::MakeXRotation(M_PI),
                           Vector3d{0, 0, height + arrowhead_height}),
           time);
+      meshcat_->SetProperty(path + "/force_C_W", "visible", true, time);
+    } else {
+      meshcat_->SetProperty(path + "/force_C_W", "visible", false, time);
     }
     // Moment vector.
     if (moment_norm >= params_.moment_threshold) {
@@ -112,6 +116,9 @@ void HydroelasticContactVisualizer::Update(
           RigidTransformd(RotationMatrixd::MakeXRotation(M_PI),
                           Vector3d{0, 0, height + arrowhead_height}),
           time);
+      meshcat_->SetProperty(path + "/moment_C_W", "visible", true, time);
+    } else {
+      meshcat_->SetProperty(path + "/moment_C_W", "visible", false, time);
     }
 
     // Contact surface
--- a/multibody/meshcat/test/hydroelastic_contact_visualizer_test.cc
+++ b/multibody/meshcat/test/hydroelastic_contact_visualizer_test.cc
@@ -19,7 +19,7 @@ using Eigen::Matrix3Xi;
 using geometry::Meshcat;
 
 // Helper to query meshcat whether an item is visible or not.
-bool visible(Meshcat& meshcat, std::string_view path) {
+bool visible(const Meshcat& meshcat, std::string_view path) {
   std::string property = meshcat.GetPackedProperty(path, "visible");
   msgpack::object_handle oh =
       msgpack::unpack(property.data(), property.size());
--- a/multibody/meshcat/test/hydroelastic_contact_visualizer_test.cc
+++ b/multibody/meshcat/test/hydroelastic_contact_visualizer_test.cc
@@ -5,6 +5,7 @@
 #include <gtest/gtest.h>
 
 #include "drake/common/test_utilities/expect_no_throw.h"
+#include "drake/geometry/meshcat_types.h"
 
 namespace drake {
 namespace multibody {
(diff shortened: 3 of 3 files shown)


DamrongGuoy (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NRtx0bp8U0s39-j81ca:b7bdky2)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @joemasterjohn)

joemasterjohn (commented):
I don't believe this totally fixes the bug you're referring to. Just to clarify here:

**The problem**:
When looking at the last frame/step of hydro contact simulation visualization, meshcat shows a rather large moment vector (~10 cm) indicating a torque magnitude of around .3 Nm (given the default value of `ContactVisualizerParams::newton_meters_per_meter`). But this disagrees with the values read from the contact results port that show a torque of magnitude 1e-12. 

**The bug**: 
`HydroelasticContactVisualizer` has logic to hide "inactive" contacts if both the force and torque component are below a parameterized magnitude threshold. If however one of either force or torque is above its threshold (the force in the above described example) then it enters a block to scale the drawn vectors to the corresponding lengths from the contact results. But the blocks rescaling and transforming the force arrow and torque arrow are separate and each contain their own conditional statement saying to not perform the rescaling if the corresponding magnitude is below the threshold.

Thus if the previous frame contained a valid force and moment and drew them accordingly, but the next frame had only a force above the threshold the contact at that path would be visible but only the force would be rescaled. The result being the moment vector is shown, but with length corresponding to the magnitude of the moment from the previous frame's contact.

**The (proposed) fix**:
Leave in the logic that requires both force and moment to be below their respective thresholds in order to hide/delete a contact visualization. Remove the individual clauses (I'll call them (1) and (2) here to refer in the code below) that decide whether to scale either force or moment vectors.

This will result in the smaller of the two vectors being scaled by its small magnitude and being displayed properly.

Note: The same effect can be achieved by keeping track of the path for the force vector and moment vector separately and hiding them on an individual basis. This would require much more work in terms or restructuring logic though.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSCXFnD401E0U8KlYp3:b-ptsnq3)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @DamrongGuoy)

___
*[`multibody/meshcat/hydroelastic_contact_visualizer.cc` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-NSCeYzIAhqoslBvYn3U:-NSCeYzJC3XFEleDyXGH:b-ushn1d) ([raw file](https://github.com/RobotLocomotion/drake/blob/467523632b044d658015a9b7072135b7a6a31cf6/multibody/meshcat/hydroelastic_contact_visualizer.cc#L71)):*
> ```cpp
>     // Force vector.
>     {
>       if (force_norm >= params_.force_threshold) {
> ```

This is what I reffered to as (1) above and can be removed.



_Code quote:_
```cpp
      if (force_norm >= params_.force_threshold) {
```
___
*[`multibody/meshcat/hydroelastic_contact_visualizer.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-NSCe4TZ9s8S2saieSPW:-NSCe4TZ9s8S2saieSPX:b-z05j0i) ([raw file](https://github.com/RobotLocomotion/drake/blob/467523632b044d658015a9b7072135b7a6a31cf6/multibody/meshcat/hydroelastic_contact_visualizer.cc#L99)):*
> ```cpp
>     // Moment vector.
>     {
>       if (moment_norm >= params_.moment_threshold) {
> ```

This is what I reffered to as (2) above and can be removed.



_Code quote:_
```cpp
      if (moment_norm >= params_.moment_threshold) {
```

DamrongGuoy (commented):
@joemasterjohn  I do not understand what you said.  I'll set up a meeting with examples that my solution works.  If you have an example to support your explanation to show, that would be very helpful too.

Let's fix the bug together.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSDzijq4XTEBsMKJV90:b-tmllcm)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @DamrongGuoy)

DamrongGuoy (commented):
@joemasterjohn I changed to our agreed solution.  PTAL.

Thanks for your help.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSHq7nQDccXiRPKqs_O:b2sh79g)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @joemasterjohn)

joemasterjohn (commented):
:lgtm: thanks Damrong! Maybe just for the sake of documenting how we tested this fix, could you provide a link to the example notebook with the Euler disk and somescreenshots of the before/after so the fix is apparent? I think writing up a whole manual test for such a small visual bug would be very cumbersome, but we can check with the platform reivewer.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSHxIZeA1fsJg4Wkh3o:bkkvs7w)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

jwnimmer-tri:
I agree manual testing would be a pain.  However, you can write an automated test for visibility checkboxes.  (See `MeshcatTest, SetPropertyBool` for example.)  A unit test should be able to feed in different contact results, and then check what visibility came out.

DamrongGuoy (commented):
@joemasterjohn .  This is the work-in-progress branch of Euler disk:
https://github.com/DamrongGuoy/drake/blob/testCylinderBoxShowHydroMeshNotebook/tutorials/hydro_compliant_cylinder_compliant_box.ipynb

The following picture is before the fix.  The blue moment vector is very long even though the white Euler's disk already stopped. (It stopped [precession](https://en.wikipedia.org/wiki/Precession) suddenly, so it grabbed the previous long moment vector to draw). The printout of ContactResult says the moment should be practically zero `tau=[2.6063725371243063e-12, -4.630148582002595e-12, 1.5155123087971682e-20],`
```
ContactReporter::Publish() called at time=4.0
num_hydroelastic_contacts() = 1

hydroelastic_contact_info(0): 0-th hydroelastic contact patch
F_Ac_W(): spatial force (on body A, at centroid of contact surface, in World frame) = \
SpatialForce(
  tau=[2.6063725371243063e-12, -4.630148582002595e-12, 1.5155123087971682e-20],
  f=[-3.6069054088718166e-11, -2.231427822929886e-11, 9.809999999114792],
)
```


The following picture is after the fix. When the white disk stops, we don't see the blue moment vector. This is desirable because, again, the moment vector is practically zero `tau=[2.6063725371243063e-12, -4.630148582002595e-12, 1.5155123087971682e-20].`
```
ContactReporter::Publish() called at time=4.0
num_hydroelastic_contacts() = 1

hydroelastic_contact_info(0): 0-th hydroelastic contact patch
F_Ac_W(): spatial force (on body A, at centroid of contact surface, in World frame) = \
SpatialForce(
  tau=[2.6063725371243063e-12, -4.630148582002595e-12, 1.5155123087971682e-20],
  f=[-3.6069054088718166e-11, -2.231427822929886e-11, 9.809999999114792],
)
```



I'll see how to add a unit test as @jwnimmer-tri suggested.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSIAGSlCfr3kIGw-ZDy:bq3b1b5)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

DamrongGuoy (commented):
I have added the unit test.  @joemasterjohn and @xuchenhan-tri PTAL.

Thanks @jwnimmer-tri .

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSIXDiH0ml7NaXPczv9:b-6ga24e)* status: LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @joemasterjohn and @xuchenhan-tri)

xuchenhan-tri (commented):
platform :lgtm: 

> Reviewed 1 of 1 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSIwD1B5BFeSFvx9om9:bzus18)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

DamrongGuoy (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19112#-:-NSIwr4ZC9M4Bb7q0JMz:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees joemasterjohn,xuchenhan-tri(platform) (waiting on @DamrongGuoy)

PR #19150, opened 2023-04-06
[multibody] Use MatrixBlock to store constraint Jacobians

xuchenhan-tri:
This involves several related changes:

1. Update BlockSparseMatrix to use MatrixBlock
2. SupernodalSolver uses MatrixBlock to leverage sparsity when the matrix stored in MatrixBlock is sparse.
3. SapConstraint uses MatrixBlocks instead of Eigen dense matrices.
4. ContactPairKinematics uses MatrixBlocks instead of Eigen dense matrices.
5. Estimate Delassus operator as J * diag(A)⁻¹ * Jᵀ.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -45,6 +45,7 @@ drake_cc_library(
     srcs = ["block_sparse_matrix.cc"],
     hdrs = ["block_sparse_matrix.h"],
     deps = [
+        ":matrix_block",
         "//common:default_scalars",
         "//common:essential",
         "//common:unused",
@@ -169,6 +170,7 @@ drake_cc_library(
     srcs = ["supernodal_solver.cc"],
     hdrs = ["supernodal_solver.h"],
     interface_deps = [
+        ":matrix_block",
         "//common:essential",
     ],
     deps = [
--- a/multibody/contact_solvers/block_sparse_matrix.cc
+++ b/multibody/contact_solvers/block_sparse_matrix.cc
@@ -48,9 +48,7 @@ void BlockSparseMatrix<T>::Multiply(const Eigen::Ref<const VectorX<T>>& x,
   for (const auto& [ib, jb, Bij] : blocks_) {
     const auto xj = x.segment(col_start_[jb], block_col_size_[jb]);
     auto yi = y->segment(row_start_[ib], block_row_size_[ib]);
-    // N.B. noalias() is necessary to remove Eigen's aliasing assumption and
-    // avoid evaluation into a temporary.
-    yi.noalias() += Bij * xj;
+    Bij.MultiplyAndAddTo(xj, &yi);
   }
 }
 
@@ -64,9 +62,7 @@ void BlockSparseMatrix<T>::MultiplyByTranspose(
   for (const auto& [ib, jb, Bij] : blocks_) {
     const auto xi = x.segment(row_start_[ib], block_row_size_[ib]);
     auto yj = y->segment(col_start_[jb], block_col_size_[jb]);
-    // N.B. noalias() is necessary to remove Eigen's aliasing assumption and
-    // avoid evaluation into a temporary.
-    yj.noalias() += Bij.transpose() * xi;
+    Bij.TransposeAndMultiplyAndAddTo(xi, &yj);
   }
 }
 
@@ -79,14 +75,18 @@ MatrixX<T> BlockSparseMatrix<T>::MakeDenseMatrix() const {
     const int rows = block_row_size_[ib];
     const int js = col_start_[jb];
     const int cols = block_col_size_[jb];
-    A.block(is, js, rows, cols) = Bij;
+    A.block(is, js, rows, cols) = Bij.MakeDenseMatrix();
   }
   return A;
 }
 
 template <typename T>
-void BlockSparseMatrixBuilder<T>::PushBlock(int i, int j,
-                                            const MatrixX<T>& Bij) {
+void BlockSparseMatrixBuilder<T>::PushBlock(int i, int j, MatrixX<T> Bij) {
+  PushBlock(i, j, MatrixBlock<T>(std::move(Bij)));
+}
+
+template <typename T>
+void BlockSparseMatrixBuilder<T>::PushBlock(int i, int j, MatrixBlock<T> Bij) {
   if (blocks_.size() == blocks_.capacity()) {
     throw std::runtime_error(
         "Exceeded the maximum number of non-zero blocks capacity specified at "
@@ -110,9 +110,9 @@ void BlockSparseMatrixBuilder<T>::PushBlock(int i, int j,
       throw std::runtime_error(
           fmt::format("Block ({}, {}) already added.", i, j));
     }
-    blocks_.push_back({i, j, Bij});
     block_row_size_[i] = Bij.rows();
     block_col_size_[j] = Bij.cols();
+    blocks_.emplace_back(i, j, std::move(Bij));
   }
 }
 
--- a/multibody/contact_solvers/block_sparse_matrix.h
+++ b/multibody/contact_solvers/block_sparse_matrix.h
@@ -8,6 +8,7 @@
 #include "drake/common/default_scalars.h"
 #include "drake/common/drake_copyable.h"
 #include "drake/common/eigen_types.h"
+#include "drake/multibody/contact_solvers/matrix_block.h"
 
 namespace drake {
 namespace multibody {
@@ -18,11 +19,12 @@ template <typename T>
 class BlockSparseMatrixBuilder;
 
 // This class provides a representation for sparse matrices with a structure
-// consisting of dense blocks of non-zeros. While other storage formats such as
-// CRS (Compressed Row Storage) are popular (E.g. Eigen::SparseMatrix), a data
-// structure tailored to block-sparse matrices enables efficient algorithms
-// capable of exploiting highly optimized operations with dense blocks (e.g. via
-// AVX instructions).
+// consisting of MatrixBlocks, which are either dense blocks or blocks with
+// specific properties such as sparse, diagonal, or identity. (See
+// MatrixBlock). While other storage formats such as CRS (Compressed Row
+// Storage) are popular (E.g. Eigen::SparseMatrix), a data structure tailored to
+// block-sparse matrices enables efficient algorithms capable of exploiting
+// highly optimized operations with individual blocks.
 //
 // Instances of this class are meant to be built with BlockSparseMatrixBuilder
 // to ensure the consistency of block entries provided by users.
@@ -44,7 +46,7 @@ class BlockSparseMatrix {
   // A non-zero block entry is specified with the triplet {i, j, Bij}, where
   // (i,j) are the i-th and j-th block row and column respectively and Bij is
   // the dense block entry.
-  typedef std::tuple<int, int, MatrixX<T>> BlockTriplet;
+  typedef std::tuple<int, int, MatrixBlock<T>> BlockTriplet;
 
   // Constructs a zero sized matrix.
   // While non-empty matrices must be built with BlockSparseMatrixBuilder to
@@ -69,7 +71,7 @@ class BlockSparseMatrix {
   // Access to the b-th block. b must be in the range 0 to num_blocks()-1.
   // Blocks are indexed in the order they were added using a builder via
   // BlockSparseMatrixBuilder::PushBlock().
-  const MatrixX<T>& get_block(int b) const {
+  const MatrixBlock<T>& get_block(int b) const {
     DRAKE_DEMAND(b < num_blocks());
     return std::get<2>(blocks_[b]);
   }
@@ -197,7 +199,8 @@ class BlockSparseMatrixBuilder {
   // added block to column j or an exception is thrown.
   // @note Blocks of size zero are ignored.
   // @throws if block (i,j) was already added.
-  void PushBlock(int i, int j, const MatrixX<T>& Bij);
+  void PushBlock(int i, int j, MatrixX<T> Bij);
+  void PushBlock(int i, int j, MatrixBlock<T> Bij);
 
   // Makes a new BlockSparseMatrix.
   // If successful, the new BlockSparseMatrix is guaranteed to be properly
--- a/multibody/contact_solvers/sap/BUILD.bazel
+++ b/multibody/contact_solvers/sap/BUILD.bazel
@@ -55,6 +55,7 @@ drake_cc_library(
     deps = [
         "//common:default_scalars",
         "//common:essential",
+        "//multibody/contact_solvers:matrix_block",
     ],
 )
 
--- a/multibody/contact_solvers/sap/sap_constraint.cc
+++ b/multibody/contact_solvers/sap/sap_constraint.cc
@@ -11,7 +11,7 @@ namespace contact_solvers {
 namespace internal {
 
 template <typename T>
-SapConstraint<T>::SapConstraint(int clique, VectorX<T> g, MatrixX<T> J)
+SapConstraint<T>::SapConstraint(int clique, VectorX<T> g, MatrixBlock<T> J)
     : first_clique_(clique),
       g_(std::move(g)),
       first_clique_jacobian_(std::move(J)) {
@@ -23,8 +23,8 @@ SapConstraint<T>::SapConstraint(int clique, VectorX<T> g, MatrixX<T> J)
 
 template <typename T>
 SapConstraint<T>::SapConstraint(int first_clique, int second_clique,
-                                VectorX<T> g, MatrixX<T> J_first_clique,
-                                MatrixX<T> J_second_clique)
+                                VectorX<T> g, MatrixBlock<T> J_first_clique,
+                                MatrixBlock<T> J_second_clique)
     : first_clique_(first_clique),
       second_clique_(second_clique),
       g_(std::move(g)),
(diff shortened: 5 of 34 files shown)


xuchenhan-tri (commented):
-(status: do not review) +@amcastro-tri for feature review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSNfUlY1-MkgQgh6U3f:bgvwxyh)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri)

amcastro-tri (commented):
I love how everything fits together in terms of MatrixBlock.
Thus far only a few questions and asking for documentation in a couple palces.

> Reviewed 20 of 34 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSgP3ylA0OB0lXXbbND:b-f6hs1u)* status: 6 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_matrix.h` line 23 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgNDu36uZIZvYpYfa9:-NSgNDu36uZIZvYpYfaA:b-12mzco) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/block_sparse_matrix.h#L23)):*
> ```objc
> // This class provides a representation for sparse matrices with a structure
> // consisting of MatrixBlocks, which are either dense blocks or sparse matrices
> // composing of dense blocks of non-zeros themselves (see MatrixBlock). While
> ```

will this always be true and should it be part of this docs? I mean, what if we add a third variant to MatrixBlock, wouldn't this go away?
Maybe be more generic like: "... with a structure consisting of blocks represented as a MatrixBlock".
Maybe expand to: " Typically dense blocks though in general blocks with specific properties (such as sparse, diagonal, or identity matrices).
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 23 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcDOs6RSYrNKyagqO:-NSgcDOt301_UqKFMGtV:bnrx71k) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L23)):*
> ```cpp
> template <typename MatrixType>
> void ComputeWeightMatrixTerms(
>     const vector<MatrixBlock<double>>& jacobian_row_data,
> ```

could we add some minimum docs to this method?
To include:
 - expected format and size of jacobian_row_data. Are there any ordering assumptions?
 - expected format and size of weight_matrix. 
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 27 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcdrh0_ec3XgS1Qix:-NSgcdri6To06fBS6H_t:b-atd9sm) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L27)):*
> ```cpp
>     MatrixType* yptr) {
>   std::vector<MatrixBlock<double>> GJs;
>   for (size_t j = 0; j < jacobian_row_data.size(); ++j) {
> ```

not sure if I understand the format.
Since you are rewriting this anway, consider using better notation.
For instance, i'd reserve `i` and `j` for row column indexes and maybe use something like `k` for entries in a linear array.
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 29 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcu0e2iyc0djhyExf:-NSgcu0f9VnKt7Z-oqQq:b-u6sodd) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L29)):*
> ```cpp
>   for (size_t j = 0; j < jacobian_row_data.size(); ++j) {
>     const MatrixBlock<double>& J = jacobian_row_data[j];
>     GJs.emplace_back(
> ```

To help future developers working in this file, document the format of the expected result of this computation.
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 236 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgTm1e5zPVoFdlj3uX:-NSgTm1e5zPVoFdlj3uY:b1hmcor) ([raw file](https://github.com/RobotLocomotion/drake/blob/c7c002a876fcce54441cf820b2a05c9fc4a98ed5/multibody/contact_solvers/supernodal_solver.cc#L236)):*
> ```cpp
> 
>   Compute_Ji_transpose_Gi_Ji(jacobian_row_data_, *weight_matrix_, weight_start_,
>                              weight_end_, &submatrix_data_.G, &G_times_J_);
> ```

Just to make sure, it looks like we were not using `G_times_J_` before? that's why you remove it?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 86 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgdFBR4MlTZ83JVak1:-NSgdFBR4MlTZ83JVak2:bf5h0dz) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/sap/sap_constraint.h#L86)):*
> ```objc
>    clique. that takes a dense Jacobian. */
>   SapConstraint(int clique, VectorX<T> g, MatrixX<T> J)
>       : SapConstraint(clique, std::move(g), MatrixBlock<T>(std::move(J))) {}
> ```

do we need the alternative signatures taking dense blocks?

amcastro-tri (commented):
> Reviewed 12 of 34 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: 6 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
Comments addressed. PTAL, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSgqLZNBNkl783hg4vC:b-pr0h6x)* status: 6 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri)

___
*[`multibody/contact_solvers/block_sparse_matrix.h` line 23 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgNDu36uZIZvYpYfa9:-NSgqpBW9NW2E8lB4FHF:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/block_sparse_matrix.h#L23)):*

will this always be true and should it be part of this docs? I mean, what if we add a third variant to MatrixBlock, wouldn't this go away?
Maybe be more generic like: "... with a structure consisting of blocks represented as a MatrixBlock".
Maybe expand to: " Typically dense blocks though in general blocks with specific properties (such as sparse, diagonal, or identity matrices).
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 23 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcDOs6RSYrNKyagqO:-NSgtiTO7OymcOECVmTd:bevftus) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L23)):*

could we add some minimum docs to this method?
To include:
 - expected format and size of jacobian_row_data. Are there any ordering assumptions?
 - expected format and size of weight_matrix. 


</blockquote></details>

Done, PTAL.
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 27 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcdrh0_ec3XgS1Qix:-NSgwIALCgnJKnTV-Z3a:b-bz40b2) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L27)):*

not sure if I understand the format.
Since you are rewriting this anway, consider using better notation.
For instance, i'd reserve `i` and `j` for row column indexes and maybe use something like `k` for entries in a linear array.
</blockquote></details>

I changed up the indices. PTAL.
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 29 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgcu0e2iyc0djhyExf:-NSgwQGL9n2XTRria8RW:bf0z58k) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/supernodal_solver.cc#L29)):*

To help future developers working in this file, document the format of the expected result of this computation.
</blockquote></details>

I added some doc for the function. Does that help?
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 236 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgTm1e5zPVoFdlj3uX:-NSgwdXoBnrdpg_y-drz:bluswy6) ([raw file](https://github.com/RobotLocomotion/drake/blob/c7c002a876fcce54441cf820b2a05c9fc4a98ed5/multibody/contact_solvers/supernodal_solver.cc#L236)):*

Just to make sure, it looks like we were not using `G_times_J_` before? that's why you remove it?
</blockquote></details>

I believe G_times_J_ was used as a scratch pad to store intermediate values of the calculation `J^T * G * J`. I changed that to a local variable `GJs` instead because I don't think it is going to prevent heap allocation with MatrixBlock. See if you agree.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 86 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgdFBR4MlTZ83JVak1:-NSgoOJn31oBJ13MmMkq:b9m5yin) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/sap/sap_constraint.h#L86)):*

do we need the alternative signatures taking dense blocks?
</blockquote></details>

It's not strictly required, but it makes transitioning to `MatrixBlock` easier. Otherwise, I'd need to spam the `MatrixBlock` keyword all over our existing unit tests.

xuchenhan-tri (commented):
See https://github.com/RobotLocomotion/drake/pull/19169 for the CI failure. I will rerun CI after (potentially) addressing the next round of reviews.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NShKWJE0dHO5551aGka:b5s6yah)* status: 6 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri)

rpoyner-tri (commented):
:lgtm:

> Reviewed 30 of 34 files at r1, 2 of 2 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSl_2Ne8rD0jqA9z3oA:bnfp4nl)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri and @frankpermenter)

amcastro-tri (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/supernodal_solver.cc` line 37 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSkiDN4-gHdfzTav9mI:-NSmCAVVDnh6pMSZ9uO2:b-akmw33) ([raw file](https://github.com/RobotLocomotion/drake/blob/b9373ffa536d773912cdeda1c9800ed04e4f07d8/multibody/contact_solvers/supernodal_solver.cc#L37)):*

The data in `jacobian_row_data` is indeed a (compressed) row of the jacobian matrix. I expanded on the layout of J in the function doc.

> ..., however in this G*J below it'd seem to indicate it is a column instead?

Can you be specific about which lines you are referring to here?
</blockquote></details>

I was referring to this very line, 37 in r3 when I made the comment. 
I believe I was able to undertanding now. As you say here, `jacobian_row_data[i]` corresponds to the `i-th` block row of J (not sure why you say compressed though. When the MatrixBlock is a dense matrix it'd simply resolve to the entire dense row?).
___
*[`multibody/contact_solvers/supernodal_solver.cc` line 23 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSmBg63333SATeWOcxr:-NSmBg64EGjFxXEKhCCL:b6pualx) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d34a4759c0baaa38959e463efb1f26c9c5b583e/multibody/contact_solvers/supernodal_solver.cc#L23)):*
> ```cpp
>  entries in `jacobian_row_data`. More specifically, the layout of J looks like
> 
>  [ jacobian_row_data[0], jacobian_row_data[1], ..., jacobian_row_data.back()]
> ```

In Matlab notation for instance, this would correspond to column concatenation. Bottom line, without a convention, this notation might only add confusion.
Reading the code below, now I understand that `jacobian_row_data[i]` corresponds to the `i-th` block row of J.
If correct, why not to say "Where jacobian_row_data[i] stores the i-th block row of the Jacobian J".
If incorrect, then I am still misunderstanding the docs.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 86 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSgdFBR4MlTZ83JVak1:-NSmIafYDrlUd1iWDhKZ:b8rgpxj) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c740461812f8d8048d9792eddbbccc49e26d513/multibody/contact_solvers/sap/sap_constraint.h#L86)):*
> then calling print on a MatrixX would cause unexpected copying.

I see your point, definitely bad to have this unexpected copying. I believe that's precisely why we enforce the use of "explicit" in Drake.

xuchenhan-tri (commented):
I launched an Anzu CI against this PR to guard against the unlikely case that the change in the Delassus esitmation implementation has an effect on Anzu CI.
+(status: do not merge) for now.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSmMGmf4U7fbVNFonxt:b-orenb2)* status: labeled "do not merge" (waiting on @xuchenhan-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: labeled "do not merge" (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
Ok anzu ci is fine -(status: do not merge) 
@amcastro-tri Please make another pass at the semantics of the J^T*G*J function and merge if you think the comments are clear.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-:-NSmUf_U0CBr9JjuZucj:b-ver0ou)* status: labeled "do not merge" (waiting on @xuchenhan-tri)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),amcastro-tri (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),amcastro-tri (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/supernodal_solver.cc` line 23 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19150#-NSmBg63333SATeWOcxr:-NSqJ4ox4zFvB2qEOwqz:b-yg9gdr) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d34a4759c0baaa38959e463efb1f26c9c5b583e/multibody/contact_solvers/supernodal_solver.cc#L23)):*
> This point is not obvious without additional context. 

Agreed. I think the context of jacobians and weight matrices is hurting understanding instead of helping it. I changed the name of the parameters and the documentation to intentional weaken the link to the physical meaning of the matrices and focus more on the algebraic operations themselves.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19150)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),amcastro-tri (waiting on @xuchenhan-tri)

PR #19145, opened 2023-04-06
[doc] Clarify stability policy for struct aggregate initialization

jwnimmer-tri:
See https://github.com/RobotLocomotion/drake/pull/19105 and https://github.com/RobotLocomotion/drake/pull/19144.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19145)

Issue #18055, opened 2022-10-07
Mujoco XML parsing should use the diagnostic policy mechanism for errors and warnings


rpoyner-tri:
**Is your feature request related to a problem? Please describe.**
The mujoco parser does not yet use the diagnostic policy mechanism: errors are all hardcoded throw and warnings are all hardcoded to text logging.

**Describe the solution you'd like**
Use the diagnostic policy mechanism to issue Errors and Warnings. See the examples of URDF and SDFormat parsers for detail. Also, add testing to verify that making errors non-fatal will not crash the parser.

**Describe alternatives you've considered**
There are no alternatives, but limping along with the existing implementation is not strictly wrong.

**Additional context**
See also similar complaint about model directives at #18052.

rpoyner-tri:
There is subsequent work to do, The mujoco parser should provide some warning that the result of parsing w.r.t. collision filters may be surprising. See https://github.com/RobotLocomotion/drake/pull/18351#pullrequestreview-1294881692 for a fuller discussion.

rpoyner-tri:
A second piece of follow-up work is to retire the peculiar added MbP method GetMutableSceneGraphPreFinalize(), introduced in #18351.

RussTedrake:
I've addressed the "mujoco parser should provide some warning" in #18351.

rpoyner-tri:
https://github.com/RobotLocomotion/drake/pull/19157 addresses most of it, leaving a few delegated throws where MbP has nicer messages. I left a TODO for those, but I think the bulk of this issue is done and can close.

PR #19157, opened 2023-04-07
[parsing] Port mujoco parser to diagnostic policy

rpoyner-tri:
This patch updates most of the mujoco parser diagnostics to include file and line number, and to obey policy configuration from the top-level Parser.

In addition, it fixes a bug found trying to use std::optional<bool> for a tri-state variable -- one dropped '*' in the wrong place led to an unreachable branch, and probably other problems.

Finally, it moves the affected files closer to clang-format compliance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157)

### DIFF
--- a/multibody/parsing/BUILD.bazel
+++ b/multibody/parsing/BUILD.bazel
@@ -573,6 +573,7 @@ drake_cc_googletest(
     ] + _DM_CONTROL_MUJOCO_FILES,
     deps = [
         ":detail_mujoco_parser",
+        ":diagnostic_policy_test_base",
         "//common:find_resource",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
--- a/multibody/parsing/detail_mujoco_parser.cc
+++ b/multibody/parsing/detail_mujoco_parser.cc
@@ -21,6 +21,7 @@
 #include "drake/math/rotation_matrix.h"
 #include "drake/multibody/parsing/detail_make_model_name.h"
 #include "drake/multibody/parsing/detail_tinyxml.h"
+#include "drake/multibody/parsing/detail_tinyxml2_diagnostic.h"
 #include "drake/multibody/tree/ball_rpy_joint.h"
 #include "drake/multibody/tree/prismatic_joint.h"
 #include "drake/multibody/tree/revolute_joint.h"
@@ -35,44 +36,24 @@ using Eigen::Matrix3d;
 using Eigen::Vector2d;
 using Eigen::Vector3d;
 using Eigen::Vector4d;
-using geometry::GeometryInstance;
 using geometry::GeometryId;
+using geometry::GeometryInstance;
 using math::RigidTransformd;
 using math::RotationMatrixd;
 using tinyxml2::XMLAttribute;
 using tinyxml2::XMLDocument;
 using tinyxml2::XMLElement;
+using tinyxml2::XMLNode;
 
 namespace {
 
+// TODO(rpoyner-tri): replace with std::string_view::ends_with once c++20
+// features are available.
 bool EndsWith(std::string_view value, std::string_view ending) {
-  if (ending.size() > value.size()) { return false; }
+  if (ending.size() > value.size()) {
+    return false;
+  }
   return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
-}
-
-// For tags that exist in the MuJoCo XML schema, but are not (yet) supported
-// here, we print a warning to the console.  This method checks the node for
-// the child element, and prints the warning if any occurrences exist.
-void WarnUnsupportedElement(const XMLElement& node, const std::string& tag) {
-  if (node.FirstChildElement(tag.c_str())) {
-    drake::log()->warn(
-        "The tag '{}' found as a child of '{}' is currently unsupported and "
-        "will be ignored.",
-        tag, node.Name());
-  }
-}
-
-// For tags that exist in the MuJoCo XML schema, but are not (yet) supported
-// here, we print a warning to the console.  This method checks the node for
-// the child element, and prints the warning if any occurrences exist.
-void WarnUnsupportedAttribute(const XMLElement& node,
-                              const std::string& attribute) {
-  if (node.Attribute(attribute.c_str())) {
-    drake::log()->warn(
-        "The attribute '{}' found in a {} tag is currently unsupported and "
-        "will be ignored.",
-        attribute, node.Name());
-  }
 }
 
 // Any attributes from `default` that are not specified in `node` will be added
@@ -88,8 +69,11 @@ void ApplyDefaultAttributes(const XMLElement& default_node, XMLElement* node) {
 
 class MujocoParser {
  public:
-  explicit MujocoParser(const ParsingWorkspace& workspace)
-      : workspace_(workspace), plant_(workspace.plant) {
+  explicit MujocoParser(const ParsingWorkspace& workspace,
+                        const DataSource& data_source)
+      : workspace_(workspace),
+        diagnostic_(&workspace.diagnostic, &data_source),
+        plant_(workspace.plant) {
     // Clang complains that the workspace_ field is unused. Nerf the warning
     // for now; it will be used soon.
     unused(workspace_);
@@ -102,17 +86,20 @@ class MujocoParser {
 
     // Check that only one of the orientation variants are supplied:
     const int num_orientation_attrs =
-        (node->Attribute("quat") != nullptr ? 1 : 0)
-        + (node->Attribute("axisangle") != nullptr ? 1 : 0)
-        + (node->Attribute("euler") != nullptr ? 1 : 0)
-        + (node->Attribute("xyaxes") != nullptr ? 1 : 0)
-        + (node->Attribute("zaxis") != nullptr ? 1 : 0);
+        (node->Attribute("quat") != nullptr ? 1 : 0) +
+        (node->Attribute("axisangle") != nullptr ? 1 : 0) +
+        (node->Attribute("euler") != nullptr ? 1 : 0) +
+        (node->Attribute("xyaxes") != nullptr ? 1 : 0) +
+        (node->Attribute("zaxis") != nullptr ? 1 : 0);
     if (num_orientation_attrs > 1) {
-      throw std::logic_error(fmt::format(
-          "Element {} has more than one orientation attribute specified "
-          "(perhaps through defaults). There must be no more than one instance "
-          "of `quat`, `axisangle`, `euler`, `xyaxes`, or `zaxis`.",
-          node->Name()));
+      Error(
+          *node,
+          fmt::format(
+              "Element {} has more than one orientation attribute specified "
+              "(perhaps through defaults). There must be no more than one "
+              "instance of `quat`, `axisangle`, `euler`, `xyaxes`, or `zaxis`.",
+              node->Name()));
+      return {};
     }
 
     Vector4d quat;  // MuJoCo uses w,x,y,z order.
@@ -169,11 +156,11 @@ class MujocoParser {
 
     std::string joint_name;
     if (!ParseStringAttribute(node, "joint", &joint_name)) {
-      drake::log()->warn(
-          "The motor '{}' does not use the 'joint' transmission specification. "
-          "Currently only the 'joint' attribute is supported.  This motor will "
-          "be ignored.",
-          name);
+      Warning(*node, fmt::format(
+                         "The motor '{}' does not use the 'joint' transmission "
+                         "specification. Currently only the 'joint' attribute "
+                         "is supported.  This motor will be ignored.",
+                         name));
       return;
     }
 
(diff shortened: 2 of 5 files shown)


rpoyner-tri (commented):
+@RussTedrake or delegate for feature review.
+(release notes: fix) +(release notes: feature) (I think both, happy to quibble about release notes)

The one set of messages I left as "throw" are ones that have a really nice message from MultibodyPlant. Perhaps a future pass can consider whether to convert those to diagnostic policy somehow.

See issue #18055 for context.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-:-NSRg32-0M2djGrvnyK8:b-a2efvp)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers (waiting on @RussTedrake)

RussTedrake (commented):
:lgtm:  Thanks for doing it!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-:-NSV9ZIn7ZKwO62Q521r:bv5boa8)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1084 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSV8TZlD8JXsPHBkN0s:-NSV8TZlD8JXsPHBkN0t:b-v77wkf) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1084)):*
> ```cpp
>               fmt::format(
>                   "Compiler attribute coordinate={} is not supported yet.",
>                   coordinate));
> ```

btw -- in other calls to Error you manually return.  Is this one intentionally different?

RussTedrake (commented):
+@ggould-tri for platform review, per rotation.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-:-NSV9fAe7d4BaUwq7b5P:b-9wrk23)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @RussTedrake)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1084 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSV8TZlD8JXsPHBkN0s:-NSc1ajC24orU9Vvj9qW:b-wuh1sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1084)):*

btw -- in other calls to Error you manually return.  Is this one intentionally different?
</blockquote></details>

This was intentional on my part. Certainly, exercising the post-error paths will be rare (tests only), unless we change the policy options are the Parser level; public API users will still get an exception. My main reason for using explicit `return` or `continue` in post-error paths is to avoid later outright crashes (null pointer dereference or similar). In this case, the `compiler` attributes are mostly just setting variables for later use. Certainly any subsequent model would be wrong, so the Error() is justified, but the post-error path (AFAICT) would not result in an outright crash.

It does make the testing a bit more convenient, but if you think a `return` here would be better, we could go that route and juggle the test a bit.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1084 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSV8TZlD8JXsPHBkN0s:-NSc9lVNBBVGDP19erxL:b-zg9z4w) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1084)):*

This was intentional on my part. Certainly, exercising the post-error paths will be rare (tests only), unless we change the policy options are the Parser level; public API users will still get an exception. My main reason for using explicit `return` or `continue` in post-error paths is to avoid later outright crashes (null pointer dereference or similar). In this case, the `compiler` attributes are mostly just setting variables for later use. Certainly any subsequent model would be wrong, so the Error() is justified, but the post-error path (AFAICT) would not result in an outright crash.

It does make the testing a bit more convenient, but if you think a `return` here would be better, we could go that route and juggle the test a bit.
</blockquote></details>

I don't feel strongly that you must return.  Perhaps a comment would be enough to help the reader know that it's intentional.

ggould-tri (commented):
:lgtm:

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-:-NSfj3OAEZC-ytmnF2JW:bnfp4nl)* status: 6 unresolved discussions (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfdtuZ3rhnaV_ZvQUZ:-NSfdtuZ3rhnaV_ZvQU_:bjb7th0) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L50)):*
> ```cpp
> namespace {
> 
> bool EndsWith(std::string_view value, std::string_view ending) {
> ```

minor:  Consider TODO'ing this on C++20 support (when we drop Focal, we get `std::string_view::ends_with`, if in fact we don't have it already).
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 973 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfevCbCiQ9TnVAg3io:-NSfevCbCiQ9TnVAg3ip:b7u5urd) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L973)):*
> ```cpp
>         }
>         // The file not existing at this stage is not an error. Absorb errors
>         // from canonical().
> ```

minor:  I think that's equivalent to `std::filesystem::weakly_canonical`, unless you care strongly about discarding structural errors like `/..`?



_Code quote:_
```cpp
        // The file not existing at this stage is not an error. Absorb errors
        // from canonical().
```
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 987 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSffaAa6os_PcZZZmbC:-NSffaAa6os_PcZZZmbD:b-fi2jyo) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L987)):*
> ```cpp
>                        [](unsigned char c) {
>                          return std::tolower(c);
>                        });        if (extension != ".obj") {
> ```

minor:  This is incorrect formatting.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSffyzADav_w8jVxNXM:-NSffyzADav_w8jVxNXN:b-ykbanl) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1162)):*
> ```cpp
> 
>     auto geom_id_from_name = [&inspector, &geom_ids,
>                               frame_group](const std::string& name) {
> ```

minor:  This is a curious indentation choice, breaking the captures and putting the remaining captures and args together.  Maybe it would look clearer regrouped?



_Suggestion:_
```cpp
    auto geom_id_from_name =
        [&inspector, &geom_ids, frame_group](
             const std::string& name) {
```
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfgP0b5yEtlschWqil:-NSfgP0b5yEtlschWqim:b-itoay0) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1396)):*
> ```cpp
>   Angle angle_{kDegree};
>   std::map<std::string, XMLElement*> default_geometry_{};
>   enum InertiaFromGeometry { kFalse, kTrue, kAuto };
> ```

BTW:  Thanks.  Also I'm delighted this is in the same PR as some file-not-found logic :-)
___
*[`multibody/parsing/test/detail_mujoco_parser_test.cc` line 73 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfiKzN4aAjayOxDd3K:-NSfiKzN4aAjayOxDd3L:bmhrgko) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/test/detail_mujoco_parser_test.cc#L73)):*
> ```cpp
>     internal::CollisionFilterGroupResolver resolver{&plant_};
>     ParsingWorkspace w{options_, package_map_, diagnostic_policy_,
>                        &plant_, &resolver, NoSelect};
> ```

minor, author's discretion:  In the spirit of #19145, it might be better if these somewhat opaque struct stuffing blocks used field names.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157)* status: 1 unresolved discussion (waiting on @ggould-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 987 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSffaAa6os_PcZZZmbC:-NSfvqnD1GA4Fg9rRf-7:bxicadz) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L987)):*

minor:  This is incorrect formatting.
</blockquote></details>

Good catch. Weird that survives lints, though.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSffyzADav_w8jVxNXM:-NSfw-o32-YEar_mfoYc:b-z4jtt2) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1162)):*

minor:  This is a curious indentation choice, breaking the captures and putting the remaining captures and args together.  Maybe it would look clearer regrouped?
</blockquote></details>

IIRC, this is what `clang-format` recommended. I'm inclined to take its advice here, since I intend to turn on clang-format by default for parsing/ soon-ish.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfgP0b5yEtlschWqil:-NSfxGTu0CxkAF4Cjodm:b-xcppzr) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1396)):*

BTW:  Thanks.  Also I'm delighted this is in the same PR as some file-not-found logic :-)
</blockquote></details>

I was somewhat astounded by the subtlety of the bug induced by optional<bool>. Maybe we should consider style guide advice?
___
*[`multibody/parsing/test/detail_mujoco_parser_test.cc` line 73 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfiKzN4aAjayOxDd3K:-NSfx_vnCCZrntmOELlw:b-31f0yx) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/test/detail_mujoco_parser_test.cc#L73)):*

minor, author's discretion:  In the spirit of #19145, it might be better if these somewhat opaque struct stuffing blocks used field names.
</blockquote></details>

I'm inclined to punt on this. The struct in question will never be part of public API, and there are likely dozens of these initializations that exist.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19157)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),RussTedrake(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 987 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSffaAa6os_PcZZZmbC:-NSkFLP53vVMyEEQuUCe:b-ilio75) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L987)):*

Good catch. Weird that survives lints, though.
</blockquote></details>

It is everything I expect of a linter.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19157#-NSfgP0b5yEtlschWqil:-NSkFUP1BqfTFStTsKeb:b-3g83x7) ([raw file](https://github.com/RobotLocomotion/drake/blob/915d136fb2813400c70026db828957a3e5338761/multibody/parsing/detail_mujoco_parser.cc#L1396)):*

I was somewhat astounded by the subtlety of the bug induced by optional<bool>. Maybe we should consider style guide advice?
</blockquote></details>

Yeah, optional's choice of "helpful" auto-unboxing and pointer-like representation is kind of the opposite of where modern languages went with this feature, for exactly these sorts of reasons.  I would support an outright ban on `optional<bool>` in favor of explicit enums if there weren't so many of them around... maybe a project for a future time.

PR #19168, opened 2023-04-10
Use vtkOBJImporter to load an mtl file

zachfang:
A current working prototype to use `vtkOBJImporter` to load an MTL file associated with an OBJ file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)


zachfang:
The main changes are:
1. Replace `vtkOBJReader` with `vtkOBJImporter` with a built-in MTL loading from the VTK library.
2. `vtkOBJImporter` creates textured `vtkActor` with a connected mapper. We then set the desired properties for the actors or the mappers.
3. As the prior points, `ImplementObj()` now has a different workflow from `ImplementGeometry(vtkPolyDataAlgorithm* source, ...)` which is shared by other primitives. My proposal is to treat `ImplementObj()` differently and factor on the common actor/mapper operations out to separate functions for reuse.
4. If there is no associated MTL file according to the importer, we then fall back to the original texture loading routine. Thus this follows the MTL-win discussion.

A quick example for rendering the `box.obj` with a non-uniform texture.
color_000

zachfang (commented):
CC\ @jwnimmer-tri @SeanCurtis-TRI for some high-level reviews. I will work on adding tests and nailing down edge cases next.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-:-NSgv1zCDOyBcScpe_ie:b-4hud21)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgu_g0B9xvYWCtORap:-NSgu_g0B9xvYWCtORaq:b-2eqtss) (no related file):*
Working. Currently, the code can render color, depth, and label images. However, there is a segfault in the end, which I think it's due to the VTK pointers. Investigating.

___
___
*[`geometry/render/test/meshes/box.obj.mtl` line 10 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgtddl-LpzeSjct6ye:-NSgtddl-LpzeSjct6yf:bm3trgg) ([raw file](https://github.com/RobotLocomotion/drake/blob/c15d53d8ffbeecf2fb0858dc2b871e6733f0dbb4/geometry/render/test/meshes/box.obj.mtl#L10)):*
> ```
> # N.B. The next line has trailing whitespace on purpose, to prove that Drake
> # ignores it during parsing.
> map_Kd -s 1 1 1 box.png 
> ```

Working. Not sure what Drake parsing indicates here, the VTK importer couldn't parse the PNG with trailing spaces and thus I removed it for now during my testing.



_Code quote:_
```
map_Kd -s 1 1 1 box.png·
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgu4Kh8JWbnTUOiDdP:-NSgu4Kh8JWbnTUOiDdQ:ba98t9a) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.cc#L102)):*
> ```cpp
> }
> 
> void SetColorActorTexture(const vtkSmartPointer<vtkActor>& color_actor,
> ```

FYI, this is the original code without changes (except I need to pass ` default_diffuse`).



_Code quote:_
```cpp
void SetColorActorTexture(const vtkSmartPointer<vtkActor>& color_actor,
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 532 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSguR7E92RukQ24c4On:-NSguR7E92RukQ24c4Oo:bxgxv5h) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.cc#L532)):*
> ```cpp
>   const std::string file_directory =
>       std::filesystem::path{file_name}.parent_path();
>   // Create three actors for color, depth, and label pipelines.
> ```

FYI. From what I read, there isn't a deep copy method for `vtkActor`. 



_Code quote:_
```cpp
// Create three actors for color, depth, and label pipelines.
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 541 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSguFES1mlpCZZfvL5i:-NSguFES1mlpCZZfvL5j:b-88422x) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.cc#L541)):*
> ```cpp
> 
>   vtkActorCollection* actor_collection = vtk_renderer->GetActors();
>   // TODO(zachfang): Figure out why there are two actors from each import.
> ```

Working. I don't know why there are two actors created for each import.



_Code quote:_
```cpp
// TODO(zachfang): Figure out why there are two actors from each import.
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`geometry/render/test/meshes/box.obj.mtl` line 10 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgtddl-LpzeSjct6ye:-NSgwVJ-0d8AYg67MYpx:b-twt3j5) ([raw file](https://github.com/RobotLocomotion/drake/blob/c15d53d8ffbeecf2fb0858dc2b871e6733f0dbb4/geometry/render/test/meshes/box.obj.mtl#L10)):*

Working. Not sure what Drake parsing indicates here, the VTK importer couldn't parse the PNG with trailing spaces and thus I removed it for now during my testing.
</blockquote></details>

In the wild, some mtl files (e.g., the YCB dataset) have trailing spaces. If VTK isn't correctly ignoring those, then we'll need to either patch VTK to fix it or find a work-around (like reading the MTL from a string which we've cleaned up).

I'm not sure whether that bug needs to block this pull request, but in any case we need ensure that this _test file_ (which is used elsewhere) keeps the trailing space.  If we need a separate `vtk_box.obj` for VTK testing modulo the VTK parsing bug, then so be it.

SeanCurtis-TRI (commented):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkx1bQ8IzUCtmg6xbX:-NSmALv_6lwb4ximcqZn:b9kvks9) (no related file):*

Now that @SeanCurtis-TRI has more time available, I'll suggest deferring to his readout for this topic, and not reading too much into my intuition.
</blockquote></details>

I started an initial review, but stopped at what I suspect is a correctness issue due to vtk. I'll build something off of this branch to confirm.

___
___
*[`geometry/render/test/meshes/4_color_texture.png` line  at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSktCJbA2To8ToYoolZ:-NSm-zLT8m9rp2410pGz:b-8p6dj) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render/test/meshes/4_color_texture.png#L)):*

nit Is there a reason we need to use a new file?

My intuition is that we should replace the existing `box.png` with this non-uniform texture, so that _all_ geometry tests are properly sensitive to texture coordinates (which is what I assume you were going for here?).
</blockquote></details>

Particularly as this is a copy of this png located in a different directory.

As for making *this* particular image "box.png", it's worth noting that various tests will fail. There are tests that coarsely test for the existence of textured objects (but don't want to worry about the details). As such, the solid green renders pixels that can be sampled as evidence of texture. When the texture varies over its surface, we have to be more precise in what we look for (making the tests more brittle).

I think it might also modify the `meshcat_manul_test` -- I believe, it loads the same box (currently green), and its textural description would no longer be valid.

It's worth noting, none of that is an objection to replacing the existing `box.png`, just a note on the broad impact it might have.
___
*[`geometry/render/test/meshes/box.obj.mtl` line 10 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgtddl-LpzeSjct6ye:-NSm0fS848nW0dYkC8Rv:bvrxub5) ([raw file](https://github.com/RobotLocomotion/drake/blob/c15d53d8ffbeecf2fb0858dc2b871e6733f0dbb4/geometry/render/test/meshes/box.obj.mtl#L10)):*

In the wild, some mtl files (e.g., the YCB dataset) have trailing spaces. If VTK isn't correctly ignoring those, then we'll need to either patch VTK to fix it or find a work-around (like reading the MTL from a string which we've cleaned up).

I'm not sure whether that bug needs to block this pull request, but in any case we need ensure that this _test file_ (which is used elsewhere) keeps the trailing space.  If we need a separate `vtk_box.obj` for VTK testing modulo the VTK parsing bug, then so be it.
</blockquote></details>

The vtk obj importer needs a lot of patching. It has *very* limited support for the mtl file spec ([currently, the only supported texture is `map_Kd`](https://github.com/Kitware/VTK/blob/master/IO/Import/vtkOBJImporterInternals.h#L28). It might be good to plan on putting something in the queue for that.
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 227 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkurEQ92EXajm9iM-1:-NSm1d-zDbBuX2g2Skpd:bxkh4v8) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.h#L227)):*

This scope of what's going on here is pretty confusing.  This function is _modifying its argument_, not the internal state of `this`.

For starters, it should be marked `static` to clarify that it doesn't touch `this`.

Then the comment should be more direct.  Something along the lines of "Modifies depth_actor to use the proper shaders for a depth sensor".
</blockquote></details>

I'd further recommend changing the signature to something like (so it looks more obviously Drake-like in modifying the input parameter):

```c++
static void SetDepthShader(vtkActor* depth_actor);
```

Then, at the call sites, you just need to add `.Get()`. I.e.,

```c++
SetDepthShader(actors[ImageType::kDepth].Get());
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 227 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSm2mbVClMq18hsbxbY:-NSm2mbVClMq18hsbxbZ:b-ld7m25) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.h#L227)):*
> ```objc
>   void SetDefaultLightPosition(const Vector3<double>& X_DL) override;
> 
>   // Sets vertex and fragment shaders to the depth_actor's mapper.
> ```

In a related vein, for mutating the color and label actors, you made those standalone functions in an anonymous namespace in the .cc file (passing the `RenderEngine` parameters as necessary. Why not do the same for the depth actor? Why particularly does *this* one have to be a member method?
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSm28Zt4XxuA-MQlxQ6:-NSm28Zt4XxuA-MQlxQ7:bc53d5p) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.cc#L102)):*
> ```cpp
> }
> 
> void SetColorActorTexture(const vtkSmartPointer<vtkActor>& color_actor,
> ```

nit Similarly, if you change the input parameter to `vtkActor* color_actor`, it will be more obvious that you're modifying the actor.

Same for the label actor.
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 533 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSm4pdF1AWWMqQZiZGn:-NSm4pdF1AWWMqQZiZGo:bycxyjw) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.cc#L533)):*
> ```cpp
>       std::filesystem::path{file_name}.parent_path();
>   // Create three actors for color, depth, and label pipelines.
>   for (int i = 0; i < kNumPipelines; ++i) {
> ```

I'm really worried about this code. As I read the vtkImporter, I believe *multiple* actors may be added *per OBJ*. If the file has disparate groups with different materials applied, I believe each group becomes its own actor.  So, we should really be taking *all* of the actors the importer creates.

That does raise the question about why there are two actors for each import for the simple OBJs we're testing against. (Without reading the code, I *suspect* we may be getting the actual mesh and a meaningless default object.) But the one thing we can't do is unilaterally drop actors.

We should create a test obj to exercise this. Let me try to create a test case to illustrate this.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkx1bQ8IzUCtmg6xbX:-NSmBDyx-4bNDGHZmTCY:b56ws80) (no related file):*

I started an initial review, but stopped at what I suspect is a correctness issue due to vtk. I'll build something off of this branch to confirm.
</blockquote></details>

If VTK's MTL support is really so awful, maybe we should just give up on trying to use it?

We've already opened the door to textures coming from _geometry properties_ when the mesh file itself doesn't have any provided texturing.  Therefore, we already need render code to handle those properties.

We could keep the VTK render engine OBJ loader as-is, and just write the code to parse an MTL file into geometry properties textures.

Maybe that's the best path to the goal?

___

SeanCurtis-TRI:

If VTK's MTL support is really so awful, maybe we should just give up on trying to use it?

We've already opened the door to textures coming from _geometry properties_ when the mesh file itself doesn't have any provided texturing.  Therefore, we already need render code to handle those properties.

We could keep the VTK render engine OBJ loader as-is, and just write the code to parse an MTL file into geometry properties textures.

Maybe that's the best path to the goal?
</blockquote></details>

It's looking pretty whiffy to me. But, that said, we still have similar issues. Any OBJ that uses multiple materials isn't supported by us either. So, do we patch the upstream or re-invent the wheel?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkx1bQ8IzUCtmg6xbX:-NSmCj4J9hpjr27XlJM7:bl2hrni) (no related file):*

It's looking pretty whiffy to me. But, that said, we still have similar issues. Any OBJ that uses multiple materials isn't supported by us either. So, do we patch the upstream or re-invent the wheel?
</blockquote></details>

Patching upstream is an incredible pain in the ass.  If the upstream code was already correct, battle-hardened, well-tested, etc. then I would be huge fan of using it here. The wheel would already have been invented.

If upstream doesn't even have wheels in the first place (as seems to be the case), then implementing the feature in Drake is not "re-inventing" but rather just "inventing". I assume we'll need an MTL parser for the GL render engine anyway, so the only question is the VTK actor calls, which don't seem like a big deal.

___

SeanCurtis-TRI:

Patching upstream is an incredible pain in the ass.  If the upstream code was already correct, battle-hardened, well-tested, etc. then I would be huge fan of using it here. The wheel would already have been invented.

If upstream doesn't even have wheels in the first place (as seems to be the case), then implementing the feature in Drake is not "re-inventing" but rather just "inventing". I assume we'll need an MTL parser for the GL render engine anyway, so the only question is the VTK actor calls, which don't seem like a big deal.
</blockquote></details>

So, the right thing to do is make better use of the tiny obj loader we already have. It's just a case of turning its data into vtk Entities.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkx1bQ8IzUCtmg6xbX:-NSmE8NR49F1joLt7BNb:biy4ci9) (no related file):*

So, the right thing to do is make better use of the tiny obj loader we already have. It's just a case of turning its data into vtk Entities.
</blockquote></details>

Yup.  It would be plausible to imagine paying Kitware to upstream our forthcoming code back into the vtkOBJImporter after the fact, so in a year or three the Ubuntu distributions would have it available for everyone to use. We can ask for a bid on that, but for now I think making progress on our own would be best.

___

zachfang (commented):
I plan to spend more time on this today, but these are some quick replies before I come back in a few hours.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-:-NSmOHx75Jueq50fUtcR:b792yzf)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSgu_g0B9xvYWCtORap:-NSmM5lx-pz6nOWt9LWw:bogszas) (no related file):*

Working. Currently, the code can render color, depth, and label images. However, there is a segfault in the end, which I think it's due to the VTK pointers. Investigating.
</blockquote></details>

One of the segfaults is from freeing `vtkRenderer` that is used for the importer. Using `vtkSmartPointer<vtkRenderer>` resolved the issue. I am looking into another one when loading primitives (assuming this avenue is still worth pursuing).

___
___
*[`geometry/render/test/meshes/4_color_texture.png` line  at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSktCJbA2To8ToYoolZ:-NSmMq9h87NpJGgbrrs9:b-xyp73x) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render/test/meshes/4_color_texture.png#L)):*

Particularly as this is a copy of this png located in a different directory.

As for making *this* particular image "box.png", it's worth noting that various tests will fail. There are tests that coarsely test for the existence of textured objects (but don't want to worry about the details). As such, the solid green renders pixels that can be sampled as evidence of texture. When the texture varies over its surface, we have to be more precise in what we look for (making the tests more brittle).

I think it might also modify the `meshcat_manul_test` -- I believe, it loads the same box (currently green), and its textural description would no longer be valid.

It's worth noting, none of that is an objection to replacing the existing `box.png`, just a note on the broad impact it might have.
</blockquote></details>

One of my intentions is to test which PNG is being loaded, e.g., through an MTL or name association. Other than that, it's a development artifact that can be removed afterward.
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 227 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSm2mbVClMq18hsbxbY:-NSmNZGzCKFFtHsirb0l:bp8oxzj) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render_vtk/internal_render_engine_vtk.h#L227)):*

In a related vein, for mutating the color and label actors, you made those standalone functions in an anonymous namespace in the .cc file (passing the `RenderEngine` parameters as necessary. Why not do the same for the depth actor? Why particularly does *this* one have to be a member method?
</blockquote></details>

The depth shader setting needs a private static member field, `uniform_setting_callback_`.

SeanCurtis-TRI:

Yup.  It would be plausible to imagine paying Kitware to upstream our forthcoming code back into the vtkOBJImporter after the fact, so in a year or three the Ubuntu distributions would have it available for everyone to use. We can ask for a bid on that, but for now I think making progress on our own would be best.
</blockquote></details>

Ok....from a quick face to face. 

There seem to be enough issues/complexities down this road that it behooves us to try an alternate path. So, I'm going to try a quick prototype using `tiny_obj_loader` to parse the obj/mtl files and funnel that directly into VTK. The goal is to convince ourselves as cheaply as possible that it's worth going all in. Stay tuned...

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

___
*[`geometry/render/test/meshes/4_color_texture.png` line  at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSktCJbA2To8ToYoolZ:-NSmXoXqAGLLmFfiza_R:bn8ohrx) ([raw file](https://github.com/RobotLocomotion/drake/blob/342b87bd8350bdb42e364fe3d0d2a6bccd83abda/geometry/render/test/meshes/4_color_texture.png#L)):*

One of my intentions is to test which PNG is being loaded, e.g., through an MTL or name association. Other than that, it's a development artifact that can be removed afterward.
</blockquote></details>

OK I'll withdraw this discussion until the tests are in place.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19168)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @zachfang)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19168#-NSkx1bQ8IzUCtmg6xbX:-NSq3ryD5eo4IDEtAi2K:bj61j5o) (no related file):*

Ok....from a quick face to face. 

There seem to be enough issues/complexities down this road that it behooves us to try an alternate path. So, I'm going to try a quick prototype using `tiny_obj_loader` to parse the obj/mtl files and funnel that directly into VTK. The goal is to convince ourselves as cheaply as possible that it's worth going all in. Stay tuned...
</blockquote></details>

A quick summary of where I stand:

Zack's primary objective (though perhaps I did not clearly state it) was to de-risk the question of having VTK load the materials for us -- is it practical to call into it, and can it do what we need?  The answers were "mostly yes" and "no". The current call pattern needs some love (w.r.t. multiple actors) but the blocker is that VTK's MTL support is so impoverished as to be worthless.

Therefore, we give up on using VTK to load materials, and so the question is -- what else should we do instead?

The next best guess is to call tinyobjloader for parsing and then create the VTK object texturing giblets by hand in Drake code. Sean is de-risking that idea now.

___

PR #19141, opened 2023-04-05
New SpatialInertia factory method SolidBoxWithMass().

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19141)

Issue #18416, opened 2022-12-07
SpatialInertia factory functions for primitive shapes.


mitiguy:
Implement SpatialInertia factory functions for primitive shapes.
- [x] "Prototype PR #18407" that creates a partial fit, form, function for a SpatialInertia factory method for a uniform-density solid box.  Although this is a "prototype PR", it is intended to be merged into master, albeit with a delay on  Python bindings until other primitive shapes also have C++ code.  This "prototype PR" provide a template for the function name, parameter names, documentation, and use in C++.   PR #18407 prototypes the name, unit test, and utility of the newly proposed function.   
1. There were several candidate names including MakeSolidBox(), SolidBox(), SolidBoxViaDensity(), SolidBoxFromDensity() and the winning candidate SolidBoxWithDensity().
2. There were two candidate ways to unit test this function.  The test implemented in the final commit of PR #18407 is shorter and clearer than the initial test (which was correct but overly laborious).  
3. The utility of SolidBoxWithDensity() was investigated by replacing code in geometry_spatial_inertia.c with its shorter call to SolidBoxWithDensity().  Similarly, a few other call sites demonstrated that SolidBoxWithDensity() has utility in more than one part of Drake's C++ code.
- [x] PR #18483 implements all the primitive shapes that are used in geometry_spatial_inertia.cc (SolidCylinderWithDensity(), SolidBoxWithDensity(), SolidCapsuleWithDensity(), SolidEllipsoidWithDensity().
- [x] Create PR to update the documentation and signatures of factories of primitive shapes in unit_inertia.h.  For example, update the signature of a cylinder and capsule which have a unit vector direction that is implied through a default argument or hard-coded and unchangeable.  Update documentation in unit_inertia.h to reflect the improved argument names in spatial_inertia.h.  Deprecate Python bindings.  Note this may be a "breaking change". 
- [x] PRs to implement the remaining primitive shapes used in unit_inertia.cc include PR #18864, PR #18865, PR #18866, PR #18889, PR #18891, PR #18950.
- [x] PR #18618 implements SolidTetrahedronWithDensity() in spatial_inertia.cc and SolidTetrahedron() in unit_inertia.cc.  
- [ ] NOT DONE, BUT MAYBE NOT NEEDED.  Update geometry_spatial_inertia.cc with calls to tetrahedron in unit_inertia.cc and spatial_inertia.cc.  Code that Sean found for calculating mass properties (volume, centroid, inertia) seems to be working. 
[...]

PR #19172, opened 2023-04-11
Python bind SpatialInertia::SolidBoxWithMass().

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19172)

### DIFF
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -707,7 +707,7 @@ class TestPlant(unittest.TestCase):
         SpatialInertia.SolidBoxWithDensity(
             density=1000, lx=0.1, ly=0.2, lz=0.3)
         SpatialInertia.SolidBoxWithMass(
-            mass=0.123, length=0.2)
+            mass=0.123, lx=0.1, ly=0.2, lz=0.3)
         SpatialInertia.SolidCubeWithDensity(
             density=1000, length=0.4)
         SpatialInertia.SolidCapsuleWithDensity(
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -706,6 +706,8 @@ class TestPlant(unittest.TestCase):
             I_SScm_E=RotationalInertia(Ixx=2.0, Iyy=2.3, Izz=2.4))
         SpatialInertia.SolidBoxWithDensity(
             density=1000, lx=0.1, ly=0.2, lz=0.3)
+        SpatialInertia.SolidBoxWithMass(
+            mass=0.123, length=0.2)
         SpatialInertia.SolidCubeWithDensity(
             density=1000, length=0.4)
         SpatialInertia.SolidCapsuleWithDensity(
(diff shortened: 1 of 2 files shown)


mitiguy:
PR https://github.com/RobotLocomotion/drake/pull/19172 is work to helping resolve part of issue https://github.com/RobotLocomotion/drake/issues/18416.

mitiguy (commented):
Feature review +@DamrongGuoy  (after it passes CI).
Please also do Platform review. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19172#-:-NShlh0TEsDoiFMBEO50:b-uai4ck)* status: LGTM missing from assignee DamrongGuoy, needs platform reviewer assigned, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

DamrongGuoy (commented):
:lgtm:

I'm not a platform reviewer, so +@rpoyner-tri for platform review tomorrow (Tuesday).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19172#-:-NShrikn6wb7hyLUPZgc:bh9cber)* status: LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
:lgtm:
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19172#-:-NSlBA9K0QdWACg2HQix:b-9kn1by)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),DamrongGuoy (waiting on @mitiguy)

jwnimmer-tri:
For the record, this little 5-line PR has a non-zero cost for release engineers.  The #19141 was already pretty small (~100 lines).  Maybe next time do the new function and the bindings in the same PR?  It would be lot easier for reviewers and release managers.

Or failing that, use "release notes: none" for PRs like this with bindings that appear within the same release cycle as the new C++ function.

DamrongGuoy:
@mitiguy,  what you described is the opposite of what @jwnimmer-tri asked, which is what I like too.

Right now you are doing this:
PR1 five  C++ functions
PR2 five python bindings

@jwnimmer-tri asked that:
PR1 2 C++ functions + 2 python bindings
PR2 2 C++ functions + 2 python bindings
PR3 1 C++ function + 1 python binding

It would satisfy my need as well if we have both C++ SolidBoxWithMass() and its python binding in the same PR.

We are trying to make Python bindings our first-class citizens for the benefit of the user majority.

I wonder whether we have Style Guide advising that C++ and Python should preferably sync in the same PR?

DamrongGuoy:
Sounds good to me.  First do C++ API until everybody agrees, then add Python binding in the same PR.

Perhaps the reviewer should, then, set `status:do not merge` until python binding is available.

jwnimmer-tri:
I think we can leave it up to the PR author how exactly to handle the code review logistics.  For some people, keeping the Python and C++ in sync from the first draft onward is not onerous, so we shouldn't be too prescriptive about the mechanism.  The more important point is the result to aim for, i.e., that for a new public API typically Python bindings happen concurrently or almost-concurrently with the C++ API.

> @jwnimmer-tri asked that:
> PR1 2 C++ functions + 2 python bindings
> PR2 2 C++ functions + 2 python bindings
> PR3 1 C++ function + 1 python binding

To be precise, I suggested just one PR with all 5 function in both C++ and Python.  The reason to split into multiple PRs would be if the lines of code blew through the 1500 ceiling, which the "inertia of primitive shape" functions and test did not.

PR #19162, opened 2023-04-09
Implements MultibodyPlant::GetPositionNames()

RussTedrake:
Towards #10736, to help with MultibodyPlant introspection.

PR #19174, opened 2023-04-11
Adds GetVelocityNames, GetStateNames, and GetActuatorNames to MultibodyPlant

### LINKED ISSUE #10736: MbP introspection
A number of user requests regarding topology and state introspection for an MBP model have been made. These include: 

- [x] Retrieve the floating base bodies in an MBP. Proposed API: **std::vector\<BodyIndex\> GetFloatingBaseBodies().**
- [x] Free body state introspection. Proposed API: **Body::is_floating_base()**, **Body::is_quaternion_floating_base()**, **Body::get_position_start()**, **Body::get_velocity_start()**, etc. Where the `get_foo_index()` methods would throw if the body is not a floating base.
- [x] Get the base for an articulated body for a given model instance (per request in #9747). Proposed API: **BodyIndex GetUniqueBaseBody(ModelInstanceIndex instance)**.
- [ ] **MBP::PrintStateNames(), MBP::PrintStateNames(ModelInstanceIndex)** (or similar name), prints a table of "index into the state vs. state name".
- [ ] **MBP::PrintInputActuationNames(), MBP::PrintInputActuationNames(ModelInstanceIndex)** (or similar name), prints a table for input vectors `u`, also index vs. actuator name (it could also include actuated joint name).
- [ ] ~~**std::string MBP::get_state_name(int i)**, to retrieve the name of the i-th entry in a bare state vector.~~

cc'ing @mposa, @RussTedrake, @sammy-tri, @siyuanfeng-tri

RussTedrake:
Resolves #10736.

+@xuchenhan-tri for feature review, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -1146,7 +1146,54 @@ void DoScalarDependentDefinitions(py::module m, T) {
             [](const Class* self, const Context<T>& context, State<T>* state) {
               self->SetDefaultState(context, state);
             },
-            py::arg("context"), py::arg("state"), cls_doc.SetDefaultState.doc);
+            py::arg("context"), py::arg("state"), cls_doc.SetDefaultState.doc)
+        .def("GetPositionNames",
+            overload_cast_explicit<std::vector<std::string>, bool, bool>(
+                &Class::GetPositionNames),
+            py::arg("add_model_instance_prefix") = false,
+            py::arg("always_add_suffix") = true,
+            cls_doc.GetPositionNames.doc_2args)
+        .def("GetPositionNames",
+            overload_cast_explicit<std::vector<std::string>, ModelInstanceIndex,
+                bool, bool>(&Class::GetPositionNames),
+            py::arg("model_instance"),
+            py::arg("add_model_instance_prefix") = false,
+            py::arg("always_add_suffix") = true,
+            cls_doc.GetPositionNames.doc_3args)
+        .def("GetVelocityNames",
+            overload_cast_explicit<std::vector<std::string>, bool, bool>(
+                &Class::GetVelocityNames),
+            py::arg("add_model_instance_prefix") = false,
+            py::arg("always_add_suffix") = true,
+            cls_doc.GetVelocityNames.doc_2args)
+        .def("GetVelocityNames",
+            overload_cast_explicit<std::vector<std::string>, ModelInstanceIndex,
+                bool, bool>(&Class::GetVelocityNames),
+            py::arg("model_instance"),
+            py::arg("add_model_instance_prefix") = false,
+            py::arg("always_add_suffix") = true,
+            cls_doc.GetVelocityNames.doc_3args)
+        .def("GetStateNames",
+            overload_cast_explicit<std::vector<std::string>, bool>(
+                &Class::GetStateNames),
+            py::arg("add_model_instance_prefix") = false,
+            cls_doc.GetStateNames.doc_1args)
+        .def("GetStateNames",
+            overload_cast_explicit<std::vector<std::string>, ModelInstanceIndex,
+                bool>(&Class::GetStateNames),
+            py::arg("model_instance"),
+            py::arg("add_model_instance_prefix") = false,
+            cls_doc.GetStateNames.doc_2args)
+        .def("GetActuatorNames",
+            overload_cast_explicit<std::vector<std::string>, bool>(
+                &Class::GetActuatorNames),
+            py::arg("add_model_instance_prefix"),
+            cls_doc.GetActuatorNames.doc_1args)
+        .def("GetActuatorNames",
+            overload_cast_explicit<std::vector<std::string>, ModelInstanceIndex,
+                bool>(&Class::GetActuatorNames),
+            py::arg("model_instance"), py::arg("add_model_instance_prefix"),
+            cls_doc.GetActuatorNames.doc_2args);
   }
 
   {
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -1225,6 +1225,43 @@ class TestPlant(unittest.TestCase):
         # Test existence of context resetting methods.
         plant.SetDefaultState(context, state=context.get_mutable_state())
 
+        self.assertEqual(
+            plant.GetPositionNames(add_model_instance_prefix=False,
+                                   always_add_suffix=False),
+            ["ShoulderJoint", "ElbowJoint"])
+        self.assertEqual(
+            plant.GetPositionNames(model_instance=instance,
+                                   add_model_instance_prefix=False,
+                                   always_add_suffix=False),
+            ["ShoulderJoint", "ElbowJoint"])
+        self.assertEqual(
+            plant.GetVelocityNames(model_instance=instance,
+                                   add_model_instance_prefix=False,
+                                   always_add_suffix=False),
+            ["ShoulderJoint", "ElbowJoint"])
+        self.assertEqual(
+            plant.GetVelocityNames(add_model_instance_prefix=False,
+                                   always_add_suffix=False),
+            ["ShoulderJoint", "ElbowJoint"])
+        self.assertEqual(
+            plant.GetStateNames(add_model_instance_prefix=False), [
+                                    "ShoulderJoint_q", "ElbowJoint_q",
+                                    "ShoulderJoint_w", "ElbowJoint_w"
+                                ])
+        self.assertEqual(
+            plant.GetStateNames(model_instance=instance,
+                                add_model_instance_prefix=False), [
+                                    "ShoulderJoint_q", "ElbowJoint_q",
+                                    "ShoulderJoint_w", "ElbowJoint_w"
+                                ])
+        self.assertEqual(
+            plant.GetActuatorNames(add_model_instance_prefix=False),
+            ["ElbowJoint"])
+        self.assertEqual(
+            plant.GetActuatorNames(model_instance=instance,
+                                   add_model_instance_prefix=False),
+            ["ElbowJoint"])
+
         # Test existence of default free body pose setting.
         body = plant.GetBodyByName("Link1")
         X_WB_default = RigidTransform_[float]()
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -1362,30 +1362,41 @@ void MultibodyPlant<T>::SetDefaultPositions(ModelInstanceIndex model_instance,
 
 template <typename T>
 std::vector<std::string> MultibodyPlant<T>::GetPositionNames(
-    std::optional<ModelInstanceIndex> model_instance,
-    std::optional<bool> add_model_instance_prefix,
+    bool add_model_instance_prefix, bool always_add_suffix) const {
+  DRAKE_MBP_THROW_IF_NOT_FINALIZED();
+  std::vector<std::string> names(num_positions());
+
+  for (int joint_index = 0; joint_index < num_joints(); ++joint_index) {
+    const Joint<T>& joint = get_joint(JointIndex(joint_index));
+    const std::string prefix =
+        add_model_instance_prefix
+            ? fmt::format("{}_", GetModelInstanceName(joint.model_instance()))
+            : "";
+    for (int i = 0; i < joint.num_positions(); ++i) {
+      const std::string suffix =
+          always_add_suffix || joint.num_positions() > 1
+              ? fmt::format("_{}", joint.position_suffix(i))
+              : "";
+      names[joint.position_start() + i] =
+          fmt::format("{}{}{}", prefix, joint.name(), suffix);
+    }
+  }
+  return names;
+}
+
+template <typename T>
+std::vector<std::string> MultibodyPlant<T>::GetPositionNames(
+    ModelInstanceIndex model_instance, bool add_model_instance_prefix,
     bool always_add_suffix) const {
   DRAKE_MBP_THROW_IF_NOT_FINALIZED();
-  std::vector<std::string> names;
-  std::vector<JointIndex> joint_indices;
-  int position_offset = 0;
-  if (model_instance) {
-    names.resize(num_positions(*model_instance));
-    joint_indices = GetJointIndices(*model_instance);
-    // The offset into the position array is the position_start of the first
-    // mobilizer in the tree; here we just take the minimum.
-    position_offset = num_positions();
-    for (const auto& joint_index : joint_indices) {
-      position_offset =
-          std::min(position_offset, get_joint(joint_index).position_start());
-    }
-  } else {
-    names.resize(num_positions());
-    joint_indices.resize(num_joints());
-    std::iota(joint_indices.begin(), joint_indices.end(), JointIndex(0));
-  }
-  if (!add_model_instance_prefix) {
-    add_model_instance_prefix = !model_instance.has_value();
+  std::vector<std::string> names(num_positions(model_instance));
+  std::vector<JointIndex> joint_indices = GetJointIndices(model_instance);
+  // The offset into the position array is the position_start of the first
+  // mobilizer in the tree; here we just take the minimum.
+  int position_offset = num_positions();
+  for (const auto& joint_index : joint_indices) {
+    position_offset =
+        std::min(position_offset, get_joint(joint_index).position_start());
   }
 
   for (const auto& joint_index : joint_indices) {
@@ -1397,12 +1408,14 @@ std::vector<std::string> MultibodyPlant<T>::GetPositionNames(
                  static_cast<int>(names.size()));
 
     const std::string prefix =
-        *add_model_instance_prefix
-            ? fmt::format("{}_", GetModelInstanceName(joint.model_instance()))
+        add_model_instance_prefix
+            ? fmt::format("{}_", GetModelInstanceName(model_instance))
             : "";
     for (int i = 0; i < joint.num_positions(); ++i) {
-      const std::string suffix = always_add_suffix || joint.num_positions() > 1
-          ? fmt::format("_{}", joint.position_suffix(i)) : "";
+      const std::string suffix =
+          always_add_suffix || joint.num_positions() > 1
+              ? fmt::format("_{}", joint.position_suffix(i))
+              : "";
       names[joint.position_start() + i - position_offset] =
           fmt::format("{}{}{}", prefix, joint.name(), suffix);
     }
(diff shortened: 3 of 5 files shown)


xuchenhan-tri (commented):
:lgtm: feature with a few minor things.
+@rpoyner-tri for platform review please (if you haven't timed out)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-:-NSm6ctN4RHfvcJPZ4Ot:b-naq9pj)* status: 4 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @RussTedrake)

___
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSlSkBU354Lc8Da46AS:bn4ypo0) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*
> ```objc
>   ///
>   /// The returned names are guaranteed to be unique if the @p model_instance is
>   /// provided OR if @p add_model_instance_prefix is `true` (the default).
> ```

The default is std::nullopt, not `true`. Ditto for all new functions.



_Code quote:_
```objc
@p add_model_instance_prefix is `true` (the default)
```
___
*[`multibody/plant/multibody_plant.h` line 2396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm0mF_A4x1237Pxujy:-NSm0mF_A4x1237Pxujz:b-smn495) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2396)):*
> ```objc
>   /// Returns a list of string names corresponding to each element of the
>   /// multibody state vector. These strings take the form
>   /// `{model_instance_name}_{joint_name}_{joint_position_suffix}`, but the
> ```

minor, typo?



_Suggestion:_
```objc
joint_state_suffix
```
___
*[`multibody/plant/multibody_plant.cc` line 1475 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm3Pll2r83MAOy8jMP:-NSm3Pll2r83MAOy8jMQ:b-3wwvf0) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1475)):*
> ```cpp
>   std::vector<std::string> velocity_names = GetVelocityNames(
>       model_instance, add_model_instance_prefix, true /* always_add_suffix */);
>   names.insert(names.end(), velocity_names.begin(), velocity_names.end());
> ```

BTW, you can move the velocity_names into names
`names.insert(names.end(), std::make_move_iterator(velocity_names.begin()), std::make_move_iterator(velocity_names.end()));`
___
*[`multibody/plant/multibody_plant.cc` line 1512 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm4GQJ6sgJRSMtCABb:-NSm4GQJ6sgJRSMtCABc:b3h552s) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1512)):*
> ```cpp
>     DRAKE_DEMAND(actuator.input_start() >= offset);
>     DRAKE_DEMAND(actuator.input_start() - offset <=
>                  static_cast<int>(names.size()));
> ```

nit, I believe this should be a < here.



_Suggestion:_
```cpp
    DRAKE_DEMAND(actuator.input_start() - offset <
                 static_cast<int>(names.size()));
```

rpoyner-tri (commented):
:lgtm: optional boolery

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-:-NSmPZH3D5C3yWX_IqbC:b-wktde7)* status: 5 unresolved discussions (waiting on @RussTedrake and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmLpHa-gBZHPWp2UVi:b-bmll1t) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

The default is std::nullopt, not `true`. Ditto for all new functions.
</blockquote></details>

The documentation (like everything else) of optional<bool> is tricky. The whole point of it here (aside from historical path dependence) is that the caller can say `{}` to get default behavior without needing to remember whether default behavior is spelled true or false. Of course (in the trivial sense) the default is std::nullopt, but the doc is trying to indicate "the behavior if unset". I'm not sure I have better words than the existing text, but perhaps there are some.
___
*[`multibody/plant/multibody_plant.cc` line 1439 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSmJqNj3TAH6h_1mqfe:-NSmJqNj3TAH6h_1mqff:bvy7q16) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1439)):*
> ```cpp
>     std::iota(joint_indices.begin(), joint_indices.end(), JointIndex(0));
>   }
>   if (!add_model_instance_prefix) {
> ```

I know these lines are likely copied, but (having recently fought with subtle `optional<bool>`-induced bugs), they read like a Zen koan.

"""If there is no value, then make the value be the opposite of its emptiness."""

Perhaps we could afford the few extra cycles of:
```
add_model_instance_prefix = add_model_instance_prefix.value_or(true);
```
or even
```
const bool add_prefix = add_model_instance_prefix.value_or(true);
```
and relieve the code below of having to remember to `*`-deref the optional bool?

I accept (with some regret) the need for optional<bool> as a parameter type, but I hope that we can make the code that reads such variables less fragile.

jwnimmer-tri:
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmRh916ym7_xUmDZqu:b-ss09t4) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*
> ... is that the caller can say {} to get default behavior without needing to remember whether default behavior is spelled true or false.

Is that actually helpful though?  It seems like a huge readability burden for almost no residual value.

Coming into this discussion cold, I would have thought this to be the most readable solution, for both our docs and implementation as well as the call sites:

```c++
  std::vector<std::string> GetPositionNames(
      std::optional<ModelInstanceIndex> model_instance = std::nullopt,
      bool add_model_instance_prefix = true,
      bool always_add_suffix = true) const;
```

I would need to see a sample call site that benefits from the `optional<bool>` to be convinced otherwise.

jwnimmer-tri:
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmSm36EyRHedUkIZWu:b-8q7ca8) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

> ... is that the caller can say {} to get default behavior without needing to remember whether default behavior is spelled true or false.

Is that actually helpful though?  It seems like a huge readability burden for almost no residual value.

Coming into this discussion cold, I would have thought this to be the most readable solution, for both our docs and implementation as well as the call sites:

```c++
  std::vector<std::string> GetPositionNames(
      std::optional<ModelInstanceIndex> model_instance = std::nullopt,
      bool add_model_instance_prefix = true,
      bool always_add_suffix = true) const;
```

I would need to see a sample call site that benefits from the `optional<bool>` to be convinced otherwise.
</blockquote></details>

Ah, I see.  The default isn't "true", the default is "the opposite of whether I passed a model instance index". So it really is a tribool.

That means the solution is less clear, even though the signature is still defectively confusing.

jwnimmer-tri:
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmTpGwBOhMxcOJO9ze:be537cm) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

Ah, I see.  The default isn't "true", the default is "the opposite of whether I passed a model instance index". So it really is a tribool.

That means the solution is less clear, even though the signature is still defectively confusing.
</blockquote></details>

Actually, I think my proposed signature is fine.  It means that people who want pass a model instance index and want a really short name need to call `GetPositionNames(index, false, false)` but that seems better?  Opting-in to terseness seems to be the shape we're aiming for here.  The defaults are obvious (verbosity) and don't require carefully reading the docs to understand nullability semantics.

RussTedrake:
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmlMkT7y18V9Dj8iLd:bb57us0) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

Actually, I think my proposed signature is fine.  It means that people who want pass a model instance index and want a really short name need to call `GetPositionNames(index, false, false)` but that seems better?  Opting-in to terseness seems to be the shape we're aiming for here.  The defaults are obvious (verbosity) and don't require carefully reading the docs to understand nullability semantics.
</blockquote></details>

I've been using my own python versions of these methods for some time.  having the model instance and not the prefix is by far the most common case.  I have many call sites that do this.

Can you suggest a call site where would pass a model instance and actually want the model instance name stapled on the front (the same string on the front of every name)?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: 5 unresolved discussions (waiting on @RussTedrake and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmnl7E0l4LIHkeOUDU:b-sz6nsi) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*
> I've been using my own python versions of these methods for some time. having the model instance and not the prefix is by far the most common case. I have many call sites that do this.

That use case was not clear from the PR overview nor the linked issue.

The Python signature is also only half the story.  It's possible that what's ergonomic at Python call sites is not ergonomic in C++, and we need to accommodate both.

> Can you suggest a call site where would pass a model instance and actually want the model instance name stapled on the front (the same string on the front of every name)?

A user might assemble a collection of position names one model instance at a time by calling this function repeatedly, e.g., on the set of robot_model_instances but not the set of manipuland_model_instances. They might still dump all of the position names into a single dictionary of all robots' positions, where they want them to be globally unique.

That's not the right question to ask though. In both the current draft and my proposal, the user has the _option_ of having the prefix or dropping the prefix.  The only debate is around the defaults.

My premise was that saying `GetPositionNames(model_instance=iiwa, add_model_instance_prefix=False)` at the call site leads to clearer code than a confusing docstring with a confusing tribool.  The True/False/None is sometimes clear enough in Python but it's *extraordinarily bad* in C++ to the point where I intend to forbid it in the style guide.

Maybe I'm wrong, though?  To answer that we need _a sample of call sites_ and we don't have that in evidence yet.

Thinking more carefully, I think the root problem here is abuse of overloading.

As I understand it, nobody is advocating that we should allow passing "None" or "null" or "{}" to this function, nor be able to pass "null" as the model instance.  The only reason we have a tribool in play is because we want a different default value contingent on whether or not there was a model instance index also passed.

In short: optional is not a trick to get around function overloading. We should only use optional when a user literally wants to pass "null" as the argument value, e.g., if they _don't know_ whether or not they have a model instance.  I don't think that applies for these functions. The user either wants all names, or they definitely have an index and only want the names for that one index.  That's what overloading is for.

If there were two functions (one taking a model instance, and the other not) the confusion about the default would vanish.  This came up in the prior PR #19162 review from Xuchen and got rejected.  I think he was probably right to argue the point:

```c++
  std::vector<std::string> GetPositionNames(
      bool add_model_instance_prefix = true,
      bool always_add_suffix = true) const;

  std::vector<std::string> GetPositionNames(
      ModelInstanceIndex model_instance,
      bool add_model_instance_prefix = false,
      bool always_add_suffix = true) const;
```

Those two signatures are _very clear_ about what's happening and why, and seem like they would match your current preferred spelling of call sites, with no disruption.

The objection from the prior PR was implementation effort.  It's not _that_ difficult to have both functions share a common implementation function, even with two public entry points. If the only argument against two functions was a few more lines of boilerplate in the implementation, I think that loses out versus API clarity and the whole debate.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: 1 unresolved discussion (waiting on @rpoyner-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSn0GKM1u8oYMsUOTPw:b-91jwsd) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

> I've been using my own python versions of these methods for some time. having the model instance and not the prefix is by far the most common case. I have many call sites that do this.

That use case was not clear from the PR overview nor the linked issue.

The Python signature is also only half the story.  It's possible that what's ergonomic at Python call sites is not ergonomic in C++, and we need to accommodate both.

> Can you suggest a call site where would pass a model instance and actually want the model instance name stapled on the front (the same string on the front of every name)?

A user might assemble a collection of position names one model instance at a time by calling this function repeatedly, e.g., on the set of robot_model_instances but not the set of manipuland_model_instances. They might still dump all of the position names into a single dictionary of all robots' positions, where they want them to be globally unique.

That's not the right question to ask though. In both the current draft and my proposal, the user has the _option_ of having the prefix or dropping the prefix.  The only debate is around the defaults.

My premise was that saying `GetPositionNames(model_instance=iiwa, add_model_instance_prefix=False)` at the call site leads to clearer code than a confusing docstring with a confusing tribool.  The True/False/None is sometimes clear enough in Python but it's *extraordinarily bad* in C++ to the point where I intend to forbid it in the style guide.

Maybe I'm wrong, though?  To answer that we need _a sample of call sites_ and we don't have that in evidence yet.

Thinking more carefully, I think the root problem here is abuse of overloading.

As I understand it, nobody is advocating that we should allow passing "None" or "null" or "{}" to this function, nor be able to pass "null" as the model instance.  The only reason we have a tribool in play is because we want a different default value contingent on whether or not there was a model instance index also passed.

In short: optional is not a trick to get around function overloading. We should only use optional when a user literally wants to pass "null" as the argument value, e.g., if they _don't know_ whether or not they have a model instance.  I don't think that applies for these functions. The user either wants all names, or they definitely have an index and only want the names for that one index.  That's what overloading is for.

If there were two functions (one taking a model instance, and the other not) the confusion about the default would vanish.  This came up in the prior PR #19162 review from Xuchen and got rejected.  I think he was probably right to argue the point:

```c++
  std::vector<std::string> GetPositionNames(
      bool add_model_instance_prefix = true,
      bool always_add_suffix = true) const;

  std::vector<std::string> GetPositionNames(
      ModelInstanceIndex model_instance,
      bool add_model_instance_prefix = false,
      bool always_add_suffix = true) const;
```

Those two signatures are _very clear_ about what's happening and why, and seem like they would match your current preferred spelling of call sites, with no disruption.

The objection from the prior PR was implementation effort.  It's not _that_ difficult to have both functions share a common implementation function, even with two public entry points. If the only argument against two functions was a few more lines of boilerplate in the implementation, I think that loses out versus API clarity and the whole debate.
</blockquote></details>

Done.  

Tell me what you really think.  ;-)
___
*[`multibody/plant/multibody_plant.h` line 2396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm0mF_A4x1237Pxujy:-NSn6QC2-JLu6R2CsAs4:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2396)):*

minor, typo?
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 1439 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSmJqNj3TAH6h_1mqfe:-NSn6SLp64id8X3txNyR:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1439)):*

I know these lines are likely copied, but (having recently fought with subtle `optional<bool>`-induced bugs), they read like a Zen koan.

"""If there is no value, then make the value be the opposite of its emptiness."""

Perhaps we could afford the few extra cycles of:
```
add_model_instance_prefix = add_model_instance_prefix.value_or(true);
```
or even
```
const bool add_prefix = add_model_instance_prefix.value_or(true);
```
and relieve the code below of having to remember to `*`-deref the optional bool?

I accept (with some regret) the need for optional<bool> as a parameter type, but I hope that we can make the code that reads such variables less fragile.
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 1475 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm3Pll2r83MAOy8jMP:-NSn6ZWA3tIR9hI-IS6-:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1475)):*

BTW, you can move the velocity_names into names
`names.insert(names.end(), std::make_move_iterator(velocity_names.begin()), std::make_move_iterator(velocity_names.end()));`
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 1512 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSm4GQJ6sgJRSMtCABb:-NSn6pKn25kcoVkcjvDq:bkz5ckv) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1512)):*

nit, I believe this should be a < here.
</blockquote></details>

Done.  Good catch.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: 1 unresolved discussion (waiting on @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.cc` line 1439 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSmJqNj3TAH6h_1mqfe:-NSmo9tjDhXaB02MZjuc:bqi0t0o) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.cc#L1439)):*

Done.
</blockquote></details>

Ah, see I misread it myself in the first comment. However, glad to see how things turned out :confetti_ball:

rpoyner-tri (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: 1 unresolved discussion (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: 1 unresolved discussion (waiting on @RussTedrake)

___
*[`multibody/plant/multibody_plant.h` line 2357 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSqENEp3mqaALYatY5T:-NSqENEp3mqaALYatY5U:bfcja2d) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2357)):*
> ```objc
>   /// added iff `model_instance == std::nullopt`.
>   /// @param always_add_suffix (optional). If true, then the suffix is always
>   /// added. If false, then the suffix is only added for joints that have more
> ```

minor, the explanation of the flag `always_add_suffix` which I found helpful is removed in the latest revision. Is this intentional?



_Suggestion:_
```objc
  /// @param always_add_suffix (optional). If true, then the suffix is always
  /// added. If false, then the suffix is only added for joints that have more
  /// than one position (in this case, not adding would lead to ambiguity).
```
___
*[`multibody/plant/multibody_plant.h` line 2361 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSlSkBU354Lc8Da46AR:-NSmU12JAjCZ389rtReF:bxzsyol) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2361)):*

Done.  

Tell me what you really think.  ;-)
</blockquote></details>

Thanks @jwnimmer-tri @RussTedrake. I think the new API is much clearer.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),xuchenhan-tri(platform) (waiting on @rpoyner-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 2357 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19174#-NSqENEp3mqaALYatY5T:-NSr7H893EIbKmeOFH5M:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bca124d10b984349906c9b0e760ecf0a90b7e69/multibody/plant/multibody_plant.h#L2357)):*

minor, the explanation of the flag `always_add_suffix` which I found helpful is removed in the latest revision. Is this intentional?
</blockquote></details>

Done.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19174)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),xuchenhan-tri(platform) (waiting on @RussTedrake)

PR #19176, opened 2023-04-11
QueryObject allows signed distance between a filtered geometry pair

### LINKED ISSUE #19135: ComputeSignedDistancePairClosestPoints on a filtered pair

QueryObject.ComputeSignedDistancePairClosestPoints has this to say:

```c++
  /** A variant of ComputeSignedDistancePairwiseClosestPoints() which computes
   the signed distance (and witnesses) between a specific pair of geometries
   indicated by id.
   ...
   @throws std::exception if ... the pair (A, B) has been marked as filtered ... */
  SignedDistancePair<T> ComputeSignedDistancePairClosestPoints(
      GeometryId geometry_id_A, GeometryId geometry_id_B) const;
```

Note the case of throwing `std::exception`.

There are cases where it's useful to compute the signed distance, even for objects which cannot collide.  This measurement can inform control algorithms or teleop panels.  Right now, there does not appear to be any way to easily calculate a distance for the case of a non-colliding pair.

Change `ComputeSignedDistancePairClosestPoints` to ignore collision filters in this function, so that it returns a value in this case instead of throwing.

Maybe we could make "filtered throws vs filtered returns" a function argument?  I don't see any reason the user ever _wants_ an exception in this case, though.

**Additional context**

[Slack thread](https://app.slack.com/client/T0JNB2DS4/C2WBPQDB7/thread/C2WBPQDB7-1680472087.332559)

SeanCurtis-TRI:
Historically, requesting signed distance between two geometries that had been filtered would throw. While it might make sense to have the geometries filtered for contact, it might still be convenient to know how far they are. This changes the API to allow the caller to demand that collision filters be ignored.

closes #19135

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176)

### DIFF
--- a/geometry/proximity/distance_to_shape_callback.cc
+++ b/geometry/proximity/distance_to_shape_callback.cc
@@ -236,8 +236,9 @@ bool Callback(fcl::CollisionObjectd* object_A_ptr,
   const EncodedData encoding_a(*object_A_ptr);
   const EncodedData encoding_b(*object_B_ptr);
 
-  const bool can_collide = data.collision_filter.CanCollideWith(
-      encoding_a.id(), encoding_b.id());
+  const bool can_collide =
+      data.collision_filter == nullptr ||
+      data.collision_filter->CanCollideWith(encoding_a.id(), encoding_b.id());
 
   if (can_collide) {
     // Throw if the geometry-pair isn't supported.
--- a/geometry/proximity/distance_to_shape_callback.h
+++ b/geometry/proximity/distance_to_shape_callback.h
@@ -40,7 +40,8 @@ struct CallbackData {
    all aliased in the data and require the aliased parameters to remain valid at
    least as long as the CallbackData instance.
 
-   @param collision_filter_in     The collision filter system. Aliased.
+   @param collision_filter_in     The collision filter system. Aliased. If null,
+                                  collision filters will not be considered.
    @param X_WGs_in                The T-valued poses. Aliased.
    @param max_distance_in         The maximum distance at which a pair is
                                   reported.
@@ -50,23 +51,22 @@ struct CallbackData {
       const std::unordered_map<GeometryId, math::RigidTransform<T>>* X_WGs_in,
       double max_distance_in,
       std::vector<SignedDistancePair<T>>* nearest_pairs_in)
-      : collision_filter(*collision_filter_in),
+      : collision_filter(collision_filter_in),
         X_WGs(*X_WGs_in),
         max_distance(max_distance_in),
         nearest_pairs(*nearest_pairs_in) {
-    DRAKE_DEMAND(collision_filter_in != nullptr);
     DRAKE_DEMAND(X_WGs_in != nullptr);
     DRAKE_DEMAND(nearest_pairs_in != nullptr);
   }
 
   /* The collision filter system.  */
-  const CollisionFilter& collision_filter;
+  const CollisionFilter* collision_filter{};
 
   /* The T-valued poses of all geometries.  */
   const std::unordered_map<GeometryId, math::RigidTransform<T>>& X_WGs;
 
   /* The maximum distance at which a pair's distance will be reported.  */
-  const double max_distance;
+  const double max_distance{};
 
   /* The distance query parameters.  */
   fcl::DistanceRequestd request;
--- a/geometry/proximity/test/distance_sphere_to_shape_test.cc
+++ b/geometry/proximity/test/distance_sphere_to_shape_test.cc
@@ -689,6 +689,12 @@ GTEST_TEST(Callback, RespectCollisionFiltering) {
   threshold = std::numeric_limits<double>::max();
   Callback<double>(&sphere_A, &sphere_B, &data, threshold);
   EXPECT_EQ(results.size(), 0u);
+
+  // However, if we explicitly request collision filters be ignored, it will do
+  // so.
+  data.collision_filter = nullptr;
+  Callback<double>(&sphere_A, &sphere_B, &data, threshold);
+  EXPECT_EQ(results.size(), 1u);
 }
 
 // Confirms that regardless of the order of the two objects, the result always
--- a/geometry/proximity_engine.cc
+++ b/geometry/proximity_engine.cc
@@ -603,8 +603,8 @@ class ProximityEngine<T>::Impl : public ShapeReifier {
     std::vector<SignedDistancePair<T>> witness_pairs;
     double max_distance = std::numeric_limits<double>::infinity();
     // All these quantities are aliased in the callback data.
-    shape_distance::CallbackData<T> data{&collision_filter_, &X_WGs,
-                                         max_distance, &witness_pairs};
+    shape_distance::CallbackData<T> data{nullptr, &X_WGs, max_distance,
+                                         &witness_pairs};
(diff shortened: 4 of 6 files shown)


SeanCurtis-TRI (commented):
+@DamrongGuoy for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-:-NSlKSg66jUrSpsxjAHf:b-feiw5h)* status: LGTM missing from assignee DamrongGuoy, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @DamrongGuoy)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176)* status: LGTM missing from assignee DamrongGuoy, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @DamrongGuoy)

___
*[`geometry/query_object.h` line 612 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSlKnJL0iXz-5eC5Koc:-NSlODMB0mxgAvxCYuVl:bwilfwa) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fa11fe30a951bf094ea79321501fdd9f32426fb/geometry/query_object.h#L612)):*

My hypothesis is that no user ever would want an exception in this case.  Could you explain a plausible use case where the user wants it?

I can _maybe_ get behind this as a deprecation shim (though I don't think we strictly need to deprecated), but that would at least mean a dated TODO here to drop the option.
</blockquote></details>

Oops. Sorry.  Went off in the wrong direction.

DamrongGuoy (commented):
:lgtm: with minor questions.

+@sammy-tri for platform review tomorrow (Thursday), please.

> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-:-NSs1DuQBsIMtqQBquk_:b8cb003)* status: 4 unresolved discussions, LGTM missing from assignee sammy-tri(platform), missing label for release notes (waiting on @sammy-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSs1aBY7aixNsM35H6g:-NSs1aBY7aixNsM35H6h:b-b3uc9e) (no related file):*
I'm just curious why there were a number of revert files (geometry_state.h, geometry_state.cc, proximity_engine.h, query_object.cc, test/geometry_state_test.cc).  Was it a different early design?

___
___
*[`geometry/proximity_engine.cc` line 635 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSrsmod9s00HrKC9RNj:-NSrsmoe-ubgILVq5WWL:b-6uwzj2) ([raw file](https://github.com/RobotLocomotion/drake/blob/dec251cc306277742090f3ec1ecb64824e4d57da/geometry/proximity_engine.cc#L635)):*
> ```cpp
>           "The geometry pair ({}, {}) does not support a signed distance query",
>           id_A, id_B));
>     }
> ```

BTW, kcov said this block is not tested.  However, since it's there already, it's not a defect of this PR.  It's just my observation in case you think it's relevant.





_Code quote:_
```cpp
    if (witness_pairs.size() == 0) {
      throw std::runtime_error(fmt::format(
          "The geometry pair ({}, {}) does not support a signed distance query",
          id_A, id_B));
    }
```
___
*[`geometry/query_object.h` line 620 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSrwiVU0DH6qQkgKX9m:-NSrwiVU0DH6qQkgKX9n:b-xg7o1f) ([raw file](https://github.com/RobotLocomotion/drake/blob/dec251cc306277742090f3ec1ecb64824e4d57da/geometry/query_object.h#L620)):*
> ```objc
>                           scalar support table.
>    @warning For Mesh shapes, their convex hulls are used in this query. It is
>             *not* computationally efficient or particularly accurate.  */
> ```

BTW, since we're touching this file, I saw that Doxygen shows `It is not*`  instead of `It is *not*`. I'm not sure why.  Again, it's not a defect of this PR.


___
*[`geometry/test/proximity_engine_test.cc` line 909 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSryJ5Q79fh9vzjrn-h:-NSryJ5Q79fh9vzjrn-i:b4b1kwh) ([raw file](https://github.com/RobotLocomotion/drake/blob/dec251cc306277742090f3ec1ecb64824e4d57da/geometry/test/proximity_engine_test.cc#L909)):*
> ```cpp
>   }
> 
>   // Case: the pair is filtered.
> ```

nit, should we update the explanation here?  I assume it checks this case, right?
```
This query is unique among the distance queries in that it doesn't respect collision filters. As long as the geometry ids refer to geometries with proximity roles, signed distance can be computed (subject to supported scalar tables above).
```



_Suggestion:_
```cpp
 // Case: regardless of the collision filter, we still compute the signed distance.
```

sammy-tri (commented):
:lgtm: (pending the remaining nits)

> Reviewed 5 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-:-NSvnpuVCYVvCuum4GRW:b-kvkecs)* status: 4 unresolved discussions, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176)* status: missing label for release notes (waiting on @DamrongGuoy and @sammy-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSs1aBY7aixNsM35H6g:-NTF3Ge083ielVjZiKDS:b-onptxj) (no related file):*

I'm just curious why there were a number of revert files (geometry_state.h, geometry_state.cc, proximity_engine.h, query_object.cc, test/geometry_state_test.cc).  Was it a different early design?
</blockquote></details>

The first revision implemented something different. In backing out that "solution", some of the related changes got undone.

___
___
*[`geometry/proximity_engine.cc` line 635 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19176#-NSrsmod9s00HrKC9RNj:-NTF2TA4E1R0xY1PASJA:b1dlono) ([raw file](https://github.com/RobotLocomotion/drake/blob/dec251cc306277742090f3ec1ecb64824e4d57da/geometry/proximity_engine.cc#L635)):*

BTW, kcov said this block is not tested.  However, since it's there already, it's not a defect of this PR.  It's just my observation in case you think it's relevant.



</blockquote></details>

Upon closer examination, I think that code is no longer relevant. The `Callback` code has two reasons for returning without writing to `witness_pairs`:

  - The pair is filtered -- this call site has gone out of its way to make sure this is not the case.
  - The distance exceeds the "max" distance. Again, we initialize the max distance to infinity to preclude this possibility.

Therefore, we simply assume that it'll always return a value (so long as it doesn't throw for some alternate reason).

sammy-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19176)* status: missing label for release notes (waiting on @SeanCurtis-TRI)

PR #19192, opened 2023-04-12
Throw exceptions for spatial inertia primitives with negative mass or density.

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192)

### DIFF
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -6,6 +6,13 @@
 
 namespace drake {
 namespace multibody {
+// namespace {
+
+template <typename T>
+static const boolean<T> is_positive_and_finite(const T& value) {
+  using std::isfinite;
+  return isfinite(value) && value > 0;
+}
 
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::MakeUnitary() {
@@ -18,9 +25,9 @@ SpatialInertia<T> SpatialInertia<T>::MakeUnitary() {
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::PointMass(
     const T& mass, const Vector3<T>& position) {
-  if (mass <= 0) {
+  if (!is_positive_and_finite(mass)) {
     const std::string error_message = fmt::format(
-        "{}(): The mass of a particle is negative or zero: {}.",
+        "{}(): The mass of a particle is not positive and finite: {}.",
         __func__, mass);
     throw std::logic_error(error_message);
   }
@@ -39,17 +46,18 @@ SpatialInertia<T> SpatialInertia<T>::PointMass(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidBoxWithDensity(
     const T& density, const T& lx, const T& ly, const T& lz) {
-  if (density <= 0) {
+  if (!is_positive_and_finite(density)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid box's density is negative or zero: {}.",
+        "{}(): A solid box's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  // Ensure lx, ly, lz are positive.
-  if (lx <= 0 || ly <= 0 || lz <= 0) {
+  if (!is_positive_and_finite(lx) ||
+      !is_positive_and_finite(ly) ||
+      !is_positive_and_finite(lz)) {
     const std::string error_message = fmt::format(
-        "{}(): One or more dimensions of a solid box is negative or zero: "
-        "({}, {}, {}).", __func__, lx, ly, lz);
+        "{}(): One or more dimensions of a solid box is not positive and "
+        "finite: ({}, {}, {}).", __func__, lx, ly, lz);
     throw std::logic_error(error_message);
   }
 
@@ -61,16 +69,18 @@ SpatialInertia<T> SpatialInertia<T>::SolidBoxWithDensity(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidBoxWithMass(
     const T& mass, const T& lx, const T& ly, const T& lz) {
-  if (mass <= 0) {
+  if (!is_positive_and_finite(mass)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid box's mass is negative or zero: {}.", __func__, mass);
+        "{}(): A solid box's mass is not positive and finite: {}.",
+        __func__, mass);
     throw std::logic_error(error_message);
   }
-  // Ensure lx, ly, lz are positive.
-  if (lx <= 0 || ly <= 0 || lz <= 0) {
+  if (!is_positive_and_finite(lx) ||
+      !is_positive_and_finite(ly) ||
+      !is_positive_and_finite(lz)) {
     const std::string error_message = fmt::format(
-        "{}(): One or more dimensions of a solid box is negative or zero: "
-        "({}, {}, {}).", __func__, lx, ly, lz);
+        "{}(): One or more dimensions of a solid box is not positive and "
+        "finite: ({}, {}, {}).", __func__, lx, ly, lz);
     throw std::logic_error(error_message);
   }
   const Vector3<T> p_BoBcm_B = Vector3<T>::Zero();
@@ -81,16 +91,15 @@ SpatialInertia<T> SpatialInertia<T>::SolidBoxWithMass(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidCubeWithDensity(
     const T& density, const T& length) {
-  if (density <= 0) {
+  if (!is_positive_and_finite(density)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cube's density is negative or zero: {}.",
+        "{}(): A solid cube's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  // Ensure length is positive.
-  if (length <= 0) {
+  if (!is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): The length of a solid cube is negative or zero: {}.",
+        "{}(): The length of a solid cube is not positive and finite: {}.",
         __func__, length);
     throw std::logic_error(error_message);
   }
@@ -103,15 +112,15 @@ SpatialInertia<T> SpatialInertia<T>::SolidCubeWithDensity(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidCubeWithMass(
     const T& mass, const T& length) {
-  if (mass <= 0) {
+  if (!is_positive_and_finite(mass)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cube's mass is negative or zero: {}.", __func__, mass);
+        "{}(): A solid cube's mass is not positive and finite: {}.",
+        __func__, mass);
     throw std::logic_error(error_message);
   }
-  // Ensure length is positive.
-  if (length <= 0) {
+  if (!is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): The length of a solid cube is negative or zero: {}.",
+        "{}(): The length of a solid cube is not positive and finite: {}.",
         __func__, length);
     throw std::logic_error(error_message);
   }
@@ -125,17 +134,16 @@ template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidCapsuleWithDensity(
     const T& density, const T& radius, const T& length,
     const Vector3<T>& unit_vector) {
-  if (density <= 0) {
+  if (!is_positive_and_finite(density)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid capsule's density is negative or zero: {}.",
+        "{}(): A solid capsule's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  // Ensure radius and length are positive.
-  if (radius <= 0 || length <= 0) {
+  if (!is_positive_and_finite(radius) || !is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid capsule's radius = {} or length = {} is "
-        "negative or zero.", __func__, radius, length);
+        "{}(): A solid capsule's radius = {} or length = {} is not "
+        "positive and finite.", __func__, radius, length);
     throw std::logic_error(error_message);
   }
 
@@ -153,17 +161,16 @@ template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensity(
     const T& density, const T& radius, const T& length,
     const Vector3<T>& unit_vector) {
-  if (density <= 0) {
+  if (!is_positive_and_finite(density)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cylinder's density is negative or zero: {}.",
+        "{}(): A solid cylinder's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  // Ensure radius and length are positive.
-  if (radius <= 0 || length <= 0) {
+  if (!is_positive_and_finite(radius) || !is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cylinder's radius = {} or length = {} is "
-        "negative or zero.", __func__, radius, length);
+        "{}(): A solid cylinder's radius = {} or length = {} is not "
+        "positive and finite.", __func__, radius, length);
     throw std::logic_error(error_message);
   }
 
@@ -194,17 +201,16 @@ template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensityAboutEnd(
     const T& density, const T& radius, const T& length,
     const Vector3<T>& unit_vector) {
-  if (density <= 0) {
+  if (!is_positive_and_finite(density)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cylinder's density is negative or zero: {}.",
+        "{}(): A solid cylinder's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  // Ensure radius and length are positive.
-  if (radius <= 0 || length <= 0) {
+  if (!is_positive_and_finite(radius) || !is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): A solid cylinder's radius = {} or length = {} is "
-        "negative or zero.", __func__, radius, length);
+        "{}(): A solid cylinder's radius = {} or length = {} is not "
+        "positive and finite.", __func__, radius, length);
     throw std::logic_error(error_message);
   }
   SpatialInertia<T> M_BBcm_B =
@@ -218,11 +224,10 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensityAboutEnd(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::ThinRodWithMass(
     const T& mass, const T& length, const Vector3<T>& unit_vector) {
-  // Ensure mass and length are positive.
-  if (mass <= 0 || length <= 0) {
+  if (!is_positive_and_finite(mass) || !is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): A thin rod's mass = {} or length = {} is negative or zero.",
-        __func__, mass, length);
+        "{}(): A thin rod's mass = {} or length = {} is not positive and "
+        "finite.", __func__, mass, length);
     throw std::logic_error(error_message);
   }
 
@@ -251,11 +256,10 @@ SpatialInertia<T> SpatialInertia<T>::ThinRodWithMass(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::ThinRodWithMassAboutEnd(
     const T& mass, const T& length, const Vector3<T>& unit_vector) {
-  // Ensure mass and length are positive.
-  if (mass <= 0 || length <= 0) {
+  if (!is_positive_and_finite(mass) || !is_positive_and_finite(length)) {
     const std::string error_message = fmt::format(
-        "{}(): A thin rod's mass = {} or length = {} is negative or zero.",
-        __func__, mass, length);
+        "{}(): A thin rod's mass = {} or length = {} is not positive and "
+        "finite.", __func__, mass, length);
     throw std::logic_error(error_message);
   }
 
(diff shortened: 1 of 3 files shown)


xuchenhan-tri (commented):
Looks good to me so far.
-(release notes: none)  +(release notes: fix) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NSvgTlI0CRsFaIGlBix:bkzb9ul)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @mitiguy)

mitiguy (commented):
@xuchenhan-tri   Ready for another look.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NSwfL36DbeSWjSwYTiB:b-symz1x)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
:lgtm: 
+@SeanCurtis-TRI for Monday platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NSx1mc_BB85pWJpNbOF:b-6b4ptn)* status: LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM: With some passing style thoughts.

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NTEhFG62ENz9K5XPYFg:bpnhhn9)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEeqdxB8O4EdGXmXBk:-NTEeqdxB8O4EdGXmXBl:bjpfmye) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.h#L143)):*
> ```objc
>   /// @retval M_BBo_B B's spatial inertia about Bo, expressed in B.
>   /// @throws std::exception if any of lx, ly, lz are zero or negative
>   /// or if density is negative.
> ```

BTW It's not obvious to me, but what's the motivation for supporting zero mass or density?
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfcCS-nTxxDWVkc0u:-NTEfcCS-nTxxDWVkc0v:b2f7gha) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L42)):*
> ```cpp
> SpatialInertia<T> SpatialInertia<T>::SolidBoxWithDensity(
>     const T& density, const T& lx, const T& ly, const T& lz) {
>   // Ensure density is non-negative.
> ```

BTW Just a thought. But this comment doesn't add much to the very clear code just below it. And, however slight that redundancy may be, it gets multiplied by every instances in this file. Is it really worth it?
___
*[`multibody/tree/spatial_inertia.cc` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfwQ70BP8Lqhccz9K:-NTEfwQ70BP8Lqhccz9L:bwhwkgy) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L45)):*
> ```cpp
>   if (density < 0) {
>     const std::string error_message = fmt::format(
>         "{}(): A solid box's density is negative: {}.", __func__, density);
> ```

BTW Have you thought about how these error messages appear?

```
SolidBoxWithDensity(); A solid box's density is negative: -1.
```

"solid box" appears twice. Surely, the function name already provides the context of what is being specified? Why not simply:


```
SolidBoxWithDensity(): given density is negative: -1.
```

Think about reading compilation errors with templated code. A mountain of characters obscure the dozen or so you really want to find that tells you the problem. For that kind of reason, I tend to favor error messages without redundancy.

jwnimmer-tri:
BTW Remember that in floating point, "(x > 0)" vs "not (x <= 0)" are different conditions.  By gold-plating these functions with overly florid error handling code instead of using the simple, sufficient, and canonical `DRAKE_THROW_UNLESS(mass > 0);`, you're allowing NaNs to leak through until later.  Your habit of writing error-rejecting code by hand is really a travesty.  If you want the error message from `DRAKE_THROW_UNLESS()` to contain more details, you should fix the macro instead of copy-pasting brittle and broken code in hundreds of places.

mitiguy (commented):
Thanks Jeremy.  The verbosity of these messages is (ideally) to help user avoid (or diagnose) downstream "Delta" problems. 
@SeanCurtis-TRI  and @xuchenhan-tri  -- after passing CI, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NTH1JhG5SdCLYQNeU0w:b-x56fut)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI and @xuchenhan-tri)

___
*[`multibody/tree/spatial_inertia.h` line 143 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEeqdxB8O4EdGXmXBk:-NTGbLn-EiGrZW80GqJw:b-l8s0ck) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.h#L143)):*

BTW It's not obvious to me, but what's the motivation for supporting zero mass or density?
</blockquote></details>

Done.  For now, we will not support zero mass or density until there is a use-case.
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfcCS-nTxxDWVkc0u:-NTGbVU8Clzcom2Dj0LT:bi4ek30) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L42)):*

BTW Just a thought. But this comment doesn't add much to the very clear code just below it. And, however slight that redundancy may be, it gets multiplied by every instances in this file. Is it really worth it?
</blockquote></details>

Done.  Removed, here and elsewhere.
___
*[`multibody/tree/spatial_inertia.cc` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfwQ70BP8Lqhccz9K:-NTGc2yeCoXasylE8HDc:b-bi02gn) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L45)):*

BTW Have you thought about how these error messages appear?

```
SolidBoxWithDensity(); A solid box's density is negative: -1.
```

"solid box" appears twice. Surely, the function name already provides the context of what is being specified? Why not simply:


```
SolidBoxWithDensity(): given density is negative: -1.
```

Think about reading compilation errors with templated code. A mountain of characters obscure the dozen or so you really want to find that tells you the problem. For that kind of reason, I tend to favor error messages without redundancy.
</blockquote></details>

I prefer it as is.  For this example, the difference is only 8 characters.  For some other examples it is slightly more (but not many).
FYI/tangent: I sometimes uses a grep-like search to of my C++ code to find an error message.  Replacing the object's name with "given" provides a more non-unique search result.

jwnimmer-tri:
My main point is that the **software implementation tactics** in evidence here for printing more descriptive messages are really, really bad.  The cost in lines of code to test and maintain is too high, and allows for too many typos the same boilerplate is copy-pasted one place to the next.  It's not a bad goal, it's just bad software.  If we want better messages, we should create better technology to emit them.  Writing 5+ lines of code when 1 should suffice is a failure to identify and create appropriate software abstractions.  There is nothing stopping us from having a 1-line macro call that conveys all of the same information as your copy-pasted code is doing here.

SeanCurtis-TRI (commented):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 142 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJp8Aw5YrOH5HapkKM:-NTJp8Aw5YrOH5HapkKN:beohakp) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.h#L142)):*
> ```objc
>   /// @param[in] lz length of the box in the Bz direction (meters).
>   /// @retval M_BBo_B B's spatial inertia about Bo, expressed in B.
>   /// @throws std::exception if density ≤ 0 or any of lx, ly, lz are zero or
> ```

BTW This looks like it's been through multiple independent editors.

"Density <= 0 but lx zero or negative?"

Strange to express the same idea two different ways in the same sentence. And that happens in all of these instances.

-------------------------------------------------

And to Jeremy's point, this is trying to craft a shopping list of values of density that *could* be bad: zero, negative. However, it's not a *complete* list (i.e. density = NaN is not included). In these cases, it's easier to say (and test) what is required:

```c++
/// @throws std::exception if density, lx, ly, or lz is not finite and positive.
```

And then the test in the code becomes:

```c++
if (!(std::isfinite(value) && value > 0)) {
  // Error message.
}
```

I'd emphasize, this is about how the condition is *phrased*. The second question is what mechanism is used to *effectively* give feedback to the user. That's orthogonal.
___
*[`multibody/tree/spatial_inertia.cc` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfwQ70BP8Lqhccz9K:-NTJrYYABbIZIdxQ7bSg:b1fbjrc) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L45)):*

I prefer it as is.  For this example, the difference is only 8 characters.  For some other examples it is slightly more (but not many).
FYI/tangent: I sometimes uses a grep-like search to of my C++ code to find an error message.  Replacing the object's name with "given" provides a more non-unique search result.
</blockquote></details>

re: tangent.  Surely, the string  "SolidBoxWithDensity()" already tells you everything you need to know regarding where to look in the code. If it doesn't, why print the function name at all?
___
*[`multibody/tree/spatial_inertia.cc` line 21 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJvZDr2BlbvAl7CM5b:-NTJvZDr2BlbvAl7CM5c:b-27fyyu) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.cc#L21)):*
BTW Continuing the theme.

These *six* lines of code produce an error message that looks like this:

> SolidBoxWithDensity(): A solid box's density is negative or zero: -1.

Alternatively, this *one* line of code:

```c++
DRAKE_THROW_UNLESS(density > 0);
```

Produces this error message:

> Failure at multibody/tree/spatial_inertia.cc:42 in SolidBoxWithDensity(): condition 'density > 0' failed.

As I compare/contrast the two error messages:

1. The former includes the offending value.
2. The latter includes file and line number.
3. The latter is uniformly phrased throughout all of Drake (everywhere `DRAKE_THROW_UNLESS` is used) so it leverages consistency.

The *only* lost piece of information between the different messages is the failing value of the tested quantity. That said, I'm generally a fan of reporting the offending value. I believe that for easy errors, just seeing the value can tell you what you've done ("Oops! That's an anachronistic initialization value!")

What we'd like to work towards is something like:

```c++
DRAKE_THROW_UNLESS_VALUE(density > 0, density);
```

Which would produce a message like:

> Failure at multibody/tree/spatial_inertia.cc:42 in SolidBoxWithDensity(): condition 'density > 0' failed. density = -1.

Paul, would you find the resulting error message sufficiently compelling that you'd transition over to it? And, if not, what's lacking?
___
*[`multibody/tree/spatial_inertia.cc` line 66 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJsAP3CtSL4HbdQw1Y:-NTJsAP3CtSL4HbdQw1Z:bsctzby) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.cc#L66)):*
> ```cpp
>   if (mass <= 0) {
>     const std::string error_message = fmt::format(
>         "{}(): A solid box's mass is negative or zero: {}.", __func__, mass);
> ```

btw: Continuing the theme of proper conditioning of the failure:

"mass isn't positive and finite". or "mass must be positive and finite".

xuchenhan-tri (commented):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI and @xuchenhan-tri)

___
*[`multibody/tree/spatial_inertia.h` line 142 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJp8Aw5YrOH5HapkKM:-NTMEp5J-3LP0KriDtUO:bhbvg6w) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.h#L142)):*

BTW This looks like it's been through multiple independent editors.

"Density <= 0 but lx zero or negative?"

Strange to express the same idea two different ways in the same sentence. And that happens in all of these instances.

-------------------------------------------------

And to Jeremy's point, this is trying to craft a shopping list of values of density that *could* be bad: zero, negative. However, it's not a *complete* list (i.e. density = NaN is not included). In these cases, it's easier to say (and test) what is required:

```c++
/// @throws std::exception if density, lx, ly, or lz is not finite and positive.
```

And then the test in the code becomes:

```c++
if (!(std::isfinite(value) && value > 0)) {
  // Error message.
}
```

I'd emphasize, this is about how the condition is *phrased*. The second question is what mechanism is used to *effectively* give feedback to the user. That's orthogonal.
</blockquote></details>

Done here and elsewhere -- with local file static function.
Note: I'll change static to anonymous function on next commit.
___
*[`multibody/tree/spatial_inertia.cc` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTEfwQ70BP8Lqhccz9K:-NTMF4Rb1TaxbrhsyYh7:b-mev0br) ([raw file](https://github.com/RobotLocomotion/drake/blob/70649ada3ab5a71a94300fa26fb0c463ae20d049/multibody/tree/spatial_inertia.cc#L45)):*

re: tangent.  Surely, the string  "SolidBoxWithDensity()" already tells you everything you need to know regarding where to look in the code. If it doesn't, why print the function name at all?
</blockquote></details>

From what I understand from our conversation, the pending new macro(s) may resolve this.  For now, I prefer it as is.
___
*[`multibody/tree/spatial_inertia.cc` line 21 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJvZDr2BlbvAl7CM5b:-NTMEync06kmtRAmMbOn:b-ku3rpz) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.cc#L21)):*

BTW Continuing the theme.

These *six* lines of code produce an error message that looks like this:

> SolidBoxWithDensity(): A solid box's density is negative or zero: -1.

Alternatively, this *one* line of code:

```c++
DRAKE_THROW_UNLESS(density > 0);
```

Produces this error message:

> Failure at multibody/tree/spatial_inertia.cc:42 in SolidBoxWithDensity(): condition 'density > 0' failed.

As I compare/contrast the two error messages:

1. The former includes the offending value.
2. The latter includes file and line number.
3. The latter is uniformly phrased throughout all of Drake (everywhere `DRAKE_THROW_UNLESS` is used) so it leverages consistency.

The *only* lost piece of information between the different messages is the failing value of the tested quantity. That said, I'm generally a fan of reporting the offending value. I believe that for easy errors, just seeing the value can tell you what you've done ("Oops! That's an anachronistic initialization value!")

What we'd like to work towards is something like:

```c++
DRAKE_THROW_UNLESS_VALUE(density > 0, density);
```

Which would produce a message like:

> Failure at multibody/tree/spatial_inertia.cc:42 in SolidBoxWithDensity(): condition 'density > 0' failed. density = -1.

Paul, would you find the resulting error message sufficiently compelling that you'd transition over to it? And, if not, what's lacking?
</blockquote></details>

Thanks Sean -- I look forward to trying out your new macro(s).
___
*[`multibody/tree/spatial_inertia.cc` line 66 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJsAP3CtSL4HbdQw1Y:-NTMLrwk7rIO0pym_hR0:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.cc#L66)):*

btw: Continuing the theme of proper conditioning of the failure:

"mass isn't positive and finite". or "mass must be positive and finite".
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
For the record, we've discussed the state of the exception-throwing cruft and have agreed to a plan.

I'm going to add a macro that'll have the efficiency of `DRAKE_THROW_UNLESS` (in terms of lines of code at the call site) in a follow-up PR. At that point, Paul will revisit the bespoke throwing machinery to make use of the new macro. For now, we'll leave the heavy machinery for two reasons:

- In the interim, the bad values are printed (which is good).
- The verbose spellings will clearly shout out for replacement. It's far less obvious to recognize that a `DRAKE_THROW_UNLESS` needs upgrading.
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NTOi27nDClAWf0ISKac:b-koomju)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 142 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTJp8Aw5YrOH5HapkKM:-NTOifEz5shBH88ynKKh:b-f9gy51) ([raw file](https://github.com/RobotLocomotion/drake/blob/b0945e3099c3ad3fb87ea4de4dccb418cfbe0dd7/multibody/tree/spatial_inertia.h#L142)):*

Done here and elsewhere -- with local file static function.
Note: I'll change static to anonymous function on next commit.

</blockquote></details>

BTW One last thought for future reference.

In all of these functions, we have a simple rule: "all parameters must be finite an positive". Rather than verbosely enumerating the parameters, we could simply directly say that.

```c++
/// @throws std::exception if any parameter is not positive and finite.
```

And then you'd have some really nice common boilerplate that looks the same from method to method.

(Note: I'm not blocking on this, just wanted to share the thought.)
___
*[`multibody/tree/spatial_inertia.cc` line 12 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTOj6vt0mEwqZsjPI_a:-NTOj6vt0mEwqZsjPI_b:b-l4i86u) ([raw file](https://github.com/RobotLocomotion/drake/blob/f08cc21a30c7b23c7d20bd2c00cd80b17c741c03/multibody/tree/spatial_inertia.cc#L12)):*
> ```cpp
> 
> template <typename T>
> static const boolean<T> is_positive_and_finite(const T& value) {
> ```

BTW I know we've had this conversation before. It's more common in Drake to use an anonymous namespace (as you have commented out above) instead of `static` to arrange internal linkage for a symbol.

mitiguy (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-:-NTPlKvF9qRcvVaqo3Jb:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.cc` line 12 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTOj6vt0mEwqZsjPI_a:-NTPkOi17ZZ4OYbL43uH:b-bfshgz) ([raw file](https://github.com/RobotLocomotion/drake/blob/f08cc21a30c7b23c7d20bd2c00cd80b17c741c03/multibody/tree/spatial_inertia.cc#L12)):*

BTW I know we've had this conversation before. It's more common in Drake to use an anonymous namespace (as you have commented out above) instead of `static` to arrange internal linkage for a symbol.
</blockquote></details>

Done.  I am thinking about moving this into a more common header file in another PR.  If this is rejected, I'll go back and fix.

SeanCurtis-TRI:
*[`multibody/tree/spatial_inertia.cc` line 12 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTOj6vt0mEwqZsjPI_a:-NTPnEoG5-f1nm514aiV:b-oqdsti) ([raw file](https://github.com/RobotLocomotion/drake/blob/f08cc21a30c7b23c7d20bd2c00cd80b17c741c03/multibody/tree/spatial_inertia.cc#L12)):*

Done.  I am thinking about moving this into a more common header file in another PR.  If this is rejected, I'll go back and fix.
</blockquote></details>

FYI If you say "Done" it implies that you accepted the suggestion. In this case, you didn't. Instead, you left a piece of zombie comment  in place (`// namespace {`).

jwnimmer-tri:
*[`multibody/tree/spatial_inertia.cc` line 12 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19192#-NTOj6vt0mEwqZsjPI_a:-NTPnceJ00lG5UT_FVt0:b7ivh71) ([raw file](https://github.com/RobotLocomotion/drake/blob/f08cc21a30c7b23c7d20bd2c00cd80b17c741c03/multibody/tree/spatial_inertia.cc#L12)):*

FYI If you say "Done" it implies that you accepted the suggestion. In this case, you didn't. Instead, you left a piece of zombie comment  in place (`// namespace {`).
</blockquote></details>

`is_positive_and_finite` is unsuitable for a project-wide header.  It looks like an abstraction but it doesn't actually abstract anything.  Functions are for abstraction.  They are not for changing copy-pasted C++ syntax into equivalent copy-pasted function name sentences.  `&&` is the canonical way to spell `and` in C++.  Saying `and` is strictly worse.  Likewise for `isfinite` is the canonical spelling for `is_..._finite`.  Repaving nature C++ code with extremely-special-purpose prose that is identical is bad style.

mitiguy:
Sean -- As we discussed f2f earlier this morning. I'll replace the "static" with an anonymous namespace and rename the function from is_positive_and_finite() to something such as is_positive_finite() or ispositivefinite() -- whatever you and Jeremy regard as the best name for its purpose.  I'll be starting this work tomorrow with a next PR.

Also, from what I understood from our conversation today, I thought you used something like is_positive_finite() in another file.  Hmmm.

mitiguy:
Agreed.  Since it needs to be templatized on T, it is advantageous that it is its own function.

jwnimmer-tri:
I suppose it depends on how you define "nice".  For me, "nice" means syntax that anyone passing by would recognize at a glance, without needing to jump to somewhere else in the code to read a one-line subroutine:

```c++
using std::isfinite;
DRAKE_THROW_UNLESS(isfinite(mass) && mass > 0, mass);
```

PR #19193, opened 2023-04-12
[multibody] Fix a bug in contact solver results computation

### LINKED ISSUE #18682: [contact_manager] Recursive-ness in discrete updates?
### What happened?

I am writing controllers that rely on cascades of discrete updates, because while they can be weird, relying on good implementations of this can make things a nicer place in the real robot world.

Per https://github.com/RobotLocomotion/drake/pull/18551#issuecomment-1384319905, I am using multiple initializations to arrive at "steady state" for discrete state.

I naively expected that if I saw the exact same `(q, v, tau)` for a plant, then I wouldn't have to worry about any other state.

However, it seems like there is extra state, possibly in the contact manager, that will "accumulate" across my calls.

I've added the example to #18235, in 9a475540c5:
https://github.com/RobotLocomotion/drake/blob/9a475540c5d844c2b5b2d5b37eca785f0f758882/examples/instability/test/plant_hidden_state_test.py#L257-L279

See output:
https://gist.github.com/EricCousineau-TRI/c8c5ebf28c6c36acdd8c70727848674d

I can workaround by using the following hack:
https://github.com/RobotLocomotion/drake/blob/9a475540c5d844c2b5b2d5b37eca785f0f758882/examples/instability/test/plant_hidden_state_test.py#L167-L170

\cc @xuchenhan-tri @DamrongGuoy w.r.t. other issues (#18236, #18241)
\cc @jwnimmer-tri @sherm1 w.r.t. #18551, #12649

FYI @dmcconachie-tri When we did our rubber duck session last week, I was afraid of stuff like this haha

### Version

d2c7504c49f7fee1cecb1fe82d21c11420b3d9f1

Ubuntu 22.04

compiled from source code using Bazel

### Relevant log output

_No response_

xuchenhan-tri:
Closes #18682.

ContactSolverResults should depend on actuation input.

### DIFF
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -1074,9 +1074,11 @@ drake_cc_googletest(
     name = "discrete_update_manager_test",
     deps = [
         ":dummy_model",
+        ":multibody_plant_config_functions",
         ":multibody_plant_core",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
+        "//multibody/parsing",
         "//systems/analysis:simulator",
         "//systems/framework:abstract_value_cloner",
     ],
--- a/multibody/plant/discrete_update_manager.cc
+++ b/multibody/plant/discrete_update_manager.cc
@@ -37,7 +37,8 @@ void DiscreteUpdateManager<T>::DeclareCacheEntries() {
       systems::ValueProducer(
           this, &DiscreteUpdateManager<T>::CalcContactSolverResults),
       {systems::System<T>::xd_ticket(),
-       systems::System<T>::all_parameters_ticket()});
+       systems::System<T>::all_parameters_ticket(),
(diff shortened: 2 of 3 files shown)


xuchenhan-tri:
*[`multibody/plant/test/discrete_update_manager_test.cc` line 333 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19193#-NSwoKgC6Dt3xBUloG-V:-NSwoKgC6Dt3xBUloG-W:b6ulgl5) ([raw file](https://github.com/RobotLocomotion/drake/blob/f6731762de33996182af252b547ec1452d5ed6cd/multibody/plant/test/discrete_update_manager_test.cc#L333)):*
> ```cpp
>   const double nonzero_actuation = 1.2;
>   u_input.FixValue(&plant_context, Vector1<double>(nonzero_actuation));
>   simulator.AdvanceTo(dt);
> ```

This should be 2*dt.

PR #19194, opened 2023-04-12
[prototype] RenderEngineVtk uses mtl file to get diffuse properties

SeanCurtis-TRI:
First stage, can I simply replace the old geoemtry with new geometry that has the same effect?

1. I need to convert general OBJ expression into buffer representation.
    - RenderEngineGl has code that does this, I'm stealing it outright.
2. Replace vtkOBJReader with this new function that returns a poly algorithm.
    - Tests pass.
3. We'll cheat for dealing with the materials.
     - Given the current logic, we'll simply stomp on the perception properties. Not a good solution, but this merely proves that the data *can* flow from tiny obj to vtk.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19194)

### DIFF
--- a/geometry/render/BUILD.bazel
+++ b/geometry/render/BUILD.bazel
@@ -16,6 +16,7 @@ drake_cc_package_library(
         ":render_camera",
         ":render_engine",
         ":render_label",
+        ":render_mesh",
     ],
 )
 
@@ -59,6 +60,18 @@ drake_cc_library(
         "//common:essential",
         "//common:hash",
         "//systems/sensors:image",
+    ],
+)
+
+drake_cc_library(
+    name = "render_mesh",
+    srcs = ["render_mesh.cc"],
+    hdrs = ["render_mesh.h"],
+    interface_deps = [
+        "@tinyobjloader",
+    ],
+    deps = [
+        "//common:essential",
     ],
 )
 
--- a/geometry/render/BUILD.bazel
+++ b/geometry/render/BUILD.bazel
@@ -68,10 +68,12 @@ drake_cc_library(
     srcs = ["render_mesh.cc"],
     hdrs = ["render_mesh.h"],
     interface_deps = [
-        "@tinyobjloader",
+        "//common:diagnostic_policy",
+        "//common:essential",
+        "//geometry:rgba",
     ],
     deps = [
-        "//common:essential",
+        "@tinyobjloader",
     ],
 )
 
@@ -114,6 +116,18 @@ drake_cc_googletest(
         ":render_label",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
+    name = "render_mesh_test",
+    data = [":test_models"],
+    deps = [
+        ":render_mesh",
+        "//common:temp_directory",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+        "//multibody/parsing:diagnostic_policy_test_base",
     ],
 )
 
--- a/geometry/render/render_mesh.cc
+++ b/geometry/render/render_mesh.cc
@@ -1,9 +1,15 @@
 #include "drake/geometry/render/render_mesh.h"
 
+#include <algorithm>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
 #include <map>
+#include <set>
 #include <tuple>
 
 #include <fmt/format.h>
+#include <tiny_obj_loader.h>
 
 #include "drake/common/drake_assert.h"
 #include "drake/common/eigen_types.h"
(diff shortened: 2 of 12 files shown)


SeanCurtis-TRI (commented):
cc @jwnimmer-tri @zachfang 

Here's the prototype. Both diffuse color (Kd) and the diffuse texture map (map_Kd) get from the .mtl file to the VTK render engine. It's predicated upon some copy pasta and some hacks (as documented above). Putting aside the ugliness of the *implementation*, we can focus on the availability of data as provided by tiny obj.

The data is solid, with a couple of caveats -- things I'd like to change about tiny obj if I could:

- Image paths: If image names are relative, they are relative to the **.mtl** file and *not* the **.obj** file. While it's typical for them to be collocated there is no requirement. The code currently assumes they are collocated. Modifying tiny obj loader so that it can tell the user the path to the mtl file it read would resolve things nicely (and be largely harmless).
- tiny_obj_loader default values: like SDFomat parser, tiny_obj_loader has defaults for *everything*. That means, once its been parsed into memory, it becomes impossible, in some cases, to tell if a value has been *set* by the author of the mtl file or by the parser. This is likewise regrettable. Changes in this regard would be *much* bigger.

Something that doesn't need *changing* but would need to be clarified in the future:

- There are some semantics of the parser of which I'm not entirely clear.
  - I assume that "shapes" are based on the "g" or "o" tags, and within a single shape, triangles can have an arbitrary material.  So, even in the presence of a single shape, we'd still end up with multiple materials. We'll have to think carefully about the definition of what Drake will support/ignore/reject and how that definition will evolve with time.

If you run `bazel test //geometry/render_vtk:internal_render_engine_vtk_test` the "mesh tests" will run for two reasons:
  - the change I made to the box.obj.mtl is no longer compatible with what those meshes expect
  - RenderEngineVtkTest.MeshTest must fail because its semantics are no longer supported.

Take a look at the comments in `internal_render_engine_vtk_test` for some thoughts on why the old tests don't work because this constitutes a change of semantics (i.e. MTL wins vs `PerceptionProperties` win).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19194#-:-NSrZQwmAMdOBHkdmS9r:b-77g8c)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI:
*[`geometry/render/test/meshes/box.obj.mtl` line 10 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19194#-NSrcH1v2VUQeMhx4Zhm:-NSrcH1v2VUQeMhx4Zhn:b-6kx73w) ([raw file](https://github.com/RobotLocomotion/drake/blob/ef1e8936b8664cac41a38ab8a31093d9e8ae707e/geometry/render/test/meshes/box.obj.mtl#L10)):*
> ```
> # N.B. The next line has trailing whitespace on purpose, to prove that Drake
> # ignores it during parsing.
> map_Kd -s 1 1 1 ../diag_gradient.png 
> ```

BTW The trailing space shows that tiny obj strips whitespace.  :)

SeanCurtis-TRI (commented):
Per F2F:

- In this *interim* step, we're going to have the following material semantics when accounting for mtl files:
    - If a single material defined in the .mtl is applied to *all* triangles and the *specified* diffuse texture (map_Kd) is readable, we'll use the mtl material.
    - Otherwise, we fall back to the current behavior (with appropriate warnings as to why).
- The copypasta'd `MeshData` is going to evolve to store the material associated with the mesh.
- The VTK implementation will get cleaned up to remove the hacks and get new tests.

What *won't* be addressed in the short-term (but will possibly be looked at in parallel):
  - Modifying tiny_obj_loader to return the path to the mtl file that was read.
  - Modifying tiny_obj_loader so users can distinguish between values that were defaulted in their absence vs set by the parser.
  - Modulation of the .mtl material with ("phong", "diffuse") -- e.g.., given a generic "box" with a generic box-like texture, a user can create multiple, varied iterations by modulating the texture color with non-white rgb values. In this first pass, it'll be *either* the .mtl wins *or* the geometry properties.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19194#-:-NSvKyq6CKwOpcN3yvG_:bhehc6j)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

SeanCurtis-TRI (commented):
Agreed.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19194#-:-NYQjKs97e7ipPMPHE7b:b-pbdqz3)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

Issue #19164, opened 2023-04-10
Improve default names for automatic floating base joints


RussTedrake:
See extended discussion in [slack](https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1681051422016589).

The choice to use e.g. `$world_body1` was made [during code review](https://github.com/RobotLocomotion/drake/pull/18390#pullrequestreview-1205100703), but this choice did not consider all use cases.  People writing optimization problems who reference the joint positions by names in constraints don't want something that verbose.  Python users using `namedview` can't eat the `$` in the name.

Request: the default names should be something more like `{"_"*N}{body_name}` for the smallest N that works. 

Although we don't have proper depreciation, the slack message includes some relevant ideas from @jwnimmer-tri .

Related to #19162.

sherm1:
Q: if the system consists of four model instances of the same humanoid (say), we'll add floating joints to the torsos. The proposed scheme makes those joints `torso`, `_torso`, `__torso`, and `___torso` with no clear mapping between name and model instance. Is that OK? We could instead guarantee to add underscores in order of model instance index -- is that worthwhile?

sherm1:
A single model instance can include multiple floating bodies -- naming the joints for the bodies as in Russ's proposal seems like the natural approach and least likely to require disambiguation.

sherm1:
Ah, sorry. I forgot that we can have the same names in multiple model instances. So the only disambiguation we need is if the model contains both a floating body named "xyz" and an _unrelated_ joint named "xyz" in the same model instance (very unlikely!). In that case we would need to make body xyz's joint be "_xyz" to avoid a joint name conflict.

jwnimmer-tri:
Fair enough.  I'm happy to allow whoever implements this to think carefully about the naming choices and back it up with unit tests.

PR #19196, opened 2023-04-13
Provide automatic sanitization of field names in `namedview`

RussTedrake:
Related to #19164 -- MbP uses names which are not immediately useable as `namedview` fields.

After this PR (and now that #19174 has landed), we can do e.g.
```
PositionView = namedview("PositionView", plant.GetPositionNames())
StateView = namedview("StateView", plant.GetStateNames())
```

+@EricCousineau-TRI for feature review, please?  WDYT?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196)

### DIFF
--- a/bindings/pydrake/common/containers.py
+++ b/bindings/pydrake/common/containers.py
@@ -150,13 +150,19 @@ class NamedViewBase:
         return cls([0] * len(cls._fields))
 
 
-def _sanitize_field_name(x: str):
-    if x[0].isnumeric():
-        x = '_' + x
-    x = re.sub('[^a-zA-Z0-9_]', '_', x)
-    x = re.sub('__+', '_', x)
-    assert x.isidentifier(), f"Sanitization failed on {x}"
-    return x
+def _sanitize_field_name(name: str):
+    result = name
+    # Ensure the first character is a valid opener (e.g., no numbers allowed).
+    if not result[0].isidentifier():
+        result = "_" + result
+    # Ensure that each additional character is valid in turn, avoiding the
+    # special case for opening characters by prepending "_" during the check.
+    for i in range(1, len(result)):
+        if not ("_" + result[i]).isidentifier():
+            result = result[:i] + "_" + result[i+1:]
+    result = re.sub("__+", "_", result)
+    assert result.isidentifier(), f"Sanitization failed on {name} => {result}"
+    return result
 
 
 def namedview(name, fields, *, sanitize_field_names=True):
--- a/bindings/pydrake/common/containers.py
+++ b/bindings/pydrake/common/containers.py
@@ -172,9 +172,9 @@ def namedview(name, fields, *, sanitize_field_names=True):
     Similar to ``namedtuple``.
 
     If ``sanitize_field_names`` is True (the default), then any characters in
-    ``fields`` which are not valid in Python identifier (A-z, 0-9, and _ )
-    will be automatically replaced with `_`. Leading numbers will have `_`
-    inserted, and duplicate `_` will be replaced by a single `_`.
+    ``fields`` which are not valid in Python identifiers will be automatically
+    replaced with `_`. Leading numbers will have `_` inserted, and duplicate
+    `_` will be replaced by a single `_`.
 
     Example:
         ::
--- a/bindings/pydrake/common/containers.py
+++ b/bindings/pydrake/common/containers.py
@@ -1,6 +1,7 @@
 """Provides extensions for containers of Drake-related objects."""
 
 import numpy as np
+import re
 
 
 class _EqualityProxyBase:
@@ -71,7 +72,6 @@ class EqualToDict(_DictKeyWrap):
     `lhs.EqualTo(rhs)`.
     """
     def __init__(self, *args, **kwargs):
-
         class Proxy(_EqualityProxyBase):
             def __eq__(self, other):
                 T = type(self.value)
@@ -118,8 +118,7 @@ class NamedViewBase:
         if not hasattr(self, name):
             raise AttributeError(
                 "Cannot add attributes! The fields in this named view are"
-                f"{self.get_fields()}, but you tried to set '{name}'."
-            )
+                f"{self.get_fields()}, but you tried to set '{name}'.")
         object.__setattr__(self, name, value)
 
     def __len__(self):
@@ -142,21 +141,34 @@ class NamedViewBase:
     @staticmethod
     def _item_property(i):
         # Maps an item (at a given index) to a property.
-        return property(
-            fget=lambda self: self[i],
-            fset=lambda self, value: self.__setitem__(i, value))
+        return property(fget=lambda self: self[i],
+                        fset=lambda self, value: self.__setitem__(i, value))
 
     @classmethod
     def Zero(cls):
         """Constructs a view onto values set to all zeros."""
-        return cls([0]*len(cls._fields))
+        return cls([0] * len(cls._fields))
 
 
-def namedview(name, fields):
+def _sanitize_field_name(x: str):
+    if x[0].isnumeric():
+        x = '_' + x
+    x = re.sub('[^a-zA-Z0-9_]', '_', x)
+    x = re.sub('__+', '_', x)
+    assert x.isidentifier(), f"Sanitization failed on {x}"
+    return x
+
+
+def namedview(name, fields, *, sanitize_field_names=True):
     """
     Creates a class that is a named view with given ``fields``. When the class
     is instantiated, it must be given the object that it will be a proxy for.
     Similar to ``namedtuple``.
+
+    If ``sanitize_field_names`` is True (the default), then any characters in
+    ``fields`` which are not valid in Python identifier (A-z, 0-9, and _ )
+    will be automatically replaced with `_`. Leading numbers will have `_`
+    inserted, and duplicate `_` will be replaced by a single `_`.
 
     Example:
         ::
@@ -186,7 +198,10 @@ def namedview(name, fields):
 
     For more details, see ``NamedViewBase``.
     """
-    base_cls = (NamedViewBase,)
+    base_cls = (NamedViewBase, )
+    if sanitize_field_names:
+        fields = [_sanitize_field_name(f) for f in fields]
+    assert len(set(fields)) == len(fields), "Field names must be unique"
     type_dict = dict(_fields=tuple(fields))
     for i, field in enumerate(fields):
         type_dict[field] = NamedViewBase._item_property(i)
--- a/bindings/pydrake/common/test/containers_test.py
+++ b/bindings/pydrake/common/test/containers_test.py
@@ -130,3 +130,25 @@ class TestNamedView(unittest.TestCase):
         self.assertEqual(len(view), 2)
         self.assertEqual(view.a, 0)
         self.assertEqual(view.b, 0)
+
+    def test_name_sanitation(self):
+        MyView = namedview("MyView",
+                           ["$world_base", "iiwa::iiwa", "no spaces", "2var"])
+        self.assertEqual(MyView.get_fields(),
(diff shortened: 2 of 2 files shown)


EricCousineau-TRI (commented):
:lgtm_strong: on code and testing
However, I'm not sure if perhaps there's another mapping that could be used. (I can't think of one, though)

@jwnimmer-tri Can I ask your opinion, esp. re: your experience remapping for Python template types?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-:-NSsRb7B6L-EKR0h2Y-c:b-n30ta9)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`bindings/pydrake/common/containers.py` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsRJsUBNIrShPtJjtq:-NSsRJsUBNIrShPtJjtr:b-fo0qp0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L156)):*
> ```Python
> 
> 
> def namedview(name, fields, sanitize_field_names=True):
> ```

nit Boolean argument may serve better as kwarg-only.

Consider making it `(name, fields, *, sanitize_field_names=True)`
___
*[`bindings/pydrake/common/containers.py` line 183 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsUjjf0mupZE9bUvSb:-NSsUjjf0mupZE9bUvSc:b-ku0j4l) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L183)):*
> ```Python
>             # Get an array from the view that is a *copy* of the original
>             # vector.
>             value_copy = np.array(view)
> ```

nit Consider adding example from testing?
___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsT_8F36DOc9aBR3dd:-NSsT_8F36DOc9aBR3de:brdah8h) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*
> ```Python
>     base_cls = (NamedViewBase,)
>     if sanitize_field_names:
>         fields = [re.sub('[^a-zA-Z0-9_]', '_', f) for f in fields]
> ```

There may be non-uniqueness for this remapping.
e.g. `my:field` and `my.field` should remap to `my_field`, which could collide with an already specified `my_field`.

Can you assert that the number of unique fields hasn't changed after the mapping?

Ideal is that we assert all fields are unique after the `if` and before `type_dict = ...` line, which can effectively protected against that, and non-unique fields in general.
This is more correct IMO, but it could cause an explicit error where there wasn't one before.

jwnimmer-tri (commented):
> However, I'm not sure if perhaps there's another mapping that could be used. (I can't think of one, though)
> Can I ask your opinion, esp. re: your experience remapping for Python template types?

I think demoting non-identifier characters to underscores (as we have here) is probably best.

Another alternative would be to mangle them into special _unicode_ characters like 𝒞 for "colon" or something, but that would defeat our purpose here of users typing these names in by hand into their own code to get & set the positions. The template mangling I did could use unicode only because the user never had to key in those names; they only needed to read the names on their screen in crash dumps and not be _too_ confused.

The one suggestion I would made is to de-duplicate the underscores.  Any run of 1 or more invalid characters could perhaps be replaced with a single `_`.  I'm not sure that keeping the repeated underscore buys is anything. You could even imagine folding any pre-existing underscores into the replacement underscores as well, i.e., doing a final regex`s/__+/_/g;` as the last renaming step.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-:-NSsVsoB0EfHQEkgP9mO:breq8xd)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsW23O9bAspn-1fnOE:-NSsW23O9bAspn-1fnOF:b-tsu9hp) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*
> ```Python
>     base_cls = (NamedViewBase,)
>     if sanitize_field_names:
>         fields = [re.sub('[^a-zA-Z0-9_]', '_', f) for f in fields]
> ```

nit This regex is not the correct validator for valid python identifiers. Use https://docs.python.org/3/library/stdtypes.html#str.isidentifier instead. Something like this:

```python
def _make_valid_identifier(x : str):
    assert len(x) > 0
    # Ensure the first character is a valid opener.
    if not x[0].isidentifier():
        x = '_' + x
    # Ensure that each additional character is valid in turn, starting from
    # the prior character to avoid the special rules for a first character.
    for i in range(1, len(x))
        if not x[i-1:i+1].isidentifier():
            x[i] = '_'
    assert x.isidentifier()
    return x
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsW23O9bAspn-1fnOE:-NSsa7rVC2qhaXJysmGp:b-z9kwd3) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*

nit This regex is not the correct validator for valid python identifiers. Use https://docs.python.org/3/library/stdtypes.html#str.isidentifier instead. Something like this:

```python
def _make_valid_identifier(x : str):
    assert len(x) > 0
    # Ensure the first character is a valid opener.
    if not x[0].isidentifier():
        x = '_' + x
    # Ensure that each additional character is valid in turn, starting from
    # the prior character to avoid the special rules for a first character.
    for i in range(1, len(x))
        if not x[i-1:i+1].isidentifier():
            x[i] = '_'
    assert x.isidentifier()
    return x
```
</blockquote></details>

To be clear, the main difference for unicode.  If someone names their model instance ロボット we should preserve that here, not rewrite it to `____`.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/containers.py` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsRJsUBNIrShPtJjtq:-NSuGE3Y2sq8yTFzcNel:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L156)):*

nit Boolean argument may serve better as kwarg-only.

Consider making it `(name, fields, *, sanitize_field_names=True)`
</blockquote></details>

Done.
___
*[`bindings/pydrake/common/containers.py` line 183 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsUjjf0mupZE9bUvSb:-NSu1OBcDV6MsJIx-q6F:bkqtfm1) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L183)):*

nit Consider adding example from testing?
</blockquote></details>

Done.  (I think; if what you meant is that i should test the values in my existing test).
___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsW23O9bAspn-1fnOE:-NSuA72eEMDcpgicOXwY:b-o0xl2b) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*

To be clear, the main difference for unicode.  If someone names their model instance ロボット we should preserve that here, not rewrite it to `____`.
</blockquote></details>

Done.  ?  The `_make_valid_identifier` method you provided doesn't work... (str objects don't support item assignment), and making it work gets ugly fast.  All to do what I believe is exactly the same as what I had in a one-liner.  I've added an assert instead, but am happy to take a working suggestion you like better.
___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsT_8F36DOc9aBR3dd:-NSuBiG182PhQ2Axw8lp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*

There may be non-uniqueness for this remapping.
e.g. `my:field` and `my.field` should remap to `my_field`, which could collide with an already specified `my_field`.

Can you assert that the number of unique fields hasn't changed after the mapping?

Ideal is that we assert all fields are unique after the `if` and before `type_dict = ...` line, which can effectively protected against that, and non-unique fields in general.
This is more correct IMO, but it could cause an explicit error where there wasn't one before.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196)* status: 1 unresolved discussion, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsW23O9bAspn-1fnOE:-NSuqZL-CENtwrCrbQi_:bnli365) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*

Okay, I'll develop something that works.
</blockquote></details>

Done (pushed).

jwnimmer-tri:

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196)* status: LGTM missing from assignee sammy-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI and @sammy-tri)

___
*[`bindings/pydrake/common/containers.py` line 196 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-NSsW23O9bAspn-1fnOE:-NSv9aq4BdM9NqfgqdVm:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2dab37ae1c3227ea0bba3ce0f0d059bc4bd52d4/bindings/pydrake/common/containers.py#L196)):*

Done (pushed).
</blockquote></details>

Thanks!

sammy-tri (commented):
:lgtm: +(status: squashing now) 

> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19196#-:-NSvdOejCofxYd1YcIW_:b1pnxgs)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),EricCousineau-TRI(platform) (waiting on @RussTedrake)

PR #19201, opened 2023-04-14
Cspace free path setup program

AlexandreAmice:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)

### DIFF
--- a/common/polynomial.cc
+++ b/common/polynomial.cc
@@ -357,6 +357,9 @@ Polynomial<T> Polynomial<T>::Integral(const T& integration_constant) const {
 }
 
 template <typename T>
+bool Polynomial<T>::is_univariate() const {return is_univariate_;}
+
+template <typename T>
 bool Polynomial<T>::operator==(const Polynomial<T>& other) const {
   // Comparison of unsorted vectors is faster copying them into std::set
   // btrees rather than using std::is_permutation().
--- a/common/polynomial.h
+++ b/common/polynomial.h
@@ -300,6 +300,9 @@ class Polynomial {
    */
   Polynomial Integral(const T& integration_constant = 0.0) const;
 
+  /** Returns true if this is a univariate polynomial */
+  bool is_univariate() const;
+
   bool operator==(const Polynomial& other) const;
 
   Polynomial& operator+=(const Polynomial& other);
--- a/common/test/polynomial_test.cc
+++ b/common/test/polynomial_test.cc
@@ -618,5 +618,17 @@ GTEST_TEST(PolynomialTest, SubsitutionTest) {
                   .CoefficientsAlmostEqual(x * x + 3 * pow(x, 4) + 3 * y));
 }
 
+GTEST_TEST(PolynomialTest, IsUnivariateTrue) {
+  const VectorXd coefficients = Eigen::Vector3d(0.25, 0.0, 2.0);
+  Polynomial<double> dut(coefficients);
+  EXPECT_TRUE(dut.is_univariate());
+}
+
+GTEST_TEST(PolynomialTest, IsUnivariateFalse) {
+  Polynomiald x("x");
+  Polynomiald y("y");
+  EXPECT_FALSE((x+y).is_univariate());
+}
+
 }  // namespace
 }  // namespace drake
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -1477,6 +1477,13 @@ GetCollisionGeometries(const multibody::MultibodyPlant<double>& plant,
   return ret;
 }
 
+int CspaceFreePolytope::GetSeparatingPlaneIndex(
+    const SortedPair<geometry::GeometryId>& pair) const {
+  return (map_geometries_to_separating_planes_.count(pair) == 0)
+             ? -1
+             : map_geometries_to_separating_planes_.at(pair);
+}
+
 // Explicit instantiation
 template VectorX<symbolic::Polynomial> CspaceFreePolytope::CalcDminusCs<double>(
     const Eigen::Ref<const Eigen::MatrixXd>&,
--- a/geometry/optimization/cspace_free_polytope.h
+++ b/geometry/optimization/cspace_free_polytope.h
@@ -481,10 +481,19 @@ class CspaceFreePolytope {
  protected:
   [[nodiscard]] const symbolic::Variables& get_s_set() const { return s_set_; }
 
-  [[nodiscard]] std::vector<PlaneSeparatesGeometries>&
+  [[nodiscard]] const std::vector<PlaneSeparatesGeometries>&
   get_mutable_plane_geometries() {
     return plane_geometries_;
   }
+
+  const geometry::SceneGraph<double>& get_scene_graph() const {
+    return scene_graph_;
+  }
+
+  // Returns the index of the plane which will separate the geometry pair.
+  // Returns -1 if the pair is not in map_geometries_to_separating_planes_.
+  int GetSeparatingPlaneIndex(
+      const SortedPair<geometry::GeometryId>& pair) const;
 
  private:
   // Forward declaration the tester class. This tester class will expose the
--- a/geometry/optimization/dev/BUILD.bazel
+++ b/geometry/optimization/dev/BUILD.bazel
@@ -5,6 +5,10 @@ load(
     "drake_cc_package_library",
 )
 load("//tools/lint:lint.bzl", "add_lint_tests")
+load(
+    "@drake//tools/skylark:test_tags.bzl",
+    "mosek_test_tags",
+)
 
 exports_files(["cspace_free_polytope.cc"])
 
@@ -13,10 +17,12 @@ drake_cc_library(
     srcs = ["cspace_free_path.cc"],
     hdrs = ["cspace_free_path.h"],
     deps = [
+        "//common:polynomial",
         "//common/symbolic:monomial_util",
         "//geometry/optimization:c_iris_collision_geometry",
         "//geometry/optimization:c_iris_separating_plane",
         "//geometry/optimization:cspace_free_polytope",
+        "//geometry/optimization/dev:polynomial_positive_on_path",
         "//solvers:choose_best_solver",
         "//solvers:mathematical_program",
         "//solvers:mosek_solver",
@@ -51,6 +57,19 @@ drake_cc_googletest(
         "//common/test_utilities:symbolic_test_util",
         "//geometry/optimization:c_iris_test_utilities",
         "//geometry/optimization:cspace_free_polytope",
+    ],
+)
+
+drake_cc_googletest(
+    name = "cspace_free_path_with_mosek_test",
+    timeout = "long",
+    tags = mosek_test_tags(),
+    deps = [
+        ":c_iris_path_test_utilities",
+        "//common/test_utilities:symbolic_test_util",
+        "//common/trajectories:bezier_curve",
+        "//geometry/optimization:c_iris_test_utilities",
+        "//solvers:mosek_solver",
     ],
 )
 
--- a/geometry/optimization/dev/cspace_free_path.cc
+++ b/geometry/optimization/dev/cspace_free_path.cc
@@ -1,5 +1,7 @@
 #include "drake/geometry/optimization/dev/cspace_free_path.h"
 
+#include <vector>
+
 namespace drake {
 namespace geometry {
 namespace optimization {
@@ -27,6 +29,46 @@ initialize_path_map(CspaceFreePath* cspace_free_path,
   return ret;
 }
 
+PlaneSeparatesGeometriesOnPath::PlaneSeparatesGeometriesOnPath(
+    const PlaneSeparatesGeometries& plane_geometries,
+    const symbolic::Variable& mu,
+    const std::unordered_map<symbolic::Variable, symbolic::Polynomial>&
+        path_with_y_subs,
+    const symbolic::Variables& indeterminates,
+    symbolic::Polynomial::SubstituteAndExpandCacheData* cached_substitutions)
+    : plane_index{plane_geometries.plane_index} {
+  auto substitute_and_create_condition =
+      [this, &cached_substitutions, &path_with_y_subs, &indeterminates, &mu](
+          const symbolic::RationalFunction& rational, bool positive_side) {
+        symbolic::Variables parameters;
+        for (const auto& var : rational.numerator().indeterminates()) {
+          parameters.insert(path_with_y_subs.at(var).decision_variables());
+        }
+        symbolic::Polynomial path_numerator{
+            rational.numerator().SubstituteAndExpand(path_with_y_subs,
+                                                     cached_substitutions)};
+
+        // The current y_slacks along with mu.
+        symbolic::Variables cur_indeterminates{
+            intersect(indeterminates, rational.numerator().indeterminates())};
+        cur_indeterminates.insert(mu);
+
+        path_numerator.SetIndeterminates(cur_indeterminates);
+        if (positive_side) {
+          positive_side_conditions.emplace_back(path_numerator, mu, parameters);
+        } else {
+          negative_side_conditions.emplace_back(path_numerator, mu, parameters);
+        }
+      };
+
+  for (const auto& rational : plane_geometries.positive_side_rationals) {
+    substitute_and_create_condition(rational, true);
+  }
+  for (const auto& rational : plane_geometries.negative_side_rationals) {
+    substitute_and_create_condition(rational, false);
+  }
+}
+
 CspaceFreePath::CspaceFreePath(const multibody::MultibodyPlant<double>* plant,
                                const geometry::SceneGraph<double>* scene_graph,
                                SeparatingPlaneOrder plane_order,
@@ -35,49 +77,127 @@ CspaceFreePath::CspaceFreePath(const multibody::MultibodyPlant<double>* plant,
                                const Options& options)
     : CspaceFreePolytope(plant, scene_graph, plane_order, q_star, options),
       mu_(symbolic::Variable("mu")),
+      max_degree_(static_cast<int>(maximum_path_degree)),
       path_(initialize_path_map(this, maximum_path_degree)) {
   this->GeneratePathRationals();
 }
 
 void CspaceFreePath::GeneratePathRationals() {
   // plane_geometries_ currently has rationals in terms of the configuration
-  // space variable. We replace each of these PlaneSeparatesGeometries with a
-  // new one that has rationals in terms of the path variable.
+  // space variable. We create PlaneSeparatesGeometriesOnPath objects which are
+  // in terms of the path variable and can be used to construct the
+  // certification program once a path is chosen.
   symbolic::Polynomial::SubstituteAndExpandCacheData cached_substitutions;
 
   // Add the auxilliary variables for matrix SOS constraints to the substitution
   // map.
   std::unordered_map<symbolic::Variable, symbolic::Polynomial>
       path_with_y_subs = path_;
+  symbolic::Variables indeterminates{mu_};
   for (int i = 0; i < y_slack().size(); ++i) {
     path_with_y_subs.emplace(y_slack()(i), symbolic::Polynomial(y_slack()(i)));
+    indeterminates.insert(y_slack()(i));
   }
 
-  auto generate_substituted_rationals =
-      [&path_with_y_subs, &cached_substitutions](
-          const std::vector<symbolic::RationalFunction>& rationals) {
-        std::vector<symbolic::RationalFunction> path_rationals;
-        path_rationals.reserve(rationals.size());
-        for (const auto& rational : rationals) {
-          const symbolic::Polynomial path_numerator{
-              rational.numerator().SubstituteAndExpand(path_with_y_subs,
-                                                       &cached_substitutions)};
-          const symbolic::Polynomial path_denominator{
-              rational.denominator().SubstituteAndExpand(
-                  path_with_y_subs, &cached_substitutions)};
-          path_rationals.emplace_back(path_numerator, path_denominator);
-        }
-        return path_rationals;
-      };
-
-  std::vector<PlaneSeparatesGeometries> path_plane_geometries;
   for (const auto& plane_geometry : this->get_mutable_plane_geometries()) {
-    path_plane_geometries.emplace_back(
-        generate_substituted_rationals(plane_geometry.positive_side_rationals),
-        generate_substituted_rationals(plane_geometry.negative_side_rationals),
-        plane_geometry.plane_index);
+    plane_geometries_on_path_.emplace_back(plane_geometry, mu_,
+                                           path_with_y_subs, indeterminates,
+                                           &cached_substitutions);
   }
-  this->get_mutable_plane_geometries() = std::move(path_plane_geometries);
+}
+
+[[nodiscard]] CspaceFreePath::PathSeparationCertificateProgram
+CspaceFreePath::MakeIsGeometrySeparableOnPathProgram(
+    const SortedPair<geometry::GeometryId>& geometry_pair,
+    const VectorX<Polynomiald>& path) const {
+  // Fail fast as building the program can be expensive.
+  int plane_index{GetSeparatingPlaneIndex(geometry_pair)};
+  if (plane_index < 0) {
+    throw std::runtime_error(fmt::format(
+        "GetIsGeometrySeparableProgram(): geometry pair ({}, {}) does not need "
+        "a separation certificate",
+        get_scene_graph().model_inspector().GetName(geometry_pair.first()),
+        get_scene_graph().model_inspector().GetName(geometry_pair.second())));
+  }
+
+  DRAKE_DEMAND(rational_forward_kin().s().rows() == path.rows());
+  // Now we convert the vector of common::Polynomial to a map from the
+  // configuration space variable s to symbolic::Polynomial in mu.
+  std::unordered_map<symbolic::Variable, symbolic::Polynomial>
+      cspace_var_to_sym_path;
+  for (int i = 0; i < path.rows(); ++i) {
+    DRAKE_DEMAND(path(i).is_univariate());
+    DRAKE_DEMAND(path(i).GetDegree() <= static_cast<int>(max_degree_));
+    symbolic::Polynomial::MapType sym_path_map;
+    for (const auto& monom : path(i).GetMonomials()) {
+      sym_path_map.insert(
+          {symbolic::Monomial(mu_, monom.GetDegree()), monom.coefficient});
+    }
+    cspace_var_to_sym_path.emplace(rational_forward_kin().s()(i),
+                                   symbolic::Polynomial{sym_path_map});
+  }
+
+  return ConstructPlaneSearchProgramOnPath(
+      plane_geometries_on_path_.at(plane_index), cspace_var_to_sym_path);
+}
+
+[[nodiscard]] CspaceFreePath::PathSeparationCertificateProgram
+CspaceFreePath::ConstructPlaneSearchProgramOnPath(
+    const PlaneSeparatesGeometriesOnPath& plane_geometries_on_path,
+    const std::unordered_map<symbolic::Variable, symbolic::Polynomial>& path)
+    const {
+  PathSeparationCertificateProgram ret{path};
+  ret.plane_index = plane_geometries_on_path.plane_index;
+  ret.prog->AddIndeterminate(mu_);
+  ret.prog->AddIndeterminates(this->y_slack());
+
+  // construct the parameter to value map
+  symbolic::Environment param_eval_map;
+  for (const auto& [config_space_var, eval_path] : path) {
+    const symbolic::Polynomial symbolic_path{path_.at(config_space_var)};
+    for (const auto& [mu_monom, mu_var_coeff] :
+         symbolic_path.monomial_to_coefficient_map()) {
+      // Find the monomial with the matching degree. If it doesn't exist
+      // evaluate it to 0.
+      const auto evaled_monom_iter =
+          eval_path.monomial_to_coefficient_map().find(mu_monom);
+      const double mu_var_coeff_eval{
+          evaled_monom_iter == eval_path.monomial_to_coefficient_map().end()
+              ? 0
+              : evaled_monom_iter->second.Evaluate()};
+      param_eval_map.insert(*mu_var_coeff.GetVariables().begin(),
+                            mu_var_coeff_eval);
+    }
+  }
+
+  // Now add the separation conditions to the program
+  for (const auto& condition :
+       plane_geometries_on_path.positive_side_conditions) {
+    condition.AddPositivityConstraintToProgram(param_eval_map, ret.prog.get());
+  }
+  for (const auto& condition :
+       plane_geometries_on_path.negative_side_conditions) {
+    condition.AddPositivityConstraintToProgram(param_eval_map, ret.prog.get());
+  }
+  return ret;
+}
+
+std::optional<CspaceFreePath::SeparationCertificateResult>
+CspaceFreePath::SolvePathSeparationCertificateProgram(
+    const CspaceFreePath::PathSeparationCertificateProgram& certificate_program,
+    const FindSeparationCertificateGivenPolytopeOptions& options) const {
+  std::optional<CspaceFreePath::SeparationCertificateResult> ret =
+      SolveSeparationCertificateProgram(certificate_program, options);
+  if (ret.has_value()) {
+    // SeparationCertificateResult computes the planes as if it is in s. We now
+    // replace the s variables with the path that was certified.
+    for (int i = 0; i < 3; ++i) {
+      ret.value().a(i) =
+          ret.value().a(i).SubstituteAndExpand(certificate_program.path);
+    }
+    ret.value().b = ret.value().b.SubstituteAndExpand(certificate_program.path);
+  }
+  return ret;
 }
 
 }  // namespace optimization
--- a/geometry/optimization/dev/cspace_free_path.h
+++ b/geometry/optimization/dev/cspace_free_path.h
@@ -1,25 +1,69 @@
 #pragma once
 
+#include <list>
 #include <map>
 #include <optional>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
-#include "geometry/optimization/cspace_free_polytope.h"
+#include <Eigen/Core>
 
 #include "drake/common/eigen_types.h"
+#include "drake/common/polynomial.h"
 #include "drake/geometry/optimization/c_iris_separating_plane.h"
+#include "drake/geometry/optimization/cspace_free_polytope.h"
+#include "drake/geometry/optimization/dev/polynomial_positive_on_path.h"
 
 namespace drake {
 namespace geometry {
 namespace optimization {
 
+/**
+ Contains the information to enforce a pair of geometries are separated by a
+ plane. The conditions are that certain rational functions should be always
+ positive.
+ @param plane_geometries Contains the information required to separate two
+ geometries over a range of configurations in the tangential configuration space
+ variables s. These condition are rational function in the variables s and
+ (sometimes) in auxillary variables y.
+ @param mu The variable used to represent time in the path.
+ @param path_with_y_subs Maps the configuration space variables s to Polynomials
+ of mu and maps the auxillary variables y to symbolic::Polynomial(y).
+ @param indeterminates Must contain mu and the auxillary variables y needed to
+ implement the separating plane condition.
+ @param cached_substitutions SubstituteAndExpandCacheData which can be used to
+ speed up the substitutions of the path.
+ */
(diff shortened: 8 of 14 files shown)


AlexandreAmice (commented):
+@hongkai-dai for feature review. I know that the two changes in bezier curve and polynomial will likely need to be merged in a different PR, but wanted to get your thoughts on the current changes since this one may take a while.

Thanks!
 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NT185fh-ojxmR4Fae33:b-o7mz0b)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
Checkpoint.

> Reviewed 4 of 10 files at r4, 7 of 11 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NT59G3P3qlM37HwefdG:by96gdd)* status: 8 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.cc` line 46 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1NH648hAYlSXUxs_M:-NT1NH648hAYlSXUxs_N:b73a7nn) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/common/trajectories/bezier_curve.cc#L46)):*
> ```cpp
> 
> template <typename T>
> MatrixX<T> BezierCurve<T>::value_no_clamp(const T& time) const {
> ```

What would you think if we combine `value` and `value_no_clamp` functions? We can create a function
```
BezierCurve::value(const T& time, bool clamp_time=true) {
  using std::clamp;
  const T ctime = clamp_time? clamp(time, T{start_time_}, T{end_time_}): time;
  MatrixX<T> v = VectorX<T>::Zero(rows());
  for (int i = 0; i < control_points_.cols(); ++i) {
    v += BernsteinBasis(i, ctime) * control_points_.col(i);
  }
  return v;

}
```
For backward-comparability we can set `clamp_time` to true by default.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 14 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1Ob031mxU8J-svc23:-NT1Ob031mxU8J-svc24:b-1hj3bn) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L14)):*
> ```objc
> 
> #include "drake/common/eigen_types.h"
> #include "drake/common/polynomial.h"
> ```

I wonder if you could use `symbolic::Polynomial` instead of `common::Polynomial`. Ideally we would like to have only one Polynomial class in Drake. `symbolic::Polynomial` is more powerful than `common::Polynomial`.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 72 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT58NkW06buXS1wDbzt:-NT58NkW06buXS1wDbzu:boti610) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L72)):*
> ```objc
>       : public CspaceFreePolytope::SeparationCertificateProgram {
>     PathSeparationCertificateProgram(
>         const std::unordered_map<symbolic::Variable, symbolic::Polynomial>&
> ```

I think it would help if you could document what the key and values are in this map. I initially thought that we have only one path, and the path is parameterized by μ ∈ [0,1], but it seems that here you have many candidate paths?
___
*[`geometry/optimization/dev/cspace_free_path.h` line 83 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT5944_6ESTwcuq90xE:-NT5944_6ESTwcuq90xF:bebjt43) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L83)):*
> ```objc
>   [[nodiscard]] int max_degree() const { return max_degree_; }
> 
>   [[nodiscard]] PathSeparationCertificateProgram
> ```

Could you document this function? It certifies if the geometry_pair is separated on all paths?
___
*[`geometry/optimization/dev/polynomial_positive_on_path.cc` line 3 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1P6Pf-6zo7pcuyBgx:-NT1P6Pf-6zo7pcuyBgy:b-6nzv6e) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/polynomial_positive_on_path.cc#L3)):*
> ```cpp
> #include "drake/geometry/optimization/dev/polynomial_positive_on_path.h"
> 
> #include <iostream>
> ```

Remove this?
___
*[`geometry/optimization/dev/polynomial_positive_on_path.cc` line 117 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1P4Om5TipDWHQqt5F:-NT1P4Om5TipDWHQqt5G:b-6nzv6e) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/polynomial_positive_on_path.cc#L117)):*
> ```cpp
>   const symbolic::Polynomial p_evaled{p_.EvaluatePartial(env)};
>   for (const auto& item : p_evaled.monomial_to_coefficient_map()) {
>     std::cout << item.second << std::endl;
> ```

Remove this?
___
*[`geometry/optimization/dev/test/c_iris_path_test_utilities.h` line 52 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1PTKp7Z5qNGO9r8Pg:-NT1PTKp7Z5qNGO9r8Ph:bu9i35q) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/test/c_iris_path_test_utilities.h#L52)):*
> ```objc
> 
>   [[nodiscard]] const std::vector<PlaneSeparatesGeometries>&
>   get_plane_geometries() {
> ```

This should be a const function?
___
*[`geometry/optimization/dev/test/c_iris_path_test_utilities.h` line 57 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1PVbUEAe3PgNSutWj:-NT1PVbUEAe3PgNSutWk:b-89a4l7) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/test/c_iris_path_test_utilities.h#L57)):*
> ```objc
> 
>   [[nodiscard]] const std::vector<PlaneSeparatesGeometriesOnPath>&
>   get_path_plane_geometries() {
> ```

Ditto

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 8 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`common/trajectories/bezier_curve.cc` line 46 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1NH648hAYlSXUxs_M:-NT5ojVg62whr2wsQO0o:beb9qw8) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/common/trajectories/bezier_curve.cc#L46)):*

What would you think if we combine `value` and `value_no_clamp` functions? We can create a function
```
BezierCurve::value(const T& time, bool clamp_time=true) {
  using std::clamp;
  const T ctime = clamp_time? clamp(time, T{start_time_}, T{end_time_}): time;
  MatrixX<T> v = VectorX<T>::Zero(rows());
  for (int i = 0; i < control_points_.cols(); ++i) {
    v += BernsteinBasis(i, ctime) * control_points_.col(i);
  }
  return v;

}
```
For backward-comparability we can set `clamp_time` to true by default.
</blockquote></details>

Good idea! I'll file that separately, but will make the change for now so that I can pass CI and we can continue the review.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 14 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1Ob031mxU8J-svc23:-NT5lJO3Bq3p19bN-jKy:bwrrvny) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L14)):*

I wonder if you could use `symbolic::Polynomial` instead of `common::Polynomial`. Ideally we would like to have only one Polynomial class in Drake. `symbolic::Polynomial` is more powerful than `common::Polynomial`.
</blockquote></details>

The reason I default to using `common::Polynomial` is the this is what is used by the `trajectory` library. I may overload to allow `symbolic::Polynomial` later.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 72 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT58NkW06buXS1wDbzt:-NT5mEGd4ym1vTNt8RjM:b-72xrcp) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L72)):*

I think it would help if you could document what the key and values are in this map. I initially thought that we have only one path, and the path is parameterized by μ ∈ [0,1], but it seems that here you have many candidate paths?
</blockquote></details>

Done. It is indeed one path, but each configuration space variable tracks a different polynomial path.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 83 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT5944_6ESTwcuq90xE:-NT5oUVE7_Ll-P2PljLx:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L83)):*

Could you document this function? It certifies if the geometry_pair is separated on all paths?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/polynomial_positive_on_path.cc` line 3 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1P6Pf-6zo7pcuyBgx:-NT5oaYZE10QGFztO7r5:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/polynomial_positive_on_path.cc#L3)):*

Remove this?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/polynomial_positive_on_path.cc` line 117 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1P4Om5TipDWHQqt5F:-NT5oaugB0I-ieiInBSk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/polynomial_positive_on_path.cc#L117)):*

Remove this?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/test/c_iris_path_test_utilities.h` line 52 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1PTKp7Z5qNGO9r8Pg:-NT5ofgV5rZomrsBIWnd:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/test/c_iris_path_test_utilities.h#L52)):*

This should be a const function?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/test/c_iris_path_test_utilities.h` line 57 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1PVbUEAe3PgNSutWj:-NT5og1u4o7fHuxFvHAs:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/test/c_iris_path_test_utilities.h#L57)):*

Ditto
</blockquote></details>

Done.

hongkai-dai (commented):
> Reviewed 4 of 7 files at r7.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 7 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_path.h` line 14 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT1Ob031mxU8J-svc23:-NTF4KSK62RLyn1q0yBs:b-kufkxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L14)):*

The reason I default to using `common::Polynomial` is the this is what is used by the `trajectory` library. I may overload to allow `symbolic::Polynomial` later.
</blockquote></details>

I see, thanks!
___
*[`geometry/optimization/dev/cspace_free_path.h` line 72 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT58NkW06buXS1wDbzt:-NTF4ai0DU7EbpSr8L4A:b-lg05o9) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L72)):*

Done. It is indeed one path, but each configuration space variable tracks a different polynomial path.
</blockquote></details>

___
*[`geometry/optimization/dev/cspace_free_path.h` line 32 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFTFS54Q4KwmI5IlgC:-NTFTFS54Q4KwmI5IlgD:bibt46i) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L32)):*
> ```objc
>       const PlaneSeparatesGeometries& plane_geometries,
>       const symbolic::Variable& mu,
>       const std::unordered_map<symbolic::Variable, symbolic::Polynomial>&
> ```

Could you document what `path_with_y_subs` is? I initially thought that it maps each tangential configuration variable `s` to the the polynomial (of indeterminates mu and y). But from the code it seems that maps the variable `y` to Polynomial(y)?
___
*[`geometry/optimization/dev/cspace_free_path.h` line 34 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFNeOmB_TRdt_8eZ3t:-NTFNeOmB_TRdt_8eZ3u:b-pt56q1) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L34)):*
> ```objc
>       const std::unordered_map<symbolic::Variable, symbolic::Polynomial>&
>           path_with_y_subs,
>       const symbolic::Variables& indeterminates,
> ```

Could you document what `indeterminates` here is? We have many symbolic variables that potentially can be indeterminates. I think here you mean `indeterminates` include `mu`, `y`, but not the tangential configuration variable `s`.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 145 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFLoPl9OnnT75lOIZE:-NTFLoPl9OnnT75lOIZF:b-dctoku) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L145)):*
> ```objc
> 
>   // We have the invariant plane_geometries_on_path_[i].plane_index == i. We use
>   // a list instead of a vector as PlaneSeparatesGeometriesOnPath contains
> ```

We do use a vector here?
___
*[`geometry/optimization/dev/cspace_free_path.h` line 155 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFVq6U37R0tQGNs0ve:-NTFVq6U37R0tQGNs0vf:b-tw83ce) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L155)):*
> ```objc
>    need to be non-negative on the path.
>    @param[in] path maps each tangential-configuration space variable to a
>    univariate polynomial of degree less than max_degree_.
> ```

BTW, maybe also document that since the path is given (we are not searching for the path), each of the univariate polynomial should have no decision variables?
___
*[`geometry/optimization/dev/cspace_free_path.cc` line 125 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFU1I52IjFiCvQMx-n:-NTFU1I52IjFiCvQMx-o:bwhlz9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.cc#L125)):*
> ```cpp
>   DRAKE_DEMAND(rational_forward_kin().s().rows() == path.rows());
>   std::unordered_map<symbolic::Variable, symbolic::Polynomial>
>       cspace_var_to_sym_path;
> ```

Could you document what this variable is? I think it maps each `s` to a polynomial of `mu`.
___
*[`geometry/optimization/dev/cspace_free_path.cc` line 194 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFX0WxCg3VZ212DqWG:-NTFX0WxCg3VZ212DqWH:bxfseqa) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.cc#L194)):*
> ```cpp
>     for (int i = 0; i < 3; ++i) {
>       ret.value().a(i) =
>           ret.value().a(i).SubstituteAndExpand(certificate_program.path);
> ```

I didn't realize it when I review the previous PR, that `a` and `b` are both polynomials of `s`. What would you think if we say `a` and `b` are both polynomials of `mu`, namely `a(mu)`? Somehow I feel it is more general to write `a` and `b` as polynomial of `mu` directly, compared to writing `a` being a polynomial of `s`, and then `s` being a polynomial of `mu`, namely `a(mu(s))`.

hongkai-dai (commented):
> Reviewed 1 of 7 files at r7, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 6 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/cspace_free_path.cc` line 194 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFX0WxCg3VZ212DqWG:-NTQfaNk84bZj0X3P-_F:b36eof9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.cc#L194)):*

I didn't realize it when I review the previous PR, that `a` and `b` are both polynomials of `s`. What would you think if we say `a` and `b` are both polynomials of `mu`, namely `a(mu)`? Somehow I feel it is more general to write `a` and `b` as polynomial of `mu` directly, compared to writing `a` being a polynomial of `s`, and then `s` being a polynomial of `mu`, namely `a(mu(s))`.
</blockquote></details>

Based on our slack discussion, changing the formulation to `a(mu)` is a lot of work (but might allow us to search in a richer set of polynomials). We can leave it to future work. I mark myself satisfied.

hongkai-dai (commented):
> Reviewed 1 of 1 files at r8, 2 of 2 files at r10, 2 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 1 unresolved discussion, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/dev/test/cspace_free_path_test.cc` line 42 at r11](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTVp8tk5yvEj9c-UqEQ:-NTVp8tl4NHdlcUFFuL7:b-qw0x07) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f30583b26931cbdc832df600de1762e0242f2d1/geometry/optimization/dev/test/cspace_free_path_test.cc#L42)):*
> ```cpp
>       : mu_{tester.get_mu()} {
>     std::default_random_engine generator;
>     std::uniform_real_distribution<> coeff_distribution(-10, 10);
> ```

Generally we do not use random number in Drake's unit test. It makes it hard to reproduce the test result with random numbers. I would suggest that you just choose the coefficient with arbitrary but fixed numbers, for example 42.

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 1 unresolved discussion, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/dev/cspace_free_path.h` line 72 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NT58NkW06buXS1wDbzt:-NTPpCvFEZxBSFUGJnfY:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8243329196d650c2d439157936f62eb008ded9bf/geometry/optimization/dev/cspace_free_path.h#L72)):*

</blockquote></details>

Done.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 32 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFTFS54Q4KwmI5IlgC:-NTPouZ-6Ug0r36eOOhh:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L32)):*

Could you document what `path_with_y_subs` is? I initially thought that it maps each tangential configuration variable `s` to the the polynomial (of indeterminates mu and y). But from the code it seems that maps the variable `y` to Polynomial(y)?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 34 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFNeOmB_TRdt_8eZ3t:-NTPpC1vCMfhFF2qfhPT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L34)):*

Could you document what `indeterminates` here is? We have many symbolic variables that potentially can be indeterminates. I think here you mean `indeterminates` include `mu`, `y`, but not the tangential configuration variable `s`.
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/cspace_free_path.h` line 145 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFLoPl9OnnT75lOIZE:-NTPpJUGEj7aI83FWEnF:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.h#L145)):*

We do use a vector here?
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/cspace_free_path.cc` line 125 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTFU1I52IjFiCvQMx-n:-NTPr-z_76fXBraofwBH:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ce9f5b876602c776f4ffd6982bb5df3d760cda9/geometry/optimization/dev/cspace_free_path.cc#L125)):*

Could you document what this variable is? I think it maps each `s` to a polynomial of `mu`.
</blockquote></details>

Done.
___
*[`geometry/optimization/dev/test/cspace_free_path_test.cc` line 42 at r11](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTVp8tk5yvEj9c-UqEQ:-NTYuI7x3XDPvJU7fJ9D:bdf9grc) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f30583b26931cbdc832df600de1762e0242f2d1/geometry/optimization/dev/test/cspace_free_path_test.cc#L42)):*

Generally we do not use random number in Drake's unit test. It makes it hard to reproduce the test result with random numbers. I would suggest that you just choose the coefficient with arbitrary but fixed numbers, for example 42.
</blockquote></details>

I do this since this is the way the class `common::Polynomial` is tested. I will fix the seed rather than give the illusion that the coefficients are in anyway meaningful. Does that work for you?

hongkai-dai (commented):
:lgtm: 

+@ggould-tri for platform review (part of the code is in the dev folder that doesn't require platform review).

> Reviewed 7 of 7 files at r12.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTkP8GUBGE5FDY2bsKf:bp7eeex)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @ggould-tri)

___
*[`geometry/optimization/dev/test/cspace_free_path_test.cc` line 42 at r11](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTVp8tk5yvEj9c-UqEQ:-NTkP0In76skpCLunScj:b-zi6bb4) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f30583b26931cbdc832df600de1762e0242f2d1/geometry/optimization/dev/test/cspace_free_path_test.cc#L42)):*

I do this since this is the way the class `common::Polynomial` is tested. I will fix the seed rather than give the illusion that the coefficients are in anyway meaningful. Does that work for you?
</blockquote></details>

That works.

ggould-tri (commented):
> Reviewed 3 of 10 files at r4, 2 of 7 files at r7, 1 of 1 files at r9, 2 of 2 files at r10, 1 of 2 files at r11, 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 4 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTndZjo77UFn25caoXL:-NTndZjo77UFn25caoXM:bpd3s36) (no related file):*
FYI:  The commit message on the Github PR is not adequate for the release engineer (especially with the API-breaking change) and the uncurated commits here will need a new message when squashed.

___
___
*[`geometry/optimization/cspace_free_polytope.h` line 484 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTnryKS03af79I0gemM:-NTnryKS03af79I0gemN:b-tos1cd) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L484)):*
> ```objc
>   [[nodiscard]] const symbolic::Variables& get_s_set() const { return s_set_; }
> 
>   [[nodiscard]] const std::vector<PlaneSeparatesGeometries>&
> ```

Returning a constref to mutable data is misleading:  The caller is going to see the contents of the allegedly-constant return value changed out from under them if they call any non-const method of this class.
This needs, at minimum, a suitably scary comment demanding that the caller not call any non-const methods for the lifetime of this alias.



_Code quote:_
```objc
const std::vector<PlaneSeparatesGeometries>&
```
___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTne5DSEy_49BCy3Hym:-NTne5DSEy_49BCy3Hyn:b-t0acwz) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*
> ```objc
> 
>   [[nodiscard]] const std::vector<PlaneSeparatesGeometries>&
>   get_plane_geometries() {
> ```

As written this is a breaking change without deprecation in the public API for subclassing this class and is outside of the deprecation window (it's been in two releases).
You should either retain the old method (possibly adding a deprecation tag) or else add `+(release notes: breaking change)` to this PR.
___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTnqcm9BSWasbxaIuGN:-NTnqcm9BSWasbxaIuGO:b-ar9mep) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*
> ```objc
> 
>   [[nodiscard]] const std::vector<PlaneSeparatesGeometries>&
>   get_plane_geometries() {
> ```

minor:  Should be method-const so that callers can call it from const methods and/or constrefs to the class (which, again see above, would probably be wise of them to do from a lifetime/mutability point of view).



_Suggestion:_
```objc
  get_plane_geometries() const {
```
___
*[`geometry/optimization/cspace_free_polytope.cc` line 1483 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTnfZGY2Esese-fn-9U:-NTnfZGY2Esese-fn-9V:brh3sju) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.cc#L1483)):*
> ```cpp
>     const SortedPair<geometry::GeometryId>& pair) const {
>   auto geometry_pair_it = map_geometries_to_separating_planes_.find(pair);
>   return (geometry_pair_it == map_geometries_to_separating_planes_.end())
> ```

minor:  `count()` is less verbose and will be easily replaced with the actually correct `contains()` when C++20 support arrives.



_Suggestion:_
```cpp
  return (map_geometries_to_separating_planes_.count(pair))
```

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 3 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @ggould-tri)

___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTne5DSEy_49BCy3Hym:-NTnv2DVB3HdpNQp57S_:b40s8y1) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*

As written this is a breaking change without deprecation in the public API for subclassing this class and is outside of the deprecation window (it's been in two releases).
You should either retain the old method (possibly adding a deprecation tag) or else add `+(release notes: breaking change)` to this PR.
</blockquote></details>

This change was originally introduced because I thought I would need to change the plane_geometries in a subclass, but actually I only need to access (not change) them. In this case, which do you recommend? 

It sounds to me that perhaps leaving this method as `get_mutable_plane_geometries()` is most appropriate given that `plane_geometries_` is not a const member? Does the answer change given that this member is protected?
___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTnqcm9BSWasbxaIuGN:-NTnvbNkDl8DTkuWcwyQ:b37x7p6) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*

minor:  Should be method-const so that callers can call it from const methods and/or constrefs to the class (which, again see above, would probably be wise of them to do from a lifetime/mutability point of view).
</blockquote></details>

Should I add the `const` even if we leave this as `get_mutable_plane_geometries`?

hongkai-dai (commented):
> Reviewed 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201)* status: 3 unresolved discussions, LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @AlexandreAmice and @ggould-tri)

ggould-tri (commented):
Apologies for being slow to get back to this -- I've had other things diverting my attention.  Anyway, :lgtm: for whichever of these paths you choose:
 * Only the mutable form of the method; no further action.
 * Add the new non-mutable form of the method, and an admonitory "const reference to mutable data" comment on the API.
 * Add a breaking change tag to this PR; remove the mutable method; add the non-mutable method (and comment).

> Reviewed 1 of 1 files at r13, 8 of 8 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTtk3rQ1HPQ-GYGR0ev:bvdo2v)* status: missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTnqcm9BSWasbxaIuGN:-NTo0UqT5Paybjp-vPc0:bk78484) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*

Should I add the `const` even if we leave this as `get_mutable_plane_geometries`?
</blockquote></details>

No, a const method returning a mutation handle would be misleading.
___
*[`geometry/optimization/cspace_free_polytope.h` line 485 at r12](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-NTne5DSEy_49BCy3Hym:-NTo0cj576NOiVxmUAvM:b-zh833l) ([raw file](https://github.com/RobotLocomotion/drake/blob/f067598203b9bda71c66fbe3b512089eda64458b/geometry/optimization/cspace_free_polytope.h#L485)):*

This change was originally introduced because I thought I would need to change the plane_geometries in a subclass, but actually I only need to access (not change) them. In this case, which do you recommend? 

It sounds to me that perhaps leaving this method as `get_mutable_plane_geometries()` is most appropriate given that `plane_geometries_` is not a const member? Does the answer change given that this member is protected?
</blockquote></details>

`protected` is, I think, irrelevant to the analysis:  As written, subclassability is part of this class's public API.

Leaving the method as it was is certainly the safest tack, and it makes clear that this is a dangerous method returning a dangerous reference.  That will make the consuming code in `dev` a bit more capable than it needs to be (and makes the for loop over the result a bit suspcious looking) but that might be for the best anyway.  It does mean that this can't be called from const methods.

AlexandreAmice (commented):
Thanks! Once I pass CI I will merge then. I went with only the mutable form

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTtnJ2K9FnOCUHHwYYV:bv4juko)* status: missing label for release notes (waiting on @ggould-tri)

AlexandreAmice (commented):
What should I include for release notes? Technically there is the functionality for checking whether a `common::Polynomial` is univariate?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTuSlLI-RrYBpq1hLOw:bmw3i4q)* status: missing label for release notes (waiting on @AlexandreAmice)

ggould-tri (commented):
I'd be happy with None or Feature.  If you don't have a strong preference I'd go with none, since the PR headline doesn't flag the (_de minimis_) feature change to the release engineer anyway.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTxq82p2vQ4SLJ1nVAg:bc2ctzh)* status: missing label for release notes (waiting on @AlexandreAmice)

AlexandreAmice (commented):
+(release notes: none) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19201#-:-NTy2lYc08MORX18he6o:b-safpmt)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),hongkai-dai (waiting on @AlexandreAmice)

PR #19205, opened 2023-04-15
Add BezierCurve::GetExpression()

AlexandreAmice:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)

### DIFF
--- a/bindings/pydrake/test/trajectories_test.py
+++ b/bindings/pydrake/test/trajectories_test.py
@@ -24,6 +24,7 @@ from pydrake.trajectories import (
     Trajectory,
     Trajectory_
 )
+from pydrake.symbolic import Variable, Expression
 
 
 # Custom trajectory class used to test Trajectory subclassing in python.
@@ -100,6 +101,11 @@ class TestTrajectories(unittest.TestCase):
         b = curve.BernsteinBasis(i=0, time=1.5, order=1)
         self.assertIsInstance(b, T)
         numpy_compare.assert_float_equal(curve.control_points(), points)
+
+        curve_expression = curve.GetExpression(time=Variable("t"))
+        self.assertEqual(curve_expression.shape, (2,))
+        for expr in curve_expression:
+            self.assertTrue(isinstance(expr, Expression))
 
     @numpy_compare.check_all_types
     def test_bspline_trajectory(self, T):
--- a/bindings/pydrake/trajectories_py.cc
+++ b/bindings/pydrake/trajectories_py.cc
@@ -276,7 +276,10 @@ struct Impl {
               py::arg("time"), py::arg("order") = std::nullopt,
               cls_doc.BernsteinBasis.doc)
           .def("control_points", &Class::control_points,
-              cls_doc.control_points.doc);
+              cls_doc.control_points.doc)
+          .def("GetExpression", &Class::GetExpression,
+              py::arg("time") = symbolic::Variable("t"),
+              cls_doc.GetExpression.doc);
 
       DefCopyAndDeepCopy(&cls);
       py::implicitly_convertible<Class, copyable_unique_ptr<Trajectory<T>>>();
--- a/common/trajectories/BUILD.bazel
+++ b/common/trajectories/BUILD.bazel
@@ -153,6 +153,8 @@ drake_cc_library(
     ],
     deps = [
         ":trajectory",
+        "//common:drake_bool",
+        "//common/symbolic:polynomial",
         "//math:binomial_coefficient",
     ],
 )
@@ -293,6 +295,7 @@ drake_cc_googletest(
     name = "bezier_curve_test",
     deps = [
         ":bezier_curve",
+        "//common/symbolic:polynomial",
         "//common/test_utilities",
     ],
 )
--- a/common/trajectories/bezier_curve.cc
+++ b/common/trajectories/bezier_curve.cc
@@ -1,5 +1,7 @@
 #include "drake/common/trajectories/bezier_curve.h"
 
+#include "drake/common/drake_bool.h"
+#include "drake/common/symbolic/polynomial.h"
 #include "drake/math/binomial_coefficient.h"
 
 namespace drake {
@@ -34,12 +36,37 @@ std::unique_ptr<Trajectory<T>> BezierCurve<T>::Clone() const {
 template <typename T>
 MatrixX<T> BezierCurve<T>::value(const T& time) const {
   using std::clamp;
-  const T ctime = clamp(time, T{start_time_}, T{end_time_});
-  MatrixX<T> v = VectorX<T>::Zero(rows());
-  for (int i = 0; i < control_points_.cols(); ++i) {
-    v += BernsteinBasis(i, ctime) * control_points_.col(i);
+  return EvaluateT(clamp(time, T{start_time_}, T{end_time_}));
+}
+
+template <typename T>
+VectorX<symbolic::Expression> BezierCurve<T>::GetExpression(
+    symbolic::Variable time) const {
+  if constexpr (scalar_predicate<T>::is_bool) {
+    MatrixX<symbolic::Expression> control_points{control_points_.rows(),
+                                                 control_points_.cols()};
+
+    if constexpr (std::is_same_v<T, double>) {
+      control_points = control_points_.template cast<symbolic::Expression>();
+    } else {
+      // AutoDiffXd.
+      for (int r = 0; r < control_points.rows(); ++r) {
+        for (int c = 0; c < control_points.cols(); ++c) {
+          control_points(r, c) =
+              symbolic::Expression(control_points_(r, c).value());
+        }
+      }
+    }
+    return BezierCurve<symbolic::Expression>(start_time_, end_time_,
+                                             control_points)
+        .GetExpression(time);
+  } else {
+    VectorX<symbolic::Expression> ret{EvaluateT(symbolic::Expression(time))};
+    for (int i = 0; i < ret.rows(); ++i) {
+      ret(i) = ret(i).Expand();
+    }
+    return ret;
   }
-  return v;
 }
 
 template <typename T>
@@ -75,6 +102,15 @@ MatrixX<T> BezierCurve<T>::DoEvalDerivative(const T& time,
   MatrixX<T> v = VectorX<T>::Zero(rows());
   for (int i = 0; i < points.cols(); ++i) {
     v += BernsteinBasis(i, ctime, order_ - derivative_order) * points.col(i);
+  }
+  return v;
+}
+
+template <typename T>
+VectorX<T> BezierCurve<T>::EvaluateT(const T& time) const {
+  VectorX<T> v = VectorX<T>::Zero(rows());
+  for (int i = 0; i < control_points_.cols(); ++i) {
+    v += BernsteinBasis(i, time) * control_points_.col(i);
   }
   return v;
 }
--- a/common/trajectories/bezier_curve.h
+++ b/common/trajectories/bezier_curve.h
@@ -66,6 +66,11 @@ class BezierCurve final : public trajectories::Trajectory<T> {
            a trajectory defined over [0, 1]. */
   MatrixX<T> value(const T& time) const override;
 
+  /** Extracts the expanded underlying polynomial expression of this curve in
+   terms of variable `time`. */
+  VectorX<symbolic::Expression> GetExpression(
+      symbolic::Variable time = symbolic::Variable("t")) const;
+
   Eigen::Index rows() const override { return control_points_.rows(); }
 
   Eigen::Index cols() const override { return 1; }
@@ -83,6 +88,8 @@ class BezierCurve final : public trajectories::Trajectory<T> {
 
   std::unique_ptr<trajectories::Trajectory<T>> DoMakeDerivative(
       int derivative_order) const override;
+
+  VectorX<T> EvaluateT(const T& time) const;
 
   double start_time_{};
   double end_time_{};
--- a/common/trajectories/test/bezier_curve_test.cc
+++ b/common/trajectories/test/bezier_curve_test.cc
@@ -3,6 +3,7 @@
 #include <gflags/gflags.h>
 #include <gtest/gtest.h>
 
+#include "drake/common/symbolic/polynomial.h"
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
 
 namespace drake {
@@ -24,6 +25,7 @@ GTEST_TEST(BezierCurveTest, Linear) {
   EXPECT_EQ(curve.end_time(), 3.0);
   EXPECT_TRUE(
       CompareMatrices(curve.value(2.5), Eigen::Vector2d(1.5, 5), 1e-14));
+  EXPECT_TRUE(CompareMatrices(curve.value(0), Eigen::Vector2d(1, 3), 1e-14));
 
   auto deriv = curve.MakeDerivative();
   BezierCurve<double>& deriv_bezier =
@@ -79,6 +81,10 @@ GTEST_TEST(BezierCurveTest, Quadratic) {
     EXPECT_TRUE(CompareMatrices(curve.value(sample_time),
                                 Eigen::Vector2d((1 - t) * (1 - t), t * t),
                                 1e-14));
+    EXPECT_TRUE(CompareMatrices(curve.value(sample_time),
+                                Eigen::Vector2d((1 - t) * (1 - t), t * t),
+                                1e-14));
+
     EXPECT_TRUE(CompareMatrices(deriv->value(sample_time),
                                 Eigen::Vector2d(-2 * (1 - t), 2 * t), 1e-14));
     EXPECT_TRUE(CompareMatrices(curve.EvalDerivative(sample_time),
@@ -87,6 +93,14 @@ GTEST_TEST(BezierCurveTest, Quadratic) {
                                 Eigen::Vector2d(2, 2), 1e-14));
     EXPECT_TRUE(CompareMatrices(curve.EvalDerivative(sample_time, 2),
                                 Eigen::Vector2d(2, 2), 1e-14));
+  }
+
+  // Extract the symoblic exprssion for the bezier curve.
+  VectorX<symbolic::Expression> curve_expression{
+      curve.GetExpression(symbolic::Variable("t"))};
+  for (int i = 0; i < curve_expression.rows(); i++) {
+    EXPECT_TRUE(curve_expression(i).is_polynomial());
+    EXPECT_EQ(symbolic::Polynomial(curve_expression(i)).TotalDegree(), 2);
   }
 }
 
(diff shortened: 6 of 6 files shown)


AlexandreAmice (commented):
+@RussTedrake for feature review. Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NT5vBCdCst3wnveS49l:bn3gxtq)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

RussTedrake (commented):
:lgtm: -- just a few small requests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NT8Z3_UEiOVSjaRs_fz:bwrpabr)* status: 2 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.h` line 70 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NT8Y8CHA9uMw8GUpHpL:-NT8Y8CHA9uMw8GUpHpM:b5b0pkq) ([raw file](https://github.com/RobotLocomotion/drake/blob/e1f9ea925fcbc0c44f800a05c82aada9fd0d14a8/common/trajectories/bezier_curve.h#L70)):*
> ```objc
> 
>   /** Evaluates the curve at the given time with the option to avoid the default
>    behavior of clamping time. This can be particularly useful when attempting
> ```

could we add just a little more detail / context here?  In the docstring for value it is more clear what "clamping time" means.  I think saying something more verbose like that will help the more casual readers.
___
*[`common/trajectories/test/bezier_curve_test.cc` line 92 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NT8YXyk9ZnP85W2_DnD:-NT8YXylEi9q2LMtgGTf:b-f1ldb0) ([raw file](https://github.com/RobotLocomotion/drake/blob/e1f9ea925fcbc0c44f800a05c82aada9fd0d14a8/common/trajectories/test/bezier_curve_test.cc#L92)):*
> ```cpp
>         CompareMatrices(curve.value(sample_time, false),
>                         Eigen::Vector2d((1 - sample_time) * (1 - sample_time),
>                                         sample_time * sample_time),
> ```

Could we add a few line test to check the behavior you wanted / promised in the doc string?  Show that for a `BezierCurve<Expression>` you can call `value(..., false)` and get an `Expression` for which `is_polynomial() == true`?  

(that will help ensure that this remains true even as others make changes to the code!)

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 2 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`common/trajectories/bezier_curve.h` line 70 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NT8Y8CHA9uMw8GUpHpL:-NT9Pz9a7PsESeMfxWSx:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/e1f9ea925fcbc0c44f800a05c82aada9fd0d14a8/common/trajectories/bezier_curve.h#L70)):*

could we add just a little more detail / context here?  In the docstring for value it is more clear what "clamping time" means.  I think saying something more verbose like that will help the more casual readers.
</blockquote></details>

Done.
___
*[`common/trajectories/test/bezier_curve_test.cc` line 92 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NT8YXyk9ZnP85W2_DnD:-NT9QHF22JBU07IL_Eof:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/e1f9ea925fcbc0c44f800a05c82aada9fd0d14a8/common/trajectories/test/bezier_curve_test.cc#L92)):*

Could we add a few line test to check the behavior you wanted / promised in the doc string?  Show that for a `BezierCurve<Expression>` you can call `value(..., false)` and get an `Expression` for which `is_polynomial() == true`?  

(that will help ensure that this remains true even as others make changes to the code!)
</blockquote></details>

Done.

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NT9YFT6CRDFXmqiWYkc:b-6cxuoj)* status: LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
As I read this PR, it feels like the purpose is to be able to get the polynomial underlying Bezier curve and this particular implementation is a roundabout way of doing it that. Is that right?

I'd propose the following:

1. Move *most* of the body of `value()` into a private function: call it `EvaluateT()` (the clamping is left behind).
2. The current body of `value()` calls two things: `clamp()` and `EvaluateT()`.
3. Add a new public method called something like: `GetBezierExpression`.
    - It could be `SFINAE`'d into only supporting `T = Expression` or it could do a conversion under the hood to return an expression based on arbitrary T-valued `BezierCurve`s.
    - It could hard-code the variable `t` or it could request the independent variable (with a default value of a variable `t`).

This eliminates the weird semantics of unclamping `t` for the other cases and makes the use case of the new function abundantly clear (and possibly more general if you chose to support arbitrary T).

Let me know if these thoughts aren't clear or if you'd like some help as to what that might look like. I'm happy to push a PR against your PR so you can see what I mean in more concrete terms.

> Reviewed 1 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTEcFFI5yCHy03y5DXX:b-oigj2e)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEdS2e5YRdlE4CuLXR:-NTEdS2f1tj29poa6eO0:be2nf9x) (no related file):*
BTW Do you have an *explicit* plan vis a vis python bindings? I'm not sure if its absence is part of a strategy or oversight.

___
___
*[`common/trajectories/bezier_curve.h` line 69 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEUTaI55CxIn9GoUTC:-NTEUTaI55CxIn9GoUTD:b-jtcebb) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/bezier_curve.h#L69)):*
> ```objc
>   MatrixX<T> value(const T& time) const override;
> 
>   /** Evaluates the curve at the given time. If clamp_time is true and t does
> ```

BTW this could be simplified and expressed in terms of the *other* implementation without duplicating documentation.

```c++
  /** Evaluates the curve at the given time.
   
   Calling `value(t, true)` is equivalent to calling `value(t)` (see above).

   If `T = symbolic::Expression`, passing `clamp_time = false` has the effect of
   extracting the underlying polynomial expression of this curve. */
```
___
*[`common/trajectories/bezier_curve.h` line 72 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEU-hP5fZFN3Ct9iHW:-NTEU-hP5fZFN3Ct9iHX:b-qqo1ev) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/bezier_curve.h#L72)):*
> ```objc
>    not lie in the range [start_time(), end_time()], the trajectory will silently
>    be evaluated at the closest valid value of time to `time`. For example,
>    `value(-1)` will return `value(0)` for a trajectory defined over [0, 1]. When
> ```

nit: You omitted the required argument. (Although, if you like my previous suggestion, this would simply evaporate.)



_Suggestion:_
```objc
value(-1, true)
```
___
*[`common/trajectories/bezier_curve.h` line 74 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEVI0V4SWpc90O9mQF:-NTEVI0V4SWpc90O9mQG:b-l45mnf) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/bezier_curve.h#L74)):*
> ```objc
>    `value(-1)` will return `value(0)` for a trajectory defined over [0, 1]. When
>    the curve is of type Expression, calling this method with time =
>    symbolic::Variable("t") and clamp_time = false enables the extraction of the
> ```

Extraction of the polynomial seems to be the *point* of this function.

1. Are there any warnings we should offer up when `T != symbolic::Expression`? Perhaps that the result is meaningless?
2. Had you considered making this a `T = symbolic::Expression` only sugar function that more directly just returns the polynomial expression in `t`? Or is there utility of this function beyond that application?
___
*[`common/trajectories/bezier_curve.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEZ1_3C5m8yAuFnJW6:-NTEZ1_42wYeOHiqtenv:b-vbhcce) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/bezier_curve.cc#L42)):*
> ```cpp
> MatrixX<T> BezierCurve<T>::value(const T& time, bool clamp_time) const {
>   using std::clamp;
>   const T ctime = clamp_time ? clamp(time, T{start_time_}, T{end_time_}) : time;
> ```

The ternary operator is causing me mild concern. This may add more weight to the idea of rather than adding a general solution (stop clamping) to get a particular purpose (polynomial expression), the true answer may be to simply give a function that returns the polynomial expression without all of the concomitant traps of the general solution.
___
*[`common/trajectories/test/bezier_curve_test.cc` line 38 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTE_3a04dkN4yKbxmN5:-NTE_3a04dkN4yKbxmN6:b-f2i9ex) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/test/bezier_curve_test.cc#L38)):*
> ```cpp
>       curve_sym.value(symbolic::Variable("t"), false)};
>   for (int i = 0; i < curve_sym_expression.rows(); i++) {
>     EXPECT_TRUE(curve_sym_expression(i).is_polynomial());
> ```

BTW Is this test sufficient to support the documented purpose (polynomial expression of curve)? These properties are necessary, but are they sufficient? It feels like it would be nice if there were *one* test that actually looked at the expression and said, "Yep. That's the expected expression." (Coupled with arguments along the lines of, "if it worked for this simple, low-order example, we have a reasonable basis to believe it works equally well for high-order scenarios.")

However, I'm willing to stipulate that might be overly naive and there's a very cogent argument for why the correctness of these tested properties implies the correctness of the full assertion. If you're testing strategy relies on this kind of argument, you'd do your future readers a service in presenting it.
___
*[`common/trajectories/test/bezier_curve_test.cc` line 96 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTEaSPR-A8q2byFR-dc:-NTEaSPR-A8q2byFR-dd:b-aa8esl) ([raw file](https://github.com/RobotLocomotion/drake/blob/c91e025be51d6da61cef7eb3df4e511334b415f8/common/trajectories/test/bezier_curve_test.cc#L96)):*
> ```cpp
>                                 Eigen::Vector2d((1 - t) * (1 - t), t * t),
>                                 1e-14));
>     EXPECT_TRUE(CompareMatrices(curve.value(sample_time, true),
> ```

btw: This assertion is an equivalent spelling of the previous test. It would be really nice if the quantity `Eigen::Vector2d(1 - t) * (1 - t), t * t)` were captured in a local variable and used both times. That would strongly underscore the equivalency of the two calls.

SeanCurtis-TRI (commented):
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTEeV2s6PsGKPihfVtM:b-l8xbmn)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

SeanCurtis-TRI (commented):
Lookin' good. Probably worth getting a refresh on feature review with the significant changes. I'd also suggest curating your commits so we can see what the new commit message will be.

> Reviewed 1 of 4 files at r4, 3 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTPmXiy9tIg9XQExUx7:byr16rk)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`common/trajectories/test/bezier_curve_test.cc` line 143 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTPlnNd0XIfQnrgDZjQ:-NTPlnNd0XIfQnrgDZjR:b-asrsw5) ([raw file](https://github.com/RobotLocomotion/drake/blob/0baeb3f9d696477c3226519ababe1171d626a709/common/trajectories/test/bezier_curve_test.cc#L143)):*
> ```cpp
> 
>   // Tests that the call to GetExpression returns a polynomial of appropriate
>   // degree and that all scalar types return the same expression. Whether the
> ```

BTW Worth noting that this works because the times and control points of the Bezier curves don't include variables?
___
*[`common/trajectories/test/bezier_curve_test.cc` line 157 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTPlNUz7VD_ZFBjKmiL:-NTPlNUz7VD_ZFBjKmiM:b-yhpi5y) ([raw file](https://github.com/RobotLocomotion/drake/blob/0baeb3f9d696477c3226519ababe1171d626a709/common/trajectories/test/bezier_curve_test.cc#L157)):*
> ```cpp
>             curve_sym.GetExpression(t)};
>         const VectorX<symbolic::Expression> curve_expression_double{
>             curve_sym.GetExpression(t)};
> ```

nit: some copypasta here. I'm pretty sure only one of these should invoke `curve_sym` and the others should be `curve` and `curve_ad`.
___
*[`bindings/pydrake/test/trajectories_test.py` line 106 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTPm7fA6Yw0bHNUgJpE:-NTPm7fA6Yw0bHNUgJpF:bchi8a) ([raw file](https://github.com/RobotLocomotion/drake/blob/ac84ac24de31f170a394564f122f2ad3217e2fc0/bindings/pydrake/test/trajectories_test.py#L106)):*
> ```Python
> 
>         curve_expression = curve.GetExpression(time=Variable("t"))
>         self.assertEqual(curve_expression.shape, (2,1))
> ```

BTW You don't have to test for the correctness of the `curve_expression`'s contents. You could simply assert that the return value *is* a matrix of expressions.

AlexandreAmice (commented):
@RussTedrake could you refresh the feature review for this?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTPvQr56IQ59HefLbP0:bso30qq)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTPxswt2her9YCoUASa:bmxilld)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform) (waiting on @AlexandreAmice)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 1 unresolved discussion (waiting on @AlexandreAmice)

___
*[`bindings/pydrake/test/trajectories_test.py` line 106 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTPy2tx1PzUXQYp8T2H:-NTPy2tx1PzUXQYp8T2I:by1eyj0) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5596793a61f9f47952492fd7c02a5c226d4f79c/bindings/pydrake/test/trajectories_test.py#L106)):*
> ```Python
> 
>         curve_expression = curve.GetExpression(time=Variable("t"))
>         self.assertEqual(curve_expression.shape, (2,1))
> ```

nit: A comma between 2 and 1 is what CI is complaining about.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 2 unresolved discussions (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.cc` line 66 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTQCcGB87XBNwbry1o_:-NTQCcGB87XBNwbry1oa:bqvlli9) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5596793a61f9f47952492fd7c02a5c226d4f79c/common/trajectories/bezier_curve.cc#L66)):*
> ```cpp
>     MatrixX<symbolic::Expression> ret{EvaluateT(symbolic::Expression(time))};
>     for (int i = 0; i < ret.rows(); ++i) {
>       for (int j = 0; j < ret.rows(); ++j) {
> ```

nit: I think this is your crash.



_Suggestion:_
```cpp
ret.cols();
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform) (waiting on @AlexandreAmice)

SeanCurtis-TRI (commented):
+@hongkai-dai for follow up feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTVhM460k11TUKibhAW:b-j9yv1l)* status: LGTM missing from assignee hongkai-dai (waiting on @hongkai-dai)

hongkai-dai (commented):
:lgtm: 

> Reviewed 1 of 4 files at r4, 1 of 3 files at r5, 1 of 2 files at r6, 1 of 2 files at r7, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTVo3mR4wb17WgQZSyW:bg7igkv)* status: 1 unresolved discussion (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.h` line 71 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTVknvU2vEVgicb6H5J:-NTVknvU2vEVgicb6H5K:bm6vft5) ([raw file](https://github.com/RobotLocomotion/drake/blob/bb5f3969af257e99722454427965e7c34422fdaf/common/trajectories/bezier_curve.h#L71)):*
> ```objc
>   /** Extracts the expanded underlying polynomial expression of this curve in
>    terms of variable `time`. */
>   MatrixX<symbolic::Expression> GetExpression(
> ```

Curious why this return `Matrix<symbolic::Expression>` instead of `VectorX<symbolic::Expression>`. My understanding is that the value is always an Eigen vector at each instance in time.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 2 unresolved discussions (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.h` line 73 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTVsTSx0OnEYnZS0dub:-NTVsTSx0OnEYnZS0duc:b-wxxuea) ([raw file](https://github.com/RobotLocomotion/drake/blob/bb5f3969af257e99722454427965e7c34422fdaf/common/trajectories/bezier_curve.h#L73)):*
> ```objc
>   MatrixX<symbolic::Expression> GetExpression(
>       symbolic::Variable time = symbolic::Variable("t")) const;
> ```

BTW, why not to pass `time` by const reference here?

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 1 unresolved discussion (waiting on @hongkai-dai and @SeanCurtis-TRI)

___
*[`common/trajectories/bezier_curve.h` line 71 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTVknvU2vEVgicb6H5J:-NTWlXhVE-PLG7QF9d1j:bv1fyjc) ([raw file](https://github.com/RobotLocomotion/drake/blob/bb5f3969af257e99722454427965e7c34422fdaf/common/trajectories/bezier_curve.h#L71)):*

Curious why this return `Matrix<symbolic::Expression>` instead of `VectorX<symbolic::Expression>`. My understanding is that the value is always an Eigen vector at each instance in time.
</blockquote></details>

It is so that it matches the return type of `value` and `EvaluateT`. Not sure if that is important or not.

hongkai-dai (commented):
> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 1 unresolved discussion (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.h` line 71 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTVknvU2vEVgicb6H5J:-NTZtlep0gWvpO-nNeTz:bh0ywo) ([raw file](https://github.com/RobotLocomotion/drake/blob/bb5f3969af257e99722454427965e7c34422fdaf/common/trajectories/bezier_curve.h#L71)):*

It is so that it matches the return type of `value` and `EvaluateT`. Not sure if that is important or not.
</blockquote></details>

I see, I think we want `value` to return MatrixX because it is inherited from the base class `Trajectory`. I am not sure if we want `EvaluateT` and `GetExpression` to return `MatrixX` as well? It seems more accurate to return `VectorX` for these two methods?

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205)* status: 1 unresolved discussion (waiting on @hongkai-dai)

___
*[`common/trajectories/bezier_curve.h` line 71 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTVknvU2vEVgicb6H5J:-NTZwovn-T1ffu7KIcnZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/bb5f3969af257e99722454427965e7c34422fdaf/common/trajectories/bezier_curve.h#L71)):*

I see, I think we want `value` to return MatrixX because it is inherited from the base class `Trajectory`. I am not sure if we want `EvaluateT` and `GetExpression` to return `MatrixX` as well? It seems more accurate to return `VectorX` for these two methods?
</blockquote></details>

Done.

hongkai-dai (commented):
:lgtm:

> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTZyFWzEt-yC-_R1INo:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),hongkai-dai,RussTedrake(platform) (waiting on @AlexandreAmice)

SeanCurtis-TRI (commented):
A bit of clean up due to the change of return type.

> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-:-NTZzig8Eh9GhN_PKLLI:bqas8mh)* status: 2 unresolved discussions (waiting on @AlexandreAmice)

___
*[`common/trajectories/bezier_curve.cc` line 65 at r10](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTZzEFR1Z0O1-_huDmU:-NTZzEFR1Z0O1-_huDmV:bl0dg9m) ([raw file](https://github.com/RobotLocomotion/drake/blob/a3d1111331e26f4ecc4a4d7b3b8b23d4214a634f/common/trajectories/bezier_curve.cc#L65)):*
> ```cpp
>   } else {
>     VectorX<symbolic::Expression> ret{EvaluateT(symbolic::Expression(time))};
>     for (int i = 0; i < ret.rows(); ++i) {
> ```

nit: This shouldn't be rows *and* cols anymore. "Yes" for the control points. "No" for the return value.
___
*[`bindings/pydrake/test/trajectories_test.py` line 106 at r10](https://reviewable.io/reviews/RobotLocomotion/drake/19205#-NTZzccM43y4sQQG0EjZ:-NTZzccM43y4sQQG0Ej_:b-lveomt) ([raw file](https://github.com/RobotLocomotion/drake/blob/a3d1111331e26f4ecc4a4d7b3b8b23d4214a634f/bindings/pydrake/test/trajectories_test.py#L106)):*
> ```Python
> 
>         curve_expression = curve.GetExpression(time=Variable("t"))
>         self.assertEqual(curve_expression.shape, (2, 1))
> ```

nit: Your CI failure is here. The shape is no longer (2, 1). Now it's simply (2, ).

PR #19206, opened 2023-04-16
[common] Fix FindResource vs non-source Bazel builds

jwnimmer-tri:
When a user installs Drake via pip or apt, but still uses Bazel as their build & test system, then previously we'd get confused and try to use Bazel runfiles to find Drake resources. Instead, we should fall back to the other mechanism (library-relative paths) in this case.

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19206)

### DIFF
--- a/common/find_resource.cc
+++ b/common/find_resource.cc
@@ -207,6 +207,12 @@ std::optional<string> MaybeGetInstallResourceRoot() {
   return std::nullopt;
 }
 
+Result MakeResultFrom(const string& resource_path, RlocationOrError other) {
+  return other.error.empty()
+             ? Result::make_success(resource_path, std::move(other.abspath))
+             : Result::make_error(resource_path, std::move(other.error));
+}
+
 }  // namespace
 
 const char* const kDrakeResourceRootEnvironmentVariableName =
@@ -243,14 +249,19 @@ Result FindResource(const string& resource_path) {
         *guess, resource_path);
   }
 
-  // (2) Check the Runfiles.
+  // (2) Check the Runfiles. If and only if we have Drake's runfiles (not just
+  // any old runfiles) should we consider runfiles as a source for FindResource.
+  // Downstream projects that use Bazel but don't build Drake from source will
+  // have runfiles but not Drake's runfiles; in that case we should skip this
+  // option and continue to option (3) instead.
   if (HasRunfiles()) {
-    auto rlocation_or_error = FindRunfile(resource_path);
-    if (rlocation_or_error.error.empty()) {
-      return Result::make_success(
-          resource_path, rlocation_or_error.abspath);
+    if (FindRunfile(kSentinelRelpath).error.empty()) {
+      return MakeResultFrom(resource_path, FindRunfile(resource_path));
+    } else {
+      log()->debug(
+          "FindResource ignoring Bazel runfiles with no sentinel file {}.",
+          kSentinelRelpath);
     }
-    return Result::make_error(resource_path, rlocation_or_error.error);
   }
 
   // (3) Check the `libdrake_marker.so` location in the install tree.
--- a/common/test/resource_tool_installed_test.py
+++ b/common/test/resource_tool_installed_test.py
@@ -2,6 +2,7 @@
 """
 
 import os
+from pathlib import Path
 import unittest
 import subprocess
 import sys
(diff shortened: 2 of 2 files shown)


jwnimmer-tri:
Okay, I manually tested with `drake-blender` that this new wheel allows removing the current path hack:

https://github.com/RobotLocomotion/drake-blender/blob/b554945b4e334d420c632e6360fb5a0cc61bfa8a/examples/ball_bin.py#L191-L195

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19206#-:-NTGODv38sJ7OG43vMEP:bnfp4nl)* status: needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri:

xuchenhan-tri (commented):
platform :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19206#-:-NTKgL5g8Xk8aYJvzCbP:bzus18)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`common/find_resource.cc` line 261 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19206#-NTKf1pC-kubYxWb_o8Q:-NTKf1pC-kubYxWb_o8R:bbrdnrz) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e68bdc784e43084a74a77c86fbf7413daa2eafe/common/find_resource.cc#L261)):*
> ```cpp
>       return MakeResultFrom(resource_path, FindRunfile(resource_path));
>     } else {
>       log()->debug("FindResource ignoring runfiles with no Drake sentinel.");
> ```

btw, is "sentinel" a term of art for bazel runfiles? I would not understand what this log message means had I not have access to the definition of the variable `kSentinelRelpath` and the accompanying comment.

I think what would have helped me is a sentence about what's the significance of a runfile not having Drake sentinel (i.e. we are not building from source using bazel?)



_Code quote:_
```cpp
sentinel
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19206)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),xuchenhan-tri(platform) (waiting on @rpoyner-tri and @xuchenhan-tri)

___
*[`common/find_resource.cc` line 261 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19206#-NTKf1pC-kubYxWb_o8Q:-NTL0ohy6Ubf5-6Aqnqv:bmka10a) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e68bdc784e43084a74a77c86fbf7413daa2eafe/common/find_resource.cc#L261)):*

btw, is "sentinel" a term of art for bazel runfiles? I would not understand what this log message means had I not have access to the definition of the variable `kSentinelRelpath` and the accompanying comment.

I think what would have helped me is a sentence about what's the significance of a runfile not having Drake sentinel (i.e. we are not building from source using bazel?)
</blockquote></details>

The message it's at "debug" level, which basically means "copy-paste this output to a drake developer when reporting a problem" so I'm okay if it's a little bit obscure.  On the other hand, what would probably help the user (and what's done elsewhere in this file) is to print out the sentinel path when it's gone missing, so I've added that into the message here to help clarify what "sentinel file" is referring to.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19206)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri)

___
*[`common/find_resource.cc` line 261 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19206#-NTKf1pC-kubYxWb_o8Q:-NTL3RaOC2aLQjH3Nek5:b-rl00pv) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e68bdc784e43084a74a77c86fbf7413daa2eafe/common/find_resource.cc#L261)):*

The message it's at "debug" level, which basically means "copy-paste this output to a drake developer when reporting a problem" so I'm okay if it's a little bit obscure.  On the other hand, what would probably help the user (and what's done elsewhere in this file) is to print out the sentinel path when it's gone missing, so I've added that into the message here to help clarify what "sentinel file" is referring to.
</blockquote></details>

That does help, thanks.

PR #19218, opened 2023-04-18
Exposes SAP near rigid threshold

amcastro-tri:
This PR exposes the parameter that controls the "near rigid" approximation used by SAP.
This parameter is used to avoid ill-conditioning for very stiff contact models. It turns out that our default value is too conservative and often leads to softened contact constraints. This is specially true in models with large mass ratios, typically gripper fingers on a much massive robot. 

While we should probably set this parameter to zero by default (meaning no near rigid conditioning), this PR simply exposes this parameters so that they can set it to zero themselves in cases for which softening is observed.

Marking high priority since several users run into this kind of problems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -972,6 +972,14 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py::arg("contact_solver"), cls_doc.set_discrete_contact_solver.doc)
         .def("get_discrete_contact_solver", &Class::get_discrete_contact_solver,
             cls_doc.get_discrete_contact_solver.doc)
+        .def("set_sap_near_rigid_threshold",
+            &Class::set_sap_near_rigid_threshold,
+            py::arg("near_rigid_threshold") =
+                MultibodyPlantConfig{}.sap_near_rigid_threshold,
+            cls_doc.set_sap_near_rigid_threshold.doc)
+        .def("get_sap_near_rigid_threshold",
+            &Class::get_sap_near_rigid_threshold,
+            cls_doc.get_sap_near_rigid_threshold.doc)
         .def_static("GetDefaultContactSurfaceRepresentation",
             &Class::GetDefaultContactSurfaceRepresentation,
             py::arg("time_step"),
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -2451,6 +2451,8 @@ class TestPlant(unittest.TestCase):
         for model in models:
             plant.set_discrete_contact_solver(model)
             self.assertEqual(plant.get_discrete_contact_solver(), model)
+        plant.get_sap_near_rigid_threshold()
+        plant.set_sap_near_rigid_threshold(near_rigid_threshold=0.03)
 
     def test_contact_surface_representation(self):
         for time_step in [0.0, 0.1]:
--- a/multibody/contact_solvers/sap/sap_friction_cone_constraint.cc
+++ b/multibody/contact_solvers/sap/sap_friction_cone_constraint.cc
@@ -22,7 +22,7 @@ SapFrictionConeConstraint<T>::SapFrictionConeConstraint(int clique,
   DRAKE_DEMAND(p.mu >= 0.0);
   DRAKE_DEMAND(p.stiffness > 0.0);
   DRAKE_DEMAND(p.dissipation_time_scale >= 0.0);
-  DRAKE_DEMAND(p.beta > 0.0);
+  DRAKE_DEMAND(p.beta >= 0.0);
   DRAKE_DEMAND(p.sigma > 0.0);
   DRAKE_DEMAND(this->first_clique_jacobian().rows() == 3);
 }
@@ -40,7 +40,7 @@ SapFrictionConeConstraint<T>::SapFrictionConeConstraint(
   DRAKE_DEMAND(p.mu >= 0.0);
   DRAKE_DEMAND(p.stiffness > 0.0);
   DRAKE_DEMAND(p.dissipation_time_scale >= 0.0);
-  DRAKE_DEMAND(p.beta > 0.0);
+  DRAKE_DEMAND(p.beta >= 0.0);
   DRAKE_DEMAND(p.sigma > 0.0);
   DRAKE_DEMAND(this->first_clique_jacobian().rows() == 3);
   DRAKE_DEMAND(this->second_clique_jacobian().rows() == 3);
--- a/multibody/contact_solvers/sap/test/sap_friction_cone_constraint_test.cc
+++ b/multibody/contact_solvers/sap/test/sap_friction_cone_constraint_test.cc
@@ -45,7 +45,7 @@ GTEST_TEST(SapFrictionConeConstraint, SingleCliqueConstraint) {
   const double mu = 0.5;
   const double stiffness = 1.0e5;
   const double dissipation_time_scale = 0.01;
-  const double beta = 0.1;
+  const double beta = 0.0;
   const double sigma = 1.0e-4;
   const int clique = 12;
   const double phi0 = -2.5e-3;
@@ -71,7 +71,7 @@ GTEST_TEST(SapFrictionConeConstraint, TwoCliquesConstraint) {
   const double mu = 0.5;
   const double stiffness = 1.0e5;
   const double dissipation_time_scale = 0.01;
-  const double beta = 0.1;
+  const double beta = 0.0;
   const double sigma = 1.0e-4;
   const int clique0 = 12;
   const int clique1 = 13;
--- a/multibody/plant/compliant_contact_manager.cc
+++ b/multibody/plant/compliant_contact_manager.cc
@@ -970,7 +970,10 @@ void CompliantContactManager<T>::ExtractModelInfo() {
       // plant to symbolic and perform other supported queries such as
       // introspection and kinematics.
       if constexpr (!std::is_same_v<T, symbolic::Expression>) {
-        sap_driver_ = std::make_unique<SapDriver<T>>(this);
+        const double near_rigid_threshold =
+            plant().get_sap_near_rigid_threshold();
+        sap_driver_ =
+            std::make_unique<SapDriver<T>>(this, near_rigid_threshold);
       }
       break;
     case DiscreteContactSolver::kTamsi:
--- a/multibody/plant/deformable_driver.cc
+++ b/multibody/plant/deformable_driver.cc
@@ -1,5 +1,6 @@
 #include "drake/multibody/plant/deformable_driver.h"
 
+#include <limits>
 #include <memory>
 #include <string>
 #include <unordered_set>
@@ -244,12 +245,12 @@ void DeformableDriver<T>::AppendDiscreteContactPairs(
 
   for (const DeformableContactSurface<double>& surface :
        deformable_contact.contact_surfaces()) {
-    /* We use an arbitrarily large stiffness as the default stiffness so that
-     the contact is in near-rigid regime and the compliance is only used as
-     stabilization. */
-    const double default_contact_stiffness = 1.0e12;
-    const T k = GetCombinedPointContactStiffness(
-        surface.id_A(), surface.id_B(), default_contact_stiffness, inspector);
+    /* While our discrete solvers might model constraints as compliant, an
+    infinite stiffness indicates to use the stiffest approximation possible
+    without sacrifycing numerical conditioning. SAP will use the "near rigid"
+    regime approximation in this case. */
+    const T k = std::numeric_limits<double>::infinity();
+
     // TODO(xuchenhan-tri): Currently, body_B is guaranteed to be
     // non-deformable. When we support deformable vs. deformable contact, we
     // need to update this logic for retrieving body names.
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -336,6 +336,7 @@ MultibodyPlant<T>::MultibodyPlant(const MultibodyPlant<U>& other)
     num_collision_geometries_ = other.num_collision_geometries_;
     contact_model_ = other.contact_model_;
     contact_solver_enum_ = other.contact_solver_enum_;
+    sap_near_rigid_threshold_ = other.sap_near_rigid_threshold_;
     contact_surface_representation_ = other.contact_surface_representation_;
     // geometry_query_port_ is set during DeclareSceneGraphPorts() below.
     // geometry_pose_port_ is set during DeclareSceneGraphPorts() below.
@@ -600,6 +601,19 @@ template <typename T>
 DiscreteContactSolver MultibodyPlant<T>::get_discrete_contact_solver()
     const {
   return contact_solver_enum_;
+}
+
+template <typename T>
+void MultibodyPlant<T>::set_sap_near_rigid_threshold(
+    double near_rigid_threshold) {
+  DRAKE_MBP_THROW_IF_FINALIZED();
+  DRAKE_THROW_UNLESS(near_rigid_threshold >= 0.0);
+  sap_near_rigid_threshold_ = near_rigid_threshold;
+}
+
+template <typename T>
+double MultibodyPlant<T>::get_sap_near_rigid_threshold() const {
+  return sap_near_rigid_threshold_;
 }
 
 template <typename T>
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -1766,6 +1766,27 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   /// Returns the contact solver type used for discrete %MultibodyPlant models.
   DiscreteContactSolver get_discrete_contact_solver() const;
 
+  /// Non-negative dimensionless number typically in the range [0.0, 1.0],
+  /// though larger values are allowed even if uncommon. This parameter controls
+  /// the "near rigid" regime of the SAP solver, β in section V.B of [Castro et
+  /// al., 2021]. It essentially controls a threshold value for the maximum
+  /// amount of stiffness SAP can handle robustly. Beyond this value, stiffness
+  /// saturates as explained in [Castro et al., 2021]. A value of 1.0 is a
+  /// conservative choice to avoid ill-conditioning that might lead to softer
+  /// than expected contact. If this is your case, consider turning off this
+  /// approximation by setting this parameter to zero. For difficult cases where
+  /// ill-conditioning is a problem, a small but non-zero number can be used,
+  /// e.g. 1.0e-3.
+  /// @throws std::exception if near_rigid_threshold is negative.
+  /// @throws std::exception if called post-finalize.
+  void set_sap_near_rigid_threshold(
+      double near_rigid_threshold =
+          MultibodyPlantConfig{}.sap_near_rigid_threshold);
+
+  /// @returns the SAP near rigid regime threshold.
+  /// @see See set_sap_near_rigid_threshold().
+  double get_sap_near_rigid_threshold() const;
+
   /// Return the default value for contact representation, given the desired
   /// time step. Discrete systems default to use polygons; continuous systems
   /// default to use triangles.
@@ -5225,6 +5246,11 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   // with the default value in multibody_plant_config.h; there are already
   // assertions in the cc file that enforce this.
   DiscreteContactSolver contact_solver_enum_{DiscreteContactSolver::kTamsi};
+
+  // Near rigid regime parameter from [Castro et al., 2021]. Refer to
+  // set_near_rigid_threshold() for details.
+  double sap_near_rigid_threshold_{
+      MultibodyPlantConfig{}.sap_near_rigid_threshold};
 
   // User's choice of the representation of contact surfaces in discrete
   // systems. The default value is dependent on whether the system is
--- a/multibody/plant/multibody_plant_config.h
+++ b/multibody/plant/multibody_plant_config.h
@@ -21,6 +21,7 @@ struct MultibodyPlantConfig {
     a->Visit(DRAKE_NVP(stiction_tolerance));
     a->Visit(DRAKE_NVP(contact_model));
     a->Visit(DRAKE_NVP(discrete_contact_solver));
+    a->Visit(DRAKE_NVP(sap_near_rigid_threshold));
     a->Visit(DRAKE_NVP(contact_surface_representation));
     a->Visit(DRAKE_NVP(adjacent_bodies_collision_filters));
   }
@@ -52,6 +53,22 @@ struct MultibodyPlantConfig {
   /// - "tamsi"
   /// - "sap"
   std::string discrete_contact_solver{"tamsi"};
+
+  // TODO(amcastro-tri): Change default to zero, or simply eliminate.
+  /// Non-negative dimensionless number typically in the range [0.0, 1.0],
+  /// though larger values are allowed even if uncommon. This parameter controls
+  /// the "near rigid" regime of the SAP solver, β in section V.B of [Castro et
+  /// al., 2021]. It essentially controls a threshold value for the maximum
+  /// amount of stiffness SAP can handle robustly. Beyond this value, stiffness
+  /// saturates as explained in [Castro et al., 2021].
+  /// A value of 1.0 is a conservative choice to avoid numerical
+  /// ill-conditioning. However, this might introduce artificial softening of
+  /// the contact constraints. If this is your case try:
+  ///   1. Set this parameter to zero.
+  ///   2. For difficult problems (hundreds of contacts for instance), you might
+  ///      need to use a low value if the solver fails to converge.
+  ///      For instance, set values in the range (1e-3, 1e-2).
+  double sap_near_rigid_threshold{1.0};
 
   /// Configures the MultibodyPlant::set_contact_surface_representation().
   /// Refer to drake::geometry::HydroelasticContactRepresentation for details.
--- a/multibody/plant/multibody_plant_config_functions.cc
+++ b/multibody/plant/multibody_plant_config_functions.cc
@@ -22,6 +22,7 @@ AddResult AddMultibodyPlant(
   result.plant.set_discrete_contact_solver(
       internal::GetDiscreteContactSolverFromString(
           config.discrete_contact_solver));
+  result.plant.set_sap_near_rigid_threshold(config.sap_near_rigid_threshold);
   result.plant.set_contact_surface_representation(
       internal::GetContactSurfaceRepresentationFromString(
           config.contact_surface_representation));
--- a/multibody/plant/sap_driver.cc
+++ b/multibody/plant/sap_driver.cc
@@ -38,8 +38,11 @@ namespace multibody {
 namespace internal {
 
 template <typename T>
-SapDriver<T>::SapDriver(const CompliantContactManager<T>* manager)
-    : manager_(manager) {
+SapDriver<T>::SapDriver(const CompliantContactManager<T>* manager,
+                        double near_rigid_threshold)
+    : manager_(manager), near_rigid_threshold_(near_rigid_threshold) {
+  DRAKE_DEMAND(manager != nullptr);
+  DRAKE_DEMAND(near_rigid_threshold >= 0.0);
   // Collect joint damping coefficients into a vector.
   joint_damping_ = VectorX<T>::Zero(plant().num_velocities());
   for (JointIndex j(0); j < plant().num_joints(); ++j) {
@@ -160,7 +163,6 @@ std::vector<RotationMatrix<T>> SapDriver<T>::AddContactConstraints(
   // Parameters used by SAP to estimate regularization, see [Castro et al.,
   // 2021].
   // TODO(amcastro-tri): consider exposing these parameters.
-  constexpr double beta = 1.0;
   constexpr double sigma = 1.0e-3;
 
   const std::vector<DiscreteContactPair<T>>& contact_pairs =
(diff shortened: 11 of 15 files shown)


amcastro-tri (commented):
+@xuchenhan-tri for feature review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-:-NTJytD6E_qoPCVWyGzr:bzgdhpr)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
:lgtm: with some concerns about defaulting to zero.

> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-:-NTKOHl05iSHzpdgzku1:b-du6zgf)* status: 8 unresolved discussions, needs at least two assigned reviewers (waiting on @amcastro-tri)

___
*[`multibody/plant/multibody_plant.h` line 1776 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOzUT9uT39FfnNr__:-NTKOzUT9uT39FfnNr_a:bia579c) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant.h#L1776)):*
> ```objc
>   /// ill-conditioning is a problem, a small but non-zero number can be used,
>   /// e.g. 1.0e-3.
>   /// @throws std::exception iff called post-finalize.
> ```

Should also throw for illegal values.



_Code quote:_
```objc
/// @throws std::exception iff called post-finalize.
```
___
*[`multibody/plant/multibody_plant.cc` line 610 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOj7V6_CBqo0Nm4ex:-NTKOj7V6_CBqo0Nm4ey:b-t1aqcc) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant.cc#L610)):*
> ```cpp
>     double near_rigid_parameter) {
>   DRAKE_MBP_THROW_IF_FINALIZED();
>   sap_near_rigid_parameter_ = near_rigid_parameter;
> ```

We should guard against illegal values here.
___
*[`multibody/plant/multibody_plant_config.h` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKLMW-5VFGvJ-1NebG:-NTKLMW-5VFGvJ-1NebH:b-b7l2ad) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant_config.h#L64)):*
> ```objc
>   /// is your case try:
>   ///   1. Set this parameter to zero.
>   ///   2. For difficult problems (hundreds of contacts for instance), you might
> ```

nit, can we mention this recipe in the error message when sap_solver fails to converge?
___
*[`multibody/plant/sap_driver.h` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKJmelAvl6ancXics3:-NTKJmelAvl6ancXics4:bwornn3) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.h#L64)):*
> ```objc
>   // services to perform solver-agnostic multibody computations, e.g. contact
>   // kinematics. The given `manager` must outlive this driver.
>   // @pre manager != nullptr.
> ```

nit, Do we still need this constructor? Seems like we can just give the two arg constructor a default value.
___
*[`multibody/plant/sap_driver.h` line 230 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKKB_011_h80eOhTL4:-NTKKB_011_h80eOhTL5:b4ben8g) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.h#L230)):*
> ```objc
>   // Near rigid regime parameter for contact constraints. A value of 1.0 is a
>   // conservative choice to avoid ill-conditioning.
>   double near_rigid_parameter_{1.0};
> ```

Are there plans to allow this parameter to be different for different constraints? Right now deformable contacts sets an arbitrarily large number of constraints and rely on a non-zero value of beta.

For example, with the `deformable_torus` example currently in master, if we set beta to zero, the solver blows up (and in a somewhat unexpected way)
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  NewtonWithBisectionFallback did not converge.
|dx| = 2.7755575615628914e-17. x_lower = 0.3259535186072462. x_upper = 0.32595351860724625. x = 0.3259535186072462. f = -4.5449060514348925e-07.
Aborted (core dumped)```
```

___
*[`multibody/plant/sap_driver.cc` line 43 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOslR4_SUM_L2ayAk:-NTKOslR4_SUM_L2ayAl:b-t1aqcc) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.cc#L43)):*
> ```cpp
> SapDriver<T>::SapDriver(const CompliantContactManager<T>* manager,
>                         double near_rigid_parameter)
>     : manager_(manager), near_rigid_parameter_(near_rigid_parameter) {
> ```

We should guard against illegal values here.
___
*[`multibody/plant/sap_driver.cc` line 220 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKMXg8-szWCi1X6t-u:-NTKMXg8-szWCi1X6t-v:bed1unt) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.cc#L220)):*
> ```cpp
>   // TODO(amcastro-tri): consider exposing these parameters.
>   // "Near-rigid" parameter. See [Castro et al., 2021].
>   constexpr double kBeta = 0.1;
> ```

BTW, do we plan to default beta's for other types of constraints to 0 as well?
___
*[`multibody/plant/test/sap_driver_test.cc` line 188 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKKVvs0omJ1ipb3GQf:-NTKKVvs0omJ1ipb3GQg:b4n277i) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/test/sap_driver_test.cc#L188)):*
> ```cpp
>     EXPECT_EQ(constraint->parameters().dissipation_time_scale,
>               discrete_pair.dissipation_time_scale);
>     EXPECT_EQ(constraint->parameters().beta,
> ```

BTW, do we want to change all "beta"s to "near_rigid_parameter"?



_Code quote:_
```cpp
beta
```

amcastro-tri (commented):
> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 6 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 1776 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOzUT9uT39FfnNr__:-NTKxe3sDJZuilF0TDjN:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant.h#L1776)):*

Should also throw for illegal values.
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 610 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOj7V6_CBqo0Nm4ex:-NTKxg7qBLTz0-feScb6:bombmv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant.cc#L610)):*

We should guard against illegal values here.
</blockquote></details>

Good call. Done.
___
*[`multibody/plant/multibody_plant_config.h` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKLMW-5VFGvJ-1NebG:-NTKyb6T6CU82IJDCVVB:bcvjgzp) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/multibody_plant_config.h#L64)):*

nit, can we mention this recipe in the error message when sap_solver fails to converge?
</blockquote></details>

Good idea. Done.
___
*[`multibody/plant/sap_driver.h` line 230 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKKB_011_h80eOhTL4:-NTLBaLq6gqzyxHvngbF:b-omh6ds) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.h#L230)):*

Are there plans to allow this parameter to be different for different constraints? Right now deformable contacts sets an arbitrarily large number of constraints and rely on a non-zero value of beta.

For example, with the `deformable_torus` example currently in master, if we set beta to zero, the solver blows up (and in a somewhat unexpected way)
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  NewtonWithBisectionFallback did not converge.
|dx| = 2.7755575615628914e-17. x_lower = 0.3259535186072462. x_upper = 0.32595351860724625. x = 0.3259535186072462. f = -4.5449060514348925e-07.
Aborted (core dumped)```
```

</blockquote></details>

Good catch!
See if you like what I did. Essentially I am using `stiffness=infinity` to indicate these "rigid" constraints should be treated differently (with beta=1.0 by default). Then DeformableDriver::AppendDiscreteContactPairs() set stiffness=infinity for its constraints.

For all other constraint types we've been using `beta=0.1` pretty successfully, so I won't change for now.
___
*[`multibody/plant/sap_driver.cc` line 43 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKOslR4_SUM_L2ayAk:-NTLDHXv5xXPjHk1_Ceg:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.cc#L43)):*

We should guard against illegal values here.
</blockquote></details>

Done.
___
*[`multibody/plant/sap_driver.cc` line 220 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKMXg8-szWCi1X6t-u:-NTLDJ331j2i-JfEX8EM:b2t2cjb) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/sap_driver.cc#L220)):*

BTW, do we plan to default beta's for other types of constraints to 0 as well?
</blockquote></details>

thus far these haven't been a problem, particularly because beta is smaller (and the max stiffness goes with the square of beta).

I might consider even lowering this number. But I'll work first on the new SAP formulation. It has better numerics and I believe we'll be able to push these limits further.
___
*[`multibody/plant/test/sap_driver_test.cc` line 188 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTKKVvs0omJ1ipb3GQf:-NTLDo3u4mBU20qyrQWG:b-x65734) ([raw file](https://github.com/RobotLocomotion/drake/blob/b04ad8f6d1e8afe7dcddc0da80eff724a8f6e04c/multibody/plant/test/sap_driver_test.cc#L188)):*

BTW, do we want to change all "beta"s to "near_rigid_parameter"?
</blockquote></details>

Internally I actually like the shorter "beta" better.

amcastro-tri (commented):
Comments addressed. PTAL @xuchenhan-tri 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-:-NTLE8IJ5DhIYZBdNfhP:blgebol)* status: 6 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

jwnimmer-tri (commented):
> Reviewed 9 of 13 files at r1, 4 of 6 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @amcastro-tri)

___
*[`multibody/plant/multibody_plant.cc` line 610 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLLxFP1D3KNT1EBFbu:-NTLLxFP1D3KNT1EBFbv:btk10cc) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant.cc#L610)):*
> ```cpp
>     double near_rigid_parameter) {
>   DRAKE_MBP_THROW_IF_FINALIZED();
>   DRAKE_THROW_UNLESS(near_rigid_parameter >= 0.0);
> ```

The API docs give both a lower bound and upper bound on this argument.  We should check the upper bound here as well, I would think?  Or else maybe explain why not in a comment.
___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTLHXxf3vEFoXdyTnhv:b49put1) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*
> ```objc
>   ///      need to use a low value if the solver fails to converge.
>   ///      For instance, set values in the range (1e-3, 1e-2).
>   double sap_near_rigid_parameter{1.0};
> ```

As a naive reader, "parameter" here seems either redundant, or maybe overly non-specific.  Everything in this struct is a parameter.  Is this some kind of "parameter parameter"?   Is there any possibly better name we could use here?  I searched through the paper and wasn't able to find any definition of "near-rigid parameter" as a dimensionless quantity.
___
*[`multibody/plant/sap_driver.h` line 60 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLIwcX7TAODUgKM26X:-NTLIwcY1_AG8SbPjPEY:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L60)):*
> ```objc
>   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(SapDriver);
> 
>   // Constructs a driver with the provided `near_rigid_regime` parameter. This
> ```

typo



_Code quote:_
```objc
near_rigid_regime
```
___
*[`multibody/plant/sap_driver.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLInKxDARMW9CRTBL-:-NTLInKxDARMW9CRTBL0:bpx6lva) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L67)):*
> ```objc
>   // @pre manager is not nullptr.
>   // @pre near_rigid_parameter is positive.
>   SapDriver(const CompliantContactManager<T>* manager,
> ```

nit Because the all arguments but the first are defaulted, we still need `explicit`.



_Suggestion:_
```objc
  explicit SapDriver(const CompliantContactManager<T>* manager,
                     double near_rigid_parameter = 1.0);
```
___
*[`multibody/plant/sap_driver.h` line 223 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLK-el-60iR3c2plaH:-NTLK-el-60iR3c2plaI:bocgu40) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L223)):*
> A value of 1.0 is a conservative choice to avoid ill-conditioning.

This is useful information, but it's placement here with the next line misleads the reader into thinking that the `{1.0}` here is a default value.  It's not a default, its dead code because the constructor's initializer unconditionally overwrites whatever appears here.
___
*[`multibody/plant/sap_driver.h` line 225 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLJQejDHW-9cGJtC_f:-NTLJQejDHW-9cGJtC_g:bkqzxnr) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L225)):*
> ```objc
>   // Near rigid regime parameter for contact constraints. A value of 1.0 is a
>   // conservative choice to avoid ill-conditioning.
>   double near_rigid_parameter_{1.0};
> ```

BTW Consider being more clear about the lifecycle of this field:



_Suggestion:_
```objc
const double near_rigid_parameter_;
```
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2454 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLMh_2Bp_GQCo9IeAF:-NTLMh_2Bp_GQCo9IeAG:b9ck70m) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/bindings/pydrake/multibody/test/plant_test.py#L2454)):*
> ```Python
>             plant.set_discrete_contact_solver(model)
>             self.assertEqual(plant.get_discrete_contact_solver(), model)
>         # Defaults specified in MultibodyPlantConfig
> ```

BTW It's fine to keep this code I suppose, but you don't really need it.
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2457 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLMsISBM-MQySGgGKg:-NTLMsISBM-MQySGgGKh:b-czvu2e) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/bindings/pydrake/multibody/test/plant_test.py#L2457)):*
> ```Python
>         self.assertEqual(plant.get_sap_near_rigid_parameter(),
>                          MultibodyPlantConfig().sap_near_rigid_parameter)
>         plant.set_sap_near_rigid_parameter(0.03)
> ```

nit Missing kwarg name in function call.



_Suggestion:_
```Python
near_rigid_parameter=0.03
```

jwnimmer-tri (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @amcastro-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTOiRbM07pQ_G3ZZ6LL:bab0913) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*
How about `sap_near_rigid_beta` in that case?

> This is β in section V.B.

That seems worth adding into the API comment here nearby the paper cite, to help people navigate the paper?
___
*[`multibody/plant/sap_driver.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLInKxDARMW9CRTBL-:-NTOihbs-ogkV5Y_mHjC:bkc1a1v) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L67)):*

thanks! I was wondering about that but I figured clang format would complain?
</blockquote></details>

You mean `cpplint` -- and yes, sometimes I think it would but it doesn't always notice since it uses regexs to parse the code and those are imperfect.
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2454 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLMh_2Bp_GQCo9IeAF:-NTOj14V2-I2wGErOulB:bcbfjux) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/bindings/pydrake/multibody/test/plant_test.py#L2454)):*

could you expand @jwnimmer-tri? I was trying to at least smoke test the APIs. 
</blockquote></details>

(Be sure to expand the diff suggestion in reviewable.  It's a little unclear by default.)

I'm saying this would be sufficient test case, for example:

```python
plant.get_sap_near_rigid_parameter()
plant.set_sap_near_rigid_parameter(near_rigid_parameter=0.03)
```

That test code would sufficient to catch any plausible bugs, e.g., that the function name is mis-spelled in the bindings.  The fact that "get" returns what got "set" is covered by the C++ unit test.  For simple bindings, we don't need a _functional_ unit test in Python, just an _existence_ test.

xuchenhan-tri (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @amcastro-tri and @jwnimmer-tri)

___
*[`multibody/plant/multibody_plant.h` line 1769 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTPIHIvCTF1jfo9tabZ:-NTPIHIvCTF1jfo9tab_:b-vc4ppb) ([raw file](https://github.com/RobotLocomotion/drake/blob/ed2fbd689ea9fbbc9ffabe63060adcea3e022716/multibody/plant/multibody_plant.h#L1769)):*
> ```objc
>   DiscreteContactSolver get_discrete_contact_solver() const;
> 
>   /// Positive dimensionless number typically in the range [0.0, 1.0], though
> ```

nit non-negative



_Code quote:_
```objc
Positive
```
___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTPIS2H5WmJKOtt5yLX:b88gku0) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

How about `sap_near_rigid_beta` in that case?

> This is β in section V.B.

That seems worth adding into the API comment here nearby the paper cite, to help people navigate the paper?
</blockquote></details>

No comment on the naming part. Ditto on comment to cite the paper, ideally including the section number as you did in this thread.
___
*[`multibody/plant/multibody_plant_config.h` line 58 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTPIJTiDNXngkv6dQIm:-NTPIJTiDNXngkv6dQIn:bvnwty5) ([raw file](https://github.com/RobotLocomotion/drake/blob/ed2fbd689ea9fbbc9ffabe63060adcea3e022716/multibody/plant/multibody_plant_config.h#L58)):*
> ```objc
> 
>   // TODO(amcastro-tri): Change default to zero, or simply eliminate.
>   /// Positive dimensionless number typically in the range [0.0, 1.0], though
> ```

nit, non-negative



_Code quote:_
```objc
Positive
```

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTUPFxD6OFe7Tg_T4i9:b-rnrzan) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

No comment on the naming part. Ditto on comment to cite the paper, ideally including the section number as you did in this thread.
</blockquote></details>

Nice idea. I added reference to section and symbol in the paper.

I still personally prefer `near_rigid_parameter` to the contender option `sap_near_rigid_beta`.
___
*[`multibody/plant/sap_driver.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLInKxDARMW9CRTBL-:-NTUPa4I-GNxPXClwDKu:bgec3p0) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/sap_driver.h#L67)):*

You mean `cpplint` -- and yes, sometimes I think it would but it doesn't always notice since it uses regexs to parse the code and those are imperfect.
</blockquote></details>

ah, got it. Thanks.
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2454 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLMh_2Bp_GQCo9IeAF:-NTUPiFk6xf58vgWEE0Z:b8a8n4t) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/bindings/pydrake/multibody/test/plant_test.py#L2454)):*
> Be sure to expand the diff suggestion in reviewable.

Wow! I didn't realize it was hiding the diff by default! weird.

Done.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-:-NTURPpWALJsLx3_mFxV:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @amcastro-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTUREuF7r5WtIOSQ8e-:b-ej2d7l) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

Nice idea. I added reference to section and symbol in the paper.

I still personally prefer `near_rigid_parameter` to the contender option `sap_near_rigid_beta`.


</blockquote></details>

So long as you're sure.  It would be pretty difficult to change later.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @amcastro-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTV-1qbD8CyQjk4c1pH:b-dx89s6) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

So long as you're sure.  It would be pretty difficult to change later.
</blockquote></details>

Sherm gave me what I think it's a really good suggestion: `sap_near_rigid_threshold`
WDYT?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 1 unresolved discussion (waiting on @amcastro-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTV2dlt88DvdlBtP2fU:b6kklz7) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

Sherm gave me what I think it's a really good suggestion: `sap_near_rigid_threshold`
WDYT?
</blockquote></details>

I like it better just from an English perspective.

I don't know the math to understand if its communicates the facts better, but if you like it, I think it's a winner.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: 1 unresolved discussion (waiting on @amcastro-tri and @jwnimmer-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTV42daC21PxKJ7JGUw:bb73djm) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

I like it better just from an English perspective.

I don't know the math to understand if its communicates the facts better, but if you like it, I think it's a winner.
</blockquote></details>

SGTM

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant_config.h` line 67 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19218#-NTLHXxf3vEFoXdyTnhu:-NTV4kWx0mS9PQBCGB-g:b-bargnq) ([raw file](https://github.com/RobotLocomotion/drake/blob/d19f03f726d346770211b92c619b4de9ae57ce75/multibody/plant/multibody_plant_config.h#L67)):*

SGTM
</blockquote></details>

Thanks for the feedback. I made the change.

jwnimmer-tri (commented):
> Reviewed 11 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19218)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @amcastro-tri)

PR #19221, opened 2023-04-18
[release] mkdir and chkdir after checking URLs

EricCousineau-TRI:
Without this, running without wheels built doesn't do any work, but still creates a directory that needs to be deleted before running.

Negative test case:
```
# Run once. Let it fail.
$ bazel run //tools/release_engineering:download_release_candidate -- --timestamp 20380101 --version v10.0.0
# Run it again. On master, complains about directory existing. This PR, it will try again.
$ bazel run //tools/release_engineering:download_release_candidate -- --timestamp 20380101 --version v10.0.0
````

Positive test cases (from docstring)
```
$ bazel run //tools/release_engineering:download_release_candidate -- \
      --timestamp 20230315 --find-git-sha
$ bazel run //tools/release_engineering:download_release_candidate -- \
      --timestamp 20230315 --version v1.14.0
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19221)

### DIFF
--- a/tools/release_engineering/download_release_candidate.py
+++ b/tools/release_engineering/download_release_candidate.py
@@ -7,12 +7,12 @@ Use bazel to build the tool:
 Here's an example of how to obtain the git sha for the release.
 
   bazel run //tools/release_engineering:download_release_candidate -- \
-      --timestamp 20220303 --find-git-sha
+      --timestamp 20230315 --find-git-sha
 
 Here's an example of how download the release artifacts:
 
   bazel run //tools/release_engineering:download_release_candidate -- \
-      --timestamp 20220303 --version v1.0.0
+      --timestamp 20230315 --version v1.14.0
 
 TODO(jwnimmer-tri) Rename this tool to something more general, like
 `release_candidate` (without the "download" part).
@@ -107,9 +107,9 @@ def _check_urls(*, urls):
         raise UserError(error_message.getvalue())
 
 
-def _download_binaries(*, timestamp, staging, version):
-    """Downloads the binaries as specified, and returns a list of (relative)
-    paths.
+def _generate_and_check_download_urls(*, timestamp, staging, version):
+    """Generates URLs for the binaries as specified, and
+    (relative)paths.
 
     The `timestamp` is a string like "YYYYMMDD".
     The `staging` is a bool (whether to download staged wheels / debian).
@@ -153,8 +153,14 @@ def _download_binaries(*, timestamp, staging, version):
 
     # Make sure all can be downloaded (fail-fast).
     _check_urls(urls=download_urls)
+    return base_url_filename_pairs
 
-    # Download.
+
+def _download_binaries(base_url_filename_pairs):
+    """Downloads URLs given results of _generate_and_check_download_urls.
+
+    Returns a list of relative paths.
+    """
     result = []
     for base_url, filename in base_url_filename_pairs:
         _download_with_sha(base_url=base_url, filename=filename)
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI (commented):
+@jwnimmer-tri for review please! +(status: single reviewer ok) 
if you don't think it's worth the change, I can close the PR - either way from my perspective

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-:-NTKkNE63bMQkYH4ykOW:b53axnl)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @jwnimmer-tri)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19221)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @jwnimmer-tri)

___
*[`tools/release_engineering/download_release_candidate.py` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-NTKkhbhDyKsvVbgKP0n:-NTKkhbhDyKsvVbgKP0o:b-4hnbnn) ([raw file](https://github.com/RobotLocomotion/drake/blob/93df47090fbc370b721b820d68e00b58707d0ee5/tools/release_engineering/download_release_candidate.py#L14)):*
> ```Python
> Here's an example of how download the release artifacts:
> 
>   bazel run //tools/release_engineering:download_release_candidate -- \
> ```

BTW Ah, this doesn't work for old releases period 'cause staging?

jwnimmer-tri (commented):
I am fine to merge this if you like it as an improvement.  (For myself, I do `rm /tmp/thing` if I made any typos and need to recover.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-:-NTL2r4O8YclzeH3CfNv:b-xwq90k)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @EricCousineau-TRI)

___
*[`tools/release_engineering/download_release_candidate.py` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-NTKkhbhDyKsvVbgKP0n:-NTL1Y6Z132JeHaAWk8p:bb7867z) ([raw file](https://github.com/RobotLocomotion/drake/blob/93df47090fbc370b721b820d68e00b58707d0ee5/tools/release_engineering/download_release_candidate.py#L14)):*

BTW Ah, this doesn't work for old releases period 'cause staging?
</blockquote></details>

Yup.
___
*[`tools/release_engineering/download_release_candidate.py` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-NTL2JT_4nzICYXtOpw9:-NTL2JT_4nzICYXtOpwA:babl470) ([raw file](https://github.com/RobotLocomotion/drake/blob/a6ec4bd62faeb8e421d97c65edd081c918ee7298/tools/release_engineering/download_release_candidate.py#L112)):*
> ```Python
> def _generate_and_check_download_urls(*, timestamp, staging, version):
>     """Generates URLs for the binaries as specified, and
>     (relative)paths.
> ```

The new return type shape and its data format is not sufficiently well-explained.  We'd probably want a type hint (for the shape) and examples (of what it contains).

On the other hand, we probably don't need to pass `base_url_filename_pairs` around.  If we passed around the `binaries` dict instead, I think it would suffice for what the code needs, and be easier to explain here in terms of a return value.

jwnimmer-tri (commented):
:lgtm: both, modulo discussions.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19221#-:-NURww7O2ICp4BXsQHCS:b55ufd5)* status: 1 unresolved discussion, missing label for release notes (waiting on @EricCousineau-TRI)

PR #14272, opened 2020-11-01
workspace: Upgrade fmt to latest release 7.1.0

jwnimmer-tri:
> Changed the default floating point format to not include .0 for consistency with std::format and std::to_chars. It is possible to get the decimal point and trailing zero with the # specifier.

-- https://github.com/fmtlib/fmt/releases/tag/7.1.0

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/14272)

PR #19229, opened 2023-04-19
Extend DRAKE_THROW_UNLESS with value arguments

SeanCurtis-TRI:
DRAKE_THROW_UNLESS historically took a single parameter: a boolean expression serving as a throwing condition. if the condition was false, a message would print out the failed condition.

However, if the condition were expressed in lvalues, the message could be inscrutable:

    DRAKE_THROW_UNLESS(x > threshold)

would include the following information:

    "Condition 'x > threshold' failed."

with no insight as to what x or threshold are.

Now, up to four value expressions can be passed along with the condition:

    DRAKE_THROW_UNLESS(x > threshold, x, threshold)

Producing (something like):

    "Condition 'x > threshold' failed. x = -inf, threshold = 1e-09."

### DIFF
--- a/common/BUILD.bazel
+++ b/common/BUILD.bazel
@@ -801,8 +801,10 @@ drake_cc_googletest(
 drake_cc_googletest(
     name = "drake_throw_test",
     deps = [
+        ":autodiff",
         ":essential",
         "//common/test_utilities:expect_no_throw",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/common/drake_assert_and_throw.cc
+++ b/common/drake_assert_and_throw.cc
@@ -50,10 +50,23 @@ void Abort(const char* condition, const char* func, const char* file,
 }
 
 // Declared in drake_throw.h.
+// If the suffix should end with punctuation, the caller must provide it.
 void Throw(const char* condition, const char* func, const char* file,
-           int line) {
+           int line, std::string_view suffix) {
   std::ostringstream what;
   PrintFailureDetailTo(what, condition, func, file, line);
+  if (!suffix.empty()) {
+    what << " " << suffix;
+  }
+  throw assertion_error(what.str().c_str());
+}
+
+// Declared in drake_throw.h.
+void ThrowWithSuffix(const std::string& condition, const char* func,
+                     const char* file, int line, const std::string& suffix) {
+  std::ostringstream what;
+  PrintFailureDetailTo(what, condition.c_str(), func, file, line);
+  what << " " << suffix;
   throw assertion_error(what.str().c_str());
 }
 
--- a/common/drake_throw.h
+++ b/common/drake_throw.h
@@ -1,37 +1,104 @@
 #pragma once
 
+#include <string>
 #include <type_traits>
+#include <vector>
+
+#include <fmt/format.h>
 
 #include "drake/common/drake_assert.h"
+#include "drake/common/eigen_types.h"
+#include "drake/common/fmt_eigen.h"
 
-/// @file
-/// Provides a convenient wrapper to throw an exception when a condition is
-/// unmet.  This is similar to an assertion, but uses exceptions instead of
-/// ::abort(), and cannot be disabled.
+/** @file
+ Provides a convenient wrapper to throw an exception when a condition is
+ unmet.  This is similar to an assertion, but uses exceptions instead of
+ ::abort(), and cannot be disabled.
+ */
 
 namespace drake {
 namespace internal {
 // Throw an error message.
 [[noreturn]] void Throw(const char* condition, const char* func,
-                        const char* file, int line);
+                        const char* file, int line,
+                        std::string_view suffix = std::string_view());
+
+/* The following infrastructure lets us iterate through a list of macro variadic
+ arguments (up to length four).
+ See https://codecraft.co/2014/11/25/variadic-macros-tricks/ for explanation.
+*/
+
+// Support up to *four* variadic arguments.
+#define _GET_NTH_ARG(_1, _2, _3, _4, N, ...) N
+// Macros for encoding a value expression into its value. _e is short for
+// _encode.
+#define _e(s, x) ::drake::internal::RecordExpressionValue(#x, x, s);
+#define _e_0(...)
+#define _e_1(s, value) _e(s, value)
+#define _e_2(s, value, ...) _e(s, value) _e_1(s, __VA_ARGS__)
+#define _e_3(s, value, ...) _e(s, value) _e_2(s, __VA_ARGS__)
+#define _e_4(s, value, ...) _e(s, value) _e_3(s, __VA_ARGS__)
+
+#define ENCODE_EACH(s, ...)                                                   \
+    /* NOLINTNEXTLINE(whitespace/comma) */                                    \
+    _GET_NTH_ARG(__VA_ARGS__ __VA_OPT__(,) _e_4, _e_3, _e_2, _e_1, _e_0)      \
+        (s, __VA_ARGS__)
+
+
+// Records the spelling of an expression and its reported value as:
+//
+//   expr = value
+//
+// adding it to the vector of recorded expression values.
+template <typename T>
+void RecordExpressionValue(const char* expr, const T& value,
+                           std::vector<std::string>* recorded) {
+  std::string value_s;
+  if constexpr (::drake::is_eigen_type<T>::value) {
+    value_s = fmt::format("{}", fmt_eigen(value));
+  } else {
+    if constexpr (std::is_convertible_v<T, std::string>) {
+      // Give string-like values surrounding quotes.
+      value_s = fmt::format("\"{}\"", value);
+    } else {
+      value_s = fmt::format("{}", value);
+    }
+  }
+  recorded->emplace_back(fmt::format("{} = {}", expr, value_s));
+}
+
+[[noreturn]] void ThrowWithSuffix(const std::string& condition,
+                                  const char* func, const char* file, int line,
+                                  const std::string& suffix);
 }  // namespace internal
 }  // namespace drake
 
-/// Evaluates @p condition and iff the value is false will throw an exception
-/// with a message showing at least the condition text, function name, file,
-/// and line.
-///
-/// The condition must not be a pointer, where we'd implicitly rely on its
-/// nullness. Instead, always write out "!= nullptr" to be precise.
-///
-/// Correct: `DRAKE_THROW_UNLESS(foo != nullptr);`
-/// Incorrect: `DRAKE_THROW_UNLESS(foo);`
-///
-/// Because this macro is intended to provide a useful exception message to
-/// users, we should err on the side of extra detail about the failure. The
-/// meaning of "foo" isolated within error message text does not make it
-/// clear that a null pointer is the proximate cause of the problem.
-#define DRAKE_THROW_UNLESS(condition)                                         \
+/** Evaluates @p condition and iff the value is false will throw an exception
+ with a message showing at least the condition text, function name, file,
+ and line.
+
+ In addition to the @p condition, up to four value expressions can be provided.
+ Each value expression and its value will be included in the error message. For
+ example:
+
+    `DRAKE_THROW_UNLESS(v1.norm() == 1, v1, v1.norm());`
+
+ Will include the the values of the vector v1 as well as its norm in the
+ message. If too many value expressions are specified, this will most likely
+ produce a compiler error referencing "ENCODE_EACH".
+
+ The condition must not be a pointer, where we'd implicitly rely on its
+ nullness. Instead, always write out "!= nullptr" to be precise.
+
+ Correct: `DRAKE_THROW_UNLESS(foo != nullptr);`
+ Incorrect: `DRAKE_THROW_UNLESS(foo);`
+
+ Because this macro is intended to provide a useful exception message to
+ users, we should err on the side of extra detail about the failure. The
+ meaning of "foo" isolated within error message text does not make it
+ clear that a null pointer is the proximate cause of the problem.
+ */
+#define DRAKE_THROW_UNLESS(condition, ...)                                    \
   do {                                                                        \
     typedef ::drake::assert::ConditionTraits<                                 \
         typename std::remove_cv_t<decltype(condition)>> Trait;                \
@@ -42,6 +109,11 @@ namespace internal {
         "DRAKE_THROW_UNLESS(foo != nullptr), do not write DRAKE_THROW_UNLESS" \
         "(foo) and rely on implicit pointer-to-bool conversion.");            \
     if (!Trait::Evaluate(condition)) {                                        \
-      ::drake::internal::Throw(#condition, __func__, __FILE__, __LINE__);     \
+      std::vector<std::string> expr_strings;                                  \
+      ENCODE_EACH(&expr_strings, __VA_ARGS__);        \
+      ::drake::internal::Throw(                                               \
+          #condition, __func__, __FILE__, __LINE__,                           \
+          fmt::format("{}{}", fmt::join(expr_strings, ", "),                  \
+                      expr_strings.size() > 0 ? "." : ""));                   \
     }                                                                         \
   } while (0)
--- a/common/test/drake_throw_test.cc
+++ b/common/test/drake_throw_test.cc
@@ -1,16 +1,82 @@
 #include "drake/common/drake_throw.h"
 
 #include <stdexcept>
+#include <string>
 
 #include <gtest/gtest.h>
 
 #include "drake/common/test_utilities/expect_no_throw.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 
 namespace {
 
 GTEST_TEST(DrakeThrowTest, BasicTest) {
   DRAKE_EXPECT_NO_THROW(DRAKE_THROW_UNLESS(true));
   EXPECT_THROW(DRAKE_THROW_UNLESS(false), std::runtime_error);
+}
+
+// The test tests two things:
+//
+//   1. Provided value expressions get encoded as expected.
+//   2. Regression guaranteeing that four value expressions are supported.
+//
+// If we have hiccups with value expressions of various types, these tests can
+// be extended to explicitly test those types.
+GTEST_TEST(DrakeThrowTest, ThrowWithValues) {
+  // Create a collection of lvalues.
+  const std::string msg = "a string";
+  const double d = 7;
+  const int i = 13;
+  const char bytes[] = "beef";
+  // const Eigen::Vector3d bytes(1, 2, 3);
+
+  // Ensure it works with up to four values.
+  DRAKE_EXPECT_NO_THROW(DRAKE_THROW_UNLESS(true, d));
+  DRAKE_EXPECT_NO_THROW(DRAKE_THROW_UNLESS(true, d, i));
+  DRAKE_EXPECT_NO_THROW(DRAKE_THROW_UNLESS(true, d, i, msg));
+  DRAKE_EXPECT_NO_THROW(DRAKE_THROW_UNLESS(true, d, i, msg, bytes));
+
+  // Note: Our DRAKE_EXPECT_THROWS_MESSAGE doesn't like the do {} while(0) body
+  // of the DRAKE_THROW_UNLESS expression. To placate it, we place the
+  // invocation in a lambda to make things happier.
+  auto f1 = [&](int count) {
+    switch (count) {
+      case 1:
+        DRAKE_THROW_UNLESS(false, d);
+        return;
+      case 2:
+        DRAKE_THROW_UNLESS(false, d, msg);
+        return;
+      case 3:
+        DRAKE_THROW_UNLESS(false, d, msg, i);
+        return;
+      case 4:
+        DRAKE_THROW_UNLESS(false, d, msg, i, bytes);
+        return;
+    }
+    DRAKE_UNREACHABLE();
+  };
+
+  DRAKE_EXPECT_THROWS_MESSAGE(f1(1), ".*d = 7.");
+  DRAKE_EXPECT_THROWS_MESSAGE(f1(2), ".*d = 7, msg = \"a string\".");
+  DRAKE_EXPECT_THROWS_MESSAGE(f1(3), ".*d = 7, msg = \"a string\", i = 13.");
+  DRAKE_EXPECT_THROWS_MESSAGE(
+      f1(4), ".*d = 7, msg = \"a string\", i = 13, bytes = \"beef\".");
+}
+
+GTEST_TEST(DrakeThrowTest, ThrowWithEigenValues) {
+  const std::string details = "These are details";
+  const Eigen::Vector3d v1(1, 2, 3);
+
+  auto f1 = [&]() {
+    DRAKE_THROW_UNLESS(false, v1, details);
+  };
+  DRAKE_EXPECT_THROWS_MESSAGE(f1(), ".*v1 = 1\n2\n3, details = .*");
+
+  auto f2 = [&]() {
+    DRAKE_THROW_UNLESS(false, v1.transpose(), details);
+  };
+  DRAKE_EXPECT_THROWS_MESSAGE(f2(), ".*v1.transpose.. = 1 2 3, details = .*");
 }
 
 }  // namespace
--- a/geometry/shape_specification.cc
+++ b/geometry/shape_specification.cc
@@ -1,6 +1,7 @@
 #include "drake/geometry/shape_specification.h"
 
 #include <algorithm>
+#include <cmath>
 #include <filesystem>
 #include <limits>
 
@@ -13,6 +14,13 @@
 
 namespace drake {
 namespace geometry {
+namespace {
+
+bool is_positive_finite(double value) {
+  return std::isfinite(value) && value > 0;
+}
+
+}
 
 using math::RigidTransform;
 
@@ -44,12 +52,9 @@ Shape::Shape(ShapeTag<S>) {
 Box::Box(double width, double depth, double height)
     : Shape(ShapeTag<Box>()),
       size_(width, depth, height) {
-  if (width <= 0 || depth <= 0 || height <= 0) {
-    throw std::logic_error(
-        fmt::format("Box width, depth, and height should all be > 0 (were {}, "
-                    "{}, and {}, respectively).",
-                    width, depth, height));
-  }
+  DRAKE_THROW_UNLESS(is_positive_finite(width), width);
+  DRAKE_THROW_UNLESS(is_positive_finite(depth), depth);
+  DRAKE_THROW_UNLESS(is_positive_finite(height), height);
 }
 
 Box::Box(const Vector3<double>& measures)
@@ -61,12 +66,8 @@ Box Box::MakeCube(double edge_size) {
 
 Capsule::Capsule(double radius, double length)
     : Shape(ShapeTag<Capsule>()), radius_(radius), length_(length) {
-  if (radius <= 0 || length <= 0) {
-    throw std::logic_error(
-        fmt::format("Capsule radius and length should both be > 0 (were {} "
-                    "and {}, respectively).",
-                    radius, length));
-  }
+  DRAKE_THROW_UNLESS(is_positive_finite(radius), radius);
+  DRAKE_THROW_UNLESS(is_positive_finite(length), length);
 }
 
 Capsule::Capsule(const Vector2<double>& measures)
@@ -76,21 +77,15 @@ Convex::Convex(const std::string& filename, double scale)
     : Shape(ShapeTag<Convex>()),
       filename_(std::filesystem::absolute(filename)),
       scale_(scale) {
-  if (std::abs(scale) < 1e-8) {
-    throw std::logic_error("Convex |scale| cannot be < 1e-8.");
-  }
+  DRAKE_THROW_UNLESS(std::isfinite(scale) && std::abs(scale) >= 1e-8, scale);
 }
 
 Cylinder::Cylinder(double radius, double length)
     : Shape(ShapeTag<Cylinder>()),
       radius_(radius),
       length_(length) {
-  if (radius <= 0 || length <= 0) {
-    throw std::logic_error(
-        fmt::format("Cylinder radius and length should both be > 0 (were {} "
-                    "and {}, respectively).",
-                    radius, length));
-  }
+  DRAKE_THROW_UNLESS(is_positive_finite(radius), radius);
+  DRAKE_THROW_UNLESS(is_positive_finite(length), length);
 }
 
 Cylinder::Cylinder(const Vector2<double>& measures)
@@ -98,12 +93,9 @@ Cylinder::Cylinder(const Vector2<double>& measures)
 
 Ellipsoid::Ellipsoid(double a, double b, double c)
     : Shape(ShapeTag<Ellipsoid>()), radii_(a, b, c) {
-  if (a <= 0 || b <= 0 || c <= 0) {
-    throw std::logic_error(
-        fmt::format("Ellipsoid lengths of principal semi-axes a, b, and c "
-                    "should all be > 0 (were {}, {}, and {}, respectively).",
-                    a, b, c));
-  }
+  DRAKE_THROW_UNLESS(is_positive_finite(a), a);
+  DRAKE_THROW_UNLESS(is_positive_finite(b), b);
+  DRAKE_THROW_UNLESS(is_positive_finite(c), c);
 }
 
 Ellipsoid::Ellipsoid(const Vector3<double>& measures)
@@ -146,19 +138,14 @@ Mesh::Mesh(const std::string& filename, double scale)
     : Shape(ShapeTag<Mesh>()),
       filename_(std::filesystem::absolute(filename)),
       scale_(scale) {
-  if (std::abs(scale) < 1e-8) {
-    throw std::logic_error("Mesh |scale| cannot be < 1e-8.");
-  }
+  DRAKE_THROW_UNLESS(std::isfinite(scale) && std::abs(scale) >= 1e-8, scale);
 }
 
 MeshcatCone::MeshcatCone(double height, double a, double b)
     : Shape(ShapeTag<MeshcatCone>()), height_(height), a_(a), b_(b) {
-  if (height <= 0 || a <= 0 || b <= 0) {
-    throw std::logic_error(fmt::format(
-        "MeshcatCone parameters height, a, and b should all be > 0 (they were "
-        "{}, {}, and {}, respectively).",
-        height, a, b));
-  }
+  DRAKE_THROW_UNLESS(is_positive_finite(height), height);
+  DRAKE_THROW_UNLESS(is_positive_finite(a), a);
+  DRAKE_THROW_UNLESS(is_positive_finite(b), b);
 }
 
 MeshcatCone::MeshcatCone(const Vector3<double>& measures)
@@ -166,10 +153,7 @@ MeshcatCone::MeshcatCone(const Vector3<double>& measures)
 
 Sphere::Sphere(double radius)
     : Shape(ShapeTag<Sphere>()), radius_(radius) {
-  if (radius < 0) {
-    throw std::logic_error(
-        fmt::format("Sphere radius should be >= 0 (was {}).", radius));
-  }
+  DRAKE_THROW_UNLESS(std::isfinite(radius) && radius >= 0, radius);
 }
 
 ShapeReifier::~ShapeReifier() = default;
--- a/geometry/shape_specification.h
+++ b/geometry/shape_specification.h
@@ -112,17 +112,17 @@ class Box final : public Shape {
   /** Constructs a box with the given `width`, `depth`, and `height`, which
    specify the box's dimension along the canonical x-, y-, and z-axes,
    respectively.
-   @throws std::exception if `width`, `depth` or `height` are not strictly
-   positive. */
+   @throws std::exception if any measure is not finite positive. */
   Box(double width, double depth, double height);
 
   /** Constructs a box with a vector of measures: width, depth, and height --
    the box's dimensions along the canonical x-, y-, and z-axes, respectively.
-   @throws std::exception if the measures are not strictly positive. */
+   @throws std::exception if any measures is not finite positive. */
   explicit Box(const Vector3<double>& measures);
 
   /** Constructs a cube with the given `edge_size` for its width, depth, and
-   height. */
+   height.
+   @throw std::exception if edge_Size is not finite positive. */
   static Box MakeCube(double edge_size);
 
   /** Returns the box's dimension along the x axis. */
@@ -153,12 +153,12 @@ class Capsule final : public Shape {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Capsule)
 
   /** Constructs a capsule with the given `radius` and `length`.
-   @throws std::exception if `radius` or `length` are not strictly positive.
+   @throws std::exception if any measure is not finite positive.
    */
   Capsule(double radius, double length);
 
   /** Constructs a capsule with a vector of measures: radius and length.
-   @throws std::exception if the measures are not strictly positive. */
+   @throws std::exception if any measure is not finite positive. */
   explicit Capsule(const Vector2<double>& measures);
 
   double radius() const { return radius_; }
@@ -216,12 +216,11 @@ class Cylinder final : public Shape {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Cylinder)
 
   /** Constructs a cylinder with the given `radius` and `length`.
-   @throws std::exception if `radius` or `length` are not strictly positive.
-   */
+   @throws std::exception if any measure is not finite positive. */
   Cylinder(double radius, double length);
 
   /** Constructs a cylinder with a vector of measures: radius and length.
-   @throws std::exception if the measures are not strictly positive. */
+   @throws std::exception if any measure is not finite positive. */
   explicit Cylinder(const Vector2<double>& measures);
 
   double radius() const { return radius_; }
@@ -248,14 +247,13 @@ class Ellipsoid final : public Shape {
   /** Constructs an ellipsoid with the given lengths of its principal
    semi-axes, with a, b, and c measured along the x-, y-, and z- axes of the
    canonical frame, respectively.
-   @throws std::exception if `a`, `b`, or `c` are not strictly positive.
-   */
+   @throws std::exception if any measure is not finite positive. */
   Ellipsoid(double a, double b, double c);
 
   /** Constructs an ellipsoid with a vector of measures: the lengths of its
    principal semi-axes, with a, b, and c measured along the x-, y-, and z- axes
    of the canonical frame, respectively.
-   @throws std::exception if the measures are not strictly positive. */
+   @throws std::exception if any measure is not finite positive. */
   explicit Ellipsoid(const Vector3<double>& measures);
 
   double a() const { return radii_(0); }
@@ -348,13 +346,12 @@ class MeshcatCone final : public Shape {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(MeshcatCone)
 
   /** Constructs the parameterized cone.
-   @throws std::exception if `height`, `a`, or `b` are not strictly positive.
-   */
+   @throws std::exception if any measure is not finite positive. */
   explicit MeshcatCone(double height, double a = 1.0, double b = 1.0);
 
   /** Constructs a cone with a vector of measures: height and principal
    semi-axes.
-   @throws std::exception if the measures are not strictly positive. */
+   @throws std::exception if any measure is not finite positive. */
   explicit MeshcatCone(const Vector3<double>& measures);
 
   double height() const { return height_; }
@@ -374,8 +371,8 @@ class Sphere final : public Shape {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Sphere)
 
   /** Constructs a sphere with the given `radius`.
-   @throws std::exception if `radius` is negative. Note that a zero radius is
-   considered valid. */
+   @throws std::exception if `radius` is not finite *non-negative*. Note that a
+                              zero radius is considered valid. */
   explicit Sphere(double radius);
 
   double radius() const { return radius_; }
--- a/geometry/test/shape_specification_test.cc
+++ b/geometry/test/shape_specification_test.cc
@@ -17,6 +17,8 @@ namespace drake {
 namespace geometry {
 namespace {
 
+using Eigen::Vector2d;
+using Eigen::Vector3d;
 using math::RigidTransformd;
 using std::unique_ptr;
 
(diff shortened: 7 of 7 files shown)


SeanCurtis-TRI:

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ2lvx55xWQ7yvZ4jQ:-NTQ2lvx55xWQ7yvZ4jR:b-jj5gtj) (no related file):*
This uses `__VA_OPT__`; need to see what kind of support we have across our various compilers.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-NTQ32sf1XAkl32yuuKm:b-4hfim3) (no related file):*
This brings invocations of `fmt::format` into the macro. There's roughly a 3:1 usage pattern of `DRAKE_THROW_UNLESS` in .cc vs .h files. Dragging formatting into header files is probably a bad thing.

Discussion: rather than "overloading" `DRAKE_THROW_UNLESS`, should we use a different symbol which must de jure only be used in .cc files?

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ3f5XAC_pH380zNHS:-NTQ3f5Y6EOJ2PSqc5o-:ba7aqrn) (no related file):*
Discussion: An earlier attempt I made to give explicit feedback about "too many value expressions" was *not* compatible with passing value expressions that created temporaries (e.g., `v.transpose()`). I surrendered the compile-time feedback in favor of that compatibility and simply rely strong representations in the documentation. If anyone has an epiphany to bring this back, I'm all ears.

___

jwnimmer-tri:
When you're ready for a partner here, you can assign me and we can co-iterate.  I can help with the question you posed above, and also there will be things that need repair -- `DRAKE_THROW_UNLESS` needs to be suitable for inline functions so we need to bifurcate this implementation somewhat.

SeanCurtis-TRI (commented):
I'll let the basic CI get through and then ping you.

Yeah, I figure I'll end up leaving `DRAKE_THROW_UNLESS` alone and make this `DRAKE_THROW_UNLESS_VALUES` that can only be used in .cc files (or some such thing).

I'd also like to role `DRAKE_THROW_UNLESS_DESCRIP` in short order. That's the one that takes two arguments: condition and something compatible with a `std::string_view`. That variant is much easier to implement than this variadic monstrosity and doesn't need specific admonition about whether it can be used inline or not because it will be apparent at the call site.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-:-NTQ5D2iClGQYlkKD_Qt:bag30q8)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri:

This brings invocations of `fmt::format` into the macro. There's roughly a 3:1 usage pattern of `DRAKE_THROW_UNLESS` in .cc vs .h files. Dragging formatting into header files is probably a bad thing.

Discussion: rather than "overloading" `DRAKE_THROW_UNLESS`, should we use a different symbol which must de jure only be used in .cc files?
</blockquote></details>

This is a non-issue.  Basically everything in Drake includes `fmt/format.h` already.  If we needed something beyond like `fmt/chrono.h` here that might be another story, but `format.h` is just fine.

SeanCurtis-TRI (commented):
Ok...the gods are simply angry! A quick `bazel build //...` will reveal that.

My best diagnosis is that the extra headers into `drake_throw.h` seems to be making other contexts *angry*.

@jwnimmer-tri I'll tag you in now. I'm going to start troubleshooting, but appreciate any cycles you can contribute to the cause.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-:-NTQ6VjxAX2NKxbFQtLf:bdqaaxr)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri:
I'll look, but at background priority so probably not today.

SeanCurtis-TRI (commented):
Ok, glx vs eigen battle was introduced. Simply by re-ordering headers in `internal_opengl_context.cc` fixed that collision. I'm going to let CI continue on the "bad" version just to give focal and mac a change to complain about `__VA_OPT__`. I'll push my local fix after that.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-:-NTQ8SneEAboRkGhHCrM:b-jdjkff)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-NTQ8t_lATlPEB6zmZDS:bm5zqdk) (no related file):*

This is a non-issue.  Basically everything in Drake includes `fmt/format.h` already.  If we needed something beyond like `fmt/chrono.h` here that might be another story, but `format.h` is just fine.
</blockquote></details>

Sounds good.

However, passing value expressions causes *invocations* of `fmt::format` and (possibly) `fmt::join`, so that overload should probably be limited to .cc files. And with that kind of rule, that emphasizes that need to split it out into its own macro. It's much easier to say `DRAKE_THROW_UNLESS` can be used anywhere and `DRAKE_THROW_UNLESS_VALUES` should only be used in .cc files.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-NTQAp1vAhCEnK0eNwar:b8k88ap) (no related file):*

Sounds good.

However, passing value expressions causes *invocations* of `fmt::format` and (possibly) `fmt::join`, so that overload should probably be limited to .cc files. And with that kind of rule, that emphasizes that need to split it out into its own macro. It's much easier to say `DRAKE_THROW_UNLESS` can be used anywhere and `DRAKE_THROW_UNLESS_VALUES` should only be used in .cc files.
</blockquote></details>

The most important invariant is that the _macro body itself_ cannot bloat the object code _of the already-inline function it's called from_.  Nothing about that prevents the macro from calling a helper function that calls fmt.  We just can't call fmt from directly within the macro.

___

SeanCurtis-TRI (commented):
mac (which wasn't affected by the glx vs eigen thing) passed. Yay!

It doesn't seem that any of the focal complained about using `__VA_OPT__`, so I've updated the branch with the expectation that this newest version will pass CI. I won't retrigger mac because nothing of import to the mac has really changed. I don't particularly have any worries about the jammy-clang combo, but can hit that just to be safe.

I'll leave this in draft mode and when you've got the cycles to visit this, we'll iterate on it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-:-NTQMsd84vmXHpD8e-lG:blvldj5)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-NTQNCTB9rX4-1QD35lF:b-wt189n) (no related file):*
> We just can't call fmt from directly within the macro.

Technically, I'm calling a function (itself defined in the header) that calls `fmt` from within the macro (as many as four times). So, I consider that having already offended.

I look forward to seeing what ideas you might have to make it more compliant. Handling the value expression string, the value expression type, and the conversion of that value into a string left me painted into a corner. I couldn't figure out how to get formatting operations out of the header and into the .cc file.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTTt2XY2Qt3eCHg14B3:-NTTt2XY2Qt3eCHg14B4:b-khs2w0) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/common/drake_throw.h#L45)):*
> ```objc
>         "(foo) and rely on implicit pointer-to-bool conversion.");            \
>     if (!Trait::Evaluate(condition)) {                                        \
>       ::drake::internal::Throw(#condition, __func__, __FILE__, __LINE__);     \
> ```

The ideal goal here is to make a _single_ function call to a function marked `noreturn`.

That means the function would be templated (on the extra value types) and also variadic (`template <typename Extras...>`).  It would also take a variadic number of `const char*` or similar, for the stringification of the extra values.

Within that function is where we should zip up the extra names and formatted values, and then call into the cc function that assembles the full message and throws.

SeanCurtis-TRI:
*[`common/drake_throw.h` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTTt2XY2Qt3eCHg14B3:-NTU629h7Zo9eYw1t2Qn:b2rlas3) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/common/drake_throw.h#L45)):*

The ideal goal here is to make a _single_ function call to a function marked `noreturn`.

That means the function would be templated (on the extra value types) and also variadic (`template <typename Extras...>`).  It would also take a variadic number of `const char*` or similar, for the stringification of the extra values.

Within that function is where we should zip up the extra names and formatted values, and then call into the cc function that assembles the full message and throws.
</blockquote></details>

Let me ponder an alternative to get around the macro variadic limit...I think I *might* be able to do *something*.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

___
*[`common/drake_throw.h` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTTt2XY2Qt3eCHg14B3:-NTVgeoMDTg1xUNcdCaf:b-pyxs1) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/common/drake_throw.h#L45)):*

The good news is that's how I interpreted it.  :) I've got an idea for making it work. 
</blockquote></details>

Ok, new version that I believe better satisfies the needs. PTAL

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/internal_opengl_context.cc` line 19 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTViDye1WSkkSb-JkOp:-NTViDye1WSkkSb-JkOq:b-qrf9xs) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/geometry/render_gl/internal_opengl_context.cc#L19)):*
> ```cpp
> // opengl_includes.h. See opengl_context.h for where pimpl is applied.
> // This *must* follow the drake includes in case any of those bring in
> // drake/common/eigen_types.h. The two headers interfer if Eigen is second.
> ```

nit typo



_Suggestion:_
```cpp
interfere
```

jwnimmer-tri (commented):
> Reviewed 1 of 7 files at r1, 2 of 2 files at r2, 2 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 35 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVwFo28ZTboLDe38L4:-NTVwFo28ZTboLDe38L5:b-78xgsq) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L35)):*
> ```objc
>     // This should be a const char* value expression; simply return it.
>     return value;
>   } else  if constexpr (::drake::is_eigen_type<T>::value) {
> ```

The full `eigen_types.h` header is pretty fat.  Probably we should copy-paste the tiny snippet we need inline here, and drop the extra include.

Ah, or perhaps a better idea -- calling `DRAKE_THROW_UNLESS(!M.isZero(), M);` is out of scope.  If you want that, you need to write `DRAKE_THROW_UNLESS(!M.isZero(), fmt_eigen(M));`.

WDYT?

If we hate the extra `fmt_eigen` in the displayed message, we could have `Throw` strip it off automatically.



_Code quote:_
```objc
is_eigen_type
```
___
*[`common/drake_throw.h` line 43 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVvaRI0YmmPc_bEdFf:-NTVvaRI0YmmPc_bEdFg:b-bnl638) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L43)):*
> ```objc
>     // Characters get single quotes.
>     return fmt::format("'{}'", value);
>   } else {
> ```

See #14272 in yaml_write_archive.  We should special-case floats so that they are the same on all platforms.  Something like:

```c++
    if constexpr (std::is_floating_point_v<T>) {
      // Different versions of fmt disagree on whether to omit the trailing
      // ".0" when formatting integer-valued floating-point numbers.  Force
      // the ".0" in all cases by using the "#" option for floats.
      return fmt::format("{:#}", value));
```
___
*[`common/drake_throw.h` line 55 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVtzC85KBXs-N9J2rW:-NTVtzC85KBXs-N9J2rX:bkavra0) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L55)):*
> ```objc
>   static_assert(N % 2 == 0,
>                 "There should be an even number: up to 4 (name, value) pairs.");
>   std::stringstream ss;
> ```

At some point I'll need to dive into the object code to help guide us on how to craft the helper functions, but my gut feeling is that the optimal choice will be to avoid stringstream in the header and instead change `Throw` to take the values directly, from this local variable (supplanting`ss`):

```c++
std::array<std::pair<std::string_view, std::string>, 4> name_values;
```

The for-loop below would just populate `name_values` from `pairs`.  The `Throw` would only print non-empty pairs, and deal with all of the colon/period filler syntax.

The other design choice would be to use `fmt::basic_memory_buffer` instead of `stringstream` and keep all of the appending in the header.

I won't know which is better until we look at sample compiled code.

Either way, if the call site has _no_ extra values, then we shouldn't pay for any extra cost in `ThrowWithValues()`.  Any local variables need to be constexpr-guarded by `N > 0`.  Or possibly just have an 0-arg overload that gets used instead of this one, which just calls `Throw()` directly.
___
*[`common/drake_throw.h` line 92 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVqBo_31OkOWp7Pifl:-NTVqBo_31OkOWp7Pifm:bik6pwa) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L92)):*
> ```objc
> }  // namespace drake
> 
> /** Evaluates @p condition and iff the value is false will throw an exception
> ```

BTW I fully endorse switching away from `///`!

But if you're going to do that work, please switch to the canonical 0-space-hang instead of 1-space-hang so we don't have to switch it again later.
___
*[`geometry/shape_specification.h` line 120 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVqu8P4M_ZI0jlPqN8:-NTVqu8P4M_ZI0jlPqN9:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/geometry/shape_specification.h#L120)):*
> ```objc
>   /** Constructs a box with a vector of measures: width, depth, and height --
>    the box's dimensions along the canonical x-, y-, and z-axes, respectively.
>    @throws std::exception if any measures is not finite positive. */
> ```

typo



_Suggestion:_
```objc
measure
```
___
*[`geometry/shape_specification.h` line 125 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVqwot-orbz4zrJIXi:-NTVqwot-orbz4zrJIXj:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/geometry/shape_specification.h#L125)):*
> ```objc
>   /** Constructs a cube with the given `edge_size` for its width, depth, and
>    height.
>    @throw std::exception if edge_Size is not finite positive. */
> ```

typo



_Suggestion:_
```objc
edge_size
```
___
*[`geometry/test/shape_specification_test.cc` line 479 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVrbsd2PPse7Ch45wq:-NTVrbsd2PPse7Ch45wr:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/geometry/test/shape_specification_test.cc#L479)):*
> ```cpp
> // parameters. We test the error message to confirm that the erroneous value
> // is part of the message. We don't worry about the *rest* of the message,
> // relyibng DRAKE_THROW_UNLESS to do the right thing.
> ```

typo



_Suggestion:_
```cpp
relying
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-NTW0dHW06gx4wr2Eo8l:b-o367bz) (no related file):*

> We just can't call fmt from directly within the macro.

Technically, I'm calling a function (itself defined in the header) that calls `fmt` from within the macro (as many as four times). So, I consider that having already offended.

I look forward to seeing what ideas you might have to make it more compliant. Handling the value expression string, the value expression type, and the conversion of that value into a string left me painted into a corner. I couldn't figure out how to get formatting operations out of the header and into the .cc file.
</blockquote></details>

OK what's here in the latest draft seems good to me.  We'll probably need to tweak some compiler annotations on the `ThrowWithValues` but the gist is right.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTW_pov5FB831oO7TVe:-NTW_pov5FB831oO7TVf:bo0bit0) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L24)):*
> ```objc
> [[noreturn]] void Throw(const char* condition, const char* func,
>                         const char* file, int line,
>                         std::string_view suffix = std::string_view());
> ```

I did some experimenting in godbolt.  I think a good starting point for object code efficiency will be passing the bonus name-value pairs using a wrapper struct, instead of streaming in the header:

```c++
struct ThrowValuesBuf {
  std::array<std::pair<std::string_view, std::string>, 4> values;
};
```



_Suggestion:_
```objc
                        const ThrowValuesBuf& buf);
```
___
*[`common/drake_throw.h` line 44 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTWagtz1M3BhiVC67I1:-NTWagtz1M3BhiVC67I2:b-69stmf) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L44)):*
> ```objc
>     return fmt::format("'{}'", value);
>   } else {
>     return fmt::format("{}", value);
> ```

BTW simplify



_Suggestion:_
```objc
return fmt::to_string(value);
```
___
*[`common/drake_throw.h` line 49 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTWaqvo-p9FU9H1legl:-NTWaqvo-p9FU9H1legm:burphbj) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L49)):*
> ```objc
> 
> template <typename... ValuePairs>
> [[noreturn]] void ThrowWithValues(const char* condition, const char* func,
> ```

BTW I still need to test some real programs instead of godbolt, but my guess is we'll want to flag this as cold.

https://xania.org/201209/forcing-code-out-of-line-in-gcc



_Suggestion:_
```objc
[[noreturn]] __attribute__((noinline, cold)) void ThrowWithValues
```
___
*[`common/drake_throw.h` line 55 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTVtzC85KBXs-N9J2rW:-NTWaNni7-h8424PvHNd:b-vq7fnv) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L55)):*

At some point I'll need to dive into the object code to help guide us on how to craft the helper functions, but my gut feeling is that the optimal choice will be to avoid stringstream in the header and instead change `Throw` to take the values directly, from this local variable (supplanting`ss`):

```c++
std::array<std::pair<std::string_view, std::string>, 4> name_values;
```

The for-loop below would just populate `name_values` from `pairs`.  The `Throw` would only print non-empty pairs, and deal with all of the colon/period filler syntax.

The other design choice would be to use `fmt::basic_memory_buffer` instead of `stringstream` and keep all of the appending in the header.

I won't know which is better until we look at sample compiled code.

Either way, if the call site has _no_ extra values, then we shouldn't pay for any extra cost in `ThrowWithValues()`.  Any local variables need to be constexpr-guarded by `N > 0`.  Or possibly just have an 0-arg overload that gets used instead of this one, which just calls `Throw()` directly.
</blockquote></details>

Here's a better starting point IMO:

(1) Loop I through N/2 (i.e., iterating in pairs), not N.
(2) Convert the name to a string_view.
(3) Convert the value to a string.

Let the code in `*.cc` stitch all of the name-values together into the unified message.

```c++
  ThrowValuesBuf buf;
  // This is a "constexpr for" loop for 0 <= I < N/2.
  auto pairs_tuple = std::forward_as_tuple(std::forward<ValuePairs>(pairs)...);
  [&]<size_t... I>(std::integer_sequence<size_t, I...> &&) {
    ((
      (buf.values[I].first = std::get<2*I>(std::move(pairs_tuple)),
       buf.values[I].second = PairFormat(std::get<2*I+1>(std::move(pairs_tuple)))),
     ...));
  }(std::make_index_sequence<N / 2>{});
  Throw(condition, func, file, line, buf);
```
___
*[`common/drake_throw.h` line 78 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTWbYD11zbzBkKEodT9:-NTWbYD11zbzBkKEodTA:bd0knan) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L78)):*
> ```objc
> // _encode.
> #define _e_0(...)
> #define _e_1(value) #value, value
> ```

BTW Avoid template instantiation bloat by calling the function using `char*` not `char[2]` or `char[3]` etc.



_Suggestion:_
```objc
static_cast<const char*>(#value)
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 14 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 31 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTZ-m5B51-e2pKTglkE:-NTZ-m5B51-e2pKTglkF:bbfxtmb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L31)):*
> ```objc
>  receiving special treatment. */
> template <bool IsExpression, typename T>
> std::string PairFormat(const T& value) {
> ```

BTW As the final optimization for object code size, we can cherry-pick the most common types of T and opt them in to the cc file by adding lines like this to header:

```c++
template<> std::string PairFormat<double>(const double& value);
```

... with the definition in the cc file.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 14 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTW_pov5FB831oO7TVe:-NTZ1jPdAm5vavNcuFok:badetli) ([raw file](https://github.com/RobotLocomotion/drake/blob/4b8b3650b957e571e88a23d4f62bc8d0e9e6b3be/common/drake_throw.h#L24)):*

I did some experimenting in godbolt.  I think a good starting point for object code efficiency will be passing the bonus name-value pairs using a wrapper struct, instead of streaming in the header:

```c++
struct ThrowValuesBuf {
  std::array<std::pair<std::string_view, std::string>, 4> values;
};
```
</blockquote></details>

Tiny amendment: prefer `const char*` to `string_view` in the pair.  It's one less thing for the call site to fuss with.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19229)* status: 15 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19229#-NTdS65r6jmnuUO12hS7:-NTdS65r6jmnuUO12hS8:b-gn0qb3) (no related file):*
BTW I suspect that for symmetry, we should amend `DRAKE_DEMAND` and `DRAKE_ASSERT` likewise.  It's plausible that for our developers to want to print out more value-specific details even for internal crashes.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 15 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/19229#-NTQ32sf1XAkl32yuuKl:-Ncf98HB7jU3zBkg6-W5:btynr5d) (no related file):*

OK what's here in the latest draft seems good to me.  We'll probably need to tweak some compiler annotations on the `ThrowWithValues` but the gist is right.
</blockquote></details>

Sorry, I hadn't meant to trigger further conversation. It's certainly not done. I was cleaning things up before I left Tokyo and I figured I should rebase this so I could tackle it again when I got home. I'm pretty sure I haven't touched any of your original comments yet.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 16 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`common/drake_throw.h` line 64 at r4](https://reviewable.io/reviews/robotlocomotion/drake/19229#-Nic1wfW2nqrMqH60Pxl:-Nic1wfX2XgLgz0icozh:b-439ijy) ([raw file](https://github.com/robotlocomotion/drake/blob/4b30756470f98f1699a98ef544ee26147ad2563f/common/drake_throw.h#L64)):*
> ```objc
>       value_s = fmt::format("\"{}\"", value);
>     } else {
>       value_s = fmt::format("{}", value);
> ```

One more recent observation... with DRAKE_THROW_UNLESS often we do _not_ unit test that the thrown message happens correctly, we just assume that the macro works and does the obvious thing by code-inspection only.

Now that we're calling the type-specific formatter, we have the possibility that the formatter might _fail_.  Possibly we should update our styleguide to mandate that formatters can never throw?  Maybe that's obvious to everyone, but it might be worth writing down.

SeanCurtis-TRI (commented):
-(status: do not review) +(release notes: feature) 

+@jwnimmer-tri we're ready for proper review now.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229#-:-O8wKLTJ4jMx9w3BtnJ9:b-i36td1)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

SeanCurtis-TRI (commented):
A ping to move this back onto boards again.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229#-:-OC-R__wES6XW-i0zXgz:b-4yg6y)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

jwnimmer-tri (commented):
I think to make progress, we can probably land this without too many more changes, given suitable documentation and TODOs.  Details below.

> Reviewed 1 of 8 files at r5, 1 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229#-:-OC0CmPXEbV-mi0KC8p_:b-dcjefc)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake/19229#-NTdS65r6jmnuUO12hS7:-OC03l2b9xJmyHliduHi:b4nxkjd) (no related file):*

Both of these are good points. I propose we do them both in follow-up PRs. Once we like `DRAKE_THROW_UNLESS` we can propagate it. And I'd like to play file games independent of code changes (and since the code changes are already in flight, I'd play the file games after just to maximize the benefit to my inherent laziness.
</blockquote></details>

I'd be satisfied with a TODO comment.  (I didn't check if there is one yet.)
___
*[`common/drake_throw.h` line 64 at r4](https://reviewable.io/reviews/robotlocomotion/drake/19229#-Nic1wfW2nqrMqH60Pxl:-OC03vNb4scDeNx_n4Fi:bbucx33) ([raw file](https://github.com/robotlocomotion/drake/blob/4b30756470f98f1699a98ef544ee26147ad2563f/common/drake_throw.h#L64)):*
> ... we should update our styleguide to mandate that formatters can never throw ...

Per f2f yes this is probably a good idea.  However, it should appear a Doxygen in `drake/common/fmt.h` not the styleguide.  (The Doxygen is a lot easier to maintain.)

Need not be a blocker here, though.
___
*[`common/drake_throw.h` line 4 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-O9jixmT39DAMbW76uzw:-O9jixmT39DAMbW76uzx:bthdowk) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L4)):*
> ```objc
> 
> #include <array>
> #include <filesystem>
> ```

nit The `<filesystem>` is too big of a header to include here (in effect adding it to every single file in Drake).

Rather at any call sites which are trying to pass a `fs::path` to a `DRAKE_THROW_UNLESS`, we should ask them to call it like `DRAKE_THROW_UNLESS(my_path.something(), my_path.string());` -- to convert the path to a string before throwing.  Yes, then the "my_path.string() = ..." output will be polluted a little bit with the `.string()` detail, but that's a worthwhile tradeoff for reducing the `#include` fanout.
___
*[`common/drake_throw.h` line 26 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-O9jjqzW2LIhc6IwSknF:-O9jjqzW2LIhc6IwSknG:b-rdcs9n) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L26)):*
> ```objc
>  overloads for built-in C++ types and a templated catch-all option. */
> std::string StringifyParam(const char* value);
> std::string StringifyParam(char value);
> ```

I am somewhat dubious about using overloading here, as supposed to `is_same_v` inside of the template function body.  What if the user passes an `int64_t`, does the overload resolution choose something silly?

I imagine the main reason for these functions was the concern over object code size.  I'll discuss that in the other thread, just below.

Beyond that, the only _other_ only thing it seems we're trying to with this overload set is choose the optimal formatter specification for a given type?  I would rather just do that in the header file with `if constexpr` so that strings are printed with `fmt_debug_string` and leave the rest alone.
___
*[`common/drake_throw.h` line 32 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OC05vgR9k1bE0Ajg_gq:-OC05vgR9k1bE0Ajg_gr:b6zsxbx) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L32)):*
> ```objc
> std::string StringifyParam(const std::string& value);
> std::string StringifyParam(const std::string_view& value);
> template <typename T>
> ```

This inline template function is too much object code at call sites, in case DRAKE_THROW_UNLESS with extra argument is called from a header file.

I have ideas how to fix this, but maybe for now the easier solution is documentation -- document that `DRAKE_THROW_UNLESS(predicate);` is fair game in a header file, but `DRAKE_THROW_UNLESS(predicate, detail);` must only ever be used in cc files (for now).

I can circle back later with implementation improvements which make it safe enough to be inline anywhere, and then we could remove that limitation from the docs.

((My imagined solution is to handle user-defined types by calling either `x.to_string()` or ADL `to_string(x)` directly, without going through `fmt`, whichever the user has provided.  Then we can (by induction) know that the user's to_string is not too big to inline.  Then for POD types, we plumb through a bunch of helpers like you have here, but more thoroughly.))

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`common/drake_throw.h` line 64 at r4](https://reviewable.io/reviews/robotlocomotion/drake/19229#-Nic1wfW2nqrMqH60Pxl:-OC949-y5EHWzO0ATBFS:b6uibqn) ([raw file](https://github.com/robotlocomotion/drake/blob/4b30756470f98f1699a98ef544ee26147ad2563f/common/drake_throw.h#L64)):*

> ... we should update our styleguide to mandate that formatters can never throw ...

Per f2f yes this is probably a good idea.  However, it should appear a Doxygen in `drake/common/fmt.h` not the styleguide.  (The Doxygen is a lot easier to maintain.)

Need not be a blocker here, though.
</blockquote></details>

I've added a note.
___
*[`common/drake_throw.h` line 4 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-O9jixmT39DAMbW76uzw:-OC92Rpv0IsUGzspOl_b:b58xv03) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L4)):*

nit The `<filesystem>` is too big of a header to include here (in effect adding it to every single file in Drake).

Rather at any call sites which are trying to pass a `fs::path` to a `DRAKE_THROW_UNLESS`, we should ask them to call it like `DRAKE_THROW_UNLESS(my_path.something(), my_path.string());` -- to convert the path to a string before throwing.  Yes, then the "my_path.string() = ..." output will be polluted a little bit with the `.string()` detail, but that's a worthwhile tradeoff for reducing the `#include` fanout.
</blockquote></details>

Done

It's a shame the `std` namespace is so hairy. A forward declaration would've been a nice solution.

I went a different route.

From your comment below, I applied `fmt_debug_string` to the string-like types. However, when passing `my_path.string()` to `DRAKE_THROW_UNLESS`, not only would we get the undesirable `string()` in the output message (which I briefly toyed with stripping out a la `fmt_eigen`), but the path would be treated as a string (surrounding double quotes, escapes, etc.) That didn't seem the right thing to do.

So, I've added smarts to invoke `string()` or `to_string()` as they are available (and preferred over `fmt::format`). 

Thoughts?
___
*[`common/drake_throw.h` line 26 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-O9jjqzW2LIhc6IwSknF:-OC9Pb9K4VseZmzUpP-d:bhr68us) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L26)):*

I am somewhat dubious about using overloading here, as supposed to `is_same_v` inside of the template function body.  What if the user passes an `int64_t`, does the overload resolution choose something silly?

I imagine the main reason for these functions was the concern over object code size.  I'll discuss that in the other thread, just below.

Beyond that, the only _other_ only thing it seems we're trying to with this overload set is choose the optimal formatter specification for a given type?  I would rather just do that in the header file with `if constexpr` so that strings are printed with `fmt_debug_string` and leave the rest alone.
</blockquote></details>

Well, object code size is one thing, but then I felt the overload resolution would be faster than the `if constexpr` resolution and recompiling across the code base.

Still, I've backed it out to the current version.
___
*[`common/drake_throw.h` line 32 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OC05vgR9k1bE0Ajg_gq:-OC96IcgEDfTp0TGHhLb:b-qt2jfu) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L32)):*

This inline template function is too much object code at call sites, in case DRAKE_THROW_UNLESS with extra argument is called from a header file.

I have ideas how to fix this, but maybe for now the easier solution is documentation -- document that `DRAKE_THROW_UNLESS(predicate);` is fair game in a header file, but `DRAKE_THROW_UNLESS(predicate, detail);` must only ever be used in cc files (for now).

I can circle back later with implementation improvements which make it safe enough to be inline anywhere, and then we could remove that limitation from the docs.

((My imagined solution is to handle user-defined types by calling either `x.to_string()` or ADL `to_string(x)` directly, without going through `fmt`, whichever the user has provided.  Then we can (by induction) know that the user's to_string is not too big to inline.  Then for POD types, we plumb through a bunch of helpers like you have here, but more thoroughly.))
</blockquote></details>

Documentation updated.

I also have a whiff of the `x.to_string() as noted before.

jwnimmer-tri (commented):
> Reviewed 1 of 9 files at r7, 2 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`-- commits` line 30 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK5C7l-hVgDNiIEFpW:-OCK5C7l-hVgDNiIEFpX:b-nhq0p3):*
Please carve the inertia_visualizer fix into its own ahead-of-time pull request.  It is too much clutter to consider here.
___
*[`common/drake_throw.h` line 4 at r7](https://reviewable.io/reviews/robotlocomotion/drake/19229#-O9jixmT39DAMbW76uzw:-OCK6SOZ0MnQramBiinx:bdv8bz9) ([raw file](https://github.com/robotlocomotion/drake/blob/5d60de393dab9afb0209edd7a912c459728f6d27/common/drake_throw.h#L4)):*
> From your comment below, I applied `fmt_debug_string` to the string-like types. However, when passing `my_path.string()` to `DRAKE_THROW_UNLESS`, not only would we get the undesirable `string()` in the output message (which I briefly toyed with stripping out a la `fmt_eigen`), but the path would be treated as a string (surrounding double quotes, escapes, etc.) That didn't seem the right thing to do.

My suggestion of using quoting for paths was *only* because that's what (I thought) the existing implementation did.

> So, I've added smarts to invoke `string()` or `to_string()` as they are available (and preferred over `fmt::format`).
>
> Thoughts?

I think designing behavior in the absence of call-site evidence means we will be guessing and therefore probably get it wrong.  (The unit test does not count as valid example call-site code.)

How about we incrementally build out this function and its capabilities as we find acute need for enhancements?  The current call site only uses floating point numbers, so let's only implement that for now, with a TODO to support more argument types later on.

Then in a separate PR, where we port a half dozen or more `fs::path` checks to this new idiom, we can debate about how this sugar should work.
___
*[`common/fmt.h` line 144 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK2Ob96jzzfmC09YmD:-OCK2Ob96jzzfmC09YmE:bvl35ee) ([raw file](https://github.com/robotlocomotion/drake/blob/3603f84b244d2f88fc0e8adaad3791f90f359301/common/fmt.h#L144)):*
> ```objc
> @param EXPR An expression to `return` from the format_as function; it can
> refer to the given `ARG` name which will be of type `const TYPE& ARG`. The
> evaluated expression _must never throw_.
> ```

nit This is a smidge too strong: it's not realistic to write `to_string` implementations without memory allocation.  We should allow the EXPR to throw `std::bad_alloc` (only).

Specifically we should allow throwing anything that `std::string`'s default allocator might throw, which is `std::allocator<char>::allocate` which is `std::bad_alloc` (only).
___
*[`geometry/shape_specification.h` line 161 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK3XsO6xZAS7uHiJ9O:-OCK3XsO6xZAS7uHiJ9P:bw1b7fw) ([raw file](https://github.com/robotlocomotion/drake/blob/3603f84b244d2f88fc0e8adaad3791f90f359301/geometry/shape_specification.h#L161)):*
> ```objc
>    specify the box's dimension along the canonical x-, y-, and z-axes,
>    respectively.
>    @throws std::exception if any measure is not finite positive. */
> ```

BTW These doc improvements are plausible to scope into this pull request, but in service of keeping the more important changes displayed even more crisply, I would encourage you to land this file as a separate PR.
___
*[`common/drake_throw.h` line 67 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK6uwF3ub2Slv2FlGg:-OCK6uwF3ub2Slv2FlGh:b-k33gj2) ([raw file](https://github.com/robotlocomotion/drake/blob/3603f84b244d2f88fc0e8adaad3791f90f359301/common/drake_throw.h#L67)):*
> ```objc
>     return value.to_string();
>   } else {
>     return fmt::to_string(value);
> ```

BTW My mental model of keeping this inline-able is that we do *not* have this catch-all fallback at the end.  Do we have any concrete examples of needing to call `DRAKE_THROW_UNLESS` on something which didn't match any of the preceding conditions?

Of course we have a documentation disclaimer for now that users cannot use ARGS when invoked in header files, but that is something that only code reviews would catch, so is not going to be perfect.  If we could aim to have this function be safe to inline anywhere down the road, that would let us remove that caveat and life gets simpler.  To do that, we probably cannot have a catch-all, so I'm curious what the motivating use case for it would be?
___
*[`visualization/inertia_visualizer.cc` line 180 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK5W8kEf-pAORtGUFg:-OCK5W8kEf-pAORtGUFh:b-mhakdr) ([raw file](https://github.com/robotlocomotion/drake/blob/3603f84b244d2f88fc0e8adaad3791f90f359301/visualization/inertia_visualizer.cc#L180)):*
> ```cpp
>     // The body is a point mass. We'll simply introduce a small sphere with
>     // a one-centimeter radius.
>     abc << 0.01, 0.01, 0.01;
> ```

This is not a good fix for a point mass.  The invariant of the current visualization mode is that the ellipsoid's volume communicates the mass.

Thus, the missing code to be added looks something like this:

```c++
std:tie(...) = ...;
if (radii.isZero(0.0)) {
  // The body is a point mass. Treat it like a sphere.
  radii = Vector3d::Ones();
}
// An ellipse that has one ...
```
___
*[`geometry/shape_specification.cc` line 176 at r8](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OCK4DrR3EpfXm4S7lue:-OCK4DrR3EpfXm4S7luf:b-bt7l3u) ([raw file](https://github.com/robotlocomotion/drake/blob/3603f84b244d2f88fc0e8adaad3791f90f359301/geometry/shape_specification.cc#L176)):*
> ```cpp
>   // fence over which ridiculous vectors will trip.
>   if (norm < 1e-10) {
>     throw std::logic_error("Can't make pose from a zero vector.");
> ```

BTW This logic could be `DRAKE_THROW_UNLESS` as well?

SeanCurtis-TRI:
Behold! It's aliiiiiiiive!

SeanCurtis-TRI:
-(status: defer ci)

jwnimmer-tri:
Thanks for the update.  I will look at this, but it will be a deep dive and my brain isn't up for that this week.  Stay tuned next week.

SeanCurtis-TRI (commented):
No worries. It's been hanging around for a couple of years. A few more weeks won't harm it. And, in the meantime, I can look at how I made CI angry.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229#-:-Oa1XKZP2FVBMnr-JL3S:bnp2wsq)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

SeanCurtis-TRI:
As for CI, once #23464 merges, this CI should be happy.

SeanCurtis-TRI:

Before I dig into the implementation too much more, we have a decision to make.  The main challenge is making the new feature suitable for inline functions (i.e., to be used in header file cheap getters / setters).

There's two ways we could go:

(1) Forbid the use of the new feature in header files, with enforcement in CI.  Probably by adding linter to `drakelint.py`, either by having it find and reject `DRAKE_THROW_UNLESS(condition, detail1, ...etc...);` while allowing `DRAKE_THROW_UNLESS(condition);`, or by giving the new feature a new name like `DRAKE_THROW_UNLESS_DETAIL(condition, detail1, ...etc...);` so the linter can just look for the disallowed screamword in headers.

(2) Continue fine-tuning the implementation to be suitable for inline functions, so it can be used in headers.

The advantage of (1) is that is a _vastly_ less work to get landed.  We wouldn't need to worry about which types can be cheaply formatted, for example.  But I'm not sure how much you think we want/need to use this function in headers.
</blockquote></details>

I vote in favor of (1)

- It gets the feature out there sooner.
- We can always solve the problem of being header friendly in the future and, at that time, remove the admonition against being in header files. Much harder to go the other way.
- Before fully embracing the alternate name, I'd like to take a stab at the linter being smarter. My motivation would be in optimistic anticipation of it eventually becoming header-safe.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`common/drake_throw.h` line 28 at r11](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OcAco5b6PNXYyHg-u32:-OcAvCPU33-jBtntUVUA:busuc5h) ([raw file](https://github.com/robotlocomotion/drake/blob/2bcdbcfc7deda4a1c34c7cb74030cf6a82583298/common/drake_throw.h#L28)):*

As it stands, this function can't be defined inline in the header file (it doesn't meet our guidelines).  Since the template argument is a closed set, it could be defined (and then explicitly instantiated) in the cc file, and _needs_ to be that way if any of the code in its body is non-trivial.

I suppose technically, we _could_ keep this function in the header if all it did was the `if constexpr` if-else chain.  Then in the body of each `if`, it would need to call a non-inlined helper function.  (Currently, `fmt_floating_point is inline, but we could move its definition to cc with a little work.)
</blockquote></details>

See #23638.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 12 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`common/drake_throw.h` line 20 at r11](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OcAzbN84XsOkjkWInal:-OcGSOAoCSJ0yUaldrsn:bmvvv3d) ([raw file](https://github.com/robotlocomotion/drake/blob/2bcdbcfc7deda4a1c34c7cb74030cf6a82583298/common/drake_throw.h#L20)):*

We should really do this first.  See #23639.
</blockquote></details>

(Ready for rebasing now.)
___
*[`common/drake_throw.h` line 28 at r11](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OcAco5b6PNXYyHg-u32:-OcGSSJz26WmKArrH8BR:b-nhaas4) ([raw file](https://github.com/robotlocomotion/drake/blob/2bcdbcfc7deda4a1c34c7cb74030cf6a82583298/common/drake_throw.h#L28)):*

See #23638.
</blockquote></details>

(Ready for rebase now.)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229)* status: 14 unresolved discussions, needs at least two assigned reviewers

___
*[`common/drake_throw.h` line 66 at r11](https://reviewable.io/reviews/robotlocomotion/drake/19229#-OcArT4K6kP3IEKqufEd:-OcbqGeq1Qt7u0lmZth1:b-yz4c5m) ([raw file](https://github.com/robotlocomotion/drake/blob/2bcdbcfc7deda4a1c34c7cb74030cf6a82583298/common/drake_throw.h#L66)):*

Yeah.  For now I'm happy to let this ride.  Once we start having uses of this function in header files, I can check the specifics of each use and see if anything new cropped up.
</blockquote></details>

(The Gemini summary of "object code compiler what is a cold clone section" is pretty good, if you want a reference.)

jwnimmer-tri (commented):
Ready for platform.

> @jwnimmer-tri reviewed 4 of 4 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19229#-:-OclU3Ad8rec2KEwtxqr:boxefeq)* status: needs at least two assigned reviewers

SeanCurtis-TRI:
Happy mac!

PR #19204, opened 2023-04-15
[plant] Add MultibodyPlant::GetDefaultPositions()

EricCousineau-TRI:
Resolves #19203 (at least one way)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19204)

PR #19230, opened 2023-04-19
[multibody] Set random state to default when distribution is missing

### LINKED ISSUE #19106: MultibodyPlant random contexts fall back to zero positions, not defaults
### What happened?

See [StackOverview](https://stackoverflow.com/questions/75881315/pydrake-montecarlosimulation-pose-of-freebody-being-reset) for the original discovery.

Right now, if I don't set up any random distributions on a MultibodyPlant and then I call `SetRandomContext`, the state I get back is all zeros (except for the '1' in any quaternions).  This happens even if I've previously called `SetDefaultFreeBodyPose` on a free body.

The thing I assumed I would get would be the default joint postures as-if from CreateDefaultContext, for state where I haven't configured any random distribution yet.

### Version

v1.14.0

Other

Unknown / Other

### Relevant log output

_No response_

xuchenhan-tri:
Closes #19106.

Currently, default positions for floating bodies are stored separately from the default positions of joint dofs. This is violates single source of truth and is the root cause of #19106.

This PR makes sure that for floating bodies the default pose value requested at MbP::SetDefaultFreeBodyPose() is stored in the floating joint connecting world and the body and MbP::GetDefaultFreeBodyPose() retrieves the default value from the floating joint if it exists. This fixes #19106 as a side effect since now there's single source of truth for default positions for all dofs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)


xuchenhan-tri (commented):
-(status: do not review) +@sherm1 for feature review please.

I wonder if the added throw condition in MbP::Set/GetDefaultFreeBodyPose constitutes a breaking change. PTAL. FWIW, the accompanied anzu ci is in flight.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTR6zIu0gxBpg2wwEtH:bi20lls)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers (waiting on @sherm1)

sherm1 (commented):
Nice! Feature :lgtm: with a few minor comments, ptal.

> Reviewed 2 of 6 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTV5Swk-I-s6B0PSNW6:bo1wqpg)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

___
*[`multibody/plant/test/multibody_plant_test.cc` line 1146 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTV1QRK5eaOWhP7Jvw8:-NTV1QRLAPoXV1QkIzKs:b-pn2xr4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/test/multibody_plant_test.cc#L1146)):*
> ```cpp
>   EXPECT_TRUE(
>       CompareMatrices(plant.GetDefaultFreeBodyPose(free_body).GetAsMatrix4(),
>                       new_X_WB.GetAsMatrix4(), kTolerance));
> ```

BTW is there a logical reason a tolerance is needed here but not above? I can see why it's not exact here since the rotation matrix gets put into a quaternion and then reconstructed. I guess the current implementation stores X_WB exactly pre-finalize? Consider allowing a tolerance above also so that the test would still pass if we decided to go with a quaternion there also, and so that the next reader won't have to worry about the significance of the two tests being different.

Also, slightly weird that the floating body defaults will change a little bit pre- and post-finalize. Would be better to switch the pre-finalize defaults to (quaternion, vec3) or Vector7 here so that they don't change when transferred to the floating joint.
___
*[`multibody/tree/multibody_tree.cc` line 971 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTV4_T4CJgApAMW436F:-NTV4_T54O92R1EeBsiC:butsknv) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/tree/multibody_tree.cc#L971)):*
> ```cpp
>   } else {
>     // Use body representation pre-finalize.
>     default_body_poses_[body.index()] = X_WB;
> ```

BTW consider Vector7 as internal rep here to avoid pre/post weirdness

jwnimmer-tri:
Without quite thinking through the deprecation transition yet, what about this proposal:

Change `SetDefaultFreeBodyPose` to **add a floating joint** for the body (idempotently, if called twice) and then set the defaults on the joint.  Maybe return the `const Joint<T>&` for convenience.

sherm1 (commented):
Oh, I like that idea! Then the defaults are always stored with the joint, and the user is guaranteed to get a free body where they expected one. We'd still have to barf if this was called post-finalize on a non-free body since we can't add a joint then.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTVAwRO-uZ6uYjoeYim:bta10b7)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

jwnimmer-tri:
Okay, deprecation proposal:

I don't think we should deprecate the current names.  They are too widely used.  We could try this:

SetDefaultFreeBodyPose:
* If the body already has a floating joint, set its pose.  Same pre- or post-finalize.
* Otherwise, if the body has any joint (in either direction), console warn (once?) and do nothing.  This is deprecated behavior (for the release notes).  After 3 months, this throws instead of warning.  Same pre- or post-finalize.
* Otherwise, add a floating joint (and set its pose).  This can only happen pre-finalize.  Post-finalize, there will always have already been some joint.

GetDefaultFreeBodyPose:
* If there is floating joint, then return its pose.  Same pre- or post-finalize.
* If there is either no joint or a non-floating joint (in either direction), then return Identity.  Warn (once?).  This is deprecated behavior (for the release notes).  After 3 months, this throws instead of warning.  Same pre- or post-finalize.

I think that's the closest we can come to a graceful transition?

sherm1 (commented):
:+1: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTV_ge7AlMcv54e7TMm:b-8fx2u7)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @xuchenhan-tri)

xuchenhan-tri:
Per f2f, the deprecation proposal is still breaking in certain scenario.

If `SetDefaultFreeBodyPose` is called a body without any joint, and later a non-floating joint is added to the same body, an exception will be thrown during finalize.

The consensus is that breaking in that case is fine because the user is doing a non-sensible thing to begin with.

xuchenhan-tri:
We could also be nice and log a warning instead of throwing and switch to throwing after the grace-period.

xuchenhan-tri (commented):
> If SetDefaultFreeBodyPose is called a body without any joint, and later a non-floating joint is added to the same body, an exception will be thrown during finalize.

It seems like this case may actually be quite common. See for example in our parser code:https://github.com/RobotLocomotion/drake/blob/a3459238a422d27a3b4ef1bde9d71ffcf00e0f2b/multibody/parsing/detail_sdf_parser.cc#L941.

That makes me think the semantic of "set the pose if the body is indeed free but discard the pose if it's not" is ingrained and if we throw an exception for this use case it would be breaking many workflows.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTWanfr8ZD-Og8rejzQ:btcaecs)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @sherm1)

___
*[`multibody/plant/test/multibody_plant_test.cc` line 1146 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTV1QRK5eaOWhP7Jvw8:-NTV6s886rQAsKqEa36x:bombmv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/test/multibody_plant_test.cc#L1146)):*

BTW is there a logical reason a tolerance is needed here but not above? I can see why it's not exact here since the rotation matrix gets put into a quaternion and then reconstructed. I guess the current implementation stores X_WB exactly pre-finalize? Consider allowing a tolerance above also so that the test would still pass if we decided to go with a quaternion there also, and so that the next reader won't have to worry about the significance of the two tests being different.

Also, slightly weird that the floating body defaults will change a little bit pre- and post-finalize. Would be better to switch the pre-finalize defaults to (quaternion, vec3) or Vector7 here so that they don't change when transferred to the floating joint.
</blockquote></details>

Good call. Done.
___
*[`multibody/tree/multibody_tree.cc` line 971 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTV4_T4CJgApAMW436F:-NTV7-oFDkG2-rMMvY0p:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/tree/multibody_tree.cc#L971)):*

BTW consider Vector7 as internal rep here to avoid pre/post weirdness
</blockquote></details>

Done

jwnimmer-tri:
> It seems like this case may actually be quite common. See for example in our parser code:

So far, both examples you've found (one in Drake, one in Anzu) were code that @EricCousineau-TRI wrote.  It's probably a vote for being more careful with the deprecation that I initially offered, but I'm not sure it's a vote for "Silently let users call functions that go nowhere, do nothing".

Possibly the "take a bigger swing" I mentioned on the call will be the right answer.  (Deprecate the copy-pasted wall of free body functions on the MbP itself, leaving only a little remaining sugar for obtaining the Joint associated with a free body.  Then the user only ever interacts with the free body's joint)

xuchenhan-tri (commented):
We can't blame anyone who use the API that way because that is how [the documentation describes it].(https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_multibody_plant.html#a054aef89f09b26103692515fea766e9e)

I support "the bigger swing" in principle but it's a bit more than I can chew at the moment. I'd imagine that involves non-trivial effort in cleaning up the deprecation mess because as you said earlier in the thread "they are too widely used."

I propose we proceed this PR as is to fix the issue referenced and also unblock #19204. I'll write up a separate issue capturing our discussion and your proposal and I can come back to this when I have a bigger chunk of time. Does that sound reasonable?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTWg7Jg55CqUl62HP9Y:b1x3hto)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @sherm1)

jwnimmer-tri (commented):
I have no problem keeping this small, non-breaking, non-deprecating.

In that case, I think we need to trim the PR to have no behavioral changes at all, other than the random vs default bugfix.  I'll post a few notes along those lines below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTYyX5WA07BCfbE8KnW:b4nmopq)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @sherm1 and @xuchenhan-tri)

___
*[`multibody/plant/multibody_plant.h` line 2644 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTYyjFWCEaHiij_ZXge:-NTYyjFX2tkq8a9QsEm2:b-ruqv58) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/multibody_plant.h#L2644)):*
> ```objc
>   ///   Default pose of the body.
>   /// @throws std::exception if called post-finalize and body.is_floating() is
>   /// false.
> ```

If we're keeping this small (non-breaking), we cannot add new throwing conditions.
___
*[`multibody/plant/multibody_plant.h` line 2654 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTYypirA3JhkwGzgiAO:-NTYypirA3JhkwGzgiAP:bp3bym4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/multibody_plant.h#L2654)):*
If we're keeping this small (non-breaking), we 100% need to echo back the call to SetDefaultFreeBodyPose, even if the body wasn't free, even across a Finalize.

That means we need to keep the dense `X_WB` list of poses still, so we can echo back any configured non-identity poses for non-free-bodies.

Possibly that means switching to this as a representation type:

```
std::vector<std::variant<
  math::RigidTransform<double>,
  JointIndex
>> X_WB_default_list_;
```

When we add a joints for a free body, we copy the default transform into the joint and replace it with the associated JointIndex in this list.

Note that this needs to happen for _user-added floating joints vs the world_ as well, not just the automatically-added joints.  The invariant is that any floating body (a body connected to the world with a floating joint) has a JointIndex in this map instead of a transform.

EricCousineau-TRI:
> Possibly the "take a bigger swing" I mentioned on the call will be the right answer.

If we want to avoid any transition / deprecation, but remove multiple-sources-of-truth, we use "has floating joint?" vs. "is finalized?", i.e.,

- If body has a floating joint, that is used for default free body pose.
- If a body does not have a floating joint, then it uses the pose list. If that value never gets used, that's OK in my opinion.

More correct, and no narrowing of functionality.
No deprecation needed,. Yes, it allows a user to specify poses that may be garbage, but we narrow functionality at that point.

WDYT?

NOTE: Ah, Jeremy posted effectively equivalent via `vector<variant<RigidTransform, JointIndex>>` - I like it!

> So far, both examples you've found (one in Drake, one in Anzu) were code that @EricCousineau-TRI wrote.

Ah, yup, that was for parsing. Forgot about that!

xuchenhan-tri (commented):
I made a few changes to limit the scope of the PR to be non-breaking following @jwnimmer-tri 's suggestion. PTAL, thanks! Since the semantics of `Set/GetDefaultFreeBodyPose` have been reset to what they currently mean in master today, it might be easier to just do a full review (base vs. r4).

+@jwnimmer-tri for the second reviewer if you are up for it since you've already spent some time looking at this PR. Otherwise, feel free to kick it to the on-call platform after @sherm1 takes another pass.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTpIP5P-Pjtw-8QeqOC:b1e64l3)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @sherm1)

___
*[`multibody/plant/multibody_plant.h` line 2644 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTYyjFWCEaHiij_ZXge:-NTo6Z8e7H7UufZpmScl:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/multibody_plant.h#L2644)):*

If we're keeping this small (non-breaking), we cannot add new throwing conditions.
</blockquote></details>

Done
___
*[`multibody/plant/multibody_plant.h` line 2654 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTYypirA3JhkwGzgiAO:-NTp0uFj9nXwIH2QaWls:b-sx81dc) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/multibody_plant.h#L2654)):*

If we're keeping this small (non-breaking), we 100% need to echo back the call to SetDefaultFreeBodyPose, even if the body wasn't free, even across a Finalize.

That means we need to keep the dense `X_WB` list of poses still, so we can echo back any configured non-identity poses for non-free-bodies.

Possibly that means switching to this as a representation type:

```
std::vector<std::variant<
  math::RigidTransform<double>,
  JointIndex
>> X_WB_default_list_;
```

When we add a joints for a free body, we copy the default transform into the joint and replace it with the associated JointIndex in this list.

Note that this needs to happen for _user-added floating joints vs the world_ as well, not just the automatically-added joints.  The invariant is that any floating body (a body connected to the world with a floating joint) has a JointIndex in this map instead of a transform.
</blockquote></details>

Done (I believe), PTAL.
___
*[`multibody/plant/test/multibody_plant_test.cc` line 1146 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTV1QRK5eaOWhP7Jvw8:-NTo6c-g3019PHg9c7-n:b3i5866) ([raw file](https://github.com/RobotLocomotion/drake/blob/5ee3769f422a80e8b3f71953d67d64015b863cd5/multibody/plant/test/multibody_plant_test.cc#L1146)):*

Good call. Done.
</blockquote></details>

Done (for real this time. Didn't push for the last time around.)

jwnimmer-tri (commented):
> Reviewed 2 of 6 files at r1, 1 of 4 files at r3, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @sherm1 and @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.cc` line 799 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLK_K7G5FULD0PQST:-NTpLK_K7G5FULD0PQSU:b-7ubhx1) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L799)):*
> ```cpp
>       // Set default positions for the floating joints.
>       // TODO(xuchenhan-tri): This assumes that the only type of floating
>       // joint is the quaternion floating joint. This may change pending
> ```

nit Seems to be missing the rest of the sentence.



_Code quote:_
```cpp
This may change pending
```
___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpLr8fA-dmG7Qdunr6:bs1iz0u) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*
> ```cpp
>       const RigidTransform<double>& default_pose = GetDefaultFreeBodyPose(body);
>       const Vector<double, 7> default_q = ToGeneralizedPositions(default_pose);
>       joint->set_default_positions(default_q);
> ```

We have a function `QuaternionFloatingJoint::SetDefaultPose` which takes a RigidTransform directly.  We've already asserted on the joint subclass just above, so we can `safely dynamic_cast` the `joint` to a `QuaternionFloatingJoint&` to call this function.

It seems to me like using that direct function would be a lot less brittle than the `ToGeneralizedPositions` helper function that duplicates the logic for quaternion position ordering vs the joint subclass.

Relatedly, it's not clear to me why the variant stores a Vector7 instead of a RigidTransformd.

sherm1 (commented):
I like this narrower version! Feature :lgtm:

> Reviewed 2 of 4 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTpRHAU3saSGAHcGlCM:bnwqwq2)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpQ9Ls9QNoaef2V8u8:brv9sxt) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*
> why the variant stores a Vector7 instead of a RigidTransformd

That way the actual default pose doesn't change (by roundoff) when the pose is transferred to the joint later. Joints can only hold default poses via their q's. Seems better to avoid that weirdness.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @sherm1 and @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpRkMP3qJrsYmSq7kd:bq968m6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*

> why the variant stores a Vector7 instead of a RigidTransformd

That way the actual default pose doesn't change (by roundoff) when the pose is transferred to the joint later. Joints can only hold default poses via their q's. Seems better to avoid that weirdness.

</blockquote></details>

That's nice in theory, but the current code doesn't do that.  It's always laundering through RigidTransform (sometimes twice).

I am OK with storing 7 doubles instead of RigidTransform, but in that case `pair<Quaternion, Vector3>` is the correct data structure, and we'll need to use the Joint-specific functions to set translation and quaternion separately.  The crux of the defect is that duplicating the logic for QuaterjoinFloatingPoint position layout here in the tree class is too brittle.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpYmSEAt30WQl_1eZr:bjn92uu) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*

That's nice in theory, but the current code doesn't do that.  It's always laundering through RigidTransform (sometimes twice).

I am OK with storing 7 doubles instead of RigidTransform, but in that case `pair<Quaternion, Vector3>` is the correct data structure, and we'll need to use the Joint-specific functions to set translation and quaternion separately.  The crux of the defect is that duplicating the logic for QuaterjoinFloatingPoint position layout here in the tree class is too brittle.
</blockquote></details>

When we want to know the pose of a body that has _any_ joint we have to "launder" its coordinates q into a pose X(q). My point is just that we should store the actual pose the same way pre- and post-finalize so the apparent pose stays bit identical pre and post.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @sherm1 and @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpZpHfA4xoMk0pTSuO:bj9kuo2) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*

When we want to know the pose of a body that has _any_ joint we have to "launder" its coordinates q into a pose X(q). My point is just that we should store the actual pose the same way pre- and post-finalize so the apparent pose stays bit identical pre and post.
</blockquote></details>

That's fine, but the current code doesn't do that.  The current code stores 7q, then when it needs use that data converts the 7q to X then back to 7q before feeding it into the joint.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @sherm1)

___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpaoXlBSHciYaQDc9p:b-m3cboe) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*

That's fine, but the current code doesn't do that.  The current code stores 7q, then when it needs use that data converts the 7q to X then back to 7q before feeding it into the joint.
</blockquote></details>

OK Looks much better now.

It might be worth a chesterton's fence comment somewhere about _why_ we store the default using the pair thing ("to match the joint without any numerical conversions").

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 1 of 5 files at r2, 1 of 2 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTpcvKY2ypgWJJKNCuB:b-vtj4vo)* status: 2 unresolved discussions (waiting on @sherm1 and @xuchenhan-tri)

___
*[`bindings/pydrake/multibody/test/plant_test.py` line 1273 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpcLnu4rHiTA6GCbX6:-NTpcLnv4LmU74QXwa3W:bmpxxrj) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/bindings/pydrake/multibody/test/plant_test.py#L1273)):*
> ```Python
>         numpy_compare.assert_float_equal(
>             plant.GetDefaultFreeBodyPose(body=body).GetAsMatrix4(),
>             X_WB_default.GetAsMatrix4())
> ```

nit I would not expect any impact on the bindings for this PR. Now that there are no more behavior changes, do we still need to carve this out into a separate test case? It _seems_ like the test code is identical, but it's a little tough to tell for sure.

sherm1 (commented):
> Reviewed 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions (waiting on @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.h` line 3091 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpbU-nASEnkqyaYgeH:-NTpbU-nASEnkqyaYgeI:bdv38o8) ([raw file](https://github.com/RobotLocomotion/drake/blob/6360a3e7580f475bf0b2cdcb551a154023e21ad3/multibody/tree/multibody_tree.h#L3091)):*
> ```objc
>   // and the body during Finalize(), and the default poses can be retrieved via
>   // the joints' default positions. These associated joints are guaranteed to be
>   // QuaternionFloatingJoints.
> ```

minor: This is unlikely to be true in the near future. IIRC we've already had a request from Russ to infer RpyFloatingJoints optionally to avoid the quaternion length constraint which makes some kinds of analysis difficult. Less imminent, if a free body is known to be a point mass we ought to infer a translation-only (3 dof) joint for it.

I would suggest instead something like "When a quaternion floating joint is used for a floating body (very common), the pose will be bit identical when transferred from the default free body pose."

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 2 unresolved discussions (waiting on @jwnimmer-tri and @sherm1)

___
*[`multibody/tree/multibody_tree.h` line 3091 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpbU-nASEnkqyaYgeH:-NTpeI7s92vO_gBkojr5:b7ri8xx) ([raw file](https://github.com/RobotLocomotion/drake/blob/6360a3e7580f475bf0b2cdcb551a154023e21ad3/multibody/tree/multibody_tree.h#L3091)):*

minor: This is unlikely to be true in the near future. IIRC we've already had a request from Russ to infer RpyFloatingJoints optionally to avoid the quaternion length constraint which makes some kinds of analysis difficult. Less imminent, if a free body is known to be a point mass we ought to infer a translation-only (3 dof) joint for it.

I would suggest instead something like "When a quaternion floating joint is used for a floating body (very common), the pose will be bit identical when transferred from the default free body pose."
</blockquote></details>

In this revision, we are making the assumption that these joints are quaternion floating joints in the code (by dynamic casting) regardless how we phrase the comment here. When another type of floating joints is introduced, we would need to rework this logic and possibly loosen the bit-wise identical guarantee and just store a generic RigidTransform instead.
___
*[`multibody/tree/multibody_tree.cc` line 799 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLK_K7G5FULD0PQST:-NTpQnlgEHVmgOmptuSu:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L799)):*

nit Seems to be missing the rest of the sentence.
</blockquote></details>

Done.
___
*[`multibody/tree/multibody_tree.cc` line 803 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpLr8fA-dmG7Qdunr5:-NTpf60N7OZUsP_dwxM6:b1ji3dl) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f0f302c26b2073f3ff0e8ec123721ede072020f/multibody/tree/multibody_tree.cc#L803)):*

OK Looks much better now.

It might be worth a chesterton's fence comment somewhere about _why_ we store the default using the pair thing ("to match the joint without any numerical conversions").
</blockquote></details>

Done, added note near declaration of the member variable.
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 1273 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpcLnu4rHiTA6GCbX6:-NTpdfX64COnKH6gVyCd:bgtczeq) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/bindings/pydrake/multibody/test/plant_test.py#L1273)):*

nit I would not expect any impact on the bindings for this PR. Now that there are no more behavior changes, do we still need to carve this out into a separate test case? It _seems_ like the test code is identical, but it's a little tough to tell for sure.
</blockquote></details>

Definitely not necessary, but the original test is somewhat nonsensical even though it's allowed (setting free body pose of a non-free body).
I can revert this if you think this should belong to the future, bigger swing PR.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 1 unresolved discussion (waiting on @sherm1)

___
*[`bindings/pydrake/multibody/test/plant_test.py` line 1273 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpcLnu4rHiTA6GCbX6:-NTpgImE-VWSA2tEn0uM:bofjmeh) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f9707a408cdaa308fddc73198dde6bbc7ffdc1/bindings/pydrake/multibody/test/plant_test.py#L1273)):*

Definitely not necessary, but the original test is somewhat nonsensical even though it's allowed (setting free body pose of a non-free body).
I can revert this if you think this should belong to the future, bigger swing PR.
</blockquote></details>

Ah okay, I think this is good to keep.

(I hadn't identified that Link1 was not floating when trying to understand the changes.)

jwnimmer-tri:
For the record, there one more rough edge here still (though I don't think it needs to happen in this PR, since the focus here is Random vs Default): if the user calls `AddJoint<QuaternionFloatingJoin>(..., world, ...)` we should switch that body's default pose variant from pair to JointIndex and copy the default into the Joint.  As it stands, any changes to defaults the user makes to that new joint will be _overwritten_ by the X_WB map.

xuchenhan-tri (commented):
It can't happen in this PR because that would be breaking. The current documentation of `MbP::SetDefaultFreeBodyPose()` dictates that X_WB map always wins.

When we do get to that breaking change eventually, the dual body/joint representation of the default pose shouldn't exist any more so the point is moot.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-:-NTphnvK6nofQ6T9P74k:b-vvpp01)* status: 1 unresolved discussion (waiting on @sherm1)

jwnimmer-tri:
SGTM.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 1 unresolved discussion (waiting on @xuchenhan-tri)

___
*[`multibody/tree/multibody_tree.h` line 3091 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpbU-nASEnkqyaYgeH:-NTplkTM7kPCtzExeKt_:bim8rx8) ([raw file](https://github.com/RobotLocomotion/drake/blob/6360a3e7580f475bf0b2cdcb551a154023e21ad3/multibody/tree/multibody_tree.h#L3091)):*

In this revision, we are making the assumption that these joints are quaternion floating joints in the code (by dynamic casting) regardless how we phrase the comment here. When another type of floating joints is introduced, we would need to rework this logic and possibly loosen the bit-wise identical guarantee and just store a generic RigidTransform instead.
</blockquote></details>

Better I think to delete   ~These associated joints are guaranteed to be
QuaternionFloatingJoints.~ here. I know that's enforced currently elsewhere but if we relax that there won't necessarily be any change required here (free bodies still need a whole pose saved before we know what joint they get). It won't be obvious to whomever makes the change that they need to come back here and delete that sentence.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: 1 unresolved discussion (waiting on @sherm1 and @xuchenhan-tri)

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri and @sherm1)

___
*[`multibody/tree/multibody_tree.h` line 3091 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19230#-NTpbU-nASEnkqyaYgeH:-NTpmsdT66kEpTcIuL-_:b-dlxpgy) ([raw file](https://github.com/RobotLocomotion/drake/blob/6360a3e7580f475bf0b2cdcb551a154023e21ad3/multibody/tree/multibody_tree.h#L3091)):*

Better I think to delete   ~These associated joints are guaranteed to be
QuaternionFloatingJoints.~ here. I know that's enforced currently elsewhere but if we relax that there won't necessarily be any change required here (free bodies still need a whole pose saved before we know what joint they get). It won't be obvious to whomever makes the change that they need to come back here and delete that sentence.

</blockquote></details>

I see your point. Done.

jwnimmer-tri (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19230)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform) (waiting on @xuchenhan-tri)

Issue #18773, opened 2023-02-13
Support a parser option: like URDF_USE_INERTIA_FROM_FILE


RussTedrake:
Failing to parse assets due to poorly specified inertial parameters is a real problem (see, for instance: https://github.com/RobotLocomotion/drake/issues/14436#issuecomment-1427468439)

Other simulators support an option like `URDF_USE_INERTIA_FROM_FILE`.  In bullet, the default behavior is actually to *ignore* the inertial parameters in the file.  I think the default behavior from MuJoCo is more appropriate for us.  

[pybullet documentation](https://github.com/bulletphysics/bullet3/blob/master/docs/pybullet_quickstart_guide/PyBulletQuickstartGuide.md.html#loadurdf): "URDF_USE_INERTIA_FROM_FILE: by default, Bullet recomputed the inertia tensor based on mass and volume of the collision shape. If you can provide more accurate inertia tensor, use this flag."

[mujoco documentation](https://mujoco.readthedocs.io/en/latest/XMLreference.html#compiler) 
inertiafromgeom: [false, true, auto], “auto”
This attribute controls the automatic inference of body masses and inertias from geoms attached to the body. If this setting is “false”, no automatic inference is performed. In that case each body must have explicitly defined mass and inertia with the [inertial](https://mujoco.readthedocs.io/en/latest/XMLreference.html#body-inertial) element, or else a compile error will be generated. If this setting is “true”, the mass and inertia of each body will be inferred from the geoms attached to it, overriding any values specified with the inertial element. The default setting “auto” means that masses and inertias are inferred automatically only when the inertial element is missing in the body definition. One reason to set this attribute to “true” instead of “auto” is to override inertial data imported from a poorly designed model. In particular, a number of publicly available URDF models have seemingly arbitrary inertias which are too large compared to the mass. This results in equivalent inertia boxes which extend far beyond the geometric boundaries of the model. Note that the built-in OpenGL visualizer can render equivalent inertia boxes.

Note that MuJoCo offers a few other inertial-parsing options that we would do well to support, also.

jwnimmer-tri:
Another path to the same goal would be to provide a tool that rewrites the XML file on disk to fix the inertia data (either editing in place, or creating a new copy of the file).

[...]

PR #19238, opened 2023-04-20
[parsing] Improve URDF inertia invalidity handling

rpoyner-tri:
This patch brings inertia invalidity reporting under diagnostic policy, and relaxes inertia problems from hard throws to warnings.

The move to diagnostic policy means that inertia warnings give file names and line numbers, which make it easier to fix problems.

The move from hard throws to warnings means that models with non-physical inertias can still be used with non-simulation workflows, such as visualization, kinematics, and analysis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)

### DIFF
--- a/multibody/parsing/detail_urdf_parser.cc
+++ b/multibody/parsing/detail_urdf_parser.cc
@@ -159,6 +159,32 @@ SpatialInertia<double> UrdfParser::ExtractSpatialInertiaAboutBoExpressedInB(
     ParseScalarAttribute(mass, "value", &body_mass);
   }
 
+  // If physical validity checks fail below, return a prepared plausible
+  // inertia instead. Use a plausible guess for the mass, letting actual mass
+  // validity checking happen later.
+  const double plausible_dummy_mass =
+      (std::isfinite(body_mass) && body_mass > 0.0) ? body_mass : 1.0;
+  // Construct a dummy inertia for a solid sphere, with the density of water,
+  // and radius deduced from the plausible mass.
+  static constexpr double kPlausibleDensity{1000};  // Water is 1000 kg/m³.
+  const double plausible_volume = plausible_dummy_mass / kPlausibleDensity;
+  // Volume = (4/3)π(radius)³, so radius = ³√(3/(4π))(volume).
+  const double plausible_radius =
+      std::cbrt((3.0 / (4.0 * M_PI)) * plausible_volume);
+  // Create Mdum_BBo_B, a plausible spatial inertia for B about Bo (B's origin).
+  // To do this, create Mdum_BBcm (a plausible spatial inertia for B about
+  // Bcm) as a solid sphere and then shift that spatial inertia from Bcm to Bo.
+  // Note: It is unwise to directly create a solid sphere about Bo as this does
+  // not guarantee that the spatial inertia about Bcm is valid. Bcm (B's center
+  // of mass) is the ground-truth point for validity tests.
+  const SpatialInertia<double> Mdum_BBcm =
+    SpatialInertia<double>::SolidSphereWithDensity(
+       kPlausibleDensity, plausible_radius);
+  // Bi's origin is at the COM as documented in
+  // http://wiki.ros.org/urdf/XML/link#Elements
+  const Vector3d& p_BoBcm_B = X_BBi.translation();
+  const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
+
   double ixx = 0;
   double ixy = 0;
   double ixz = 0;
@@ -176,12 +202,26 @@ SpatialInertia<double> UrdfParser::ExtractSpatialInertiaAboutBoExpressedInB(
     ParseScalarAttribute(inertia, "izz", &izz);
   }
 
-  const RotationalInertia<double> I_BBcm_Bi(ixx, iyy, izz, ixy, ixz, iyz);
+  // Yes, catching exceptions violates the coding standard. It is done here to
+  // capture math-aware exceptions into parse-time warnings, since non-physical
+  // inertias are all too common, and the thrown messages are actually pretty
+  // useful.
+  RotationalInertia<double> I_BBcm_Bi;
+  try {
+    // Use the factory method here; it doesn't change its diagnostic behavior
+    // between release and debug builds.
+    I_BBcm_Bi = RotationalInertia<double>::MakeFromMomentsAndProductsOfInertia(
+            ixx, iyy, izz, ixy, ixz, iyz);
+  } catch (const std::exception& e) {
+    Warning(*node, e.what());
+    return Mdum_BBo_B;
+  }
 
   // If this is a massless body, return a zero SpatialInertia.
-  if (body_mass == 0. && I_BBcm_Bi.get_moments().isZero() &&
+  if (body_mass == 0.0 && I_BBcm_Bi.get_moments().isZero() &&
       I_BBcm_Bi.get_products().isZero()) {
-    return SpatialInertia<double>(body_mass, {0., 0., 0.}, {0., 0., 0});
+    return SpatialInertia<double>(
+        0.0, Vector3d::Zero(), UnitInertia<double>{0.0, 0.0, 0.0});
   }
   // B and Bi are not necessarily aligned.
   const math::RotationMatrix<double> R_BBi(X_BBi.rotation());
@@ -189,12 +229,14 @@ SpatialInertia<double> UrdfParser::ExtractSpatialInertiaAboutBoExpressedInB(
   // Re-express in frame B as needed.
   const RotationalInertia<double> I_BBcm_B = I_BBcm_Bi.ReExpress(R_BBi);
 
-  // Bi's origin is at the COM as documented in
-  // http://wiki.ros.org/urdf/XML/link#Elements
-  const Vector3d p_BoBcm_B = X_BBi.translation();
-
-  return SpatialInertia<double>::MakeFromCentralInertia(
-      body_mass, p_BoBcm_B, I_BBcm_B);
+  try {
+    return SpatialInertia<double>::MakeFromCentralInertia(
+        body_mass, p_BoBcm_B, I_BBcm_B);
+  } catch (const std::exception& e) {
+    Warning(*node, e.what());
+    return Mdum_BBo_B;
+  }
+  DRAKE_UNREACHABLE();
 }
 
 void UrdfParser::ParseBody(XMLElement* node, MaterialMap* materials) {
--- a/multibody/parsing/test/detail_urdf_parser_test.cc
+++ b/multibody/parsing/test/detail_urdf_parser_test.cc
@@ -1207,7 +1207,7 @@ TEST_F(UrdfParserTest, PointMass) {
   EXPECT_TRUE(body.default_rotational_inertia().get_products().isZero());
 }
 
-TEST_F(UrdfParserTest, BadInertia) {
+TEST_F(UrdfParserTest, BadInertiaFormats) {
   // Test various mis-formatted inputs.
   constexpr const char* base = R"""(
     <robot name='point_mass'>
@@ -1248,36 +1248,57 @@ TEST_F(UrdfParserTest, BadInertia) {
   EXPECT_THAT(TakeError(), MatchesRegex(".*Expected single value.*izz.*"));
 }
 
-// TODO(rpoyner-tri): these tests don't test the parser but rather error
-// behavior of underlying implementation components. Consider moving or
-// removing them.
-class ZeroMassNonZeroInertiaTest : public UrdfParserTest {
- public:
-  void ParseZeroMassNonZeroInertia() {
-    AddModelFromUrdfString(R"""(
-<robot name='bad'>
-  <link name='bad'>
-    <inertial>
-      <mass value="0"/>
-      <inertia ixx="1" ixy="0" ixz="0" iyy="1" iyz="0" izz="1"/>
-    </inertial>
-  </link>
-</robot>)""", "");
-  }
-};
+TEST_F(UrdfParserTest, BadInertiaValues) {
+  // Test various invalid input values.
+  constexpr const char* base = R"""(
+    <robot name='test'>
+      <link name='test'>
+        <inertial>
+          <mass {}/>
+          <inertia {}/>
+        </inertial>
+      </link>
+    </robot>)""";
 
-TEST_F(ZeroMassNonZeroInertiaTest, ExceptionType) {
-  // Test that attempt to parse links with zero mass and non-zero inertia fails.
-  if (!::drake::kDrakeAssertIsArmed) {
-    EXPECT_THROW(ParseZeroMassNonZeroInertia(), std::runtime_error);
-  }
-}
+  const int num_builtin_models = plant_.num_model_instances();
 
-TEST_F(ZeroMassNonZeroInertiaTest, Message) {
+  // Absurd rotational inertia values.
+  AddModelFromUrdfString(
+      fmt::format(base, "value='1'",
+                  "ixx='1' ixy='4' ixz='9' iyy='16' iyz='25' izz='36'"), "a");
+  EXPECT_THAT(TakeWarning(), MatchesRegex(".*rot.*inertia.*"));
+  // Test some inertia values found in the wild.
+  AddModelFromUrdfString(
+      fmt::format(
+          base, "value='0.038'",
+          "ixx='4.30439933333e-05' ixy='9.57068e-06' ixz='5.1205e-06' "
+          "iyy='1.44451933333e-05' iyz='1.342825e-05' izz='4.30439933333e-05'"),
+      "b");
+  EXPECT_THAT(TakeWarning(), MatchesRegex(".*rot.*inertia.*"));
+  // Negative mass.
+  AddModelFromUrdfString(
+      fmt::format(base, "value='-1'",
+                  "ixx='1' ixy='0' ixz='0' iyy='1' iyz='0' izz='1'"), "c");
+  EXPECT_THAT(TakeWarning(), MatchesRegex(".*mass > 0.*"));
   // Test that attempt to parse links with zero mass and non-zero inertia fails.
-  const std::string expected_message = ".*condition 'mass > 0' failed.";
-  DRAKE_EXPECT_THROWS_MESSAGE(
-      ParseZeroMassNonZeroInertia(), expected_message);
+  AddModelFromUrdfString(
+      fmt::format(base, "value='0'",
+                  "ixx='1' ixy='0' ixz='0' iyy='1' iyz='0' izz='1'"), "d");
+  EXPECT_THAT(TakeWarning(), MatchesRegex(".*mass > 0.*"));
+
+  plant_.Finalize();
+  // Do some basic sanity checking on the plausible mass and inertia generated
+  // when warnings are issued.
+  for (ModelInstanceIndex k(num_builtin_models);
+       k < plant_.num_model_instances(); ++k) {
+    SCOPED_TRACE(fmt::format("model instance {}", k));
+    const auto& body = dynamic_cast<const RigidBody<double>&>(
+        plant_.GetBodyByName("test", k));
+    const double mass = body.default_mass();
+    EXPECT_GT(mass, 0);
+    EXPECT_TRUE(std::isfinite(mass));
+    EXPECT_TRUE(body.default_rotational_inertia().CouldBePhysicallyValid());
+  }
 }
 
 TEST_F(UrdfParserTest, BushingParsing) {


rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUpkpsAxCUJT5PMjdy:-NTUpkpsAxCUJT5PMjdz:b-m6zviy) (no related file):*
Here is an piece of diagnostic output from a model found on Github:


There were a lot more warnings, but the above gives the general idea.

___

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 203 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUuJke5Rn-PT6vAruy:-NTUuJke5Rn-PT6vAruz:b5lsy4z) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L203)):*
> ```cpp
>     return SpatialInertia<double>::MakeFromCentralInertia(
>         body_mass, p_BoBcm_B, I_BBcm_B);
>   } catch (const std::exception& e) {
> ```

To me the biggest risk here is that we might demote programming errors on our part into warnings instead of throws. The scope of that risk is the amount of code within the try-block.

Some of the lines within here _seem_ to me like they would be our fault if they throw, not the users's fault.

It is possible to narrow the try-block(s) to surround fewer operations?

We could create a `const SpatialInertia<double> dummy_point_mass;` up above and then reuse it in a few try-blocks, e.g.,

```c++
// Yes, catching exceptions violates the coding standard. It is done here to
// funnel the plethora of possible exceptions into parse-time warnings, since
// non-physical inertias are all too common, and the thrown messages are
// actually pretty useful.
RotationalInertia<double> I_BBcm_Bi;
try {
  I_BBcm_Bi = RotationalInertia<double>(ixx, iyy, izz, ixy, ixz, iyz);
} catch (const std::exception& e) {
  Warning(*node, e.what());
  return dummy_point_mass;
}

// <-- NOTE: No try-catch around this part. -->

// If this is a massless body, return a zero SpatialInertia.
if (body_mass == 0. && I_BBcm_Bi.get_moments().isZero() &&
    I_BBcm_Bi.get_products().isZero()) {
  return SpatialInertia<double>(body_mass, {0., 0., 0.}, {0., 0., 0});
}

...
```
___
*[`multibody/parsing/detail_urdf_parser.cc` line 207 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUsX1j0opM_XvQnuRn:-NTUsX1j0opM_XvQnuRo:b-qtiayc) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L207)):*
> ```cpp
>     // Return an arbitrary but valid inertia, to allow parsing to
>     // continue.
>     return SpatialInertia<double>::PointMass(1., {0., 0., 0.});
> ```

On first glance, I would have expected to use `body_mass` here, so at least that part of the model is correct, even if the moments are misshapen.  Is there any difficulty with using it? I suppose we would need to clamp it to be `0 < body_mass < inf` but that isn't _too_ onerous.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 207 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUtyIxEo-eKX8wI-Sd:-NTUtyIxEo-eKX8wI-Se:b8wl5j1) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L207)):*
> ```cpp
>     // Return an arbitrary but valid inertia, to allow parsing to
>     // continue.
>     return SpatialInertia<double>::PointMass(1., {0., 0., 0.});
> ```

nit GSG wants `0.0` not `0.` but in any case there's a more canonical spelling.



_Suggestion:_
```cpp
Vector3d::Zero()
```

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUpkpsAxCUJT5PMjdy:-NTUyU2z2K-iJ61qN-Iz:b-83sjoh) (no related file):*

Here is an piece of diagnostic output from a model found on Github:


There were a lot more warnings, but the above gives the general idea.
</blockquote></details>

Nice!

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 203 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUuJke5Rn-PT6vAruy:-NTUzGAo8aVtnb0Zt2el:brjd3qa) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L203)):*
> I see now that my particular example, the body_mass is still unverified,  ...

Never mind, I was right the first time.  The mass is already checked to be zero, so that line cannot throw.

rpoyner-tri (commented):
Excitingly, the patch as currently written emits different error messages in debug builds than in release builds. Thanks, DRAKE_ASSERT_VOID(). I'll at the very least need to adjust the test.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTUqp866sh2yg1dnmlx:b-5xr5rw)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 203 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUuJke5Rn-PT6vAruy:-NTUyqO5C_3TWefNa5fW:b9kgv7s) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L203)):*

> I see now that my particular example, the body_mass is still unverified,  ...

Never mind, I was right the first time.  The mass is already checked to be zero, so that line cannot throw.
</blockquote></details>

I don't know that I have the full catalog, but certainly the RotIn ctor and the ReExpress operation and the SpatIn factory method can throw. See also test results above; we currently get different throws under debug and release mode. The useful content is essentially the same (it's almost always the triangle inequality).
___
*[`multibody/parsing/detail_urdf_parser.cc` line 207 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTUsX1j0opM_XvQnuRn:-NTUy7Fa6vY9j7hXN7N4:b-8i2vo2) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff18089873d96a1fcc9dc44d76f3b23a3a73372f/multibody/parsing/detail_urdf_parser.cc#L207)):*

On first glance, I would have expected to use `body_mass` here, so at least that part of the model is correct, even if the moments are misshapen.  Is there any difficulty with using it? I suppose we would need to clamp it to be `0 < body_mass < inf` but that isn't _too_ onerous.
</blockquote></details>

In this branch, body mass hasn't been checked. It won't be crazy (inf, nan), but could be negative, or zero. I can write an expression that guesses a more plausible mass.

jwnimmer-tri:
My first reaction there is that we should avoid ever hitting Debug-only throwing, either by promoting the throws to be Release as well, or else by calling a different equivalent API that already throws, or by creating a variation of the function that does always throw, etc.  The user should get the same Parser output no matter the build flavor.

rpoyner-tri (commented):
Agreed. No doubt the D_A_V()'s are there because the methods are used in Mbp, so I'll have to step carefully.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTV4MMK2Wx4Ex0Nh6EU:b-w7axmd)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
Seems good.  I will check back tomorrow with fresh eyes before LGTM.

You're welcome to assign platform in the meantime if you like.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTVLYd0-7QGGO4Na26Z:b-shq4us)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 166 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVIiCrB6NDgIwt1T2y:-NTVIiCs-hy9wCO16hln:b-99bdeu) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L166)):*
> ```cpp
>   // validity checking happen later.
>   auto dummy_inertia = SpatialInertia<double>::PointMass(
>       body_mass > 0.0 ? body_mass : 1.0,
> ```

nit IIRC the PointMass will throw on +inf.

In another thread you claimed it can't be inf, but my reading of `ConvertToVector` is that it allows infinities, because that's how istringstream rolls.

I suppose a unit test would be authoritative one way or another.



_Suggestion:_
```cpp
(std::isfinite(body_mass) && body_mass > 0.0)
```
___
*[`multibody/parsing/detail_urdf_parser.cc` line 204 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVJ3HS0tOYoSYu1qjS:-NTVJ3HT7OUlm6WC51qw:bwbnc0u) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L204)):*
> ```cpp
>   if (body_mass == 0.0 && I_BBcm_Bi.get_moments().isZero() &&
>       I_BBcm_Bi.get_products().isZero()) {
>     return SpatialInertia<double>(body_mass, {0.0, 0.0, 0.0}, {0.0, 0.0, 0.0});
> ```

BTW If you prefer, see the suggestion below.

If nothing else, consider `s/body_mass/0.0/` since every time I read this line I always forget that the argument is a known constant, not actually a variable.



_Suggestion:_
```cpp
SpatialInertia<double>::PointMass(0.0, Vector3d::Zero())
```
___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1254 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVKxVY95s_xmngfGwd:-NTVKxVY95s_xmngfGwe:b-8t4rv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/test/detail_urdf_parser_test.cc#L1254)):*
> ```cpp
>   // Test various invalid input values.
>   constexpr const char* base = R"""(
>     <robot name='point_mass'>
> ```

nit It's technically not a point mass unless the inertia is all zeros.



_Suggestion:_
```cpp
foo
```

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 166 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVIiCrB6NDgIwt1T2y:-NTVQcGJ8g9jJBcuYSKa:b-5icum4) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L166)):*

nit IIRC the PointMass will throw on +inf.

In another thread you claimed it can't be inf, but my reading of `ConvertToVector` is that it allows infinities, because that's how istringstream rolls.

I suppose a unit test would be authoritative one way or another.
</blockquote></details>

How would you propose to spell infinity in the input? Everything I've tried ends up erroring from ParseScalarAttribute(). The analogous unit test is here: https://github.com/RobotLocomotion/drake/blob/master/multibody/parsing/test/detail_tinyxml_test.cc#L104-L105
___
*[`multibody/parsing/detail_urdf_parser.cc` line 204 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVJ3HS0tOYoSYu1qjS:-NTVR3Mx0KM1zN1UY_Bz:b-zcaony) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L204)):*

BTW If you prefer, see the suggestion below.

If nothing else, consider `s/body_mass/0.0/` since every time I read this line I always forget that the argument is a known constant, not actually a variable.
</blockquote></details>

PointMass throws on 0 input mass: https://github.com/RobotLocomotion/drake/blob/master/multibody/tree/spatial_inertia.cc#L28-L32

I've improved the spelling as much as I know how.

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 166 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVIiCrB6NDgIwt1T2y:-NTVa1ar2O20U3AhwMs0:bahftvs) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L166)):*

How would you propose to spell infinity in the input? Everything I've tried ends up erroring from ParseScalarAttribute(). The analogous unit test is here: https://github.com/RobotLocomotion/drake/blob/master/multibody/parsing/test/detail_tinyxml_test.cc#L104-L105
</blockquote></details>

Aha.  TIL it's platform-specific.

https://stackoverflow.com/questions/70971921/how-to-parse-floating-point-infinity-from-stdistream

With Apple Clang and libc++ (on macsim) the goldbolt code in that SO link shows that `inf` in the XML would be parsed as infinity.  So using `value='inf'` and testing on macOS would show the problem.

In any case, I'm fine without a unit test now that I've tested manually.  But I do think it's worth spending the code to guard against inf.

jwnimmer-tri (commented):
:lgtm: feature.

> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTZYmshAMjPPUpU5imJ:b-n37teh)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/test/detail_urdf_parser_test.cc` line 1280 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTZY_NQ72zAojsXeG2R:-NTZY_NQ72zAojsXeG2S:bin6gqp) ([raw file](https://github.com/RobotLocomotion/drake/blob/27c1116e9944daa4b29265a206fb0e77d2157789/multibody/parsing/test/detail_urdf_parser_test.cc#L1280)):*
> ```cpp
>       fmt::format(base, "value='-1'",
>                   "ixx='1' ixy='0' ixz='0' iyy='1' iyz='0' izz='1'"), "c");
>   EXPECT_THAT(TakeWarning(), MatchesRegex(".*mass > 0.*fail.*"));
> ```

BTW Gut feeling -- matching vs "fail" here is not critical for testing and hazards churn here as the RotationalInertia messages evolve.

Ditto below.



_Suggestion:_
```cpp
".*mass > 0.*"
```

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 166 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTVIiCrB6NDgIwt1T2y:-NTZZypH9hm3ZpeuLsDv:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/560d4fad774f47bfbdf10bfea962e02c107e82dd/multibody/parsing/detail_urdf_parser.cc#L166)):*

Aha.  TIL it's platform-specific.

https://stackoverflow.com/questions/70971921/how-to-parse-floating-point-infinity-from-stdistream

With Apple Clang and libc++ (on macsim) the goldbolt code in that SO link shows that `inf` in the XML would be parsed as infinity.  So using `value='inf'` and testing on macOS would show the problem.

In any case, I'm fine without a unit test now that I've tested manually.  But I do think it's worth spending the code to guard against inf.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: needs at least two assigned reviewers (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+@ggould-tri for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTZ_Fkp31nxu_DaeQHk:b-irfx5k)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

rpoyner-tri (commented):
See https://github.com/RobotLocomotion/drake/issues/18773 for discussion and context.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTaKM3L0feO7OeQUXRC:b-4rg7za)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

ggould-tri:
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTnbzlDDGv5MMGjEODG:bk9lnay) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

In general my opinion is that once there are inertia parsing errors, the user should expect their dynamics to be wonky, so it's OK to ignore the offset.  The big picture goal is to still allow kinematics in this case.

On the other hand, we _do_ have a valid `X_BBi` already parsed, so it would be natural to use its `.translation()` here instead of `Vector3d::Zero()`.  The user could plausibly claim that unless we _warned_ about ignoring the translation, we should obey it when populating the plant.
</blockquote></details>

I agree if the dynamics are merely "wonky", but not if they crash the sim.  Possibly someone should actually test this with a free body with bad inertia.  (I won't have a chance to do that test for a little while, though; @rpoyner-tri do you by any chance have an example to hand?).

A small nonzero moment would be more polite (I don't know offhand if just dropping in a translation of the point does that -- the COM will still be on the particle, so the translation relative to the COM will be zero, right?).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri and @sherm1)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTno6pI4rOenc7G-Po0:bahb6op) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

I agree if the dynamics are merely "wonky", but not if they crash the sim.  Possibly someone should actually test this with a free body with bad inertia.  (I won't have a chance to do that test for a little while, though; @rpoyner-tri do you by any chance have an example to hand?).

A small nonzero moment would be more polite (I don't know offhand if just dropping in a translation of the point does that -- the COM will still be on the particle, so the translation relative to the COM will be zero, right?).
</blockquote></details>

Tagging @sherm1 for advice.

The situation we have here is that the user has some typo in their URDF relating to the moments or products of inertia.  We have a valid `X_BBi` transform and `mass`; it's only ixx,iyy,izz,ixy,ixz,iyz that we don't know.

Our goal here to warn the user that they have a typo and therefore that we will be ignoring their inertia values, but then still create the Body and give them our best guess at a MultibodyPlant, so they can still use the MbP for kinematics.  If they try simulate the dynamics of course we'd expect a mismatch between what happens vs what they _intended_ to happen.

What's the best "dummy inertia" to use in this case of a user's typo?

The current draft here uses a point mass.  Grant is worried that a point mass on a free body would be trouble.  Should we use a sphere inertia instead of a point, for example?

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @mitiguy and @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTnvoamEij5Pg6Rp3pG:bsrjv09) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

Tagging @sherm1 for advice.

The situation we have here is that the user has some typo in their URDF relating to the moments or products of inertia.  We have a valid `X_BBi` transform and `mass`; it's only ixx,iyy,izz,ixy,ixz,iyz that we don't know.

Our goal here to warn the user that they have a typo and therefore that we will be ignoring their inertia values, but then still create the Body and give them our best guess at a MultibodyPlant, so they can still use the MbP for kinematics.  If they try simulate the dynamics of course we'd expect a mismatch between what happens vs what they _intended_ to happen.

What's the best "dummy inertia" to use in this case of a user's typo?

The current draft here uses a point mass.  Grant is worried that a point mass on a free body would be trouble.  Should we use a sphere inertia instead of a point, for example?
</blockquote></details>

Yes, a point mass is a poor choice here since it is likely to cause new problems. I would suggest something like [SpatialInertia::SolidSphereWithDensity(1000,.1)](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_spatial_inertia.html#a1354f4723023135536629d866ef5de09). That would be a mass of 1.3kg and inertia equivalent to a ball of ice 20cm in diameter (please check my math). An easier-to-picture version might be a smaller steel ball (density 8000 kg/m^3). A radius of 5cm gives 8000 4/3 π 0.05³ = 1.3 kg, less inertia than an ice ball but still enough to avoid trouble.

(SolidSphereWithMass would be easier to think about but I'm not sure we have that one yet -- @mitiguy ? )

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTo2sc_EkDJYm6sfJRU:b-1icuph) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

Yes, a point mass is a poor choice here since it is likely to cause new problems. I would suggest something like [SpatialInertia::SolidSphereWithDensity(1000,.1)](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_spatial_inertia.html#a1354f4723023135536629d866ef5de09). That would be a mass of 1.3kg and inertia equivalent to a ball of ice 20cm in diameter (please check my math). An easier-to-picture version might be a smaller steel ball (density 8000 kg/m^3). A radius of 5cm gives 8000 4/3 π 0.05³ = 1.3 kg, less inertia than an ice ball but still enough to avoid trouble.

(SolidSphereWithMass would be easier to think about but I'm not sure we have that one yet -- @mitiguy ? )
</blockquote></details>

In that case my vote is for `SolidSphereWithDensity` such that `mass` matches.

The question is whether we fix radius and solve for density (from mass), or fix density and solve for radius (from mass).  The latter seems more natural to me, but I don't mind either way.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTo8v5jEo4w0sr9DLZ-:b7804gs) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

In that case my vote is for `SolidSphereWithDensity` such that `mass` matches.

The question is whether we fix radius and solve for density (from mass), or fix density and solve for radius (from mass).  The latter seems more natural to me, but I don't mind either way.
</blockquote></details>

I agree - my preference would be to pick a reasonable material, then solve for the r that gives the desired mass of that stuff.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NToIOt0CHda_oouT056:bibnaij) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

I agree - my preference would be to pick a reasonable material, then solve for the r that gives the desired mass of that stuff.
</blockquote></details>

Here's a version with a sold sphere at water density -- PTAL

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NToIaqp6a1aW-BO6C7V:-NToIaqq54Z66wSLpDHx:b-9mvmw3) ([raw file](https://github.com/RobotLocomotion/drake/blob/7273055564b0bf56272010ac2ddff080611c1733/multibody/parsing/detail_urdf_parser.cc#L168)):*
> ```cpp
>       (std::isfinite(body_mass) && body_mass > 0.0) ? body_mass : 1.0;
>   // Construct a dummy inertia for a solid sphere, with the density of water,
>   // and radius deduced from the plausible mass.
> ```

nit For clarity...



_Suggestion:_
```cpp
  // Construct a dummy inertia for a solid sphere, with the density of water,
  // and radius deduced from the plausible mass. This does NOT use the X_BBi
  // translation vector, so note that our plausible inertia might be in the
  // wrong place.
```

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NToZSfw4uyH4MDvi0PW):*
minor:  Given the semantic depth of the inertia changes, this PR headline is hiding the ball a bit -- you don't just report invalidity, you fix it.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri and @sherm1)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NToIaqp6a1aW-BO6C7V:-NTowelu7XigsOOI1QSq:bvrsdww) ([raw file](https://github.com/RobotLocomotion/drake/blob/7273055564b0bf56272010ac2ddff080611c1733/multibody/parsing/detail_urdf_parser.cc#L168)):*

Oh, good point. This could still fail if we don't put the inertia at the body origin. The plausible inertia being generated is a central inertia (about the sphere origin). If the code is expecting the inertia at the body origin instead, the central inertia should be shifted from the center to the body origin (by the negative of the specified center of mass vector). There is a convenient SpatialInertia::Shift method for that purpose.

I don't think documenting that it is in the wrong place is good enough. It is very likely to violate physicality if we generate a central inertia but then Drake puts it somewhere away from the center. Those tests operate by shifting the inertia back to the com (which shrinks the inertias) and then looking for trouble.
</blockquote></details>

I think I have done what @sherm1 was asking. PTAL

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @sherm1)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 175 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTp6OLD410kSg6hhuvu:-NTp6OLD410kSg6hhuvv:bw61bqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/873e51919553e7663facc73749b5f6bf984451d0/multibody/parsing/detail_urdf_parser.cc#L175)):*
> ```cpp
>       std::cbrt((3.0 / (4.0 * M_PI)) * plausible_volume);
>   auto dummy_inertia = SpatialInertia<double>::SolidSphereWithDensity(
>       kPlausibleDensity, plausible_radius).ShiftInPlace(-X_BBi.translation());
> ```

BTW This looks good to me. If there isn't one already, a unit test with a bad inertia and Bo != Bcm checking that we got the expected about-Bo inertia would be nice to have! Paul @mitiguy could likely generate a sphere inertia but about Bo calculated in some other way so the test is more than a tautology.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @jwnimmer-tri, @rpoyner-tri, and @sherm1)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 175 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTp6OLD410kSg6hhuvu:-NTpJCK79vBDuXU9yfzI:b-5uhj93) ([raw file](https://github.com/RobotLocomotion/drake/blob/873e51919553e7663facc73749b5f6bf984451d0/multibody/parsing/detail_urdf_parser.cc#L175)):*

BTW This looks good to me. If there isn't one already, a unit test with a bad inertia and Bo != Bcm checking that we got the expected about-Bo inertia would be nice to have! Paul @mitiguy could likely generate a sphere inertia but about Bo calculated in some other way so the test is more than a tautology.
</blockquote></details>

Yeah, I had thought about a unit test, but not (yet) about what it should look like. I will investigate.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 175 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTp6OLD410kSg6hhuvu:-NTulhXA05TYYNReu5k_:bnbxqki) ([raw file](https://github.com/RobotLocomotion/drake/blob/873e51919553e7663facc73749b5f6bf984451d0/multibody/parsing/detail_urdf_parser.cc#L175)):*

Yeah, I had thought about a unit test, but not (yet) about what it should look like. I will investigate.
</blockquote></details>

Added some minimal sanity checking of plausible inertias -- PTAL

jwnimmer-tri (commented):
Oops, I never gave my LGTM, did I?  I thought I'd done it.

:lgtm: feature.  Certainly the "parsing best practices" part of this code is in great shape now.

Possibly you should have an actual dynamicist review check math for the "dummy" inertia, though?  I don't know if I fully understand it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTurxhA8bkDBn-tVX5u:b-2k1ne)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+@mitiguy for specialist review of the "plausible dummy inertia" code.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTxeN2V6sPNFn2CWcwU:b-2edkca)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @mitiguy)

jwnimmer-tri:
In our f2f on Monday, we talked about a helper function that both SDFormat and URDF (and more?) would use.  For clarity, here's the kind of signature I had in mind during that discussion.

```c++
/* Combines the given user inputs into a SpatialInertia. When any data is invalid,
warns into the policy and returns a best-effort approximation instead.
@param I_BBcm_Bi is ordered as ixx, iyy, izz, ixy, ixz, iyz */
SpatialInertia<double> ParseSpatialInertia(
    const DiagnosticPolicy& diagnostic,
    const RigidTransformd& X_BBi,
    double mass,
    const Vector6d& I_BBcm_Bi)
```

rpoyner-tri (commented):
I'm not against it. I figured I would roll the refactoring into the SDFormat patch, when we at least reach rule-of-2.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTy-Jjk84JI8qH-fPa2:bonmtyl)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @mitiguy)

rpoyner-tri (commented):
FYI @joemasterjohn here's the PR where the parser tries to substitute a "plausible inertia". PTAL if you have time.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NTyD2Gk1xAvnAdKZHV_:b-qn855a)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @mitiguy)

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTnYeam1_tGd9uBIpLy:-NTpPqIk5aVHCPD8rEr9:b-cqkh3d) ([raw file](https://github.com/RobotLocomotion/drake/blob/a31ce81798ae546977faa5274b3d5243907c002e/multibody/parsing/detail_urdf_parser.cc#L168)):*

Here's a version with a sold sphere at water density -- PTAL
</blockquote></details>

Note: PR #19268 for SolidSphereWithMass() is in review.
___
*[`multibody/parsing/detail_urdf_parser.cc` line 175 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NTp6OLD410kSg6hhuvu:-NTyqkbR-fScvkq3itNG:btxg7ff) ([raw file](https://github.com/RobotLocomotion/drake/blob/873e51919553e7663facc73749b5f6bf984451d0/multibody/parsing/detail_urdf_parser.cc#L175)):*

Added some minimal sanity checking of plausible inertias -- PTAL
</blockquote></details>

Nit  Consider semi-monogram notation and documentation to clarify the intended dummy_inertia.  Also a note as to why the shift is needed is helpful.  Consider code such as:



_Code snippet:_
```cpp
  // Create Mdum_BBo_B, a plausible spatial inertia for B about Bo (B's origin).
  // To do this, create Mdum_BBcm (a plausible spatial inertia for B about
  // Bcm) as a solid sphere and then shift that spatial inertia from Bcm to Bo.
  // Note: It is unwise to directly create a solid sphere about Bo as this does
  // not guarantee that the spatial inertia about Bcm is valid. Bcm (B's center
  // of mass) is the ground-truth point for validty tests.
  const SpatialInertia<double> Mdum_BBcm = 
    SpatialInertia<double>::SolidSphereWithDensity(
       kPlausibleDensity, plausible_radius);
  const Vector3<double>& p_BoBcm_B = X_BBi.translation();  // Bi's origin is Bcm.
  const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
```

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

jwnimmer-tri (commented):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @jwnimmer-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 190 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NU2dhKHAo0rEGtBEf7Q:-NU2dhKIBwiMZ5GALsg7:bsdsrt) ([raw file](https://github.com/RobotLocomotion/drake/blob/51dd0115d597adc926bf7cb93d660d0fec3b23e0/multibody/parsing/detail_urdf_parser.cc#L190)):*
> ```cpp
>   // Tell the user we are constructing a plausible but wrong model, when input
>   // inertia is not usable.
>   static constexpr char warn_format[] =
> ```

nit https://drake.mit.edu/styleguide/cppguide.html#Constant_Names
___
*[`multibody/parsing/detail_urdf_parser.cc` line 192 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NU2eATk7VSQytOlNsku:-NU2eATk7VSQytOlNskv:braa2og) ([raw file](https://github.com/RobotLocomotion/drake/blob/51dd0115d597adc926bf7cb93d660d0fec3b23e0/multibody/parsing/detail_urdf_parser.cc#L192)):*
> ```cpp
>   static constexpr char warn_format[] =
>       "{}\n"
>       "Substituting a plausible-guess inertia '{}' instead.";
> ```

Working

I'm going to re-test locally to see if this is actually an improvement.  On paper, it seems like not.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 2 unresolved discussions, LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 192 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NU2eATk7VSQytOlNsku:-NU2eGjODPXg-EZkWKAk:b-qt26ww) ([raw file](https://github.com/RobotLocomotion/drake/blob/51dd0115d597adc926bf7cb93d660d0fec3b23e0/multibody/parsing/detail_urdf_parser.cc#L192)):*
> Added some text about the substitute inertia to warning text.

Could you explain the thought process?

I put a typo in the Pendulum and ran it.  Here's what I see now:

```
jwnimmer@call-cps:~/jwnimmer-tri/drake$ bazel-bin/multibody/parsing/parser_manual_test examples/pendulum/Pendulum.urdf 
parsing examples/pendulum/Pendulum.urdf
/home/jwnimmer/jwnimmer-tri/drake/examples/pendulum/Pendulum.urdf:66: warning: MakeFromMomentsAndProductsOfInertia(): The rotational inertia
[1  0  0]
[0  0  0]
[0  0  0]
did not pass the test CouldBePhysicallyValid().
The associated principal moments of inertia:
0  0  1
do not satisfy the triangle inequality.
Substituting a plausible-guess inertia '
 mass = 0.5
 Center of mass = [0  0  0]
 Inertia about point P, I_BP =
[0.000484861            0            0]
[          0  0.000484861            0]
[          0            0  0.000484861]
' instead.
Parsed 1 models.
```

First of all, as a nit: the single quotes are useless.  It should be something like "... plausible inertia instead:\n" and then the string output.

What's still not clear to me though is why adding even more voluminous text to the parser output is to the user's advantage.

If we printed out an URL XML stanza with a suggested fix, I could probably buy that.  But as it stands, what do you think a user would do with this extra information?  To me, it's just clutter that makes it more difficult to find the more important messages that will also be nearby.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: 1 unresolved discussion, LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @jwnimmer-tri)

___
*[`multibody/parsing/detail_urdf_parser.cc` line 192 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-NU2eATk7VSQytOlNsku:-NU2hka10D0rhJCvVkDc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/51dd0115d597adc926bf7cb93d660d0fec3b23e0/multibody/parsing/detail_urdf_parser.cc#L192)):*

> Added some text about the substitute inertia to warning text.

Could you explain the thought process?

I put a typo in the Pendulum and ran it.  Here's what I see now:

```
jwnimmer@call-cps:~/jwnimmer-tri/drake$ bazel-bin/multibody/parsing/parser_manual_test examples/pendulum/Pendulum.urdf 
parsing examples/pendulum/Pendulum.urdf
/home/jwnimmer/jwnimmer-tri/drake/examples/pendulum/Pendulum.urdf:66: warning: MakeFromMomentsAndProductsOfInertia(): The rotational inertia
[1  0  0]
[0  0  0]
[0  0  0]
did not pass the test CouldBePhysicallyValid().
The associated principal moments of inertia:
0  0  1
do not satisfy the triangle inequality.
Substituting a plausible-guess inertia '
 mass = 0.5
 Center of mass = [0  0  0]
 Inertia about point P, I_BP =
[0.000484861            0            0]
[          0  0.000484861            0]
[          0            0  0.000484861]
' instead.
Parsed 1 models.
```

First of all, as a nit: the single quotes are useless.  It should be something like "... plausible inertia instead:\n" and then the string output.

What's still not clear to me though is why adding even more voluminous text to the parser output is to the user's advantage.

If we printed out an URL XML stanza with a suggested fix, I could probably buy that.  But as it stands, what do you think a user would do with this extra information?  To me, it's just clutter that makes it more difficult to find the more important messages that will also be nearby.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

mitiguy (commented):
@rpoyner-tri   Looks good.  Overly generous with co-author credit -- but I like the idea that I am co-responsible for what follows :)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NU2xigqBOvhg6n-nw79:b-v1pcvn)* status: LGTM missing from assignees ggould-tri(platform),mitiguy (waiting on @rpoyner-tri)

rpoyner-tri (commented):
It's Monday, and @ggould-tri stamp is still blocking this patch.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NUMdbhmD-SMSvMWzJZe:bclqu9l)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

ggould-tri (commented):
Uh, didn't realize this was blocking.  すみません  :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19238#-:-NUMficc0f6TMD2v6ax-:b-8z4cfu)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),jwnimmer-tri(platform),mitiguy

PR #19240, opened 2023-04-20
Change is_positive_and_finite() to is_positive_finite().  

mitiguy:
Also, use anonymous namespace instead of static.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19240)

### DIFF
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -6,13 +6,15 @@
 
 namespace drake {
 namespace multibody {
-// namespace {
+
+namespace {
 
 template <typename T>
-static const boolean<T> is_positive_and_finite(const T& value) {
+const boolean<T> is_positive_finite(const T& value) {
   using std::isfinite;
   return isfinite(value) && value > 0;
 }
+}  // anonymous namespace
 
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::MakeUnitary() {
@@ -25,7 +27,7 @@ SpatialInertia<T> SpatialInertia<T>::MakeUnitary() {
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::PointMass(
     const T& mass, const Vector3<T>& position) {
-  if (!is_positive_and_finite(mass)) {
+  if (!is_positive_finite(mass)) {
     const std::string error_message = fmt::format(
         "{}(): The mass of a particle is not positive and finite: {}.",
         __func__, mass);
@@ -46,15 +48,15 @@ SpatialInertia<T> SpatialInertia<T>::PointMass(
 template <typename T>
 SpatialInertia<T> SpatialInertia<T>::SolidBoxWithDensity(
     const T& density, const T& lx, const T& ly, const T& lz) {
-  if (!is_positive_and_finite(density)) {
+  if (!is_positive_finite(density)) {
     const std::string error_message = fmt::format(
         "{}(): A solid box's density is not positive and finite: {}.",
         __func__, density);
     throw std::logic_error(error_message);
   }
-  if (!is_positive_and_finite(lx) ||
-      !is_positive_and_finite(ly) ||
-      !is_positive_and_finite(lz)) {
+  if (!is_positive_finite(lx) ||
+      !is_positive_finite(ly) ||
+      !is_positive_finite(lz)) {
     const std::string error_message = fmt::format(
         "{}(): One or more dimensions of a solid box is not positive and "
         "finite: ({}, {}, {}).", __func__, lx, ly, lz);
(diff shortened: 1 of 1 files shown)


mitiguy (commented):
Feature review +@SeanCurtis-TRI 
For both Feature and Platform reviews. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-:-NTVhZ7C4W4155BoUJ6o:bejaeq0)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM: This is fine as far as the PR goes. But this doesn't necessarily make hoisting it out of this file appreciably easier. For me, the primary value of this PR is not the renaming, but the clean up of the namespacing (and elimination of the checked-in, commented-out code from the previous PR).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-:-NTo_7JpDml0cZ6EN8Wp:bg90tk3)* status: 1 unresolved discussion (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.cc` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-NToZln_6dGIpO6cZ4u_:-NToZln_6dGIpO6cZ4ua:b-d6hptb) ([raw file](https://github.com/RobotLocomotion/drake/blob/a54dd41c1a14ccb33d63e8c7881a2c3a73dcc275/multibody/tree/spatial_inertia.cc#L17)):*
> ```cpp
>   return isfinite(value) && value > 0;
> }
> }  // anonymous namespace
> ```

nit This is how anonymous namespaces get documented by our auto-formatting.



_Suggestion:_
```cpp
}  // namespace
```

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19240)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`multibody/tree/spatial_inertia.cc` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-NToZln_6dGIpO6cZ4u_:-NTp4CcV0HL9CKH_jPIG:b10ezyv) ([raw file](https://github.com/RobotLocomotion/drake/blob/a54dd41c1a14ccb33d63e8c7881a2c3a73dcc275/multibody/tree/spatial_inertia.cc#L17)):*

nit This is how anonymous namespaces get documented by our auto-formatting.
</blockquote></details>

OK.  I'll fix in next PR (should be in today).

mitiguy (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-:-NTp4JsODEgVp4BUAXrR:bm9o57a)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI:
*[`multibody/tree/spatial_inertia.cc` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19240#-NToZln_6dGIpO6cZ4u_:-NTp5qEL-A2ULQi9nxO1:b-ovvbz2) ([raw file](https://github.com/RobotLocomotion/drake/blob/a54dd41c1a14ccb33d63e8c7881a2c3a73dcc275/multibody/tree/spatial_inertia.cc#L17)):*

OK.  I'll fix in next PR (should be in today).
</blockquote></details>

In the future, I will be withholding my approval tag. Not responding to a comment and pushing a "bad" PR is exactly how you got to be doing this PR in the first place.

There are many reasons to defer a fix to the next PR. Knowingly introducing typos doesn't seem to be one.

mitiguy:
Sean -- I agree with your overarching point on the :lgtm.
Feel free to hold back on future :lgtm as you see fit.

However, please consider the context.  It was marked as a nit and there are at least 17 other occurrences of  this in our codebase:
`}  // anonymous namespace`

I viewed your response as: "I (or autoformatting) prefer strawberry over vanilla ice-cream" rather than this is a defect and it violates the style guide.  If this really is a defect, I wonder why it is OK in 17 other places?  My mindset is to be agreeable with a "strawberry" preference so I'll change it a pending PR on SolidSphereWithMass() to "strawberry", i.e., changing to:
`
}  // namespace`

jwnimmer-tri:
> I wonder why it is OK in 17 other places?

Sometimes, formatting bugs are missed during code review.

The 17 places are about 1.5% of the total.  The other 1124 places in the code spell it the conventional way.

I'll improve our linter to flag the unusual spelling.

jwnimmer-tri:
=> https://github.com/RobotLocomotion/drake/pull/19270

SeanCurtis-TRI:
Paul, reading your comment, I think there are one or two ideas that bear elaboration. You raised two issues: what does "nit" mean and what do you do for "strawberry vs vanilla" issues.

1. Meaning of nit. This isn't guaranteed to be universal (although it seems to generally be the case), "nit" is synonymous with "minor". It indicates a small defect that is simply resolved and most likely doesn't require discussion. A classic example would be a misspelling (although there are plenty of non-linted C++ issues that could likewise qualify).
    - There are, of course, multiple good reasons to have a discussion in response to a "nit". The nit is mistaken. The nit comes with a suggestion to address the defect, but the author resolves the defect in an alternate manner.
    - The point of the nit is that it should block until its addressed, but it doesn't require a song and dance to address.
2. "strawberry vs vanilla". These are frequently prefixed with "btw" or "fyi". Some reviewers eschew these entirely in favor of keeping the review conversation purely about defects and correctness. Sometimes they are used if the reviewer suspects a good alternative might not have occurred to the author. These are 100% author's discretion and can simply be dismissed with a "no thank you". (Technically, they can be silently dismissed by clicking "Resolved", but it's considered good form to communicate that you read it and haven't chosen otherwise.)

So, ignoring about whether my "nit" communicated my intended severity or not, you say you interpreted it as a BTW. However, your response to the BTW isn't consistent with a BTW. If it is *truly* an aesthetic change that you don't care for, you should say so and not plan on doing it in a follow up PR *at all*. If you're planning on doing it, and it's as small and trivial as this one was, then it makes the most sense to simply do it in the PR where the discussion took place. Deferring changes to future PRs is usually reserved for uglier changes and they usually get bookmarked with a TODO.

I hope this clarifies two things: what is (probably) being communicated with the various prefixes (nit, minor, btw, fyi) and responses consistent to those dispositions.

PR #19158, opened 2023-04-07
[visualization] Add render camera preview

jwnimmer-tri:
Caveat: This won't work in Deepnote until #18862 and related are finished.  It uses a GL native window as the preview for now.

[Screencast from 04-07-2023 12:48:36 PM.webm](https://user-images.githubusercontent.com/17596505/230668625-da9f88df-2c4b-4b2b-9e62-025b3c5b9897.webm)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19158)

PR #19243, opened 2023-04-21
[systems] Add DiagramBuilder::RemoveSystem

jwnimmer-tri:
Towards #19158.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243)

### DIFF
--- a/bindings/pydrake/systems/framework_py_semantics.cc
+++ b/bindings/pydrake/systems/framework_py_semantics.cc
@@ -548,6 +548,8 @@ void DoScalarDependentDefinitions(py::module m) {
           py::keep_alive<1, 0>(),
           // Keep alive, ownership: `system` keeps `self` alive.
           py::keep_alive<3, 1>(), doc.DiagramBuilder.AddNamedSystem.doc)
+      .def("RemoveSystem", &DiagramBuilder<T>::RemoveSystem, py::arg("system"),
+          doc.DiagramBuilder.RemoveSystem.doc)
       .def("empty", &DiagramBuilder<T>::empty, doc.DiagramBuilder.empty.doc)
       .def("already_built", &DiagramBuilder<T>::already_built,
           doc.DiagramBuilder.already_built.doc)
--- a/bindings/pydrake/systems/test/general_test.py
+++ b/bindings/pydrake/systems/test/general_test.py
@@ -842,6 +842,16 @@ class TestGeneral(unittest.TestCase):
         html = GenerateHtml(system, initial_depth=2)
         self.assertRegex(html, r'key: "zoh"')
 
+    def test_diagram_builder_remove(self):
+        builder = DiagramBuilder()
+        source = builder.AddSystem(ConstantVectorSource([0.0]))
+        adder = builder.AddSystem(Adder(1, 1))
+        builder.ExportOutput(source.get_output_port())
+        builder.RemoveSystem(adder)  # N.B. Deletes 'adder'; don't use after!
+        diagram = builder.Build()
+        self.assertEqual(diagram.num_input_ports(), 0)
+        self.assertEqual(diagram.num_output_ports(), 1)
+
     def test_diagram_fan_out(self):
         builder = DiagramBuilder()
         adder = builder.AddSystem(Adder(7, 1))
--- a/systems/framework/diagram.h
+++ b/systems/framework/diagram.h
@@ -47,8 +47,12 @@ class OwnedSystems {
   decltype(auto) begin() const { return vec_.begin(); }
   decltype(auto) end() const { return vec_.end(); }
   decltype(auto) operator[](size_t i) const { return vec_[i]; }
+  decltype(auto) operator[](size_t i) { return vec_[i]; }
   void push_back(std::unique_ptr<System<T>>&& sys) {
     vec_.push_back(std::move(sys));
+  }
+  void pop_back() {
+    vec_.pop_back();
   }
 
  private:
--- a/systems/framework/diagram_builder.cc
+++ b/systems/framework/diagram_builder.cc
@@ -11,12 +11,106 @@
 
 namespace drake {
 namespace systems {
+namespace {
+
+/* Erases the i'th element of vec, shifting everything after it over by one. */
+template <typename StdVector>
+void VectorErase(StdVector* vec, size_t i) {
+  DRAKE_DEMAND(vec != nullptr);
+  const size_t size = vec->size();
+  DRAKE_DEMAND(i < size);
+  for (size_t hole = i; (hole + 1) < size; ++hole) {
+    (*vec)[hole] = std::move((*vec)[hole + 1]);
+  }
+  vec->pop_back();
+}
+
+}  // namespace
 
 template <typename T>
 DiagramBuilder<T>::DiagramBuilder() {}
 
 template <typename T>
 DiagramBuilder<T>::~DiagramBuilder() {}
+
+template <typename T>
+void DiagramBuilder<T>::RemoveSystem(const System<T>& system) {
+  ThrowIfAlreadyBuilt();
+  if (systems_.count(&system) == 0) {
+    throw std::logic_error(fmt::format(
+        "Cannot RemoveSystem on {} because it has not been added to this "
+        "DiagramBuilder",
+        system.GetSystemPathname()));
+  }
+  const size_t system_index = std::distance(
+      registered_systems_.begin(),
+      std::find_if(registered_systems_.begin(), registered_systems_.end(),
+                   [&system](const std::unique_ptr<System<T>>& item) {
+                     return item.get() == &system;
+                   }));
+  DRAKE_DEMAND(system_index < registered_systems_.size());
+
+  // Un-export any input ports associated with this system.
+  // First, undo the ConnectInput.
+  std::set<std::string> disconnected_diagram_input_port_names;
+  for (size_t i = 0; i < input_port_ids_.size();) {
+    const InputPortLocator& locator = input_port_ids_[i];
+    if (locator.first == &system) {
+      const size_t num_erased = diagram_input_set_.erase(locator);
+      DRAKE_DEMAND(num_erased == 1);
+      disconnected_diagram_input_port_names.insert(
+          std::move(input_port_names_[i]));
+      VectorErase(&input_port_ids_, i);
+      VectorErase(&input_port_names_, i);
+    } else {
+      ++i;
+    }
+  }
+  // Second, undo the DeclareInput (iff it was the last connected system).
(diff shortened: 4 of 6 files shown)


jwnimmer-tri:
+@rpoyner-tri for feature review, please.

+@sherm1 for high-level feature review, please -- as well as platform review.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243)* status: 1 unresolved discussion, LGTM missing from assignees rpoyner-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri, @rpoyner-tri, and @sherm1)

___
*[`systems/framework/diagram_builder.cc` line 98 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NToL0ce5MpGtZut4DqZ:-NToL0ce5MpGtZut4Dq_:b-uhfbzm) ([raw file](https://github.com/RobotLocomotion/drake/blob/b016700ce99257d1715ab845331580bc99fb17ab/systems/framework/diagram_builder.cc#L98)):*
> ```cpp
>   }
> 
>   // Disconnect any internal connections associated with this system.
> ```

Whoops.  I forgot to test coverage for this block.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243)* status: LGTM missing from assignees rpoyner-tri(platform),sherm1(platform) (waiting on @rpoyner-tri and @sherm1)

___
*[`systems/framework/diagram_builder.cc` line 98 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NToL0ce5MpGtZut4DqZ:-NToM5nV8_2JGtEB1bB9:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b016700ce99257d1715ab845331580bc99fb17ab/systems/framework/diagram_builder.cc#L98)):*

Whoops.  I forgot to test coverage for this block.
</blockquote></details>

Done.

rpoyner-tri (commented):
:lgtm:

> Reviewed 2 of 6 files at r1, 2 of 4 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-:-NTonnsP3TChqj5GVTyK:bnfp4nl)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform) (waiting on @jwnimmer-tri and @sherm1)

___
*[`systems/framework/diagram_builder.h` line 255 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTolZov39reZKCIqsCs:-NTolZov39reZKCIqsCt:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/f88beb5383a7a607c3a6871a289be25512475d85/systems/framework/diagram_builder.h#L255)):*
> ```objc
>   /// decrementing the indices of all higher-numbered ports that remain.
>   ///
>   /// @warning Because a DigramBuilder owns the objects it contains, the system
> ```

typo



_Suggestion:_
```objc
DiagramBuilder
```

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243)* status: LGTM missing from assignee sherm1(platform) (waiting on @sherm1)

sherm1 (commented):
It's not clear to me how this is related to render camera preview in #19158 -- please explain if that is the motivation for this PR.

High-level feature, and platform :lgtm: with a few minor comments

> Reviewed 2 of 6 files at r1, 2 of 4 files at r2, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-:-NTp3a9G5KCd8dxYws5P:b188xs8)* status: 3 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`systems/framework/diagram_builder.h` line 256 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTowk2y5m3tu9R2bIrV:-NTowk2y5m3tu9R2bIrW:b-nn3jbv) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.h#L256)):*
> ```objc
>   ///
>   /// @warning Because a DiagramBuilder owns the objects it contains, the system
>   /// will be deleted.
> ```

BTW consider mentioning that this must be done prior to Build() -- might not be obvious to someone who doesn't realize Build() transfers ownership out.
___
*[`systems/framework/diagram_builder.cc` line 51 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NToxZPQC9upLXb830q5:-NToxZPQC9upLXb830q6:b-w0i7kl) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.cc#L51)):*
> ```cpp
>                      return item.get() == &system;
>                    }));
>   DRAKE_DEMAND(system_index < registered_systems_.size());
> ```

BTW I doubt there is a performance concern here, but you could use a find_if==end result to issue the above error message without doing the search twice. I think system_index==registered_systems_.size() is the same condition as systems_.count()==0.
___
*[`systems/framework/test/diagram_builder_test.cc` line 64 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTp3JF540hhmKFhHIik:-NTp3JF540hhmKFhHIil:b5anzbr) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/test/diagram_builder_test.cc#L64)):*
> ```cpp
>   //
>   // This setup is carefully chosen such that removing 'adder' will cover all
>   // branching conditions within the implementation.
> ```

BTW confirmed with kcov

The only thing that wasn't covered was CheckInvariants() which is called only in Debug. You could consider a manual call here in the test cases FWIW.

jwnimmer-tri (commented):
> It's not clear to me how this is related to render camera preview in #19158 -- please explain if that is the motivation for this PR.

That PR contains a call to this new function:

```py
            # Disable LCM image transmission. It has a non-trivial cost, and
            # at the moment Meldis can't display LCM images anyway.
            self._builder.builder().RemoveSystem(camera_publisher)
````

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-:-NTp5m0q4bS50VX4m0qo:blaktl8)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),sherm1(platform) (waiting on @jwnimmer-tri)

___
*[`systems/framework/diagram_builder.h` line 256 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTowk2y5m3tu9R2bIrV:-NTp6rFYCV4Sp8N6NJO3:b8s1pek) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.h#L256)):*

BTW consider mentioning that this must be done prior to Build() -- might not be obvious to someone who doesn't realize Build() transfers ownership out.
</blockquote></details>

That's true of all functions in the the header.  I won't try to fix it in this PR.
___
*[`systems/framework/diagram_builder.cc` line 51 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NToxZPQC9upLXb830q5:-NTp6IUG3LcxSrGUaSme:b-8ol42t) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.cc#L51)):*

BTW I doubt there is a performance concern here, but you could use a find_if==end result to issue the above error message without doing the search twice. I think system_index==registered_systems_.size() is the same condition as systems_.count()==0.
</blockquote></details>

I don't think performance carries any weight here.
___
*[`systems/framework/test/diagram_builder_test.cc` line 64 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTp3JF540hhmKFhHIik:-NTp6zPgEgKBCSgxdhc0:b-qm2pht) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/test/diagram_builder_test.cc#L64)):*

BTW confirmed with kcov

The only thing that wasn't covered was CheckInvariants() which is called only in Debug. You could consider a manual call here in the test cases FWIW.
</blockquote></details>

For my own taste, running `bazel test //systems/framework:diagram_builder_test -c dbg` during development was sufficient.  Checking the invariants buys us the most when they are running continuously in all test cases, not just one line here.

jwnimmer-tri:
Merging now, to unblock a train of coming PRs.  If there are other lingering improvements I'm happy to fix in a follow-up.

sherm1:
*[`systems/framework/diagram_builder.cc` line 18 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTtEara27z_JyrTXIF_:-NTtEara27z_JyrTXIFa:b6ht1q) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.cc#L18)):*
> ```cpp
> /* Erases the i'th element of vec, shifting everything after it over by one. */
> template <typename StdVector>
> void VectorErase(StdVector* vec, size_t i) {
> ```

FYI (irrelevent here where performance is not a concern, but I should have at least mentioned it): 
When an std::vector is used as a set (rather than a sequence) you can do O(1) erasure of element i by moving the last element of the vector to position i and shortening the vector by 1. Less renumbering required also.

jwnimmer-tri:
*[`systems/framework/diagram_builder.cc` line 18 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19243#-NTtEara27z_JyrTXIF_:-NTtGMlzERxlzZjVEL9f:bl7yu53) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a1d98b2a3b41b35c1bb7aa3ff4f6e5c37fbd89a/systems/framework/diagram_builder.cc#L18)):*

FYI (irrelevent here where performance is not a concern, but I should have at least mentioned it): 
When an std::vector is used as a set (rather than a sequence) you can do O(1) erasure of element i by moving the last element of the vector to position i and shortening the vector by 1. Less renumbering required also.
</blockquote></details>

I actually did that initially.  It fails to meet the API promise that the relative order of the systems and ports remains unchanged.

PR #19245, opened 2023-04-22
[parsing] Improve SDFormat inertia handling

rpoyner-tri:
This patch brings inertia invalidity reporting under diagnostic policy, and relaxes inertia problems from hard throws to warnings.

The move to diagnostic policy means that inertia warnings give file names and line numbers, which make it easier to fix problems. In the case of SDFormat, this patch demotes the useless error message from the underlying sdformat library to a warning, and emits more detailed warnings in the later translation-to-Drake phase.

The move from hard throws to warnings means that models with non-physical inertias can still be used with non-simulation workflows, such as visualization, kinematics, and analysis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)

### DIFF
--- a/multibody/parsing/BUILD.bazel
+++ b/multibody/parsing/BUILD.bazel
@@ -488,6 +488,7 @@ drake_cc_googletest(
     name = "detail_common_test",
     deps = [
         ":detail_misc",
+        "//common/test_utilities",
     ],
 )
 
--- a/multibody/parsing/detail_common.cc
+++ b/multibody/parsing/detail_common.cc
@@ -9,6 +9,7 @@ namespace multibody {
 namespace internal {
 
 using drake::internal::DiagnosticPolicy;
+using Eigen::Vector3d;
 
 bool EndsWithCaseInsensitive(std::string_view str, std::string_view ext) {
   if (ext.size() > str.size()) { return false; }
@@ -242,6 +243,77 @@ void ParseCollisionFilterGroupCommon(
         next_sibling_element, has_attribute, read_string_attribute,
         read_bool_attribute, read_tag_string);
   }
+}
+
+SpatialInertia<double> ParseSpatialInertia(
+    const DiagnosticPolicy& diagnostic,
+    const math::RigidTransformd& X_BBi,
+    double mass,
+    const InertiaInputs& inertia_Bi_Bi) {
+  // If physical validity checks fail below, return a prepared plausible
+  // inertia instead. Use a plausible guess for the mass, letting actual mass
+  // validity checking happen later.
+  const double plausible_dummy_mass =
+      (std::isfinite(mass) && mass > 0.0) ? mass : 1.0;
+  // Construct a dummy inertia for a solid sphere, with the density of water,
+  // and radius deduced from the plausible mass.
+  static constexpr double kPlausibleDensity{1000};  // Water is 1000 kg/m³.
+  const double plausible_volume = plausible_dummy_mass / kPlausibleDensity;
+  // Volume = (4/3)π(radius)³, so radius = ³√(3/(4π))(volume).
+  const double plausible_radius =
+      std::cbrt((3.0 / (4.0 * M_PI)) * plausible_volume);
+  // Create Mdum_BBo_B, a plausible spatial inertia for B about Bo (B's origin).
+  // To do this, create Mdum_BBcm (a plausible spatial inertia for B about
+  // Bcm) as a solid sphere and then shift that spatial inertia from Bcm to Bo.
+  // Note: It is unwise to directly create a solid sphere about Bo as this does
+  // not guarantee that the spatial inertia about Bcm is valid. Bcm (B's center
+  // of mass) is the ground-truth point for validity tests.
+  const SpatialInertia<double> Mdum_BBcm =
+    SpatialInertia<double>::SolidSphereWithDensity(
+       kPlausibleDensity, plausible_radius);
+  const Vector3d& p_BoBcm_B = X_BBi.translation();
+  const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
+
+  // Yes, catching exceptions violates the coding standard. It is done here to
+  // capture math-aware exceptions into parse-time warnings, since non-physical
+  // inertias are all too common, and the thrown messages are actually pretty
+  // useful.
+  //
+  // The safety and correctness of the catch blocks here relies on the promises
+  // of the *Inertia classes to satisfy the "basic exception guarantee".
+  // See https://en.cppreference.com/w/cpp/language/exceptions
+  RotationalInertia<double> I_BBcm_Bi;
+  try {
+    // Use the factory method here; it doesn't change its diagnostic behavior
+    // between release and debug builds.
+    I_BBcm_Bi = RotationalInertia<double>::MakeFromMomentsAndProductsOfInertia(
+            inertia_Bi_Bi.ixx, inertia_Bi_Bi.iyy, inertia_Bi_Bi.izz,
+            inertia_Bi_Bi.ixy, inertia_Bi_Bi.ixz, inertia_Bi_Bi.iyz);
+  } catch (const std::exception& e) {
+    diagnostic.Warning(e.what());
+    return Mdum_BBo_B;
+  }
+
+  // If this is a massless body, return a zero SpatialInertia.
+  if (mass == 0.0 && I_BBcm_Bi.get_moments().isZero() &&
+      I_BBcm_Bi.get_products().isZero()) {
+    return SpatialInertia<double>(
+        0.0, Vector3d::Zero(), UnitInertia<double>{0.0, 0.0, 0.0});
+  }
+  // B and Bi are not necessarily aligned.
+  const math::RotationMatrix<double>& R_BBi = X_BBi.rotation();
+
+  // Re-express in frame B as needed.
+  const RotationalInertia<double> I_BBcm_B = I_BBcm_Bi.ReExpress(R_BBi);
+
+  try {
+    return SpatialInertia<double>::MakeFromCentralInertia(
+        mass, p_BoBcm_B, I_BBcm_B);
+  } catch (const std::exception& e) {
+    diagnostic.Warning(e.what());
+    return Mdum_BBo_B;
+  }
+  DRAKE_UNREACHABLE();
 }
 
 }  // namespace internal
--- a/multibody/parsing/detail_common.h
+++ b/multibody/parsing/detail_common.h
@@ -16,6 +16,7 @@
 #include "drake/multibody/plant/coulomb_friction.h"
 #include "drake/multibody/plant/multibody_plant.h"
 #include "drake/multibody/tree/linear_bushing_roll_pitch_yaw.h"
+#include "drake/multibody/tree/spatial_inertia.h"
 
 namespace drake {
 namespace multibody {
@@ -212,6 +213,47 @@ void ParseCollisionFilterGroupCommon(
         read_bool_attribute,
     const std::function<std::string(const ElementNode&, const char*)>&
         read_tag_string);
+
+// This struct helps label and order the rotational inertia inputs at call
+// sites of ParseSpatialInertia. Units of all quantities are kg⋅m².
+//
+// These quantities should have been parsed from the URDF
+// `/robot/link/inertial/inertia` tag or from the SDFormat
+// `//model/link/inertial/inertia` tag.
+struct InertiaInputs {
+  // Moments.
+  double ixx{};
+  double iyy{};
+  double izz{};
+  // Products of inertia.
+  double ixy{};
+  double ixz{};
+  double iyz{};
+};
+
+// Combines the given user inputs into a SpatialInertia. When any data is
+// invalid, emits a warning into the diagnostic policy and returns a
+// best-effort approximation instead.
+//
+// The inertia math defines an "inertia frame" (called here Bi), with
+// origin at the center of mass of the body, and axes arbitrary, but sometimes
+// chosen to zero out the products of inertia.
+//
+// Conveniently, both URDF and SDFormat inputs support compatible formats for
+// specifying the inertia frame. Callers should have previously constructed the
+// transform `X_BBi` from the URDF `/robot/link/inertial/origin` tag or from
+// the SDFormat `//model/link/inertial/pose` tag.
+//
+// @param diagnostic  The error-reporting channel.
+// @param X_BBi       Pose of the inertia frame expressed in the body's frame.
+// @param mass        Mass of the body.
+// @param inertia_Bi_Bi  The moments and products of inertia about Bi's origin,
+//                      expressed in frame Bi.
+SpatialInertia<double> ParseSpatialInertia(
+    const drake::internal::DiagnosticPolicy& diagnostic,
+    const math::RigidTransformd& X_BBi,
+    double mass,
+    const InertiaInputs& inertia_Bi_Bi);
 
 }  // namespace internal
 }  // namespace multibody
--- a/multibody/parsing/detail_sdf_diagnostic.cc
+++ b/multibody/parsing/detail_sdf_diagnostic.cc
@@ -87,6 +87,7 @@ bool SDFormatDiagnostic::PropagateErrors(
 
 bool IsError(const sdf::Error& report) {
   switch (report.Code()) {
+    case sdf::ErrorCode::LINK_INERTIA_INVALID:
     case sdf::ErrorCode::ELEMENT_DEPRECATED:
     case sdf::ErrorCode::VERSION_DEPRECATED:
     case sdf::ErrorCode::NONE: {
--- a/multibody/parsing/detail_sdf_parser.cc
+++ b/multibody/parsing/detail_sdf_parser.cc
@@ -159,24 +159,6 @@ std::string GetRelativeBodyName(
   }
 }
 
-// Given a gz::math::Inertial object, extract a RotationalInertia object
-// for the rotational inertia of body B, about its center of mass Bcm and,
-// expressed in the inertial frame Bi (as specified in <inertial> in the SDF
-// file.)
-RotationalInertia<double> ExtractRotationalInertiaAboutBcmExpressedInBi(
-    const gz::math::Inertiald &inertial) {
-  // TODO(amcastro-tri): Verify that gz::math::Inertial::MOI() ALWAYS is
-  // expresed in the body frame B, regardless of how a user might have
-  // specified frames in the sdf file. That is, that it always returns R_BBcm_B.
-  // TODO(amcastro-tri): Verify that gz::math::Inertial::MassMatrix()
-  // ALWAYS is in the inertial frame Bi, regardless of how a user might have
-  // specified frames in the sdf file. That is, that it always returns
-  // M_BBcm_Bi.
-  const gz::math::Matrix3d I = inertial.MassMatrix().Moi();
-  return RotationalInertia<double>(I(0, 0), I(1, 1), I(2, 2),
-                                   I(1, 0), I(2, 0), I(2, 1));
-}
-
 // This takes an `sdf::SemanticPose`, which defines a pose relative to a frame,
 // and resolves its value with respect to another frame.
 math::RigidTransformd ResolveRigidTransform(
@@ -220,17 +202,10 @@ std::string ResolveJointChildLinkName(
 // frame origin Bo and, expressed in body frame B, from a gz::Inertial
 // object.
 SpatialInertia<double> ExtractSpatialInertiaAboutBoExpressedInB(
+    const SDFormatDiagnostic& diagnostic,
+    const sdf::ElementPtr link_element,
     const gz::math::Inertiald& Inertial_BBcm_Bi) {
   double mass = Inertial_BBcm_Bi.MassMatrix().Mass();
-
-  const RotationalInertia<double> I_BBcm_Bi =
-      ExtractRotationalInertiaAboutBcmExpressedInBi(Inertial_BBcm_Bi);
-
-  // If this is a massless body, return a zero SpatialInertia.
-  if (mass == 0. && I_BBcm_Bi.get_moments().isZero() &&
-      I_BBcm_Bi.get_products().isZero()) {
-    return SpatialInertia<double>(mass, {0., 0., 0.}, {0., 0., 0});
-  }
 
   // Pose of the "<inertial>" frame Bi in the body frame B.
   // TODO(amcastro-tri): Verify we don't get funny results when X_BBi is not
@@ -241,20 +216,18 @@ SpatialInertia<double> ExtractSpatialInertiaAboutBoExpressedInB(
   // give us X_BI. Verify this.
   const RigidTransformd X_BBi = ToRigidTransform(Inertial_BBcm_Bi.Pose());
 
-  // B and Bi are not necessarily aligned.
-  const RotationMatrixd R_BBi = X_BBi.rotation();
-
-  // Re-express in frame B as needed.
-  const RotationalInertia<double> I_BBcm_B = I_BBcm_Bi.ReExpress(R_BBi);
-
-  // Bi's origin is at the COM as documented in
-  // http://sdformat.org/spec?ver=1.6&elem=link#inertial_pose
-  const Vector3d p_BoBcm_B = X_BBi.translation();
-
-  // Return the spatial inertia M_BBo_B of body B, about its body frame origin
-  // Bo, and expressed in the body frame B.
-  return SpatialInertia<double>::MakeFromCentralInertia(
-      mass, p_BoBcm_B, I_BBcm_B);
+  // TODO(amcastro-tri): Verify that gz::math::Inertial::MOI() ALWAYS is
+  // expresed in the body frame B, regardless of how a user might have
+  // specified frames in the sdf file. That is, that it always returns R_BBcm_B.
+  // TODO(amcastro-tri): Verify that gz::math::Inertial::MassMatrix()
+  // ALWAYS is in the inertial frame Bi, regardless of how a user might have
+  // specified frames in the sdf file. That is, that it always returns
+  // M_BBcm_Bi.
+  const gz::math::Matrix3d I = Inertial_BBcm_Bi.MassMatrix().Moi();
+  return ParseSpatialInertia(diagnostic.MakePolicyForNode(*link_element),
+                             X_BBi, mass,
+                             {.ixx = I(0, 0), .iyy = I(1, 1), .izz = I(2, 2),
+                              .ixy = I(1, 0), .ixz = I(2, 0), .iyz = I(2, 1)});
 }
 
 // Helper method to retrieve a Body given the name of the link specification.
@@ -924,7 +897,8 @@ std::optional<std::vector<LinkInfo>> AddLinksFromSpecification(
     const gz::math::Inertiald& Inertial_Bcm_Bi = link.Inertial();
 
     const SpatialInertia<double> M_BBo_B =
-        ExtractSpatialInertiaAboutBoExpressedInB(Inertial_Bcm_Bi);
+        ExtractSpatialInertiaAboutBoExpressedInB(
+            diagnostic, link_element, Inertial_Bcm_Bi);
 
     // Add a rigid body to model each link.
     const RigidBody<double>& body =
--- a/multibody/parsing/detail_urdf_parser.cc
+++ b/multibody/parsing/detail_urdf_parser.cc
@@ -160,84 +160,18 @@ SpatialInertia<double> UrdfParser::ExtractSpatialInertiaAboutBoExpressedInB(
     ParseScalarAttribute(mass, "value", &body_mass);
   }
 
-  // If physical validity checks fail below, return a prepared plausible
-  // inertia instead. Use a plausible guess for the mass, letting actual mass
-  // validity checking happen later.
-  const double plausible_dummy_mass =
-      (std::isfinite(body_mass) && body_mass > 0.0) ? body_mass : 1.0;
-  // Construct a dummy inertia for a solid sphere, with the density of water,
-  // and radius deduced from the plausible mass.
-  static constexpr double kPlausibleDensity{1000};  // Water is 1000 kg/m³.
-  const double plausible_volume = plausible_dummy_mass / kPlausibleDensity;
-  // Volume = (4/3)π(radius)³, so radius = ³√(3/(4π))(volume).
-  const double plausible_radius =
-      std::cbrt((3.0 / (4.0 * M_PI)) * plausible_volume);
-  // Create Mdum_BBo_B, a plausible spatial inertia for B about Bo (B's origin).
-  // To do this, create Mdum_BBcm (a plausible spatial inertia for B about
-  // Bcm) as a solid sphere and then shift that spatial inertia from Bcm to Bo.
-  // Note: It is unwise to directly create a solid sphere about Bo as this does
-  // not guarantee that the spatial inertia about Bcm is valid. Bcm (B's center
-  // of mass) is the ground-truth point for validity tests.
-  const SpatialInertia<double> Mdum_BBcm =
-    SpatialInertia<double>::SolidSphereWithDensity(
-       kPlausibleDensity, plausible_radius);
-  // Bi's origin is at the COM as documented in
-  // http://wiki.ros.org/urdf/XML/link#Elements
-  const Vector3d& p_BoBcm_B = X_BBi.translation();
-  const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
-
-  double ixx = 0;
-  double ixy = 0;
-  double ixz = 0;
-  double iyy = 0;
-  double iyz = 0;
-  double izz = 0;
-
+  InertiaInputs inputs;
   XMLElement* inertia = node->FirstChildElement("inertia");
   if (inertia) {
-    ParseScalarAttribute(inertia, "ixx", &ixx);
-    ParseScalarAttribute(inertia, "ixy", &ixy);
-    ParseScalarAttribute(inertia, "ixz", &ixz);
-    ParseScalarAttribute(inertia, "iyy", &iyy);
-    ParseScalarAttribute(inertia, "iyz", &iyz);
-    ParseScalarAttribute(inertia, "izz", &izz);
+    ParseScalarAttribute(inertia, "ixx", &inputs.ixx);
+    ParseScalarAttribute(inertia, "ixy", &inputs.ixy);
+    ParseScalarAttribute(inertia, "ixz", &inputs.ixz);
+    ParseScalarAttribute(inertia, "iyy", &inputs.iyy);
+    ParseScalarAttribute(inertia, "iyz", &inputs.iyz);
+    ParseScalarAttribute(inertia, "izz", &inputs.izz);
   }
-
-  // Yes, catching exceptions violates the coding standard. It is done here to
-  // capture math-aware exceptions into parse-time warnings, since non-physical
-  // inertias are all too common, and the thrown messages are actually pretty
-  // useful.
-  RotationalInertia<double> I_BBcm_Bi;
-  try {
-    // Use the factory method here; it doesn't change its diagnostic behavior
-    // between release and debug builds.
-    I_BBcm_Bi = RotationalInertia<double>::MakeFromMomentsAndProductsOfInertia(
-            ixx, iyy, izz, ixy, ixz, iyz);
-  } catch (const std::exception& e) {
-    Warning(*node, e.what());
-    return Mdum_BBo_B;
-  }
-
-  // If this is a massless body, return a zero SpatialInertia.
-  if (body_mass == 0.0 && I_BBcm_Bi.get_moments().isZero() &&
-      I_BBcm_Bi.get_products().isZero()) {
-    return SpatialInertia<double>(
-        0.0, Vector3d::Zero(), UnitInertia<double>{0.0, 0.0, 0.0});
-  }
-  // B and Bi are not necessarily aligned.
-  const math::RotationMatrix<double> R_BBi(X_BBi.rotation());
-
-  // Re-express in frame B as needed.
-  const RotationalInertia<double> I_BBcm_B = I_BBcm_Bi.ReExpress(R_BBi);
-
-  try {
-    return SpatialInertia<double>::MakeFromCentralInertia(
-        body_mass, p_BoBcm_B, I_BBcm_B);
-  } catch (const std::exception& e) {
-    Warning(*node, e.what());
-    return Mdum_BBo_B;
-  }
-  DRAKE_UNREACHABLE();
+  return ParseSpatialInertia(diagnostic_.MakePolicyForNode(node),
+                             X_BBi, body_mass, inputs);
 }
 
 void UrdfParser::ParseBody(XMLElement* node, MaterialMap* materials) {
--- a/multibody/parsing/test/detail_common_test.cc
+++ b/multibody/parsing/test/detail_common_test.cc
@@ -3,6 +3,8 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "drake/common/test_utilities/diagnostic_policy_test_base.h"
+
 namespace drake {
 namespace multibody {
 namespace internal {
@@ -23,6 +25,7 @@ using geometry::internal::kHydroGroup;
 using geometry::internal::kMaterialGroup;
 using geometry::internal::kRezHint;
 using std::optional;
+using testing::MatchesRegex;
 
 GTEST_TEST(EndsWithCaseInsensitiveTest, BasicTests) {
   EXPECT_TRUE(EndsWithCaseInsensitive("something", "thing"));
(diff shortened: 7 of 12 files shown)


rpoyner-tri (commented):
See https://github.com/RobotLocomotion/drake/issues/18773 for discussion and context.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NTaKTo547ZTMuEU75zv:b-5kqrf4)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

rpoyner-tri (commented):
https://github.com/RobotLocomotion/drake/pull/19238 is the analogous patch for URDF.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NTaKe2s1-RSx9KXWYoe:bt05ux8)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NTnvch-4MfvA-qxQa2n:-NTnvch-4MfvA-qxQa2o:b-cdwpup) (no related file):*
This patch in `r1` has the same "dummy inertia" choice as the URDF patch -- there is ongoing discussion there. This patch should follow the resolution of that discussion.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NTnvch-4MfvA-qxQa2n:-NTpJt8l4ricv0artieH:b-uvse1z) (no related file):*

This patch in `r1` has the same "dummy inertia" choice as the URDF patch -- there is ongoing discussion there. This patch should follow the resolution of that discussion.
</blockquote></details>

Some considerations so far:

* It probably makes sense to carve out the portion of the code that has both the "magic" catch blocks, and the dummy inertia. Even at "rule of 2", it's magic enough code that sharing would be worth the trouble.
*  #19238 as it is so far likely needs some sanity testing on the dummy inertia.
* perhaps this PR can take on the task of separating out the inertia building, and the related unit test to a new shared location.

___

rpoyner-tri (commented):
+(status: do not review)  while the URDF patch percolates.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NTtDJ-y-sMLZBvCC1Pk:b-rck0co)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

rpoyner-tri (commented):
-(status: do not review) +(release notes: fix) 
+@amcastro-tri for feature review, based on his high number of surviving code lines in detail_sdf_parser.cc, and not being today's on-call platform reviewer.
FYI @mitiguy  -- shouldn't be much new math, only moved math.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NUSpLx666rVbXkjI1w3:b-7xvam0)* status: 2 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri)

amcastro-tri (commented):
at this point a few questions only

> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NUbj_E3CBMbzA_z8Z89:bmd8dww)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 218 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcNaZ6FPAoIOnv95w:-NUbcNaZ6FPAoIOnv95x:b-xbobfh) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L218)):*
> ```objc
> 
> // This helper struct serves to label and order the inertia inputs in the
> // (otherwise unmanageable) parameter list of ParseSpatialInertia.
> ```

minor, it's not unmanageable thanks to you! so I'd keep the docs to "Helper struct to store inputs for the components of a rotational inertia."
___
*[`multibody/parsing/detail_common.h` line 219 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbebe-5TfsbLeptkvl:-NUbebe-5TfsbLeptkvm:b-tj2hsi) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L219)):*
> ```objc
> // This helper struct serves to label and order the inertia inputs in the
> // (otherwise unmanageable) parameter list of ParseSpatialInertia.
> struct InertiaInputs {
> ```

nit, suggestoin, you could store `mass` here and make this an in-memory representation of a full spatial inertia, less arguments for `ParseSpatialInertia()`
___
*[`multibody/parsing/detail_common.h` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbelnH8Bg3RJ8JVW4a:-NUbelnH8Bg3RJ8JVW4b:byyc5h1) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L224)):*
> ```objc
>   double iyy{};
>   double izz{};
>   // Products of inertia.
> ```

could you state units here? sometimes people do include mass as in kg⋅m², and sometimes only m².
___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUbcqwJ9jlpEenTK2g7:b-77f2kv) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*
> ```objc
> // Combines the given user inputs into a SpatialInertia. When any data is
> // invalid, emits a warning into the diagnostic policy and returns a
> // best-effort approximation instead.
> ```

I haven't read the source yet, but why to return a "best-effort approximation"? that'd require to document whatev heuristics are required to accomplish that. 
As far as I understood from your PR description, invalid inertias are allowed so that users can do kinematic things like visualizations.
If the input inertia values are invalid and therefore not meant for sim and/or dynamics queries, then probably I'd emit a warning and return an invalid inertia full of NaNs. That way users get bit early, with a trail of NaNs easy to track, and hopefully they get forced to ready your warning.
___
*[`multibody/parsing/detail_common.h` line 234 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbdpkl9CzsH9HG_h-h:-NUbdpkl9CzsH9HG_h-i:b-ubonu) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L234)):*
> ```objc
> // best-effort approximation instead.
> // @param diagnostic  The error-reporting channel.
> // @param X_BBi       Pose of the inertia frame expressed in the body's frame.
> ```

you din't quite define `Bi`. When I read the inputs to this method my immediate thoughts are:
1. About what point are `inertia_inputs` computed? pressumably Bi's origin? 
2. In what frame is `inertia_inputs` expressed? presumably Bi?
3. I'd strongly suggest to use monogram notation for `inertia_inputs`! after all, is an in-memory representation of a rotational  inertia. So maybe `inertia_inputs_Bi_Bi`?
___
*[`multibody/parsing/detail_common.cc` line 257 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbfL4K1TznfyFvP_Hg:-NUbfL4K1TznfyFvP_Hh:b-nivzgz) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L257)):*
> ```cpp
>   // validity checking happen later.
>   const double plausible_dummy_mass =
>       (std::isfinite(mass) && mass > 0.0) ? mass : 1.0;
> ```

as suggested in the header, I'd make this NaN instead. After all, that'd mean the parsed file is not meant for dynamics computations.
___
*[`multibody/parsing/detail_common.cc` line 265 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbgf218iyg5So06xxD:-NUbgf218iyg5So06xxE:b-8glvnx) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L265)):*
> ```cpp
>   const double plausible_radius =
>       std::cbrt((3.0 / (4.0 * M_PI)) * plausible_volume);
>   // Create Mdum_BBo_B, a plausible spatial inertia for B about Bo (B's origin).
> ```

maybe a dumb question, what's `dum`?
___
*[`multibody/parsing/detail_common.cc` line 275 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbg4pl7fvZxdeC8aGN:-NUbg4pl7fvZxdeC8aGO:b-akzemq) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L275)):*
> ```cpp
>        kPlausibleDensity, plausible_radius);
>   // Bi's origin is at the COM as documented in
>   // http://wiki.ros.org/urdf/XML/link#Elements
> ```

IMO this should be part of your docs in the header. That is, define Bi as being a frame located at the body's CoM, not necessarily aligned with the principal axes of the body (since clearly the producst of inertia are allowed to be non-zero)
___
*[`multibody/parsing/detail_common.cc` line 279 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbhq6d34cFz7awV8wn:-NUbhq6d34cFz7awV8wo:b-7npwwv) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L279)):*
> ```cpp
>   const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
> 
>   // Yes, catching exceptions violates the coding standard. It is done here to
> ```

is there a TODO here?
___
*[`multibody/parsing/detail_common.cc` line 292 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbjQwo2Ox6vauRyC-X:-NUbjQwo2Ox6vauRyC-Y:b-ianfad) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L292)):*
> ```cpp
>   } catch (const std::exception& e) {
>     diagnostic.Warning(e.what());
>     return Mdum_BBo_B;
> ```

Here and below, I'm proposing to regtunr a NaN inertia, i.e. `return SpatialInertia<double>()`
___
*[`multibody/parsing/detail_common.cc` line 302 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbj7mpEsoX97XtykZW:-NUbj7mpEsoX97XtykZX:b-xy527h) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L302)):*
> ```cpp
>   }
>   // B and Bi are not necessarily aligned.
>   const math::RotationMatrix<double> R_BBi(X_BBi.rotation());
> ```

nit, you could probably get a reference instead of a copy?

amcastro-tri (commented):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @rpoyner-tri)

jwnimmer-tri:
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUbjxSR3huE5VUCJ6Jn:b-7hdhn3) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*

I haven't read the source yet, but why to return a "best-effort approximation"? that'd require to document whatev heuristics are required to accomplish that. 
As far as I understood from your PR description, invalid inertias are allowed so that users can do kinematic things like visualizations.
If the input inertia values are invalid and therefore not meant for sim and/or dynamics queries, then probably I'd emit a warning and return an invalid inertia full of NaNs. That way users get bit early, with a trail of NaNs easy to track, and hopefully they get forced to ready your warning.
</blockquote></details>

It's to our advantage to still allow dynamics and simulation from slightly-buggy model input files.  The simulation might not be great (and maybe will even crash), but making users type in mathematically perfect inertia values is too much of a burden to impose on someone just getting started with Drake.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUbziGd0z8NDa6aFZUj:bjkw8sr) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*

It's to our advantage to still allow dynamics and simulation from slightly-buggy model input files.  The simulation might not be great (and maybe will even crash), but making users type in mathematically perfect inertia values is too much of a burden to impose on someone just getting started with Drake.
</blockquote></details>

I agree, that's why we should compute inertia values automatically for them.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 12 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUc-Wkt5teW1I7nYKUg:bqdmhi3) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*

So the only problem here is that the PR overview incompletely stated the goals, so you can mark yourself resolved here?
</blockquote></details>

Specific example, to help clarify the goal: assume the user has `mug.sdf` file with a slightly-wrong inertia (i.e., one that we bark about here).  They load the mug model and posture it on a table.  The robot is off doing something else on a bookshelf nearby, not actually doing anything with the mug.  Should we refuse to run the bookshelf simulation because the idle mug on the table had a typo in it's inertia?  No, we should still let them run their simulation. We warned them that the mug inertia was not going to be respected. If they they want to ignore that warning for now, we shouldn't get in their way.  That means that using NAN for the mug mass and/or inertia would be a detriment to usability.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 219 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbebe-5TfsbLeptkvl:-NUc_2wb4veZhbJGFkS6:b-rzz4zd) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L219)):*

nit, suggestoin, you could store `mass` here and make this an in-memory representation of a full spatial inertia, less arguments for `ParseSpatialInertia()`
</blockquote></details>

I think adding mass here overinflates a very tiny abstraction. Why not add the offset vector? Why not add physical validity checking? This struct is really just a cheap way to safely transport the rotational inertia components into the parse method, by allowing the ordering to explicit at call sites.
___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUcaqZS6MPsd7C5GLq4:bthb8up) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*

Specific example, to help clarify the goal: assume the user has `mug.sdf` file with a slightly-wrong inertia (i.e., one that we bark about here).  They load the mug model and posture it on a table.  The robot is off doing something else on a bookshelf nearby, not actually doing anything with the mug.  Should we refuse to run the bookshelf simulation because the idle mug on the table had a typo in it's inertia?  No, we should still let them run their simulation. We warned them that the mug inertia was not going to be respected. If they they want to ignore that warning for now, we shouldn't get in their way.  That means that using NAN for the mug mass and/or inertia would be a detriment to usability.
</blockquote></details>

1. The long game is to provide a model-rewriting tool for URDF and SDF, since their published semantics don't support "ignore your silly inertias and compute new ones" the way mujoco's format does.
2.  In the short term, having a warning, and a plausible inertia probably does a better job of avoiding user discouragement than a bag of NaNs. A plausible dummy inertia is visualizable, NaNs are not.
___
*[`multibody/parsing/detail_common.cc` line 257 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbfL4K1TznfyFvP_Hg:-NUcZhb7CnEOhRqZrpfX:b-bocff6) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L257)):*

as suggested in the header, I'd make this NaN instead. After all, that'd mean the parsed file is not meant for dynamics computations.
</blockquote></details>

I disagree. See discussions elsewhere.
___
*[`multibody/parsing/detail_common.cc` line 265 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbgf218iyg5So06xxD:-NUcZLxKCUzk287Z3SsI:b-8cvae9) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L265)):*

maybe a dumb question, what's `dum`?
</blockquote></details>

dum` is the shortened form of "dummy" that @mitiguy recommended when revising #19238 for monogram notation.
___
*[`multibody/parsing/detail_common.cc` line 279 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbhq6d34cFz7awV8wn:-NUcYQOl3Y2iXxPM6J-q:b-ppnitg) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L279)):*

is there a TODO here?
</blockquote></details>

Not really. The TODO would be something like "undertake a thorough rewrite of all of the inertia classes, and their error handling". The alternatives were explored during development of https://github.com/RobotLocomotion/drake/pull/19238 . See for example [this slack thread](https://drakedevelopers.slack.com/archives/C2ETTRXT7/p1681929257081249).
___
*[`multibody/parsing/detail_common.cc` line 292 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbjQwo2Ox6vauRyC-X:-NUcZ6wy4ui7VQrIIFLG:baxftxu) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L292)):*

Here and below, I'm proposing to regtunr a NaN inertia, i.e. `return SpatialInertia<double>()`
</blockquote></details>

I disagree. How would that be any better than the old status quo of just exploding at parse time?

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 8 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbelnH8Bg3RJ8JVW4a:-NUcj2Ef2OJeGPXJ2xnj:b-evjyww) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L224)):*

could you state units here? sometimes people do include mass as in kg⋅m², and sometimes only m².
</blockquote></details>

Fun fact: neither URDF nor SDFormat docs give a clear statement of units for these quantities, although some URDF tutorial [hints](https://wiki.ros.org/urdf/Tutorials/Adding%20Physical%20and%20Collision%20Properties%20to%20a%20URDF%20Model#Inertia) at units. I've claimed here they are kg*m^2, since we are passing them directly to the rotational inertia factory method, and since (I presume) all physical quantities in Drake are in SI units.

slug-ft^2 anyone?
___
*[`multibody/parsing/detail_common.h` line 234 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbdpkl9CzsH9HG_h-h:-NUcvw0A-zFVm7myQGik:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L234)):*

you din't quite define `Bi`. When I read the inputs to this method my immediate thoughts are:
1. About what point are `inertia_inputs` computed? pressumably Bi's origin? 
2. In what frame is `inertia_inputs` expressed? presumably Bi?
3. I'd strongly suggest to use monogram notation for `inertia_inputs`! after all, is an in-memory representation of a rotational  inertia. So maybe `inertia_inputs_Bi_Bi`?
</blockquote></details>

Done.
___
*[`multibody/parsing/detail_common.cc` line 275 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbg4pl7fvZxdeC8aGN:-NUcvaHc1C-vm9qODMqa:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.cc#L275)):*

IMO this should be part of your docs in the header. That is, define Bi as being a frame located at the body's CoM, not necessarily aligned with the principal axes of the body (since clearly the producst of inertia are allowed to be non-zero)
</blockquote></details>

Done.

rpoyner-tri (commented):
@amcastro-tri are you ready to L G T M ?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NUwdJjxAXUkIINYxMBp:b-eegb1e)* status: LGTM missing from assignees amcastro-tri,ggould-tri(platform) (waiting on @amcastro-tri and @ggould-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: LGTM missing from assignees ggould-tri(platform),amcastro-tri (waiting on @amcastro-tri and @ggould-tri)

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUwi7o51Hvc053p2krt:btk6ob0) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*
> The con is that this allows them to shoot themselves in the foot since "almost ok" simulation results are extremely difficult to debug.

This is mitigated by the fact that we are printing a warning when this mistake happens.  I expect that for a lot of users, that will make it a lot easier to debug.

Once we have a tool that rewrites the SDFormat file to have a valid inertia, the warning should be updated to cite that tool (perhaps as part of a wider `troubleshooting.html` stanza).

> TODO... revert this to NaN's when the automatic computation of intertias from geometry lands.

No.  Never should the parser add NaNs to the MbP.  In the extreme case, we could calculate the moments from the collision geometry, but I don't even think we should do that.  Time will tell what works best, but adding NaNs to the plant is never okay in any regime.

ggould-tri (commented):
:lgtm:

> Reviewed 1 of 2 files at r1, 6 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NUwjfn-9QECVcxLHh1b:bnfp4nl)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri (waiting on @amcastro-tri)

___
*[`multibody/parsing/detail_common.h` line 222 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUwgrFN3-R2EF-NHJtF:-NUwgfnuCQz9FzJzroWk:bmrqhqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.h#L222)):*
> ```objc
> // These quantities should have been parsed from the URDF
> // `/robot/link/inertial/inertia` tag or from the SDFormat
> // `//model/link/inertial/inertia` tag.
> ```

FYI:  Thank you for getting the '//' right here! :-)
(and throughout)
___
*[`multibody/parsing/detail_common.cc` line 277 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUwhx6X4tF1rWWk_g9A:-NUwhx6X4tF1rWWk_g9B:br3nyzm) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.cc#L277)):*
> ```cpp
>   const SpatialInertia<double> Mdum_BBo_B = Mdum_BBcm.Shift(-p_BoBcm_B);
> 
>   // Yes, catching exceptions violates the coding standard. It is done here to
> ```

minor:
It seems like you've already weighed up the arguments here, but I cannot avoid commenting:
(1) Would it be better to provide a critique method in `RotationalInertia`?
(2) Since the primary coding standard concern is that methods cannot be known to be exception safe, should this document the exception safety of the called method?
(3) Would it be better to use a more specific exception type to avoid issuing more serious errors as warnings?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri (waiting on @amcastro-tri)

___
*[`multibody/parsing/detail_common.cc` line 277 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUwhx6X4tF1rWWk_g9A:-NUwkk907eF3r5f_MTw9:b7i8zf5) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.cc#L277)):*

minor:
It seems like you've already weighed up the arguments here, but I cannot avoid commenting:
(1) Would it be better to provide a critique method in `RotationalInertia`?
(2) Since the primary coding standard concern is that methods cannot be known to be exception safe, should this document the exception safety of the called method?
(3) Would it be better to use a more specific exception type to avoid issuing more serious errors as warnings?
</blockquote></details>

I'm somewhat to blame for the "just catch the exception" stance here.

I would support a version of (2).  The `SpatialInerta` classes (i.e., probably `UnitInertia` as well) should probably promise that they support the "Basic exception guarantee" in their class overview docs -- that any member function of those classes that throws meets that guarantee.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri (waiting on @amcastro-tri)

___
*[`multibody/parsing/detail_common.cc` line 277 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUwhx6X4tF1rWWk_g9A:-NUwzGxF-e7PJHV-4VN2:bd53a04) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.cc#L277)):*

To be clear, is it now my responsibility to document the "basic exception guarantee" for all the *Inertia classes, or just they safety of the usage here?
</blockquote></details>

I think either would serve the purposes of this file; if it can be guaranteed at all though I'd think it could be guaranteed in the classes rather than here?

amcastro-tri (commented):
sorry for the delay @rpoyner-tri , between travel, hospital and meetings I had a pretty hectic last few days. 
Feature review done. Thanks for the patience, :lgtm: 

> Reviewed 1 of 2 files at r1, 6 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-:-NUxEwHR7GrNCYh7u_-5:bgzypj9)* status: 5 unresolved discussions

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NUxAV8LBB7mRsgE30y-:b-o0mvgb) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*

> The con is that this allows them to shoot themselves in the foot since "almost ok" simulation results are extremely difficult to debug.

This is mitigated by the fact that we are printing a warning when this mistake happens.  I expect that for a lot of users, that will make it a lot easier to debug.

Once we have a tool that rewrites the SDFormat file to have a valid inertia, the warning should be updated to cite that tool (perhaps as part of a wider `troubleshooting.html` stanza).

> TODO... revert this to NaN's when the automatic computation of intertias from geometry lands.

No.  Never should the parser add NaNs to the MbP.  In the extreme case, we could calculate the moments from the collision geometry, but I don't even think we should do that.  Time will tell what works best, but adding NaNs to the plant is never okay in any regime.
</blockquote></details>

Not sure why we keep talking about a tool to rewrite SDFs (difficult and long term) when there is a much lower hanging fruit; inertias from geometry, always correct.
BTW, Per f2f, this is already resolved.
___
*[`multibody/parsing/detail_common.h` line 241 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxBmq24YQe3xo9S7k6:-NUxBmq24YQe3xo9S7k7:b-hduzyh) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.h#L241)):*
> ```objc
> // origin at the center of mass of the body, and axes arbitrary, but sometimes
> // chosen to zero out the products of inertia. Callers should have previously
> // constructed the transform `X_BBi` from the URDF `/robot/link/inertial/origin`
> ```

minor, 
In case it simplifies your docs, there is no need to refer to either URDF/SDF/MuJoCo or any other file format here. You simply need to define "your choice" of brame Bi, namely: "frame defined to be located a the body's CoM, not necessarily aligned with the principal axes of inertia"
___
*[`multibody/parsing/detail_common.h` line 247 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxCBW46AfYQEkkeRaK:-NUxCBW5DXdRZr_DB5vU:b-2ehisb) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_common.h#L247)):*
> ```objc
> // @param X_BBi       Pose of the inertia frame expressed in the body's frame.
> // @param mass        Mass of the body.
> // @param inputs_Bi_Bi The moments and products of inertia, measured and
> ```

nit, there is no "measured" for inertias, only "about a point". In this case it'd be "The moments and products of inertia about Bi's origin, expressed in frame Bi"
___
*[`multibody/parsing/detail_sdf_diagnostic.cc` line 90 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxDo-b64CKDwj3JdeS:-NUxDo-b64CKDwj3JdeT:bht5ghk) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_sdf_diagnostic.cc#L90)):*
> ```cpp
> bool IsError(const sdf::Error& report) {
>   switch (report.Code()) {
>     case sdf::ErrorCode::LINK_INERTIA_INVALID:
> ```

maybe a dumb question, being unfamiliar with this code, I don't quite see how this change ties to everything else in this PR. Is it tested/exercised
___
*[`multibody/parsing/test/detail_common_test.cc` line 371 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxEisuCpVswQsUv-H6:-NUxEisuCpVswQsUv-H7:bcrozrw) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/test/detail_common_test.cc#L371)):*
> ```cpp
> class ParseSpatialInertiaTest : public test::DiagnosticPolicyTestBase {};
> 
> // Confirms successful parsing of friction.
> ```

nit, copy/pasta comment

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 5 unresolved discussions

___
*[`multibody/parsing/detail_sdf_diagnostic.cc` line 90 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxDo-b64CKDwj3JdeS:-NUxRsVY5ILmQPJrY2Hf:b-xidxvp) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_sdf_diagnostic.cc#L90)):*

maybe a dumb question, being unfamiliar with this code, I don't quite see how this change ties to everything else in this PR. Is it tested/exercised
</blockquote></details>

This is necessary to avoid the status-quo of the underlying SDFormat library error getting thrown. Here is an example from current master code:

```
rico@torg:~/checkout/drake$ ROS_PACKAGE_PATH=$HOME/third/kinova-movo/movo_simulation/movo_gazebo bazel-bin/multibody/parsing/parser_manual_test package://movo_gazebo/models/test_zone/model.sdf
parsing package://movo_gazebo/models/test_zone/model.sdf
error: A link named link has invalid inertia.
```
No source file-and-line, no values, no explanation whatsoever.

With the current PR applied, instead you get this:
```
rico@torg:~/checkout/drake$ ROS_PACKAGE_PATH=$HOME/third/kinova-movo/movo_simulation/movo_gazebo bazel-bin/multibody/parsing/parser_manual_test -- package://movo_gazebo/models/test_zone/model.sdf
parsing package://movo_gazebo/models/test_zone/model.sdf
warning: A link named link has invalid inertia.
/home/rico/third/kinova-movo/movo_simulation/movo_gazebo/models/test_zone/model.sdf:4: warning: MakeFromMomentsAndProductsOfInertia(): The rotational inertia
[  0    0    0]
[  0    0    0]
[  0    0  0.1]
did not pass the test CouldBePhysicallyValid().
The associated principal moments of inertia:
0  0  0.1
do not satisfy the triangle inequality.
/home/rico/third/kinova-movo/movo_simulation/movo_gazebo/models/test_zone/model.sdf:4: error: URI 'model://test_zone/meshes/building.dae' refers to unknown package 'test_zone'
```
The old, useless, message is still there as a warning. It would take a bit more effort to be rid of it entirely. However, we also get the more informative Drake error message, with the relevant file and line number.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 5 unresolved discussions

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NV-u1OpCaWjuTcXK5Yz:bmueeq8) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*
> ... always correct.

No its not.  The collision solid is often an approximation of the true geometry  and anyway the body is usually not a uniform density.

> a tool to rewrite SDFs (difficult and long term)

The tool is neither difficult to write nor to use.  Having SDFormat files that can be successfully loaded into simulators besides Drake is important.  Adding magic new Drake-specific tags to SDFormat per your proposal is the worse option.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 5 unresolved discussions

___
*[`multibody/parsing/detail_common.h` line 232 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUbcqwJ9jlpEenTK2g6:-NV02007EKDtJ_J-OXha:bhl10m7) ([raw file](https://github.com/RobotLocomotion/drake/blob/96213f8755ba305e4476a1dd8a3dd4472882c213/multibody/parsing/detail_common.h#L232)):*
> The collision solid is often an approximation of the true geometry

True, but significantly better than eyeballing an arbitrary mass and dimensions. 

> The tool is neither difficult to write nor to use.

That's like music to my ears! can't wait to start using it myself.
___
*[`multibody/parsing/detail_sdf_diagnostic.cc` line 90 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxDo-b64CKDwj3JdeS:-NV02E_00BXZvliPKt_o:b4952lx) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_sdf_diagnostic.cc#L90)):*

This is necessary to avoid the status-quo of the underlying SDFormat library error getting thrown. Here is an example from current master code:

```
rico@torg:~/checkout/drake$ ROS_PACKAGE_PATH=$HOME/third/kinova-movo/movo_simulation/movo_gazebo bazel-bin/multibody/parsing/parser_manual_test package://movo_gazebo/models/test_zone/model.sdf
parsing package://movo_gazebo/models/test_zone/model.sdf
error: A link named link has invalid inertia.
```
No source file-and-line, no values, no explanation whatsoever.

With the current PR applied, instead you get this:
```
rico@torg:~/checkout/drake$ ROS_PACKAGE_PATH=$HOME/third/kinova-movo/movo_simulation/movo_gazebo bazel-bin/multibody/parsing/parser_manual_test -- package://movo_gazebo/models/test_zone/model.sdf
parsing package://movo_gazebo/models/test_zone/model.sdf
warning: A link named link has invalid inertia.
/home/rico/third/kinova-movo/movo_simulation/movo_gazebo/models/test_zone/model.sdf:4: warning: MakeFromMomentsAndProductsOfInertia(): The rotational inertia
[  0    0    0]
[  0    0    0]
[  0    0  0.1]
did not pass the test CouldBePhysicallyValid().
The associated principal moments of inertia:
0  0  0.1
do not satisfy the triangle inequality.
/home/rico/third/kinova-movo/movo_simulation/movo_gazebo/models/test_zone/model.sdf:4: error: URI 'model://test_zone/meshes/building.dae' refers to unknown package 'test_zone'
```
The old, useless, message is still there as a warning. It would take a bit more effort to be rid of it entirely. However, we also get the more informative Drake error message, with the relevant file and line number.
</blockquote></details>

ah! that's so cool! definitely a great improvement. Thanks for the explanation.

rpoyner-tri (commented):
> Dismissed @amcastro-tri from a discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19245)* status: 4 unresolved discussions

___
*[`multibody/parsing/detail_sdf_diagnostic.cc` line 90 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19245#-NUxDo-b64CKDwj3JdeS:-NV10pKI6jSb-1mm3U7P:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/44a70af2439579bcc8e8dd541533d329426f29f7/multibody/parsing/detail_sdf_diagnostic.cc#L90)):*

ah! that's so cool! definitely a great improvement. Thanks for the explanation.
</blockquote></details>

Done.

PR #19247, opened 2023-04-22
DiscreteDerivative sets suppress_initial_transient=true by default

RussTedrake:
Also adds missing documentation and fixes up the doxygen for the related DifferentialInverseKinematicIntegrator.

+@jwnimmer-tri for both reviews, please?

I think this is a worthwhile change. Almost everyone wants the suppression; anyone who calls `set_input_history` manually has this default effectively overwritten.  The cost of not using the suppression is high (huge initial transients), and the cost of having extra suppression is low.

Inspired by https://stackoverflow.com/q/75981586/9510020.

Does this need to be labeled a breaking change?  I could find our policy on this in the developer docs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19247)

### DIFF
--- a/bindings/pydrake/systems/primitives_py.cc
+++ b/bindings/pydrake/systems/primitives_py.cc
@@ -163,7 +163,7 @@ PYBIND11_MODULE(primitives, m) {
     DefineTemplateClassWithDefault<DiscreteDerivative<T>, LeafSystem<T>>(
         m, "DiscreteDerivative", GetPyParam<T>(), doc.DiscreteDerivative.doc)
         .def(py::init<int, double, bool>(), py::arg("num_inputs"),
-            py::arg("time_step"), py::arg("suppress_initial_transient") = false,
+            py::arg("time_step"), py::arg("suppress_initial_transient") = true,
             doc.DiscreteDerivative.ctor.doc)
         .def("time_step", &DiscreteDerivative<T>::time_step,
             doc.DiscreteDerivative.time_step.doc)
@@ -386,7 +386,7 @@ PYBIND11_MODULE(primitives, m) {
         Diagram<T>>(m, "StateInterpolatorWithDiscreteDerivative",
         GetPyParam<T>(), doc.StateInterpolatorWithDiscreteDerivative.doc)
         .def(py::init<int, double, bool>(), py::arg("num_positions"),
-            py::arg("time_step"), py::arg("suppress_initial_transient") = false,
+            py::arg("time_step"), py::arg("suppress_initial_transient") = true,
             doc.StateInterpolatorWithDiscreteDerivative.ctor.doc)
         .def("suppress_initial_transient",
             &StateInterpolatorWithDiscreteDerivative<
--- a/bindings/pydrake/systems/test/primitives_test.py
+++ b/bindings/pydrake/systems/test/primitives_test.py
@@ -716,18 +716,18 @@ class TestGeneral(unittest.TestCase):
         self.assertEqual(discrete_derivative.get_input_port(0).size(), 5)
         self.assertEqual(discrete_derivative.get_output_port(0).size(), 5)
         self.assertEqual(discrete_derivative.time_step(), 0.5)
-        self.assertFalse(discrete_derivative.suppress_initial_transient())
+        self.assertTrue(discrete_derivative.suppress_initial_transient())
 
         discrete_derivative = DiscreteDerivative(
-            num_inputs=5, time_step=0.5, suppress_initial_transient=True)
-        self.assertTrue(discrete_derivative.suppress_initial_transient())
+            num_inputs=5, time_step=0.5, suppress_initial_transient=False)
+        self.assertFalse(discrete_derivative.suppress_initial_transient())
 
     def test_state_interpolator_with_discrete_derivative(self):
         state_interpolator = StateInterpolatorWithDiscreteDerivative(
             num_positions=5, time_step=0.4)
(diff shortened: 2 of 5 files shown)


jwnimmer-tri (commented):
:lgtm: both +(status: single reviewer ok).

I confirmed that the new Doxygen styling looks good when rendered.

> Does this need to be labeled a breaking change? I could find our policy on this in the developer docs.

See https://drake.mit.edu/stable.html#behavioral-changes: _"Sometimes our software will choose a default value when nothing more specific is provided. Those defaults may change as we find better heuristics, without prior notice."_

Not breaking, so +(release notes: fix).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-:-NTdoDo397ZlhsOc2knj:bis3btm)* status: 3 unresolved discussions (waiting on @jwnimmer-tri and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdpxKz6TtKf6ZyMlTu:-NTdpxKz6TtKf6ZyMlTv:b-64g11i) (no related file):*
A unit test has become inconsistent now and need some updating:

```c++
// The dynamics must be affine when the transient suppression is disabled.
// This confirms that we've correctly conditioned the extra state for the
// update counter to be declared only when strictly required.
GTEST_TEST(DiscreteDerivativeTest, IsAffine) {
  const DiscreteDerivative<double> dut(2, 0.1);
```

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdrtA1C3Zfkq8Kk6t7:-NTdrtA1C3Zfkq8Kk6t8:bdytyzb) (no related file):*
Working

Testing vs Anzu CI.

___
___
*[`systems/primitives/discrete_derivative.h` line 61 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdr4D-DmltWDZG9zcS:-NTdr4D-DmltWDZG9zcT:bnolrth) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e44efcca0234b7e3d8fd366d06a61a01a735292/systems/primitives/discrete_derivative.h#L61)):*
> ```objc
>   /// suppress_initial_transient is true (the default), then the output will be
>   /// zero for the first two time steps (see the class documentation for
>   /// details).
> ```

This comment pretty authoritatively says "output will be zero" but that won't be true if the user later calls `set_input_history`.  That seems like an important caveat to mention.

I'd be satisfied with either a caveat inline here, or with the (already-cited) class overview docs explaining the caveat.  At the moment, neither of them do.

Ditto below for the same new comment on the wrapper system.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19247)* status: 2 unresolved discussions (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdrtA1C3Zfkq8Kk6t7:-NTe55qu1UtJPCznYaYo:b4bfv0h) (no related file):*

Working

Testing vs Anzu CI.
</blockquote></details>

OK all tests passed.

___

RussTedrake (commented):
Thank you.  I did look for that documentation, but didn't manage to find it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-:-NTetpM8E9ipjhamCkue:b-8975k0)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdpxKz6TtKf6ZyMlTu:-NTeuFmG5cmYS7MhIlzh:bkz5ckv) (no related file):*

A unit test has become inconsistent now and need some updating:

```c++
// The dynamics must be affine when the transient suppression is disabled.
// This confirms that we've correctly conditioned the extra state for the
// update counter to be declared only when strictly required.
GTEST_TEST(DiscreteDerivativeTest, IsAffine) {
  const DiscreteDerivative<double> dut(2, 0.1);
```
</blockquote></details>

Done.  Good catch.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdrtA1C3Zfkq8Kk6t7:-NTeuHDp8S4-WW30cOWr:b-9755fa) (no related file):*

OK all tests passed.
</blockquote></details>

Thank you.

___
___
*[`systems/primitives/discrete_derivative.h` line 61 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19247#-NTdr4D-DmltWDZG9zcS:-NTew3bp7fcog3fY3F35:b-7m0sn) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e44efcca0234b7e3d8fd366d06a61a01a735292/systems/primitives/discrete_derivative.h#L61)):*

This comment pretty authoritatively says "output will be zero" but that won't be true if the user later calls `set_input_history`.  That seems like an important caveat to mention.

I'd be satisfied with either a caveat inline here, or with the (already-cited) class overview docs explaining the caveat.  At the moment, neither of them do.

Ditto below for the same new comment on the wrapper system.
</blockquote></details>

Done.  I've tried to make it more clear.  PTAL.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19247)* status: :shipit: complete! all discussions resolved, LGTM from assignee jwnimmer-tri(platform) (waiting on @RussTedrake)

PR #19250, opened 2023-04-23
[workspace] Switch pybind11 to upstream

jwnimmer-tri:
This is a work-in-progress branch for the epic journey of getting rid of our Drake-specific pybind11 fork.

My plan is to carve out pieces as they become ripe.

Closes #5842.
Closes #13058.

Relates #7856.
Relates #12811.
Relates #14373.
Relates #14387.
Relates #14838.

Prerequisites:
- [x] #19251
- [x] #19369
- [x] #19570
- [x] #19571
- [x] #19572

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19250)

PR #14453, opened 2020-12-14
systems: Use const-ref arguments, not unique_ptr

jwnimmer-tri:
This commit adjusts FixInputPort and DeclareAbstractState to use pass-by-const-ref.

Part of #8904.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/14453)

PR #19251, opened 2023-04-23
[geometry] Add GeometryInstance overload without unique_ptr

jwnimmer-tri:
Towards #19250.

Broadly, a C++ function that takes a `unique_ptr<T>` as input without also offering a `const T&` overload is defective in that it forces users (even C++ users) to worry about lifetimes and leaves our implementations vulnerable to unwanted aliasing, when in most cases they don't care about actually care about ownership transfer. See, e.g., #14453 for prior art.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_common.cc
+++ b/bindings/pydrake/geometry/geometry_py_common.cc
@@ -86,8 +86,8 @@ void DoScalarIndependentDefinitions(py::module m) {
     constexpr auto& cls_doc = doc.GeometryInstance;
     py::class_<Class> cls(m, "GeometryInstance", cls_doc.doc);
     cls  // BR
-        .def(py::init<const math::RigidTransform<double>&,
-                 std::unique_ptr<Shape>, const std::string&>(),
+        .def(py::init<const math::RigidTransform<double>&, const Shape&,
+                 const std::string&>(),
             py::arg("X_PG"), py::arg("shape"), py::arg("name"),
             cls_doc.ctor.doc)
         .def("id", &Class::id, cls_doc.id.doc)
@@ -96,7 +96,6 @@ void DoScalarIndependentDefinitions(py::module m) {
             "set_pose", &Class::set_pose, py::arg("X_PG"), cls_doc.set_pose.doc)
         .def("shape", &Class::shape, py_rvp::reference_internal,
             cls_doc.shape.doc)
-        .def("release_shape", &Class::release_shape, cls_doc.release_shape.doc)
         .def("name", &Class::name, cls_doc.name.doc)
         .def("set_name", &Class::set_name, cls_doc.set_name.doc)
         .def("set_proximity_properties", &Class::set_proximity_properties,
@@ -121,6 +120,12 @@ void DoScalarIndependentDefinitions(py::module m) {
         .def("perception_properties", &Class::perception_properties,
             py_rvp::reference_internal, cls_doc.perception_properties.doc);
     DefCopyAndDeepCopy(&cls);
+    constexpr char doc_release_deprecated[] =
+        "Ownership transfer doesn't make sense for Python. Just use shape() "
+        "instead. This function will be removed on or after 2023-08-01.";
+    cls.def("release_shape",
+        WrapDeprecated(doc_release_deprecated, &Class::release_shape),
+        doc_release_deprecated);
   }
 
   // GeometryProperties
--- a/bindings/pydrake/geometry/test/common_test.py
+++ b/bindings/pydrake/geometry/test/common_test.py
@@ -69,7 +69,6 @@ class TestGeometryCore(unittest.TestCase):
         geometry.set_pose(RigidTransform([1, 0, 0]))
         self.assertIsInstance(geometry.pose(), RigidTransform)
         self.assertIsInstance(geometry.shape(), mut.Shape)
-        self.assertIsInstance(geometry.release_shape(), mut.Shape)
         self.assertEqual(geometry.name(), "sphere")
         geometry.set_name("funky")
         self.assertEqual(geometry.name(), "funky")
@@ -88,6 +87,8 @@ class TestGeometryCore(unittest.TestCase):
                               mut.PerceptionProperties)
         self.assertIsInstance(geometry.perception_properties(),
                               mut.PerceptionProperties)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(geometry.release_shape(), mut.Shape)
 
     def test_geometry_properties_api(self):
         # Test perception/ illustration properties (specifically Rgba).
--- a/geometry/geometry_instance.cc
+++ b/geometry/geometry_instance.cc
@@ -6,19 +6,25 @@ namespace drake {
 namespace geometry {
 
 GeometryInstance::GeometryInstance(const math::RigidTransform<double>& X_PG,
+                                   const Shape& shape, const std::string& name)
+    : GeometryInstance(X_PG, shape.Clone(), name) {}
+
+GeometryInstance::GeometryInstance(const math::RigidTransform<double>& X_PG,
                                    std::unique_ptr<Shape> shape,
                                    const std::string& name)
-    : id_(GeometryId::get_new_id()),
-      X_PG_(X_PG),
-      shape_(std::move(shape)) {
+    : id_(GeometryId::get_new_id()), X_PG_(X_PG) {
+  DRAKE_THROW_UNLESS(shape != nullptr);
+  shape_ = std::move(shape);
   set_name(name);
 }
 
 void GeometryInstance::set_name(const std::string& name) {
   name_ = internal::CanonicalizeStringName(name);
   if (name_.empty()) {
-    throw std::logic_error("GeometryInstance given the name '" + name +
-                           "' which is an empty canonical string");
+    throw std::logic_error(fmt::format(
+        "GeometryInstance given the name '{}' which is an empty canonical "
+        "string",
+        name));
   }
 }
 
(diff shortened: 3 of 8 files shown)


jwnimmer-tri:

SeanCurtis-TRI (commented):
:LGTM: With one quesiton.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-:-NTob3e74OigACAVvH5x:b-ib26yb)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NToaFf1-Y09PEzwY6sC:bn1hsm) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*
> ```objc
> 
>   /** (Advanced) Overload that transfers ownership of `shape` (for performance).
>    The caller must not retain any pointer to `shape`.
> ```

BTW This comment seems largely inspired by thinking about python users. It strikes me as being either redundant or misleading based on the function signature. Can you explain the motivation behind this additional comment?



_Code quote:_
```objc
The caller must not retain any pointer to `shape`.
```

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-:-NTobF7N1fKTrWWGImlj:bcav8r0)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri and @SeanCurtis-TRI)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NToebGM9-RuI05SPZPK:bfy6v5c) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*

BTW This comment seems largely inspired by thinking about python users. It strikes me as being either redundant or misleading based on the function signature. Can you explain the motivation behind this additional comment?
</blockquote></details>

The caller is transferring the lifetime of `shape` to be under the control of GeometryInstance (and in all probability, eventually the SceneGraph).  Those classes might mutate the Shape, or copy it, or delete it, or otherwise alter it in ways that the caller cannot anticipate, e.g., by calling `release_shape`. The idea of this comment is to dissuade callers (even C++ callers) from trying to keep a borrowed pointer around. We don't want to promise that the borrowed pointer will remain valid.

ggould-tri (commented):
:lgtm:

(I'd modify your manifesto statement above to limit it to cases where `T` is copyable/cloneable -- for non-copyable T it is a more complicated case.  Of course we know we're in the copyable case here (per the `copyable_unique_ptr`), so your argument strikes with full force.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-:-NTod4eO7UuWT4MvvjTD:bnj2312)* status: 1 unresolved discussion (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NTok4Jg2KPvrUFwmgKI:b-189ow) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*

The caller is transferring the lifetime of `shape` to be under the control of GeometryInstance (and in all probability, eventually the SceneGraph).  Those classes might mutate the Shape, or copy it, or delete it, or otherwise alter it in ways that the caller cannot anticipate, e.g., by calling `release_shape`. The idea of this comment is to dissuade callers (even C++ callers) from trying to keep a borrowed pointer around. We don't want to promise that the borrowed pointer will remain valid.
</blockquote></details>

I guess where I'm being obtuse is how is that not equally true of every API that takes a `unique_ptr` by value? That felt intrinsic to the concept of passing ownership.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NTokLgGC9cIGOtIzjQd:bbazry9) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*

I guess where I'm being obtuse is how is that not equally true of every API that takes a `unique_ptr` by value? That felt intrinsic to the concept of passing ownership.
</blockquote></details>

Ah.  So you're saying comment is redundant, not that it's wrong?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NTomtiFBINxBe6gurpc:b-z489sf) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*

Ah.  So you're saying comment is redundant, not that it's wrong?
</blockquote></details>

My original comment is that it could be "either redundant or misleading". I believe we can both see the arguable redundancy.

By "misleading" I was referencing the idea that passing ownership doesn't *ever* preclude the ability to hang on to a raw pointer. But that behavior has always been dangerous, because you *can't* know what happens to the object after passing ownership. But we've rarely warned people about whether or not the thing that gets passed is disassembled (and destroyed) or not.  In the case of the `Shape`, it isn't and the shape stays alive indefinitely  Certainly, if you store a pointer to data that some other code owns, there is *always* the danger of it mutating, unless you've been given a clear promise to the contrary.

So, you can *always* hang on to a raw pointer with just as much safety in this API as every other similar API.

(In the case of the `GeometryInstance` passed to `SceneGraph`, it *is* decomposed. But that's a different API.)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NTorFdL-n3GicyuHiAw:bsjyq1k) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*
> In the case of the Shape, it isn't and the shape stays alive indefinitely.

I want to leave room to change this implementation behavior down the road, e.g., immediately Clone the argument and discard the the original.  From the point of view of the caller, they should assume _no possibility_ of the shape pointer remaining valid upon return.  That's what I'm trying to communicate here.

If we think that's sufficiently obvious to a regular C++ users, we could nix the extra sentence.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

___
*[`geometry/geometry_instance.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19251#-NToaFf1-Y09PEzwY6sB:-NTot-cY2nUbXssRiMwa:beuhz18) ([raw file](https://github.com/RobotLocomotion/drake/blob/7fd9895775caf7f55f95e5f479961384b1211ae2/geometry/geometry_instance.h#L95)):*

> In the case of the Shape, it isn't and the shape stays alive indefinitely.

I want to leave room to change this implementation behavior down the road, e.g., immediately Clone the argument and discard the the original.  From the point of view of the caller, they should assume _no possibility_ of the shape pointer remaining valid upon return.  That's what I'm trying to communicate here.

If we think that's sufficiently obvious to a regular C++ users, we could nix the extra sentence.
</blockquote></details>

Yep. I think the idea of `unique_ptr` means as soon as you've passed it along, you have no guarantees. So, I'd be quite happy with removing the extra sentence.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19251)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform) (waiting on @jwnimmer-tri)

PR #12079, opened 2019-09-19
bindings: Change context.FixInputPort() to port.FixValue()

mpetersen94:
Towards #11341 

@sherm1 for feature review

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/12079)

PR #19255, opened 2023-04-24
[bindings] Use PointCloud object directly for GetInputPort.FixValue

### LINKED ISSUE #12086: point_cloud: Should permit SetFrom / operator= to update fields
See discussion item here:
https://github.com/RobotLocomotion/drake/pull/12079#issuecomment-533641237

Basically, `PointClolud`'s contract does not permit the following (in C++ or Python):
```
value = AbstractValue.Make(PointCloud())
# New fields introduced, not permitted by PointCloud.SetFrom.
value.SetFrom(PointCloud(new_size=0, fields=kXYZs | kRGBs))
```

This can be fixed by relaxing the contract for `PointCloud` to be less strict, and thus allowing the "structure" of the point cloud to be mutated and allow the model values to be more generic.

(As an alternative, perhaps we more seriously consider adopting `open3d`, if the dependencies are small enough?)

zachfang:
Close #12086

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19255)

### DIFF
--- a/bindings/pydrake/geometry/test/visualizers_test.py
+++ b/bindings/pydrake/geometry/test/visualizers_test.py
@@ -7,7 +7,6 @@ import numpy as np
 
 from drake import lcmt_viewer_load_robot, lcmt_viewer_draw
 from pydrake.autodiffutils import AutoDiffXd
-from pydrake.common.value import AbstractValue
 from pydrake.common.test_utilities import numpy_compare
 from pydrake.lcm import DrakeLcm, Subscriber
 from pydrake.math import RigidTransform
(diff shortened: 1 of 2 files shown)


zachfang (commented):
Following the conversations, I think it's the same issue already being addressed by #18987.

By running `git grep AbstractValue.Make`, these are the two call sites, I think, that need fixing up. Also, I assume the original `meshcat_visualizer_test.py` is repurposed or removed because I couldn't locate Eric's TODO in #12079.
```
# TODO(eric.cousineau): Replace `AbstractValue.Make(pc)` with just
# `pc` (#12086).
```

+@jwnimmer-tri for review. Thanks!

CC @EricCousineau-TRI

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19255#-:-NTkedX27UHo34jAxCFd:b-vmo9hr)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
Agreed.

:lgtm: both.

+(release notes: none) 
+(status: single reviewer ok) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19255#-:-NTnaO7a02Y1k9it8dxa:b-uvzmw1)* status: :shipit: complete! all discussions resolved, LGTM from assignee jwnimmer-tri(platform) (waiting on @zachfang)

PR #19271, opened 2023-04-25
[geometry/optimization] Optionally return all results from GCS

mpetersen94:
This allows the caller to see the relaxed solution as well as all of the rounded solutions from calling `SolveShortestPath`. If the user doesn't request all the results, this does not add any overhead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -499,20 +499,53 @@ void DefineGeometryOptimization(py::module m) {
                 py::arg("result") = std::nullopt, py::arg("show_slacks") = true,
                 py::arg("precision") = 3, py::arg("scientific") = false,
                 cls_doc.GetGraphvizString.doc)
-            .def("SolveShortestPath",
-                overload_cast_explicit<solvers::MathematicalProgramResult,
-                    GraphOfConvexSets::VertexId, GraphOfConvexSets::VertexId,
-                    const GraphOfConvexSetsOptions&>(
-                    &GraphOfConvexSets::SolveShortestPath),
+            .def(
+                "SolveShortestPath",
+                [](const GraphOfConvexSets& self,
+                    GraphOfConvexSets::VertexId source_id,
+                    GraphOfConvexSets::VertexId target_id,
+                    const GraphOfConvexSetsOptions& options) {
+                  return self.SolveShortestPath(source_id, target_id, options);
+                },
                 py::arg("source_id"), py::arg("target_id"),
                 py::arg("options") = GraphOfConvexSetsOptions(),
                 cls_doc.SolveShortestPath.doc_by_id)
-            .def("SolveShortestPath",
-                overload_cast_explicit<solvers::MathematicalProgramResult,
-                    const GraphOfConvexSets::Vertex&,
-                    const GraphOfConvexSets::Vertex&,
-                    const GraphOfConvexSetsOptions&>(
-                    &GraphOfConvexSets::SolveShortestPath),
+            .def(
+                "SolveShortestPath",
+                [](const GraphOfConvexSets& self,
+                    const GraphOfConvexSets::Vertex& source,
+                    const GraphOfConvexSets::Vertex& target,
+                    const GraphOfConvexSetsOptions& options) {
+                  return self.SolveShortestPath(source, target, options);
+                },
+                py::arg("source"), py::arg("target"),
+                py::arg("options") = GraphOfConvexSetsOptions(),
+                cls_doc.SolveShortestPath.doc_by_reference)
+            .def(
+                "SolveShortestPathAllResults",
+                [](const GraphOfConvexSets& self,
+                    GraphOfConvexSets::VertexId source_id,
+                    GraphOfConvexSets::VertexId target_id,
+                    const GraphOfConvexSetsOptions& options) {
+                  std::vector<solvers::MathematicalProgramResult> all_results;
+                  auto result = self.SolveShortestPath(
+                      source_id, target_id, options, &all_results);
+                  return py::make_tuple(result, all_results);
+                },
+                py::arg("source_id"), py::arg("target_id"),
+                py::arg("options") = GraphOfConvexSetsOptions(),
+                cls_doc.SolveShortestPath.doc_by_id)
+            .def(
+                "SolveShortestPathAllResults",
+                [](const GraphOfConvexSets& self,
+                    const GraphOfConvexSets::Vertex& source,
+                    const GraphOfConvexSets::Vertex& target,
+                    const GraphOfConvexSetsOptions& options) {
+                  std::vector<solvers::MathematicalProgramResult> all_results;
+                  auto result = self.SolveShortestPath(
+                      source, target, options, &all_results);
+                  return py::make_tuple(result, all_results);
+                },
                 py::arg("source"), py::arg("target"),
                 py::arg("options") = GraphOfConvexSetsOptions(),
                 cls_doc.SolveShortestPath.doc_by_reference);
--- a/geometry/optimization/graph_of_convex_sets.cc
+++ b/geometry/optimization/graph_of_convex_sets.cc
@@ -620,6 +620,99 @@ void GraphOfConvexSets::AddPerspectiveConstraint(
   }
 }
 
+void GraphOfConvexSets::FillResult(
+    std::map<VertexId, std::vector<Edge*>> incoming_edges,
+    std::map<VertexId, std::vector<Edge*>> outgoing_edges,
+    std::vector<Edge*> excluded_edges,
+    std::map<VertexId, MatrixXDecisionVariable> vertex_edge_ell,
+    std::map<EdgeId, Variable> relaxed_phi, std::vector<Variable> excluded_phi,
+    VertexId target_id, const MathematicalProgram& prog,
+    const GraphOfConvexSetsOptions& options,
+    MathematicalProgramResult* result) const {
+  int num_placeholder_vars = relaxed_phi.size();
+  for (const std::pair<const VertexId, std::unique_ptr<Vertex>>& vpair :
+       vertices_) {
+    num_placeholder_vars += vpair.second->ambient_dimension();
+    num_placeholder_vars += vpair.second->ell_.size();
+  }
+  for (const Edge* e : excluded_edges) {
+    num_placeholder_vars += e->y_.size() + e->z_.size() + e->ell_.size() + 1;
+  }
+  num_placeholder_vars += excluded_phi.size();
+  std::unordered_map<symbolic::Variable::Id, int> decision_variable_index =
+      prog.decision_variable_index();
+  int count = result->get_x_val().size();
+  Eigen::VectorXd x_val(count + num_placeholder_vars);
+  x_val.head(count) = result->get_x_val();
+  for (const Edge* e : excluded_edges) {
+    for (int i = 0; i < e->y_.size(); ++i) {
+      decision_variable_index.emplace(e->y_[i].get_id(), count);
+      x_val[count++] = 0;
+    }
+    for (int i = 0; i < e->z_.size(); ++i) {
+      decision_variable_index.emplace(e->z_[i].get_id(), count);
+      x_val[count++] = 0;
+    }
+    for (int i = 0; i < e->ell_.size(); ++i) {
+      decision_variable_index.emplace(e->ell_[i].get_id(), count);
+      x_val[count++] = 0;
+    }
+    decision_variable_index.emplace(e->phi_.get_id(), count);
+    x_val[count++] = 0;
+  }
+  for (const Variable& phi : excluded_phi) {
+    decision_variable_index.emplace(phi.get_id(), count);
+    x_val[count++] = 0;
+  }
+  for (const std::pair<const VertexId, std::unique_ptr<Vertex>>& vpair :
+       vertices_) {
+    const Vertex* v = vpair.second.get();
+    const bool is_target = (target_id == v->id());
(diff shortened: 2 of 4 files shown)


RussTedrake (commented):

quick questions at the API level before I carefully review the rest.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-:-NTq2JK2ChyA_f2KHGf1:bq4t5xa)* status: 2 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94 and @wrangelvid)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NTq2LqwA-KBImoLeY_T:b-dc21g3) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*
> ```objc
>       VertexId source_id, VertexId target_id,
>       const GraphOfConvexSetsOptions& options = GraphOfConvexSetsOptions(),
>       std::vector<solvers::MathematicalProgramResult>* all_results =
> ```

since you're returning one result already, why not call this all_rounded_results?  Then the actual return value will be the relaxed result, and this will be populated with the rounded results?
___
*[`geometry/optimization/graph_of_convex_sets.h` line 527 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2drr7qGSYu5kPvVQ:-NTq2drr7qGSYu5kPvVR:brxl8ww) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L527)):*
> ```objc
>       const solvers::VectorXDecisionVariable& vars) const;
> 
>   void FillResult(
> ```

btw -- i think we want to work towards having the rounding solve just the small version of the problem (not the big problem with constraints on phi).  Would that change the way you're factoring this out?

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 2 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94 and @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NTqGcVy0qhdiDe8ja1A:b-4z8zoz) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

since you're returning one result already, why not call this all_rounded_results?  Then the actual return value will be the relaxed result, and this will be populated with the rounded results?
</blockquote></details>

Returning the relaxed result might make the API confusing. In practice, we mostly use rounding and don't care about all results. So I would prefer it returning the best result. However, I see the problem with mixing relaxed and rounded results in the vector of `all_results`

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 1 unresolved discussion, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @wrangelvid)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 527 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2drr7qGSYu5kPvVQ:-NTsLSgO6czwSb8_15Gx:b-gcxpk1) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L527)):*

btw -- i think we want to work towards having the rounding solve just the small version of the problem (not the big problem with constraints on phi).  Would that change the way you're factoring this out?
</blockquote></details>

talked to @wrangelvid about this, and he thinks it's fine.

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 5 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94 and @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NTt-SOa4kVyRvUJafAA:bfablzv) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

Returning the relaxed result might make the API confusing. In practice, we mostly use rounding and don't care about all results. So I would prefer it returning the best result. However, I see the problem with mixing relaxed and rounded results in the vector of `all_results`
</blockquote></details>

I would be fine with the header clearly stating that the first element in all_results is the full problem followed by the rounded  problems.
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 632 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTt3Dv220NksPXo-VVR:-NTt3Dv220NksPXo-VVS:b-d4psc6) ([raw file](https://github.com/RobotLocomotion/drake/blob/d2a2f576be504d6fc2f8ee2ae8f94d001d66f049/geometry/optimization/graph_of_convex_sets.cc#L632)):*
> ```cpp
>     const GraphOfConvexSetsOptions& options,
>     MathematicalProgramResult* result) const {
>   int num_placeholder_vars = relaxed_phi.size();
> ```

More comments would help parsing the number of dense for loops. 
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1671 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTt3vM73_thNZAFPV-N:-NTt3vM73_thNZAFPV-O:b-7r4dty) ([raw file](https://github.com/RobotLocomotion/drake/blob/d2a2f576be504d6fc2f8ee2ae8f94d001d66f049/geometry/optimization/test/graph_of_convex_sets_test.cc#L1671)):*
> ```cpp
>   std::vector<MathematicalProgramResult> all_results;
>   MathematicalProgramResult rounded_result =
>       spp.SolveShortestPath(source->id(), target->id(), options, &all_results);
> ```

Since `max_rounded_paths` is set to 1, you could check if the length of all_results is what you expect (2). 
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1672 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTt4BB2EhaKbCWowYZV:-NTt4BB2EhaKbCWowYZW:b-c0whfl) ([raw file](https://github.com/RobotLocomotion/drake/blob/d2a2f576be504d6fc2f8ee2ae8f94d001d66f049/geometry/optimization/test/graph_of_convex_sets_test.cc#L1672)):*
> ```cpp
>   MathematicalProgramResult rounded_result =
>       spp.SolveShortestPath(source->id(), target->id(), options, &all_results);
>   MathematicalProgramResult relaxed_result = all_results[0];
> ```

Can you verify the first result is actually relaxed? Maybe check the phi values? 
___
*[`bindings/pydrake/geometry/geometry_py_optimization.cc` line 539 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTt54ahC57f7GSGFvSd:-NTt54ahC57f7GSGFvSe:b26tnl4) ([raw file](https://github.com/RobotLocomotion/drake/blob/d2a2f576be504d6fc2f8ee2ae8f94d001d66f049/bindings/pydrake/geometry/geometry_py_optimization.cc#L539)):*
> ```cpp
>                 cls_doc.SolveShortestPath.doc_by_id)
>             .def(
>                 "SolveShortestPathAllResults",
> ```

Add this call to the python test.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 6 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94 and @wrangelvid)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NU6avj0-vGwqUUZvsDb:busuhcj) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

I would be fine with the header clearly stating that the first element in all_results is the full problem followed by the rounded  problems.
</blockquote></details>

actually, rounding is off by default (not sure I love that), so people currently have to opt-in to using rounding.  the default behavior when we set convex_relaxation=True is to get the relaxed result as the primary output.

So I'm still in favor of having this be the rounded results.
___
*[`bindings/pydrake/geometry/geometry_py_optimization.cc` line 525 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NU6bLPM4dUrwt1ldVyg:-NU6bLPN1dEHkxpyqz0-:b-gjv2r6) ([raw file](https://github.com/RobotLocomotion/drake/blob/d2a2f576be504d6fc2f8ee2ae8f94d001d66f049/bindings/pydrake/geometry/geometry_py_optimization.cc#L525)):*
> ```cpp
>                 cls_doc.SolveShortestPath.doc_by_reference)
>             .def(
>                 "SolveShortestPathAllResults",
> ```

btw -- it's a little unfortunate that we have a name here that doesn't match the c++.  i can't immediately think of something better, because I agree that in python we want to had back those option outputs as the tuple.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 6 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94, @RussTedrake, and @wrangelvid)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NU8BseU770KwScL_EwY:bl9moxd) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

actually, rounding is off by default (not sure I love that), so people currently have to opt-in to using rounding.  the default behavior when we set convex_relaxation=True is to get the relaxed result as the primary output.

So I'm still in favor of having this be the rounded results.
</blockquote></details>

Russ asked me to weigh in here.

Perhaps your challenge is trying to cram too much meaning into a return value or argument name. 

What do you think about defining a tiny little struct that wraps `solvers::MathematicalProgramResult` alongside the meta-data you want about the answer e.g. whether it's relaxed, or rounded, or the random seed, or how many trials it took, etc.  The function return value and/or this output argument could communicate that wrapper struct instead of the raw MPResult directly.

You'll probably also be better served by defining a second function with a different name, rather than trying to use overloading.  That would help keep C++ and Python more uniform.  In that case you could use a list-typed return value, instead of an output argument.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19271)* status: 6 unresolved discussions, LGTM missing from assignees wrangelvid,RussTedrake(platform), missing label for release notes (waiting on @mpetersen94 and @wrangelvid)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NUPDZaq45QcSKmPLEkV:bvcg7lf) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

Russ asked me to weigh in here.

Perhaps your challenge is trying to cram too much meaning into a return value or argument name. 

What do you think about defining a tiny little struct that wraps `solvers::MathematicalProgramResult` alongside the meta-data you want about the answer e.g. whether it's relaxed, or rounded, or the random seed, or how many trials it took, etc.  The function return value and/or this output argument could communicate that wrapper struct instead of the raw MPResult directly.

You'll probably also be better served by defining a second function with a different name, rather than trying to use overloading.  That would help keep C++ and Python more uniform.  In that case you could use a list-typed return value, instead of an output argument.
</blockquote></details>

Thanks Jeremy.  @mpetersen94 -- are you able to move this forward?

mpetersen94:
*[`geometry/optimization/graph_of_convex_sets.h` line 486 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19271#-NTq2LqwA-KBImoLeY_S:-NUUo0MI0SZuDm926D9e:bqyorth) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1b7d81cf2a4852b147544cac443382d46ae7a7b/geometry/optimization/graph_of_convex_sets.h#L486)):*

Thanks Jeremy.  @mpetersen94 -- are you able to move this forward?
</blockquote></details>

Yes. I'll take a stab at making these structural changes over the next couple of days.

RussTedrake:
I am closing this in favor of my preferred approach (to break the rounding out of the main solve method into its own method).

PR #19279, opened 2023-04-26
ConvexSet can report when its just a point

jwnimmer-tri:
This allows algorithms (#19239) to customize themselves for degenerate sets, for performance.

+@wrangelvid +@RussTedrake what do you think of this proposal?  I think we could use this in `EdgesBetweenSubgraphs::RegionsConnectThroughSubspace` instead of the `dynamic_cast<Point*>`.

I only ported a few sets to the new function.  I suspect that some of the other types of sets (like a cartesian product of points) can also be degenerate.  We could iron out those details if the proposal seems useful.

Requires #19311.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -46,6 +46,8 @@ void DefineGeometryOptimization(py::module m) {
         .def("IntersectsWith", &ConvexSet::IntersectsWith, py::arg("other"),
             cls_doc.IntersectsWith.doc)
         .def("IsBounded", &ConvexSet::IsBounded, cls_doc.IsBounded.doc)
+        .def("MaybeGetPoint", &ConvexSet::MaybeGetPoint,
+            cls_doc.MaybeGetPoint.doc)
         .def("PointInSet", &ConvexSet::PointInSet, py::arg("x"),
             py::arg("tol") = 1e-8, cls_doc.PointInSet.doc)
         .def("AddPointInSetConstraints", &ConvexSet::AddPointInSetConstraints,
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -45,6 +45,7 @@ class TestGeometryOptimization(unittest.TestCase):
         point = mut.Point(p)
         self.assertEqual(point.ambient_dimension(), 3)
         np.testing.assert_array_equal(point.x(), p)
+        np.testing.assert_array_equal(point.MaybeGetPoint(), p)
         point.set_x(x=2*p)
         np.testing.assert_array_equal(point.x(), 2*p)
         point.set_x(x=p)
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -211,6 +211,7 @@ drake_cc_googletest(
         ":convex_set",
         ":test_utilities",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -52,13 +52,18 @@ CartesianProduct::CartesianProduct(const ConvexSet& setA, const ConvexSet& setB)
 CartesianProduct::CartesianProduct(const ConvexSets& sets,
                                    const Eigen::Ref<const MatrixXd>& A,
                                    const Eigen::Ref<const VectorXd>& b)
-    : ConvexSet(A.cols()), sets_(sets), A_(A), b_(b) {
+    : ConvexSet(A.cols()),
+      sets_(sets),
+      A_(A),
+      b_(b),
+      A_decomp_(Eigen::ColPivHouseholderQR<Eigen::MatrixXd>(*A_)) {
   const int y_ambient_dimension = SumAmbientDimensions(sets);
   const int x_ambient_dimension = ambient_dimension();
   DRAKE_THROW_UNLESS(A_->rows() == y_ambient_dimension);
   DRAKE_THROW_UNLESS(b_->rows() == y_ambient_dimension);
   DRAKE_THROW_UNLESS(A_->cols() == x_ambient_dimension);
-  DRAKE_THROW_UNLESS(A_->colPivHouseholderQr().rank() == A_->cols());
+  // Ensure that A is injective.
+  DRAKE_THROW_UNLESS(A_decomp_->rank() == A_->cols());
 }
 
 CartesianProduct::CartesianProduct(const QueryObject<double>& query_object,
@@ -84,6 +89,8 @@ CartesianProduct::CartesianProduct(const QueryObject<double>& query_object,
 
   A_ = X_GF.rotation().matrix();
   b_ = X_GF.translation();
+  // N.B. We leave A_decomp_ unset here. It's only used by MaybeGetPoint and
+  // there it's irrelevant because a cylinder is never a point anyway.
 }
 
 CartesianProduct::~CartesianProduct() = default;
@@ -105,6 +112,43 @@ bool CartesianProduct::DoIsBounded() const {
     }
   }
   return true;
+}
+
+std::optional<VectorXd> CartesianProduct::DoMaybeGetPoint() const {
+  // Check if all sets are points.
+  std::vector<VectorXd> points;
+  for (const auto& s : sets_) {
+    if (std::optional<VectorXd> point = s->MaybeGetPoint()) {
+      points.push_back(std::move(*point));
+    } else {
+      return std::nullopt;
+    }
+  }
+
+  // Stack the points.
+  const int y_ambient_dimension = A_ ? A_->rows() : ambient_dimension();
+  int start = 0;
+  VectorXd y(y_ambient_dimension);
+  for (const VectorXd& point : points) {
+    const int point_size = point.size();
+    y.segment(start, point_size) = point;
+    start += point_size;
+  }
+  DRAKE_DEMAND(start == y.size());
+
+  // When A and b are NOT in use, x is just y.
+  if (!A_.has_value()) {
+    return y;
+  }
+
+  // When A_ is in use, either A was passed to the constructor or we denote a
+  // cylinder. In the former case, we have the QR decomp already (set in the
+  // constructor). In the latter case, we can't possibly have cleared the "all
+  // sets are points" checks earlier in this function.
+  DRAKE_DEMAND(A_decomp_.has_value());
+
+  // Solve for x.
+  return A_decomp_->solve(y - *b_);
 }
 
 bool CartesianProduct::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
--- a/geometry/optimization/cartesian_product.h
+++ b/geometry/optimization/cartesian_product.h
@@ -37,7 +37,8 @@ class CartesianProduct final : public ConvexSet {
   CartesianProduct(const ConvexSet& setA, const ConvexSet& setB);
 
   /** Constructs the product of convex sets in the transformed coordinates:
-  {x | y = Ax + b, y ∈ Y₁ × Y₂ × ⋯ × Yₙ}. `A` must be full column rank. */
+  {x | y = Ax + b, y ∈ Y₁ × Y₂ × ⋯ × Yₙ}.
+  @throws std::exception when `A` is not full column rank. */
   CartesianProduct(const ConvexSets& sets,
                    const Eigen::Ref<const Eigen::MatrixXd>& A,
                    const Eigen::Ref<const Eigen::VectorXd>& b);
@@ -76,6 +77,8 @@ class CartesianProduct final : public ConvexSet {
 
   bool DoIsBounded() const final;
 
+  std::optional<Eigen::VectorXd> DoMaybeGetPoint() const final;
+
   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
                     double tol) const final;
 
@@ -113,6 +116,12 @@ class CartesianProduct final : public ConvexSet {
   // in the implementation.
   std::optional<Eigen::MatrixXd> A_{std::nullopt};
   std::optional<Eigen::VectorXd> b_{std::nullopt};
+
+  // When an `A` is passed to the constructor, we'll compute its decomposition
+  // and store it here for later use. Note that even though the constructor for
+  // a scene graph cylinder sets A_, it does not set A_decomp_.
+  std::optional<Eigen::ColPivHouseholderQR<Eigen::MatrixXd>> A_decomp_{
+      std::nullopt};
 };
 
 }  // namespace optimization
--- a/geometry/optimization/convex_set.cc
+++ b/geometry/optimization/convex_set.cc
@@ -31,6 +31,10 @@ bool ConvexSet::IntersectsWith(const ConvexSet& other) const {
   return result.is_success();
 }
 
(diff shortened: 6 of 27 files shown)


wrangelvid (commented):
I like it since other convex sets don't have to be of instance Point to degenerate. For instance, the nonnegative orthant in R^2 with another hyperplane going thru the origin, but pointing away, would degenerate to a point.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-:-NTvFW319JjClcv6w0n8:b-z3ec7q)* status: LGTM missing from assignees wrangelvid,RussTedrake(platform) (waiting on @RussTedrake)

RussTedrake (commented):
For me it's slightly questionable as a general (beautiful) method that belongs at the level of ConvexSet.  (for instance, if we started having MaybeGetHPolyhedron, that seems quite wrong).  But since it seems to be useful in the application, i'm ok with it.  

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-:-NU1_GYC36ERq16HQpeZ:b-w370gl)* status: LGTM missing from assignee wrangelvid (waiting on @wrangelvid)

jwnimmer-tri:
Okay, I'll work on cleaning this up into mergeable shape.

I'll also underline that the new function isn't asking about the runtime type, it's asking whether the set has degenerated into a single point. From my survey so far, all kinds of sets (other than the Hyperellipsoid (?)) can become degenerate in this way.

It would be similarly reasonable to have a base class function to "convert yourself to an hpolyhedron, if you're able to be represented as such" which could be satisfied by some (but not all) sets.

In the absence of limiting `class ConvexSet` to have a closed set of subclasses and a visitor pattern, representational operators like this need to go onto the base class as member functions.

jwnimmer-tri:
The other way to think about the new function... say that a user created their own point-like subclass like this:

```c++
class F32Point final : public ConvexSet {
  explicit F32Point(const VectorX<float>& x);
  const VectorX<float>& x() const { return x_; }
  ...
```

We would want our algorithms to treat that like the built-in `Point` class when optimizing.

Using an explicit `dynamic_cast` to call a derived class function `Point::x()` gets in the way of truly generic programming.

jwnimmer-tri:
*[`geometry/optimization/convex_set.h` line 92 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NU8XokC4KOcxdmlkMi9:-NU8XokC4KOcxdmlkMiA:b-j6ylgg) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf9d09e49de6f3710bf50626211f2e28ca5977be/geometry/optimization/convex_set.h#L92)):*
> ```objc
> 
>   /** If this set contains exactly one point, returns the value of that point.
>    * Otherwise, returns nullopt. */
> ```

If the ambient_dimension is zero, should this return None or an empty array?



_Code quote:_
```objc
  /** If this set contains exactly one point, returns the value of that point.
   * Otherwise, returns nullopt. */
```

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: 1 unresolved discussion, LGTM missing from assignee wrangelvid (waiting on @wrangelvid)

___
*[`geometry/optimization/convex_set.h` line 92 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NU8XokC4KOcxdmlkMi9:-NU9Wzj7C_8uvfcpiAqR:b-eeq4ea) ([raw file](https://github.com/RobotLocomotion/drake/blob/cf9d09e49de6f3710bf50626211f2e28ca5977be/geometry/optimization/convex_set.h#L92)):*

If the ambient_dimension is zero, should this return None or an empty array?
</blockquote></details>

I guess if the ambient space is ∅, then the cardinality of (any set in) that space is zero (not 1).  So I'd say nullopt.

jwnimmer-tri:
BTW this is on hold until #19311 lands.  It's not possible to write a correct `MaybeGetPoint` implementation until the other bugs are fixed first.

jwnimmer-tri (commented):
Okay, all of the prep work is done for this.  I'll work on adding unit tests next.

> Reviewed 16 of 16 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-:-NXMnGprAXQvJPxDwKjF:b-k9awc)* status: needs platform reviewer assigned, needs at least two assigned reviewers

jwnimmer-tri:
This is ready for feature review now.  +@RussTedrake or +@wrangelvid whoever has time, please go ahead.  (You can unassign the other reviewer, we probably only need one.)

jwnimmer-tri (commented):
> Reviewed 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: LGTM missing from assignee wrangelvid

RussTedrake (commented):
-@wrangelvid 

:lgtm: apart from one issue flagged below.  thanks!

> Reviewed 15 of 16 files at r4, 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-:-NXPhXhiBP48L55tKFgc:b-f6nsjj)* status: 3 unresolved discussions, needs at least two assigned reviewers

___
*[`geometry/optimization/cartesian_product.cc` line 134 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPiUyd5WhPoPJCmJIP:-NXPiUyd5WhPoPJCmJIQ:b3lpldd) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/cartesian_product.cc#L134)):*
> ```cpp
>   if (A_.has_value()) {
>     DRAKE_DEMAND(b_.has_value());
>     return *A_ * x + *b_;
> ```

The set is defined in x, not in y.
When A and b are defined, the value you build up in "stack the points" is a vector y, not x.  The set would only be a point if A is invertible, and you should return (inv(A)(y - b)).  
___
*[`geometry/optimization/intersection.cc` line 43 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPjMMpEYYhkFr7nu7Z:-NXPjMMpEYYhkFr7nu7_:b9ccq96) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/intersection.cc#L43)):*
> ```cpp
>   // of the other sets_, so that AddPointInSetConstraints and related could
>   // do a lot less work. It's not clear yet whether that optimization would
>   // be a good trade-off.
> ```

I don't think this comment is worth the clutter.  It's true that for many of the computations we do on convex sets, we can trade-off between precomputing vs on-demand computation.  I don't think DoMaybeGetPoint deserves special status in that regard.
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 152 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPkHtTAyMI3hrimkUm:-NXPkHtTAyMI3hrimkUn:bjgwnkt) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/test/cartesian_product_test.cc#L152)):*
> ```cpp
>   CartesianProduct S(H1, H2);
>   EXPECT_TRUE(S.IsBounded());
>   EXPECT_FALSE(S.MaybeGetPoint().has_value());
> ```

better add a test for the y = Ax + b code path.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: 3 unresolved discussions, needs at least two assigned reviewers

___
*[`geometry/optimization/cartesian_product.cc` line 134 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPiUyd5WhPoPJCmJIP:-NXQGkRkE-oardA5B9F3:b-8sykde) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/cartesian_product.cc#L134)):*

Technically, I guess that as long as A is injective, and y is in the range of Ax + b, then we could return the point.
</blockquote></details>

For now, I've supposed that inverting the matrix isn't worth the effort.  WDYT?
___
*[`geometry/optimization/intersection.cc` line 43 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPjMMpEYYhkFr7nu7Z:-NXQGjkuDnmyUGAPzsq4:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/intersection.cc#L43)):*

I don't think this comment is worth the clutter.  It's true that for many of the computations we do on convex sets, we can trade-off between precomputing vs on-demand computation.  I don't think DoMaybeGetPoint deserves special status in that regard.
</blockquote></details>

Done.
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 152 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPkHtTAyMI3hrimkUm:-NXQGih79A1mvs3kGikw:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/test/cartesian_product_test.cc#L152)):*

better add a test for the y = Ax + b code path.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> Reviewed 2 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cartesian_product.cc` line 134 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPiUyd5WhPoPJCmJIP:-NXS1YknBrA0hqWvA172:b-kyagzr) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/cartesian_product.cc#L134)):*

I think it would be nicer to handle it (the documentation of the MaybeGetPoint method is a bit non-committal on this point, but would probably lead me to expect success).  I do think your solution is better than doing the inverse case without properly handling the injective case.  

Maybe you can at least amend your new comment to add the injective comment (so I don't forget it if we go back and implement this later)?
</blockquote></details>

I'm in no particular hurry to land this PR, so I think we can take some time to fix it now.

I've pushed a first attempt, and I'm curious to get your feedback.

I think we probably want to add a test case with a non-injective A, but I haven't tried to make that yet.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cartesian_product.cc` line 134 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPiUyd5WhPoPJCmJIP:-NXVPtjE5UvrQihi0qia:b9bkkpe) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/cartesian_product.cc#L134)):*

I'm in no particular hurry to land this PR, so I think we can take some time to fix it now.

I've pushed a first attempt, and I'm curious to get your feedback.

I think we probably want to add a test case with a non-injective A, but I haven't tried to make that yet.
</blockquote></details>

Looks great.
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 198 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXVOPXg3hDLKmWQYVWV:-NXVOPXg3hDLKmWQYVWW:b-nlyeza) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b34f5857b49bbecb25f29f739c06612e8656277/geometry/optimization/test/cartesian_product_test.cc#L198)):*
> ```cpp
>   EXPECT_TRUE(S.PointInSet(in, kTol));
>   ASSERT_TRUE(S.MaybeGetPoint().has_value());
>   EXPECT_TRUE(CompareMatrices(S.MaybeGetPoint().value(), in, kTol));
> ```

to test the injective path, how about
```
A = [[ 1, 0 ], [0, 1], [0, 0]]
b = [ arbitrary; arbitrary; 0]
y = [ arbitrary; arbitrary; 0]
```

e.g. in this case the cartesian product of, for example, a single 3D point (which is itself) would also be a point in 2D.

RussTedrake (commented):
:lgtm: (again)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-:-NXW3_XY-ukamrX7g0j9:b-ptz332)* status: needs at least two assigned reviewers

___
*[`geometry/optimization/cartesian_product.cc` line 134 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXPiUyd5WhPoPJCmJIP:-NXW396R28mtM2TD8A2q:b-rmmu1e) ([raw file](https://github.com/RobotLocomotion/drake/blob/fcdfe11b1ee4cee7d93f5377e26d76b4623fd820/geometry/optimization/cartesian_product.cc#L134)):*

> I think we probably want to add a test case with a non-injective A, but I haven't tried to make that yet.

Upon further study, I think A is always injective.  We've already asserted in our constructor that is has full column rank.  Therefore, I've removed the `isInjective` check here in `MaybeGetPoint`.
</blockquote></details>

Oh.  Nice.  Should the doc for the constructor be updated to say it throws if A is not full column rank? (instead of just "A must be full column rank.")

jwnimmer-tri (commented):
> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: needs at least two assigned reviewers

jwnimmer-tri:

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: 1 unresolved discussion

___
*[`geometry/optimization/test/cartesian_product_test.cc` line 229 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXkRnjd38GOGuMGCEyK:-NXkaBJQ0Mhm_QY8jIlM:bvxetyj) ([raw file](https://github.com/RobotLocomotion/drake/blob/e36c8e06559ca7c5ca34ea35ca0bc7c902d1ec21/geometry/optimization/test/cartesian_product_test.cc#L229)):*

nit, it appears that a test on `MaybeGetPoint` on a `CartesianProduct` constructed from a cylinder is missing.
</blockquote></details>

See if I added what you had in mind.

xuchenhan-tri (commented):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19279)* status: :shipit: complete! all discussions resolved, LGTM from assignees RussTedrake(platform),xuchenhan-tri(platform)

___
*[`geometry/optimization/test/cartesian_product_test.cc` line 229 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19279#-NXkRnjd38GOGuMGCEyK:-NXkbNvfE-J7iPmdjDtA:b-k614gu) ([raw file](https://github.com/RobotLocomotion/drake/blob/e36c8e06559ca7c5ca34ea35ca0bc7c902d1ec21/geometry/optimization/test/cartesian_product_test.cc#L229)):*

See if I added what you had in mind.
</blockquote></details>

Yes, thanks!

Issue #6138, opened 2017-05-18
Robust enforcement to use the lower triangle in RotationalInertia


amcastro-tri:
As discussed in this [Reviewable comment](https://reviewable.io/reviews/robotlocomotion/drake/5945#-Kjk5x7d8BPlefOZzSIo), the eforcement to use the lower triangular part of the Eigen member in `RotationalInertia` is brittle and there an oportunity for improvement. 

@jwnimmer-tri proposed that we could enforce this all throughout the class in a robust way by placing the Eigen matrix member within a struct with private access. Then that struct would only allow to perform valid operations, like so:
 ```c++
class RotationalInertia {
  public:
    // RotationalInertia methods ...
  private:
   // Internal representation of the rotational inertia as a symmetric matrix.
   // Only access that doesn't break this symmetry is allowed.
   struct SymmetricMatrixRepresentaion {
     /// Access symmetric matrix elements with static assertion.
     template <int i, int j>
     const T& get_value() const {
       static_assert(j <=  i, "Only the lower triangle must be used.");
       return matrix3x3_(i, j);
     }

     /// When indexes cannot be resolved at compile time (as in IsNaN())
     /// you can use this method.
     const T& get_value(int i, int j) const {
       check_and_swap(&i, &j);
       DRAKE_ASSERT(j <=  i);
       return matrix3x3_(i, j);
     }

     // Returns a const Eigen view expression to the symmetric part of the matrix
     // in use by this RotationalInertia.
     const Eigen::SelfAdjointView<const Matrix3<T>, Eigen::Lower>
     get_symmetric_matrix_view() const {
       return matrix3x3_.template selfadjointView<Eigen::Lower>();
     }
     
     // Returns a mutable Eigen view expression to the symmetric part of the
     // matrix in use by RotationalInertia.
     // Note: operator=() is not defined for Eigen::SelfAdjointView and therefore
     // we need to return a TriangularView here.
     Eigen::TriangularView<Matrix3<T>, Eigen::Lower>
     get_mutable_triangular_view() {
       return matrix3x3_.template triangularView<Eigen::Lower>();
      }
    private:
     Matrix3<T> matrix3x3_{Matrix3<T>::Constant(std::numeric_limits<
         typename Eigen::NumTraits<T>::Literal>::quiet_NaN())};
   } I_SP_E_;
};
```

[...]

PR #19281, opened 2023-04-26
Calculate principal moments of inertia and their axes.

mitiguy:


mitiguy (commented):
Feature review +@sherm1 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NUC_rWd3gzhrL1rvLmH:bwuh3l8)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

sherm1 (commented):
Feature pass done. I'm not convinced this is correct, ptal.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NUN0sUj7q4R93Ci2MYL:bd6fqou)* status: 10 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`multibody/tree/rotational_inertia.h` line 541 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMu0yrCnKQcMcxLIr-:-NUMu0yrCnKQcMcxLIr0:bn9wh61) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L541)):*
> ```objc
> 
>   /// Forms the 3 principal moments of inertia and their 3 associated principal
>   /// directions for `this` unit inertia about-point P, expressed-in a frame E.
> ```

minor: why "unit inertia" here -- shouldn't this be rotational inertia?
___
*[`multibody/tree/rotational_inertia.h` line 542 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMunE_DPuGhcQ_CsEj:-NUMunE_DPuGhcQ_CsEk:bxixs3j) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L542)):*
> ```objc
>   /// Forms the 3 principal moments of inertia and their 3 associated principal
>   /// directions for `this` unit inertia about-point P, expressed-in a frame E.
>   /// @param[out] R_EP 3x3 right-handed orthonormal matrix which happens to be
> ```

minor: this appears to be an optional return but the documentation doesn't explain that
___
*[`multibody/tree/rotational_inertia.h` line 543 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMtdI_14LEMFWqMGEB:-NUMtdI_14LEMFWqMGEC:bz3d3ku) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L543)):*
> ```objc
>   /// directions for `this` unit inertia about-point P, expressed-in a frame E.
>   /// @param[out] R_EP 3x3 right-handed orthonormal matrix which happens to be
>   /// the rotation matrix relating the expressed-in frame E to the frame P. The
> ```

minor: frame P is undefined at this point (you mentioned point P above but it's not clear what the axes are)
___
*[`multibody/tree/rotational_inertia.h` line 549 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMsmfDAouBk76vDZIm:-NUMsmfDAouBk76vDZIn:b-lp8p1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L549)):*
> ```objc
>   /// the returned 3 principal moments of inertia.
>   /// @returns The 3 principal moments of inertia [Ixx Iyy Izz] sorted in
>   /// ascending order.
> ```

BTW it took me a while to decipher this sentence. Is there a better way to say it? I think you're saying we calculate the principal moments in ascending order, then call the smallest one Ixx and the largest Izz. As written it isn't clear that Ixx etc are _defined_ by the ordering, rather than _dictating_ the ordering.
___
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NUMvYka5KDJ68jShRQ5:bsew6tq) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*
> ```cpp
>   static_assert(is_lower_triangular_order(1, 0), "Invalid indices");
>   static_assert(is_lower_triangular_order(2, 0), "Invalid indices");
>   static_assert(is_lower_triangular_order(2, 1), "Invalid indices");
> ```

minor: these static_asserts don't make sense with constant arguments. Of course those are lower triangular! These should be removed.
___
*[`multibody/tree/rotational_inertia.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMzvlo6t0eRhmssra8:-NUMzvlo6t0eRhmssra9:bqn40q1) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L38)):*
> ```cpp
>     // Sort the eigenvalues in ascending order.
>     Vector3<double> moments(I_double(0, 0), I_double(1, 1), I_double(2, 2));
>     std::sort(moments.data(), moments.data() + moments.size());
> ```

This doesn't seem correct to me. If you reorder the principal moments so that, for example, what was Ixx becomes Izz, shouldn't that produce a non-identity rotation matrix?

What would happen if you removed the special case and applied the eigen solver even to a diagonal inertia -- wouldn't the returned Eigenvectors be non-identity if the eigenvalues were reordered?
___
*[`multibody/tree/rotational_inertia.cc` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMwiHDCebZx71Fn4Fs:-NUMwiHDCebZx71Fn4Ft:b-u0jjjh) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L56)):*
> ```cpp
>   // The eigen solver orders the storage of eigenvalues in increasing value and
>   // orders the storage of their corresponding eigenvectors similarly.
>   // Note: Since the unit inertia 3x3 matrix should be positive semi-definite,
> ```

minor: _unit_ inertia?
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 532 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMxDz57wQhnTmiHEJd:-NUMxDz57wQhnTmiHEJe:b-gwi2zg) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L532)):*
> ```cpp
>       UnitInertia<double>::SolidEllipsoid(a, b, c);
> 
>   // Ensure UnitInertia can calculate principal moments and principal axes.
> ```

BTW why the unit inertia test here? Isn't that tested elsewhere? If there is some reason to retest here that should be explained. If it is just to obtain numbers to use later, it should say that rather than look like a test.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 543 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMzBHl-bywYxDa9XB0:-NUMzBHl-bywYxDa9XB1:b-ym8q3m) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L543)):*
> ```cpp
> 
>   // For mass = 3 kg, ensure RotationalInertia calculates the same principal
>   // axes as UnitInertia (above) with thrice the moments of inertia.
> ```

minor: should state what is being tested. Here it is just the special case of an exactly-principal-already inertia.

Actually this isn't quite sufficient because the inertias come in already sorted into ascending order. That means the test would pass even if someone accidentally deleted the std::sort in the special case. Would be better to reorder a,b,c above so that the inertias are not properly ordered initially.

See my comment above -- shouldn't the eigenvectors be non-identity if the principal moments are reordered?
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUN0YQp4iJ-d2w9Prtx:-NUN0YQp4iJ-d2w9Prty:bqibxmm) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L570)):*
> ```cpp
>   EXPECT_NEAR(std::abs(Px_B.dot(Cx_B)), 1.0, kTolerance);
>   EXPECT_NEAR(std::abs(Py_B.dot(Cy_B)), 1.0, kTolerance);
>   EXPECT_NEAR(std::abs(Pz_B.dot(Cz_B)), 1.0, kTolerance);
> ```

minor: with the `std::abs()` here wouldn't this match a left hand frame to a right hand frame? This test doesn't seem strict enough.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 10 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NUN1ZTHEa4m_B0Q2PjN:bm245zf) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*

minor: these static_asserts don't make sense with constant arguments. Of course those are lower triangular! These should be removed.
</blockquote></details>

This is a pre-existing pattern that was carefully designed.  These lines are guarding that the hard-coded indices just below like `(1, 0)` are the proper hard-coded values, versus erroneous constants like `(0, 1)`.

It is true that these assertions will never fail.  That's why they are assertions!  No assertions are ever expected to fail, they are stating an invariant about the code.

Possibly there is a defect that the documentation is unclear, since you didn't follow along while reading the code.  But I think the code is properly written.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 10 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NUN2K680QjDuQhLW0fC:b-ndgtc7) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*

This is a pre-existing pattern that was carefully designed.  These lines are guarding that the hard-coded indices just below like `(1, 0)` are the proper hard-coded values, versus erroneous constants like `(0, 1)`.

It is true that these assertions will never fail.  That's why they are assertions!  No assertions are ever expected to fail, they are stating an invariant about the code.

Possibly there is a defect that the documentation is unclear, since you didn't follow along while reading the code.  But I think the code is properly written.
</blockquote></details>

Perhaps moving the `static_assert` to _immediately_ prior to the lines in question would help clarify:

```c++
  I_double(0, 0) = ExtractDoubleOrThrow(I_BP_E(0, 0));
  I_double(1, 1) = ExtractDoubleOrThrow(I_BP_E(1, 1));
  I_double(2, 2) = ExtractDoubleOrThrow(I_BP_E(2, 2));
  static_assert(is_lower_triangular_order(1, 0), "Invalid indices");
  static_assert(is_lower_triangular_order(2, 0), "Invalid indices");
  static_assert(is_lower_triangular_order(2, 1), "Invalid indices");
  I_double(1, 0) = ExtractDoubleOrThrow(I_BP_E(1, 0));
  I_double(2, 0) = ExtractDoubleOrThrow(I_BP_E(2, 0));
  I_double(2, 1) = ExtractDoubleOrThrow(I_BP_E(2, 1));
```

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 10 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NUN6auD8F9lu-XCvePz:bufoitb) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*

Perhaps moving the `static_assert` to _immediately_ prior to the lines in question would help clarify:

```c++
  I_double(0, 0) = ExtractDoubleOrThrow(I_BP_E(0, 0));
  I_double(1, 1) = ExtractDoubleOrThrow(I_BP_E(1, 1));
  I_double(2, 2) = ExtractDoubleOrThrow(I_BP_E(2, 2));
  static_assert(is_lower_triangular_order(1, 0), "Invalid indices");
  static_assert(is_lower_triangular_order(2, 0), "Invalid indices");
  static_assert(is_lower_triangular_order(2, 1), "Invalid indices");
  I_double(1, 0) = ExtractDoubleOrThrow(I_BP_E(1, 0));
  I_double(2, 0) = ExtractDoubleOrThrow(I_BP_E(2, 0));
  I_double(2, 1) = ExtractDoubleOrThrow(I_BP_E(2, 1));
```

</blockquote></details>

If the test was checking that `operator()(i,j)` in `I_BP_E(1,0)` was actually accessing the 1,0 element rather than 0,1 I would agree the test was meaningful. But in fact it is just testing that that i >= j which is obviously true when i==1 and j==0! And that is the definition of the lower triangle of a matrix. So I don't see how this is actually asserting anything but a tautology.

My preference would be to replace it with a comment that says "be sure to access only the lower triangle". OTOH it is harmless so I can't object to it being there if someone loves it. However in that case I think it needs a comment so the next reader doesn't waste time like I did trying to figure out what it was doing when it appears to do nothing.

mitiguy (commented):
@sherm1  Next pass done PTAL

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NV0etsK5zK3h78W6UaM:bf5u5ze)* status: 4 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @sherm1)

___
*[`multibody/tree/rotational_inertia.h` line 541 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMu0yrCnKQcMcxLIr-:-NUwqeJeE9_mGlgKmhX4:b-snh328) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L541)):*

minor: why "unit inertia" here -- shouldn't this be rotational inertia?
</blockquote></details>

Done here and elsewhere.
___
*[`multibody/tree/rotational_inertia.h` line 542 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMunE_DPuGhcQ_CsEj:-NUx4_sw8Zaa1bj87_0Z:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L542)):*

minor: this appears to be an optional return but the documentation doesn't explain that
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.h` line 543 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMtdI_14LEMFWqMGEB:-NUxHyeG0QIPSwnnFaS0:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L543)):*

minor: frame P is undefined at this point (you mentioned point P above but it's not clear what the axes are)
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.h` line 549 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMsmfDAouBk76vDZIm:-NUxLIvf5NyKpuAvzfSi:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.h#L549)):*

BTW it took me a while to decipher this sentence. Is there a better way to say it? I think you're saying we calculate the principal moments in ascending order, then call the smallest one Ixx and the largest Izz. As written it isn't clear that Ixx etc are _defined_ by the ordering, rather than _dictating_ the ordering.
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NUxNj9j0GyimSGEb4EF:bh98f5j) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*

If the test was checking that `operator()(i,j)` in `I_BP_E(1,0)` was actually accessing the 1,0 element rather than 0,1 I would agree the test was meaningful. But in fact it is just testing that that i >= j which is obviously true when i==1 and j==0! And that is the definition of the lower triangle of a matrix. So I don't see how this is actually asserting anything but a tautology.

My preference would be to replace it with a comment that says "be sure to access only the lower triangle". OTOH it is harmless so I can't object to it being there if someone loves it. However in that case I think it needs a comment so the next reader doesn't waste time like I did trying to figure out what it was doing when it appears to do nothing.
</blockquote></details>

Done.  Comments added.
___
*[`multibody/tree/rotational_inertia.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMzvlo6t0eRhmssra8:-NUxcj4ZBJo8kjYitW4V:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L38)):*

This doesn't seem correct to me. If you reorder the principal moments so that, for example, what was Ixx becomes Izz, shouldn't that produce a non-identity rotation matrix?

What would happen if you removed the special case and applied the eigen solver even to a diagonal inertia -- wouldn't the returned Eigenvectors be non-identity if the eigenvalues were reordered?
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.cc` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMwiHDCebZx71Fn4Fs:-NUxM7_N7YQ5CSRv39t1:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L56)):*

minor: _unit_ inertia?
</blockquote></details>

Done.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @sherm1)

___
*[`multibody/tree/test/rotational_inertia_test.cc` line 532 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMxDz57wQhnTmiHEJd:-NV0f4fD2lj1U2SLzxA6:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L532)):*

BTW why the unit inertia test here? Isn't that tested elsewhere? If there is some reason to retest here that should be explained. If it is just to obtain numbers to use later, it should say that rather than look like a test.
</blockquote></details>

Done.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 543 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMzBHl-bywYxDa9XB0:-NV0f0TQ0WfTW21RQ4FL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L543)):*

minor: should state what is being tested. Here it is just the special case of an exactly-principal-already inertia.

Actually this isn't quite sufficient because the inertias come in already sorted into ascending order. That means the test would pass even if someone accidentally deleted the std::sort in the special case. Would be better to reorder a,b,c above so that the inertias are not properly ordered initially.

See my comment above -- shouldn't the eigenvectors be non-identity if the principal moments are reordered?
</blockquote></details>

Done.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUN0YQp4iJ-d2w9Prtx:-NV0f6qk11bEvjRLWcHn:b-dg1yhk) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L570)):*

minor: with the `std::abs()` here wouldn't this match a left hand frame to a right hand frame? This test doesn't seem strict enough.
</blockquote></details>

I'll fix this.

sherm1 (commented):
Feature :lgtm: pending one open issue from earlier, a suggested simplification, and a few comments

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NV0xBGY5qmoAdFMs-u1:b-5kpiwe)* status: 5 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/rotational_inertia.h` line 547 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0kihu2oPgSmSOpaMf:-NV0kihu2oPgSmSOpaMg:b-cek67p) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.h#L547)):*
> ```objc
>   /// I_BPo_E (body B's rotational inertia about-point Po) and frame P contains
>   /// right-handed orthogonal unit vectors Px, Py, Pz. The 1ˢᵗ column of R_EP is
>   /// Px_E (Px expressed in frame E) which is parallel to the principal acis
> ```

typo: acis -> axis
___
*[`multibody/tree/rotational_inertia.cc` line 38 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0lUBQA5Zt7XkCMWD4:-NV0lUBQA5Zt7XkCMWD5:bphhdrr) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L38)):*
> ```cpp
>     double Imin = I_double(0, 0);
>     double Imed = I_double(1, 1);
>     double Imax = I_double(2, 2);
> ```

BTW consider a comment here noting that this is an initial guess (i.e. not yet actually Imin, Imed, Imax)
___
*[`multibody/tree/rotational_inertia.cc` line 52 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0uTu5429h8j60jpS4:-NV0uTu5429h8j60jpS5:b-umr5z7) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L52)):*
> ```cpp
>       std::swap(imed, imax);  // imed = 1 or 2 and imax = 0 or 1.
>     }
>     DRAKE_ASSERT(Imin <= Imed && Imed <= Imax);
> ```

BTW consider this more-compact alternative. Also, adding a comment showing what's known at each point makes it easier to follow, IMO



_Suggestion:_
```cpp
  auto Imin = std::make_pair(Ixx,0);  // Initial guess.
  auto Imed = std::make_pair(Iyy,1);
  auto Imax = std::make_pair(Izz,2);

  if (Imin.first > Imed.first) std::swap(Imin, Imed);
  // Imin <= Imed
  if (Imin.first > Imax.first) std::swap(Imin, Imax);
  // Imin <= Imax, Imin <= Imid
  if (Imed.first > Imax.first) std::swap(Imed, Imax);
  // Imin <= Imed <= Imax
```
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 535 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0xSRB2zk0v4wxx5rE:-NV0xSRB2zk0v4wxx5rF:b-itzxwa) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/test/rotational_inertia_test.cc#L535)):*
> ```cpp
>   drake::math::RotationMatrix<double> R_BP;
> 
>   // Verify principal moments of inertia for straightforward case.
> ```

minor: test comments should be clear about when you are testing the bespoke sorting method for diagonal inertias vs. the general case when there are products of inertia. Should be clear that both cases get tested somewhere.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 6 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/rotational_inertia.cc` line 52 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV198_OEwnETaKEL4zf:-NV198_OEwnETaKEL4zg:b-lph1se) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L52)):*
> ```cpp
>       std::swap(imed, imax);  // imed = 1 or 2 and imax = 0 or 1.
>     }
>     DRAKE_ASSERT(Imin <= Imed && Imed <= Imax);
> ```

BTW I couldn't resist taking one more shot. Here it is in four lines. Performance should be close to optimal due to fixed sizes.
```c++
  using Pair = std::pair<double,int>;
  std::array I{Pair{Ixx,0},Pair{Iyy,1},Pair{Izz,2}};
  std::sort(I.begin(), I.end(), [](const Pair& l, const Pair& r) {
    return l.first < r.first;});
// Here I[0] <= I[1] <= I[2]
```

mitiguy (commented):
@sherm1   Seems like this is ready for another look.
Although CI failed sometimes, it may be unrelated:
"abort due to malloc #1 while max_num_allocations = 0 in effect"

> Reviewed 1 of 4 files at r1, 2 of 3 files at r2, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NV2Dzgz42buOKsVW6y6:b-ljybzr)* status: 2 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

___
*[`multibody/tree/rotational_inertia.h` line 547 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0kihu2oPgSmSOpaMf:-NV1GH002TZ9Cq7obs4z:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.h#L547)):*

typo: acis -> axis
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.cc` line 38 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0lUBQA5Zt7XkCMWD4:-NV1I2Ss4afq7NPVyszl:b-3zbf4g) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L38)):*

BTW consider a comment here noting that this is an initial guess (i.e. not yet actually Imin, Imed, Imax)
</blockquote></details>

No longer needed -- better code from you.
___
*[`multibody/tree/rotational_inertia.cc` line 52 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0uTu5429h8j60jpS4:-NV1XuYqCJ2ByJ-l81N9:bdmpcyt) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L52)):*

BTW consider this more-compact alternative. Also, adding a comment showing what's known at each point makes it easier to follow, IMO
</blockquote></details>

Done.  Used your next code.
___
*[`multibody/tree/rotational_inertia.cc` line 52 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV198_OEwnETaKEL4zf:-NV1Y--02wPdLHTWJIOS:b-yhtdi9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/rotational_inertia.cc#L52)):*

BTW I couldn't resist taking one more shot. Here it is in four lines. Performance should be close to optimal due to fixed sizes.
```c++
  using Pair = std::pair<double,int>;
  std::array I{Pair{Ixx,0},Pair{Iyy,1},Pair{Izz,2}};
  std::sort(I.begin(), I.end(), [](const Pair& l, const Pair& r) {
    return l.first < r.first;});
// Here I[0] <= I[1] <= I[2]
```

</blockquote></details>

Done.  I like this one.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUN0YQp4iJ-d2w9Prtx:-NV2Dpku6YM_Xm8_TVBp:bkfu4t4) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/test/rotational_inertia_test.cc#L570)):*

I'll fix this.
</blockquote></details>

Better comments added.
Also added a test of right-handedness of frame.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 535 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV0xSRB2zk0v4wxx5rE:-NV1hMVjD4P99LWpsSvm:bgoehq7) ([raw file](https://github.com/RobotLocomotion/drake/blob/bc9d0aa8ac32f310a1432e2cecd6be2a1fadc65e/multibody/tree/test/rotational_inertia_test.cc#L535)):*

minor: test comments should be clear about when you are testing the bespoke sorting method for diagonal inertias vs. the general case when there are products of inertia. Should be clear that both cases get tested somewhere.
</blockquote></details>

Done  (I think).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 3 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/rotational_inertia.h` line 560 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV2Fm750LvUOmaICSlq:-NV2Fm750LvUOmaICSlr:b-3adfyl) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.h#L560)):*
> ```objc
>   /// that can be converted to a double (discarding any supplemental scalar data
>   /// such as derivatives of an AutoDiffXd).
>   Vector3<double> CalcPrincipalMomentsAndAxesOfInertia(
> ```

BTW If the user is calling `CalcPrincipalMomentsAndAxesOfInertia` instead of `CalcPrincipalMomentsOfInertia`, surely they want the `R_EP` information as well?  It seems like in practice nobody would ever pass `nullptr` here.

If that premise is correct, then why is this function formulated as an awkward split between one return value and one output-only argument, instead of the smoother alternative of returning both values directly?

```c++
const auto& [moments, axes] = foo.CalcPrincipalMomentsAndAxesOfInertia();
```

sherm1 (commented):
That CI failure is mysterious and I don't see how the addition of a new method here could have caused it. Should be resolved though before merging.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NV5hwKn6arMyyi1_JX5:b-apf6et)* status: 4 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/rotational_inertia.h` line 560 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV2Fm750LvUOmaICSlq:-NV5eM8b-Z_ez0KR8Y7v:b-gqpg40) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.h#L560)):*

BTW If the user is calling `CalcPrincipalMomentsAndAxesOfInertia` instead of `CalcPrincipalMomentsOfInertia`, surely they want the `R_EP` information as well?  It seems like in practice nobody would ever pass `nullptr` here.

If that premise is correct, then why is this function formulated as an awkward split between one return value and one output-only argument, instead of the smoother alternative of returning both values directly?

```c++
const auto& [moments, axes] = foo.CalcPrincipalMomentsAndAxesOfInertia();
```
</blockquote></details>

Good catch! I agree this should not be optional.
___
*[`multibody/tree/rotational_inertia.h` line 560 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5hBpf7urd3ZXGg9lb:-NV5hBpf7urd3ZXGg9lc:blgdudd) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.h#L560)):*
> ```objc
>   /// that can be converted to a double (discarding any supplemental scalar data
>   /// such as derivatives of an AutoDiffXd).
>   Vector3<double> CalcPrincipalMomentsAndAxesOfInertia(
> ```

BTW consider whether you want to document the stable behavior for diagonal inertias with duplicate elements. I don't think it's necessary to make that a promise but it is nice behavior to return an un-permuted identity matrix for a sphere!
___
*[`multibody/tree/rotational_inertia.cc` line 53 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5fjWa-fwDUlDoNYKl:-NV5fjWa-fwDUlDoNYKm:b326kgo) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.cc#L53)):*
> ```cpp
>       const uint index_min = I[0].second;  // Index associated with Imin.
>       const uint index_med = I[1].second;  // Index associated with Imed.
>       const uint index_max = I[2].second;  // Index associated with Imax.
> ```

nit: these should be `int` not unsigned which we avoid like the plague
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 581 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5ghgaBXZGu48ybieJ:-NV5ghgaBXZGu48ybieK:b-clg4h6) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/test/rotational_inertia_test.cc#L581)):*
> ```cpp
>   // Rotate body B by 30 degrees and verify principal moments/axes. This tests
>   // a general case for a rotational inertia with non-zero products of inertia.
>   I_BBcm_B = RotationalInertia<double>(Imin, Imed, Imax);
> ```

minor: since we want a stable sort you need a test where two or all of the principal moments are the same to make sure the identity matrix doesn't get improperly mangled

xuchenhan-tri:
The additional heap allocation comes from the std::sort introduced in this PR.

jwnimmer-tri:
FYI you can use `std::sort` and still be stable in this case, if you sort the whole pair instead of just the first element of the pair.  The second element already provides the stability you need.

sherm1 (commented):
> The additional heap allocation comes from the std::sort introduced in this PR.

Rats. I hoped with a fixed-size input it wouldn't use the heap :(. Back to Plan B.'

> you can use `std::sort` and still be stable in this case, if you sort the whole pair

!! -- maybe stable_sort is more of a heap pig than sort (doubtful)?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NV7z7wD-W711_708eQc:bi3e43v)* status: 4 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

jwnimmer-tri:
The https://en.cppreference.com/w/cpp/algorithm/stable_sort page explains that `stable_sort` will allocate extra storage, in order to reduce big-O complexity.

mitiguy (commented):
FWIW:  Documentation on [stable_sort](https://en.cppreference.com/w/cpp/algorithm/stable\_sort) and [sort](https://en.cppreference.com/w/cpp/algorithm/sort) suggest that both may allocate memory.  The documentation for [<b>sort</b>](https://en.cppreference.com/w/cpp/algorithm/sort)  includes:

Exceptions ...
If the algorithm fails to allocate memory, std::bad_alloc is thrown.

> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVB6GWi7jaSjW15KMvd:b9zn9ln)* status: 1 unresolved discussion, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

___
*[`multibody/tree/rotational_inertia.h` line 560 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV2Fm750LvUOmaICSlq:-NVB7BGt1RmaFA7D8D1q:b34w5ub) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.h#L560)):*

Good catch! I agree this should not be optional.
</blockquote></details>

I moved/renamed the original method to a private subsection of the RotationalInertia class.  The following public methods call the privatemethod, with the 2nd argument being nullptr or not, depending on what is needed.

public:
CalcPrincipalMomentsOfInertia();
CalcPrincipalMomentsAndAxesOfInertia();

private:
CalcPrincipalMomentsAndMaybeAxesOfInertia();
___
*[`multibody/tree/rotational_inertia.h` line 560 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5hBpf7urd3ZXGg9lb:-NVBdKJt66lW-AldXHOT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.h#L560)):*

BTW consider whether you want to document the stable behavior for diagonal inertias with duplicate elements. I don't think it's necessary to make that a promise but it is nice behavior to return an un-permuted identity matrix for a sphere!
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.cc` line 53 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5fjWa-fwDUlDoNYKl:-NV5u0xM9-IM9sc8Bs2v:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/rotational_inertia.cc#L53)):*

nit: these should be `int` not unsigned which we avoid like the plague
</blockquote></details>

Done.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 581 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NV5ghgaBXZGu48ybieJ:-NV5xkTjB4e_Ewm4YMc3:b-ffuyvg) ([raw file](https://github.com/RobotLocomotion/drake/blob/87821c5225a1b2fedcf59f8d8448c0c078d87e08/multibody/tree/test/rotational_inertia_test.cc#L581)):*

minor: since we want a stable sort you need a test where two or all of the principal moments are the same to make sure the identity matrix doesn't get improperly mangled
</blockquote></details>

Done.  Tested when all three principal moments are equal -- which should always respond to identity matrix.

jwnimmer-tri:
> FWIW: Documentation on [stable_sort](https://en.cppreference.com/w/cpp/algorithm/stable_sort) and [sort](https://en.cppreference.com/w/cpp/algorithm/sort) suggest that both may allocate memory. The documentation for [sort](https://en.cppreference.com/w/cpp/algorithm/sort) includes: Exceptions ... If the algorithm fails to allocate memory, std::bad_alloc is thrown.

That's only if you provide an ExecutionPolicy.

mitiguy (commented):
Thanks Jeremy.  Sherm is on vacation today and already provided an LGTM.
What do you think of this being ready for Platform review (and ideally subsequent merge)?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVBuuW8DwZx-rZLB1Fi:bg3c50q)* status: needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

jwnimmer-tri:
There are no open discussion and it has feature LGTM.  I think its ready for platform review.

mitiguy (commented):
Platform review +@xuchenhan-tri 
Note: The schedule has Sammy, but it may be too late for East coast -- and since Xuchen is willing.  Thank you Xuchen.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVC1i89C783kLCA_GH-:b-c5gtek)* status: LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
Platform review done. I have a few comments and questions. Probably worth another round before stamping.

> Reviewed 1 of 4 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVCBm2R6C5rX9pza5HE:bv63eoe)* status: 8 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @xuchenhan-tri)

___
*[`multibody/tree/rotational_inertia.h` line 527 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC3lFPET265Cax94MX:-NVC3lFPET265Cax94MY:b-190zxk) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L527)):*
> ```objc
>   /// @retval The 3 principal moments of inertia [Imin Imed Imax], sorted in
>   /// ascending order (Imin ≤ Imed ≤ Imax).
>   /// @throws std::exception if the eigenvalue solver fails or if scalar type T
> ```

nit, it's not clear how and where an eigenvalue solver is involved. It'd be better to state the throw condition on what the user can point their fingers to.  Ditto for the function below.

In fact, when does the eigenvalue solver fail? Do we ever expect it to happen?



_Code quote:_
```objc
/// @throws std::exception if the eigenvalue solver fails
```
___
*[`multibody/tree/rotational_inertia.h` line 531 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC3XWaCVzxNfan9puw:-NVC3XWaCVzxNfan9pux:bj02ycw) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L531)):*
> ```objc
>   /// @note: This method only works for a rotational inertia with scalar type T
>   /// that can be converted to a double (discarding any supplemental scalar data
>   /// such as derivatives of an AutoDiffXd).
> ```

BTW, since we only support three scalar types, would it be more concise to say symbolic is not supported for this function? Ditto for the function below.



_Code quote:_
```objc
  /// @note: This method only works for a rotational inertia with scalar type T
  /// that can be converted to a double (discarding any supplemental scalar data
  /// such as derivatives of an AutoDiffXd).
```
___
*[`multibody/tree/rotational_inertia.h` line 540 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC4KYwBzVJvMcUnWlj:-NVC4KYwBzVJvMcUnWlk:b9xu9px) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L540)):*
> ```objc
>   /// Forms the 3 principal moments of inertia and their 3 associated principal
>   /// directions for `this` rotational inertia.
>   /// @returns 3 principal moments of inertia [Ixx Iyy Izz], sorted in ascending
> ```

BTW, is the deviation from the notation from the function above (Imin, Imed, Imax) intentional? It'd be nice to see consistent notations across two closely-related functions.



_Code quote:_
```objc
[Ixx Iyy Izz]
```
___
*[`multibody/tree/rotational_inertia.cc` line 33 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7UCr3L9Xaq-tfg9W:-NVC7UCr3L9Xaq-tfg9X:bnzy8vy) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L33)):*
> ```cpp
>   // elements and the principal directions are simple (e.g., [1, 0, 0]).
>   const bool is_diagonal = (I_double(1, 0) == 0 && I_double(2, 0) == 0 &&
>                             I_double(2, 1) == 0);
> ```

BTW, one of the first things that the eigen SelfAdjointEigenSolver does is to check if the off diagonal entries are close enough to zero. So I wonder if this special case is really necessary? How much do we gain compare to the bloat in code?



_Code quote:_
```cpp
  // If all products of inertia are zero, no need to calculate eigenvalues or
  // eigenvectors as the principal moments of inertia are just the diagonal
  // elements and the principal directions are simple (e.g., [1, 0, 0]).
  const bool is_diagonal = (I_double(1, 0) == 0 && I_double(2, 0) == 0 &&
                            I_double(2, 1) == 0);
```
___
*[`multibody/tree/rotational_inertia.cc` line 68 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7wSj4krxojuKZud2:-NVC7wSj4krxojuKZud3:bbvi2kz) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L68)):*
> ```cpp
>   // Calculate eigenvalues, possibly with eigenvectors.
>   Eigen::SelfAdjointEigenSolver<Matrix3<double>> eig_solve;
>   const int compute_Eigenvectors = principal_directions != nullptr ?
> ```

nit snake_case for variables.



_Suggestion:_
```cpp
compute_eigenvectors
```
___
*[`multibody/tree/rotational_inertia.cc` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC8Za40jcYM15BpqBf:-NVC8Za40jcYM15BpqBg:b-i1zwf8) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L70)):*
> ```cpp
>   const int compute_Eigenvectors = principal_directions != nullptr ?
>       Eigen::ComputeEigenvectors : Eigen::EigenvaluesOnly;
>   eig_solve.compute(I_double, compute_Eigenvectors);
> ```

BTW, are you aware of computeDirect which is supposed to be faster? https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html#afe520161701f5f585bcc4cedb8657bd1



_Code quote:_
```cpp
eig_solve.compute(I_double, compute_Eigenvectors);
```
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 609 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVCBGEN49qQiHQO3GH2:-NVCBGEN49qQiHQO3GH3:b-7ilvee) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/test/rotational_inertia_test.cc#L609)):*
> ```cpp
>   const Vector3<double> Px_B = R_BP.col(0), Cx_B = R_BP.col(0);
>   const Vector3<double> Py_B = R_BP.col(1), Cy_B = R_BP.col(1);
>   const Vector3<double> Pz_B = R_BP.col(2), Cz_B = R_BP.col(2);
> ```

typos



_Suggestion:_
```cpp
  const Vector3<double> Px_B = R_BP.col(0), Cx_B = R_BC.col(0);
  const Vector3<double> Py_B = R_BP.col(1), Cy_B = R_BC.col(1);
  const Vector3<double> Pz_B = R_BP.col(2), Cz_B = R_BC.col(2);
```
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 613 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVCBSPM3cjhki9w5-pd:-NVCBSPM3cjhki9w5-pe:b-uvhqw0) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/test/rotational_inertia_test.cc#L613)):*
> ```cpp
>   EXPECT_NEAR(std::abs(Py_B.dot(Cy_B)), 1.0, kTolerance);  // Py parallel to Cy.
>   EXPECT_NEAR(std::abs(Pz_B.dot(Cz_B)), 1.0, kTolerance);  // Pz parallel to Cz.
>   EXPECT_NEAR(Px_B.cross(Py_B).dot(Pz_B), 1.0, kTolerance);  // Right-handed.
> ```

nit, seems to me that this is a consequence of the previous three statements.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 8 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @xuchenhan-tri)

___
*[`multibody/tree/rotational_inertia.cc` line 33 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7UCr3L9Xaq-tfg9W:-NVSOQeJ4Ysul8zXt_rG:b-a71m5z) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L33)):*

BTW, one of the first things that the eigen SelfAdjointEigenSolver does is to check if the off diagonal entries are close enough to zero. So I wonder if this special case is really necessary? How much do we gain compare to the bloat in code?
</blockquote></details>

BTW I wonder if it would also stable-sort the eigenvectors in case they are equal (i.e. can we guarantee that if the eigenvalues are all three the same then we get eigenvectors that form an identity matrix, rather than some perturbation of an identity?)
___
*[`multibody/tree/rotational_inertia.cc` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC8Za40jcYM15BpqBf:-NVSOw_w1L6bVu-vc-9g:bqixffj) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L70)):*

BTW, are you aware of computeDirect which is supposed to be faster? https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html#afe520161701f5f585bcc4cedb8657bd1
</blockquote></details>

BTW possible accuracy problem with that methd?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 8 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy and @xuchenhan-tri)

___
*[`multibody/tree/rotational_inertia.cc` line 33 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7UCr3L9Xaq-tfg9W:-NVV93YM3feyv1Rx9EZd:b-777iu6) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L33)):*

BTW I wonder if it would also stable-sort the eigenvectors in case they are equal (i.e. can we guarantee that if the eigenvalues are all three the same then we get eigenvectors that form an identity matrix, rather than some perturbation of an identity?)
</blockquote></details>

I think the eigenvectors would be form the identity because the algorithm starts with an identity matrix and iteratively tag on rotation matrices to form eigenvectors.
___
*[`multibody/tree/rotational_inertia.cc` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC8Za40jcYM15BpqBf:-NVVBWUSEZV2rxOxW1V8:besdi5g) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L70)):*

BTW possible accuracy problem with that methd?
</blockquote></details>

According to the Eigen doc, the relative error is 1e-8 for eigenvalues. It may or may not be too coarse depending on the use case.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281)* status: 4 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/tree/rotational_inertia.h` line 527 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC3lFPET265Cax94MX:-NVW94sx7S1P_57PcdHS:bywzx2l) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L527)):*

nit, it's not clear how and where an eigenvalue solver is involved. It'd be better to state the throw condition on what the user can point their fingers to.  Ditto for the function below.

In fact, when does the eigenvalue solver fail? Do we ever expect it to happen?
</blockquote></details>

Done.  Updated here and elsewhere.
___
*[`multibody/tree/rotational_inertia.h` line 531 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC3XWaCVzxNfan9puw:-NVW9tVo8C2nV5CKATEt:bywzx2l) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L531)):*

BTW, since we only support three scalar types, would it be more concise to say symbolic is not supported for this function? Ditto for the function below.
</blockquote></details>

Done.  Updated here and elsewhere.
___
*[`multibody/tree/rotational_inertia.h` line 540 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC4KYwBzVJvMcUnWlj:-NVWB1KJ1zIcHIKi0G4Q:b-80fghs) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.h#L540)):*

BTW, is the deviation from the notation from the function above (Imin, Imed, Imax) intentional? It'd be nice to see consistent notations across two closely-related functions.
</blockquote></details>

Per f2f, the deviation is intentional.  Summary of our conversation:
For the method that just calculates the moments of inertia, there are 3 of them, ordered from minimum to maximum, hence the names [Imin, Imed, Imax].  There is no need to inject x, y, z as the RotationalInertia inherently has a unitary vector basis with x, y, z, so no need to needlessly create confusion for a simple method.

For the method that calculates the moments of inertia and principal directions, the directions are called Px, Py, Pz, so there is some semblance of another xyz basis in play -- hence the names [Ixx  Iyy  Izz].  I considered using [Imin, Imed, Imax], everywhere, but as you and I agreed, this might need extra verbage to associate Px with Imin, Py with Imed, and Pz with Imax.
___
*[`multibody/tree/rotational_inertia.cc` line 33 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7UCr3L9Xaq-tfg9W:-NVWCwJz-rqyW8yhZjt5:bkx992u) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L33)):*

I think the eigenvectors would be form the identity because the algorithm starts with an identity matrix and iteratively tag on rotation matrices to form eigenvectors.
</blockquote></details>

1. The std::sort() of the 3 pairs that contain equal eigenvalues and an index of 0, 1, 2 will sort and eventually produce an identity rotation matrix.  This is now documented and tested.
2.  The specialized code that test for no-products of inertia is from a prior function in Drake Master that only calculated moments of inertia (did not calculate products of inertia).  If this code-bloat should be removed, it can be done in another PR (if desirable).
___
*[`multibody/tree/rotational_inertia.cc` line 68 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC7wSj4krxojuKZud2:-NVWEWpM8Zvaj6n87ZNL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L68)):*

nit snake_case for variables.
</blockquote></details>

Done.
___
*[`multibody/tree/rotational_inertia.cc` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVC8Za40jcYM15BpqBf:-NVWGUak4IOLtzWxZGqD:bspm2cj) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/rotational_inertia.cc#L70)):*

According to the Eigen doc, the relative error is 1e-8 for eigenvalues. It may or may not be too coarse depending on the use case.
</blockquote></details>

Added TODO for possible future investigation.
However, lower priority as this function should not be in an inner-loop.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 609 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVCBGEN49qQiHQO3GH2:-NVWH0dyEBsJcDG_yAT2:bx3qkc4) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/test/rotational_inertia_test.cc#L609)):*

typos
</blockquote></details>

Done.  Good eyes.  Otherwise, the test is pointless.
___
*[`multibody/tree/test/rotational_inertia_test.cc` line 613 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NVCBSPM3cjhki9w5-pd:-NVWHDHJ5nrfgcbz5gxP:b-hcuqr8) ([raw file](https://github.com/RobotLocomotion/drake/blob/eb54cab2d2601f2de0de333c76644e7cae874acb/multibody/tree/test/rotational_inertia_test.cc#L613)):*

nit, seems to me that this is a consequence of the previous three statements.
</blockquote></details>

Done  (I think -- based on change above).

xuchenhan-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVWOBJ74-FFnOQblNEn:bnfp4nl)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

mitiguy (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-:-NVWUUAWBw-c2bpXLBCd:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees sherm1(platform),xuchenhan-tri(platform)

jwnimmer-tri:
*[`multibody/tree/rotational_inertia.cc` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19281#-NUMvYka5KDJ68jShRQ4:-NW4vSVRBh9s6oIRWmm-:b-odaxou) ([raw file](https://github.com/RobotLocomotion/drake/blob/a152fa2f95fbe2ea39ab43a6b7588687e43d9934/multibody/tree/rotational_inertia.cc#L19)):*

Done.  Comments added.
</blockquote></details>

See also #6138.

Issue #17662, opened 2022-08-01
Visualize inertia ellipsoids


jwnimmer-tri:
I'd like to have a checkbox in our visualizer(s) to display the inertia ellipsoids for our MbP's rigid bodies.  This would help users debug their SDFormat models by visualizing the masses & etc to make sure everything's in the right place and of an appropriate scale.

EricCousineau-TRI:
Howdy @trowell-tri! Any chance this is in your queue for next couple of weeks?
Excited to try this out to check on models, e.g. on the IIWA, related to [this post in Drake Slack](https://drakedevelopers.slack.com/archives/C2WBPQDB7/p1666731346091319) and torque control (landed via #18295, relatedly Anzu issue 9395).

trowell-tri:
Yes! I hope to start working on this later next week, and get something usable if not final this month.

EricCousineau-TRI:
Awesome, thanks!

RussTedrake:
fwiw -- I saw two student bugs just last night related to having center of masses outside of the collision geometry (leading to confusing behavior in sim).  SDF makes it far too easy to make this mistake.  This feature will really help.

RussTedrake:
two MORE student projects have had the same issue.  i can't imagine this will land in time to help the projects this semester, but it is badly needed.  

As a one-off, I'm recommending people use something like this to at least visualize the center of mass:
```
    plant_context = plant.GetMyContextFromRoot(context)
    body = plant.GetBodyByName("base_link")
    p_BCom = body.CalcCenterOfMassInBodyFrame(plant_context)
    X_WB = body.EvalPoseInWorld(plant_context)
    p_WCom = X_WB @ p_BCom
    meshcat.SetObject("center_of_mass", Sphere(0.01), Rgba(0, 1, 0, 1))
    meshcat.SetTransform("center_of_mass", RigidTransform(p_WCom))    
```

trowell-tri:
Plan of attack:
• Add geometry to scene graph
• Easiest to affix geometry to existing bodies?
• SceneGraph may not support wireframes (but I'm not sure that's the way we want to go anyway)
• Group under a magic name with own alpha slider

jwnimmer-tri:
New brainwave from today: by default, we should add the inertia geoms with alpha=0 so they are invisible.  Then, for specific visualizers that opt-in, those visualizer front-ends can offer a control to unhide the geoms.

EricCousineau-TRI:
[...]

PR #19287, opened 2023-04-26
[meshcat] Enable alpha adjustment of invisible geometry

trowell-tri:
If geometry is invisible, the meshcat alpha sliders will never make it visible because MeshcatVisualizer simply multiplies geometry alpha by the slider value. To improve this somewhat, check if all geometry in a visualizer is invisible and if so use the absolute alpha slider value as the new alpha instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)

### DIFF
--- a/geometry/meshcat_visualizer.cc
+++ b/geometry/meshcat_visualizer.cc
@@ -1,5 +1,6 @@
 #include "drake/geometry/meshcat_visualizer.h"
 
+#include <algorithm>
 #include <memory>
 #include <optional>
 #include <string>
@@ -257,9 +258,18 @@ void MeshcatVisualizer<T>::SetTransforms(
 
 template <typename T>
 void MeshcatVisualizer<T>::SetColorAlphas() const {
+  double max_alpha = 0.0;
+  for (const auto& [geom_id, path] : geometries_) {
+    max_alpha = std::max(max_alpha, colors_[geom_id].a());
+  }
+
   for (const auto& [geom_id, path] : geometries_) {
     Rgba color = colors_[geom_id];
-    color.update({}, {}, {}, alpha_value_ * color.a());
+    if (max_alpha == 0.0) {
+      color.update({}, {}, {}, alpha_value_);
+    } else {
+      color.update({}, {}, {}, alpha_value_ * color.a());
+    }
     meshcat_->SetProperty(path, "color",
                           {color.r(), color.g(), color.b(), color.a()});
   }
--- a/geometry/meshcat_visualizer_params.h
+++ b/geometry/meshcat_visualizer_params.h
@@ -58,21 +58,21 @@ struct MeshcatVisualizerParams {
   bool visible_by_default{true};
 
   /** When using the hydroelastic contact model, collision geometries that are
-  _declared_ as geometric primitives are frequently represented by some
-  discretely tessellated mesh when computing contact. It can be quite helpful
-  in assessing contact behavior to visualize these discrete meshes (in place of
-  the idealized primitives).
+   _declared_ as geometric primitives are frequently represented by some
+   discretely tessellated mesh when computing contact. It can be quite helpful
+   in assessing contact behavior to visualize these discrete meshes (in place
+   of the idealized primitives).
 
-  To visualize these representations it is necessary to request visualization
-  of geometries with the Role::kProximity role (see the role field). It is
-  further necessary to explicitly request the hydroelastic meshes where
-  available (setting show_hydroelastic to `true`).
+   To visualize these representations it is necessary to request visualization
+   of geometries with the Role::kProximity role (see the role field). It is
+   further necessary to explicitly request the hydroelastic meshes where
+   available (setting show_hydroelastic to `true`).
 
-  Setting this `show_hydroelastic` to `true` will have no apparent effect if
-  none of the collision meshes have a hydroelastic mesh associated with them.
+   Setting this `show_hydroelastic` to `true` will have no apparent effect if
+   none of the collision meshes have a hydroelastic mesh associated with them.
 
-  This option is ignored by MeshcatVisualizer<T> when T is not `double`, e.g.
-  if T == AutoDiffXd. */
+   This option is ignored by MeshcatVisualizer<T> when T is not `double`, e.g.
+   if T == AutoDiffXd. */
   bool show_hydroelastic{false};
 
   /** (Advanced) For a given geometry, if the GeometryProperties for our `role`
--- a/geometry/test/meshcat_visualizer_test.cc
+++ b/geometry/test/meshcat_visualizer_test.cc
@@ -321,23 +321,6 @@ TEST_F(MeshcatVisualizerWithIiwaTest, ScalarConversion) {
   ad_diagram->ForcedPublish(*ad_context);
 }
 
-TEST_F(MeshcatVisualizerWithIiwaTest, UpdateAlphaSliders) {
-  MeshcatVisualizerParams params;
-  params.enable_alpha_slider = true;
-  SetUpDiagram(params);
-  systems::Simulator<double> simulator(*diagram_);
-
-  // Simulate for a moment and publish to populate the visualizer.
-  simulator.AdvanceTo(0.1);
-  diagram_->ForcedPublish(*context_);
-
-  meshcat_->SetSliderValue("visualizer α", 0.5);
-
-  // Simulate and publish again to cause an update.
-  simulator.AdvanceTo(0.1);
-  diagram_->ForcedPublish(*context_);
-}
-
 // When opted-in by the user, we should display the hydroelastic tessellation
 // instead of the primitive shape.
 GTEST_TEST(MeshcatVisualizerTest, HydroGeometry) {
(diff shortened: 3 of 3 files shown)


trowell-tri (commented):
Toward #17662

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NTzYYYW9eiXhovZnO63:bv0cxzh)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

sherm1 (commented):
+@SeanCurtis-TRI for feature review, please (or reassign)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NTzlqDT6BC93sdkuD-V:bq7wl1w)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
:lgtm: co-feature, and platform.  I'll let Sean finish feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NTzpTSN77h7Ngtza4TW:b-ofmglp)* status: 7 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @trowell-tri)

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzm-stAq1L7cmpMWpS):*
nit This commit _also_ reflows a header's whitespace.  To help out anyone who runs `git blame` on the header and wonders what the hydroelastic config has to do with alpha blending, the commit message here needs to mention the sidecar change to the whitespace.
___
*[`geometry/test/meshcat_visualizer_test.cc` line 493 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzmU5s111bxJI9op_G:-NTzmU5t3ZS2wWJYKUIh:bflluou) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L493)):*
> ```cpp
> GTEST_TEST(MeshcatVisualizerTest, AlphaSliderResults) {
>   struct Scenario {
>     double geometry_alpha;
> ```

nit https://drake.mit.edu/styleguide/cppguide.html#Variable_and_Array_Initialization requires braces.

Hopefully the scenarios list still compiles with this change.  If the compiler hates it, then I suppose we can omit the `{}` with a comment mentioning the reason:

```c++
// We can't initialize this here, or else the list of scenarios below fails to compile.
double geometry_alpha;
```



_Suggestion:_
```cpp
double geometry_alpha{};
```
___
*[`geometry/test/meshcat_visualizer_test.cc` line 495 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzoSpc1I7yliwW62Ln:-NTzoSpd7eZIuOyg7HnL:b-y0jy0b) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L495)):*
> ```cpp
>     double geometry_alpha;
>     // The visualizer sets the minimum slider value to 0.02, so expected values
>     // for small slider values need to adjust for that fact.
> ```

I don't see any reason to unit test here what happens with a slider value set to 0.0.  The lower bound on the slider is orthogonal to the logic in `SetColorAlphas`, which is the component we're interested in testing here.  The test cases would be simplified and more clear if we just punted the `slider_value == 0.0` coverage.
___
*[`geometry/test/meshcat_visualizer_test.cc` line 518 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTznFWAA73Cc9EcPkm4:-NTznFWAA73Cc9EcPkm5:b32aji5) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L518)):*
> ```cpp
>     plant.Finalize();
> 
>     auto& inspector = scene_graph.model_inspector();
> ```

This stanza should have a comment explaining the overall work, since it's a bit of a wall of text currently.  We're no longer "Load a simple model with one geometry", we're something more like "Set the one geometry's alpha to scenario.geometry_alpha.".
___
*[`geometry/test/meshcat_visualizer_test.cc` line 536 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTznkYe1ATICYIczRcu:-NTznkYf8lfKuFyYXmDZ:b-tjdpvb) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L536)):*
> ```cpp
>     // Create the visualizer.
>     MeshcatVisualizerParams params;
>     params.include_unspecified_accepting = true;
> ```

nit This setting is already at its default value, and is not really relevant to this test so it's misleading to mention it explicitly.
___
*[`geometry/test/meshcat_visualizer_test.cc` line 547 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzo5AJEOMXvzN7-51o:-NTzo5AJEOMXvzN7-51p:b-ufuu1q) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L547)):*
> ```cpp
>     const std::string geom_path =
>         fmt::format("visualizer/box/box/{}", geom_id.get_value());
>     for (auto value_pair : scenario.slider_and_expected_values) {
> ```

nit Using `.first` and `.second` from a std::pair is nearly always a readability defect.  We can give them names in the for-loop.



_Suggestion:_
```cpp
const auto& [slider_value, expected_alpha] : scenario.slider_and_expected_values
```
___
*[`geometry/test/meshcat_visualizer_test.cc` line 564 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzp98QB-J9O42rTwDj:-NTzp98RA02lMY_a5AcW:bx4u1g1) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L564)):*
> ```cpp
>       auto data = oh.get().as<internal::SetPropertyData<std::vector<double>>>();
>       EXPECT_EQ(data.property, "color");
>       EXPECT_EQ(data.value.size(), 4);
> ```

nit If the size isn't four, we should stop before reading from index `[3]` (which might segfault).



_Suggestion:_
```cpp
ASSERT_EQ(data.value.size(), 4);
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)* status: 8 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI and @trowell-tri)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 496 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzpiFe7b6tDB_Styiy:-NTzpiFe7b6tDB_Styiz:b-7w1g68) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L496)):*
> ```cpp
>     // The visualizer sets the minimum slider value to 0.02, so expected values
>     // for small slider values need to adjust for that fact.
>     std::vector<std::pair<double, double>> slider_and_expected_values;
> ```

BTW I suspect that a `Scenario` with three distinct, well-named fields would read a lot better that the `pair` stuff.

I don't think the nested looping is really buying is anything.  Creating a few extra diagrams doesn't seem like it would hurt anything.

trowell-tri (commented):
I meant to mention this before, but `meshcat_visualizer_test.cc` also includes a test `UpdateAlphaSliders` which is strictly a subset of this test except that it uses `Simulator` and I wonder if that test should just be removed?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NTzvbFfDPoKyk93UzPG:b2eskzv)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 495 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzoSpc1I7yliwW62Ln:-NTztwVO-BD4XWsXpIyF:b-72hdw7) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L495)):*

I don't see any reason to unit test here what happens with a slider value set to 0.0.  The lower bound on the slider is orthogonal to the logic in `SetColorAlphas`, which is the component we're interested in testing here.  The test cases would be simplified and more clear if we just punted the `slider_value == 0.0` coverage.
</blockquote></details>

I'll remove it. I just felt that not checking the other endpoint might be leaving a trap for some future maintainer. "Why don't we test zero, that's where half the problems are?"
___
*[`geometry/test/meshcat_visualizer_test.cc` line 496 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzpiFe7b6tDB_Styiy:-NTzuDcL40wSnHe28Emv:bdxx4e2) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L496)):*

BTW I suspect that a `Scenario` with three distinct, well-named fields would read a lot better that the `pair` stuff.

I don't think the nested looping is really buying is anything.  Creating a few extra diagrams doesn't seem like it would hurt anything.
</blockquote></details>

I originally wrote it that way and expected complaints about the looping *there* so I switched it. :)

Is there a better idiom to use here, by the way?
___
*[`geometry/test/meshcat_visualizer_test.cc` line 547 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzo5AJEOMXvzN7-51o:-NTzvG2Y7G49ybOe9rHD:bat4yyo) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L547)):*

nit Using `.first` and `.second` from a std::pair is nearly always a readability defect.  We can give them names in the for-loop.
</blockquote></details>

Ooh that's way better. But I'm removing the pair anyway.

jwnimmer-tri (commented):
Let's do +@rpoyner-tri not -@SeanCurtis-TRI for co-feature review, since Rico did the original alpha sliders.  Sean is still welcome to weigh in from the peanut gallery of course.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU2vU5iCAyv3JCvv1a1:b-npjndh)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @trowell-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @trowell-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NU2pZuh601N-YbbfXf3:-NU2vlkNCLVYqRcXKy_a:b1avhas) (no related file):*

> I meant to mention this before, but meshcat_visualizer_test.cc also includes a test UpdateAlphaSliders which is strictly a subset of this test except that it uses Simulator and I wonder if that test should just be removed?

Working to check...
</blockquote></details>

It's pretty cheap in lines of code, and a nice no-crash acceptance test with the full system.  My suggestion is to co-locate the two test cases within the file, and then adjust their test names and overview comments to explain the difference... once is an acceptance test with an entire Simulator and a real robot; the other is a narrow functional test for the acute arithmetic.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri and @trowell-tri)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 496 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzpiFe7b6tDB_Styiy:-NU2w9kZDj7MnHPF_uMV:b-tbl2ct) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L496)):*
> Is there a better idiom to use here, by the way?

http://google.github.io/googletest/advanced.html#value-parameterized-tests but in this case it might not be worth it.

rpoyner-tri (commented):
:lgtm: feature.

BTW can I say that I hate that the "Participants" table status does not properly show if :lgtm: has been granted? If you comment anything after :lgtm: , your status goes to discussing, even if no discussions are still open.

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3os1sE4BhfuDTr82o:b-rqtbwt)* status: 2 unresolved discussions (waiting on @trowell-tri)

rpoyner-tri (commented):
At this late hour, probably +@SeanCurtis-TRI for platform review, Monday schedule.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3pWJe9i8FcAZzi6Bc:b-suglgi)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI and @trowell-tri)

trowell-tri (commented):
I believe that @jwnimmer-tri gave LGTM for platform? Or do I misunderstand?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3qFmjEJqqusonmmzc:b5dav5j)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

rpoyner-tri (commented):
Ah, I  see. He did claim platform. -@SeanCurtis-TRI 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3qlJRBDDHWrNP1Yr5:bnzoj0p)* status: 2 unresolved discussions (waiting on @trowell-tri)

rpoyner-tri (commented):
Sadly, the "am I a platform reviewer at all?" is a machine checkable predicate, while "am I platform reviewer for this PR?" is not.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3qv11C-CWzBXHtc6B:bodaiy)* status: 2 unresolved discussions (waiting on @trowell-tri)

trowell-tri (commented):
I'm sure LLMs will fix all this.

The latest update moves one test function, changes the scenario data table and removes some looping.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-:-NU3srXB-Tuogvq23K-5:bquz5dy)* status: 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 495 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzoSpc1I7yliwW62Ln:-NU3rtXZC2i0cyYqS8xB:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L495)):*

If you're worried about a trap, it would be OK to leave a comment breadcrumb here explaining it.  Something like "We don't test using alpha == 0.0 because that's outside the configured range of the slider".
</blockquote></details>

Done.
___
*[`geometry/test/meshcat_visualizer_test.cc` line 496 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19287#-NTzpiFe7b6tDB_Styiy:-NU3sOELDOA2LMU33v5i:b-5axmzo) ([raw file](https://github.com/RobotLocomotion/drake/blob/9ff4e150f55e7fb4b46193f3e6c422e0236f6b2a/geometry/test/meshcat_visualizer_test.cc#L496)):*

> Is there a better idiom to use here, by the way?

http://google.github.io/googletest/advanced.html#value-parameterized-tests but in this case it might not be worth it.
</blockquote></details>

Whoops, I missed the restructuring suggestion. Now the looping comment makes more sense and yes this does look a little better. Done.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),jwnimmer-tri(platform) (waiting on @trowell-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19287)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),jwnimmer-tri(platform) (waiting on @trowell-tri)

PR #19103, opened 2023-03-30
[WIP] Downgrade SNOPT to 7.2

aykut-tri:
(WIP; please don't review)

I’ve been working on downgrading SNOPT to 7.2. At this point, I got it into a working state for [a relaxed IK problem for iiwa](https://github.com/aykut-tri/drake/blob/downgrade_snopt/examples/kuka_iiwa_arm/iiwa_relaxed_ik.cc). However, it fails in tests, such as [the acrobot swing up](https://github.com/RobotLocomotion/drake/blob/master/examples/acrobot/test/run_swing_up_traj_optimization.cc).

Also, even for the working example, there are some memory issues (as shown in [this valgrind output](https://github.com/RobotLocomotion/drake/files/11114454/valgrind_iiwa_relaxed_ik.txt), during the LU decomposition of linear constraints) that I cannot find the root cause for. Something noteworthy is [the workspace memory resizing operations](https://github.com/RobotLocomotion/drake/compare/master...aykut-tri:downgrade_snopt?expand=1#diff-0c3815e5a67a9470653f46577afa91ff6fa611452496ef742a086ca5ce46bd35R1397-R1429) fail w/o [this change for increasing the initial size](https://github.com/RobotLocomotion/drake/compare/master...aykut-tri:downgrade_snopt?expand=1#diff-0c3815e5a67a9470653f46577afa91ff6fa611452496ef742a086ca5ce46bd35R388-R390) that was copied [from the adapter version compatible with 7.2](https://github.com/RobotLocomotion/drake/blob/4a7615d1c08c5c3bab7a9db7dbd94c15618073af/solvers/snopt_solver.cc#L165).

Any help would be greatly appreciated, and I'd be happy to provide further details.

---

Relates to #17609.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19103)

PR #19296, opened 2023-04-27
[pydrake] Handle nulls in ImportError reporting

jwnimmer-tri:
The fields of an ImportError might be None; don't crash with an even more obscure error in that case.

Noticed during https://github.com/RobotLocomotion/drake/pull/19103.

My cost / benefit on adding a regression test for this was in favor for punting.  We regression test the non-None case already; injecting the None seemed difficult.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19296)

### DIFF
--- a/bindings/pydrake/__init__.py
+++ b/bindings/pydrake/__init__.py
@@ -41,7 +41,8 @@ except ImportError:
 try:
     from . import common
 except ImportError as e:
-    if '/pydrake/' in e.path and 'cannot open shared object file' in e.msg:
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
+@sammy-tri for both reviews per schedule, please.

jwnimmer-tri:
@drake-jenkins-bot mac-arm-monterey-unprovisioned-clang-wheel-experimental-release please

PR #19301, opened 2023-04-28
[planning] Adds a time cost to GcsTrajectoryOptimization

wrangelvid:
This adds a minimum time objective. Rather than finding the shortest path, this allows users to find the fastest path. It becomes particularly useful once velocity bounds will be introduced.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)

### DIFF
--- a/bindings/pydrake/planning/planning_py_trajectory_optimization.cc
+++ b/bindings/pydrake/planning/planning_py_trajectory_optimization.cc
@@ -376,6 +376,8 @@ void DefinePlanningTrajectoryOptimization(py::module m) {
               return out;
             },
             subgraph_doc.regions.doc)
+        .def("AddTimeCost", &Class::Subgraph::AddTimeCost,
+            py::arg("weight") = 1.0, subgraph_doc.AddTimeCost.doc)
         .def("AddPathLengthCost",
             py::overload_cast<const Eigen::MatrixXd&>(
                 &Class::Subgraph::AddPathLengthCost),
@@ -416,6 +418,8 @@ void DefinePlanningTrajectoryOptimization(py::module m) {
         .def("AddEdges", &Class::AddEdges, py_rvp::reference_internal,
             py::arg("from_subgraph"), py::arg("to_subgraph"),
             py::arg("subspace") = py::none(), cls_doc.AddEdges.doc)
+        .def("AddTimeCost", &Class::AddTimeCost, py::arg("weight") = 1.0,
+            cls_doc.AddTimeCost.doc)
         .def("AddPathLengthCost",
             py::overload_cast<const Eigen::MatrixXd&>(
                 &Class::AddPathLengthCost),
--- a/bindings/pydrake/planning/test/trajectory_optimization_test.py
+++ b/bindings/pydrake/planning/test/trajectory_optimization_test.py
@@ -320,13 +320,19 @@ class TestTrajectoryOptimization(unittest.TestCase):
             np.array([[5.0, 5.0, 4.4, 4.4], [2.8, 5.0, 5.0, 2.8]])
         ]
 
-        # We add a the path length cost to the entire graph.
+        # We add a path length cost to the entire graph.
         # This can be called ahead of time or after adding the regions.
         gcs.AddPathLengthCost(weight=1.0)
         # This cost is equivalent to the above.
         # It will be added twice, which is unnecessary,
         # but we do it to test the binding.
         gcs.AddPathLengthCost(weight_matrix=np.eye(dimension))
+
+        # Add a mimimum time cost to the entire graph.
+        gcs.AddTimeCost(weight=1.0)
+        # Add the cost again, which is unnecessary for the optimization
+        # but useful to check the binding with the default values.
+        gcs.AddTimeCost()
 
         # Add two subgraphs with different orders.
         main1 = gcs.AddRegions(
@@ -415,6 +421,14 @@ class TestTrajectoryOptimization(unittest.TestCase):
         # This weight matrix penalizes movement in the y direction three
         # times more than in the x direction only for the main2 subgraph.
         main2.AddPathLengthCost(weight_matrix=np.diag([1.0, 3.0]))
+
+        # Adding this cost checks the python binding. It won't contribute to
+        # the solution since we already added the minimum time cost to the
+        # whole graph.
+        main2.AddTimeCost(weight=1.0)
+        # Add the cost again, which is unnecessary for the optimization
+        # but useful to check the binding with the default values.
+        main2.AddTimeCost()
 
         options = GraphOfConvexSetsOptions()
         options.convex_relaxation = True
--- a/planning/trajectory_optimization/gcs_trajectory_optimization.cc
+++ b/planning/trajectory_optimization/gcs_trajectory_optimization.cc
@@ -32,6 +32,7 @@ using solvers::Binding;
 using solvers::Constraint;
 using solvers::Cost;
 using solvers::L2NormCost;
+using solvers::LinearCost;
 using solvers::LinearEqualityConstraint;
 using symbolic::DecomposeLinearExpressions;
 using symbolic::Expression;
@@ -141,6 +142,17 @@ Subgraph::Subgraph(
 }
 
 Subgraph::~Subgraph() = default;
+
+void Subgraph::AddTimeCost(double weight) {
+  // The time cost is the sum of duration variables ∑ hᵢ
+  auto time_cost =
+      std::make_shared<LinearCost>(weight * Eigen::VectorXd::Ones(1), 0.0);
+
+  for (Vertex* v : vertices_) {
+    // The duration variable is the last element of the vertex.
+    v->AddCost(Binding<LinearCost>(time_cost, v->x().tail(1)));
+  }
+}
 
 void Subgraph::AddPathLengthCost(const MatrixXd& weight_matrix) {
   /*
@@ -332,6 +344,10 @@ Subgraph& GcsTrajectoryOptimization::AddRegions(
                                     h_min, h_max, std::move(name), this);
 
   // Add global costs to the subgraph.
+  for (double weight : global_time_costs_) {
+    subgraph->AddTimeCost(weight);
+  }
+
   if (order > 0) {
     // These costs rely on the derivative of the trajectory.
     for (const MatrixXd& weight_matrix : global_path_length_costs_) {
@@ -369,6 +385,14 @@ EdgesBetweenSubgraphs& GcsTrajectoryOptimization::AddEdges(
     const ConvexSet* subspace) {
   return *subgraph_edges_.emplace_back(
       new EdgesBetweenSubgraphs(from_subgraph, to_subgraph, subspace, this));
+}
+
+void GcsTrajectoryOptimization::AddTimeCost(double weight) {
+  // Add time cost to each subgraph.
+  for (std::unique_ptr<Subgraph>& subgraph : subgraphs_) {
+    subgraph->AddTimeCost(weight);
+  }
+  global_time_costs_.push_back(weight);
 }
 
 void GcsTrajectoryOptimization::AddPathLengthCost(
--- a/planning/trajectory_optimization/gcs_trajectory_optimization.h
+++ b/planning/trajectory_optimization/gcs_trajectory_optimization.h
@@ -80,6 +80,12 @@ class GcsTrajectoryOptimization final {
       return regions_;
     }
 
+    /** Adds a minimum time cost to all regions in the subgraph. The cost is the
+    sum of the time scaling variables.
+    @param weight is the relative weight of the cost.
+    */
+    void AddTimeCost(double weight = 1.0);
+
     /** Adds multiple L2Norm Costs on the upper bound of the path length.
     Since we cannot directly compute the path length of a Bézier curve, we
     minimize the upper bound of the path integral by minimizing the sum of
@@ -262,6 +268,16 @@ class GcsTrajectoryOptimization final {
       const Subgraph& from_subgraph, const Subgraph& to_subgraph,
       const geometry::optimization::ConvexSet* subspace = nullptr);
 
+  /** Adds a minimum time cost to all regions in the whole graph. The cost is
+  the sum of the time scaling variables.
+
+  This cost will be added to the entire graph. Note that this cost
+  will be applied even to subgraphs added in the future.
+
+  @param weight is the relative weight of the cost.
+  */
+  void AddTimeCost(double weight = 1.0);
+
   /** Adds multiple L2Norm Costs on the upper bound of the path length.
   Since we cannot directly compute the path length of a Bézier curve, we
   minimize the upper bound of the path integral by minimizing the sum of
@@ -337,6 +353,7 @@ class GcsTrajectoryOptimization final {
   // Store the subgraphs by reference.
   std::vector<std::unique_ptr<Subgraph>> subgraphs_;
   std::vector<std::unique_ptr<EdgesBetweenSubgraphs>> subgraph_edges_;
+  std::vector<double> global_time_costs_;
   std::vector<Eigen::MatrixXd> global_path_length_costs_;
 };
 
--- a/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc
+++ b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc
@@ -57,6 +57,9 @@ GTEST_TEST(GcsTrajectoryOptimizationTest, Basic) {
   EXPECT_TRUE(CompareMatrices(traj.value(traj.end_time()), goal, 1e-6));
 }
 
+// TODO(wrangelvid) Add a dedicated test comparing time cost to path length
+// cost. Velocity bounds are required for that.
+
 GTEST_TEST(GcsTrajectoryOptimizationTest, InvalidPositions) {
   /* Positions passed into GcsTrajectoryOptimization must be greater than 0.*/
   DRAKE_EXPECT_THROWS_MESSAGE(GcsTrajectoryOptimization(0),
@@ -416,6 +419,7 @@ TEST_F(SimpleEnv2D, MultiStartGoal) {
   gcs.AddEdges(regions, target);
 
   gcs.AddPathLengthCost();
+  gcs.AddTimeCost();
 
   if (!GurobiOrMosekSolverAvailable()) {
     return;
@@ -507,6 +511,7 @@ TEST_F(SimpleEnv2D, IntermediatePoint) {
 
   // We can add different costs to the individual subgraphs.
   main1.AddPathLengthCost(5);
+  main1.AddTimeCost(1);
 
   // This weight matrix penalizes movement in the y direction three times more
   // than in the x direction.


RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 2 unresolved discussions, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 272 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6YnKo3z-mWRoxp5rZ:-NU6YnKo3z-mWRoxp5r_:b-3ce9a1) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/gcs_trajectory_optimization.h#L272)):*
> ```objc
> 
>   /** Adds a minimum time cost to all regions in the whole graph. The cost is
>   the sum of the time scaling variables.
> ```

Better say again here "Note that this cost will be applied even to subgraphs added in the future."
___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NU6ZDRC7IQK1pX0HN1s:brove9s) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*
> ```cpp
>   // We can add different costs to the individual subgraphs.
>   main1.AddPathLengthCost(5);
>   main1.AddTimeCost(1);
> ```

somewhere in this file we should have a test that establishes the correctness of the time cost implementation (not just coverage).  e.g. a new test with a  simple minimum time GCStrajopt instance where you know the right answer.

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 2 unresolved discussions, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 272 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6YnKo3z-mWRoxp5rZ:-NU7QdBp0MWQopu9I45O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/gcs_trajectory_optimization.h#L272)):*

Better say again here "Note that this cost will be applied even to subgraphs added in the future."
</blockquote></details>

Done.
___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NU7OxHC8eBnuetulqU1:b-vqss5p) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

somewhere in this file we should have a test that establishes the correctness of the time cost implementation (not just coverage).  e.g. a new test with a  simple minimum time GCStrajopt instance where you know the right answer.  
</blockquote></details>

I was hoping to wait for my next PR that adds the velocity bounds to properly test it.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 1 unresolved discussion, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @wrangelvid)

___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NU9SEWa9NYlX376hSSb:b-9ca0dn) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

I was hoping to wait for my next PR that adds the velocity bounds to properly test it.
</blockquote></details>

I see.  That would be a good reason to PR them both at the same time.  But if you want to do just this one, at least add a TODO to say that the test is coming.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 1 unresolved discussion, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @wrangelvid)

___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NU9STgK8TNfxEEZxETR:bgp0vpb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

I see.  That would be a good reason to PR them both at the same time.  But if you want to do just this one, at least add a TODO to say that the test is coming.
</blockquote></details>

I'd say go ahead and add it to this PR, if it's ready?

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 1 unresolved discussion, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @wrangelvid)

___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUEYh3YBOAuiqLkU4Yc:boroc80) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

I'd say go ahead and add it to this PR, if it's ready?
</blockquote></details>

Done. I would rather PR often and keep them small in size.

RussTedrake (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-:-NUGz8-n7096S-YpHoWW:bnfp4nl)* status: 1 unresolved discussion, needs at least two assigned reviewers, missing label for release notes (waiting on @wrangelvid)

___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUGylqi59nhEkAzkVrM:b9jks12) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

Done. I would rather PR often and keep them small in size.
</blockquote></details>

That philosophy is exactly right, but the minimal viable element is one that permits testing of correctness.

We're likely to get the platform review tomorrow (Monday), so if you want to put in the velocity constraints today in this PR, then I can review and they will go through at the same time.  But I'm aok with this as is.

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-:-NUGzCTc-P1G0IP4XIL3:b-6cxuoj)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @SeanCurtis-TRI and @wrangelvid)

SeanCurtis-TRI (commented):
> Reviewed 3 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @wrangelvid)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMkmWaAc5LEzNHSt48):*
nit: probably better to replace this commit message with the PR summary. It is far more expressive. Note: This can be done at the time you merge of your next pass on tweaks based comments. (The former is slightly preferred because it can be reviewed and is less likely to be forgotten at the last second.)
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 85 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMl16zAASaOg-2L-TA:-NUMl17-5EEPSg1dNDNV:bp4nchu) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.h#L85)):*
> ```objc
>     /** Adds a minimum time cost to all regions in the subgraph. The cost is the
>     sum of the time scaling variables.
>     @param weight is the relative weight of the cost.
> ```

btw: Generally it's better to think of these doxygen directives as a definition. E.g., "foo: the thing that foos" rather than a sentence ("foo is the thing that foos").

I wouldn't bother addressing it in this PR as this sentence-centric style appears throughout this file and, therefore, the language in this PR is consistent with this file.

This did inspire me to take a look at Drake generally and there are 128 instances of this pattern (compared to a further 2800 that use the definition-like pattern). I'll probably do a clean up PR in the near future. In the meantime, you might bear it in mind.
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 151 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMnuR741aLTZC-tBq-:-NUMnuR741aLTZC-tBq0:b-o6javu) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L151)):*
> ```cpp
>       std::make_shared<LinearCost>(weight * Eigen::VectorXd::Ones(1), 0.0);
> 
>   for (Vertex* v : vertices_) {
> ```

nit: As long as you don't need to modify it, you should make the code enforce that.



_Suggestion:_
```cpp
for (const Vertex* v : vertices_) {
```
___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUMp2Lg3oPxfyk1k54X:b-7znyub) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

That philosophy is exactly right, but the minimal viable element is one that permits testing of correctness.

We're likely to get the platform review tomorrow (Monday), so if you want to put in the velocity constraints today in this PR, then I can review and they will go through at the same time.  But I'm aok with this as is.
</blockquote></details>

I see the TODO added in the most recent revision. I'm content.

Although, I'd have loved to seen the test where  the optimizer achieves a path with zero duration because it used infinite velocity to teleport across each region.  :)
___
*[`bindings/pydrake/planning/test/trajectory_optimization_test.py` line 332 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMpOxpCoQSiN94gzxg:-NUMpOxpCoQSiN94gzxh:b-yi3kf0) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/bindings/pydrake/planning/test/trajectory_optimization_test.py#L332)):*
> ```Python
> 
>         # Add a mimimum time cost to the entire graph.
>         gcs.AddTimeCost(weight=1.0)
> ```

nit: A property of the *bindings* is that `weight` has been bound with a default value. As test coverage for that, it would be appropriate to call `gcs.AddTimeCost()` to make sure that is a valid invocation. (Same for the test below.)

SeanCurtis-TRI (commented):
:LGTM: With the nits indicated below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-:-NUNdQAMA2cPJb42VDvx:bobxvy6)* status: 5 unresolved discussions, missing label for release notes (waiting on @wrangelvid)

SeanCurtis-TRI (commented):
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-:-NUNdVE66gPd0ZBX1xGp:b-l8xbmn)* status: 5 unresolved discussions

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMkmWaAc5LEzNHSt48):*

nit: probably better to replace this commit message with the PR summary. It is far more expressive. Note: This can be done at the time you merge of your next pass on tweaks based comments. (The former is slightly preferred because it can be reviewed and is less likely to be forgotten at the last second.)
</blockquote></details>

Done.
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 85 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMl16zAASaOg-2L-TA:-NUPYnW3BenKR07Wkfkx:b-ze3amd) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.h#L85)):*

btw: Generally it's better to think of these doxygen directives as a definition. E.g., "foo: the thing that foos" rather than a sentence ("foo is the thing that foos").

I wouldn't bother addressing it in this PR as this sentence-centric style appears throughout this file and, therefore, the language in this PR is consistent with this file.

This did inspire me to take a look at Drake generally and there are 128 instances of this pattern (compared to a further 2800 that use the definition-like pattern). I'll probably do a clean up PR in the near future. In the meantime, you might bear it in mind.
</blockquote></details>

Thanks for bringing this up. Is that referenced in the drake style guide? I based most of this files stile of other files I found in the repository. I will follow up with a clean-up PR
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 151 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMnuR741aLTZC-tBq-:-NUPZOkB2y8IPZkIyYP_:bs7wumb) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L151)):*

nit: As long as you don't need to modify it, you should make the code enforce that.
</blockquote></details>

Done.
___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUPZQbo32uxeoShgQY5:b3tn1bo) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

I see the TODO added in the most recent revision. I'm content.

Although, I'd have loved to seen the test where  the optimizer achieves a path with zero duration because it used infinite velocity to teleport across each region.  :)
</blockquote></details>

That's what I wanted to do too but I decided against it since adding the time cost without any velocity bounds is meaningless.

Case 1: No velocity bounds, time cost, path length cost:
The duration scaling is fully decoupled from the rest of the problem so the time cost will just turn down the duration scaling to the minimum value and not effect the trajectory shape nor the path in the GCS graph. In practice I found that with MOSEK and the GCS framework the duration variables will be set to the minimum regardless of adding the time cost. 

Case 2: No velocity bounds, time cost:
This one is funny. When is set to be greater than 0 `h_min > 0`, all the time variables in the vertices will be minimized to h_min. Thus the solution that minimizes the total trajectory time $\sum_i h_i$, will contain the minimum number of edges. Without the velocity bounds, the trajectory in the sets might look strange since infinite speed allows them to take any shape. 

Thus, I find it more meaningful to include the velocity bounds and construct a test case where the minimum length objective will yield a different path in the graph.
___
*[`bindings/pydrake/planning/test/trajectory_optimization_test.py` line 332 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMpOxpCoQSiN94gzxg:-NUPbRl86b64KVFg1BT2:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/bindings/pydrake/planning/test/trajectory_optimization_test.py#L332)):*

nit: A property of the *bindings* is that `weight` has been bound with a default value. As test coverage for that, it would be appropriate to call `gcs.AddTimeCost()` to make sure that is a valid invocation. (Same for the test below.)
</blockquote></details>

Done.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19301)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMkmWaAc5LEzNHSt48):*

Done.
</blockquote></details>

FYI Some notes on [git common practice](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53).

- The first sentence should be a succinct summary of what effect the PR has (not neccessarily *how* it does it).
    - In Drake release notes, those summaries become the one line reports and we prefer to use the imperative mood (see [some release notes](https://drake.mit.edu/release_notes/v1.15.0.html) for an example). In this case, your PR title is almost perfect (I simply changed "Adds" to "Add").
- An empty line follows the summary and any interesting implementation details or observations can be included as you see fit. 
 - All lines should have hard breaks.

These conventions make working with git commits far simpler. In that spirit, I'd suggest the following commit message (note the hard line breaks):

```bash
Add a time cost to GcsTrajectoryOptimization

This adds a minimum time objective. Rather than finding the shortest path,
this allows users to find the fastest path. It becomes particularly useful
once velocity bounds will be introduced.
```

We can always patch it when we merge.
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 85 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMl16zAASaOg-2L-TA:-NUS2uLh96_-P1vVgYsH:bum9wdw) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.h#L85)):*

Thanks for bringing this up. Is that referenced in the drake style guide? I based most of this files stile of other files I found in the repository. I will follow up with a clean-up PR
</blockquote></details>

Don't worry about that. Like I said, there are enough instances outside of this file, that I'll do a clean up of all of them at the same time.
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 151 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NUMnuR741aLTZC-tBq-:-NUS4TupD6rv4U7bf9aM:b-973o2h) ([raw file](https://github.com/RobotLocomotion/drake/blob/8c2522fcce1c99a1deb489d9d1e14808ef6bf7fc/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L151)):*

Done.
</blockquote></details>

I owe you a *huge* apology. This *cannot* be `const`. I focused on the `v->x().tail(1)` stuff and thought "that's const access". I *completely* ignored the fact that we're calling `v->AddCost(...)` which is a __*non*__ const method. I introduced your CI problems. Again, apologies for misleading you.
___
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUS3D2p8jcwcaA9IQXR:b-wer3ve) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

That's what I wanted to do too but I decided against it since adding the time cost without any velocity bounds is meaningless.

Case 1: No velocity bounds, time cost, path length cost:
The duration scaling is fully decoupled from the rest of the problem so the time cost will just turn down the duration scaling to the minimum value and not effect the trajectory shape nor the path in the GCS graph. In practice I found that with MOSEK and the GCS framework the duration variables will be set to the minimum regardless of adding the time cost. 

Case 2: No velocity bounds, time cost:
This one is funny. When is set to be greater than 0 `h_min > 0`, all the time variables in the vertices will be minimized to h_min. Thus the solution that minimizes the total trajectory time $\sum_i h_i$, will contain the minimum number of edges. Without the velocity bounds, the trajectory in the sets might look strange since infinite speed allows them to take any shape. 

Thus, I find it more meaningful to include the velocity bounds and construct a test case where the minimum length objective will yield a different path in the graph.

</blockquote></details>

I used to do pedestrian simulation and this is similar to the classic canard when people assert human trajectories optimize for time -- can't be true, otherwise people would be sprinting everywhere.  :)

SeanCurtis-TRI:
*[`planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc` line 511 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19301#-NU6ZDRC7IQK1pX0HN1r:-NUSbW-x1ntL6RUkR-qg:b-5s3wxl) ([raw file](https://github.com/RobotLocomotion/drake/blob/3632462d89aa93488532c254953fdb366d97eb2b/planning/trajectory_optimization/test/gcs_trajectory_optimization_test.cc#L511)):*

Well, here is an example: Let's say I want to go from my dorm room to our lab in Stata. If I walk, I am speed limited by 1.5 m/s. There is a direct walking path. I could also take my e-scooter that drivers 13 m/s, however I can't traverse stairs with it so I need to take a detour. Using GCSTrajectoryOptimization, we would build a subgraph for each mode and enforce different velocity bounds. The shortest path objective would tell me to walk, while the fastest path objective would recommend to take the scooter.
</blockquote></details>

A good example. One of the papers I worked looked at an energy-optimization basis for trajectory planning. Humans have a sweet spot for energy per unit distance traveled. Unsurprisingly, the most optimal walking speed (in that sense) aligns well with *average* walking speed -- assuming that one doesn't have a pathological gait. It fuels my belief that humans are natural optimizers. Even when it seems we aren't, we are. We're just bad at it or picking horrible cost functions.  :)

PR #19311, opened 2023-04-29
Fix ConvexSet move operations

### LINKED ISSUE #19310: Invalid move implementations for ConvexSet subclasses
Reading the code, I'm pretty sure that a moved-from ConvexSet subclass does not meet the "valid but indeterminate state" requirement.  The subclass's fields will be empty (e.g., the Point's `x_`), but the superclass does not annotate `reset_after_move` on the ambient dimension. The `cloner_` is probably null as well.

jwnimmer-tri:
Requires #19321.  Closes #19310.  Towards #19279.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)

### DIFF
--- a/geometry/optimization/convex_set.h
+++ b/geometry/optimization/convex_set.h
@@ -7,6 +7,7 @@
 
 #include "drake/common/copyable_unique_ptr.h"
 #include "drake/common/drake_assert.h"
+#include "drake/common/reset_after_move.h"
 #include "drake/geometry/geometry_ids.h"
 #include "drake/geometry/query_object.h"
 #include "drake/geometry/shape_specification.h"
@@ -180,7 +181,9 @@ class ConvexSet : public ShapeReifier {
   /** Implements non-virtual base class serialization. */
   template <typename Archive>
   void Serialize(Archive* a) {
-    a->Visit(MakeNameValue("ambient_dimension", &ambient_dimension_));
+    // Visit the mutable reference inside the reset_after_move wrapper.
+    int& ambient_dimension = ambient_dimension_;
+    a->Visit(DRAKE_NVP(ambient_dimension));
   }
 
   /** Non-virtual interface implementation for Clone(). */
@@ -238,7 +241,21 @@ class ConvexSet : public ShapeReifier {
   DoToShapeWithPose() const = 0;
 
  private:
-  int ambient_dimension_{0};
+  // The reset_after_move wrapper adjusts ConvexSet's default move constructor
+  // and move assignment operator to set the ambient dimension of a moved-from
+  // object back to zero. This is essential to keep the ambient dimension in
+  // sync with any moved-from member fields in concrete ConvexSet subclasses.
+  //
+  // For example, the `Eigen::VectorXd x_` member field of `Point` will be moved
+  // out, ending up with `x_.size() == 0` on the moved-from Point. To maintain
+  // the invariant that `x_.size() == ambient_dimension_`, we need to zero the
+  // dimension when `x_` is moved-from.
+  //
+  // Similarly, for subclasses that are composite sets (e.g., CartesianProduct)
+  // the `ConvexSets sets_` vector becomes empty when moved-from. To maintain
+  // an invariant like `∑(set.size() for set in sets_) == ambient_dimension_`,
+  // we need to zero the dimension when `sets_` is moved-from.
+  reset_after_move<int> ambient_dimension_;
 };
 
 /** Provides the recommended container for passing a collection of ConvexSet
--- a/geometry/optimization/test/cartesian_product_test.cc
+++ b/geometry/optimization/test/cartesian_product_test.cc
@@ -66,6 +66,23 @@ GTEST_TEST(CartesianProductTest, DefaultCtor) {
   EXPECT_FALSE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
 }
 
+GTEST_TEST(CartesianProductTest, Move) {
+  const Point P1(Vector2d{1.2, 3.4}), P2(Vector2d{5.6, 7.8});
+  CartesianProduct orig(P1, P2);
+
+  // A move-constructed CartesianProduct takes over the original data.
+  CartesianProduct dut(std::move(orig));
+  EXPECT_EQ(dut.num_factors(), 2);
+  EXPECT_EQ(dut.ambient_dimension(), 4);
+
+  // The old set is in a valid but unspecified state. For convenience we'll
+  // assert that it's empty, but that's not the only valid implementation,
+  // just the one we happen to currently use.
+  EXPECT_EQ(orig.num_factors(), 0);
+  EXPECT_EQ(orig.ambient_dimension(), 0);
+  EXPECT_NO_THROW(orig.Clone());
+}
+
 GTEST_TEST(CartesianProductTest, FromSceneGraph) {
   const RigidTransformd X_WG{math::RollPitchYawd(.1, .2, 3),
                              Vector3d{.5, .87, .1}};
--- a/geometry/optimization/test/hpolyhedron_test.cc
+++ b/geometry/optimization/test/hpolyhedron_test.cc
@@ -89,6 +89,24 @@ GTEST_TEST(HPolyhedronTest, UnitBoxTest) {
   EXPECT_TRUE(CompareMatrices(b, H_scene_graph.b()));
 }
 
+GTEST_TEST(HPolyhedronTest, Move) {
+  Matrix<double, 6, 3> A;
+  A << Matrix3d::Identity(), -Matrix3d::Identity();
+  Vector6d b = Vector6d::Ones();
+  HPolyhedron orig(A, b);
+
+  // A move-constructed HPolyhedron takes over the original data.
+  HPolyhedron dut(std::move(orig));
+  EXPECT_EQ(dut.ambient_dimension(), 3);
+  EXPECT_TRUE(CompareMatrices(dut.A(), A));
+  EXPECT_TRUE(CompareMatrices(dut.b(), b));
+
+  // The old HPolyhedron is in a valid but unspecified state.
+  EXPECT_EQ(orig.A().cols(), orig.ambient_dimension());
+  EXPECT_EQ(orig.b().size(), orig.ambient_dimension());
+  EXPECT_NO_THROW(orig.Clone());
+}
+
 GTEST_TEST(HPolyhedronTest, ConstructorFromVPolytope) {
   Eigen::Matrix<double, 3, 4> vert1;
   // clang-format off
--- a/geometry/optimization/test/hyperellipsoid_test.cc
+++ b/geometry/optimization/test/hyperellipsoid_test.cc
@@ -98,6 +98,23 @@ GTEST_TEST(HyperellipsoidTest, DefaultCtor) {
   EXPECT_FALSE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
 }
 
+GTEST_TEST(HyperellipsoidTest, Move) {
+  const Eigen::Matrix3d A = Eigen::Matrix3d::Identity();
+  const Vector3d center = Vector3d::Zero();
+  Hyperellipsoid orig(A, center);
+
+  // A move-constructed Hyperellipsoid takes over the original data.
+  Hyperellipsoid dut(std::move(orig));
+  EXPECT_EQ(dut.ambient_dimension(), 3);
+  EXPECT_TRUE(CompareMatrices(dut.A(), A));
+  EXPECT_TRUE(CompareMatrices(dut.center(), center));
+
+  // The old Hyperellipsoid is in a valid but unspecified state.
+  EXPECT_EQ(orig.A().cols(), orig.ambient_dimension());
+  EXPECT_EQ(orig.center().size(), orig.ambient_dimension());
+  EXPECT_NO_THROW(orig.Clone());
+}
+
 GTEST_TEST(HyperellipsoidTest, ScaledSphereTest) {
   const double radius = 0.1;
   auto [scene_graph, geom_id] =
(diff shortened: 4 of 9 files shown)


jwnimmer-tri (commented):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri:
\CC @RussTedrake FYI, or in case you'd like to feature-review.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

jwnimmer-tri:
On second look, there are still several lingering bugs related to zero-dimensional set handling.  I've converted back to "Draft" while I work on those.

jwnimmer-tri (commented):
> Reviewed 26 of 26 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge"

___
*[`geometry/optimization/convex_set.cc` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NUN0yTy1XAZmnst4ity:-NUN0yTz64J_FmNJZ78g:bh9huxj) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fc07f737dfd54e6c402424dc89bde52f7ac5f86/geometry/optimization/convex_set.cc#L47)):*
> ```cpp
>     const Eigen::Ref<const solvers::VectorXDecisionVariable>& x,
>     const symbolic::Variable& t) const {
>   DRAKE_THROW_UNLESS(ambient_dimension() > 0);
> ```

Working

Some unit tests are hitting this.

Either they are incorrectly formulated (in which case we should fix the test), or they really do need to add this vacuously false  constraint, in which case we need to figure out how to add a vacuously false constraint to MathematicalProgram without it throwing an exception.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge"

___
*[`geometry/optimization/convex_set.cc` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NUN0yTy1XAZmnst4ity:-NUNkY9q3dKyINCQCfZZ:brc3fxy) ([raw file](https://github.com/RobotLocomotion/drake/blob/2fc07f737dfd54e6c402424dc89bde52f7ac5f86/geometry/optimization/convex_set.cc#L47)):*

Working

Some unit tests are hitting this.

Either they are incorrectly formulated (in which case we should fix the test), or they really do need to add this vacuously false  constraint, in which case we need to figure out how to add a vacuously false constraint to MathematicalProgram without it throwing an exception.
</blockquote></details>

Actually, moving this discussion to #19321.

jwnimmer-tri (commented):
> Reviewed 3 of 25 files at r7, 29 of 29 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @RussTedrake)

RussTedrake:
*[`geometry/optimization/convex_set.h` line 244 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXKkX5i5-UO3KkUL8vi:-NXKkX5i5-UO3KkUL8vj:bfobm3s) ([raw file](https://github.com/RobotLocomotion/drake/blob/ba3478aaa779190e1e73c5b82df4fe605d065ef8/geometry/optimization/convex_set.h#L244)):*
> ```objc
> 
>  private:
>   reset_after_move<int> ambient_dimension_;
> ```

in your comments about threading, you said "readability/maintainability of the code may trump other concerns".  In my view, you are making something that should be simple look very complicated.  (here an in the lines above). I think it's unlikely that developers will get this right out of the box.  I've just now read the reset_after_move docs (i was not aware)... but don't know if that is enough for me to justify this complexity?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: 1 unresolved discussion, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers

___
*[`geometry/optimization/convex_set.h` line 244 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXKkX5i5-UO3KkUL8vi:-NXL42NG4XvBTieO6wPE:b-jomc6j) ([raw file](https://github.com/RobotLocomotion/drake/blob/ba3478aaa779190e1e73c5b82df4fe605d065ef8/geometry/optimization/convex_set.h#L244)):*

in your comments about threading, you said "readability/maintainability of the code may trump other concerns".  In my view, you are making something that should be simple look very complicated.  (here an in the lines above). I think it's unlikely that developers will get this right out of the box.  I've just now read the reset_after_move docs (i was not aware)... but don't know if that is enough for me to justify this complexity?
</blockquote></details>

It didn't read as complicated at all to me, but let's run with the idea anyway.

If we assume that this is too difficult as-is, I'm not sure where you're going with the thought process.  Maybe we could add more documentation nearby here, to better explain?  Something like:

```c++
  // When this object is moved from (via the move constructor or move
  // assignment operator), the reset_after_move wrapper will reset the
  // ambient dimension to zero. This is important because subclasses
  // that are composite sets (e.g., MinkowskiSum) will have their
  // child sets moved away, leaving them with no dimensions to report.
```

The bottom line is that the move constructors are all broken on master.  Unless we have a fix that ends up with this same effect (however we spell it), the only other alternative is to remove (`= delete`) all of the move operators on `ConvexSet` and all of its subclasses, leaving only the copy operators.  Is that a better fix than this approach?

RussTedrake (commented):
:lgtm:

> Reviewed 3 of 25 files at r7, 29 of 29 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-:-NXLMcKf5TMhUsoJ5tVx:bg7igk9)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/convex_set.h` line 244 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXKkX5i5-UO3KkUL8vi:-NXLIV1z55xcEe685_VW:bp3ao57) ([raw file](https://github.com/RobotLocomotion/drake/blob/ba3478aaa779190e1e73c5b82df4fe605d065ef8/geometry/optimization/convex_set.h#L244)):*

It didn't read as complicated at all to me, but let's run with the idea anyway.

If we assume that this is too difficult as-is, I'm not sure where you're going with the thought process.  Maybe we could add more documentation nearby here, to better explain?  Something like:

```c++
  // When this object is moved from (via the move constructor or move
  // assignment operator), the reset_after_move wrapper will reset the
  // ambient dimension to zero. This is important because subclasses
  // that are composite sets (e.g., MinkowskiSum) will have their
  // child sets moved away, leaving them with no dimensions to report.
```

The bottom line is that the move constructors are all broken on master.  Unless we have a fix that ends up with this same effect (however we spell it), the only other alternative is to remove (`= delete`) all of the move operators on `ConvexSet` and all of its subclasses, leaving only the copy operators.  Is that a better fix than this approach?
</blockquote></details>

The last line of the comment helps me understand the goal.  I initially failed to see the danger in having the ambient dimension have a non-zero value sticking around.

It's complicated because it suddenly asks the develops to understand a lot more about move semantics, etc.  And one has to go and read the reset_after_move header file to learn more.  We have important contributors to the ConvexSet classes who are still coming up to speed on the more nuanced aspects of C++, and I don't want to scare them all away.

I think adding the comment here provides sufficient justification for the added complexity.

sherm1 (commented):
Platform :lgtm: with two suggestions

> Reviewed 2 of 25 files at r7, 27 of 29 files at r8, 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-:-NXMehP22BvJeQs6eENg:b-3fpmlv)* status: 2 unresolved discussions

___
*[`geometry/optimization/test/cartesian_product_test.cc` line 80 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMc0kh6HDQuvopElYf:-NXMc0kh6HDQuvopElYg:b-l4tr0j) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/cartesian_product_test.cc#L80)):*
> ```cpp
>   // The old set is in a valid but unspecified state. For convenience we'll
>   // assert that it's empty, but that's not the only valid implementation,
>   // just the one we happen to currently use.
> ```

BTW I don't think it's necessary or helpful to talk here about other implementations we might have done. Consider removing the second sentence (META). (Also the current behavior seems like the obvious choice -- likely everyone will be happy with it forever.)
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 230 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMeGzL079cUdF1tqiO:-NXMeGzL079cUdF1tqiP:b-vyrqf7) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/spectrahedron_test.cc#L230)):*
> ```cpp
>   EXPECT_NO_THROW(dut.AddPointInSetConstraints(
>       &constraints, constraints.NewContinuousVariables<6>()));
>   EXPECT_EQ(constraints.positive_semidefinite_constraints().size(), 1);
> ```

BTW not clear to me what this additional test is accomplishing -- can you add a comment explaining?

jwnimmer-tri (commented):
> Reviewed 3 of 3 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19311)* status: :shipit: complete! all discussions resolved, LGTM from assignees sherm1(platform),RussTedrake(platform)

___
*[`geometry/optimization/test/cartesian_product_test.cc` line 80 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMc0kh6HDQuvopElYf:-NXMf3CJ97-l7K7IkjVJ:b-p18htk) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/cartesian_product_test.cc#L80)):*

BTW I don't think it's necessary or helpful to talk here about other implementations we might have done. Consider removing the second sentence (META). (Also the current behavior seems like the obvious choice -- likely everyone will be happy with it forever.)
</blockquote></details>

In general I think it's important for unit tests to make it clear when they are asserting something beyond the documented contract of the API.

Would there be a less confusing phrasing you can think of?
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 230 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMeGzL079cUdF1tqiO:-NXMg5NA-mx_G3naqXCM:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/spectrahedron_test.cc#L230)):*

BTW not clear to me what this additional test is accomplishing -- can you add a comment explaining?
</blockquote></details>

Done.

sherm1:
*[`geometry/optimization/test/cartesian_product_test.cc` line 80 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMc0kh6HDQuvopElYf:-NXMl-W78X6kSoZwH-78:b-7chaor) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/cartesian_product_test.cc#L80)):*

In general I think it's important for unit tests to make it clear when they are asserting something beyond the documented contract of the API.

Would there be a less confusing phrasing you can think of?
</blockquote></details>

IMO the first sentence achieved that adequately. Several of the new tests had just the first sentence for the same condition; I thought that was fine.

jwnimmer-tri:
*[`geometry/optimization/test/cartesian_product_test.cc` line 80 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19311#-NXMc0kh6HDQuvopElYf:-NXMlwp77QeXJlqvk3Uh:b-i6fi2x) ([raw file](https://github.com/RobotLocomotion/drake/blob/55f0809def762c3c408197f25a6f7d96aed27d5a/geometry/optimization/test/cartesian_product_test.cc#L80)):*

IMO the first sentence achieved that adequately. Several of the new tests had just the first sentence for the same condition; I thought that was fine.
</blockquote></details>

Right, those were the tests that only tested the "valid but unspecified" predicate.  In the tests that assert "empty" instead of "unspecified", we have the extra text.  I think I'm OK leaving it.

PR #19312, opened 2023-04-29
Make GCS Preprocessing respect solver options

RussTedrake:
Previously, the solver passed in was not respected by the preprocessor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -249,6 +249,7 @@ drake_cc_googletest(
         "//solvers:csdp_solver",
         "//solvers:gurobi_solver",
         "//solvers:ipopt_solver",
+        "//solvers:linear_system_solver",
         "//solvers:mosek_solver",
     ],
 )
--- a/geometry/optimization/graph_of_convex_sets.cc
+++ b/geometry/optimization/graph_of_convex_sets.cc
@@ -51,6 +51,23 @@ using symbolic::Expression;
 using symbolic::Variable;
 using symbolic::Variables;
 
+namespace {
+MathematicalProgramResult Solve(const MathematicalProgram& prog,
+                                const GraphOfConvexSetsOptions& options,
+                                bool rounding) {
+  MathematicalProgramResult result;
+  auto solver_options = (rounding && options.rounding_solver_options)
+                            ? options.rounding_solver_options
+                            : options.solver_options;
+  if (options.solver) {
+    options.solver->Solve(prog, {}, solver_options, &result);
+  } else {
+    result = solvers::Solve(prog, {}, solver_options);
+  }
+  return result;
+}
+}  // namespace
+
 GraphOfConvexSets::~GraphOfConvexSets() = default;
 
 Vertex::Vertex(VertexId id, const ConvexSet& set, std::string name)
@@ -310,7 +327,8 @@ std::string GraphOfConvexSets::GetGraphvizString(
 // "Motion Planning around Obstacles with Convex Optimization":
 // https://arxiv.org/abs/2205.04422
 std::set<EdgeId> GraphOfConvexSets::PreprocessShortestPath(
-    VertexId source_id, VertexId target_id) const {
+    VertexId source_id, VertexId target_id,
+    const GraphOfConvexSetsOptions& options) const {
   DRAKE_DEMAND(vertices_.find(source_id) != vertices_.end());
   DRAKE_DEMAND(vertices_.find(target_id) != vertices_.end());
 
@@ -432,7 +450,7 @@ std::set<EdgeId> GraphOfConvexSets::PreprocessShortestPath(
     degree.at(e->v().id()).evaluator()->set_bounds(Vector1d(0), Vector1d(0));
 
     // Check if edge e = (u,v) could be on a path from start to goal.
-    auto result = Solve(prog);
+    auto result = Solve(prog, options, false);
     if (!result.is_success()) {
       unusable_edges.insert(edge_id);
     }
@@ -626,23 +644,6 @@ void GraphOfConvexSets::AddPerspectiveConstraint(
   }
 }
 
-namespace {
-MathematicalProgramResult Solve(const MathematicalProgram& prog,
-                                const GraphOfConvexSetsOptions& options,
-                                bool rounding) {
-  MathematicalProgramResult result;
-  auto solver_options = (rounding && options.rounding_solver_options)
-                            ? options.rounding_solver_options
-                            : options.solver_options;
-  if (options.solver) {
-    options.solver->Solve(prog, {}, solver_options, &result);
-  } else {
-    result = solvers::Solve(prog, {}, solver_options);
-  }
-  return result;
-}
-}  // namespace
-
 MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
     VertexId source_id, VertexId target_id,
     const GraphOfConvexSetsOptions& options) const {
@@ -651,7 +652,7 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
 
   std::set<EdgeId> unusable_edges;
   if (options.preprocessing) {
-    unusable_edges = PreprocessShortestPath(source_id, target_id);
+    unusable_edges = PreprocessShortestPath(source_id, target_id, options);
   }
 
   MathematicalProgram prog;
--- a/geometry/optimization/graph_of_convex_sets.h
+++ b/geometry/optimization/graph_of_convex_sets.h
@@ -505,8 +505,9 @@ class GraphOfConvexSets {
   /* Facilitates testing. */
   friend class PreprocessShortestPathTest;
 
-  std::set<EdgeId> PreprocessShortestPath(VertexId source_id,
-                                          VertexId target_id) const;
+  std::set<EdgeId> PreprocessShortestPath(
+      VertexId source_id, VertexId target_id,
+      const GraphOfConvexSetsOptions& options) const;
 
   // Adds a perspective constraint to the mathematical program to upper bound
   // the cost below a slack variable, ℓ. Specifically given a cost g(x) to
--- a/geometry/optimization/test/graph_of_convex_sets_test.cc
+++ b/geometry/optimization/test/graph_of_convex_sets_test.cc
@@ -1256,14 +1256,28 @@ TEST_F(PreprocessShortestPathTest, CheckResults) {
   }
 }
 
-// Confirm that the preprocessor uses the solver passed in through the options
-// structure.
-TEST_F(PreprocessShortestPathTest, Solver) {
-  // Intentionally choose a solver that cannot run the preprocessing.
+// PreprocessShortestPath() is supposed to consume the solver options. More
+// particularly, those options must be passed to it from SolveShortestPath().
+// We'll exploit friend access to confirm PreporocessShortestPath() makes use
+// of the options. Confirming the options are *passed* is trickier. This is
+// because both PreprocessShortestPath() and SolveShortestPath() invoke Solve
+// with the same options. It is difficult to pass in a set of options such that
+// we can discen the exercise in PreprocessShortestPath strictly from looking
+// at the result. For example, passing an incompatible solver (as we do below)
+// will throw an exception no matter what, even if SolveShortestPath() skips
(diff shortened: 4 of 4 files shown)


RussTedrake (commented):
(this was the reason that the pset problem was calling gurobi on your machine)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUEDY6U4BgwPlF9Omqv:bjtchob)* status: LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @TobiaMarcucci)

TobiaMarcucci (commented):
Looks good to me!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUJCYyf8qxhlBvwUHk1:b9j9n3q)* status: LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUJCf-A64v9U4ML5ybw:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUJDQ5m9Jf7M16cjxVy:b-6cxuoj)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUNy3kd2tBzr_hVIkS0:bsfbm51) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

This doesn't actually serve as a regression test of the new feature. Can we add a test that would have failed without this fix that now passes?
</blockquote></details>

I was looking for a way to do that, but don't have a great solution.  There is an internal call in the preprocessor which should now choose the solver defined in options, and before it used a different solver.  I could set up an entire solver infrastructure that counts calls to the solver, but that seems too much.

Do you have any suggestions?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUNz1skBCRk_ymXFEAR:b-a2axqm) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

I was looking for a way to do that, but don't have a great solution.  There is an internal call in the preprocessor which should now choose the solver defined in options, and before it used a different solver.  I could set up an entire solver infrastructure that counts calls to the solver, but that seems too much.

Do you have any suggestions?
</blockquote></details>

One way to check if a specific solver was ever called might be to set the solver-specific "print file" option, and then check for the presence or absence of the log file in the test. Whether that works would depend on the solver having such an option, and whether the preprocessing solver was different than other solvers called later.

Another way to check would be to set a _malformed_ solver-specific option, and check that it threw specific exception text, indicating that the options were live.

RussTedrake:
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUNzn8oAdcpi81Ik17h:b-tiaayw) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

That was going to be my suggestion. I.e., depending on the solver set, limit iterations to give a bad answer, etc.
</blockquote></details>

Thanks.  I considered the case where I could make it throw, but that felt ugly to me.  I will check the print file first (tonight).

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUO-KfZA6h5GzyOQuX9:btfxul8) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

Thanks.  I considered the case where I could make it throw, but that felt ugly to me.  I will check the print file first (tonight).
</blockquote></details>

All you're really testing is that the options get passed through. From that perspective, an exception (ideally with an expected message), may be the cleanest way to do that. It would forestall doing work unrelated to the mechanism under test.

SeanCurtis-TRI (commented):
:LGTM: (modulo the discussions below).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUOJmq31HL_p2-XKUV9:b-pf7ttr)* status: 1 unresolved discussion

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/19312#-:-NUOaopt66f-KbMdLV9Y:bnfp4nl)* status: 1 unresolved discussion

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: 1 unresolved discussion

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUPGmb0Bdtj4TBDy754:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

All you're really testing is that the options get passed through. From that perspective, an exception (ideally with an expected message), may be the cleanest way to do that. It would forestall doing work unrelated to the mechanism under test.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: 1 unresolved discussion

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1216 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NUMcxmTEuDwQqadgvIH:-NUS0SuE4inucFBgXoYX:b-b0fzx1) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c5335b3ca1bdb6a13b9b7698563ab36ac165e3f/geometry/optimization/test/graph_of_convex_sets_test.cc#L1216)):*

Done.
</blockquote></details>

The test is insufficient; see my comments below. I'm closing this thread in favor of that one.
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1261 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NURvWUU8L1BGY6Mol1G:-NURvWUU8L1BGY6Mol1H:bofbm4m) ([raw file](https://github.com/RobotLocomotion/drake/blob/7293965efbc39ee083c897a6239c1e65c547be16/geometry/optimization/test/graph_of_convex_sets_test.cc#L1261)):*
> ```cpp
> // Confirm that the preprocessor uses the solver passed in through the options
> // structure.
> TEST_F(PreprocessShortestPathTest, Solver) {
> ```

As I read this test, it doesn't satisfy the need. Here's why. Please feel free to point out my error:

1. This PR is all about making sure `PreprocessShortestPath` now consumes solver options (as passed by `SolveShortestPath`).
2. `SolveShortestPath` only calls `PreprocessShortestPath` iff `options.preproccesing` is true ([see here](https://github.com/RussTedrake/drake/blob/gcs_preprocess_options/geometry/optimization/graph_of_convex_sets.cc#L654-L656)).
3. `options.preprocessing` defaults to `false. ([see here](https://github.com/RussTedrake/drake/blob/gcs_preprocess_options/geometry/optimization/graph_of_convex_sets.h#L39)).

Therefore this invocation doesn't actually *call* `PreprocessShortestPath`, passing solver options. In fact, the exception is thrown inside the body of `SolveShortestPath` [here](https://github.com/RussTedrake/drake/blob/gcs_preprocess_options/geometry/optimization/graph_of_convex_sets.cc#L899).

Furthermore, I can set `preprocessing` to `true` or `false` in this test, arbitrarily, and it doesn't change the outcome of the test at all. It changes the *code path*; setting it to `true` causes the `Solve` invocation in `PreprocessShortestPath` to throw the same error message. Unfortunately, the observed outcome is not dispositive.

Instead, we should do the following:

1. Using the `LinearSystemSolver` you have here, you can make a *direct* call to `PreprocessShortestPath()`. That will throw the current exception message and show that the solver passed in options is being used.
2. The second half is making sure that `SolveShortestPath` is passing. This is a bit trickier.
    - Make sure that `options.preprocessing` is set to `true`.
    - Pass in a solver *capable* of doing the preprocessing work, however, pass in a `SolverOption` value for that solver that would cause the solve to report failure. Create a simple graph with a single edge. If the edge is returned as being an "unusable edge", the options matter. If it isn't included, then the options are ignored.

I can certainly help with the first step (it's trivial). The second is conceptually straightforward, but I haven't worked with the APIs enough to immediately know how to configure such a graph. That said, I'm willing to pitch in if you're swamped. Let me know and I'll use this as an excuse to better familiarize myself with the GCS code.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @RussTedrake)

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1261 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-NURvWUU8L1BGY6Mol1G:-NUxHcqu8UeUEAeUqZ_f:b41b1cz) ([raw file](https://github.com/RobotLocomotion/drake/blob/7293965efbc39ee083c897a6239c1e65c547be16/geometry/optimization/test/graph_of_convex_sets_test.cc#L1261)):*

I did a bit more exploring for number two. It feels incredibly tricky. Coming up with a combination of graph and solver settings such that the preprocessor is going to *uniquely* change the outcome requires deep understanding of the programs being solved. My superficial exploration didn't suggest anything.

So, for now, perhaps it'll be enough to do number one and then add a TODO for figuring out how to test number two.
</blockquote></details>

I went ahead and made the changes I've recommended so we can close this out.

SeanCurtis-TRI (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19312#-:-NUxNEbY3N1Xm6n5kkKA:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,SeanCurtis-TRI(platform)

PR #19316, opened 2023-05-01
Adds Spectrahedron isa ConvexSet

RussTedrake:
This adds the minimal functionality to be useful. I will follow-up with more PRs to fill in the "not implemented yet" blocks as needed.

fyi @jwnimmer-tri.  I expect this will land on your platform day tomorrow.  You have a lot of cleanups in the ConvexSet hierarchy; I tried to be consistent with most of them (but still have the copyable_unique_ptr in my bindings so far).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -17,6 +17,7 @@
 #include "drake/geometry/optimization/iris.h"
 #include "drake/geometry/optimization/minkowski_sum.h"
 #include "drake/geometry/optimization/point.h"
+#include "drake/geometry/optimization/spectrahedron.h"
 #include "drake/geometry/optimization/vpolytope.h"
 
 namespace drake {
@@ -255,6 +256,16 @@ void DefineGeometryOptimization(py::module m) {
         .def(py::pickle([](const Point& self) { return self.x(); },
             [](Eigen::VectorXd arg) { return Point(arg); }));
     py::implicitly_convertible<Point, copyable_unique_ptr<ConvexSet>>();
+  }
+
+  // Spectrahedron
+  {
+    const auto& cls_doc = doc.Spectrahedron;
+    py::class_<Spectrahedron, ConvexSet>(m, "Spectrahedron", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
+        .def(py::init<const solvers::MathematicalProgram&>(), py::arg("prog"),
+            cls_doc.ctor.doc_1args);
+    py::implicitly_convertible<Spectrahedron, copyable_unique_ptr<ConvexSet>>();
   }
 
   // VPolytope
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -206,6 +206,15 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertEqual(sum2.num_terms(), 2)
         self.assertIsInstance(sum2.term(0), mut.Point)
 
+    def test_spectrahedron(self):
+        s = mut.Spectrahedron()
+        prog = MathematicalProgram()
+        X = prog.NewSymmetricContinuousVariables(3)
+        prog.AddPositiveSemidefiniteConstraint(X)
+        prog.AddLinearEqualityConstraint(X[0, 0] + X[1, 1] + X[2, 2], 1)
+        s = mut.Spectrahedron(prog=prog)
+        self.assertEqual(s.ambient_dimension(), 6)
+
     def test_v_polytope(self):
         vertices = np.array([[0.0, 1.0, 2.0], [3.0, 7.0, 5.0]])
         vpoly = mut.VPolytope(vertices=vertices)
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -35,6 +35,7 @@ drake_cc_library(
         "intersection.cc",
         "minkowski_sum.cc",
         "point.cc",
+        "spectrahedron.cc",
         "vpolytope.cc",
     ],
     hdrs = [
@@ -45,6 +46,7 @@ drake_cc_library(
         "intersection.h",
         "minkowski_sum.h",
         "point.h",
+        "spectrahedron.h",
         "vpolytope.h",
     ],
     interface_deps = [
@@ -55,6 +57,7 @@ drake_cc_library(
     deps = [
         "//geometry:read_obj",
         "//solvers:choose_best_solver",
+        "//solvers:get_program_type",
         "//solvers:gurobi_solver",
         "//solvers:ipopt_solver",
         "//solvers:mosek_solver",
@@ -329,6 +332,16 @@ drake_cc_googletest(
         ":convex_set",
         ":test_utilities",
         "//common/test_utilities:eigen_matrix_compare",
+    ],
+)
+
+drake_cc_googletest(
+    name = "spectrahedron_test",
+    deps = [
+        ":convex_set",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+        "//solvers:solve",
     ],
 )
 
--- a/geometry/optimization/spectrahedron.cc
+++ b/geometry/optimization/spectrahedron.cc
@@ -1,0 +1,191 @@
+#include "drake/geometry/optimization/spectrahedron.h"
+
+#include <limits>
+#include <memory>
+
+#include <fmt/format.h>
+
+#include "drake/solvers/get_program_type.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
+using solvers::Binding;
+using solvers::Constraint;
+using solvers::MathematicalProgram;
+using solvers::ProgramAttribute;
+using solvers::ProgramAttributes;
+using solvers::VectorXDecisionVariable;
+using symbolic::Variable;
+
+namespace {
+
+// TODO(russt): This can be replaced by vars(indices) once we have Eigen 3.4.
+VectorXDecisionVariable GetVariablesByIndex(
+    const Eigen::Ref<const VectorXDecisionVariable>& vars,
+    std::vector<int> indices) {
+  VectorXDecisionVariable new_vars(indices.size());
+  for (int i = 0; i < ssize(indices); ++i) {
+    new_vars[i] = vars[indices[i]];
+  }
+  return new_vars;
+}
+
+}  // namespace
+
+Spectrahedron::Spectrahedron()
+    : ConvexSet(&ConvexSetCloner<Spectrahedron>, 0) {}
+
+Spectrahedron::Spectrahedron(const MathematicalProgram& prog)
+    : ConvexSet(&ConvexSetCloner<Spectrahedron>, prog.num_vars()) {
+  for (const ProgramAttribute& attr : prog.required_capabilities()) {
+    if (supported_attributes().count(attr) < 1) {
+      throw std::runtime_error(fmt::format(
+          "Spectrahedron does not support MathematicalPrograms that require "
+          "ProgramAttribute {}. If that attribute is convex, it might be "
+          "possible to add that support.",
+          attr));
+    }
+  }
+  sdp_ = prog.Clone();
+  // Remove any objective functions.
+  for (const auto& binding : sdp_->GetAllCosts()) {
+    sdp_->RemoveCost(binding);
+  }
+}
+
+Spectrahedron::~Spectrahedron() = default;
+
+const ProgramAttributes& Spectrahedron::supported_attributes() {
+  static const never_destroyed<ProgramAttributes> kSupportedAttributes{
+      ProgramAttributes{ProgramAttribute::kLinearCost,
+                        ProgramAttribute::kLinearConstraint,
+                        ProgramAttribute::kLinearEqualityConstraint,
+                        ProgramAttribute::kPositiveSemidefiniteConstraint}};
+  return kSupportedAttributes.access();
+}
+
+bool Spectrahedron::DoIsBounded() const {
+  throw std::runtime_error(
+      "Spectrahedron::IsBounded() is not implemented yet.");
+}
+
+bool Spectrahedron::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
+                                 double tol) const {
+  return sdp_->CheckSatisfied(sdp_->GetAllConstraints(), x, tol);
+}
+
+void Spectrahedron::DoAddPointInSetConstraints(
+    MathematicalProgram* prog,
+    const Eigen::Ref<const VectorXDecisionVariable>& x) const {
+  DRAKE_DEMAND(x.size() == sdp_->num_vars());
+  for (const auto& binding : sdp_->GetAllConstraints()) {
+    prog->AddConstraint(
+        binding.evaluator(),
+        GetVariablesByIndex(
+            x, sdp_->FindDecisionVariableIndices(binding.variables())));
+  }
+}
+
+std::vector<Binding<Constraint>>
+Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
+    MathematicalProgram* prog,
+    const Eigen::Ref<const VectorXDecisionVariable>& x,
+    const Variable& t) const {
+  DRAKE_DEMAND(x.size() == sdp_->num_vars());
+  std::vector<Binding<Constraint>> constraints;
+  const double kInf = std::numeric_limits<double>::infinity();
+
+  // Helper function that given a binding.variables() returns the corresponding
+  // subset of variables from `x` with `t` tacked on the end.
+  auto stack_xt = [&x, &t, this](const VectorXDecisionVariable& bind_vars) {
+    VectorXDecisionVariable xt(bind_vars.size() + 1);
+    xt << GetVariablesByIndex(x, sdp_->FindDecisionVariableIndices(bind_vars)),
+        t;
+    return xt;
+  };
+
+  // TODO(russt): Support SparseMatrix constraints.
+  for (const auto& binding : sdp_->bounding_box_constraints()) {
+    // t*lb ≤ x ≤ t*ub, implemented as
+    // [I,-lb]*[x;t] ≥ 0, [I,-ub]*[x;t] ≤ 0.
+    VectorXDecisionVariable vars = stack_xt(binding.variables());
+    MatrixXd Ab =
+        MatrixXd::Identity(binding.evaluator()->num_constraints(), vars.size());
+    // TODO(russt): Handle individual elements that are infinite.
+    if (binding.evaluator()->lower_bound().array().isFinite().any()) {
+      Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
+      prog->AddLinearConstraint(Ab, 0, kInf, vars);
+    }
+    if (binding.evaluator()->upper_bound().array().isFinite().any()) {
+      Ab.rightCols<1>() = -binding.evaluator()->upper_bound();
+      prog->AddLinearConstraint(Ab, -kInf, 0, vars);
+    }
+  }
+  for (const auto& binding : sdp_->linear_equality_constraints()) {
+    // Ax = t*b, implemented as
+    // [A,-lb]*[x;t] == 0.
+    VectorXDecisionVariable vars = stack_xt(binding.variables());
+    MatrixXd Ab(binding.evaluator()->num_constraints(), vars.size());
+    Ab.leftCols(binding.evaluator()->num_vars()) =
+        binding.evaluator()->GetDenseA();
+    Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
+    prog->AddLinearEqualityConstraint(Ab, 0, vars);
+  }
+  for (const auto& binding : sdp_->linear_constraints()) {
+    // t*lb <= Ax = t*ub, implemented as
+    // [A,-lb]*[x;t] ≥ 0, [A,-ub]*[x;t] ≤ 0.
+    VectorXDecisionVariable vars = stack_xt(binding.variables());
+    MatrixXd Ab(binding.evaluator()->num_constraints(), vars.size());
+    Ab.leftCols(binding.evaluator()->num_vars()) =
+        binding.evaluator()->GetDenseA();
+    if (binding.evaluator()->lower_bound().array().isFinite().any()) {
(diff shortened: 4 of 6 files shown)


AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 2 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake and @TobiaMarcucci)

___
*[`geometry/optimization/spectrahedron.h` line 39 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMTUjPBtygB6F3-j9L:-NUMTUjQA3lMEXSMvzLW:b9mzo3n) ([raw file](https://github.com/RobotLocomotion/drake/blob/9dbd3e86285bba2e174276e7108e5367b5605cf0/geometry/optimization/spectrahedron.h#L39)):*
> ```objc
>   bool DoIsBounded() const final;
> 
>   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
> ```

It feels unnatural to have to pass `x` as a vector to check membership in the Spectrahedron. Should we relax this to `Eigen::MatrixXd` everywhere?
___
*[`geometry/optimization/spectrahedron.cc` line 28 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMTxhI-0Cj0AIotG0c:-NUMTxhI-0Cj0AIotG0d:bv2moty) ([raw file](https://github.com/RobotLocomotion/drake/blob/9dbd3e86285bba2e174276e7108e5367b5605cf0/geometry/optimization/spectrahedron.cc#L28)):*
> ```cpp
> Spectrahedron::Spectrahedron(const MathematicalProgram& prog)
>     : ConvexSet(&ConvexSetCloner<Spectrahedron>, prog.num_vars()) {
>   DRAKE_DEMAND(solvers::GetProgramType(prog) == solvers::ProgramType::kSDP);
> ```

Does this  check that
```  
@pre @p prog must have exactly one positive-semidefinite constraint which is
  bound to *all* of the decision variables.
```

jwnimmer-tri (commented):
Thanks, looks great so far!  One nit below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-:-NUMhYA4E84qMbPfu1X1:b1mz1n)* status: 3 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake and @TobiaMarcucci)

___
*[`geometry/optimization/spectrahedron.cc` line 28 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMhe668R8SZ0bkQy9w:-NUMhe6740r5mYX9lG2P:b8trtbc) ([raw file](https://github.com/RobotLocomotion/drake/blob/ef23a2c68b5af1648f9c6bd24c97e41a710a7d10/geometry/optimization/spectrahedron.cc#L28)):*
> ```cpp
> Spectrahedron::Spectrahedron(const MathematicalProgram& prog)
>     : ConvexSet(&ConvexSetCloner<Spectrahedron>, prog.num_vars()) {
>   DRAKE_DEMAND(solvers::GetProgramType(prog) == solvers::ProgramType::kSDP);
> ```

nit When validating user input to a function, we should respond to mistakes using `throw` not `std::abort`.



_Suggestion:_
```cpp
DRAKE_THROW_UNLESS
```

TobiaMarcucci (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 4 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`geometry/optimization/spectrahedron.cc` line 60 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUNefVG4b_cALsf_L5O:-NUNefVG4b_cALsf_L5P:bpkmfb2) ([raw file](https://github.com/RobotLocomotion/drake/blob/ef23a2c68b5af1648f9c6bd24c97e41a710a7d10/geometry/optimization/spectrahedron.cc#L60)):*
> ```cpp
> 
> std::vector<Binding<Constraint>>
> Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
> ```

Just to be sure I understand this part. A Spectahedron can be defined by bounding-box constraints, linear inequality, quadratic inequalities, and semidefinite constraints? That's why one needs to go through all these cases within this method? if this is the case, this seems to be already quite close to the other approach we discussed, where each convex set in the GCS is put in canonical conic form and the perspective stuff is implemented only once for generic cone constraints?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 3 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @RussTedrake and @TobiaMarcucci)

___
*[`geometry/optimization/spectrahedron.h` line 39 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMTUjPBtygB6F3-j9L:-NUP1lWl9ErB4hNrRmRt:b-whorc3) ([raw file](https://github.com/RobotLocomotion/drake/blob/9dbd3e86285bba2e174276e7108e5367b5605cf0/geometry/optimization/spectrahedron.h#L39)):*

It feels unnatural to have to pass `x` as a vector to check membership in the Spectrahedron. Should we relax this to `Eigen::MatrixXd` everywhere?
</blockquote></details>

Done. Right now, this x can be the entire list of decision variables for a mathematical program.  (Plus this particular method is an overload from the base class).

I think this one should stay as is, but we can add additional sugar methods for the case you're thinking of in a follow-up.  I've added a TODO.
___
*[`geometry/optimization/spectrahedron.cc` line 28 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMTxhI-0Cj0AIotG0c:-NUP2Z2KCuzEhJwW_o0H:bq0guov) ([raw file](https://github.com/RobotLocomotion/drake/blob/9dbd3e86285bba2e174276e7108e5367b5605cf0/geometry/optimization/spectrahedron.cc#L28)):*

Does this  check that
```  
@pre @p prog must have exactly one positive-semidefinite constraint which is
  bound to *all* of the decision variables.
```

</blockquote></details>

No.  I decided to relax that constraint.  Good catch!
___
*[`geometry/optimization/spectrahedron.cc` line 28 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUMhe668R8SZ0bkQy9w:-NUP2omf9fXL5Ce_xjjk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/ef23a2c68b5af1648f9c6bd24c97e41a710a7d10/geometry/optimization/spectrahedron.cc#L28)):*

... and to be clear, it's fine for the `DoFoo()` NVI implementations to use `DEMAND`, since those are an internal double-check not input validation.  The input validation happens in the ConvexSet::Foo base class before it calls into the DoFoo function.
</blockquote></details>

Done.
___
*[`geometry/optimization/spectrahedron.cc` line 60 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUNefVG4b_cALsf_L5O:-NUP7e9W16TLYLo9G-wX:botybg8) ([raw file](https://github.com/RobotLocomotion/drake/blob/ef23a2c68b5af1648f9c6bd24c97e41a710a7d10/geometry/optimization/spectrahedron.cc#L60)):*

Just to be sure I understand this part. A Spectahedron can be defined by bounding-box constraints, linear inequality, quadratic inequalities, and semidefinite constraints? That's why one needs to go through all these cases within this method? if this is the case, this seems to be already quite close to the other approach we discussed, where each convex set in the GCS is put in canonical conic form and the perspective stuff is implemented only once for generic cone constraints?
</blockquote></details>

Done. Yes.  (as discussed f2f).

RussTedrake (commented):
+@jwnimmer-tri for platform review, please

@TobiaMarcucci -- hoping for an LGTM from you when you're ready.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-:-NUPIqL1COdPEUZl5EF2:b-fz5ep1)* status: 3 unresolved discussions, LGTM missing from assignees TobiaMarcucci,jwnimmer-tri(platform) (waiting on @jwnimmer-tri and @TobiaMarcucci)

jwnimmer-tri (commented):
See https://github.com/RussTedrake/drake/pull/48 with fixes and suggestions for some of the discussions below.

:lgtm: platform.

> Reviewed 3 of 6 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-:-NURgfTuAcTgzhz4UbPT:bu4g1aj)* status: 9 unresolved discussions, LGTM missing from assignee TobiaMarcucci (waiting on @TobiaMarcucci)

___
*[`geometry/optimization/spectrahedron.h` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURcukM7HDE5ome8fVf:-NURcukN-THiBpJ6yjZ0:bvh0lc9) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.h#L75)):*
> ```objc
>   copyable_unique_ptr<solvers::MathematicalProgram> sdp_{};
> 
>   static const solvers::ProgramAttributes supported_attributes_;
> ```

The class `ProgramAttributes` has a non-trivial constructor and destructor, so need special handling for https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables.

This is fixed in my meta-PR.
___
*[`geometry/optimization/spectrahedron.cc` line 40 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURg-81DHP3ti4-fTOl:-NURg-81DHP3ti4-fTOm:b-ajkuo1) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L40)):*
> ```cpp
> 
> const solvers::ProgramAttributes Spectrahedron::supported_attributes_ = {
>     ProgramAttribute::kLinearCost, ProgramAttribute::kLinearConstraint,
> ```

I don't follow the reasoning for why we allow for linear costs, but not e.g. quadratic cost or generic costs.  In the implementation, we're only checking a feasibility problem anyway and we remove _all_ costs from the input `prog`. It seems to me like we should either advertise _all_ costs in the "supported attributes" set, or possibly _no_ costs with a disclaimer in the Doxygen that we only list the supported constraints and end up punting all costs anyway.  Having just one cost listed here seems inconsistent.



_Code quote:_
```cpp
ProgramAttribute::kLinearCost,
```
___
*[`geometry/optimization/spectrahedron.cc` line 107 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURqQ5gArXUQYMdmVJP:-NURqQ5gArXUQYMdmVJQ:bmo0hn8) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L107)):*
> ```cpp
>         t;
>     MatrixXd Ab = MatrixXd::Identity(binding.evaluator()->num_constraints(),
>                 binding.evaluator()->num_vars() + 1);
> ```

nit The copy-pasted code for subsetting `[x;t]` and `num_vars + 1` somewhat obscures the bigger picture within of each loop.  Some refactoring for conciseness would help readability.

This is fixed in my meta-PR.
___
*[`geometry/optimization/spectrahedron.cc` line 130 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURjmPq0ZqVcsjiLm3F:-NURjmPq0ZqVcsjiLm3G:b-7c6kfa) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L130)):*
> ```cpp
>         binding.evaluator()->GetDenseA();
>     Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
>     prog->AddLinearConstraint(Ab, 0, 0, vars);
> ```

Here, we add a `LinearConstraint` to the new `prog`.  It has the mathematical form of a linear **equality** constraint, but not the runtime type of one.

Arguably, that's a bug in `MathematicalProgram::AddLinearConstraint` not detecting that `(ub.array() == lb.array()).all()`, but in the meantime should we avoid the pothole by calling AddLinearEqualityConstraint instead?

The program classification will not be precise otherwise (though unlikely to matter in practice).



_Suggestion:_
```cpp
prog->AddLinearEqualityConstraint(Ab, VectorXd::Zero(num_constraints), vars);
```
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURr5BM78wy6JAutzXb:-NURr5BM78wy6JAutzXc:by7b9jq) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/test/spectrahedron_test.cc#L41)):*
> ```cpp
> }
> 
> /**
> ```

nit Don't use Doxygen opener for non-Doxygen code.



_Code quote:_
```cpp
/**
```
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURs0UxAyq7H8lXpMqe:-NURs0UxAyq7H8lXpMqf:b-v36cgq) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/test/spectrahedron_test.cc#L95)):*
> ```cpp
>           0, x3, t3),
>       ".*not implemented yet.*");
> }
> ```

Working

Are these tests sufficient coverage?  It seems like we could drop some constraints and still pass the x-true, x-false spot checks.

Will explore locally.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 9 unresolved discussions, LGTM missing from assignee TobiaMarcucci (waiting on @TobiaMarcucci)

___
*[`geometry/optimization/test/spectrahedron_test.cc` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURs0UxAyq7H8lXpMqe:-NURtUN48o9QtsQ09SRS:b-dv5ck) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/test/spectrahedron_test.cc#L95)):*

Working

Are these tests sufficient coverage?  It seems like we could drop some constraints and still pass the x-true, x-false spot checks.

Will explore locally.
</blockquote></details>

Yeah, if I comment out the whole `for (const auto& binding : sdp_->linear_constraints()) { ... }` block, all tests still pass.

Probably the easiest fix is just to directly interrogate all of the constraints that came out for their _exact_ mathematical form (checking the A matrix, bounds, etc).  For your input program, it seems like the added constraints are not that complicated and we could easily desk-check the intended result.

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-:-NUSJz5C45BKUtF6fO-I:bnfp4nl)* status: 8 unresolved discussions

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 3 unresolved discussions

___
*[`geometry/optimization/spectrahedron.h` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURcukM7HDE5ome8fVf:-NUTNMUrAjrnkRcZ3qO7:bmc9pxa) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.h#L75)):*

The class `ProgramAttributes` has a non-trivial constructor and destructor, so need special handling for https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables.

This is fixed in my meta-PR.
</blockquote></details>

Done.  thanks.
___
*[`geometry/optimization/spectrahedron.cc` line 40 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURg-81DHP3ti4-fTOl:-NUTNS2m46sebHSYYyBD:b-ygvc73) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L40)):*

I don't follow the reasoning for why we allow for linear costs, but not e.g. quadratic cost or generic costs.  In the implementation, we're only checking a feasibility problem anyway and we remove _all_ costs from the input `prog`. It seems to me like we should either advertise _all_ costs in the "supported attributes" set, or possibly _no_ costs with a disclaimer in the Doxygen that we only list the supported constraints and end up punting all costs anyway.  Having just one cost listed here seems inconsistent.
</blockquote></details>

By convention, semidefinite programs can only ever have linear costs.
___
*[`geometry/optimization/spectrahedron.cc` line 107 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURqQ5gArXUQYMdmVJP:-NUTNaP99drfxZeKJqGE:b-l5r1cy) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L107)):*

nit The copy-pasted code for subsetting `[x;t]` and `num_vars + 1` somewhat obscures the bigger picture within of each loop.  Some refactoring for conciseness would help readability.

This is fixed in my meta-PR.
</blockquote></details>

Done.  Thanks.
___
*[`geometry/optimization/spectrahedron.cc` line 130 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURjmPq0ZqVcsjiLm3F:-NUTOD4A-XO76INFDTiB:b-6luv3l) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/spectrahedron.cc#L130)):*

Here, we add a `LinearConstraint` to the new `prog`.  It has the mathematical form of a linear **equality** constraint, but not the runtime type of one.

Arguably, that's a bug in `MathematicalProgram::AddLinearConstraint` not detecting that `(ub.array() == lb.array()).all()`, but in the meantime should we avoid the pothole by calling AddLinearEqualityConstraint instead?

The program classification will not be precise otherwise (though unlikely to matter in practice).
</blockquote></details>

Done.  good catch.
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURr5BM78wy6JAutzXb:-NUTOHCwBDX62x6T_Fqs:bmfxq1z) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/test/spectrahedron_test.cc#L41)):*

nit Don't use Doxygen opener for non-Doxygen code.
</blockquote></details>

Done.  
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NURs0UxAyq7H8lXpMqe:-NUTTUnz2ccbjSQeXf5Q:b-34xfb4) ([raw file](https://github.com/RobotLocomotion/drake/blob/88dae28f1c421ffe01f739e4b9d41d7a711f652d/geometry/optimization/test/spectrahedron_test.cc#L95)):*

Yeah, if I comment out the whole `for (const auto& binding : sdp_->linear_constraints()) { ... }` block, all tests still pass.

Probably the easiest fix is just to directly interrogate all of the constraints that came out for their _exact_ mathematical form (checking the A matrix, bounds, etc).  For your input program, it seems like the added constraints are not that complicated and we could easily desk-check the intended result.
</blockquote></details>

I've added the ones that are easy comparison, and more checks around the ones that are more complicated.  Reassembling all of the new matrices to check the scaling is a pretty heavy lift. 



_Code quote:_
```cpp
  Vector<double, 7> xt_test;
  xt_test << x_star, 1;
  EXPECT_TRUE(prog3.CheckSatisfied(prog3.GetAllConstraints(), xt_test));
  EXPECT_TRUE(
      prog3.CheckSatisfied(prog3.GetAllConstraints(), VectorXd::Zero(7)));
  xt_test << x_bad, 1;
  EXPECT_FALSE(prog3.CheckSatisfied(prog3.GetAllConstraints(), xt_test));
```

jwnimmer-tri (commented):
> Reviewed 1 of 2 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 1 unresolved discussion

___
*[`geometry/optimization/spectrahedron.cc` line 147 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUTYks6EwosOkbXD22-:-NUTYks6EwosOkbXD220:b6tqzwq) ([raw file](https://github.com/RobotLocomotion/drake/blob/33d2201ee67f5cb4d6cc066b171f17121f03d9dd/geometry/optimization/spectrahedron.cc#L147)):*
> ```cpp
>     if (binding.evaluator()->lower_bound().array().isFinite().any()) {
>       Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
>       prog->AddLinearConstraint(Ab, 0, kInf, vars);
> ```

If I comment out this line of code, no tests fail.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 1 unresolved discussion

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19316)* status: 1 unresolved discussion

___
*[`geometry/optimization/spectrahedron.cc` line 147 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19316#-NUTYks6EwosOkbXD22-:-NUU4CHA3m-M0ZB1XlWH:bmob7rx) ([raw file](https://github.com/RobotLocomotion/drake/blob/33d2201ee67f5cb4d6cc066b171f17121f03d9dd/geometry/optimization/spectrahedron.cc#L147)):*

If I comment out this line of code, no tests fail.
</blockquote></details>

That was because of the linear constraint having no lower bound (not the testing).  I've added a lower bound, and more tests.

PR #19321, opened 2023-05-01
Add a default constructor to all ConvexSet subclasses

jwnimmer-tri:
The user can already create such a set by moving-from an existing object. Making it directly constructible instead makes it easier to operate on and unit test these objects.

Add some missing nullptr checks.

---

Towards #19311.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -81,6 +81,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.CartesianProduct;
     py::class_<CartesianProduct, ConvexSet>(m, "CartesianProduct", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init([](const std::vector<ConvexSet*>& sets) {
           return std::make_unique<CartesianProduct>(CloneConvexSets(sets));
         }),
@@ -110,6 +111,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.HPolyhedron;
     py::class_<HPolyhedron, ConvexSet>(m, "HPolyhedron", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init<const Eigen::Ref<const Eigen::MatrixXd>&,
                  const Eigen::Ref<const Eigen::VectorXd>&>(),
             py::arg("A"), py::arg("b"), cls_doc.ctor.doc_2args)
@@ -171,6 +173,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.Hyperellipsoid;
     py::class_<Hyperellipsoid, ConvexSet>(m, "Hyperellipsoid", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init<const Eigen::Ref<const Eigen::MatrixXd>&,
                  const Eigen::Ref<const Eigen::VectorXd>&>(),
             py::arg("A"), py::arg("center"), cls_doc.ctor.doc_2args)
@@ -203,6 +206,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.Intersection;
     py::class_<Intersection, ConvexSet>(m, "Intersection", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init([](const std::vector<ConvexSet*>& sets) {
           return std::make_unique<Intersection>(CloneConvexSets(sets));
         }),
@@ -219,6 +223,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.MinkowskiSum;
     py::class_<MinkowskiSum, ConvexSet>(m, "MinkowskiSum", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init([](const std::vector<ConvexSet*>& sets) {
           return std::make_unique<MinkowskiSum>(CloneConvexSets(sets));
         }),
@@ -238,6 +243,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.Point;
     py::class_<Point, ConvexSet>(m, "Point", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init<const Eigen::Ref<const Eigen::VectorXd>&>(), py::arg("x"),
             cls_doc.ctor.doc_1args)
         .def(py::init<const QueryObject<double>&, GeometryId,
@@ -264,6 +270,7 @@ void DefineGeometryOptimization(py::module m) {
   {
     const auto& cls_doc = doc.VPolytope;
     py::class_<VPolytope, ConvexSet>(m, "VPolytope", cls_doc.doc)
+        .def(py::init<>(), cls_doc.ctor.doc)
         .def(py::init<const Eigen::Ref<const Eigen::MatrixXd>&>(),
             py::arg("vertices"), cls_doc.ctor.doc_vertices)
         .def(py::init<const HPolyhedron&>(), py::arg("H"),
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -40,6 +40,7 @@ class TestGeometryOptimization(unittest.TestCase):
         self.z = self.prog.NewContinuousVariables(2, "z")
 
     def test_point_convex_set(self):
+        mut.Point()
         p = np.array([11.1, 12.2, 13.3])
         point = mut.Point(p)
         self.assertEqual(point.ambient_dimension(), 3)
@@ -53,6 +54,7 @@ class TestGeometryOptimization(unittest.TestCase):
         # builds from shape.
 
     def test_h_polyhedron(self):
+        mut.HPolyhedron()
         hpoly = mut.HPolyhedron(A=self.A, b=self.b)
         self.assertEqual(hpoly.ambient_dimension(), 3)
         np.testing.assert_array_equal(hpoly.A(), self.A)
@@ -158,6 +160,7 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertEqual(hpoly.A().shape, (4, 3))
 
     def test_hyper_ellipsoid(self):
+        mut.Hyperellipsoid()
         ellipsoid = mut.Hyperellipsoid(A=self.A, center=self.b)
         self.assertEqual(ellipsoid.ambient_dimension(), 3)
         np.testing.assert_array_equal(ellipsoid.A(), self.A)
@@ -196,6 +199,7 @@ class TestGeometryOptimization(unittest.TestCase):
         np.testing.assert_array_equal(e_ball3.center(), [0, 0, 0])
 
     def test_minkowski_sum(self):
+        mut.MinkowskiSum()
         point = mut.Point(np.array([11.1, 12.2, 13.3]))
         hpoly = mut.HPolyhedron(A=self.A, b=self.b)
         sum = mut.MinkowskiSum(setA=point, setB=hpoly)
@@ -216,6 +220,7 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertEqual(s.ambient_dimension(), 6)
 
     def test_v_polytope(self):
+        mut.VPolytope()
         vertices = np.array([[0.0, 1.0, 2.0], [3.0, 7.0, 5.0]])
         vpoly = mut.VPolytope(vertices=vertices)
         self.assertEqual(vpoly.ambient_dimension(), 2)
@@ -289,6 +294,7 @@ class TestGeometryOptimization(unittest.TestCase):
         return length
 
     def test_cartesian_product(self):
+        mut.CartesianProduct()
         point = mut.Point(np.array([11.1, 12.2, 13.3]))
         h_box = mut.HPolyhedron.MakeBox(
             lb=[-1, -1, -1], ub=[1, 1, 1])
@@ -306,6 +312,7 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertIsInstance(sum2.factor(1), mut.HPolyhedron)
 
     def test_intersection(self):
+        mut.Intersection()
         point = mut.Point(np.array([0.1, 0.2, 0.3]))
         h_box = mut.HPolyhedron.MakeBox(
             lb=[-1, -1, -1], ub=[1, 1, 1])
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -27,18 +27,21 @@ using symbolic::Variable;
 
 namespace {
 
-int sum_ambient_dimensions(const ConvexSets& sets) {
+int SumAmbientDimensions(const ConvexSets& sets) {
   int dim = 0;
-  for (const auto& s : sets) {
-    dim += s->ambient_dimension();
+  for (const copyable_unique_ptr<ConvexSet>& set : sets) {
+    DRAKE_THROW_UNLESS(set != nullptr);
+    dim += set->ambient_dimension();
   }
   return dim;
 }
 
 }  // namespace
 
+CartesianProduct::CartesianProduct() : CartesianProduct(ConvexSets{}) {}
+
 CartesianProduct::CartesianProduct(const ConvexSets& sets)
-    : ConvexSet(sum_ambient_dimensions(sets)), sets_{sets} {}
+    : ConvexSet(SumAmbientDimensions(sets)), sets_(sets) {}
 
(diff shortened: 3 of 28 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/convex_set.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUNkW2Q42eUyJMSQJlv:-NUNkW2Q42eUyJMSQJlw:b-bzcmzv) ([raw file](https://github.com/RobotLocomotion/drake/blob/83bd0c379ad9f5019cf964cfa16563ab51404bd1/geometry/optimization/convex_set.cc#L53)):*
> ```cpp
>     const Eigen::Ref<const solvers::VectorXDecisionVariable>& x,
>     const symbolic::Variable& t) const {
>   DRAKE_THROW_UNLESS(ambient_dimension() > 0);
> ```

Working

Some unit tests are hitting this.

Either they are incorrectly formulated (in which case we should fix the test), or they really do need to add this vacuously false constraint, in which case we need to figure out how to add a vacuously false constraint to MathematicalProgram without it throwing an exception.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/convex_set.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUNkW2Q42eUyJMSQJlv:-NUNyZEH4CZvb94YFiaY:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/83bd0c379ad9f5019cf964cfa16563ab51404bd1/geometry/optimization/convex_set.cc#L53)):*

Working

Some unit tests are hitting this.

Either they are incorrectly formulated (in which case we should fix the test), or they really do need to add this vacuously false constraint, in which case we need to figure out how to add a vacuously false constraint to MathematicalProgram without it throwing an exception.
</blockquote></details>

Done

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NUO8YTfBdusw_YOwHYl:b7qub7u) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*
> ```cpp
> 
>     // Spatial non-negativity: y ∈ ϕX, z ∈ ϕX.
>     if (e->u().ambient_dimension() > 0) {
> ```

@RussTedrake @wrangelvid I need some guidance on this PR.

Here's the story...

At https://github.com/RobotLocomotion/drake/pull/19279#pullrequestreview-1406745625, we said that a `Point` with an ambient dimension of zero does **not** contain the empty vector (i.e., `VectorXd(0)`).

Therefore, for consistency `PointInSet(VectorXd(0))` must also be false.

What about the constraint-emitting functions `AddPointInSetConstraints` and `AddPointInNonnegativeScalingConstraints` though?

In the case of zero ambient dimension, they could either add exactly one vacuously-false constraint, or else add **no** constraints.  Rather then choose one or or the other, I decided to throw an error to make the user decide which they intended.

That means in `GcsTrajectoryOptimization` when we add the dummy source or target vertices (with zero dimension), the path search here would throw an exception.  I've edited the the code here to skip adding constraints for the dummy vertices.

Is this the best approach?  Should we instead declare that the constraint-adding functions are always no-ops when the ambient dimension is zero?  I was a little worried about the inconsistency of functions that ask "Is *x* in the set?" returning false, but having no associated constraint to express the membership condition.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NUPEHaVEmYvgYX2XUwk:bsvyzzc) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*

@RussTedrake @wrangelvid I need some guidance on this PR.

Here's the story...

At https://github.com/RobotLocomotion/drake/pull/19279#pullrequestreview-1406745625, we said that a `Point` with an ambient dimension of zero does **not** contain the empty vector (i.e., `VectorXd(0)`).

Therefore, for consistency `PointInSet(VectorXd(0))` must also be false.

What about the constraint-emitting functions `AddPointInSetConstraints` and `AddPointInNonnegativeScalingConstraints` though?

In the case of zero ambient dimension, they could either add exactly one vacuously-false constraint, or else add **no** constraints.  Rather then choose one or or the other, I decided to throw an error to make the user decide which they intended.

That means in `GcsTrajectoryOptimization` when we add the dummy source or target vertices (with zero dimension), the path search here would throw an exception.  I've edited the the code here to skip adding constraints for the dummy vertices.

Is this the best approach?  Should we instead declare that the constraint-adding functions are always no-ops when the ambient dimension is zero?  I was a little worried about the inconsistency of functions that ask "Is *x* in the set?" returning false, but having no associated constraint to express the membership condition.
</blockquote></details>

Your solution seems reasonable to me. Sorry for the cascade of implications.
@TobiaMarcucci -- do you want to weigh in here, too?

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NUPfQruEI5btfyazB5Q:b-fr9bb1) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*

Your solution seems reasonable to me. Sorry for the cascade of implications.
@TobiaMarcucci -- do you want to weigh in here, too?
</blockquote></details>

Seems also reasonable to me.
Making the constrain-adding functions on the empty set vertex no-ops might simplify things, but I am worried that it could lead to misunderstand. Maybe following your scheme and return an error when a user attempts to add a constraint directly on the empty set vertex would be more informative?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NURTdmcDbM7vPX_q4QO:b-2tq287) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*

Seems also reasonable to me.
Making the constrain-adding functions on the empty set vertex no-ops might simplify things, but I am worried that it could lead to misunderstand. Maybe following your scheme and return an error when a user attempts to add a constraint directly on the empty set vertex would be more informative?
</blockquote></details>

@wrangelvid -- What misunderstanding are you worried about?

In this proposal, I don't think that there are any constraints that the user thinks they have added which are now being ignored?  It puts more of a burden on us as algorithm writers (we have to check for non-empty before adding the set constraints), but the user should be fine?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NUS-Hvy2ur4pnJOemRj:bxvaxfu) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*
I think all three of us are in agreement that we must not silently no-op in this case.  (That's what I understood from @wrangelvid's first sentence, at least.)

> Maybe following your scheme and return an error when a user attempts to add a constraint directly on the empty set vertex would be more informative?

Ah, I see now; it took me a bit to parse this.  You're saying that in addition to the code already in this PR draft, we should change `GOCS::Vertex::AddContraint` to yell in case the vertex was empty.

jwnimmer-tri (commented):
This is ready for feature review now.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-:-NWxf1kB5hTGtYbrvEdv:b-4xcbdh)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 703 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NUO8YTfBdusw_YOwHYk:-NWxb46x9Lnz9lAtkrZK:b-rvq6lm) ([raw file](https://github.com/RobotLocomotion/drake/blob/d4428a378fbfa88166d2877af9eb658e52b61a08/geometry/optimization/graph_of_convex_sets.cc#L703)):*

I think all three of us are in agreement that we must not silently no-op in this case.  (That's what I understood from @wrangelvid's first sentence, at least.)

> Maybe following your scheme and return an error when a user attempts to add a constraint directly on the empty set vertex would be more informative?

Ah, I see now; it took me a bit to parse this.  You're saying that in addition to the code already in this PR draft, we should change `GOCS::Vertex::AddContraint` to yell in case the vertex was empty.
</blockquote></details>

Done.

In developing this change, I considered the question -- should the user be able to add a _cost_ on a zero-dimensional vertex (or an edge between two such vertices)?  I decided that it's fine.  The cost would already be required to not have any variables, so it would be a constant.  Adding a constant seems clear enough, with no ambiguity.

For constraints, we already had sanity-checking logic in place for Vertex:

```c++
Binding<Constraint> Vertex::AddConstraint(const Binding<Constraint>& binding) {
  DRAKE_THROW_UNLESS(ambient_dimension() > 0);
```

However, there was no checking in place for adding a zero-dimensional constraint on an edge between two zero-dimensional vertices.  I've added such a check now.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NWxsACHDYy96_Ip82Jg:-NWxsACHDYy96_Ip82Jh:bz51l85) (no related file):*
Working

Need to add Spectrahedron default ctor as well now.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NWxsACHDYy96_Ip82Jg:-NWxsz6y9fgboH8t5Tkc:b-ifhnyy) (no related file):*

Working

Need to add Spectrahedron default ctor as well now.
</blockquote></details>

Nevermind, it's already implemented and well-tested on master.

___

jwnimmer-tri:
+@RussTedrake for feature review or delegation, please.

RussTedrake (commented):
:lgtm:  thank you.

> Reviewed 7 of 23 files at r1, 1 of 1 files at r2, 21 of 21 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-:-NXFG6sIDplamELgtPO_:b-vwvtje)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/optimization/cartesian_product.cc` line 60 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NXFAMnc8_xohd3KmblZ:-NXFAMnc8_xohd3Kmbl_:b-dk1u7v) ([raw file](https://github.com/RobotLocomotion/drake/blob/91cdd731a2c7483fe7cda2a373fa0f54116749ee/geometry/optimization/cartesian_product.cc#L60)):*
> ```cpp
>   DRAKE_THROW_UNLESS(A_->rows() == x_ambient_dimension);
>   DRAKE_THROW_UNLESS(b_->rows() == x_ambient_dimension);
>   DRAKE_THROW_UNLESS(A_->cols() == y_ambient_dimension);
> ```

You've switched x and y here.

A is size [dim_y, dim_x].  
dim_x = ambient_dimension
___
*[`geometry/optimization/convex_set.h` line 87 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NXFBlkl3TsaGhNFQ1I7:-NXFBlkl3TsaGhNFQ1I8:b-pmsbcj) ([raw file](https://github.com/RobotLocomotion/drake/blob/91cdd731a2c7483fe7cda2a373fa0f54116749ee/geometry/optimization/convex_set.h#L87)):*
> ```objc
>   bool IsBounded() const {
>     if (ambient_dimension() == 0) {
>       return false;
> ```

I would say that the set with ambient dimension zero is always bounded (return true, not false).

RussTedrake (commented):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321)* status: LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

rpoyner-tri (commented):
:lgtm: distracting off-patch cruft; ignore if you like.

> Reviewed 19 of 21 files at r3, 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-:-NXGJK5F33oHE6kVJbdZ:bct672p)* status: 3 unresolved discussions

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 322 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NXGIQUL6N3BmMJd45Np:-NXGIQUL6N3BmMJd45Nq:b98hwfv) ([raw file](https://github.com/RobotLocomotion/drake/blob/37b178ee4bc1fe6be2c9828f015f1b46fcfd518b/geometry/optimization/test/graph_of_convex_sets_test.cc#L322)):*
> ```cpp
> 
> /*
> Let's me test with one edge defintely on the optimal path, and one definitely
> ```

nit grammar



_Suggestion:_
```cpp
Lets
```
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 322 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NXGIVGD5PbrJIpRYBXv:-NXGIVGD5PbrJIpRYBXw:b-qrf9xs) ([raw file](https://github.com/RobotLocomotion/drake/blob/37b178ee4bc1fe6be2c9828f015f1b46fcfd518b/geometry/optimization/test/graph_of_convex_sets_test.cc#L322)):*
> ```cpp
> 
> /*
> Let's me test with one edge defintely on the optimal path, and one definitely
> ```

nit typo



_Suggestion:_
```cpp
definitely
```
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 322 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19321#-NXGIe4H2tmROjsnd9iH:-NXGIe4H2tmROjsnd9iI:b-n005s) ([raw file](https://github.com/RobotLocomotion/drake/blob/37b178ee4bc1fe6be2c9828f015f1b46fcfd518b/geometry/optimization/test/graph_of_convex_sets_test.cc#L322)):*
> ```cpp
> 
> /*
> Let's me test with one edge defintely on the optimal path, and one definitely
> ```

nit Feel free to ignore me, but I dislike first-person comments, since change history tends to make them incomprehensible. Consider just deleting the first part of the sentence, and just start "Test with one edge".

PR #19323, opened 2023-05-01
[render_gl] Support multithreaded operation (sometimes)

jwnimmer-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)

### DIFF
--- a/geometry/render_gl/BUILD.bazel
+++ b/geometry/render_gl/BUILD.bazel
@@ -286,6 +286,26 @@ drake_cc_googletest_ubuntu_only(
     ],
 )
 
+drake_cc_googletest_ubuntu_only(
+    name = "thread_test",
+    data = [
+        "//geometry/render:test_models",
+    ],
+    tags = vtk_test_tags() + [
+        # We launch up to 3 child tasks.
+        "cpu:4",
+        # Generally, GL driver shared libraries are not tsan-clean.
+        # Instead we rely on functional testing to detect threading errors.
+        "no_tsan",
+    ],
+    deps = [
+        ":factory",
+        "//common/yaml",
+        "//geometry:geometry_instance",
+        "//systems/sensors:image_writer",
+    ],
+)
+
 add_lint_tests(
     cpplint_extra_srcs = [
         "no_factory.cc",
--- a/geometry/render_gl/internal_opengl_context.cc
+++ b/geometry/render_gl/internal_opengl_context.cc
@@ -276,7 +276,7 @@ class OpenGlContext::Impl {
     // (https://linux.die.net/man/3/xclosedisplay)
     // TODO(duy): If problems crop up in the future, this can/should be
     // investigated.
-    static Display* g_display = XOpenDisplay(0);
+    static Display* g_display = (XInitThreads(), XOpenDisplay(0));
     DRAKE_THROW_UNLESS(g_display != nullptr);
     return g_display;
   }
--- a/geometry/render_gl/test/thread_test.cc
+++ b/geometry/render_gl/test/thread_test.cc
@@ -1,0 +1,223 @@
+#include <filesystem>
+#include <future>
+#include <vector>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "drake/common/find_resource.h"
+#include "drake/common/yaml/yaml_io.h"
+#include "drake/geometry/geometry_instance.h"
+#include "drake/geometry/render_gl/factory.h"
+#include "drake/systems/sensors/image_writer.h"
+
+namespace {
+enum AsyncMode {
+  /* No asynchrony; run serially. */
+  kNone,
+  /* Use std::async(policy=std::launch::async). */
+  kTask,
+  /* Use std::thread. */
+  kThread,
+};
+std::string_view to_string(AsyncMode mode) {
+  switch (mode) {
+    case kNone:
+      return "kNone";
+    case kTask:
+      return "kTask";
+    case kThread:
+      return "kThread";
+  }
+  DRAKE_UNREACHABLE();
+}
+}  // namespace
+DRAKE_FORMATTER_AS(, , AsyncMode, x, to_string(x))
+
+namespace fs = std::filesystem;
+
+namespace drake {
+namespace geometry {
+namespace {
+
+using Eigen::AngleAxisd;
+using Eigen::Vector3d;
+using math::RigidTransformd;
+using math::RotationMatrixd;
+using render::ClippingRange;
+using render::ColorRenderCamera;
+using render::RenderCameraCore;
+using render::RenderEngine;
+using render::RenderLabel;
+using systems::sensors::CameraInfo;
+using systems::sensors::ImageRgba8U;
+
+struct Params {
+  /* Passes this object to an Archive.
+   Refer to @ref yaml_serialization "YAML Serialization" for background. */
+  template <typename Archive>
+  void Serialize(Archive* a) {
+    a->Visit(DRAKE_NVP(async_mode));
+    a->Visit(DRAKE_NVP(num_workers));
+    a->Visit(DRAKE_NVP(num_repeats));
+  }
+
+  AsyncMode async_mode{kNone};
+  int num_workers{1};
+  int num_repeats{1};
+
+  friend std::ostream& operator<<(std::ostream& os, const Params& params) {
+    os << yaml::SaveJsonString(params);
+    return os;
+  }
+};
+
+class ThreadTest : public testing::TestWithParam<Params> {};
+
+// This test case directly calls into RenderEngineGl to create images, without
+// relying on the SceneGraph and/or GeometryState infrastructure.
+TEST_P(ThreadTest, CallRenderEngineDirectly) {
+  const auto& [async_mode, num_workers, num_repeats] = GetParam();
+  DRAKE_DEMAND(num_workers > 0);
+  DRAKE_DEMAND(num_repeats > 0);
+
+  // Prepare the geometry.
+  const GeometryInstance instance(
+      RigidTransformd{},
+      Mesh(FindResourceOrThrow("drake/geometry/render/test/meshes/box.obj"), 1),
+      "box");
+  const PerceptionProperties perception_properties = []() {
+    PerceptionProperties result;
+    result.AddProperty("label", "id", RenderLabel::kDontCare);
+    return result;
+  }();
+
+  // Prepare the camera settings.
+  // The pose of the *camera* body frame C in the world frame: looking straight
+  // down from 3m above the ground.
+  const RigidTransformd X_WC(
+      RotationMatrixd{AngleAxisd(M_PI, Vector3d::UnitY()) *
+                      AngleAxisd(-M_PI_2, Vector3d::UnitZ())},
+      {0, 0, 3.0});
+  // The pose of the *sensor* frame S in the camera body frame.
+  RigidTransformd X_CS;
+  const ColorRenderCamera camera_params(RenderCameraCore(
+      "color", CameraInfo(640, 480, 2.0), ClippingRange(0.25, 10.0), X_CS));
+
+  // The worker functor for the i'th camera.
+  MatrixX<ImageRgba8U> images(num_workers, num_repeats);
+  std::atomic<int> num_errors{0};
+  auto work = [&images, &num_errors, num_repeats = num_repeats, &instance,
+               &perception_properties, &X_WC, &camera_params](int i) {
+    try {
+      // TODO(jwnimmer-tri) For the test to pass, we need to create a separate
+      // engine within each worker thread. We should try to adjust the engine
+      // so that we can call a clone of one canonical engine instead of making
+      // new ones from scratch.
+      std::unique_ptr<RenderEngine> renderer = MakeRenderEngineGl();
+      renderer->RegisterVisual(instance.id(), instance.shape(),
+                               perception_properties, instance.pose(),
+                               false /* needs update */);
+      for (int j = 0; j < num_repeats; ++j) {
+        ImageRgba8U& image = images(i, j);
+        image = ImageRgba8U(camera_params.core().intrinsics().width(),
+                            camera_params.core().intrinsics().height());
+        renderer->UpdateViewpoint(
+            X_WC * camera_params.core().sensor_pose_in_camera_body());
+        renderer->RenderColorImage(camera_params, &image);
+      }
+    } catch (std::exception& e) {
+      // Letting an exception leak out of a std::thread leads to obscure error
+      // messages. Catching, logging, and tallying is a much nicer outcome.
+      drake::log()->error("Worker {} exception: {}", i, e.what());
+      ++num_errors;
+    }
+  };
+
+  // Render on multiple threads concurrently.
+  switch (async_mode) {
+    case kNone:
+      drake::log()->info("Runnning {} workers serially", num_workers);
+      break;
+    case kTask:
+      drake::log()->info("Launching {} std::async workers", num_workers);
+      break;
+    case kThread:
+      drake::log()->info("Launching {} std::thread workers", num_workers);
+      break;
+  }
+  std::vector<std::future<void>> futures;
+  std::vector<std::thread> threads;
+  for (int i = 0; i < num_workers; ++i) {
+    switch (async_mode) {
+      case kNone:
+        work(i);
+        break;
+      case kTask:
+        futures.push_back(std::async(std::launch::async, work, i));
+        break;
+      case kThread:
+        threads.push_back(std::thread(work, i));
+        break;
+    }
+  }
+
+  // Wait for all workers to finish.
+  for (auto& thread : threads) {
+    thread.join();
+  }
+  for (auto& future : futures) {
+    future.get();
+  }
+
+  // Save the first image for offline inspection.
+  // See bazel-testlogs/geometry/render_gl/thread_test/test.outputs/output.zip.
+  const std::string filename = fmt::format("image-{}.png", async_mode);
+  if (const char* dir = std::getenv("TEST_UNDECLARED_OUTPUTS_DIR")) {
+    SaveToPng(images(0, 0), fs::path(dir) / filename);
+  }
+
+  // Check for errors.
+  EXPECT_EQ(num_errors, 0);
+
+  // Check that all output images are identical.
+  const ImageRgba8U& image_00 = images(0, 0);
+  for (int i = 0; i < num_workers; ++i) {
+    SCOPED_TRACE(i);
+    for (int j = 0; j < num_repeats; ++j) {
+      SCOPED_TRACE(j);
+      const ImageRgba8U& image_ij = images(i, j);
+      ASSERT_EQ(image_ij, image_00);
+    }
+  }
+
+  // Check that the output image has the clear color along with other color(s).
+  int num_clear_pixels = 0;
+  int num_box_pixels = 0;
+  for (int x = 0; x < image_00.width(); ++x) {
+    for (int y = 0; y < image_00.height(); ++y) {
+      const uint8_t* rgba = image_00.at(x, y);
+      if (std::vector<uint8_t>(rgba, rgba + 4) ==
+          std::vector<uint8_t>{204, 229, 255, 255}) {
+        ++num_clear_pixels;
+      } else {
+        ++num_box_pixels;
+      }
+    }
+  }
+  const double num_pixels = image_00.width() * image_00.height();
+  EXPECT_GT(num_clear_pixels / num_pixels, 0.75);
+  EXPECT_GT(num_box_pixels / num_pixels, 0.05);
+}
+
+constexpr Params kParams[] = {
+    {.async_mode = kNone, .num_workers = 3, .num_repeats = 2},
+    {.async_mode = kTask, .num_workers = 3, .num_repeats = 2},
+    {.async_mode = kThread, .num_workers = 3, .num_repeats = 2},
+};
+
+INSTANTIATE_TEST_SUITE_P(All, ThreadTest, testing::ValuesIn(kParams));
+
+}  // namespace
+}  // namespace geometry
+}  // namespace drake


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUNizHn4sCiqJ7qzscS:-NUNizHn4sCiqJ7qzscT:bh3lkyv) ([raw file](https://github.com/RobotLocomotion/drake/blob/95de376809ba7b83513f4ea263b4f8090dd33a91/geometry/render_gl/test/thread_test.cc#L81)):*
> ```cpp
>   // we create ends up crashing when used in a separate thread, so we must build
>   // one to throw away.
>   builder->scene_graph().AddRenderer("gl[dummy]", MakeRenderEngineGl());
> ```

BTW @SeanCurtis-TRI I eventually found this hack as a work-around for my problem.  Maybe later on we can dig in to why it's necessary.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NUOVIV0ARu6q6vrYjRu:b-dtnous) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*
> ```cpp
> 
>   // Save the first image for offline inspection.
>   // See bazel-testlogs/geometry/render_gl/thread_test/test.outputs/output.zip.
> ```

@SeanCurtis-TRI for the life of me, I can't figure out why my rendered image is all black even in the single-threaded case.

If I comment out the `box.obj` parsing, I successfully see the blue sky.  But with the box.obj there, the whole image is black.

I assume I'm doing something silly, but I can't spot what it is.  Could you take a look?

SeanCurtis-TRI:
*[`geometry/render_gl/test/thread_test.cc` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUNizHn4sCiqJ7qzscS:-NUvi7Mu3_N9DChpKjI1:bgbqk8s) ([raw file](https://github.com/RobotLocomotion/drake/blob/95de376809ba7b83513f4ea263b4f8090dd33a91/geometry/render_gl/test/thread_test.cc#L81)):*
Sorry about the radio silence. It's my intention to tackle the black pixels today.

BTW Way back, when I first learned graphics programming, I was introduced to a saying that has stood the test of time that is *particularly* relevant now (apologies if I've already inflicted this on you):

> The problem with OpenGl is that there's a million ways to draw a black screen.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NUxVjm1AtPkdu7tbw-p:b-ajssyf) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

@SeanCurtis-TRI any thoughts?

I had thought that maybe I had mis-posed the camera inside the box by mistake, but even moving the camera around didn't change anything.
</blockquote></details>

The fact that it's black points to a complete initialization failure. That's the default clear color. Even if you were inside the box and back face culling were enabled, you'd be getting the clear color. And the parms don't initialize to black. So, essentially, the context isn't being populated with *any* information.

What you should focus on is making sure that you can clear the image to an arbitrary color. That's your canary in the coal mine.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV-Xw379Lh7fOHq-sxm:butgwkq) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

The fact that it's black points to a complete initialization failure. That's the default clear color. Even if you were inside the box and back face culling were enabled, you'd be getting the clear color. And the parms don't initialize to black. So, essentially, the context isn't being populated with *any* information.

What you should focus on is making sure that you can clear the image to an arbitrary color. That's your canary in the coal mine.
</blockquote></details>

Without looking at the code a thought occurred to me this morning, did you change the `OpenGlContext` such that it is no longer shared across render engines? (I assume so.) I think what needs to happen is that the contents of the original context in the model engine needs to be reconstructed for each parallel context. That may well be why you're getting nothing but black.

Test the hypothesis: can you get a hold of the render engine *inside* SceneGraph and ask *it* to Render the image? (Subverting the whole `System` framework?) I predict that that will be the only instance of engine that produce non-black engines.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV01NEFCeX2Omr54KOE:b-3g62v2) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

Without looking at the code a thought occurred to me this morning, did you change the `OpenGlContext` such that it is no longer shared across render engines? (I assume so.) I think what needs to happen is that the contents of the original context in the model engine needs to be reconstructed for each parallel context. That may well be why you're getting nothing but black.

Test the hypothesis: can you get a hold of the render engine *inside* SceneGraph and ask *it* to Render the image? (Subverting the whole `System` framework?) I predict that that will be the only instance of engine that produce non-black engines.
</blockquote></details>

The code is creating multiple `RenderEngineGl` objects (one per thread), each with a unique and dedicated `OpenGlContext` that only it uses, and every `OpenGlContext` should be undergoing the same initialization as it used to.

The only shared state (so far as I see) is the `Display*` pointer which in theory the call to `XInitThreads()` should allow me to safely share.

SeanCurtis-TRI:
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV04OQT7TY3HhsP93Kk:bkgluul) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

Hmmm....so nope. That's not it. I've *now* looked at the code and see the `Context` isn't being copied. Each RenderEngine creates its own `OpenGlContext` and that only gets shared with its clones in the contexts but not across different render engines. You're not using one context per thread, but one render engine in the context per thread. (BTW, seems to me the context needs to be refreshed and the frozen to be safe.)

Yeah, I'd recognized that and I'm poking around some more. I just hadn't published the follow up comment.

BTW: I'm apparently failing on getting the image to output to the directory of my choice. What am I doing wrong?

```
bazel test  --action_env=TEST_UNDECLARED_OUTPUTS_DIR=/home/seancurtis/tmp geometry/render_gl:thread_test
```

This is sufficient to get it to save the image, but it places it deep in my cache which surprises me.

</blockquote></details>

I'll have to dig at this more in a bit...

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV05-DPDJLbyattyU_X:b9dfljn) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

I'll have to dig at this more in a bit...
</blockquote></details>

Don't set `TEST_UNDECLARED_OUTPUTS_DIR`, it's build-in: https://bazel.build/reference/test-encyclopedia.

After the test runs, the noted zipfile should exist.

SeanCurtis-TRI:
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV09pZgARDIr0XOhdkb:b-q42on5) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

Don't set `TEST_UNDECLARED_OUTPUTS_DIR`, it's build-in: https://bazel.build/reference/test-encyclopedia.

After the test runs, the noted zipfile should exist.
</blockquote></details>

Ah...so the test to see if it's defined is strictly for safety. It's obnoxious it's buried so deeply.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV0ATKQ8pvMPNs036cV:bdcfrhf) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*
> Ah...so the test to see if it's defined is strictly for safety.

Right.  When running by hand (without Bazel involvement) it would not be set.

I suppose you could do `env TEST_UNDECLARED_OUTPUTS_DIR=/tmp bazel-bin/render_gl/thread_test` in that case.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/thread_test.cc` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUNizHn4sCiqJ7qzscS:-NV9gslh2w8iOc1DvJ-H:b4yscld) ([raw file](https://github.com/RobotLocomotion/drake/blob/95de376809ba7b83513f4ea263b4f8090dd33a91/geometry/render_gl/test/thread_test.cc#L81)):*

Sorry about the radio silence. It's my intention to tackle the black pixels today.

BTW Way back, when I first learned graphics programming, I was introduced to a saying that has stood the test of time that is *particularly* relevant now (apologies if I've already inflicted this on you):

> The problem with OpenGl is that there's a million ways to draw a black screen.
</blockquote></details>

If you'd like to see the solution to this problem (sneak preview), you can see it here:

https://github.com/RobotLocomotion/drake/commit/9344b17cbdca5c6124cf1f066f3e2eb1425d02e1

I'm considering about submitting it in a PR while I figure out the black screen problem; but it'll require a new test. It'll be a very stripped down version of `thread_test_2.cc` found here:

https://github.com/RobotLocomotion/drake/commit/93220c8067826c8e7a81ea92b2c67f1a0582117d
___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NV9hPhe7Z52zMw9xPOK:b7ds28s) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

> Ah...so the test to see if it's defined is strictly for safety.

Right.  When running by hand (without Bazel involvement) it would not be set.

I suppose you could do `env TEST_UNDECLARED_OUTPUTS_DIR=/tmp bazel-bin/render_gl/thread_test` in that case.
</blockquote></details>

I'm sill pursuing the question of -- why doesn't the OpenGl context have any data path (now that the contexts are reliable).

jwnimmer-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-:-NVKeKgr0DCsTjUnUTPM:b-pq7hn8)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVKezk01b1FsuAQV4Wg:-NVKezk01b1FsuAQV4Wh:b9lka55) (no related file):*
Working

The use of of XInitThreads when in single-threaded simulations has the possibility of slowing things down (i.e., a performance regression for most users today).  I need to take some benchmarks to assess.

___
___
*[`geometry/render_gl/test/thread_test.cc` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUNizHn4sCiqJ7qzscS:-NVKedKs0LxoKv3aaxWe:b-f1992m) ([raw file](https://github.com/RobotLocomotion/drake/blob/95de376809ba7b83513f4ea263b4f8090dd33a91/geometry/render_gl/test/thread_test.cc#L81)):*

If you'd like to see the solution to this problem (sneak preview), you can see it here:

https://github.com/RobotLocomotion/drake/commit/9344b17cbdca5c6124cf1f066f3e2eb1425d02e1

I'm considering about submitting it in a PR while I figure out the black screen problem; but it'll require a new test. It'll be a very stripped down version of `thread_test_2.cc` found here:

https://github.com/RobotLocomotion/drake/commit/93220c8067826c8e7a81ea92b2c67f1a0582117d

</blockquote></details>

Done (via TODO).

The current status is good enough to enable threading correctness tests and benchmarking. We have ideas for how to fill in this pothole later.
___
*[`geometry/render_gl/test/thread_test.cc` line 191 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NUOVIV0ARu6q6vrYjRt:-NVKenoZEMVGjOH6CisH:bdlkiuh) ([raw file](https://github.com/RobotLocomotion/drake/blob/a0f829630d7b5e4414f00b208b0c1cac0ce059a2/geometry/render_gl/test/thread_test.cc#L191)):*

I'm sill pursuing the question of -- why doesn't the OpenGl context have any data path (now that the contexts are reliable). 
</blockquote></details>

Done (via #19398, and related adjustments here).

jwnimmer-tri (commented):
Ready for full feature review.  I think it's OK to land this even with the "need to throw away a context" hack.  We can remove that hack as part of your PR that resolves the root cause.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-:-NVab8Gl5bOvaUBV6AqQ:b-nxd40c)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVKezk01b1FsuAQV4Wg:-NVaYFTrB3NAy9ySKN3V:bv7bgjv) (no related file):*

Working

The use of of XInitThreads when in single-threaded simulations has the possibility of slowing things down (i.e., a performance regression for most users today).  I need to take some benchmarks to assess.
</blockquote></details>

Ok.

I pounded on `RenderBenchmark/GlColor/1/1/640/480` a *lot*.  I ensured that CPU turbo was set to off, CPU governor was set to performance, and the program was taskset-pinned to the same CPU throughout. The benchmark used 100x repetitions of that case, and each repetition run ~900 iterations internally.

I compared the master branch to this PR branch.

About 50% of the times I ran it, there was no statistically different result.

The other 50% of the times I ran it, the *second* branch I tested was statistically significant ~3-4% slower.  This was irrespective of whether I tested master then PR, or PR then master.

My conclusions from that are that:

(1) At most, this change is a ~3% performance loss.  Even if that loss is real, it seems a worthwhile trade-off versus the ability to use threads at all ever.

(2) Most likely, there is GPU boosting or throttling happening that is not being controlled by the benchmarking fixtures.  (This would explain the first-vs-second run measures.)

As such, I think this PR is okay to merge.

___

SeanCurtis-TRI (commented):
:LGTM: With some suggestions about shifting some of the code around.

> Reviewed 1 of 3 files at r1, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-:-NVezlNUBmC8Nlkv9vMF:b-e7lmwb)* status: 5 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 53 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVesTEFBTcnPrM2yz7f:-NVesTEFBTcnPrM2yz7g:b29fpf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L53)):*
> ```cpp
> using systems::Context;
> using systems::Diagram;
> using systems::System;
> ```

nit: unused
___
*[`geometry/render_gl/test/thread_test.cc` line 85 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVewrvcBJjVcgTNh7iX:-NVewrvcBJjVcgTNh7iY:b-kr9091) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L85)):*
> ```cpp
> 
>   // Create the plant and scene_graph.
>   auto builder = std::make_unique<RobotDiagramBuilder<double>>();
> ```

nit: Overwrought?



_Suggestion:_
```cpp
 RobotDiagramBuilder<double> builder;
```
___
*[`geometry/render_gl/test/thread_test.cc` line 135 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVeyXru2XTPgHRa1tVf:-NVeyXru2XTPgHRa1tVg:bhxt3c9) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L135)):*
> ```cpp
>     try {
>       for (int j = 0; j < num_repeats; ++j) {
>         // Invalidate the state to force re-calculation of the output.
> ```

BTW I'd propose hoisting everything *except* the render call out of the for loop (and out of the try/catch block) so that the only exceptions that get gracefully caught and accounted for are rendering exceptions. I did this in my own variations.
___
*[`geometry/render_gl/test/thread_test.cc` line 145 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVexNa75rehtFnsZzvM:-NVexNa75rehtFnsZzvN:b-vyojht) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L145)):*
> ```cpp
>                     scene_graph.GetMyContextFromRoot(diagram_context));
>         // Obtain our uniquely-owned RenderEngine (and so, we can const_cast).
>         RenderEngine* const renderer = const_cast<RenderEngine*>(
> ```

Sneaking out the `RenderEngine` feels a bit subversive. As long as we're doing this test, we should probably exercise the API we most expect to be exercised (i.e., the `QueryObject` API). The following diff respells it using that API.

```diff
--- a/geometry/render_gl/test/thread_test.cc
+++ b/geometry/render_gl/test/thread_test.cc
@@ -84,6 +84,8 @@ TEST_P(ThreadTest, Run) {
   // Create the plant and scene_graph.
   auto builder = std::make_unique<RobotDiagramBuilder<double>>();
 
+  const FrameId world_frame_id = builder->scene_graph().world_frame_id();
+
   if (async_mode != kNone) {
     // Add a never-used render engine. Because adding geometry activates the GL
     // context, the first render engine we create ends up crashing when used in
@@ -127,9 +129,15 @@ TEST_P(ThreadTest, Run) {
   MatrixX<ImageRgba8U> images(num_workers, num_repeats);
   std::atomic<int> num_errors{0};
   auto work = [&images, &num_errors, num_repeats = num_repeats, &renderer_names,
-               &X_WC, &camera_params, &const_diagram, &contexts](int i) {
+               &X_WC, &camera_params, &const_diagram, &contexts,
+               world_frame_id](int i) {
     Context<double>& diagram_context = *contexts.at(i);
     const std::string& renderer_name = renderer_names.at(i);
+    const RenderCameraCore& core = camera_params.core();
+    const ColorRenderCamera thread_cam(
+        {renderer_name, core.intrinsics(), core.clipping(),
+         core.sensor_pose_in_camera_body()},
+        camera_params.show_window());
     try {
       for (int j = 0; j < num_repeats; ++j) {
         // Invalidate the state to force re-calculation of the output.
@@ -141,17 +149,11 @@ TEST_P(ThreadTest, Run) {
             scene_graph.GetOutputPort("query")
                 .template Eval<QueryObject<double>>(
                     scene_graph.GetMyContextFromRoot(diagram_context));
-        // Obtain our uniquely-owned RenderEngine (and so, we can const_cast).
-        RenderEngine* const renderer = const_cast<RenderEngine*>(
-            query_object.GetRenderEngineByName(renderer_name));
-        DRAKE_THROW_UNLESS(renderer != nullptr);
         // Render an image.
         ImageRgba8U& image = images(i, j);
         image = ImageRgba8U(camera_params.core().intrinsics().width(),
                             camera_params.core().intrinsics().height());
-        renderer->UpdateViewpoint(
-            X_WC * camera_params.core().sensor_pose_in_camera_body());
-        renderer->RenderColorImage(camera_params, &image);
+        query_object.RenderColorImage(thread_cam, world_frame_id, X_WC, &image);
       }
     } catch (std::exception& e) {
       drake::log()->error("Worker {} exception: {}", i, e.what());
```
  
___
*[`geometry/render_gl/test/thread_test.cc` line 239 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVezDfw4faGRoSTgh4j:-NVezDfxE9v8F2gsdMM7:b2866j5) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L239)):*
> ```cpp
> 
> constexpr Params kParams[] = {
>     {.async_mode = kNone, .num_workers = 1, .num_repeats = 2},
> ```

My gut says having 2 workers, even for serial is a good thing as a reality check that all of the games you've played definitely work. I'm open to the argument that it is trivially so. But, either way, probably worth putting a note why serial execution *shouldn't* have multiple workers.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 135 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVeyXru2XTPgHRa1tVf:-NVfR-xk3S2Gcw2LOMk7:b-m593to) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L135)):*

BTW I'd propose hoisting everything *except* the render call out of the for loop (and out of the try/catch block) so that the only exceptions that get gracefully caught and accounted for are rendering exceptions. I did this in my own variations.
</blockquote></details>

Having an exception leak out a `std::thread` worker function is extremely bad juju -- obscure and hard to debug.  (I consider any thread-target function that doesn't have a global try-catch defective.)
___
*[`geometry/render_gl/test/thread_test.cc` line 145 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVexNa75rehtFnsZzvM:-NVfRgOyERJV3hoiALT1:b1xemo2) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L145)):*

Sneaking out the `RenderEngine` feels a bit subversive. As long as we're doing this test, we should probably exercise the API we most expect to be exercised (i.e., the `QueryObject` API). The following diff respells it using that API.

```diff
--- a/geometry/render_gl/test/thread_test.cc
+++ b/geometry/render_gl/test/thread_test.cc
@@ -84,6 +84,8 @@ TEST_P(ThreadTest, Run) {
   // Create the plant and scene_graph.
   auto builder = std::make_unique<RobotDiagramBuilder<double>>();
 
+  const FrameId world_frame_id = builder->scene_graph().world_frame_id();
+
   if (async_mode != kNone) {
     // Add a never-used render engine. Because adding geometry activates the GL
     // context, the first render engine we create ends up crashing when used in
@@ -127,9 +129,15 @@ TEST_P(ThreadTest, Run) {
   MatrixX<ImageRgba8U> images(num_workers, num_repeats);
   std::atomic<int> num_errors{0};
   auto work = [&images, &num_errors, num_repeats = num_repeats, &renderer_names,
-               &X_WC, &camera_params, &const_diagram, &contexts](int i) {
+               &X_WC, &camera_params, &const_diagram, &contexts,
+               world_frame_id](int i) {
     Context<double>& diagram_context = *contexts.at(i);
     const std::string& renderer_name = renderer_names.at(i);
+    const RenderCameraCore& core = camera_params.core();
+    const ColorRenderCamera thread_cam(
+        {renderer_name, core.intrinsics(), core.clipping(),
+         core.sensor_pose_in_camera_body()},
+        camera_params.show_window());
     try {
       for (int j = 0; j < num_repeats; ++j) {
         // Invalidate the state to force re-calculation of the output.
@@ -141,17 +149,11 @@ TEST_P(ThreadTest, Run) {
             scene_graph.GetOutputPort("query")
                 .template Eval<QueryObject<double>>(
                     scene_graph.GetMyContextFromRoot(diagram_context));
-        // Obtain our uniquely-owned RenderEngine (and so, we can const_cast).
-        RenderEngine* const renderer = const_cast<RenderEngine*>(
-            query_object.GetRenderEngineByName(renderer_name));
-        DRAKE_THROW_UNLESS(renderer != nullptr);
         // Render an image.
         ImageRgba8U& image = images(i, j);
         image = ImageRgba8U(camera_params.core().intrinsics().width(),
                             camera_params.core().intrinsics().height());
-        renderer->UpdateViewpoint(
-            X_WC * camera_params.core().sensor_pose_in_camera_body());
-        renderer->RenderColorImage(camera_params, &image);
+        query_object.RenderColorImage(thread_cam, world_frame_id, X_WC, &image);
       }
     } catch (std::exception& e) {
       drake::log()->error("Worker {} exception: {}", i, e.what());
```
  
</blockquote></details>

That depends on whether this test is scoped to find threading bugs in the GL render engine, or the whole geometry stack.  I was thinking more the former.  I imagine we'll have higher-layer acceptance tests elsewhere.
___
*[`geometry/render_gl/test/thread_test.cc` line 239 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVezDfw4faGRoSTgh4j:-NVfQkRT7X1xcznGCv-5:b-2vhvxh) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L239)):*

My gut says having 2 workers, even for serial is a good thing as a reality check that all of the games you've played definitely work. I'm open to the argument that it is trivially so. But, either way, probably worth putting a note why serial execution *shouldn't* have multiple workers.
</blockquote></details>

Sounds fine; bumped to 3.

BTW Notice now that the params are the same for all cases.  I think it's still working keep them as params; it was useful to play with these values easily while testing.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 239 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVezDfw4faGRoSTgh4j:-NVfSQ310XM8bKQw-a3f:b2p84cm) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L239)):*

Sounds fine; bumped to 3.

BTW Notice now that the params are the same for all cases.  I think it's still working keep them as params; it was useful to play with these values easily while testing.
</blockquote></details>

*worth keeping

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 135 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVeyXru2XTPgHRa1tVf:-NVj4Ffb6bau7VjS6EAt:b5ieqhc) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L135)):*

Having an exception leak out a `std::thread` worker function is extremely bad juju -- obscure and hard to debug.  (I consider any thread-target function that doesn't have a global try-catch defective.)
</blockquote></details>

Well, then I'd like to separate the two concepts.
You're doing work inside the for loop that doesn't need to be done inside the for loop (accessing the scene_graph and grabbing its query object, declaring and sizing the image -- although I'm open to the idea of wanting a "pure" starting point for each image justifies its creation inside).

Nevertheless, I'm marking myself as satisfied. 
___
*[`geometry/render_gl/test/thread_test.cc` line 145 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVexNa75rehtFnsZzvM:-NVj4oSj6JT9WYgE3Tjl:b9g83ws) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L145)):*

That depends on whether this test is scoped to find threading bugs in the GL render engine, or the whole geometry stack.  I was thinking more the former.  I imagine we'll have higher-layer acceptance tests elsewhere.
</blockquote></details>

I would *not* have inferred that purpose from the code. Introduction of `RobotDiagramBuilder`, `SceneGraph`, and system framework artifacts really obscures that fact. It's one of the reasons as I started do my debugging, I created my version of this test that was truly [focused solely on `RenderEngineGl`](https://github.com/SeanCurtis-TRI/drake/blob/WIP_multi_thread_render/geometry/render_gl/test/thread_test_2.cc). It's the same test, but eliminates everything that isn't strictly necessary.

You don't have to replace this test with that test, but it would be good to add some documentation to this file clarifying:

1. Why is it here?
2. Will it go away? (And under what circumstances.)

As to this second question, the code I'm currently working on to make `OpenGlContext` naturally thread safe comes with its own specific targeted unit tests (on both the context as well as the `RenderEngineGl` successfully using it). I feel that test would supplant this test. I'd note that *that* test has been so reduced, that it no longer has the serial-async-thread parameters (it simply uses thread). That reduction may preclude that test supplanting this test. Should I add it back in so that it *can* serve that purpose? We'll figure that out during review.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/render_gl/test/thread_test.cc` line 145 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NVexNa75rehtFnsZzvM:-NVjO6tHAryZ0y0NYdWf:b-4x6u45) ([raw file](https://github.com/RobotLocomotion/drake/blob/a94070cbd0deb3459288a3317c8146c726a409ce/geometry/render_gl/test/thread_test.cc#L145)):*
This is a good point.

I've shaved the current test down to be more like thread_test_2.  There's a TODO that I think your PR can fix.

I think both levels of testing will probably remain valuable, to help isolate faults when something regresses, so my suggestion is for your PR to also add a higher-layer test as a differently-named test case on the same fixture.

> ... no longer has the serial-async-thread parameters ...

I think it's important to test the `std::async`.  That's will be the primary use case.  Threads (`std::thread`) are dispreferred.  Testing both isn't that hard, so I think we should keep the 3-way enum characterization intact.

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-:-NW30ADV-Vt4XwO9OesS:bcav8r0)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

ggould-tri (commented):
:lgtm:

> Reviewed 1 of 3 files at r1, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-:-NW33dlj8G3AOwa4QzJW:bnfp4nl)* status: 1 unresolved discussion

___
*[`-- commits` line 2 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NW30QeK8DUthF5ddqyL):*
The release engineer is definitely going to need more info on "(sometimes)".
Also this is 'release notes: fix' but this message doesn't mention the issue being fixed; is there an issue number?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19323)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform)

___
*[`-- commits` line 2 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19323#-NW30QeK8DUthF5ddqyL):*
> The release engineer is definitely going to need more info on "(sometimes)".

I think the subject line is fine as-is.  Not every release note line needs to be self-contained and fully expository when read in isolation.  This snippet provides the gist of what's happening, and the release notes will link to the pull request where the user can go to read more.  ("Each thread must be given its own unique RenderEngineGl instance.")

> Also this is 'release notes: fix' but this message doesn't mention the issue being fixed; is there an issue number?

There is no issue number.  The (hopefully implied) fix is that multithreaded operation was unsupported, and is now supported.

PR #19337, opened 2023-05-03
[planning] Parallelizes intersection checks

wrangelvid:
AddRegions and AddEdgesBetweenRegions run time consuming intersection checks for all pairs of regions. When OMP is available, it will be called to run these loops in parallel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19337)

### DIFF
--- a/planning/trajectory_optimization/gcs_trajectory_optimization.cc
+++ b/planning/trajectory_optimization/gcs_trajectory_optimization.cc
@@ -11,6 +11,9 @@
 #include "drake/geometry/optimization/point.h"
 #include "drake/math/matrix_util.h"
 #include "drake/solvers/solve.h"
+#if defined(_OPENMP)
+#include <omp.h>
+#endif
 
 namespace drake {
 namespace planning {
@@ -261,7 +264,9 @@ EdgesBetweenSubgraphs::EdgesBetweenSubgraphs(
       std::make_shared<LinearEqualityConstraint>(
           M, VectorXd::Zero(num_positions()));
 
-  // TODO(wrangelvid) this can be parallelized.
+#if defined(_OPENMP)
+#pragma omp parallel for
+#endif
   for (int i = 0; i < from_subgraph.size(); ++i) {
     for (int j = 0; j < to_subgraph.size(); ++j) {
       if (from_subgraph.regions()[i]->IntersectsWith(
@@ -279,7 +284,6 @@ EdgesBetweenSubgraphs::EdgesBetweenSubgraphs(
         const Vertex& u = *from_subgraph.vertices_[i];
         const Vertex& v = *to_subgraph.vertices_[j];
         Edge* uv_edge = traj_opt_.AddEdge(u, v);
-        edges_.emplace_back(uv_edge);
 
         // Add path continuity constraints.
         uv_edge->AddConstraint(Binding<LinearEqualityConstraint>(
@@ -299,6 +303,11 @@ EdgesBetweenSubgraphs::EdgesBetweenSubgraphs(
             uv_edge->AddConstraint(Binding<Constraint>(constraint, vars));
           }
         }
+
+#if defined(_OPENMP)
+#pragma omp critical
+#endif
+        { edges_.emplace_back(uv_edge); }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


wrangelvid (commented):
@jwnimmer-tri I am not sure if that's the proper way of using OMP with drake. I am open to suggestions.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19337#-:-NUUl6A-AjV6mJQNhgbw:bkix9t4)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @wrangelvid)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19337)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),calderpg-tri, missing label for release notes (waiting on @calderpg-tri and @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 286 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19337#-NUYCSh3D6gI31TxFNW1:-NUYG_8T2QZNFZwXu-tn:b-6czvyl) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7e1b94cd47796de394b709132b71edf41f6ca31/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L286)):*

This line (AddEdge) is also a race condition, not just the emplace_back that already got relocated.

I assume the expensive part is IntersectsWith and RegionsConnectThroughSubspace above, not this constraint-copying code here and below?

Assuming that's true, a better formulation might be to first do the loop for IntersectsWith + RegionsConnectThroughSubspace and save that into a temporary data structure that's pre-allocated ahead of the loop, so no critical section is required.

Then once that's all calculated, go back and (serially) do this AddEdge and related code for anything that passed the expensive checks.
</blockquote></details>

In the alternative, we could somehow accumulate `(i, j, edge_constraints)` tuples as a kind of proto-Edge, and then later (serially) add those edges to the GCS object after all of the looping finishes.

The challenge there would be ensuring consistent behavior under different numbers of threads on users' machines. We don't want the order that the threads checked for edges to affect the GCS that comes out -- we want the same GCS object every time, no matter what OpenMP did. We'd need to be careful to add the proto-edges into GCS in some consistent order.  (Sorting them by `i,j` would suffice, but would not be linear time.)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19337)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),calderpg-tri, missing label for release notes (waiting on @calderpg-tri and @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 385 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19337#-NUYAf-m8f018HNC7ksc:-NUYSFtZ64HdWOy2tPqH:b-e7x379) ([raw file](https://github.com/RobotLocomotion/drake/blob/e7e1b94cd47796de394b709132b71edf41f6ca31/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L385)):*

Another approach would be to store region-region connections in a region-indexed manner first, then flatten. Something like:
```
  std::vector<std::vector<size_t>> connected_regions(regions.size());
#if defined(_OPENMP)
#pragma omp parallel for
#endif
  for (size_t i = 0; i < regions.size(); ++i) {
    for (size_t j = i + 1; j < regions.size(); ++j) {
      if (regions[i]->IntersectsWith(*regions[j])) {
        connected_regions[i].emplace_back(j);
      }
    }
  }

  std::vector<std::pair<int, int>> edges_between_regions;
  for (size_t i = 0; i < regions.size(); ++i) {
    const auto& connections = connected_regions[i];
    for (const auto j : connections) {
      edges_between_regions.emplace_back(i, j);
      edged_between_regions.emplace_back(j, i);
    }
  }
```
</blockquote></details>

Ah, this is a good point I didn't quite realize.

Just to underline it: we only need to allocate distinct storage for each `i`th index.  The for-j looping is all done on the same worker thread, so doesn't need that level of storage isolation.

Depending on what your benchmarking shows, I'll bet Calder's sample code here (along with the triangle loop trick) would be the best answer in terms of the storage representation.

wrangelvid:
Things have changed in master. The intersection checks have been moved to [CalcPairwiseIntersections](https://github.com/RobotLocomotion/drake/blob/ee96c093c0b131c00d527cd053d74bc8465a1c14/geometry/optimization/geodesic_convexity.cc#L231). I will close this PR for now. In the future, another alternative to Parallelizing with OMP could involve solving one large problem with all intersections and let the solver, like MOSEK, parallelize internally.

PR #19342, opened 2023-05-03
Mujoco mesh inertia dev

rpoyner-tri:


rpoyner-tri (commented):
Seeking high level design review from @RussTedrake @SeanCurtis-TRI . Is this on the right track? Am I misusing or misnaming things, or missing an abstraction somewhere?

If ya'll like it, I'll clean it up, add tests, and take the PR out of draft mode.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUXoFw71V6HZFSVbtGk:bsafh3v)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
from Sean f2f: yeah, this is dumb. There is a much smaller patch that should be fine. I'll work on drafting that.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUYUyTp5hqtBbmBdDdu:b-9iswzc)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
-(status: do not merge) -(status: do not review)  +(release notes: fix) 
+@RussTedrake for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUcTR0W6bV6Sx7lCtiJ:b-29q3al)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @rpoyner-tri and @RussTedrake)

rpoyner-tri (commented):
per Russ' request (too busy), turning to -@RussTedrake +@SeanCurtis-TRI for feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUccpI0-dfCuE4dF602:bprktlw)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
Issues with center of mass. I haven't looked at the tests yet. (We'll need a test for a mesh that *doesn't* have its origin at its geometric center.)

I also made a suggestion for simplifying the code. See below.

> Reviewed 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUcyOHt6t8kk1FWBpVe:b9lijvi)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUcwzBP7T-vsy-E8iLP:b-ni5zv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*
> ```cpp
>       }
>       if (mesh_.count(mesh)) {
>         geom.shape = mesh_.at(mesh)->Clone();
> ```

BTW Near as I can tell, all of these branches creates a `geometry::Shape`. The `CalcSpatialInertia(Shape)` API would almost be sufficient for everything you're doing. Simply create a shape in each branch and then, at the very end, simply ask for the `SpatialInertia` for the shape. Things would definitely get simpler.

The obstacle to doing that would be the fact that *that* API takes density and *you* could have either density or mass.

The solution:

 - Call `CalcSpatialInertia(Shape)` with unit density. This gives you a *candidate* for `M_GG_G`.
 - Figure out if you have a density value.
    - If you do, create a new `M_GG_G` with the same unit inertia and com, but multiply the mass by the new density.
- If, instead, you have mass, create a new `M_GG_G` by using the same unit inertia and com, but set the mass to the value at hand.
- Finally, do the transformation here to move it from `M_GG_G` to `M_GBo_B`.

That would simplify all of your mass value calculations to a single call -- however, you *would* have to make sure you don't invoke it if the type is `plane`.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUctJl55hI93Zl8WRQ5:b-oxf9dh) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L682)):*
> ```cpp
>         }
>       }
>       multibody::SpatialInertia<double> M_GG_G(mass, Vector3d::Zero(),
> ```

nit: This worries me. It puts the center of mass at the geometry origin. That is *not* generally true for meshes. It was valid for the old code, because it was the mass of a box whose CoM was at its geometric origin. But it won't generally be true.

So, you'll need to *capture* CoM in mesh and pipe it down here.

------------------------------------------

On a related note...

Reading the mujoco documentation has me a *bit* worried. It indicates that a mesh asset [is centered around (0,0,0) and its principal axes of inertia are the coordinate axes](https://mujoco.readthedocs.io/en/latest/XMLreference.html#asset-mesh) as part of the "compilation stage". The documentation makes *particular* reference to the __geometric center__ (and not the center of mass). Presumably we haven't been doing that (looking at l. 975, it appears not).

Fixing this behavior is outside the scope of this PR, but we should flag it.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUczfYR4ugtG1wFntmu:bkwes9z) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

BTW Near as I can tell, all of these branches creates a `geometry::Shape`. The `CalcSpatialInertia(Shape)` API would almost be sufficient for everything you're doing. Simply create a shape in each branch and then, at the very end, simply ask for the `SpatialInertia` for the shape. Things would definitely get simpler.

The obstacle to doing that would be the fact that *that* API takes density and *you* could have either density or mass.

The solution:

 - Call `CalcSpatialInertia(Shape)` with unit density. This gives you a *candidate* for `M_GG_G`.
 - Figure out if you have a density value.
    - If you do, create a new `M_GG_G` with the same unit inertia and com, but multiply the mass by the new density.
- If, instead, you have mass, create a new `M_GG_G` by using the same unit inertia and com, but set the mass to the value at hand.
- Finally, do the transformation here to move it from `M_GG_G` to `M_GBo_B`.

That would simplify all of your mass value calculations to a single call -- however, you *would* have to make sure you don't invoke it if the type is `plane`.
</blockquote></details>

I'm sympathetic, but that all feels like more than I signed up for.  However, now that you've pointed out some things, I do feel some pain for a 450-line method. Still thinking about what I want to do about all that.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUd07oW3YqVGWWpONxT:br0zmkp) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L682)):*

nit: This worries me. It puts the center of mass at the geometry origin. That is *not* generally true for meshes. It was valid for the old code, because it was the mass of a box whose CoM was at its geometric origin. But it won't generally be true.

So, you'll need to *capture* CoM in mesh and pipe it down here.

------------------------------------------

On a related note...

Reading the mujoco documentation has me a *bit* worried. It indicates that a mesh asset [is centered around (0,0,0) and its principal axes of inertia are the coordinate axes](https://mujoco.readthedocs.io/en/latest/XMLreference.html#asset-mesh) as part of the "compilation stage". The documentation makes *particular* reference to the __geometric center__ (and not the center of mass). Presumably we haven't been doing that (looking at l. 975, it appears not).

Fixing this behavior is outside the scope of this PR, but we should flag it.
</blockquote></details>

I read that same passage in the mujoco docs, and I am more confused. It appears that passage refers to the practice of constructing meshes directly in xml from vertices only, rather than from mesh files. Perhaps there is no difference? I would be curious if you get more clarity than I did on a revisit.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUghRN88Yv5ZhEt9jzc:b-woqt6x) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

I'm fine with that not happening, but you will have to carve out capturing the *one* center of mass for the mesh and transport it down below. The alternate approach I'm suggesting would actually have a net reduction in lines of code (I think) and no special cases. If you'd like, I can take a peek at it in the morning to validate my hypothesis. Right now, it's more a hunch.
</blockquote></details>

I dabbled this morning and produced https://github.com/rpoyner-tri/drake/pull/3.

The net result is a bigger (and farther reaching) change than I'd originally imagined. If you like it, I can PR the changes to `ShapeReifier` independently. The only *real* take away is the change to the test re: Mesh center of mass.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUgtrs_3zFB0yXrPEfi:b2k9rhr) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

I dabbled this morning and produced https://github.com/rpoyner-tri/drake/pull/3.

The net result is a bigger (and farther reaching) change than I'd originally imagined. If you like it, I can PR the changes to `ShapeReifier` independently. The only *real* take away is the change to the test re: Mesh center of mass.
</blockquote></details>

Cool. I found and flipped the reviewable.io bits for my fork; I'll read the PR a bit more closely.

Tentatively, I'd say let's just roll with your proposal: land the ShapeReifier PR separately, and then add your mujoco changes here with co-author credit.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUguLrTCwSA9lUfY3Tx:blkqeyv) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L682)):*

I'm going to move the conversation on mesh post-processing out of this thread and make its own conversation above.

I *am* going to double down on the fact that this is the wrong center of mass for a general mesh, regardless of what Mujoco may or may not do with mesh assets.
</blockquote></details>

Agree with your concerns. Close this thread and discuss elsewhere?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUh6IMI4cwur9wpz4-R:bn21gna) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

Cool. I found and flipped the reviewable.io bits for my fork; I'll read the PR a bit more closely.

Tentatively, I'd say let's just roll with your proposal: land the ShapeReifier PR separately, and then add your mujoco changes here with co-author credit.
</blockquote></details>

Ok, let me extract the `ShapeReifier` changes out and submit a PR.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUh5zbi9v_5E5cuWV_4:bkbgf64) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L682)):*

Agree with your concerns. Close this thread and discuss elsewhere?
</blockquote></details>

Let's leave this open until we have a confirmed fix to the mesh center of mass alive in the PR. :)

(I've started reading the mujoco doc a bit more carefully and have started crafting an executive summary....but I keep tripping over apparent inconsistencies, so I'm going to sit on it until I'm convinced they're not attributable to PEBKAC.)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUh7CR59D-b1JSNhLFh:b-8fx2u7) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

Ok, let me extract the `ShapeReifier` changes out and submit a PR.
</blockquote></details>

:+1:

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUh7Nox15A7KuUUZWCK:b-8fx2u7) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L682)):*

Let's leave this open until we have a confirmed fix to the mesh center of mass alive in the PR. :)

(I've started reading the mujoco doc a bit more carefully and have started crafting an executive summary....but I keep tripping over apparent inconsistencies, so I'm going to sit on it until I'm convinced they're not attributable to PEBKAC.)
</blockquote></details>

:+1:

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUcwzBP7T-vsy-E8iLO:-NUh8fkr5sJc2bg1zMAM:buf1mk3) ([raw file](https://github.com/RobotLocomotion/drake/blob/21812541f63cabf26fb3fd616eeb4786e7977864/multibody/parsing/detail_mujoco_parser.cc#L556)):*

:+1: 
</blockquote></details>

See #19367.

rpoyner-tri (commented):
This is now rebased to newer master.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUw0NcX4UaDf6UTtaQf:bdhm5z)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
And the PR has been rebased onto this.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NUx7Q3g7jZU6L7vmfjv:b-zbpydl)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @rpoyner-tri and @SeanCurtis-TRI)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 682 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUctJl55hI93Zl8WRQ4:-NUx_cxT0ApuBP_h7cyL:b7hemi) ([raw file](https://github.com/RobotLocomotion/drake/blob/c6d85ca56ff245812c541724fc46d7fce1dadba2/multibody/parsing/detail_mujoco_parser.cc#L682)):*

:+1: 
</blockquote></details>

I've squashed in your changes, made a new draft of commit message with co-author credit. It's a good time to read the whole thing again as feature/co-author.

SeanCurtis-TRI (commented):
:LGTM: With a couple of tweaks correcting my code.

> Reviewed 3 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NV-tW0VCk4HWM27rLcV:b4v1ykv)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 352 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-pt097GK3eqt6yAL_:-NV-pt097GK3eqt6yALa:bxbt80u) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L352)):*
> ```cpp
>     InertiaCalculator(
>         const std::map<std::string, SpatialInertia<double>>* mesh_inertia,
>         std::string name)
> ```

BTW It feels like it would be nice if this could be a `std::string_view` instead of a `std::string`, however, that is incompatible with using it as a look up in the map.

The copy isn't the *end* of the world, but as we're already aliasing the map of spatial inertias, we might consider aliasing the mesh name as well. Thoughts?
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 358 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-qD1tCDi4Srh9HBUA:-NV-qD1u1A_Vv_bonaAo:b-zbpjnc) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L358)):*
> ```cpp
>     }
> 
>     SpatialInertia<double> spatial_inertia(const geometry::Shape& shape) {
> ```

nit: Oops on the naming here. Doesn't get to be camel_case and is missing a verb.

It could be `CalcSpatialInertia`, or it could simply be `Calc` because the call site looks like:

```c++
InertiaCalculator(&mesh_, name).Calc(shape);
```

SeanCurtis-TRI:
FYI I've officially decided that mujoco is insane. Or, at least its documentation is.
From [here](https://mujoco.readthedocs.io/en/latest/XMLreference.html#asset-mesh):

> Positioning and orienting is complicated by the fact that vertex data are often designed relative to coordinate frames whose origin is not inside the mesh. In contrast, MuJoCo expects the __origin of a geom’s local frame to coincide with the geometric center of the shape.__ We resolve this discrepancy by pre-processing the mesh in the compiler, so that it is centered around (0,0,0) and its principal axes of inertia are the coordinate axes. (emphasis mine)

Some thoughts:
1. "geometric center of the shape" (also known as "centroid" or "center of figure") is equivalent to the center of mass when the density is uniform.  So, in the mesh's frame G, the center of mass should always be at (0, 0, 0).
2. It should also be aligned such that the principle axes are aligned with Gx, Gy, Gz (the basis of geometry frame G).
3. The transforms necessary to achieve this get stored with the *asset* and they are combined with the `geometry`instantiation to produce the final pose.
4. We're definitely not doing this right now.

Now, the fun and games. Further along in the *same* document:

> Fortunately most meshes used in robot models are designed in a coordinate frame centered at the joint. This makes the corresponding MJCF model intuitive: we set the body frame at the joint, so that the joint position is (0,0,0) in the body frame, and simply reference the mesh. 

"simply reference the mesh" is accompanied by an example of mujoco xml showing ` <geom>` tag referencing a mesh without any additional posing. As described, the mesh's origin was defined to be at the joint origin. Combined with the text earlier in the paragraph, that also implies that the *center of mass* of the mesh is also at the joint origin which is highly unlikely.

I'd say the text is inconclusive. I'm not sure how to define it. For our purposes, we define our `SpatialInertia` of the geometry about the *body origin* and expressed in the body frame. For now, that's got to be as good as we can do without further clarity.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NUdVoMREuQuev-t31fA:-NV02F8y6R1l4VvQR8d_:bnkj7yz) (no related file):*

FYI I've officially decided that mujoco is insane. Or, at least its documentation is.
From [here](https://mujoco.readthedocs.io/en/latest/XMLreference.html#asset-mesh):

> Positioning and orienting is complicated by the fact that vertex data are often designed relative to coordinate frames whose origin is not inside the mesh. In contrast, MuJoCo expects the __origin of a geom’s local frame to coincide with the geometric center of the shape.__ We resolve this discrepancy by pre-processing the mesh in the compiler, so that it is centered around (0,0,0) and its principal axes of inertia are the coordinate axes. (emphasis mine)

Some thoughts:
1. "geometric center of the shape" (also known as "centroid" or "center of figure") is equivalent to the center of mass when the density is uniform.  So, in the mesh's frame G, the center of mass should always be at (0, 0, 0).
2. It should also be aligned such that the principle axes are aligned with Gx, Gy, Gz (the basis of geometry frame G).
3. The transforms necessary to achieve this get stored with the *asset* and they are combined with the `geometry`instantiation to produce the final pose.
4. We're definitely not doing this right now.

Now, the fun and games. Further along in the *same* document:

> Fortunately most meshes used in robot models are designed in a coordinate frame centered at the joint. This makes the corresponding MJCF model intuitive: we set the body frame at the joint, so that the joint position is (0,0,0) in the body frame, and simply reference the mesh. 

"simply reference the mesh" is accompanied by an example of mujoco xml showing ` <geom>` tag referencing a mesh without any additional posing. As described, the mesh's origin was defined to be at the joint origin. Combined with the text earlier in the paragraph, that also implies that the *center of mass* of the mesh is also at the joint origin which is highly unlikely.

I'd say the text is inconclusive. I'm not sure how to define it. For our purposes, we define our `SpatialInertia` of the geometry about the *body origin* and expressed in the body frame. For now, that's got to be as good as we can do without further clarity.

</blockquote></details>

Whee. Thanks for the further close reading.

___
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 352 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-pt097GK3eqt6yAL_:-NV02PSw8FsbIb0ZDxzB:bkjhllc) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L352)):*

BTW It feels like it would be nice if this could be a `std::string_view` instead of a `std::string`, however, that is incompatible with using it as a look up in the map.

The copy isn't the *end* of the world, but as we're already aliasing the map of spatial inertias, we might consider aliasing the mesh name as well. Thoughts?
</blockquote></details>

If you are going to rely on aliases, the actual storage needs to have an owner with appropriate lifetime. Sounds like that might cost you more complexity. If we ever care about the performance of parsers (unlikely, given that we are tied to XML), we can revisit.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 358 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-qD1tCDi4Srh9HBUA:-NV02laU3GwEIRGljJMs:b-vd1qoa) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L358)):*

nit: Oops on the naming here. Doesn't get to be camel_case and is missing a verb.

It could be `CalcSpatialInertia`, or it could simply be `Calc` because the call site looks like:

```c++
InertiaCalculator(&mesh_, name).Calc(shape);
```
</blockquote></details>

I'll fix up the names somehow.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 352 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-pt097GK3eqt6yAL_:-NV0Z6k24wI4ajn59VSm:b-nwxs2r) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L352)):*

If you are going to rely on aliases, the actual storage needs to have an owner with appropriate lifetime. Sounds like that might cost you more complexity. If we ever care about the performance of parsers (unlikely, given that we are tied to XML), we can revisit.
</blockquote></details>

In for a penny, in for a pound?

It might be simpler if we get rid of all aliases and simply pass in references to the mesh name, mesh inertias, *and* shape in the `CalcWhatever()` method.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 352 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-pt097GK3eqt6yAL_:-NV10PF5A9lDQicS2pd0:b1kkhgi) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L352)):*

In for a penny, in for a pound?

It might be simpler if we get rid of all aliases and simply pass in references to the mesh name, mesh inertias, *and* shape in the `CalcWhatever()` method.
</blockquote></details>

I guess I'm open to proposals, but I'm fine with what is there now.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 352 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NV-pt097GK3eqt6yAL_:-NV1E5FN6BrdGn1iZYEY:b-vab2jh) ([raw file](https://github.com/RobotLocomotion/drake/blob/2f3420ca69867f0cd6b7d52f6c8f0b14e3cc81c8/multibody/parsing/detail_mujoco_parser.cc#L352)):*

I guess I'm open to proposals, but I'm fine with what is there now.
</blockquote></details>

Well, I'll stand by the BTW. I'm content to let it go and we'll let platform review weigh in on whether they have an opinion or not. As it is in an anonymous namespace, subsequent changes are painless.

rpoyner-tri (commented):
+@sammy-tri for platform review, if there is time today.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NVBPL6O3z-9n1XIetGT:bclmeew)* status: LGTM missing from assignee sammy-tri(platform) (waiting on @sammy-tri)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: LGTM missing from assignee sammy-tri(platform) (waiting on @sammy-tri)

sammy-tri (commented):
:lgtm: (modulo one nit)

> Reviewed 3 of 5 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-:-NVBpNT592yEblR4w7W8:b-nevw6d)* status: 1 unresolved discussion

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 381 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NVBosr52dsZZP0_Dzhm:-NVBosr52dsZZP0_Dzhn:b655lda) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b793e595dfc1ce8c1f8efa834c7c0a550cf378a/multibody/parsing/detail_mujoco_parser.cc#L381)):*
> ```cpp
>     const std::map<std::string, SpatialInertia<double>>& mesh_inertia_;
>     std::string name_;
>     SpatialInertia<double> M_GG_G;
> ```

nit: no trailing underscore on member field

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19342)* status: 1 unresolved discussion

___
*[`multibody/parsing/detail_mujoco_parser.cc` line 381 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19342#-NVBosr52dsZZP0_Dzhm:-NVBpURa0xRxBMxUOhVJ:b6yzc1s) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b793e595dfc1ce8c1f8efa834c7c0a550cf378a/multibody/parsing/detail_mujoco_parser.cc#L381)):*

nit: no trailing underscore on member field
</blockquote></details>

ah, good catch. Feels like a bit of an eye test to me these days.

PR #18983, opened 2023-03-13
[multibody] Add joint locking feature to SAP

joemasterjohn:
Implements functionality for joint locking when SAP is the underlying discrete solver. Prior to this PR, the locked/unlocked status of a joint was ignored when using SAP.

Now joint locking with function as intended: whatever configuration the joint was in when `joint.Lock(&context)` is called will remain constant in every subsequent timestep until `joint.Unlock(&context)` is called. Further, joint velocities for the locked joint will remain exactly 0 for every timestep while the joint is locked.

This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/18983)

PR #19349, opened 2023-05-04
Filter contact results for completely locked trees

joemasterjohn:
This change serves as a prerequisite to joint locking in SAP #18983.

Because SAP uses sparse constraint Jacobians the possibility exists for a contact constraint Jacobian to become degenerate (0-dimensional) when all DOFs of participating trees within the tree topology are locked from joint locking. This occurs becuase joint locking uses the strategy of projecting the contact problem to a reduced DOF problem to lock DOFs.

To alleviate the problem, this PR filters out `DiscreteContactPairs` corresponding to geometries that are part of completely locked trees. This has the effect of removing the corresponding `ContactResults` for these pairs. Geometry computed contact data for these types of pairs (`PenetrationAsPointPair` and `ContactSurface`) are not affected and still exist.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)

### DIFF
--- a/multibody/plant/compliant_contact_manager.cc
+++ b/multibody/plant/compliant_contact_manager.cc
@@ -348,6 +348,10 @@ void CompliantContactManager<T>::AppendDiscreteContactPairsForPointContact(
           .template Eval<geometry::QueryObject<T>>(context);
   const geometry::SceneGraphInspector<T>& inspector = query_object.inspector();
 
+  const std::vector<std::vector<int>>& per_tree_unlocked_indices =
+      this->EvalUnlockedVelocityIndicesPerTree(context);
+  const MultibodyTreeTopology& topology = this->internal_tree().get_topology();
+
   // Fill in the point contact pairs.
   const std::vector<PenetrationAsPointPair<T>>& point_pairs =
       plant().EvalPointPairPenetrations(context);
@@ -359,39 +363,51 @@ void CompliantContactManager<T>::AppendDiscreteContactPairsForPointContact(
         this->geometry_id_to_body_index().at(pair.id_B);
     const Body<T>& body_B = plant().get_body(body_B_index);
 
-    const T kA = GetPointContactStiffness(
-        pair.id_A, this->default_contact_stiffness(), inspector);
-    const T kB = GetPointContactStiffness(
-        pair.id_B, this->default_contact_stiffness(), inspector);
-    const T k = GetCombinedPointContactStiffness(
-        pair.id_A, pair.id_B, this->default_contact_stiffness(), inspector);
+    const TreeIndex& treeA_index = topology.body_to_tree_index(body_A_index);
+    const TreeIndex& treeB_index = topology.body_to_tree_index(body_B_index);
 
-    // Hunt & Crossley dissipation. Used by TAMSI, ignored by SAP.
-    const T d = GetCombinedHuntCrossleyDissipation(
-        pair.id_A, pair.id_B, kA, kB, this->default_contact_dissipation(),
-        inspector);
+    //  For joint locking, filter out contacts between bodies who belong to
+    //  trees with 0 degrees of freedom. For a contact to remain in
+    //  consideration, at least one of the trees involved has to be valid and
+    //  have a non-zero number of DOFs.
+    if ((treeA_index.is_valid() &&
+         per_tree_unlocked_indices[treeA_index].size() != 0) ||
+        (treeB_index.is_valid() &&
+         per_tree_unlocked_indices[treeB_index].size() != 0)) {
+      const T kA = GetPointContactStiffness(
+          pair.id_A, this->default_contact_stiffness(), inspector);
+      const T kB = GetPointContactStiffness(
+          pair.id_B, this->default_contact_stiffness(), inspector);
+      const T k = GetCombinedPointContactStiffness(
+          pair.id_A, pair.id_B, this->default_contact_stiffness(), inspector);
 
-    // Dissipation time scale. Used by SAP, ignored by TAMSI.
-    const double default_dissipation_time_constant = 0.1;
-    const T tau = GetCombinedDissipationTimeConstant(
-        pair.id_A, pair.id_B, default_dissipation_time_constant, body_A.name(),
-        body_B.name(), inspector);
-    const T mu =
-        GetCombinedDynamicCoulombFriction(pair.id_A, pair.id_B, inspector);
+      // Hunt & Crossley dissipation. Used by TAMSI, ignored by SAP.
+      const T d = GetCombinedHuntCrossleyDissipation(
+          pair.id_A, pair.id_B, kA, kB, this->default_contact_dissipation(),
+          inspector);
 
-    // We compute the position of the point contact based on Hertz's theory
-    // for contact between two elastic bodies.
-    const T denom = kA + kB;
-    const T wA = (denom == 0 ? 0.5 : kA / denom);
-    const T wB = (denom == 0 ? 0.5 : kB / denom);
-    const Vector3<T> p_WC = wA * pair.p_WCa + wB * pair.p_WCb;
+      // Dissipation time scale. Used by SAP, ignored by TAMSI.
+      const double default_dissipation_time_constant = 0.1;
+      const T tau = GetCombinedDissipationTimeConstant(
+          pair.id_A, pair.id_B, default_dissipation_time_constant,
+          body_A.name(), body_B.name(), inspector);
+      const T mu =
+          GetCombinedDynamicCoulombFriction(pair.id_A, pair.id_B, inspector);
 
-    const T phi0 = -pair.depth;
-    const T fn0 = k * pair.depth;  // Used by TAMSI, ignored by SAP.
+      // We compute the position of the point contact based on Hertz's theory
+      // for contact between two elastic bodies.
+      const T denom = kA + kB;
+      const T wA = (denom == 0 ? 0.5 : kA / denom);
+      const T wB = (denom == 0 ? 0.5 : kB / denom);
+      const Vector3<T> p_WC = wA * pair.p_WCa + wB * pair.p_WCb;
 
-    contact_pairs.push_back({pair.id_A, pair.id_B, p_WC, pair.nhat_BA_W, phi0,
-                             fn0, k, d, tau, mu, {} /* no surface index */,
-                             {} /* no face index */});
+      const T phi0 = -pair.depth;
+      const T fn0 = k * pair.depth;  // Used by TAMSI, ignored by SAP.
+
+      contact_pairs.push_back({pair.id_A, pair.id_B, p_WC, pair.nhat_BA_W, phi0,
+                              fn0, k, d, tau, mu, {} /* no surface index */,
+                              {} /* no face index */});
+    }
   }
 }
 
@@ -426,6 +442,10 @@ void CompliantContactManager<T>::
   const std::vector<geometry::ContactSurface<T>>& surfaces =
       this->EvalContactSurfaces(context);
 
+  const std::vector<std::vector<int>>& per_tree_unlocked_indices =
+      this->EvalUnlockedVelocityIndicesPerTree(context);
+  const MultibodyTreeTopology& topology = this->internal_tree().get_topology();
+
   const int num_surfaces = surfaces.size();
   for (int surface_index = 0; surface_index < num_surfaces; ++surface_index) {
     const auto& s = surfaces[surface_index];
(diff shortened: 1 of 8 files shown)


amcastro-tri (commented):
+@amcastro-tri for feature review

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-:-NUbaLQH22KERK1Px6uH:b-2qhvtz)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

joemasterjohn (commented):
Working through the CI failures now.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-:-NUbobKnEI4ObGTrDp01:b-5l5r0t)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri and @joemasterjohn)

amcastro-tri (commented):
> Reviewed 6 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 13 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/plant/multibody_plant.h` line 4840 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc3DZ_75LcuhmJcwiq:-NUc3DZ_75LcuhmJcwir:b-ybz2m8) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.h#L4840)):*
> ```objc
>   // ascending order, in the range [0, tree_M.num_velocities()), with the
>   // indices of the locked velocities removed.
>   void CalcJointLockingIndicesPerTree(
> ```

minor, for this method and the other ones in this family, maybe name them so that it is obvious what they return? 
That is if I name it `CalcUnlockedJointIndicesPerTree()` I do know that the return corresponds to the unlocked velocities.
___
*[`multibody/plant/multibody_plant.h` line 4842 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc3bNc6EWo7U8z5WQ7:-NUc3bNc6EWo7U8z5WQ8:bnq2tm8) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.h#L4842)):*
> ```objc
>   void CalcJointLockingIndicesPerTree(
>       const systems::Context<T>& context,
>       std::vector<std::vector<int>>* unlocked_velocity_indices) const;
> ```

I'd imagine `unlocked_velocity_indices.size()` equals the number of trees? and it is resized accordingly on output? 
mention this here in the doc.
___
*[`multibody/plant/multibody_plant.cc` line 2308 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc5rgj9kSgbGoeOQMI:-NUc5rgj9kSgbGoeOQMJ:b-qfdcpf) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2308)):*
> ```cpp
> 
> template <typename T>
> void MultibodyPlant<T>::CalcJointLockingIndices(
> ```

given this implementation, I'd rename this method to `CalcUnlockedVelocityIndices()`
___
*[`multibody/plant/multibody_plant.cc` line 2342 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc45EP7i6A9rE9m0eu:-NUc45EP7i6A9rE9m0ev:b-lx3qzj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2342)):*
> ```cpp
>   DRAKE_DEMAND(unlocked_velocity_indices != nullptr);
> 
>   const auto& joint_locking_indices = this->EvalJointLockingIndices(context);
> ```

nit, maybe rename to `unlocked_joint_indices`?
but, this evaluates to velocity indices right? so maybe `unlocked_velocity_indices`?
___
*[`multibody/plant/multibody_plant.cc` line 2350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc59nx4RnhiaPU8fCa:-NUc59nx4RnhiaPU8fCb:b-mtkemp) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2350)):*
> ```cpp
>   internal::TreeIndex tree_index{0};
>   while (joint_index < static_cast<int>(joint_locking_indices.size())) {
>     while (joint_locking_indices[joint_index] >=
> ```

here I think you can do:
```c++
for (int dof : unlocked_velocity_indices) {
  const TreeIndex tree = topology.velocity_to_tree_index(dof);
  const int tree_dof = dof - topology.tree_velocities_start(tree);
  indices[tree].push_back(tree_dof);
}
```
___
*[`multibody/plant/multibody_plant.cc` line 2899 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc6qTz3yZkYoAQle0c:-NUc6qTz3yZkYoAQle0d:b7636g1) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2899)):*
> ```cpp
>       "Joint Locking Per Tree Indices.", std::vector<std::vector<int>>(),
>       &MultibodyPlant::CalcJointLockingIndicesPerTree,
>       {this->all_parameters_ticket()});
> ```

why the dependence on parameters ticket?
___
*[`multibody/plant/test/joint_locking_test.cc` line 341 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc-fadERomgwJGTZCf:-NUc-fadERomgwJGTZCg:bxfg59) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L341)):*
> ```cpp
>   }
>   return out;
> }
> ```

minor, I'd probably add a blank line
___
*[`multibody/plant/test/joint_locking_test.cc` line 378 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc0kwGE3yKLl43uNWZ:-NUc0kwGE3yKLl43uNW_:b8sfrup) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L378)):*
> ```cpp
>     // Pose of sphere geometry S in body frame B.
>     const RigidTransformd X_BS = RigidTransformd::Identity();
>     // Set material properties for hydroelastics.
> ```

minor, well, these prooperties also affect point contact. So maybe just say "set material properties"?
___
*[`multibody/plant/test/joint_locking_test.cc` line 384 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc0r5BBIc-ESLt7rep:-NUc0r5BBIc-ESLt7req:brspkaf) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L384)):*
> ```cpp
>         CoulombFriction<double>{1.0, 1.0} /* friction */, &ball_props);
> 
>     if (config.contact_model == ContactModel::kHydroelastic) {
> ```

minor, you could probably just remove the `if`. If hydro is not used, then this property won't be used.
___
*[`multibody/plant/test/joint_locking_test.cc` line 410 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc1HzvBsJHv7hVPqTM:-NUc1HzvBsJHv7hVPqTN:bg0mu1n) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L410)):*
> ```cpp
> // resulting in degenerate contact Jacobians. We eliminate all contacts between
> // bodies where both bodies' parent trees have all of their DOFs locked from
> // joint locking.
> ```

While this is true, probably a simpler statement, like the one a user would be interested in, would read "Test that contact results are properly reported. MultibodyPlant will not report contact results between bodies that are either anchored or with all of their dofs locked."
___
*[`multibody/plant/test/joint_locking_test.cc` line 431 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc22ySEArkx1HWTi2g:-NUc22ySEArkx1HWTi2h:bmjcxp9) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L431)):*
> ```cpp
>     } else {
>       EXPECT_EQ(results.num_point_pair_contacts(), 0);
>       EXPECT_EQ(results.num_hydroelastic_contacts(), 1);
> ```

interesting, why a single hydro contact? maybe just `EXPECT_GT(num_hydro_contacts(), 0)`? since the actual number is irrelevant really.
___
*[`multibody/plant/test/joint_locking_test.cc` line 435 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc2KJ_7X1T7WDthY2t:-NUc2KJ_7X1T7WDthY2u:b-nc427e) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L435)):*
> ```cpp
>   }
> 
>   // One of the pair is locked, expect a single contact result.
> ```

nit, "One of the pairs" or "One pair"
___
*[`multibody/plant/test/joint_locking_test.cc` line 475 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc2cOBAajp3CyzmdR5:-NUc2cOBAajp3CyzmdR6:b5cyqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L475)):*
> ```cpp
>   return std::vector<FilteredContactResultsConfig>{
>       {.contact_model = ContactModel::kPoint,
>        .solver = DiscreteContactSolver::kTamsi},
> ```

not sure why it is relevant to test all hydro/point combinations, but I haven't seen the implementation yet. 
I'd imagine that's because you follow different code paths that you want to test?

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 15 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @joemasterjohn)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc98ZK0KA1PWBlVq1w:-NUc98ZK0KA1PWBlVq1x:b-k440gv) (no related file):*
seeing you used the word "filter" in your code several times made me think.
Could we accomplish this in a cleaner way by using `SceneGraph` filters? Imagine if we could filter out geometry when we do `joint.Lock(&context)`.

___
___
*[`multibody/plant/compliant_contact_manager.cc` line 351 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc7ycq6Q2m_SZf7c75:-NUc7ycr4rB8FZM0C1Hn:b-6xi2sh) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/compliant_contact_manager.cc#L351)):*
> ```cpp
>   const geometry::SceneGraphInspector<T>& inspector = query_object.inspector();
> 
>   const std::vector<std::vector<int>>& per_tree_locked_indices =
> ```

should this be "unlocked"?

DamrongGuoy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 16 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @joemasterjohn)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUvjSTw4umpSo5nD_T9:-NUvjSTw4umpSo5nD_TA:bvtua3g) (no related file):*
Quick question: "_Geometry computed contact data for these types of pairs (`PenetrationAsPointPair` and `ContactSurface`) are not affected and still exist._" means which one?

A. MeshCat will continue to show the contact surfaces but no force/moment vectors will be drawn, Or 
B. MeshCat will not show the contact surfaces at all.

I'm guessing the latter because MeshCat draws from `ContactResults`, I think.

I have no opinion on how to do it (in multibody or geometry).  I only confirm that the expected user experience makes sense, which I think B is.

___

calderpg-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 16 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @joemasterjohn)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc98ZK0KA1PWBlVq1w:-NUvj74-B94YH6UCO9W1:b-w9o4lq) (no related file):*

Two thoughts on that:

1. This would eliminate geometry level information for these contacts completely (i.e. `PenetrationAsPointPair` and `ContactSurface`) for any other user of that plant's associated SceneGraph. If that same SceneGraph was used for motion planning for instance it would introduce complexity where that application would need to do additional filtering or "unlock" to get the side effect of un-filtering. Not sure if that's a common workflow though, maybe some input from @EricCousineau-TRI or @calderpg-tri might give some insight?

2. This would have to be achieved with SceneGraph's "transient" filters so they could be removed when and unlock happens. But the transient state is shared between all clients of a SceneGraph instance. Therefore another code path could unintentionally remove a transient declaration made by MbP or, more likely, could "flatten" the transient history, breaking the behavior when plant attempts to unlock the joint.
</blockquote></details>

To the extent that joint locking is supposed to be similar in behavior to having the affected bodies welded together, then applying a filter between those bodies' geometries would be the appropriate way to update SceneGraph (were those bodies simply welded together at the start, collisions between their geometries would be filtered).

I think trying to use the _same_ MbP/SG for multiple uses where some have locked joints and the others don't (and things are being applied SG-wide, not just to specific context(s)) is "don't do this" fringe case.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 16 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @joemasterjohn)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc98ZK0KA1PWBlVq1w:-NUvm-l_Cwt9BwWfnBCA:bje32jt) (no related file):*
> To the extent that joint locking is supposed to be similar in behavior to having the affected bodies welded together, then applying a filter between those bodies' geometries would be the appropriate way to update SceneGraph ...

I agree.

> ... (were those bodies simply welded together at the start, collisions between their geometries would be filtered).

That's only true modulo [MbP::get_adjacent_bodies_collision_filters()](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_multibody_plant.html#a14ad3e6c2aa8d6de728ee3552260f428).

Given that, it might make sense to do *both* the SAP-level fix (internal to the plant, to avoid degenerate dofs) and *also* to fix the contact pair reporting back to the user to omit the collisions (iff the adjacent body filters are enabled).

___

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUvjSTw4umpSo5nD_T9:-NUvwZXh7RCK1Yjbrr-1:b-icobt7) (no related file):*

Quick question: "_Geometry computed contact data for these types of pairs (`PenetrationAsPointPair` and `ContactSurface`) are not affected and still exist._" means which one?

A. MeshCat will continue to show the contact surfaces but no force/moment vectors will be drawn, Or 
B. MeshCat will not show the contact surfaces at all.

I'm guessing the latter because MeshCat draws from `ContactResults`, I think.

I have no opinion on how to do it (in multibody or geometry).  I only confirm that the expected user experience makes sense, which I think B is.
</blockquote></details>

B. is correct. The contact surfaces will still be computed by SceneGraph, but not shown by Meshcat. The contact surfaces are only sent to meshcat if there is a corresponding `HydroelasticContactInfo` in the `ContactResults`, but that is what this PR is filtering out.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 11 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @amcastro-tri and @joemasterjohn)

___
*[`multibody/plant/multibody_plant.cc` line 2899 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc6qTz3yZkYoAQle0c:-NUw1v7E-V5yCNrPH0et:bee7bit) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2899)):*

The locked/unlocked status of a mobilizer is stored as an abstract parameter in the context, so this is the most fine grained ticket to depend on, I believe.
</blockquote></details>

Don't tell reviewable, where it will be lost to the mists of time.  Tell the next person who reads this code and will likewise be confused, by adding a code comment nearby.

amcastro-tri (commented):
Thank you @joemasterjohn. :lgtm: 
+@rpoyner-tri for platform review please (tomorrow)

> Reviewed 5 of 8 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-:-NUxIOe_2oH1csiFxKlx:b-qhufkl)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc98ZK0KA1PWBlVq1w:-NUxFy6N0XVZzFLTqEZP:b-fvc4fn) (no related file):*

> To the extent that joint locking is supposed to be similar in behavior to having the affected bodies welded together, then applying a filter between those bodies' geometries would be the appropriate way to update SceneGraph ...

I agree.

> ... (were those bodies simply welded together at the start, collisions between their geometries would be filtered).

That's only true modulo [MbP::get_adjacent_bodies_collision_filters()](https://drake.mit.edu/doxygen_cxx/classdrake_1_1multibody_1_1_multibody_plant.html#a14ad3e6c2aa8d6de728ee3552260f428).

Given that, it might make sense to do *both* the SAP-level fix (internal to the plant, to avoid degenerate dofs) and *also* to fix the contact pair reporting back to the user to omit the collisions (iff the adjacent body filters are enabled).
</blockquote></details>

We decided to move  forward and make progress with the solution original PRed by Joe. That is, we'll work with the internal discrete solvers fix. The "public" (breaking) result is that users will not see contacts reported between bodies effectively anchored to the world. We could work on updating filters if needed on a later pass. 

___
___
*[`multibody/plant/multibody_plant.cc` line 2899 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc6qTz3yZkYoAQle0c:-NUxH6C10Wi6B1t67wtr:b-uwugwp) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2899)):*

Don't tell reviewable, where it will be lost to the mists of time.  Tell the next person who reads this code and will likewise be confused, by adding a code comment nearby.
</blockquote></details>

oh, I didn't realize it was a parameter, I always thought  about it as "state". I agree with Jeremey, please state in a comment if you don't mind. 
___
*[`multibody/plant/test/joint_locking_test.cc` line 431 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc22ySEArkx1HWTi2g:-NUxHXR_21Ul6XduovPz:bwcoxnf) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L431)):*

Done. We only get a single `HydroelasticContactInfo` per surface and this test is so constrained that I know there is exactly one contact surface. But I'll change it to be the more generic clause.
</blockquote></details>

oh, right, we use polygons!
___
*[`multibody/plant/test/joint_locking_test.cc` line 475 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc2cOBAajp3CyzmdR5:-NUxHnNDB_oVN2TkWiIC:b-ox5nom) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/test/joint_locking_test.cc#L475)):*

Testing both TAMSI and SAP is actually not needed becauase the code path ony differs based on point/hydro. I just have them both here explicitly as a smoke test while I was coding this. I can remove it if you find it distracting.
</blockquote></details>

I think i't ok. Maybe add comment stating this. Also I imagine your PR train on locking might leverage the additional testing?

joemasterjohn (commented):
Because this is a breaking change I went ahead and tested Anzu against this PR at r4: 

https://robotics-enabling-jenkins.awsinternal.tri.global/job/Integrations/job/Anzu-master-Drake-PR/203/

Nothing in Anzu seems to be breaking as a result.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-:-NV-oIwg2NJsbeG6HpVV:b-rsp3z9)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform) (waiting on @rpoyner-tri)

___
*[`multibody/plant/multibody_plant.cc` line 2899 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NUc6qTz3yZkYoAQle0c:-NV-vDBo5P80NoKMY_n3:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9fbeb68fa63a5d97fb12f44ffa54e8dddfce486c/multibody/plant/multibody_plant.cc#L2899)):*

oh, I didn't realize it was a parameter, I always thought  about it as "state". I agree with Jeremey, please state in a comment if you don't mind. 
</blockquote></details>

Done.

rpoyner-tri (commented):
:lgtm: nittery

> Reviewed 5 of 8 files at r2, 1 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-:-NV0xckV6-5gDcfl1to3:bb62doi)* status: 4 unresolved discussions

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0xQrC5s27eM8KSyCh):*
nit It would be a help to release notes editors to say exactly what makes this a breaking change and what kinds of programs would break.
___
*[`multibody/plant/compliant_contact_manager.cc` line 370 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0vQKW0fWupBVxUaOJ:-NV0vQKW0fWupBVxUaOK:b-w0na1w) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L370)):*
> ```cpp
> 
>     // For joint locking, filter out contacts between bodies who belong to trees
>     // with 0 degrees of freedom. At least one of the trees involved has to be
> ```

nit readability. These two sentences seems to be talking about opposites senses of the same thing.  At first read, one could be misled to think that *filtered-out* contacts need to satisfy the clauses of the second sentence. Maybe this(?):
```
   For joint locking, filter out contacts between bodies who belong to trees
   with 0 degrees of freedom. For a contact to remain in consideration, at
   least one of the trees involved has to be valid and have a non-zero
   number of DOFs.
```
___
*[`multibody/plant/compliant_contact_manager.cc` line 469 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0wCwtDS65BCzznaad:-NV0wCwtDS65BCzznaae:b-a98cy0) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L469)):*
> ```cpp
>     // For joint locking, filter out contacts between bodies who belong to trees
>     // with 0 degrees of freedom. At least one of the trees involved has to be
>     // valid and have a non-zero number of DOFs.
> ```

nit readability. see above.
___
*[`multibody/plant/compliant_contact_manager.cc` line 915 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0wWC30vYO3nA4znpn:-NV0wWC30vYO3nA4znpo:bgrw9ne) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L915)):*
> ```cpp
>     // For joint locking, filter out contacts between bodies who belong to trees
>     // with 0 degrees of freedom. At least one of the trees involved has to be
>     // valid and have a non-zero number of DOFs.
> ```

nit readability. see above.

Having seen the comment 3 times now, I wonder if the filtering predicate can be factored into a helper method (private? anonymous namespace maybe?) and commented in just one place.

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: 2 unresolved discussions

___
*[`multibody/plant/compliant_contact_manager.cc` line 370 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0vQKW0fWupBVxUaOJ:-NV1EDu0D0EX6tEntvrE:b4dqo45) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L370)):*

nit readability. These two sentences seems to be talking about opposites senses of the same thing.  At first read, one could be misled to think that *filtered-out* contacts need to satisfy the clauses of the second sentence. Maybe this(?):
```
   For joint locking, filter out contacts between bodies who belong to trees
   with 0 degrees of freedom. For a contact to remain in consideration, at
   least one of the trees involved has to be valid and have a non-zero
   number of DOFs.
```
</blockquote></details>

Done. Thanks for the snippet.
___
*[`multibody/plant/compliant_contact_manager.cc` line 469 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0wCwtDS65BCzznaad:-NV1EOMg00PbuIg5L6g1:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L469)):*

nit readability. see above.
</blockquote></details>

Done.
___
*[`multibody/plant/compliant_contact_manager.cc` line 915 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0wWC30vYO3nA4znpn:-NV1JxSjAFoJH7ICQAhZ:bsnj4o9) ([raw file](https://github.com/RobotLocomotion/drake/blob/829768c3958a9cc63982b999d5a41e5db63462c0/multibody/plant/compliant_contact_manager.cc#L915)):*

nit readability. see above.

Having seen the comment 3 times now, I wonder if the filtering predicate can be factored into a helper method (private? anonymous namespace maybe?) and commented in just one place.
</blockquote></details>

I'm going to punt because I think it's borderline for readability. The result would look something like:
```
if(is_tree_not_world_and_has_dofs(treeM_index, per_tree_unlocked_indices) || 
   is_tree_not_world_and_has_dofs(treeN_index, per_tree_unlocked_indices)){
  ...
}
```

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),amcastro-tri

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19349#-NV0xQrC5s27eM8KSyCh):*

nit It would be a help to release notes editors to say exactly what makes this a breaking change and what kinds of programs would break.
</blockquote></details>

Done.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19349)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),amcastro-tri

PR #19351, opened 2023-05-04
RenderEngineVtk supports Wavefront material libraries

SeanCurtis-TRI:
This PR is a reference PR. It provides the context in which smaller PRs all play together. It is a reference for reviewers of the actual PRs and will close upon the completion of the PR train.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19351)


SeanCurtis-TRI (commented):
@jwnimmer-tri @zachfang 

The CI failures on this draft have importance.

## tl;dr

The `render_gltf_client/test/integration_test.py` depends on how VTK loads meshes into memory. The gltf produced will change depending on how many vertices and the like actually get placed into VTK structures. As of this PR, the gltf being produced and the reference gltf on the file system differ.

## The details

There is good news and bad news. Here's the crux of the error message:

```
  File "/media/ephemeral0/ubuntu/workspace/linux-focal-clang-bazel-experimental-everything-release/_bazel_ubuntu/9b5a763240e43846505bc95ddbdb867f/sandbox/linux-sandbox/18406/execroot/drake/bazel-out/k8-opt/bin/geometry/render_gltf_client/py/integration_test.runfiles/drake/geometry/render_gltf_client/test/integration_test.py", line 298, in test_gltf_conversion
    self._check_one_gltf(gltf, ground_truth_gltf)
  File "/media/ephemeral0/ubuntu/workspace/linux-focal-clang-bazel-experimental-everything-release/_bazel_ubuntu/9b5a763240e43846505bc95ddbdb867f/sandbox/linux-sandbox/18406/execroot/drake/bazel-out/k8-opt/bin/geometry/render_gltf_client/py/integration_test.runfiles/drake/geometry/render_gltf_client/test/integration_test.py", line 221, in _check_one_gltf
    self.assertCountEqual(actual["nodes"], expected["nodes"])
AssertionError: Element counts were not equal:
First has 1, Second has 0:  {'matrix': [0.389418342309, -0.921060994003, -0.0, 0, 0.921060701964, 0.389418218836, -0.000796326711, 0, 0.000733465472, 0.000310104228, 0.999999682932, 0, 0.027, -0.0049, -0.092, 1], 'mesh': {'primitives': [{'attributes': {'POSITION': {'bufferView': {'byteLength': 101076, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5126, 'count': 8423, 'max': [0.03326, 0.009812, 0.188148], 'min': [-0.063938, -0.056809, -0.003153], 'type': 'VEC3'}, 'TEXCOORD_0': {'bufferView': {'byteLength': 67384, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5126, 'count': 8423, 'normalized': False, 'type': 'VEC2'}}, 'indices': {'bufferView': {'byteLength': 196608, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5125, 'count': 49152, 'type': 'SCALAR'}, 'material': {'pbrMetallicRoughness': {'baseColorFactor': [0.003921568627, 0, 0, 1], 'metallicFactor': 0, 'roughnessFactor': 1}}, 'mode': 4}]}}
First has 0, Second has 1:  {'matrix': [0.389418342309, -0.921060994003, -0.0, 0, 0.921060701964, 0.389418218836, -0.000796326711, 0, 0.000733465472, 0.000310104228, 0.999999682932, 0, 0.027, -0.0049, -0.092, 1], 'mesh': {'primitives': [{'attributes': {'POSITION': {'bufferView': {'byteLength': 589824, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5126, 'count': 49152, 'max': [0.033259999007, 0.009812000208, 0.188148006797], 'min': [-0.063937999308, -0.056809000671, -0.003153000027], 'type': 'VEC3'}, 'TEXCOORD_0': {'bufferView': {'byteLength': 393216, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5126, 'count': 49152, 'normalized': False, 'type': 'VEC2'}}, 'indices': {'bufferView': {'byteLength': 196608, 'byteOffset': 0}, 'byteOffset': 0, 'componentType': 5125, 'count': 49152, 'type': 'SCALAR'}, 'material': {'pbrMetallicRoughness': {'baseColorFactor': [0.003921568627, 0, 0, 1], 'metallicFactor': 0, 'roughnessFactor': 1}}, 'mode': 4}]}}
```

Observations:
1. The "First" (in the line beginning "First has 1") refers to the *generated* gltf file being tested. The json struct following represents that generated data.
2. Similar, the "Second" (in the line beginning "First has 0, Second has 1"), refers to the reference gltf file on disk. Again, the json struct represents the cached data (e.g., `test_color_scene.gltf`).
2. The differences:
    - The *size* of the various data buffers (POSITION and TEXCOORD_0) are *much* smaller in the tested gltf. It is likely that the extra work we're doing in creating the `RenderMesh` produces a far more compact representation of the mesh than `vtkOBJReader` did. (I can peek in the VTK code to confirm this if we like.)
    - The `min` and `max` extents of `POSITION` are printed with less apparent precision. The reason for this is less obvious. It may be that there is an underlying shift to `double`-valued quantities that would affect calculations and stringification.

## Next Steps

The simplest thing to do is to change the reference .gltf files in the test to reflect these changes. I could use Zach's help in doing so. I'm not entirely sure what its provenance of the gltf reference files and how best to modify it.

Prior to doing that, we may want to beef up our confidence on the correctness of these changes vis a vis gltf render client. I'm open to all suggestions on what we can/should do to proceed with confidence.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19351#-:-NUh-Hzh5odY3Xtzfp6Q:b-o91zde)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
Quick follow up on "guesses".

The old reference gltf file has:
 - 49152 vertices. The mustard bottle obj has exactly 16384 triangles. If you multiply by three, you get, yep, you guessed it, 49152. This is clear evidence that every triangle's vertices are being redundantly added to the buffers regardless of *need*.
- The new representation has 8423 vertices. The OBJ has 8194 vertices. The fact that our vertex count is greater than the OBJ count is a nice reality check. The fact it's only *slightly* greater tells us that the mesh doesn't have insane normals and texture coordinates (which is what we would expect).
- In both cases, the same number of triangles are reported. This can be inferred by the `indices` structure. They each have 3 * 16384 entries, which is how 16384 triangles are encoded.

So, there's some evidence to encourage good feelings.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19351#-:-NUh4i7c7_6f4HAI306N:b-h26vos)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
> The simplest thing to do is to change the reference .gltf files in the test to reflect these changes. I could use Zach's help in doing so. I'm not entirely sure what its provenance of the gltf reference files and how best to modify it.

It's just (on purpose and with good cause) a change-detector test.  To update it, basically re-commit your new gltf as the gold standard.  The original PR where it went in should have advice on how it was groomed for git.  If I recall correctly, the only grooming was manually deleting the in-band texture to keep the file small.

> Prior to doing that, we may want to beef up our confidence on the correctness of these changes vis a vis gltf render client. I'm open to all suggestions on what we can/should do to proceed with confidence.

If we can trust the VTK render engine (via it's own unit tests), then the automated regression test is already doing a round-trip render and comparison:

```py
    def test_integration(self):
        """Quantitatively compares the images rendered by RenderEngineVtk and
        RenderEngineGltfClient via a fully exercised RPC pipeline.
        """
```

If we're worried that's not sensitive enough, we could momentarily ratchet up the test tolerances to see exactly what (if anything) is different, and compare those test case images to the images that occur on master.

SeanCurtis-TRI:
I can buy those arguments.

FTR, it doesn't seem like there's much guidance for *updating* the change-detection test.

In `integration_test.py`, [there is an indication](https://github.com/RobotLocomotion/drake/blob/master/geometry/render_gltf_client/test/integration_test.py#L14-L18) that the reference gltf files have been modified and suggestion that you refer to the README file for more.

The README file [has a section on the integration test](https://github.com/RobotLocomotion/drake/blob/master/geometry/render_gltf_client/test/README.md#notes-on-integration-test), but doesn't mention updating the gltf files. What it has is guidance to *check* the gltf files (which it declares would be a "rare" occurrence).

In my own efforts to update the gltf files, I found I had to go down the `--sandbox_debug` route and subsequently found that there were *many* differences between the two files, most of which didn't matter. So, I hand tweaked the changes that seemed to be relevant. It seems the minimum acceptable update would be to grab the generated file from the sandbox and delete the "buffers" key-value pair. But that doesn't seem to be how the original files got created.

So, before I provide some concrete guidance for updating the gltf files in the readme, I wanted to see if there was something more ergonomic that I simply missed.

jwnimmer-tri:
> In my own efforts to update the gltf files, I found I had to go down the --sandbox_debug route ...

Maybe the test should save (or copy) the files to `TEST_UNDECLARED_OUTPUTS_DIR` so that they are easy to access later? For example:

https://github.com/RobotLocomotion/drake-blender/blob/bab0e41247981c6d21240b8c8d7189a1e9078cc4/test/server_test.py#L127-L133

SeanCurtis-TRI (commented):
I'll roll that into a follow up PR.

I'd still like @zachfang's thoughts on if there's further steps in processing the files (once we have easy access to them). There's a great deal of whitespace difference between the gltf file that gets generated by the test vs what has been stored. So, I'd like as complete a characterization of the process for creating new reference gltf files before adding that mechanism.

We *could* choose not to care about the whitepsace differences and assume Reviewable will help us detect every *meaningful* change.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19351#-:-NUx6hT_EwsSgVHpmw7v:bfsu0w5)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI:
This chain is now 3/4 in master and the last PR is live for review (#19383). I'm now simply going to close this as having no further value.

PR #19352, opened 2023-05-04
Add RenderMaterial to geometry/render

SeanCurtis-TRI:
Introduce the RenderMaterial type and utility functions:
 - Convert from GeometryProperties to RenderMaterial
 - Define fallback logic for when a mesh doesn't specify its own materials.
 - Provide a utility for mesh parsing to report when the mesh specifies its own materials as well as material properties.

Relates #18844

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352)

### DIFF
--- a/geometry/render/BUILD.bazel
+++ b/geometry/render/BUILD.bazel
@@ -16,6 +16,7 @@ drake_cc_package_library(
         ":render_camera",
         ":render_engine",
         ":render_label",
+        ":render_material",
     ],
 )
 
@@ -62,6 +63,21 @@ drake_cc_library(
     ],
 )
 
+drake_cc_library(
+    name = "render_material",
+    srcs = ["render_material.cc"],
+    hdrs = ["render_material.h"],
+    interface_deps = [
+        "//common:diagnostic_policy",
+        "//common:essential",
+        "//geometry:geometry_properties",
+        "//geometry:rgba",
+    ],
+    deps = [
+        "@tinyobjloader",
+    ],
+)
+
 # === test/ ===
 
 filegroup(
@@ -101,6 +117,16 @@ drake_cc_googletest(
         ":render_label",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
+    name = "render_material_test",
+    data = [":test_models"],
+    deps = [
+        ":render_material",
+        "//common/test_utilities:diagnostic_policy_test_base",
+        "//geometry:geometry_roles",
     ],
 )
 
--- a/geometry/render/render_material.cc
+++ b/geometry/render/render_material.cc
@@ -1,0 +1,112 @@
+#include "drake/geometry/render/render_material.h"
+
+#include <fstream>
+#include <vector>
+
+#include <fmt/format.h>
+
+#include "drake/common/drake_assert.h"
+#include "drake/common/fmt_eigen.h"
+
+namespace drake {
+namespace geometry {
+namespace internal {
+
+using drake::internal::DiagnosticPolicy;
+
+void MaybeWarnForRedundantMaterial(
+    const GeometryProperties& props, std::string_view mesh_name,
+    const DiagnosticPolicy& policy) {
+  std::vector<std::string> ignored_props;
+  if (props.HasProperty("phong", "diffuse")) {
+    ignored_props.push_back(fmt::format(
+        "('phong', 'diffuse') = {}",
+        fmt_eigen(props.GetProperty<Rgba>("phong", "diffuse").rgba())));
+  }
+  if (props.HasProperty("phong", "diffuse_map")) {
+    ignored_props.push_back(
+        fmt::format("('phong', 'diffuse_map') = {}",
+                    props.GetProperty<std::string>("phong", "diffuse_map")));
+  }
+  if (!ignored_props.empty()) {
+    policy.Warning(
+        fmt::format("The mesh {} has its own materials, but material "
+                    "properties have been defined as well. They will "
+                    "be ignored: {}",
+                    mesh_name, fmt::join(ignored_props, ", ")));
+  }
+}
+
+RenderMaterial MakeMeshFallbackMaterial(
+    const GeometryProperties& props, const std::filesystem::path& mesh_path,
+    const Rgba& default_diffuse,
+    const DiagnosticPolicy& policy) {
+  // If a material is indicated *at all* in the properties, that
+  // defines the material.
+  if (props.HasProperty("phong", "diffuse") ||
+      props.HasProperty("phong", "diffuse_map")) {
+    // This call must be guarded because it will create a valid material even if
+    // no properties have been defined.
+
+    // Why is the default color white?
+    //  We're here because the user specified *some* drake material property.
+    //  If it's *only* diffuse_map, then the underlying color must be white to
+    //  faithfully reproduce the texture. Obviously, if they've specified a
+    //  diffuse color, that will be used. Objects that request an invalid
+    //  texture will be drawn in white.
+    return DefineMaterial(props, Rgba(1, 1, 1), policy);
+  }
+
+  // Checks for foo.png for mesh filename foo.*. This the legacy behavior we
+  // want to do away with.
+  RenderMaterial material;
+
+  // This is the fall-through condition; the final priority in the protocol.
+  material.diffuse = default_diffuse;
+
+  if (!mesh_path.empty()) {
+    std::filesystem::path alt_texture_path(mesh_path);
+    alt_texture_path.replace_extension("png");
+    // If we find foo.png but can't access it, we'll *silently* treat it like
+    // we didn't find it. No value in warning for a behavior we're cutting.
+    if (std::ifstream(alt_texture_path).is_open()) {
+      material.diffuse = Rgba(1, 1, 1);
+      material.diffuse_map = alt_texture_path;
+    }
+  }
+  return material;
+}
+
+RenderMaterial DefineMaterial(
+    const GeometryProperties& props,
+    const Rgba& default_diffuse,
+    const DiagnosticPolicy& policy) {
+  RenderMaterial material;
+
+  material.diffuse_map =
+      props.GetPropertyOrDefault<std::string>("phong", "diffuse_map", "");
+  const bool has_texture = !material.diffuse_map.empty();
+  // Default of white with a declared texture, otherwise the given default.
+  material.diffuse = props.GetPropertyOrDefault<Rgba>(
+      "phong", "diffuse", has_texture ? Rgba(1, 1, 1) : default_diffuse);
+
+  if (has_texture) {
+    // Confirm it is available.
+    if (!std::ifstream(material.diffuse_map).is_open()) {
+      // TODO(SeanCurtis-TRI): It would be good to be able to tie this into
+      // some reference to the geometry under question. Ideally, the caller
+      // would provide a custom policy that would automatically decorate this
+      // message with the additional context.
+      policy.Warning(fmt::format(
+          "The ('phong', 'diffuse_map') property referenced a map that "
+          "could not be found: {}",
+          material.diffuse_map));
+      material.diffuse_map.clear();
+    }
+  }
+  return material;
+}
+
+}  // namespace internal
+}  // namespace geometry
+}  // namespace drake
--- a/geometry/render/render_material.h
+++ b/geometry/render/render_material.h
@@ -1,0 +1,73 @@
+#pragma once
+
+#include <filesystem>
+#include <string>
+
+#include "drake/common/diagnostic_policy.h"
+#include "drake/geometry/geometry_properties.h"
+#include "drake/geometry/rgba.h"
+
+namespace drake {
+namespace geometry {
+namespace internal {
+
+/* Specifies a mesh material as currently supported by Drake. We expect this
+ definition to grow with time. */
+struct RenderMaterial {
+  /* The diffuse appearance at a point on a geometry surface is defined by the
+   channel-wise product of the `diffuse` and the image color of `diffuse_map`
+   applied as a texture according to the geometry's texture coordinates. If
+   `diffuse_map` is empty, it acts as the multiplicative identity. */
+  Rgba diffuse;
+  std::filesystem::path diffuse_map;
+};
+
+/* Dispatches a warning to the given diagnostic policy if the props contain a
+ material definition. It is assumed an intrinsic material has already been found
+ for the named mesh. */
+void MaybeWarnForRedundantMaterial(
+    const GeometryProperties& props, std::string_view mesh_name,
+    const drake::internal::DiagnosticPolicy& policy);
+
+/* If a mesh definition doesn't include a single material (e.g., as in an .mtl
+ file for an .obj mesh), this function applies a cascading priority for
+ otherwise defining a material for the mesh.
+
+ The material is defined with the following protocol:
+
+   - If the properties indicate a material at all, the material is derived
+     purely from the properties (e.g., ("phong", "diffuse_map") and
+     ("phong", "diffuse").
+   - Otherwise, if an image can be located with a "compatible name" (e.g.,
+     foo.png for a mesh foo.obj), a material with an unmodulated texture is
+     created.
+   - Finally, a diffuse material is created with the given default_diffuse
+     color value.
+
+ References to textures will be included in the material iff they can be read.
+
+ @pre The mesh (named by `mesh_filename`) is a valid mesh and did not have an
+      acceptable material definition). */
+RenderMaterial MakeMeshFallbackMaterial(
+    const GeometryProperties& props, const std::filesystem::path& mesh_path,
+    const Rgba& default_diffuse,
+    const drake::internal::DiagnosticPolicy& policy);
+
+/* Creates a RenderMaterial from the given set of geometry properties. If no
+ material properties exist, a material with the given default diffuse color is
+ returned.
+
+ If only a texture is specified, the diffuse color will be white. If the
+ texture is not available, the diffuse map will be cleared, but the color will
+ remain white, signaling a data error in the material specification.
+
+ The `default_diffuse` color is only applied in the total absence of material
+ properties. */
+RenderMaterial DefineMaterial(
+    const GeometryProperties& props,
+    const Rgba& default_diffuse = Rgba(1, 1, 1),
+    const drake::internal::DiagnosticPolicy& policy = {});
+
+}  // namespace internal
+}  // namespace geometry
+}  // namespace drake
--- a/geometry/render/test/render_material_test.cc
+++ b/geometry/render/test/render_material_test.cc
@@ -1,0 +1,220 @@
+#include "drake/geometry/render/render_material.h"
+
+#include <filesystem>
+#include <fstream>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "drake/common/find_resource.h"
+#include "drake/common/test_utilities/diagnostic_policy_test_base.h"
+#include "drake/geometry/geometry_roles.h"
+
+namespace drake {
+namespace geometry {
+namespace internal {
+namespace {
+
+namespace fs = std::filesystem;
+
+/* Confirms the expected diagnostic warnings in the presence of material
+ properties. */
+class MaybeWarnForRedundantMaterialTest
+    : public test::DiagnosticPolicyTestBase {};
+
+/* Confirm the presence of a diffuse color dispatches a warning. */
+TEST_F(MaybeWarnForRedundantMaterialTest, WarningsDispatchedDiffuseColor) {
+  PerceptionProperties props;
+  props.AddProperty("phong", "diffuse", Rgba(0.1, 0.2, 0.3, 0.4));
+  MaybeWarnForRedundantMaterial(props, "dummy_file_name", diagnostic_policy_);
+  EXPECT_THAT(TakeWarning(), testing::ContainsRegex(
+                                 "has its own materials.*'phong', 'diffuse'"));
+}
+
+/* Confirm the presence of a diffuse map dispatches a warning. */
+TEST_F(MaybeWarnForRedundantMaterialTest, WarningsDispatchedDiffuseMap) {
+  PerceptionProperties props;
+  props.AddProperty("phong", "diffuse_map", "no_such.png");
+  MaybeWarnForRedundantMaterial(props, "dummy_file_name", diagnostic_policy_);
+  EXPECT_THAT(TakeWarning(),
+              testing::MatchesRegex(
+                  ".*has its own materials.*'phong', 'diffuse_map'.*"));
+}
+
+/* Tests the DefineMaterial() function (with the potential to dispatch warnings
+ to a diagnostic policy). */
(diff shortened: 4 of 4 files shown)


jwnimmer-tri (commented):
Checkpoint.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-:-NUc5RPfEAVMKsibCiPn:by96gdd)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbstri0be5533D0QzV:-NUbstrjEJJc2i3bgEPh:bp7qf1q) (no related file):*
BTW We can land the code here for now, but I'm not really sold on this being render-specific.  I would think our illustration meshes should follow the same material precedence protocols, wouldn't they?

___
___
*[`geometry/render/render_material.h` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbn6-j32RSWm0oBXOA:-NUbn6-k5pGQ5vlMjUOW:bod55te) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L17)):*
> ```objc
> struct RenderMaterial {
>   Rgba diffuse;
>   std::string diffuse_map;
> ```

nit It's not clear what this field denotes.

If it's a resolved filename, we should use `std::filesystem::path` as a data type instead, to signify that.

If it's a not-yet-validated path (i.e., midway through parsing), then `std::string` is fine, but then we need docs here explaining what it means... I assume in that case it's a relative path fragment or something, but then the question is relative to what?  If it's situation-dependent, we can say that and then at the points-of-use we need to clarify what it's relative to.

Or maybe its a URL?
___
*[`geometry/render/render_material.h` line 24 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbs5paDRHu11-skioJ:-NUbs5pb19Z8zIhEWwFJ:b-mp5p28) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L24)):*
> ```objc
>  for the named mesh. */
> void MaybeWarnForRedundantMaterial(
>     const GeometryProperties& props, const std::string& mesh_name,
> ```

BTW We don't really need a fat string.



_Suggestion:_
```objc
    const GeometryProperties& props, std::string_view mesh_name,
```
___
*[`geometry/render/render_material.h` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbuoyQ1-2agCfy2zT-:-NUbuoyQ1-2agCfy2zT0:b-ugrzer) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L47)):*
> ```objc
>       acceptable material definition). */
> RenderMaterial MakeMeshFallbackMaterial(
>     const GeometryProperties& props, const std::string& mesh_filename,
> ```

BTW I haven't finished writing up the GSG amendment yet, but I'm going to be pushing for using `path` (forbidding `string`) when referring to a path.  It makes things clearer for the reader, and generally simplifies the code inside and implementation.



_Suggestion:_
```objc
const std::filesystem::path& mesh_filename
```
___
*[`geometry/render/render_material.h` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbvgrHAWAOMRGozSIq:-NUbvgrHAWAOMRGozSIr:b-zv3v9) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L53)):*
> ```objc
> /* Creates a RenderMaterial from the given set of geometry properties. If no
>  material properties exist, a material with the given default diffuse color is
>  returned. */
> ```

nit This doc should promise to the caller that it checks whether the file exists and warns (and resets) if not.  I assume that the call site(s) will need to rely on that promise, and anyway it helps clarify why there's a policy object here.
___
*[`geometry/render/render_material.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbx1d_7nlR2t4RHMeV:-NUbx1d_7nlR2t4RHMeW:b-nvcjb4) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.cc#L42)):*
> ```cpp
>                     "be ignored: {}{}{}",
>                     mesh_name, diffuse_s,
>                     has_diffuse && has_diffuse_map ? ", " : "", map_s));
> ```

Manually dealing with commas and whitespace stuff isn't going to scale as we get more kinds of properties.  This can all be streamlined:

```c++
void MaybeWarnForRedundantMaterial(const GeometryProperties& props,
                                   const std::string& mesh_name,
                                   const DiagnosticPolicy& policy) {
  std::vector<std::string> ignored_props;
  if (props.HasProperty("phong", "diffuse")) {
    ignored_props.push_back(fmt::format(
        "('phong', 'diffuse') = {}",
        fmt_eigen(props.GetProperty<Rgba>("phong", "diffuse").rgba())));
  }
  if (props.HasProperty("phong", "diffuse_map")) {
    ignored_props.push_back(
        fmt::format("('phong', 'diffuse_map') = {}",
                    props.GetProperty<std::string>("phong", "diffuse_map")));
  }
  if (!ignored_props.empty()) {
    policy.Warning(fmt::format(
        "The mesh {} has its own materials, but material properties have been "
        "defined as well. They will be ignored: {}",
        mesh_name, fmt::join(ignored_props, ", ")));
  }
}
```
___
*[`geometry/render/render_material.cc` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbw4om8nMvGUR3W1a3:-NUbw4om8nMvGUR3W1a4:b-oy97h7) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.cc#L100)):*
> ```cpp
>     if (!std::ifstream(material.diffuse_map).is_open()) {
>       // TODO(SeanCurtis-TRI): It would be good to be able to tie this into
>       // some reference to the geometry under question.
> ```

BTW That would be best accomplished by the caller passing in a customized policy object that amend the callback message with that detail.  It wouldn't be a feature of *this* code at all.  (In other words, the TODO might be misplaced.)

jwnimmer-tri (commented):
:lgtm:

> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-:-NUcASr78wZskn7AeIUj:bnfp4nl)* status: 13 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`geometry/render/render_material.h` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc8ByG4stc4gNmlXbF:-NUc8ByG4stc4gNmlXbG:b-88rxwe) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L27)):*
> ```objc
>     const drake::internal::DiagnosticPolicy& policy);
> 
> /* If a mesh hasn't specified an implicit material (or it has defined multiple
> ```

nit It's probably worth defining this term for future readers who haven't been pounding on this for weeks like we have.  Or at least, giving an example like "The mtllib from OBJ file".



_Code quote:_
```objc
implicit material
```
___
*[`geometry/render/test/render_material_test.cc` line 28 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc7GAG5d6XRFqv_0cg:-NUc7GAG5d6XRFqv_0ch:bbp7od9) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L28)):*
> ```cpp
>  */
> TEST_F(MaybeWarnForRedundantMaterialTest, WarningsDispatched) {
>   // Pass in diffuse color property; the actual material doesn't change.
> ```

BTW This second clause seems spurious. This is a warning-generating function, it doesn't modify nor return anything, so no "change" can occur no matter what.



_Code quote:_
```cpp
the actual material doesn't change
```
___
*[`geometry/render/test/render_material_test.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc6CFHA_4lkMrG_dyn:-NUc6CFHA_4lkMrG_dyo:b-dkc8og) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L29)):*
> ```cpp
> TEST_F(MaybeWarnForRedundantMaterialTest, WarningsDispatched) {
>   // Pass in diffuse color property; the actual material doesn't change.
>   {
> ```

Maybe this is "strawberry vs vanilla" but there are *no* shared local variables across the two separate checks here.  Why not have two separate TEST_F cases each with a useful case name?  As it stands, there's this extra `{}` nesting that doesn't seem to have any point. In fact, by grouping into a single TEST_F if the policy base class trips at the end to flag a spurious warning message, we won't know _which_ call into the function under test caused the badness.

Ditto throughout.
___
*[`geometry/render/test/render_material_test.cc` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc6dfk7r_UsACxKu8-:-NUc6dfl-mtfClIX_Nkb:byw00cn) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L36)):*
> ```cpp
>     EXPECT_THAT(
>         TakeWarning(),
>         testing::MatchesRegex(".*has its own materials.*'phong', 'diffuse'.*"));
> ```

BTW As I understand it, canonical googletest would say:



_Suggestion:_
```cpp
testing::ContainsRegex("has its own materials.*'phong', 'diffuse'"));
```
___
*[`geometry/render/test/render_material_test.cc` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc8d_EES9nRGXdX6Lb:-NUc8d_EES9nRGXdX6Lc:b-8c2xw8) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L83)):*
> ```cpp
> 
> /* When only the (phong, diffuse_map) is defined, it is used with the default
>  diffuse color. */
> ```

BTW I'm not convinced yet that this is the correct protocol.  I think possibly "default diffuse to mate with a map" (white) and "default diffuse with no map" (grey) might end up being two different defaults.

But anyway this PR is self-consistent and well-tested, so we can revisit in the call-site PRs down the road whether or not we need that extra feature, and add it in then.
___
*[`geometry/render/test/render_material_test.cc` line 148 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc9Vxs-1Oa6_zWl9o2:-NUc9Vxs-1Oa6_zWl9o3:btkpb37) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L148)):*
> ```cpp
> /* No material defined in the properties and no foo.png --> default-colored
>  material. This doesn't test the case where foo.png *does* exist, but isn't
>  available. TODO(SeanCurtis-TRI): Test unaccessible foo.png. */
> ```

I don't think we should plan to do this.  The code is destined for removal.

SeanCurtis-TRI (commented):
Comments addressed. BTW You persuaded me as to the separation of the locale's default diffuse and the `foo.png` logic. I've made a change there.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-:-NUcPCuOAVcUy1g103-a:b-y54lv5)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbstri0be5533D0QzV:-NUcDJxbEhkSQm9b2_2a:b-mz7q7o) (no related file):*

BTW We can land the code here for now, but I'm not really sold on this being render-specific.  I would think our illustration meshes should follow the same material precedence protocols, wouldn't they?
</blockquote></details>

I'd considered that. The biggest challenge is mapping geometry properties to what representations the visualizers consume. So, the names reflect what I felt confident in supporting *today*.  Fortunately, it's all internal, so we can shift later. 

___
___
*[`geometry/render/render_material.h` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbuoyQ1-2agCfy2zT-:-NUcFjVq4f7GpH0oICKM:b-pbdqz3) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.h#L47)):*

BTW I haven't finished writing up the GSG amendment yet, but I'm going to be pushing for using `path` (forbidding `string`) when referring to a path.  It makes things clearer for the reader, and generally simplifies the code inside and implementation.
</blockquote></details>

Agreed.
___
*[`geometry/render/render_material.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbx1d_7nlR2t4RHMeV:-NUcGxXL7B8s3L2t_mkX:bxwmc4t) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.cc#L42)):*

Manually dealing with commas and whitespace stuff isn't going to scale as we get more kinds of properties.  This can all be streamlined:

```c++
void MaybeWarnForRedundantMaterial(const GeometryProperties& props,
                                   const std::string& mesh_name,
                                   const DiagnosticPolicy& policy) {
  std::vector<std::string> ignored_props;
  if (props.HasProperty("phong", "diffuse")) {
    ignored_props.push_back(fmt::format(
        "('phong', 'diffuse') = {}",
        fmt_eigen(props.GetProperty<Rgba>("phong", "diffuse").rgba())));
  }
  if (props.HasProperty("phong", "diffuse_map")) {
    ignored_props.push_back(
        fmt::format("('phong', 'diffuse_map') = {}",
                    props.GetProperty<std::string>("phong", "diffuse_map")));
  }
  if (!ignored_props.empty()) {
    policy.Warning(fmt::format(
        "The mesh {} has its own materials, but material properties have been "
        "defined as well. They will be ignored: {}",
        mesh_name, fmt::join(ignored_props, ", ")));
  }
}
```

</blockquote></details>

Excellent point.
___
*[`geometry/render/test/render_material_test.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc6CFHA_4lkMrG_dyn:-NUcNYYF5_P8WOhDtLXA:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L29)):*

Maybe this is "strawberry vs vanilla" but there are *no* shared local variables across the two separate checks here.  Why not have two separate TEST_F cases each with a useful case name?  As it stands, there's this extra `{}` nesting that doesn't seem to have any point. In fact, by grouping into a single TEST_F if the policy base class trips at the end to flag a spurious warning message, we won't know _which_ call into the function under test caused the badness.

Ditto throughout.
</blockquote></details>

Done.
___
*[`geometry/render/test/render_material_test.cc` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc6dfk7r_UsACxKu8-:-NUcLVJuC13_qwqEJsfU:bb0kw4k) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L36)):*

BTW As I understand it, canonical googletest would say:
</blockquote></details>

Much nicer...I stuck with the first promising symbol my eyes latched onto.
___
*[`geometry/render/test/render_material_test.cc` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc8d_EES9nRGXdX6Lb:-NUcHkao5Pwgp3t5DShg:b-mdrshx) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L83)):*

BTW I'm not convinced yet that this is the correct protocol.  I think possibly "default diffuse to mate with a map" (white) and "default diffuse with no map" (grey) might end up being two different defaults.

But anyway this PR is self-consistent and well-tested, so we can revisit in the call-site PRs down the road whether or not we need that extra feature, and add it in then.
</blockquote></details>

I find the point compelling. The `RenderEngine` should only apply its default diffuse in the absence of all other material definition. Specifying only a diffuse texture seems a perfectly reasonable way to *fully* specify the color of the object.

In the follow up PR, I'm going to apply a patch to tinyobj so that we get the same behavior out of that (texture on white). The goal being defining a `newmtl` with only `map_Kd foo.png` should have the same result as only defining `('phong', 'diffuse_map') = "foo.png"`.
___
*[`geometry/render/test/render_material_test.cc` line 148 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUc9Vxs-1Oa6_zWl9o2:-NUcHeXg5BYacKJ-mTaL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/test/render_material_test.cc#L148)):*

I don't think we should plan to do this.  The code is destined for removal.
</blockquote></details>

Done.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352)* status: LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI)

___
*[`geometry/render/test/render_material_test.cc` line 173 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUcSVnV5biqcmB2f0dw:-NUcW9Lk5qQ8ekPeahuH:b7q6h07) ([raw file](https://github.com/RobotLocomotion/drake/blob/7c6855f68a3f5994880e34beb4faeae8b41b1064/geometry/render/test/render_material_test.cc#L173)):*

nit Trailing whitespace.  Ditto a few lines later.
</blockquote></details>

So weird that it didn't get flagged by CI.  :-/

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352)* status: LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI)

___
*[`geometry/render/test/render_material_test.cc` line 173 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUcSVnV5biqcmB2f0dw:-NUcWSJD1804cnCsnquT:b-tz8y1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/7c6855f68a3f5994880e34beb4faeae8b41b1064/geometry/render/test/render_material_test.cc#L173)):*

So weird that it didn't get flagged by CI.  :-/
</blockquote></details>

Unfortunately, cpplint can't always handle C-style `/* */` comments very well.

Fortunately, the `clang-format` linter will flag this once someone enables it for this directory.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352)* status: LGTM missing from assignee EricCousineau-TRI(platform) (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
:lgtm: platform with small grammar nits!

> Reviewed 1 of 4 files at r1, 1 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-:-NUceuBZETw0OGE5QogR:b270dmh)* status: 2 unresolved discussions

___
*[`geometry/render/test/render_material_test.cc` line 120 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUcefzh8mnB9ckWdJ2z:-NUcefzh8mnB9ckWdJ3-:b1ry5ws) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4cd852ebde08992a4fb166b88ec6f2109e5a523/geometry/render/test/render_material_test.cc#L120)):*
> ```cpp
> }
> 
> /* Tests the MakeMeshFallbackMaterial() function. This function has should only
> ```

nit grammar "has should"
___
*[`geometry/render/test/render_material_test.cc` line 169 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUcep7kD9WKwZ1XtnSI:-NUcep7kD9WKwZ1XtnSJ:b-9ikgbw) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4cd852ebde08992a4fb166b88ec6f2109e5a523/geometry/render/test/render_material_test.cc#L169)):*
> ```cpp
> }
> 
> /* The presence of either any material property should create a material.
> ```

nit grammar "either any"

SeanCurtis-TRI (commented):
Them there's some good new grammar usements.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-:-NUcgkkK6fQmLWE1-Ln4:b-t4nye7)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),EricCousineau-TRI(platform)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19352)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),EricCousineau-TRI(platform)

jwnimmer-tri:
*[`geometry/render/render_material.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbx1d_7nlR2t4RHMeV:-NUd1iZF817Gq098kBt-:b-mnkhw7) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.cc#L42)):*

Excellent point.
</blockquote></details>

I just noticed that you didn't take my line-wrapping improvement to the fmt::format message string.  It can fit on two lines, instead of three.  Any reason why not?

SeanCurtis-TRI:
*[`geometry/render/render_material.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19352#-NUbx1d_7nlR2t4RHMeV:-NUdQ3cJ8mtJx_dYBlY-:bc8ow1u) ([raw file](https://github.com/RobotLocomotion/drake/blob/56a53b8debfa8c5820252e0234bec5c163563572/geometry/render/render_material.cc#L42)):*

I just noticed that you didn't take my line-wrapping improvement to the fmt::format message string.  It can fit on two lines, instead of three.  Any reason why not?
</blockquote></details>

I hadn't noticed the formatting change. Stupidly, I should've just copied and pasted the snippet you provided. Instead, I did my own typing and then hit it with clang-format. Since I hadn't changed the literal string boundaries, clang-format was powerless to do anything. So, the consequence of being foolish.

Issue #18327, opened 2022-11-15
Add support for macOS Ventura


svenevs:
- [x] Add PR that updates `from_source.md` and `installation.md` once we have a successful build going.
- [X] Identify the locations drake warns about unsupported OS / compiler (only a few).
    - [X] Add `"13"` to list (#19350)
    
        https://github.com/RobotLocomotion/drake/blob/f2a3f717f49dacac0c066892fa2db2342a5709d4/tools/workspace/os.bzl#L168
- [x] Fix a few build error(s) with new Xcode ([reference job log](https://drake-jenkins.csail.mit.edu/job/mac-arm-ventura-clang-bazel-experimental-everything-release/3/consoleFull).
    - [x] ERROR: -wunusedbutset

        ```console
        [4:16:03 AM]  ERROR: /Users/admin/workspace/mac-arm-ventura-clang-bazel-experimental-everything-release/src/multibody/inverse_kinematics/BUILD.bazel:27:17: Compiling multibody/inverse_kinematics/constraint_relaxing_ik.cc failed: (Exit 1): wrapped_clang_pp failed: error executing command (from target //multibody/inverse_kinematics:constraint_relaxing_ik) external/local_config_cc/wrapped_clang_pp '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -g0 -O2 -DNDEBUG ... (remaining 401 arguments skipped)
        [4:16:03 AM]  
        [4:16:03 AM]  Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
        [4:16:03 AM]  multibody/inverse_kinematics/constraint_relaxing_ik.cc:55:7: error: variable 'step_ctr' set but not used [-Werror,-Wunused-but-set-variable]
        [4:16:03 AM]    int step_ctr = 0;
        ```

    - [x]  geometry/proximity/mesh_plane_intersection.cc:87:7 `int num_intersections = 0;`
    - [x] multibody/plant/multibody_plant.cc:2559:7 `int num_actuated_instances = 0;`
[...]

PR #19354, opened 2023-05-04
Avoid 'using std::move'

jwnimmer-tri:
This risks severe ADL hazards, and will be Clang error soon.

Towards #18327.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19354)


jwnimmer-tri:
+@EricCousineau-TRI for both reviews per schedule, please.

PR #19356, opened 2023-05-04
Avoid ADL std::move

jwnimmer-tri:
Amends #19354.  In that PR, I only searched for the `using std::move;` statements.  Turns out, for things like move on a `std::unique_ptr`, we were getting away with ADL sometimes, too -- even in headers!

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19356)

### DIFF
--- a/common/symbolic/generic_polynomial.cc
+++ b/common/symbolic/generic_polynomial.cc
@@ -353,7 +353,7 @@ Variables GetDecisionVariables(
 
 template <typename BasisElement>
 GenericPolynomial<BasisElement>::GenericPolynomial(MapType init)
-    : basis_element_to_coefficient_map_{move(init)},
+    : basis_element_to_coefficient_map_{std::move(init)},
       indeterminates_{
           GetIndeterminates<BasisElement>(basis_element_to_coefficient_map_)},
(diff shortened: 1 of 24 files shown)


jwnimmer-tri:
+@EricCousineau-TRI for both reviews again, please!

EricCousineau-TRI (commented):
:lgtm: x2

BTW where is `move` even coming from in these cases?!

> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19356#-:-NUcDJ81Aj7mUMJ7yzJL:b-itry9a)* status: missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri:
> BTW where is move even coming from in these cases?!

It's *actually* using ADL for realz.  The type being moved is like `std::vector` or `std::unique_ptr` which means that `move(myvector)` searches for a function named `move` in the same namespace as `myvector` which is `std::`.

EricCousineau-TRI:
ahhhhhh gotcha, thanks!

PR #19364, opened 2023-05-05
Add RenderMaterial to RenderMesh

SeanCurtis-TRI:
- `RenderMesh` can now come back with material data in addition to the geometric data.
- `LoadRenderMeshFromObj()` signature has changed to provide sufficient data for material definition.
  - `RenderEngineGl` call has changed to patch around the change. `RenderEngineGl` still doesn't take advantage of this information. That will require further modifications.
- Implementation in `LoadRenderMeshFromObj()` has been updated to use the object-oriented version of `tinyobj`.
  - Some error messages have been made more agnostic -- no longer tied to a specific render engine implementation.
- Parsing is now wholly reliant on `tinyobj`; the unused/incompatible streaming version of the parser has simply been removed.
- Patch `tinyobj` to conform to Drake behavior: if `map_Kd` is supplied but `Kd` is not, `Kd` defaults to (1, 1, 1) (in `tinyobj` it was (0.6, 0.6, 0.6)).

This is the last prerequisite to update `RenderEngineVtk` to respect .mtl files and the new material protocol.

Relates https://github.com/RobotLocomotion/drake/issues/18844

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364)

### DIFF
--- a/geometry/render/BUILD.bazel
+++ b/geometry/render/BUILD.bazel
@@ -84,7 +84,11 @@ drake_cc_library(
     srcs = ["render_mesh.cc"],
     hdrs = ["render_mesh.h"],
     interface_deps = [
+        ":render_material",
+        "//common:diagnostic_policy",
         "//common:essential",
+        "//geometry:geometry_properties",
+        "//geometry:rgba",
     ],
     deps = [
         "@tinyobjloader",
@@ -149,8 +153,11 @@ drake_cc_googletest(
     deps = [
         ":render_mesh",
         "//common:find_resource",
+        "//common:temp_directory",
+        "//common/test_utilities:diagnostic_policy_test_base",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
+        "//geometry:geometry_roles",
     ],
 )
 
--- a/geometry/render/render_material.cc
+++ b/geometry/render/render_material.cc
@@ -1,6 +1,7 @@
 #include "drake/geometry/render/render_material.h"
 
 #include <fstream>
+#include <string>
 #include <vector>
 
 #include <fmt/format.h>
--- a/geometry/render/render_material.h
+++ b/geometry/render/render_material.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <filesystem>
-#include <string>
 
 #include "drake/common/diagnostic_policy.h"
 #include "drake/geometry/geometry_properties.h"
--- a/geometry/render/render_mesh.cc
+++ b/geometry/render/render_mesh.cc
@@ -1,8 +1,10 @@
 #include "drake/geometry/render/render_mesh.h"
 
+#include <algorithm>
 #include <fstream>
-#include <istream>
 #include <map>
+#include <set>
+#include <string>
 #include <tuple>
 #include <vector>
 
@@ -16,47 +18,118 @@
 namespace drake {
 namespace geometry {
 namespace internal {
+namespace {
 
+using drake::internal::DiagnosticPolicy;
 using Eigen::Vector2d;
 using Eigen::Vector3d;
 using std::make_tuple;
 using std::map;
-using std::string;
 using std::tuple;
 using std::vector;
 
-RenderMesh LoadRenderMeshFromObj(std::istream* input_stream,
-                                 const std::string& filename) {
-  tinyobj::attrib_t attrib;
-  vector<tinyobj::shape_t> shapes;
-  vector<tinyobj::material_t> materials;
-  string warn;
-  string err;
-  /* This renderer assumes everything is triangles -- we rely on tinyobj to
-   triangulate for us. */
-  const bool do_tinyobj_triangulation = true;
+/* Constructs a RenderMaterial from a single tinyobj material specification. */
+RenderMaterial MakeMaterialFromMtl(const tinyobj::material_t& mat,
+                                   const std::filesystem::path& obj_path,
+                                   const GeometryProperties& properties,
+                                   bool has_tex_coord,
+                                   const DiagnosticPolicy& policy) {
+  RenderMaterial result;
+  result.diffuse.set(mat.diffuse[0], mat.diffuse[1], mat.diffuse[2],
+                     mat.dissolve);
+  result.diffuse_map = [&mat, &obj_path]() -> std::string {
+    if (mat.diffuse_texname.empty()) {
+      return mat.diffuse_texname;
+    }
+    std::filesystem::path tex_path(mat.diffuse_texname);
+    if (tex_path.is_absolute()) {
+      return mat.diffuse_texname;
+    }
+    // There are three potential paths: path to obj, path to mtl, and path to
+    // texture image (png). We have the path to obj. The mtl is defined relative
+    // to the obj (inside the obj) and the png is defined relative to the
+    // mtl (defined in the mtl). However, tinyobj doesn't give us access to
+    // the mtl path to resolve image paths. For now, we're making the
+    // simplifying assumption that obj and mtl are in the same directory.
+    //
+    // What if the OBJ references multiple mtl files in disparate locations?
+    // Ideally, `material_t` should  come with a string indicating either the
+    // the mtl file it came from or the directory of that mtl file. Then
+    // relative paths of the referenced images can be properly interpreted.
+    // Or what is stored in material_t should be relative to the obj.
+    std::filesystem::path obj_dir = obj_path.parent_path();
+    return (obj_dir / tex_path).lexically_normal().string();
+  }();
 
-  drake::log()->trace("LoadRenderMeshFromObj('{}')", filename);
-
-  /* Tinyobj doesn't infer the search directory from the directory containing
-   the obj file. We have to provide that directory; of course, this assumes
-   that the material library reference is relative to the obj directory.
-   Ignore material-library file.  */
-  tinyobj::MaterialReader* material_reader = nullptr;
-  const bool ret =
-      tinyobj::LoadObj(&attrib, &shapes, &materials, &warn, &err, input_stream,
-                       material_reader, do_tinyobj_triangulation);
-  if (!ret) {
-    throw std::runtime_error(
-        fmt::format("tinyobj::LoadObj failed to load file: {}", filename));
+  // If texture is specified, it must be available and applicable.
+  if (!result.diffuse_map.empty()) {
+    std::ifstream tex_file(result.diffuse_map);
+    if (!tex_file.is_open()) {
+      policy.Warning(fmt::format(
+          "The OBJ file's material requested an unavailable diffuse texture "
+          "image: {}. The image will be omitted.",
+          result.diffuse_map.string()));
+      result.diffuse_map.clear();
+    }
+    if (!has_tex_coord) {
+      policy.Warning(fmt::format(
+          "The OBJ file's material requested a diffuse texture image: {}. "
+          "However the mesh doesn't define texture coordinates. The image "
+          "will be omitted.",
+          result.diffuse_map.string()));
+      result.diffuse_map.clear();
+    }
   }
 
-  if (shapes.empty()) {
+  MaybeWarnForRedundantMaterial(properties, obj_path.string(), policy);
+  return result;
+}
+
+}  // namespace
+
+// TODO(SeanCurtis-TRI): Add troubleshooting entry on OBJ support and
+// reference it in these errors/warnings.
+
+RenderMesh LoadRenderMeshFromObj(const std::filesystem::path& obj_path,
+                                 const GeometryProperties& properties,
+                                 const Rgba& default_diffuse,
+                                 const DiagnosticPolicy& policy) {
+  tinyobj::ObjReaderConfig config;
+  config.triangulate = true;
+  config.vertex_color = false;
+  tinyobj::ObjReader reader;
+  const bool valid_parse = reader.ParseFromFile(obj_path.string(), config);
+
+  if (!valid_parse) {
+    throw std::runtime_error(fmt::format("Failed parsing the obj file: {}: {}",
+                                         obj_path.string(), reader.Error()));
+  }
+
+  // We better not get any errors if we have a valid parse.
+  DRAKE_DEMAND(reader.Error().empty());
+
+  const vector<tinyobj::shape_t>& shapes = reader.GetShapes();
+
+  bool faces_found = false;
+  for (const auto& shape : shapes) {
+    faces_found = shape.mesh.indices.size() > 0;
+    if (faces_found) break;
+  }
+  if (!faces_found) {
     throw std::runtime_error(fmt::format(
         "The OBJ data appears to have no faces; it could be missing faces or "
         "might not be an OBJ file: {}",
-        filename));
+        obj_path.string()));
   }
+
+  if (!reader.Warning().empty()) {
+    policy.Warning(reader.Warning());
+  }
+
+  const tinyobj::attrib_t& attrib = reader.GetAttrib();
+
+  // All of the materials referenced by faces.
+  std::set<int> referenced_materials;
 
   /* The parsed product needs to be further processed. The RenderMesh assumes
    that all vertex quantities (positions, normals, texture coordinates) are
@@ -104,9 +177,8 @@ RenderMesh LoadRenderMeshFromObj(std::istream* input_stream,
   //   1. If normals are absent, generate normals so that we get faceted meshes.
   //   2. Make use of smoothing groups.
   if (attrib.normals.size() == 0) {
-    throw std::runtime_error(fmt::format(
-        "OBJ has no normals; RenderEngineGl requires OBJs with normals: {}",
-        filename));
+    throw std::runtime_error(
+        fmt::format("OBJ has no normals: {}", obj_path.string()));
   }
 
   bool has_tex_coord{attrib.texcoords.size() > 0};
@@ -131,13 +203,14 @@ RenderMesh LoadRenderMeshFromObj(std::istream* input_stream,
         const int norm_index = shape_mesh.indices[v_index].normal_index;
         const int uv_index = shape_mesh.indices[v_index].texcoord_index;
         if (norm_index < 0) {
-          throw std::runtime_error(
-              fmt::format("Not all faces reference normals: {}", filename));
+          throw std::runtime_error(fmt::format(
+              "Not all faces reference normals: {}", obj_path.string()));
         }
         if (has_tex_coord) {
           if (uv_index < 0) {
-            throw std::runtime_error(fmt::format(
-                "Not all faces reference texture coordinates: {}", filename));
+            throw std::runtime_error(
+                fmt::format("Not all faces reference texture coordinates: {}",
+                            obj_path.string()));
           }
         } else {
           DRAKE_DEMAND(uv_index < 0);
@@ -167,6 +240,7 @@ RenderMesh LoadRenderMeshFromObj(std::istream* input_stream,
         ++v_index;
       }
       triangles.emplace_back(&face_vertices[0]);
+      referenced_materials.insert(shape_mesh.material_ids[f]);
     }
   }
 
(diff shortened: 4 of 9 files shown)


SeanCurtis-TRI (commented):
+(release notes: none) 

+@jwnimmer-tri for feature review (as we have some momentum).

And, as before, the full context can be found in the draft PR  #19351.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUgjiQW0krnowXGmJbq:b-m4i2rk)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
+(status: do not review) I realized I hadn't quite followed through with my clean up. Stay tuned.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUgmDEy3kDdKnssM2UU:b-aeeh9d)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
-(status: do not review) Cleaned up.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUh6qkA4Zh71EYojY-C:b-2i8i5a)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
Checkpoint on a few nits, as I take a break for the day.

> Reviewed 8 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUhJz6ZDDz9TrL0QPVw:boey3z)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`geometry/render/render_mesh.h` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhEwYD6xYxQAkHiH5w:-NUhEwYD6xYxQAkHiH5x:bx2e5f7) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L42)):*
> ```objc
>   bool has_tex_coord{kHasTexCoordDefault};
> 
>   /* The specification of the material associated with this mesh data. */
> ```

BTW Is there any invariant relating `has_tex_coord` to the contents of `material`?  If so, we should mention it here.  (I assume there's not any invariant.)
___
*[`geometry/render/render_mesh.h` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhGcRlE7O0i5c_nQXD:-NUhGcRlE7O0i5c_nQXE:b-xx8pmt) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L69)):*
> ```objc
> 
>  The `properties` and `default_diffuse` are only applied in the case where the
>  material fallback rules must be applied.
> ```

BTW For me at least, moving this sentence up near where `MakeMeshFallbackMaterial` is cited would have been easier to read.  Here, it took me a while to page back in what "in the case where the material fallback rules" was talking about.  Or instead of using it, use the exact same wording in every place this condition is cited.
___
*[`geometry/render/render_mesh.h` line 89 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhFIqA1KrI6Sb47RQ6:-NUhFIqA1KrI6Sb47RQ7:b6k7ls0) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L89)):*
> ```objc
>                            or normals, c) faces fail to reference normals, or d)
>                            faces fail to reference the texture coordinates if
>                            they are present.  */
> ```

It seems odd to throw on bad data when we have a juicy `policy` object sitting there ready to post errors.  Should we switch to calling that instead, whether in this PR or a future one?  When defaulted it would be as good as currently, and when not defaulted could be hooked by the caller to provide even more context about how we ended up here.
___
*[`geometry/render/render_mesh.cc` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhH_AwC5Qag_UdRQj0:-NUhH_AwC5Qag_UdRQj1:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L51)):*
> ```cpp
>     // doesn't give us this information. It'd be nice if it did. What if the
>     // OBJ references multiple mtl files? Ideally, `material_t` should come
>     // come with a string indicating either the mtl file it came from or
> ```

typo



_Suggestion:_
```cpp
with
```
___
*[`geometry/render/render_mesh.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhHmV2CcWCRWuwW01r:-NUhHmV2CcWCRWuwW01s:buf6vzt) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L53)):*
> ```cpp
>     // come with a string indicating either the mtl file it came from or
>     // the directory of that mtl file. Then relative paths of the referenced
>     // images can be properly interpreted.
> ```

The prior paragraph is talking about obj vs mtl paths.  But I think really what it's trying to get at is mtl vs png paths.  Can you double-check the wording?
___
*[`geometry/render/render_mesh.cc` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhIHZ452iQeB-GSNXB:-NUhIHZ452iQeB-GSNXC:bd0jr5r) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L59)):*
> ```cpp
> 
>   // If texture is specified, it must be available and applicable.
>   bool valid_texture = true;
> ```

Working

When filename is empty, why is this `true`?
___
*[`geometry/render/render_mesh.cc` line 87 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhIOfB7jXwsTFXxAg7:-NUhIOfB7jXwsTFXxAg8:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L87)):*
> ```cpp
> }  // namespace
> 
> // TODO(SeanCurtis-TRI): Add trouble shooting entry on OBJ support and
> ```

typo



_Suggestion:_
```cpp
troubleshooting
```
___
*[`geometry/render/render_mesh.cc` line 108 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhJ1om-Tn00aqv0ZSJ:-NUhJ1om-Tn00aqv0ZSK:buvruup) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L108)):*
> ```cpp
>   DRAKE_DEMAND(reader.Error().empty());
> 
>   const auto& shapes = reader.GetShapes();
> ```

nit Too-obscure `auto`



_Suggestion:_
```cpp
const vector<tinyobj::shape_t>& shapes = reader.GetShapes();
```
___
*[`geometry/render/render_mesh.cc` line 126 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhJAkI8LtOpCkOhbuy:-NUhJAkI8LtOpCkOhbuz:b-v222jl) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L126)):*
> ```cpp
>   }
> 
>   const auto& attrib = reader.GetAttrib();
> ```

nit Too-obscure auto.

jwnimmer-tri (commented):
I've finished reading, and :lgtm:.

(I still need to circle back to one more draft comment, working below.)

> Reviewed 1 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUvpo8x2a1-rDZj-Gaz:b-cxb1bd)* status: 17 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`geometry/render/test/render_mesh_test.cc` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvgD-9-Y7L1cq7vFqT:-NUvgD-9-Y7L1cq7vFqU:b-qzymow) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L56)):*
> ```cpp
>   }
> 
>   static void SetUpTestSuite() {
> ```

nit Not everyone might realize that this is a Google API.  (I had to go look it up.) A comment should help clarify:

```c++
/* Performs shared setup for all tests in the test suite. GoogleTest calls
 SetUpTestSuite() before running the first test in the test suite. */
```
___
*[`geometry/render/test/render_mesh_test.cc` line 113 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvfr9-EgcFV-Ad8GWF:-NUvfr9-EgcFV-Ad8GWG:b-f3j31f) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L113)):*
> ```cpp
> };
> 
> const fs::path LoadRenderMeshFromObjTest::kTempDir = temp_directory();
> ```

nit This path almost certainly has a non-trivial destructor, so needs to be function-local static and `never_destroyed`.
___
*[`geometry/render/test/render_mesh_test.cc` line 119 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvh3Kt7vegVxdT5P4V:-NUvh3Kt7vegVxdT5P4W:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L119)):*
> ```cpp
>   int error = -1;
>   {
>     // Case: file can't be parsed. THe simplest way to achieve a invalid parse
> ```

typo



_Suggestion:_
```cpp
The
```
___
*[`geometry/render/test/render_mesh_test.cc` line 132 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvhoZt07pjYhVChPQc:-NUvhoZt07pjYhVChPQd:bz23n0z) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L132)):*
> ```cpp
>         WriteFile("v 1 2 3", fmt::format("error{}.obj", ++error));
>     DRAKE_EXPECT_THROWS_MESSAGE(
>         LoadRenderMeshFromObj(obj_path.string(), empty_props(), kDefaultDiffuse,
> ```

nit I think maybe the `.string()` tacked onto the end here is vestigial?

Ditto throughout this whole test file.



_Suggestion:_
```cpp
obj_path
```
___
*[`geometry/render/test/render_mesh_test.cc` line 564 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvnrgsDdwtuPIWGdeU:-NUvnrgsDdwtuPIWGdeV:b-8d0m4p) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L564)):*
> ```cpp
> 
> /* The OBJ has multiple intrinsic materials *applied* to the mesh. The material
>  should be the fallback material and should be accompanied by a warning. */
> ```

nit Possibly we want to allow this to parse OK in the future (when the renderers support it)?  In which case the overview paragraph here could explain that it's OK to adjust the test case in case our API spec changes.
___
*[`geometry/render/test/render_mesh_test.cc` line 597 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvommy2BI8TRp-ObUW:-NUvommy2BI8TRp-ObUX:b-tnqp5k) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L597)):*
> ```cpp
>   const fs::path obj_path = obj_dir / "mtl_elsewhere.obj";
>   {
>     std::ofstream out(obj_path);
> ```

nit The `WriteFile` does allow for a non-default `temp_dir` argument.  Should we be using that helper function here?
___
*[`geometry/render/test/render_mesh_test.cc` line 630 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvp0Fm8IGtiuxF2D0Y:-NUvp0Fm8IGtiuxF2D0Z:bhdeopz) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L630)):*
> ```cpp
>         usemtl test_material_2
>         f 1//1 2//1 3//1)""",
>                                                   multi_mtl),
> ```

BTW is `multi_mtl` here possibly more confusing than necessary?  It seems to me like even a single-mtl would be sufficient to trigger the warning, and would help isolate exactly what we're testing.
___
*[`geometry/render/test/render_mesh_test.cc` line 676 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvpMjr994DMBFBdv6Z:-NUvpMjr994DMBFBdv6_:bif08vg) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L676)):*
> ```cpp
>   const RenderMesh mesh = LoadRenderMeshFromObj(
>       obj_path, empty_props(), kDefaultDiffuse, diagnostic_policy_);
>   EXPECT_THAT(
> ```

nit This is a little inconsistent with its brethren.  In all of the other test cases, we first do the two EXPECTs on `mesh.material` and then do the EXPECT on the warning.  This does it in the opposite order.

SeanCurtis-TRI (commented):
Comments addressed, plenty of TODOs added.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUwGXrIC7swb5-QiUNf:bx7t7n8)* status: needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

___
*[`geometry/render/render_mesh.h` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhEwYD6xYxQAkHiH5w:-NUw02rADzuhsFW_azsl:b-wp3xqd) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L42)):*

BTW Is there any invariant relating `has_tex_coord` to the contents of `material`?  If so, we should mention it here.  (I assume there's not any invariant.)
</blockquote></details>

My gut suggests there *should* now be an invariant.

A material that reports a texture for a mesh that doesn't have texture coordinates is meaningless. I suspect this is ultimately something that can simply be retired. I've put a TODO in.

It mostly affects the GL renderer. I believe resolving materials there will be the ideal time to resolve this.
___
*[`geometry/render/render_mesh.h` line 89 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhFIqA1KrI6Sb47RQ6:-NUvwmki39QT4dW0tyb5:b-obzcu9) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L89)):*

It seems odd to throw on bad data when we have a juicy `policy` object sitting there ready to post errors.  Should we switch to calling that instead, whether in this PR or a future one?  When defaulted it would be as good as currently, and when not defaulted could be hooked by the caller to provide even more context about how we ended up here.
</blockquote></details>

So, your basic view is that for cases that are true failures, we should be able to log it to the policy and return an "empty" mesh. The caller decides what to do about it (via a combination of configuring the policy passed and/or explicit handling of empty returns).

That seems a meaningful path forward. Take a look at my TODO and decide if this sufficiently covers things.
___
*[`geometry/render/render_mesh.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhHmV2CcWCRWuwW01r:-NUw59SRDEBBptYtG-PW:bvnmg98) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L53)):*

The prior paragraph is talking about obj vs mtl paths.  But I think really what it's trying to get at is mtl vs png paths.  Can you double-check the wording?
</blockquote></details>

It's actually all three. I've tried rephrasing.
___
*[`geometry/render/render_mesh.cc` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhIHZ452iQeB-GSNXB:-NUw4sxTEqeAvQpvsb0_:bjhirst) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.cc#L59)):*

Ah, I see.  If the mtl file has a color but no texmap, that's still a valid texture.  That's obvious once I've said it, but it took me long enough to figure out that it's probably worth adding a comment here.
</blockquote></details>

It all went away, anyways.
___
*[`geometry/render/test/render_mesh_test.cc` line 564 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUvnrgsDdwtuPIWGdeU:-NUwAHzl2PcTf5YrsowT:bcge7k8) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/test/render_mesh_test.cc#L564)):*

nit Possibly we want to allow this to parse OK in the future (when the renderers support it)?  In which case the overview paragraph here could explain that it's OK to adjust the test case in case our API spec changes.
</blockquote></details>

The real solution is when the TODO on `LoadRenderMeshFromObj()` and we return a set. The simple reality, we'll never generally support multiple materials *per render mesh* (that flies in the face of a meaningful shader library). It means one obj gets turned into a set of `RenderMesh` instances.

SeanCurtis-TRI (commented):
+@ggould-tri for platform, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUwhvxlB3OAagrASXYg:b4qukgs)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`geometry/render/render_mesh.h` line 89 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUhFIqA1KrI6Sb47RQ6:-NUwjEGm2e7mCb6cVcg8:bza158m) ([raw file](https://github.com/RobotLocomotion/drake/blob/093e5887a292ef277a4d79d721b48b7ecb98d66c/geometry/render/render_mesh.h#L89)):*

So, your basic view is that for cases that are true failures, we should be able to log it to the policy and return an "empty" mesh. The caller decides what to do about it (via a combination of configuring the policy passed and/or explicit handling of empty returns).

That seems a meaningful path forward. Take a look at my TODO and decide if this sufficiently covers things.
</blockquote></details>

OK I hadn't thought it through quite far enough to realize that "What does an errorful RenderMesh look like?" would be the sticking point.  The new TODO is good.

ggould-tri (commented):
:lgtm:

> Reviewed 6 of 9 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUwrVrmBGaAxIojG1D8:bnfp4nl)* status: 4 unresolved discussions

___
*[`geometry/render/render_mesh.h` line 70 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUwo5IPBhy_Fg7vqqzT:-NUwo5IPBhy_Fg7vqqzU:b-9aaot) ([raw file](https://github.com/RobotLocomotion/drake/blob/713f61bb0842e764f25ffc2446243e7bee8bf523/geometry/render/render_mesh.h#L70)):*
> ```objc
>  specification is flawed. The derivation does its best to provide the "best"
>  approximation of the declared material. But the fact it was declared and
>  applied is respected. For example:
> ```

minor:  I found it hard to follow the logic of this paragraph; consider being a bit more explicit here.



_Suggestion:_
```objc
 applied is respected. For example, the following are still errors:
```
___
*[`geometry/render/render_mesh.cc` line 283 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUwpzCG68soIGI5ktcA:-NUwpzCG68soIGI5ktcB:bfhgzoa) ([raw file](https://github.com/RobotLocomotion/drake/blob/713f61bb0842e764f25ffc2446243e7bee8bf523/geometry/render/render_mesh.cc#L283)):*
> ```cpp
>           referenced_materials.begin(), referenced_materials.end(),
>           std::back_inserter(mat_names), [&reader](int i) {
>             return i < 0 ? std::string("__default__")
> ```

BTW:  I can't figure out where this magic string is defined/required/justified.  Am I missing something?
___
*[`geometry/render/test/render_mesh_test.cc` line 234 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUwqWW-1w9MRIxf_Hg8:-NUwqWW-1w9MRIxf_Hg9:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/713f61bb0842e764f25ffc2446243e7bee8bf523/geometry/render/test/render_mesh_test.cc#L234)):*
> ```cpp
>     Vector3d n0;
>     Vector2d uv0;
>     // The expected normal and uvfor the vertices in face 1 (f1).
> ```

typo



_Suggestion:_
```cpp
    // The expected normal and uv for the vertices in face 1 (f1).
```
___
*[`geometry/render/test/render_mesh_test.cc` line 542 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-NUwr4it3skzIFcXhiJy:-NUwr4it3skzIFcXhiJz:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/713f61bb0842e764f25ffc2446243e7bee8bf523/geometry/render/test/render_mesh_test.cc#L542)):*
> ```cpp
>  relative to the mtl file, but we don't know the mtl file that got read. So,
>  we *assume* mtl and obj are in the same place. If not, it can erroneously lead
>  to not being able to located used textures. The victory condition is that when
> ```

typo



_Suggestion:_
```cpp
 to not being able to locate used textures. The victory condition is that when
```

SeanCurtis-TRI (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364#-:-NUx6EEK-8JlonpCPNog:b-zcjvte)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),jwnimmer-tri(platform)

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19364)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),jwnimmer-tri(platform)

PR #19368, opened 2023-05-05
Refactor CspaceFreePolytope

AlexandreAmice:
Refactors CspaceFreePolytope internals to make it easier to develop the path certifier.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -19,7 +19,10 @@ drake_cc_package_library(
         ":c_iris_collision_geometry",
         ":c_iris_separating_plane",
         ":convex_set",
+        ":cspace_free_internal",
         ":cspace_free_polytope",
+        ":cspace_free_structs",
+        ":cspace_separating_plane",
         ":graph_of_convex_sets",
         ":iris",
     ],
@@ -81,12 +84,23 @@ drake_cc_library(
 )
 
 drake_cc_library(
+    name = "cspace_separating_plane",
+    srcs = ["cspace_separating_plane.cc"],
+    hdrs = ["cspace_separating_plane.h"],
+    deps = [
+        "//common/symbolic:polynomial",
+        "//geometry/optimization:c_iris_collision_geometry",
+    ],
+)
+
+drake_cc_library(
     name = "c_iris_separating_plane",
     srcs = ["c_iris_separating_plane.cc"],
     hdrs = ["c_iris_separating_plane.h"],
     deps = [
         "//common/symbolic:polynomial",
         "//geometry/optimization:c_iris_collision_geometry",
+        "//geometry/optimization:cspace_separating_plane",
     ],
 )
 
@@ -101,10 +115,38 @@ drake_cc_library(
         "//common/symbolic:monomial_util",
         "//geometry/optimization:c_iris_collision_geometry",
         "//geometry/optimization:c_iris_separating_plane",
+        "//geometry/optimization:cspace_free_internal",
+        "//geometry/optimization:cspace_free_structs",
         "//solvers:choose_best_solver",
         "//solvers:mathematical_program",
         "//solvers:mosek_solver",
         "//solvers:solve",
+    ],
+)
+
+drake_cc_library(
+    name = "cspace_free_internal",
+    srcs = [
+        "cspace_free_internal.cc",
+    ],
+    hdrs = ["cspace_free_internal.h"],
+    deps = [
+        "//geometry/optimization:c_iris_collision_geometry",
+        "//geometry/optimization:cspace_free_structs",
+        "//geometry/optimization:cspace_separating_plane",
+        "//solvers:choose_best_solver",
+        "//solvers:mathematical_program",
+    ],
+)
+
+drake_cc_library(
+    name = "cspace_free_structs",
+    hdrs = ["cspace_free_structs.h"],
+    deps = [
+        "//common/symbolic:rational_function",
+        "//solvers:mathematical_program",
+        "//solvers:mathematical_program_result",
+        "//solvers:mosek_solver",
     ],
 )
 
@@ -180,6 +222,15 @@ drake_cc_googletest(
         ":cspace_free_polytope",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:symbolic_test_util",
+        "//geometry/optimization:cspace_free_internal",
+    ],
+)
+
+drake_cc_googletest(
+    name = "cspace_free_internal_test",
+    deps = [
+        ":c_iris_test_utilities",
+        ":cspace_free_internal",
     ],
 )
 
--- a/geometry/optimization/c_iris_separating_plane.h
+++ b/geometry/optimization/c_iris_separating_plane.h
@@ -40,8 +40,8 @@ struct CIrisSeparatingPlane : public CSpaceSeparatingPlane<T> {
                        SeparatingPlaneOrder m_plane_order,
                        const Eigen::Ref<const VectorX<T>>& m_decision_variables)
       : CSpaceSeparatingPlane<T>(m_a, m_b, m_positive_side_geometry,
-                              m_negative_side_geometry, m_expressed_body,
-                              m_decision_variables),
+                                 m_negative_side_geometry, m_expressed_body,
+                                 m_decision_variables),
         plane_order{m_plane_order} {}
 
   SeparatingPlaneOrder plane_order;
--- a/geometry/optimization/c_iris_separating_plane.h
+++ b/geometry/optimization/c_iris_separating_plane.h
@@ -9,18 +9,17 @@
  </pre>
  */
 #pragma once
-
 #include <utility>
 
-#include "drake/common/symbolic/polynomial.h"
-#include "drake/geometry/optimization/c_iris_collision_geometry.h"
+#include "drake/geometry/optimization/cspace_separating_plane.h"
 
 namespace drake {
 namespace geometry {
 namespace optimization {
 /** The separating plane aᵀx + b ≥ δ, aᵀx+b ≤ −δ has parameters a and b. These
  parameters a polynomial function of `s_for_plane` with the specified degree.
- `s_for_plane` is a subset of the configuration-space variable `s`, please refer
+ `s_for_plane` is a sub
+ set of the configuration-space variable `s`, please refer
  to RationalForwardKinematics class or the paper above on the meaning of s.
  */
 // TODO(hongkai.dai): I might support kConstant in the future.
@@ -28,16 +27,8 @@ enum class SeparatingPlaneOrder {
   kAffine,  ///< a and b are affine function of s.
 };
 
-/**
- Wraps the information that a pair of collision geometries are separated by a
- plane.
- One collision geometry is on the "positive" side of the separating
- plane, namely {x| aᵀx + b ≥ δ} (with δ ≥ 0}, and the other collision geometry
- is on the "negative" side of the separating plane, namely {x|aᵀx+b ≤ −δ}.
- @tparam T The type of decision_variables. T= symbolic::Variable or double.
- */
 template <typename T>
-struct CIrisSeparatingPlane {
+struct CIrisSeparatingPlane : public CSpaceSeparatingPlane<T> {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(CIrisSeparatingPlane)
 
@@ -48,28 +39,12 @@ struct CIrisSeparatingPlane {
                        multibody::BodyIndex m_expressed_body,
                        SeparatingPlaneOrder m_plane_order,
                        const Eigen::Ref<const VectorX<T>>& m_decision_variables)
-      : a{std::move(m_a)},
-        b{std::move(m_b)},
-        positive_side_geometry{m_positive_side_geometry},
-        negative_side_geometry{m_negative_side_geometry},
-        expressed_body{m_expressed_body},
-        plane_order{m_plane_order},
-        decision_variables{m_decision_variables} {}
+      : CSpaceSeparatingPlane<T>(m_a, m_b, m_positive_side_geometry,
+                              m_negative_side_geometry, m_expressed_body,
+                              m_decision_variables),
+        plane_order{m_plane_order} {}
 
-  /// Return the geometry on the specified side.
-  [[nodiscard]] const CIrisCollisionGeometry* geometry(
-      PlaneSide plane_side) const {
-    return plane_side == PlaneSide::kPositive ? positive_side_geometry
-                                              : negative_side_geometry;
-  }
-
-  Vector3<symbolic::Polynomial> a;
-  symbolic::Polynomial b;
-  const CIrisCollisionGeometry* positive_side_geometry;
-  const CIrisCollisionGeometry* negative_side_geometry;
-  multibody::BodyIndex expressed_body;
   SeparatingPlaneOrder plane_order;
-  VectorX<T> decision_variables;
 };
 
 /**
--- a/geometry/optimization/c_iris_separating_plane.h
+++ b/geometry/optimization/c_iris_separating_plane.h
@@ -29,7 +29,6 @@ enum class SeparatingPlaneOrder {
 
 template <typename T>
 struct CIrisSeparatingPlane : public CSpaceSeparatingPlane<T> {
- public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(CIrisSeparatingPlane)
 
   CIrisSeparatingPlane(Vector3<symbolic::Polynomial> m_a,
@@ -44,7 +43,7 @@ struct CIrisSeparatingPlane : public CSpaceSeparatingPlane<T> {
                                  m_decision_variables),
         plane_order{m_plane_order} {}
 
-  SeparatingPlaneOrder plane_order;
+  SeparatingPlaneOrder plane_order{SeparatingPlaneOrder::kAffine};
 };
 
 /**
--- a/geometry/optimization/cspace_free_internal.cc
+++ b/geometry/optimization/cspace_free_internal.cc
@@ -14,7 +14,7 @@ namespace drake {
 namespace geometry {
 namespace optimization {
 namespace internal {
-
+namespace {
 struct BodyPair {
   BodyPair(multibody::BodyIndex m_body1, multibody::BodyIndex m_body2)
       : body1{m_body1}, body2{m_body2} {}
@@ -29,18 +29,6 @@ struct BodyPair {
 struct BodyPairHash {
   size_t operator()(const BodyPair& p) const { return p.body1 * 100 + p.body2; }
 };
-
-int GetNumYInRational(const symbolic::RationalFunction& rational,
-                      const Vector3<symbolic::Variable>& y_slack) {
-  int count = 0;
-  for (int i = 0; i < 3; ++i) {
-    if (rational.numerator().indeterminates().find(y_slack(i)) !=
-        rational.numerator().indeterminates().end()) {
-      ++count;
-    }
-  }
-  return count;
-}
 
 // Returns true if all the bodies on the kinematics chain from `start` to `end`
 // are welded together (namely all the mobilizers in between are welded).
@@ -92,6 +80,19 @@ GetLinkCollisionPairs(
     }
   }
   return ret;
+}
+}  // namespace
+
+int GetNumYInRational(const symbolic::RationalFunction& rational,
+                      const Vector3<symbolic::Variable>& y_slack) {
+  int count = 0;
+  for (int i = 0; i < 3; ++i) {
+    if (rational.numerator().indeterminates().find(y_slack(i)) !=
+        rational.numerator().indeterminates().end()) {
+      ++count;
+    }
+  }
+  return count;
 }
 
 std::map<multibody::BodyIndex,
@@ -157,7 +158,7 @@ int GenerateCollisionPairs(
 void GenerateRationals(
     const std::vector<std::unique_ptr<
         CSpaceSeparatingPlane<symbolic::Variable>>>& separating_planes,
-    const Vector3<symbolic::Variable> y_slack,
+    const Vector3<symbolic::Variable>& y_slack,
     const Eigen::Ref<const Eigen::VectorXd>& q_star,
     const multibody::RationalForwardKinematics& rational_forward_kin,
     std::vector<PlaneSeparatesGeometries>* plane_geometries) {
@@ -172,6 +173,7 @@ void GenerateRationals(
   for (int plane_index = 0;
        plane_index < static_cast<int>(separating_planes.size());
        ++plane_index) {
+    DRAKE_ASSERT(separating_planes.at(plane_index) != nullptr);
     const auto& separating_plane = *(separating_planes.at(plane_index));
     // Compute X_AB for both side of the geometries.
     std::vector<symbolic::RationalFunction> positive_side_rationals;
@@ -264,28 +266,23 @@ void GenerateRationals(
   }
 }
 
-/**
- Solves a SeparationCertificateProgramBase with the given options.
- */
-[[nodiscard]] SeparationCertificateResultBase
-SolveSeparationCertificateProgramBase(
+void SolveSeparationCertificateProgramBase(
     const SeparationCertificateProgramBase& certificate_program,
     const FindSeparationCertificateOptions& options,
-    const CSpaceSeparatingPlane<symbolic::Variable>& separating_plane) {
-  SeparationCertificateResultBase cert_result;
+    const CSpaceSeparatingPlane<symbolic::Variable>& separating_plane,
+    SeparationCertificateResultBase* result) {
+  result->plane_index = certificate_program.plane_index;
   solvers::MakeSolver(options.solver_id)
       ->Solve(*certificate_program.prog, std::nullopt, options.solver_options,
-              &(cert_result.result));
-  if (cert_result.result.is_success()) {
-    cert_result.plane_index = certificate_program.plane_index;
-    cert_result.plane_decision_var_vals =
-        cert_result.result.GetSolution(separating_plane.decision_variables);
+              &(result->result));
+  if (result->result.is_success()) {
+    result->plane_decision_var_vals =
+        result->result.GetSolution(separating_plane.decision_variables);
     for (int i = 0; i < 3; ++i) {
-      cert_result.a(i) = cert_result.result.GetSolution(separating_plane.a(i));
+      result->a(i) = result->result.GetSolution(separating_plane.a(i));
     }
-    cert_result.b = cert_result.result.GetSolution(separating_plane.b);
+    result->b = result->result.GetSolution(separating_plane.b);
   }
-  return cert_result;
 }
 
 }  // namespace internal
--- a/geometry/optimization/cspace_free_internal.cc
+++ b/geometry/optimization/cspace_free_internal.cc
@@ -1,0 +1,294 @@
+#include "drake/geometry/optimization/cspace_free_internal.h"
+
+#include <map>
+#include <optional>
+#include <unordered_map>
+#include <utility>
+
+#include "drake/multibody/rational/rational_forward_kinematics.h"
+#include "drake/multibody/rational/rational_forward_kinematics_internal.h"
+#include "drake/solvers/choose_best_solver.h"
+#include "drake/solvers/mathematical_program_result.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+namespace internal {
+
+struct BodyPair {
+  BodyPair(multibody::BodyIndex m_body1, multibody::BodyIndex m_body2)
+      : body1{m_body1}, body2{m_body2} {}
+  bool operator==(const BodyPair& other) const {
+    return body1 == other.body1 && body2 == other.body2;
+  }
+
+  multibody::BodyIndex body1;
+  multibody::BodyIndex body2;
+};
+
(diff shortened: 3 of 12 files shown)


AlexandreAmice (commented):
+@hongkai-dai For feature review. This moves around the internals of `CspaceFreePolytope` around a bit to make the implementation of `CspaceFreePath` better. This PR tries to include most (if not all) the changes that will require a platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NUhehimBCc0lRTGHMTX:bvcftyu)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

hongkai-dai (commented):
Checkpoint.

Sorry for the slow review, I only have sporadic time these days.

> Reviewed 4 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NUmN2Cz6_PGHHEsgJEa:b-7ldksy)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/cspace_free_internal.h` line 20 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmKsQoAPdJVzrcRsOE:-NUmKsQoAPdJVzrcRsOF:bb76586) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L20)):*
> ```objc
> namespace internal {
> 
> void GenerateRationals(
> ```

BTW, document this function?
___
*[`geometry/optimization/cspace_free_internal.h` line 23 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmLaPR6mIJHEZfZX1T:-NUmLaPR6mIJHEZfZX1U:b-90al5e) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L23)):*
> ```objc
>     const std::vector<std::unique_ptr<
>         CSpaceSeparatingPlane<symbolic::Variable>>>& separating_planes,
>     const Vector3<symbolic::Variable> y_slack,
> ```

Pass by const reference?0.0000
___
*[`geometry/optimization/cspace_free_internal.h` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmLi2LB1TJS6__yR7t:-NUmLi2LB1TJS6__yR7u:blyjjdw) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L30)):*
> ```objc
> // Returns the number of y_slack variables in `rational`.
> // Not all y_slack necessarily appear in `rational`.
> int GetNumYInRational(const symbolic::RationalFunction& rational,
> ```

Add `[[nodiscard]]` before the return type?
___
*[`geometry/optimization/cspace_free_internal.h` line 35 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmMCrB70a0RGskYCUY:-NUmMCrB70a0RGskYCUZ:b-ymnrr1) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L35)):*
> ```objc
> /**
>  * Given a plant and associated scene graph, returns all
>  * the collision geometries.
> ```

BTW, also document the return type, that ret[body_index] returns all the CIrisCollisionGeometry object attached to that body?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: 6 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice)

___
*[`geometry/optimization/cspace_free_internal.h` line 66 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUrHp65209xpvG6w-io:-NUrHp65209xpvG6w-ip:be0ic4z) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L66)):*
> ```objc
>    empty.
>    */
> [[nodiscard]] SeparationCertificateResultBase
> ```

Not sure if I like this design here. We have two `Base` class here, the input `SeparationCertificateProgramBase` which allows polymorphism at the call site, but how does returning a base class work with polymorphism?

If I do
```cc
auto result = SolveSeparationCertificateProgramBase(...);
```
this `result` class will always be a `SeparationCertificateProgramBase` class, it won't be the child class type. I would think it is more natural to use an output argument instead of a return type here, like
```
void SolveSeparationCertificateProgramBase(const SeparationCertificateProgramBase&, ..., SeparationCertificateResultBase* result);
```
?
___
*[`geometry/optimization/cspace_free_internal.cc` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUrJrf96U5jcdAuWrFI:-NUrJrf96U5jcdAuWrFJ:b4t46hw) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.cc#L45)):*
> ```cpp
> }
> 
> // Returns true if all the bodies on the kinematics chain from `start` to `end`
> ```

BTW, this function doesn't show up in the header file? In that case, I would suggest to put this function within an anonymous namespace within the cc file.

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_internal.h` line 23 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmLaPR6mIJHEZfZX1T:-NUy73is0KgPbyBFQAhz:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L23)):*

Pass by const reference?0.0000
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_internal.h` line 30 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUmLi2LB1TJS6__yR7t:-NUy78Bh36YpAeviIsPq:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L30)):*

Add `[[nodiscard]]` before the return type?
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_internal.h` line 66 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUrHp65209xpvG6w-io:-NUy7Z3uDIZb89oq0bdu:b-3p6zr6) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.h#L66)):*

Not sure if I like this design here. We have two `Base` class here, the input `SeparationCertificateProgramBase` which allows polymorphism at the call site, but how does returning a base class work with polymorphism?

If I do
```cc
auto result = SolveSeparationCertificateProgramBase(...);
```
this `result` class will always be a `SeparationCertificateProgramBase` class, it won't be the child class type. I would think it is more natural to use an output argument instead of a return type here, like
```
void SolveSeparationCertificateProgramBase(const SeparationCertificateProgramBase&, ..., SeparationCertificateResultBase* result);
```
?
</blockquote></details>

Done. Great suggestion.
___
*[`geometry/optimization/cspace_free_internal.cc` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NUrJrf96U5jcdAuWrFI:-NUyFBHv6JPKvzR4KVOH:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/8061c7ba75b9b6f3ebd2702d09a534b92d740eb1/geometry/optimization/cspace_free_internal.cc#L45)):*

BTW, this function doesn't show up in the header file? In that case, I would suggest to put this function within an anonymous namespace within the cc file.
</blockquote></details>

Done

hongkai-dai (commented):
:lgtm: sorry for the delay. +@rpoyner-tri for platform review please, thanks!

> Reviewed 6 of 12 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV1An0nA84hOa_bLzHr:b-ruwjmh)* status: 3 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @AlexandreAmice and @rpoyner-tri)

___
*[`geometry/optimization/cspace_free_internal.cc` line 175 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV19fS7C5zBShYXNuOy:-NV19fS7C5zBShYXNuOz:b-yl0gm1) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.cc#L175)):*
> ```cpp
>   for (int plane_index = 0;
>        plane_index < static_cast<int>(separating_planes.size());
>        ++plane_index) {
> ```

nit. A question for platform reviewers: if we de-reference a unique_ptr, should we first check if this unique pointer is nullptr, like `DRAKE_DEMAND(separating_planes.at(plane_index).get() != nullptr)`? Even without this DRAKE_DEMAND check I think the code is fine (it will throw a segfault if the unique pointer is really a nullptr).
___
*[`geometry/optimization/cspace_free_polytope.h` line 480 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV18il24grce8QkBB1a:-NV18il24grce8QkBB1b:b-tm1821) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_polytope.h#L480)):*
> ```objc
>    collision free.
>    */
>   //  void GenerateRationals();
> ```

nit., remove this line?
___
*[`geometry/optimization/cspace_free_structs.h` line 56 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV1ARK3EruAOHqlGVHH:-NV1ARK3EruAOHqlGVHI:brxix90) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_structs.h#L56)):*
> ```objc
>  We certify that a pair of geometries is collision free by finding the
>  separating plane over a range of configuration. The Lagrangian multipliers
>  used for certifying this condition will differ. This struct contains the
> ```

Maybe say "will deffer between derived classes"?

rpoyner-tri (commented):
+(release notes: breaking change) until I can actually figure out whether the API changes are breaking or not.
Checkpoint: more tomorrow.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV1qWCt5lg149PAnY9f:bxqyqx4)* status: 6 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`geometry/optimization/cspace_free_internal.h` line 20 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV1nY4MCRWNMct9-0M6:-NV1nY4N7MCmUuoYDKZZ:b-56cqkj) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.h#L20)):*
> ```objc
> namespace internal {
> 
> /**
> ```

nit We are in the `internal` namespace, so none of this doc will be processed by Doxygen. You can still use its keywords and other syntax, but the comment blocks should be "ordinary" enclosure style: either `/* stuff */` or `// stuff`. 
___
*[`geometry/optimization/cspace_free_internal.h` line 25 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV1nEJ21ajITpvoOyaA:-NV1nEJ21ajITpvoOyaB:b-2zlkxh) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.h#L25)):*
> ```objc
>  collision free.
>  @param[in] separating_planes A vector containing the plane parameters as well
>  as information about which bodies to separate.
> ```

nit Should probably document that all of the contained unique_ptrs must not be null.
___
*[`geometry/optimization/cspace_free_internal.h` line 58 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV1oDVm00M1tFypR5sY:-NV1oDVm00M1tFypR5sZ:b5j1j3i) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.h#L58)):*
> ```objc
> 
> /**
>  * Given a plant and associated scene graph compute all the possible collision
> ```

nit This comment should indicate which of the pointer-bearing inputs require that their pointers not be null.
___
*[`geometry/optimization/cspace_free_internal.cc` line 175 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV19fS7C5zBShYXNuOy:-NV1lHlO3tFcHYklZRQ5:bamqgu0) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.cc#L175)):*

nit. A question for platform reviewers: if we de-reference a unique_ptr, should we first check if this unique pointer is nullptr, like `DRAKE_DEMAND(separating_planes.at(plane_index).get() != nullptr)`? Even without this DRAKE_DEMAND check I think the code is fine (it will throw a segfault if the unique pointer is really a nullptr).
</blockquote></details>

I looked at the docs for `unique_ptr::operator*`. It's even worse than I imagined. Looks like whether or not `*` throws on nullptr is implementation-dependent. For comparison, `operator->` is guaranteed to never throw (which means you always get Undefined Behavior instead). So, if you don't check, you are courting Undefined Behavior (of which segfault is an example), which is bad.

So I'd say you need to do something. If this is a high-performance code path, then maybe DRAKE_ASSERT is acceptable. If you think you can afford to always check, then DRAKE_DEMAND.

AlexandreAmice (commented):
Some context might be helpful. I am currently working on a project that shares a good bit of code with `cspace_free_polytope` but is not appropriate to subclass from. This PR pulls out some of the internal code that was previously in `cspace_free_polytope` so that it can be effectively used by `cspace_free_path` which is currently in `geometry/optimization/dev`. The plane is to update that code once this PR passes.

TLDR: Nothing here should be a breaking change and indeed none of the tests in `cspace_free_polytope_test` and `cspace_free_polytope_with_mosek` required changing. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV2Fl2v3qSWMWzSj06e:b3g29bc)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice and @rpoyner-tri)

___
*[`geometry/optimization/cspace_free_internal.cc` line 175 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV19fS7C5zBShYXNuOy:-NV2JOgQ7i-w4rJM1FsE:b-a427f4) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.cc#L175)):*

I looked at the docs for `unique_ptr::operator*`. It's even worse than I imagined. Looks like whether or not `*` throws on nullptr is implementation-dependent. For comparison, `operator->` is guaranteed to never throw (which means you always get Undefined Behavior instead). So, if you don't check, you are courting Undefined Behavior (of which segfault is an example), which is bad.

So I'd say you need to do something. If this is a high-performance code path, then maybe DRAKE_ASSERT is acceptable. If you think you can afford to always check, then DRAKE_DEMAND.


</blockquote></details>

This code only gets called once per object creation of `cspace_free_polytope` and will be called only once per object creation of `cspace_free_path`.  Therefore, I expect that it won't be called too frequently and so `DRAKE_DEMAND` should be fine. 
___
*[`geometry/optimization/cspace_free_structs.h` line 56 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV1ARK3EruAOHqlGVHH:-NV2KvcT4APv5HA6HU5I:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_structs.h#L56)):*

Maybe say "will deffer between derived classes"?
</blockquote></details>

Done.

hongkai-dai (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: LGTM missing from assignee rpoyner-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice and @rpoyner-tri)

___
*[`geometry/optimization/cspace_free_internal.cc` line 175 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV19fS7C5zBShYXNuOy:-NV5VQp63D5QXuxPAU7E:b2oxbwj) ([raw file](https://github.com/RobotLocomotion/drake/blob/8e34de2926095633ebdc60e7874d4d41997de90d/geometry/optimization/cspace_free_internal.cc#L175)):*

This code only gets called once per object creation of `cspace_free_polytope` and will be called only once per object creation of `cspace_free_path`.  Therefore, I expect that it won't be called too frequently and so `DRAKE_DEMAND` should be fine. 
</blockquote></details>

Thanks @rpoyner-tri for the check, that is good to now!

rpoyner-tri (commented):
:lgtm: There is one breaking change, noted below. I'll let you guys discuss how you want to handle it. Otherwise, buncha minor stuff.

> Reviewed 5 of 12 files at r1, 1 of 3 files at r2, 2 of 2 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV5Z54DAOlEa8OOB7sA:b-chtrfi)* status: 6 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`geometry/optimization/c_iris_separating_plane.h` line 32 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5LDrtEdIIcTDGrg3T:-NV5LDru59xZhxB0zoC1:b-c3ly1l) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/c_iris_separating_plane.h#L32)):*
> ```objc
> template <typename T>
> struct CIrisSeparatingPlane : public CSpaceSeparatingPlane<T> {
>  public:
> ```

nit `public:` is not necessary here; structs are `public` by default.
___
*[`geometry/optimization/c_iris_separating_plane.h` line 47 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5YfF_65PmwKbTLw2A:-NV5YfF_65PmwKbTLw2B:b-gh07b7) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/c_iris_separating_plane.h#L47)):*
> ```objc
>         plane_order{m_plane_order} {}
> 
>   SeparatingPlaneOrder plane_order;
> ```

Primitive-type (even enum!) member fields need to be initialized.
___
*[`geometry/optimization/cspace_free_polytope.h` line 708 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5Vczz4p-XB8b7PXZz:-NV5Vczz4p-XB8b7PX_-:bfwjbpb) ([raw file](https://github.com/RobotLocomotion/drake/blob/04e34592a90cce3cf1adc1c5d999fa091a19f352/geometry/optimization/cspace_free_polytope.h#L708)):*
> ```objc
> [[nodiscard]] std::map<multibody::BodyIndex,
>                        std::vector<std::unique_ptr<CIrisCollisionGeometry>>>
> GetCollisionGeometries(const multibody::MultibodyPlant<double>& plant,
> ```

nit This function moved from a public namespace to `internal`. We can either treat this as a breaking change, or retain this function as a wrapper that calls the internal function, or do a deprecation. If there are arguably few/no external users, then declaring a breaking change might still be a reasonable choice.
___
*[`geometry/optimization/cspace_free_structs.h` line 30 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5WnvjBmaVnYSjosUE:-NV5WnvjBmaVnYSjosUF:b-s6mkx1) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L30)):*
> ```objc
>   const std::vector<symbolic::RationalFunction> positive_side_rationals;
>   const std::vector<symbolic::RationalFunction> negative_side_rationals;
>   int plane_index;
> ```

Primitive-type struct members need to be initialized. 
___
*[`geometry/optimization/cspace_free_structs.h` line 62 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5WxOaDumWg0yUgEVM:-NV5WxOaDumWg0yUgEVN:b-s6mkx1) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L62)):*
> ```objc
>  */
> struct SeparationCertificateResultBase {
>   int plane_index;
> ```

Primitive-type struct members need to be initialized. 
___
*[`geometry/optimization/cspace_free_structs.h` line 80 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5X-285OmpzWUnRDbG:-NV5X-285OmpzWUnRDbH:b-s6mkx1) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L80)):*
> ```objc
>   /// plane and Lagrangian multipliers as certificate.
>   std::unique_ptr<solvers::MathematicalProgram> prog;
>   int plane_index;
> ```

Primitive-type struct members need to be initialized.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: 6 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`geometry/optimization/cspace_free_polytope.h` line 708 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5Vczz4p-XB8b7PXZz:-NV5cVkY2kCEwYmnk_xa:b-rck2ue) ([raw file](https://github.com/RobotLocomotion/drake/blob/04e34592a90cce3cf1adc1c5d999fa091a19f352/geometry/optimization/cspace_free_polytope.h#L708)):*

nit This function moved from a public namespace to `internal`. We can either treat this as a breaking change, or retain this function as a wrapper that calls the internal function, or do a deprecation. If there are arguably few/no external users, then declaring a breaking change might still be a reasonable choice.
</blockquote></details>

Thanks Rico for catching this!

I don't think users will call this function (except us who dived into C-IRIS). I am OK to move this function from the public namespace to `internal`. So I am fine to declare a breaking change.

jwnimmer-tri:
@rpoyner-tri please check me on the deprecation, then if you agree switch the label (breaking => newly deprecated).

rpoyner-tri (commented):
-(release notes: breaking change) +(release notes: newly deprecated) 

> Reviewed 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV5mANr6X-VrQLCwdPn:b-27xk)* status: 5 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

AlexandreAmice (commented):
Thanks for the review. Can you check me on the primitive type member fields needing to be initialized? 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV5qyEn5Q-OHuG6BBnq:b11prf2)* status: 4 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`geometry/optimization/c_iris_separating_plane.h` line 47 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5YfF_65PmwKbTLw2A:-NV5qcix7TMw-FRdJeMl:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/c_iris_separating_plane.h#L47)):*

Primitive-type (even enum!) member fields need to be initialized.
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_structs.h` line 30 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5WnvjBmaVnYSjosUE:-NV5qsWo-lAYMDuy4ch0:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L30)):*

Primitive-type struct members need to be initialized. 
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_structs.h` line 62 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5WxOaDumWg0yUgEVM:-NV5qvFe-0xp11oHek8u:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L62)):*

Primitive-type struct members need to be initialized. 
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_structs.h` line 80 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5X-285OmpzWUnRDbG:-NV5qx2W9rjIP7K1ouUx:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa94f42857f0018518f5a39fa68d1568f0c0b43b/geometry/optimization/cspace_free_structs.h#L80)):*

Primitive-type struct members need to be initialized. 
</blockquote></details>

Done.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

___
*[`geometry/optimization/c_iris_separating_plane.h` line 46 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-NV5tR7p1jhRXWqsfPqt:-NV5tR7p1jhRXWqsfPqu:buly23i) ([raw file](https://github.com/RobotLocomotion/drake/blob/647d3b8c45a7b2a64b6e77cc8d380fc6ec430a0d/geometry/optimization/c_iris_separating_plane.h#L46)):*
> ```objc
>         plane_order{m_plane_order} {}
> 
>   SeparatingPlaneOrder plane_order{separating_plane_order::kAffine};
> ```

nit CI is sad because it wants `SeparatingPlaneOrder plane_order{SeparatingPlaneOrder::kAffine};` instead.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @AlexandreAmice)

rpoyner-tri (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19368#-:-NV6XJ4F4HEHtND340o6:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),hongkai-dai

PR #19393, opened 2023-05-11
Improve math::AreAutoDiffVecXdEqual

RussTedrake:
Previously, it was possible that ExtractGradient (and therefore AreAutoDiffVecXdEqual) would throw, when this method should simply return false. Now we handle the case where the elements of the vector could have a different number of derivative values.

This version should also be slightly faster.

N.B. This came up for me because I had two different constraints setting values in a shared Context, but one was calling SetPositions(), and the other was calling SetPositionsAndVelocities().

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19393)

### DIFF
--- a/math/autodiff_gradient.cc
+++ b/math/autodiff_gradient.cc
@@ -11,13 +11,15 @@ bool AreAutoDiffVecXdEqual(const Eigen::Ref<const VectorX<AutoDiffXd>>& a,
   if (math::ExtractValue(a) != math::ExtractValue(b)) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


RussTedrake (commented):
+@sammy-tri for both reviews, please?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19393#-:-NV9q4AU69WIDLLsub6x:b1yb49p)* status: LGTM missing from assignee sammy-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake and @sammy-tri)

PR #19392, opened 2023-05-10
[WIP] Implements updated SAP constraints framework.

amcastro-tri:
Infrastructure towards #19320. 

Thus far a WIP PR to get some early feedback. Since this PR touches the base `SapConstraint` class, a full PR requires to update all constraints and their unit tests, even if the math is the same. 

What this PR does is to write SAP constraints in terms of the cost, its gradient (impulses) and its Hessian. While the original SAP constraints are tied to the idea of "regularization", the new formulation allows for arbitrary functional forms (eg. the H&C model we want in #19320). 

The first important aspects to review in this PR are the new virtual functions in `SapConstraint` that specific constraints must implement, namely:
 1. DoMakeData()
 2. DoCalcData()
 3. DoCalcCost()
 4. DoCalcImpulse()  (minus the gradient of the cost)
 5. DoCalcCostHessian()

The second important aspect is that all of these methods work with an `AbstractValue` data (a `void*` in C terms). Since constraints are polymorphic, the idea is to allow specific constraints to make their own data type as needed with `MakeData()`.

## What's fully complete
 1. Updated `SapFrictionConeConstraint` and unit tests
 2. Updated `SapLimitConstraint`, not the unit tests yet.
 3. Updated `SapConstraintBundle`, `SapModel` and `SapSolver` to use the new infrastrucute.
 4. `SapSolver` tests are updated and pass (good!)
 5. Examples with contact and limits work as expected (for instance, the Atlas example)

## What's missing:
 1. Update the remaining constraints to use the new framework.
 2. Update unit tests. The beauty of the new framework is that we can use AutoDiff for all constraints for unit tests. I believe I could write a base unit test for all constraints and reuse, so possibly not that much work.
 3. Beautify documentation.

Issue #1939, opened 2016-03-25
discuss/resolve the way we handle external dependencies


RussTedrake:
I've started a google doc here:
https://docs.google.com/document/d/1RMDja0UcU1t5ASRSeu7kW2ZuWMb7u6mhswWPP4QT4xU/edit?usp=sharing

Please comment (right on the google doc)!  My goal is to flesh this out, make some decisions, and the move this into sphinx documentation.

patmarion:
Perhaps the google doc can also indicate whether a dependency is required or optional, and if optional, what additional functionality it provides?

RussTedrake:
Added required/optional column

david-german-tri:
Since the Google Docs margin note conversations have died inconclusively (after a lot of great data gathering), I'll try to restart this thread.  

@jwnimmer-tri pointed out that the success criteria for changes to dependency management are unclear, and I agree.  @RussTedrake, maybe you can tell us more about the decisions you want this thread to yield?

liangfok:
Suggested success criteria: Drake successfully compiles and runs 100% of the time assuming users follow the [system configuration requirements](http://drake.mit.edu/developers.html#supported-configurations) and [installation instructions](http://drake.mit.edu/installation.html) to a T.

liangfok:
Suggested secondary success criteria: 100% of users can compile and run Drake without installing any proprietary or code with viral licenses.

david-german-tri:
friendly ping to @RussTedrake re. https://github.com/RobotLocomotion/drake/issues/1939#issuecomment-221438529

jwnimmer-tri:
I wonder if we should just close this issue?

david-german-tri:
I believe we should close this issue, yes.  The relationship to external dependencies will be more explicit and clear in Bazel, and we've already agreed that the supported, CI-tested modes will be fewer: "open source" and "everything".

PR #19402, opened 2023-05-12
Introduce new SapConstraint APIs implemented in terms of older APIs.

amcastro-tri:
Towards landing the full PR #19392.

This PR introduces the new set of `SapConstraint` APIs but implements them in terms of the old APIs so that we can make the migration towards #19392 progressively. Therefore specific constraints are not touched in this PR.

The PR also teaches `SapConstraintBundle`, `SapModel` and `SapSolver` how to use the new APIs. With these new APIs the cost assigned to each constraint no longer needs to have the form of a quadratic regularizer. This will allow us to introduce more complex models of contact.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)

### DIFF
--- a/multibody/contact_solvers/sap/BUILD.bazel
+++ b/multibody/contact_solvers/sap/BUILD.bazel
@@ -55,6 +55,8 @@ drake_cc_library(
     deps = [
         "//common:default_scalars",
         "//common:essential",
+        "//common:unused",
+        "//common:value",
         "//multibody/contact_solvers:matrix_block",
     ],
 )
@@ -182,6 +184,7 @@ drake_cc_googletest(
         ":sap_constraint_bundle",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
+        "//math:gradient",
         "//multibody/contact_solvers:block_sparse_matrix",
     ],
 )
--- a/multibody/contact_solvers/sap/sap_constraint.cc
+++ b/multibody/contact_solvers/sap/sap_constraint.cc
@@ -40,6 +40,88 @@ SapConstraint<T>::SapConstraint(int first_clique, int second_clique,
                      first_clique_jacobian().rows());
 }
 
+template <typename T>
+std::unique_ptr<AbstractValue> SapConstraint<T>::MakeData(
+    const T& time_step,
+    const Eigen::Ref<const VectorX<T>>& delassus_estimation) const {
+  DRAKE_DEMAND(delassus_estimation.size() == num_constraint_equations());
+  return DoMakeData(time_step, delassus_estimation);
+}
+
+template <typename T>
+void SapConstraint<T>::CalcData(const Eigen::Ref<const VectorX<T>>& vc,
+                                AbstractValue* data) const {
+  DRAKE_DEMAND(vc.size() == num_constraint_equations());
+  DRAKE_DEMAND(data != nullptr);
+  DoCalcData(vc, data);
+}
+
+template <typename T>
+T SapConstraint<T>::CalcCost(const AbstractValue& data) const {
+  return DoCalcCost(data);
+}
+
+template <typename T>
+void SapConstraint<T>::CalcImpulse(const AbstractValue& data,
+                                   EigenPtr<VectorX<T>> gamma) const {
+  DRAKE_DEMAND(gamma != nullptr);
+  DoCalcImpulse(data, gamma);
+}
+
+template <typename T>
+void SapConstraint<T>::CalcCostHessian(const AbstractValue& data,
+                                       MatrixX<T>* G) const {
+  DRAKE_DEMAND(G != nullptr);
+  DoCalcCostHessian(data, G);
+}
+
+template <typename T>
+std::unique_ptr<AbstractValue> SapConstraint<T>::DoMakeData(
+    const T& time_step,
+    const Eigen::Ref<const VectorX<T>>& delassus_estimation) const {
+  // TODO(amcastro-tri): make use of the full Delassus estimation, not just the
+  // mean.
+  const T wi = delassus_estimation.sum() / delassus_estimation.size();
+  VectorX<T> v_hat = CalcBiasTerm(time_step, wi);
+  VectorX<T> R = CalcDiagonalRegularization(time_step, wi);
+  SapConstraintData<T> data(std::move(R), std::move(v_hat));
+  return AbstractValue::Make(data);
+}
+
+template <typename T>
+void SapConstraint<T>::DoCalcData(const Eigen::Ref<const VectorX<T>>& vc,
+                                  AbstractValue* abstract_data) const {
+  auto& data = abstract_data->get_mutable_value<SapConstraintData<T>>();
+  const VectorX<T>& R_inv = data.R_inv();
+  const VectorX<T>& v_hat = data.v_hat();
+  data.mutable_vc() = vc;
+  data.mutable_y() = R_inv.asDiagonal() * (v_hat - vc);
+  Project(data.y(), data.R(), &data.mutable_gamma(), &data.mutable_dPdy());
+}
+
+template <typename T>
+T SapConstraint<T>::DoCalcCost(const AbstractValue& abstract_data) const {
+  const auto& data = abstract_data.get_value<SapConstraintData<T>>();
+  const VectorX<T>& R = data.R();
+  const VectorX<T>& gamma = data.gamma();
+  const T cost = 0.5 * gamma.dot(R.asDiagonal() * gamma);
+  return cost;
+}
+
+template <typename T>
+void SapConstraint<T>::DoCalcImpulse(const AbstractValue& abstract_data,
+                                     EigenPtr<VectorX<T>> gamma) const {
+  const auto& data = abstract_data.get_value<SapConstraintData<T>>();
+  *gamma = data.gamma();
+}
+
+template <typename T>
+void SapConstraint<T>::DoCalcCostHessian(const AbstractValue& abstract_data,
+                                         MatrixX<T>* G) const {
+  const auto& data = abstract_data.get_value<SapConstraintData<T>>();
+  *G = data.dPdy() * data.R_inv().asDiagonal();
+}
+
 }  // namespace internal
 }  // namespace contact_solvers
 }  // namespace multibody
--- a/multibody/contact_solvers/sap/sap_constraint.h
+++ b/multibody/contact_solvers/sap/sap_constraint.h
@@ -5,12 +5,77 @@
 
 #include "drake/common/drake_copyable.h"
 #include "drake/common/eigen_types.h"
+#include "drake/common/unused.h"
+#include "drake/common/value.h"
 #include "drake/multibody/contact_solvers/matrix_block.h"
 
 namespace drake {
 namespace multibody {
 namespace contact_solvers {
 namespace internal {
+
+// TODO(amcastro-tri): SapConstraintData is only temporary to aid the migration
+// towards #19392 and it will soon be removed.
+/* Class to store data needed for SapConstraint computations. */
+template <typename T>
+class SapConstraintData {
+ public:
+  DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(SapConstraintData);
+
+  /* Constructs data for a SapLimitConstraint.
+     @param R Regularization parameters.
+     @param v_hat Bias term.
+     @warning Data stored is left uninitialized to avoid the cost of an
+     unnecessary initialization. */
+  SapConstraintData(VectorX<T> R, VectorX<T> v_hat) {
+    const int nk = R.size();
+    parameters_.R.resize(nk);
+    parameters_.R_inv.resize(nk);
+    parameters_.v_hat.resize(nk);
+    parameters_.R = R;
+    parameters_.R_inv = R.cwiseInverse();
+    parameters_.v_hat = v_hat;
+    vc_.resize(nk);
+    y_.resize(nk);
+    gamma_.resize(nk);
+    dPdy_.resize(nk, nk);
+  }
+
+  /* Regularization R. */
+  const VectorX<T>& R() const { return parameters_.R; }
+
+  /* Inverse of the regularization, R⁻¹. */
+  const VectorX<T>& R_inv() const { return parameters_.R_inv; }
+
+  /* Constraint bias. */
+  const VectorX<T>& v_hat() const { return parameters_.v_hat; }
+
+  /* Const access. */
+  const VectorX<T>& vc() const { return vc_; }
+  const VectorX<T>& y() const { return y_; }
+  const VectorX<T>& gamma() const { return gamma_; }
+  const MatrixX<T>& dPdy() const { return dPdy_; }
+
+  /* Mutable access. */
+  VectorX<T>& mutable_vc() { return vc_; }
+  VectorX<T>& mutable_y() { return y_; }
+  VectorX<T>& mutable_gamma() { return gamma_; }
+  MatrixX<T>& mutable_dPdy() { return dPdy_; }
+
+ private:
+  // This struct stores parameters that remain const after construction.
+  struct ConstParameters {
+    VectorX<T> R;      // Regularization R.
+    VectorX<T> R_inv;  // Inverse of regularization R.
+    VectorX<T> v_hat;  // Constraint velocity bias.
+  };
+  ConstParameters parameters_;
+
+  VectorX<T> vc_;
+  VectorX<T> y_;      // Un-projected impulse y = −R⁻¹⋅(vc−v̂)
+  VectorX<T> gamma_;  // Impulse.
+  MatrixX<T> dPdy_;   // Gradient of the projection γ = P(y) w.r.t. y.
+};
 
 /* This class serves to represent constraints supported by the SapSolver as
 described in [Castro et al., 2021].
@@ -61,6 +126,15 @@ two cliques, see issue #16575.
 template <typename T>
 class SapConstraint {
  public:
+  /* We do not allow copy, move, or assignment generally to avoid slicing.
+    Protected copy construction is enabled for sub-classes to use in their
+    implementation of DoClone(). */
+  //@{
+  SapConstraint& operator=(const SapConstraint&) = delete;
+  SapConstraint(SapConstraint&&) = delete;
+  SapConstraint& operator=(SapConstraint&&) = delete;
+  //@}
+
   /* Constructor for a constraint among DOFs within a single `clique`.
    @param[in] clique
      Index of a clique in the SapContactProblem where this constraint will be
@@ -160,6 +234,85 @@ class SapConstraint {
     return second_clique_jacobian_;
   }
 
+  /* Makes data used by this constraint to perform computations. Different
+   constraints can opt to use `time_step` and a diagonal approximation of the
+   Delassus operator in `delassus_estimation` to pre-compute scale quantities to
+   condition the problem better.
+   N.B. Specific constraints can choose to ignore `time_step` and
+   `delassus_estimation` making a new data that does not depend on either of
+   them. Refer to the documentation for specific constraint types.
+   @pre delassus_estimation.size() equals num_constraint_equations(). */
+  std::unique_ptr<AbstractValue> MakeData(
+      const T& time_step,
+      const Eigen::Ref<const VectorX<T>>& delassus_estimation) const;
+
+  /* Computes constraint data as a function of constraint velocities `vc`.
+   This method can be as simple as a copy of `vc` into `data`, though specific
+   constraints might want to compute commonly occurring terms in the computation
+   of the cost, impulses and Hessian into `data` to be reused.
+   @pre vc.size() equals num_constraint_equations().
+   @pre data does not equal nullptr. */
+  void CalcData(const Eigen::Ref<const VectorX<T>>& vc,
+                AbstractValue* data) const;
+
+  /* Computes the constraint cost ℓ(vc) function of constraint velocities vc.
+   @post The cost ℓ(vc) is a convex function of vc, as required by the SAP
+   formulation. */
+  T CalcCost(const AbstractValue& data) const;
+
+  /* Computes the impulse according to:
+       γ(vc) = −∂ℓ/∂vc.
+   @pre gamma does not equal nullptr. */
+  void CalcImpulse(const AbstractValue& data, EigenPtr<VectorX<T>> gamma) const;
+
+  /* Computes the constraint Hessian as:
+       G(vc) = ∂²ℓ/∂vc² = -∂γ/∂vc.
+   @post The constraint Hessian G(vc) is
+   symmetric positive semi-definite, since ℓ(vc) is a convex function of vc as
+   required by the SAP formulation.
+   @pre G does not equal nullptr. */
+  void CalcCostHessian(const AbstractValue& data, MatrixX<T>* G) const;
+
+  /* Polymorphic deep-copy into a new instance. */
+  std::unique_ptr<SapConstraint<T>> Clone() const { return DoClone(); }
+
+ protected:
+  /* Protected copy construction is enabled for sub-classes to use in their
+   implementation of DoClone(). */
+  SapConstraint(const SapConstraint&) = default;
+
+  // @group NVI implementations. Specific constraints must implement these
+  // methods. Refer to the specific NVI documentation for details.
+  // Proper argument sizes and valid non-null pointers are already guaranteed by
+  // checks in the correspondng NVIs.
+  // TODO(amcastro-tri): Make these pure virtual per #19392.
+  // @{
+  virtual std::unique_ptr<AbstractValue> DoMakeData(
+      const T& time_step,
+      const Eigen::Ref<const VectorX<T>>& delassus_estimation) const;
+  virtual void DoCalcData(const Eigen::Ref<const VectorX<T>>& vc,
+                          AbstractValue* data) const;
+  virtual T DoCalcCost(const AbstractValue& data) const;
+  virtual void DoCalcImpulse(const AbstractValue& data,
+                             EigenPtr<VectorX<T>> gamma) const;
+  virtual void DoCalcCostHessian(const AbstractValue& data,
+                                 MatrixX<T>* G) const;
+
+  /* Clone() implementation. Derived classes must override to provide
+   polymorphic deep-copy into a new instance. */
+  virtual std::unique_ptr<SapConstraint<T>> DoClone() const = 0;
+
+  // @}
+
+  // TODO(amcastro-tri): Remove this group of methods below per #19392.
+  // N.B. To aid migration towards #19392, the default implementations all
+  // abort. This will guarantee that all current (and future) derived classes
+  // have overridden these methods; failure to do so would lead to a program
+  // abort. This allows older implementations in terms of these methods to
+  // coexist with newly migrated constraints that implement the new DoCalcFoo()
+  // APIs. Once the migration is complete, these virtual APIs will be removed
+  // and the DoCalcFoo() APIs will be made pure virtual.
+
   /* Computes the projection γ = P(y) onto the convex set specific to a
    constraint in the norm defined by the diagonal positive matrix R, i.e. the
    norm ‖x‖ = sqrt(xᵀ⋅R⋅x). Refer to [Castro et al., 2021] for details.
@@ -174,7 +327,13 @@ class SapConstraint {
   virtual void Project(const Eigen::Ref<const VectorX<T>>& y,
                        const Eigen::Ref<const VectorX<T>>& R,
                        EigenPtr<VectorX<T>> gamma,
-                       MatrixX<T>* dPdy = nullptr) const = 0;
+                       MatrixX<T>* dPdy = nullptr) const {
+    unused(y);
+    unused(R);
+    unused(gamma);
+    unused(dPdy);
+    DRAKE_UNREACHABLE();
+  }
 
   /* Computes the bias term v̂ used to compute the constraint impulses before
    projection y = −R⁻¹⋅(vc − v̂).
@@ -187,7 +346,11 @@ class SapConstraint {
    Thus far we are assuming the same scalar factor can be used for all
    constraint equations effectively.
    TODO(amcastro-tri): Consider making wi a vector quantity. */
-  virtual VectorX<T> CalcBiasTerm(const T& time_step, const T& wi) const = 0;
+  virtual VectorX<T> CalcBiasTerm(const T& time_step, const T& wi) const {
+    unused(time_step);
+    unused(wi);
+    DRAKE_UNREACHABLE();
+  }
 
   /* Computes the regularization R used to compute the constraint impulses
    before projection as y = −R⁻¹⋅(vc − v̂).
@@ -201,14 +364,11 @@ class SapConstraint {
    constraint equations effectively.
    TODO(amcastro-tri): Consider making wi a vector quantity. */
   virtual VectorX<T> CalcDiagonalRegularization(const T& time_step,
-                                                const T& wi) const = 0;
-
-  /* Derived classes must override to provide polymorphic deep-copy into a new
-   instance. */
-  virtual std::unique_ptr<SapConstraint<T>> Clone() const = 0;
-
- protected:
-  DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(SapConstraint);
+                                                const T& wi) const {
+    unused(time_step);
+    unused(wi);
+    DRAKE_UNREACHABLE();
+  }
 
  private:
   int first_clique_{-1};
--- a/multibody/contact_solvers/sap/sap_constraint_bundle.cc
+++ b/multibody/contact_solvers/sap/sap_constraint_bundle.cc
@@ -3,6 +3,7 @@
 #include <utility>
 
 #include "drake/common/default_scalars.h"
+#include "drake/common/ssize.h"
 #include "drake/multibody/contact_solvers/sap/contact_problem_graph.h"
 
 namespace drake {
@@ -14,7 +15,8 @@ template <typename T>
 SapConstraintBundle<T>::SapConstraintBundle(
     const SapContactProblem<T>* problem, const VectorX<T>& delassus_diagonal) {
   DRAKE_THROW_UNLESS(problem != nullptr);
-  DRAKE_THROW_UNLESS(delassus_diagonal.size() == problem->num_constraints());
+  DRAKE_THROW_UNLESS(delassus_diagonal.size() ==
+                     problem->num_constraint_equations());
 
   // Create vector of constraints but not in the order they were enumerated in
   // the SapProblem, but in the computationally convenient order enumerated in
@@ -22,28 +24,14 @@ SapConstraintBundle<T>::SapConstraintBundle(
   // pair of cliques are "clustered" together.
   constraints_.reserve(problem->num_constraints());
 
-  // Vector of bias velocities and diagonal matrix R.
-  vhat_.resize(problem->num_constraint_equations());
-  R_.resize(problem->num_constraint_equations());
-  int impulse_index_start = 0;
+  // Store constraints in the order specified by the graph, i.e. by clusters.
   for (const ContactProblemGraph::ConstraintCluster& e :
        problem->graph().clusters()) {
     for (int i : e.constraint_index()) {
       const SapConstraint<T>& c = problem->get_constraint(i);
       constraints_.push_back(&c);
-
-      const int ni = c.num_constraint_equations();
-      const T& wi = delassus_diagonal[i];
-
-      vhat_.segment(impulse_index_start, ni) =
-          c.CalcBiasTerm(problem->time_step(), wi);
-      R_.segment(impulse_index_start, ni) =
-          c.CalcDiagonalRegularization(problem->time_step(), wi);
-
-      impulse_index_start += ni;
     }
   }
-  Rinv_ = R_.cwiseInverse();
 
   MakeConstraintBundleJacobian(*problem);
 }
@@ -122,59 +110,87 @@ void SapConstraintBundle<T>::MakeConstraintBundleJacobian(
 }
 
 template <typename T>
-void SapConstraintBundle<T>::CalcUnprojectedImpulses(const VectorX<T>& vc,
-                                                     VectorX<T>* y) const {
-  DRAKE_DEMAND(vc.size() == num_constraint_equations());
-  DRAKE_DEMAND(y != nullptr);
-  DRAKE_DEMAND(y->size() == num_constraint_equations());
-  *y = Rinv_.asDiagonal() * (vhat_ - vc);
-}
-
-template <typename T>
-void SapConstraintBundle<T>::ProjectImpulses(
-    const VectorX<T>& y, VectorX<T>* gamma,
-    std::vector<MatrixX<T>>* dPdy) const {
-  DRAKE_DEMAND(y.size() == num_constraint_equations());
-  DRAKE_DEMAND(gamma != nullptr);
-  DRAKE_DEMAND(gamma->size() == num_constraint_equations());
-  if (dPdy != nullptr) {
-    DRAKE_DEMAND(static_cast<int>(dPdy->size()) == num_constraints());
-  }
+SapConstraintBundleData SapConstraintBundle<T>::MakeData(
+    const T& time_step, const VectorX<T>& delassus_diagonal) const {
+  DRAKE_DEMAND(delassus_diagonal.size() == num_constraint_equations());
+  SapConstraintBundleData data;
+  data.reserve(num_constraints());
   int constraint_start = 0;
   for (int i = 0; i < num_constraints(); ++i) {
     const SapConstraint<T>& c = *constraints_[i];
     const int ni = c.num_constraint_equations();
-    const auto y_i = y.segment(constraint_start, ni);
-    const auto R_i = R().segment(constraint_start, ni);
-    auto gamma_i = gamma->segment(constraint_start, ni);
-    if (dPdy != nullptr) {
-      MatrixX<T>& dPdy_i = (*dPdy)[i];
-      c.Project(y_i, R_i, &gamma_i, &dPdy_i);
-    } else {
-      c.Project(y_i, R_i, &gamma_i);
-    }
+    const auto wi = delassus_diagonal.segment(constraint_start, ni);
+    data.emplace_back(c.MakeData(time_step, wi));
+    constraint_start += ni;
+  }
+  return data;
+}
+
+template <typename T>
+void SapConstraintBundle<T>::CalcData(
+    const VectorX<T>& vc, SapConstraintBundleData* bundle_data) const {
+  DRAKE_DEMAND(bundle_data != nullptr);
+  DRAKE_DEMAND(ssize(*bundle_data) == num_constraints());
+  int constraint_start = 0;
+  for (int i = 0; i < num_constraints(); ++i) {
+    const SapConstraint<T>& c = *constraints_[i];
+    const int ni = c.num_constraint_equations();
+    const auto vc_i = vc.segment(constraint_start, ni);
+    AbstractValue& data = *(*bundle_data)[i];
+    c.CalcData(vc_i, &data);
     constraint_start += ni;
   }
 }
 
 template <typename T>
-void SapConstraintBundle<T>::ProjectImpulsesAndCalcConstraintsHessian(
-    const VectorX<T>& y, VectorX<T>* gamma, std::vector<MatrixX<T>>* G) const {
-  DRAKE_DEMAND(y.size() == num_constraint_equations());
+T SapConstraintBundle<T>::CalcCost(
+    const SapConstraintBundleData& bundle_data) const {
+  DRAKE_DEMAND(ssize(bundle_data) == num_constraints());
+  T cost = 0.0;
+  for (int i = 0; i < num_constraints(); ++i) {
+    const SapConstraint<T>& c = *constraints_[i];
+    const AbstractValue& data = *bundle_data[i];
+    cost += c.CalcCost(data);
+  }
+  return cost;
+}
+
+template <typename T>
+void SapConstraintBundle<T>::CalcImpulses(
+    const SapConstraintBundleData& bundle_data, VectorX<T>* gamma) const {
+  DRAKE_DEMAND(ssize(bundle_data) == num_constraints());
   DRAKE_DEMAND(gamma != nullptr);
   DRAKE_DEMAND(gamma->size() == num_constraint_equations());
-  DRAKE_DEMAND(static_cast<int>(G->size()) == num_constraints());
-  // G = dPdy after call to ProjectImpulses. We add in the R⁻¹ next.
-  ProjectImpulses(y, gamma, G);
+  int constraint_start = 0;
+  for (int i = 0; i < num_constraints(); ++i) {
+    const SapConstraint<T>& c = *constraints_[i];
+    const int ni = c.num_constraint_equations();
+    const AbstractValue& data = *bundle_data[i];
+    auto gamma_i = gamma->segment(constraint_start, ni);
+    c.CalcImpulse(data, &gamma_i);
+    constraint_start += ni;
+  }
+}
+
+template <typename T>
+void SapConstraintBundle<T>::CalcImpulsesAndConstraintsHessian(
+    const SapConstraintBundleData& bundle_data, VectorX<T>* gamma,
+    std::vector<MatrixX<T>>* G) const {
+  DRAKE_DEMAND(ssize(bundle_data) == num_constraints());
+  DRAKE_DEMAND(gamma != nullptr);
+  DRAKE_DEMAND(gamma->size() == num_constraint_equations());
+  DRAKE_DEMAND(ssize(*G) == num_constraints());
 
   // The regularizer Hessian is G = d²ℓ/dvc² = dP/dy⋅R⁻¹.
   int constraint_start = 0;
   for (int i = 0; i < num_constraints(); ++i) {
     const SapConstraint<T>& c = *constraints_[i];
     const int ni = c.num_constraint_equations();
-    const auto Rinv_i = Rinv().segment(constraint_start, ni);
-    const MatrixX<T>& dPdy_i = (*G)[i];
-    (*G)[i] = dPdy_i * Rinv_i.asDiagonal();
+    const AbstractValue& data = *bundle_data[i];
+    auto gamma_i = gamma->segment(constraint_start, ni);
+    auto& Gi = (*G)[i];
+    c.CalcImpulse(data, &gamma_i);
+    c.CalcCostHessian(data, &Gi);
     constraint_start += ni;
   }
 }
--- a/multibody/contact_solvers/sap/sap_constraint_bundle.h
+++ b/multibody/contact_solvers/sap/sap_constraint_bundle.h
@@ -1,7 +1,9 @@
 #pragma once
 
+#include <memory>
 #include <vector>
 
+#include "drake/common/copyable_unique_ptr.h"
 #include "drake/common/drake_copyable.h"
 #include "drake/multibody/contact_solvers/block_sparse_matrix.h"
 #include "drake/multibody/contact_solvers/sap/partial_permutation.h"
@@ -12,6 +14,8 @@ namespace drake {
 namespace multibody {
 namespace contact_solvers {
 namespace internal {
+
+using SapConstraintBundleData = std::vector<std::unique_ptr<AbstractValue>>;
 
 /* Given a contact problem, this class provides a representation for the entire
  "bundle" of constraints in the problem. This class re-arranges constraints
@@ -85,43 +89,38 @@ class SapConstraintBundle {
    ContactProblemGraph::participating_cliques(). */
   const BlockSparseMatrix<T>& J() const { return J_; }
 
-  /* Returns the diagonal regularization matrix R. Of size
-   num_constraint_equations(). */
-  const VectorX<T>& R() const { return R_; }
+  /* Makes data structure to store data needed for bundle computations.
+    `time_step` and `delassus_diagonal` can be used to pre-compute scale
+    quantities to condition the problem better.
+    @param[in] time_step The time step used in the contact problem.
+    @param[in] delassus_diagonal An estimation of the diagonal of the Delassus
+    operator, with one entry per constraint equation. Of size
+    num_constraint_equations().
+    */
+  SapConstraintBundleData MakeData(const T& time_step,
+                                   const VectorX<T>& delassus_diagonal) const;
 
-  /* Returns the diagonal of the inverse of the regularization matrix. Of size
-   num_constraint_equations(). */
-  const VectorX<T>& Rinv() const { return Rinv_; }
+  /* Updates constraint data as a function of constraint velocities `vc`. */
+  void CalcData(const VectorX<T>& vc,
+                SapConstraintBundleData* constraints_data) const;
 
-  /* Returns the bias v̂ for the entire bundle. Of size
-   num_constraint_equations().*/
-  const VectorX<T>& vhat() const { return vhat_; }
+  /* Computes the compound cost of the bundle as a function of the constraints'
+   velocities vector vc. */
+  T CalcCost(const SapConstraintBundleData& constraints_data) const;
 
-  /* Computes unprojected impulses y according to y = −R⁻¹⋅(v−v̂), where R is
-   the regularization matrix, R(), and v̂ is the bias term, vhat().
+  /* Computes the impulse gamma as a function of vc.
    @pre vc.size() equals num_constraint_equations().
-   @pre y != nullptr and y->size() equals num_constraint_equations(). */
-  void CalcUnprojectedImpulses(const VectorX<T>& vc, VectorX<T>* y) const;
+   @pre gamma != nullptr and gamma->size() equals num_constraint_equations(). */
+  void CalcImpulses(const SapConstraintBundleData& bundle_data,
+                    VectorX<T>* gamma) const;
 
-  /* Computes the projection γ = P(y) for all impulses and the gradient
-   dP/dy if dPdy != nullptr. On output dPdy[i] stores the gradient dPᵢ/dyᵢ for
-   the i-th constraint.
-   @pre y.size() equals num_constraint_equations().
-   @pre gamma != nullptr and gamma->size() equals num_constraint_equations().
-   @pre if dPdy != nullptr, then dPdy->size() equals num_constraints(). */
-  void ProjectImpulses(const VectorX<T>& y, VectorX<T>* gamma,
-                       std::vector<MatrixX<T>>* dPdy = nullptr) const;
-
-  /*  SAP's regularizer cost is defined as ℓᵣ = 1/2γᵀ⋅R⋅γ. The Hessian computed
-   with respect to vc (defined as vc = J⋅v, see class's documentation) is a
-   block diagonal matrix G where the i-th block diagonal entry corresponds to
-   Gᵢ = dPᵢ/dyᵢ⋅Rᵢ⁻¹. This method computes γ = P(y) and Hessian matrix G.
-   See Appendix E [Castro et al., 2021] for further details.
-   @pre y.size() equals num_constraint_equations().
+  /* Computes the constraints's Hessian G(vc) = −∂γ/∂vc = ∂²ℓ/∂vc².
+   @pre vc.size() equals num_constraint_equations().
    @pre gamma != nullptr and gamma->size() equals num_constraint_equations().
    @pre G != nullptr and G->size() equals num_constraints(). */
-  void ProjectImpulsesAndCalcConstraintsHessian(
-      const VectorX<T>& y, VectorX<T>* gamma, std::vector<MatrixX<T>>* G) const;
+  void CalcImpulsesAndConstraintsHessian(
+      const SapConstraintBundleData& bundle_data, VectorX<T>* gamma,
+      std::vector<MatrixX<T>>* G) const;
 
  private:
   /* This method builds the BlockSparseMatrix representation of the Jacobian
@@ -130,9 +129,6 @@ class SapConstraintBundle {
   void MakeConstraintBundleJacobian(const SapContactProblem<T>& problem);
 
   BlockSparseMatrix<T> J_;
-  VectorX<T> vhat_;
-  VectorX<T> R_;
-  VectorX<T> Rinv_;
   // Constraint references in the order dictated by the ContactProblemGraph.
   std::vector<const SapConstraint<T>*> constraints_;
 };
--- a/multibody/contact_solvers/sap/sap_friction_cone_constraint.h
+++ b/multibody/contact_solvers/sap/sap_friction_cone_constraint.h
@@ -3,7 +3,6 @@
 #include <memory>
 #include <utility>
 
-#include "drake/common/drake_copyable.h"
 #include "drake/common/eigen_types.h"
 #include "drake/multibody/contact_solvers/sap/sap_constraint.h"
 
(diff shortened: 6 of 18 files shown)


amcastro-tri (commented):
+@joemasterjohn for feature review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVFr_-k5y0ayf33EO0I:bqwin3e)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri and @joemasterjohn)

joemasterjohn (commented):
:lgtm: with just a few nits. +@SeanCurtis-TRI for platform review today if you have time.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVW42OV9MEYIAWtTAsG:b-7oe0ew)* status: 5 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @amcastro-tri and @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 256 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVVecdpEtAtupXFzyYM:-NVVecdpEtAtupXFzyYN:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L256)):*
> ```objc
>        G(vc) = ∂²ℓ/∂vc² = -∂γ/∂vc.
>    Since ℓ(vc) is required to be a convex function of vc, G(vc) must be
>    symmetric semi-positive definite.
> ```

nit: typo



_Suggestion:_
```objc
positive semi-definite.
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 268 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVVfT3Z5vPpbBgU1Gi8:-NVVfT3Z5vPpbBgU1Gi9:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L268)):*
> ```objc
> 
>   // @group NVI implementations. Specific constraints must implement these
>   // methods. Refer to the specific NIV documentation for details.
> ```

nit: typo



_Suggestion:_
```objc
NVI
```
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.h` line 117 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVVhhqL-LvOvOLGQoct:-NVVhhqL-LvOvOLGQocu:b-tmkhzr) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint_bundle.h#L117)):*
> ```objc
>                     VectorX<T>* gamma) const;
> 
>   /* Computes the constraints's Hessian G(vc)=−∂γ/∂vc = ∂²ℓ/∂vc².
> ```

nit: spacing



_Suggestion:_
```objc
G(vc) = −∂γ/∂vc = ∂²ℓ/∂vc²
```
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVVhzVeBRQDgkosLSYV:-NVVhzVeBRQDgkosLSYW:b6t5enx) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L27)):*
> ```cpp
> 
>   // Vector of bias velocities and diagonal matrix R.
>   // int impulse_index_start = 0;
> ```

nit: remove commented code.



_Code quote:_
```cpp
 // int impulse_index_start = 0;
```
___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 295 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVVok0KAZh20G9sBBzC:-NVVok0KAZh20G9sBBzD:bv3pboh) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/test/sap_model_test.cc#L295)):*
> ```cpp
>                      EigenPtr<VectorX<T>> gamma) const final {
>     const VectorX<T>& vc = data.get_value<VectorX<T>>();
>     *gamma = R_.cwiseInverse().asDiagonal() * (v_hat_ - vc);
> ```

nit: might help readability to write this in the form you write gamma other places.



_Suggestion:_
```cpp
    *gamma = -R_.cwiseInverse().asDiagonal() * (vc - v_hat_);
```

SeanCurtis-TRI (commented):
+(release notes: none)  

Breakpoint. A few small nits and some passing thoughts. I'll pick up the tests tomorrow.

> Reviewed 5 of 10 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVW5ttQ6AiSLFWg6_WU:bqg8npv)* status: 14 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWG4qQ8M_d7yQX0jXZ:-NVWG4qQ8M_d7yQX0jX_:b-g0a7ky) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L21)):*
> ```objc
> /* Structure to store data needed for SapConstraint computations. */
> template <typename T>
> struct SapConstraintData {
> ```

nit: While I know this is temporary, it *really* shouldn't be a `struct`. And, given the API you have, it doesn't seem you're hurt in changing it to `class`.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWGcyJA5BfAs6ruV9R:-NVWGcyJA5BfAs6ruV9S:b-uzu9ii) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L36)):*
> ```objc
>     parameters_.R_inv = R.cwiseInverse();
>     parameters_.v_hat = v_hat;
>     vc_.resize(nk);
> ```

BTW: In the name of "this is temporary", I'm ignoring the fact that these parameters are all uninitialized. Post construction, they must still be set (and are set without checking that invariants are maintained, e.g., they maintain the correct size.

I simply wrap my arms around myself, rock myself in my chair, muttering, "It's only temporary, it's only temporary, it's only temporary...."
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWI3aS4KZ9nluzgb3h:-NVWI3aS4KZ9nluzgb3i:blp2pk0) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L226)):*
> ```objc
>   }
> 
>   /* Makes data used by this constraint to perform computations. Different
> ```

BTW I find it odd that the documentation on the *public* API doesn't give instructions to the caller, but to the programmer that implements `DoFoo()`.

From the caller's perspective, it seems sufficient to say,

```c++
  /* Makes data used by this constraint to perform computations.
   N.B. There is no guarantee that the data for any particular constraint type
   depends on either `time_step` or `delassus_estimation`.
   @pre delassus_estimation.size() must equal num_constraint_equations(). */
  std::unique_ptr<AbstractValue> MakeData(
```

After all, it's not like the caller gets to peek inside the returned data (they may not even know the type). But it is worth noting that if the caller tweaks the parameters and doesn't see a difference, then that could very well be correct.

This then allows you, in the documentation of the actual `DoMakeData()` to emphasize that derived class can optionally use the parameters. It can also make the promise that size of `delassus_estimation` is guaranteed to be correct.

I infer that all of these new methods are the "new public api" therefore, cleaning up its documentation so the right words go to the right audience would be good.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 230 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWHW946oEjzXTqsa81:-NVWHW946oEjzXTqsa82:bo6tnsq) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L230)):*
> ```objc
>    Delassus operator in `delassus_estimation` to pre-compute scale quantities to
>    condition the problem better.
>    @pre delassus_estimation.size() must equal num_constraint_equations(). */
> ```

nit: This is a subtle linguistic point. But prerequisites should be described as *facts* and not *obligations*. It's the difference between saying the prerequisite is the existence of a law vs actual behavioral compliance.

Same applies to all of the other new methods. This is correct in the bundle code.



_Suggestion:_
```objc
delassus_estimation.size() equals num_constraint_equations()
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 245 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWJZGu-TqVCRGpQh6_:-NVWJZGu-TqVCRGpQh6a:bvhcczv) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L245)):*
> ```objc
> 
>   /* Returns the constraint cost ℓ(vc) function of constraint velocities vc.
>    The SAP formulation requires ℓ(vc) to be a convex function of vc. */
> ```

BTW We don't use this much, but this would be the ideal use of [doxygen's `@post`](https://www.doxygen.nl/manual/commands.html#cmdpost):

```
@post The cost ℓ(vc) is a convex function of `vc`.
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 251 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWJBpK2qXgKkVxgSgc:-NVWJBpL-RwNrwLPfCCJ:b-8tmtu8) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L251)):*
> ```objc
>        γ(vc) = −∂ℓ/∂vc.
>    @pre gamma must not be nullptr. */
>   void CalcImpulse(const AbstractValue& data, EigenPtr<VectorX<T>> gamma) const;
> ```

nit: Strangely, you have three functions in a row all called `CalcThing`. But their documentation says the first "Updates", the second "Returns", and the third "Computes". You probably want to unify that language.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 255 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKQlD0OCEZ9zOmdN1:-NVWKQlE3StriPqYS6EP:b-5ffsv3) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L255)):*
> ```objc
>   /* Computes the constraint Hessian as:
>        G(vc) = ∂²ℓ/∂vc² = -∂γ/∂vc.
>    Since ℓ(vc) is required to be a convex function of vc, G(vc) must be
> ```

BTW, again, if you like the `@post`, this is another good place for it.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 262 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKjvm8uEPkoLM7S4g:-NVWKjvm8uEPkoLM7S4h:ba21ual) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L262)):*
> ```objc
>   /* Derived classes must override to provide polymorphic deep-copy into a new
>    instance. */
>   virtual std::unique_ptr<SapConstraint<T>> Clone() const = 0;
> ```

nit: Why not NVI?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NVWKoXi9sgA07wh0zI0:bf954y5) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*
> ```objc
> 
>  protected:
>   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(SapConstraint);
> ```

nit: Do you really want *all* semantics existing as protected? I assume you just need copy available so derived classes can implement `Clone()`.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVWLr2f5bSs-KCBWe_J:b-l1v6ui) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*
> ```objc
> 
>   // TODO(amcastro-tri): Remove this group of methods below per #19392.
>   // N.B. We make the default implementations no-op to ease the migration
> ```

nit: Only one of these is a no-op. The other two are land mines. Probably want to explain that. Why do two explode, but one doesn't?



_Code quote:_
```objc
We make the default implementations no-op
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 309 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWL8fR262bhS2RKEsW:-NVWL8fR262bhS2RKEsX:b-p9esm7) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L309)):*
> ```objc
>     unused(gamma);
>     unused(dPdy);
>   }
> ```

BTW As these are going away and have been rendered no-ops, did you consider simply omitting the parameter names? The *only* reason to keep them is if you cared about correlating documentation with parameter names. And it doesn't seem you care much about that as it does nothing and is going to be burninated.



_Suggestion:_
```objc
  virtual void Project(const Eigen::Ref<const VectorX<T>>&,
                       const Eigen::Ref<const VectorX<T>>&,
                       EigenPtr<VectorX<T>>,
                       MatrixX<T>* = nullptr) const {}
```
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.h` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWSfOT66H_6do5EuVY:-NVWSfOT66H_6do5EuVZ:bvia6oy) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.h#L112)):*
> ```objc
> 
>   /* Computes the impulse gamma as a function of vc.
>    @pre vc.size() == num_constraint_equations().
> ```

BTW Sometimes "equals" sometimes "==".

But, unlike `sap_constraint.h`, these `@pre` are statements of fact. :)
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NVWTBMe6Itkpg14u0zn:b9m6g36) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*
> ```cpp
> T SapConstraintBundle<T>::CalcCost(
>     const SapConstraintBundleData& bundle_data) const {
>   DRAKE_DEMAND(static_cast<int>(bundle_data.size()) == num_constraints());
> ```

BTW Here and throughout, you can now use: `ssize(bundle_data)` in place of `static_cast<int>(bundle_data.size())`.
___
*[`multibody/contact_solvers/sap/test/sap_constraint_bundle_test.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWV0HN5peHaOFAiEFm:-NVWV0HN5peHaOFAiEFn:b8a9ykt) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_constraint_bundle_test.cc#L42)):*
> ```cpp
> // properly composes projections.
> // The constraint has a single numeric parameter `param` to define an arbitrary
> // projection function such that as gamma = -param * vc.
> ```

nit: grammar



_Suggestion:_
```cpp
such that gamma
```

SeanCurtis-TRI (commented):
Tests covered.

> Reviewed 1 of 10 files at r1, 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVZjslx8wDSnP8y3rz0:b5ok4ik)* status: 18 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/test/sap_constraint_bundle_test.cc` line 207 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVZfr236K_5dVlQ4OE7:-NVZfr236K_5dVlQ4OE8:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_constraint_bundle_test.cc#L207)):*
> ```cpp
>                               MatrixCompareType::relative));
> 
>   // Extract Hessian computed by automatic differentation of gamma.
> ```

nit: typo



_Suggestion:_
```cpp
differentiation
```
___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 37 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVZgmOm2lZYu7hiai3b:-NVZgmOm2lZYu7hiai3c:bgs24p0) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_model_test.cc#L37)):*
> ```cpp
> // With SAP we can model implicit springs using constraints. For these
> // constraint the projection is the identity, i.e. γ = P(y) = y.
> // For testing purposes, this is a simple constraints that models a spring
> ```

BTW typo



_Suggestion:_
```cpp
this is a simple constraint
```
___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 54 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVZh0ET59Aoo331nDUh:-NVZh0ET59Aoo331nDUi:b7qukrs) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_model_test.cc#L54)):*
> ```cpp
> 
>   // Model a spring attached to `clique`, expected to be a 3D particle.
>   explicit SpringConstraint(int clique, Vector3<T> x0, T k, T tau_d)
> ```

nit: This isn't  necessary.
___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 103 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVZiBg68Arf4Yq6icF-:-NVZiBg68Arf4Yq6icF0:blcjkbe) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_model_test.cc#L103)):*
> ```cpp
>  private:
>   T k_{0.0};      // Stiffness, in N/m.
>   T tau_d_{0.0};  // Dissipation time scale, in N⋅s/m.
> ```

nit Are these units right? If so, how is `(time_step + tau_d_)` valid in `DoMakeData()`?



_Code quote:_
```cpp
 N⋅s/m.
```

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWG4qQ8M_d7yQX0jXZ:-NV_2sf4CRUqFaGfP6jo:b3pyec8) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L21)):*

nit: While I know this is temporary, it *really* shouldn't be a `struct`. And, given the API you have, it doesn't seem you're hurt in changing it to `class`.
</blockquote></details>

good catch! I made it evolve into a class and forgot to change `struct` by `class`.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWGcyJA5BfAs6ruV9R:-NV_3CkQ5b9epnBCmd5n:b-rd4f9a) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L36)):*

BTW: In the name of "this is temporary", I'm ignoring the fact that these parameters are all uninitialized. Post construction, they must still be set (and are set without checking that invariants are maintained, e.g., they maintain the correct size.

I simply wrap my arms around myself, rock myself in my chair, muttering, "It's only temporary, it's only temporary, it's only temporary...."
</blockquote></details>

Why do we care about setting them to a particular value? they'll get overwritten the first time we do `CalcData()`. 
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWI3aS4KZ9nluzgb3h:-NV_3n9r-x82oXiR9vf6:b-obpnjq) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L226)):*

BTW I find it odd that the documentation on the *public* API doesn't give instructions to the caller, but to the programmer that implements `DoFoo()`.

From the caller's perspective, it seems sufficient to say,

```c++
  /* Makes data used by this constraint to perform computations.
   N.B. There is no guarantee that the data for any particular constraint type
   depends on either `time_step` or `delassus_estimation`.
   @pre delassus_estimation.size() must equal num_constraint_equations(). */
  std::unique_ptr<AbstractValue> MakeData(
```

After all, it's not like the caller gets to peek inside the returned data (they may not even know the type). But it is worth noting that if the caller tweaks the parameters and doesn't see a difference, then that could very well be correct.

This then allows you, in the documentation of the actual `DoMakeData()` to emphasize that derived class can optionally use the parameters. It can also make the promise that size of `delassus_estimation` is guaranteed to be correct.

I infer that all of these new methods are the "new public api" therefore, cleaning up its documentation so the right words go to the right audience would be good.
</blockquote></details>

This will always be intended to authors of a constraint, who will be both user and programmer (notice all this is `internall::`, so not for a typical Drake user).

I can be convinced, but I personally, being myself both programmer and user, find it a lot more useful to find all docs in a single place. Then for the `DoFoo()` implementations I can simply refer to the NVIs docs.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 245 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWJZGu-TqVCRGpQh6_:-NV_4pZoAkm4NWxtUbLA:bfzb3i1) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L245)):*

BTW We don't use this much, but this would be the ideal use of [doxygen's `@post`](https://www.doxygen.nl/manual/commands.html#cmdpost):

```
@post The cost ℓ(vc) is a convex function of `vc`.
```
</blockquote></details>

nice! I have never seen that before. 
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 262 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKjvm8uEPkoLM7S4g:-NV_5V0c45v4PEVN1AXm:b-h6zms3) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L262)):*

nit: Why not NVI?
</blockquote></details>

I wanted to focus on the `DoCalcFoo()` APIs in this PR. Also, why would I make the `Clone()` method and NVI? just for naming consistency and grouping all virtual methods in one place?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NV_5wGkBtEsEEZDLkbm:blzq4xp) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*

nit: Do you really want *all* semantics existing as protected? I assume you just need copy available so derived classes can implement `Clone()`.
</blockquote></details>

Don't I need move ctor's to be protected for the derived classes to have move semantics?
Not sure I understood your comment though.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NV_7vtsAZ5Iy7iZ6c26:bjag60d) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

nit: Only one of these is a no-op. The other two are land mines. Probably want to explain that. Why do two explode, but one doesn't?
</blockquote></details>

the three of them are no-ops? not sure I followed.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 309 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWL8fR262bhS2RKEsW:-NV_84gU9LVysPz7JknF:b50slep) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L309)):*

BTW As these are going away and have been rendered no-ops, did you consider simply omitting the parameter names? The *only* reason to keep them is if you cared about correlating documentation with parameter names. And it doesn't seem you care much about that as it does nothing and is going to be burninated.
</blockquote></details>

yeah, the reason was to keep the docs in place. I figured I'd burn the whole thing together at once, rather than in pieces. That is my preference at least.
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.h` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWSfOT66H_6do5EuVY:-NV_8L6u6XDJfIXPJCEa:b4c9lyr) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.h#L112)):*

BTW Sometimes "equals" sometimes "==".

But, unlike `sap_constraint.h`, these `@pre` are statements of fact. :)
</blockquote></details>

:thumbsup: 
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NV_9-w01vCSQyXOCyma:b-pshzrf) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*

BTW Here and throughout, you can now use: `ssize(bundle_data)` in place of `static_cast<int>(bundle_data.size())`.
</blockquote></details>

agh... The puget I'm on uses Focal :-/, and Vince is using the one where I updated to Jammy this week. 
Next pass I guess.
___
*[`multibody/contact_solvers/sap/test/sap_model_test.cc` line 103 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVZiBg68Arf4Yq6icF-:-NV_9lnk7b4j3jYndZUW:b-e05hjm) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/test/sap_model_test.cc#L103)):*

nit Are these units right? If so, how is `(time_step + tau_d_)` valid in `DoMakeData()`?
</blockquote></details>

godo catch!

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 87 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NV_Tl311In2DDp-KB2k:-NV_Tl311In2DDp-KB2l:b-rhahr) ([raw file](https://github.com/RobotLocomotion/drake/blob/5383a7b233056ffc5b122da1a17689923b10866d/multibody/contact_solvers/sap/sap_constraint.cc#L87)):*
> ```cpp
>   const VectorX<T> v_hat = CalcBiasTerm(time_step, wi);
>   const VectorX<T> R = CalcDiagonalRegularization(time_step, wi);
>   SapConstraintData<T> data(std::move(R), std::move(v_hat));
> ```

nit `move`ing from a const doesn't accomplish anything.

You could make them non-const, but really I think just lose the moves.  This is one-time setup code, I think.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWGcyJA5BfAs6ruV9R:-NV_kSei8e6WSZHJ2EW4:bo1uutm) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L36)):*

Why do we care about setting them to a particular value? they'll get overwritten the first time we do `CalcData()`. 
</blockquote></details>

It's discipline. You've got instances floating around in indeterminate state. I could take the result of `MakeData()` and query these uninitialized quantities and nothing in the design would hinder me or warn me from doing that. So, you've got instances of the data type whose semantics differ based on what has been done to them and no simple way to detect that from the outside.

Again, I'm not marking this as a defect (as it is a) a stop gap and b) internal).
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWI3aS4KZ9nluzgb3h:-NV_lBvIDl58q58bPFgE:bdq03mb) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L226)):*
To reiterate one of my points. Anyone implementing `DoMakeData()` should be able to rely on the fact that it is invoked that the parameters have been validated. 

1. That promise isn't documented anywhere.
2. It doesn't make sense to document on `MakeData` because `MakeData` is the one declaring the prerequisite.

> more useful to find all docs in a single place.

This works best when *all* of the docs *can* be kept in a single place. But when you have contradictory docs (this must have the right size, the size is guaranteed), then that model doesn't quite work.

Remember, you're not writing this documentation for *you*. You're writing it for a developer you've never met.

That said, I tagged it BTW, and I'll mark myself satisfied, deferring to author preference.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 230 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWHW946oEjzXTqsa81:-NV_qMtJ41fKEDReL_cx:b-aezns6) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L230)):*

nit: This is a subtle linguistic point. But prerequisites should be described as *facts* and not *obligations*. It's the difference between saying the prerequisite is the existence of a law vs actual behavioral compliance.

Same applies to all of the other new methods. This is correct in the bundle code.
</blockquote></details>

You corrected it here, but not on others: `CalcData`, `CalcImpulse`, `CalcCostHessian`, etc.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 262 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKjvm8uEPkoLM7S4g:-NV_m1I69SFaeLYPMCPf:blfwpv9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L262)):*

I wanted to focus on the `DoCalcFoo()` APIs in this PR. Also, why would I make the `Clone()` method and NVI? just for naming consistency and grouping all virtual methods in one place?
</blockquote></details>

Here were my arguments in favor of NVI:
  - Drake typically prefers NVI.
  -  It's always better to have it up front than to try and add it later.
  -  It also makes your class more consistent -- all of the methods any derived class has to implement are of the `DoFoo()` variety. 
  - I can find  three other classes that declare a virtual `Clone()`, but nine classes that declare a virtual `DoClone()`.

Argument against:
  - The [drake styleguide](https://drake.mit.edu/styleguide/cppguide.html#Copyable_Movable_Types) explicitly says, "If your base class needs to be copyable, provide a public virtual Clone() method." So, as legal arguments go, that wins.

So, for all those reasons, I'm marking myself as satisfied. This design is, de jure, not a defect.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NV_p-WSBv3XYPcTqdda:bqpifcc) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*

Don't I need move ctor's to be protected for the derived classes to have move semantics?
Not sure I understood your comment though.
</blockquote></details>

Move semantics in derived classes. That *sounds* odd to me. Is that an explicit intent? Because that implies that you're not accessing your constraints via pointers/references on the base class, but by accessing them via their derived type explicitly. Will that be happening a lot?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NV_pkX7BXGNMdJbCvpw:b7sgqhj) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

the three of them are no-ops? not sure I followed.
</blockquote></details>

Two call `DRAKE_UNREACHABLE`. Those are the land mines. If they get called, things abort.
___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 87 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NV_Tl311In2DDp-KB2k:-NV_k4FeDOSmHwmSOozx:b-2g8npe) ([raw file](https://github.com/RobotLocomotion/drake/blob/5383a7b233056ffc5b122da1a17689923b10866d/multibody/contact_solvers/sap/sap_constraint.cc#L87)):*

nit `move`ing from a const doesn't accomplish anything.

You could make them non-const, but really I think just lose the moves.  This is one-time setup code, I think.
</blockquote></details>

D'oh! I glossed over the `const`.  
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NV_qfLADwPZUFCGwGDa:b-vqkjuf) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*

agh... The puget I'm on uses Focal :-/, and Vince is using the one where I updated to Jammy this week. 
Next pass I guess.
</blockquote></details>

No worries.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWGcyJA5BfAs6ruV9R:-NV_tEQK0l8Lz3cIFEJV:bm2j06a) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L36)):*

It's discipline. You've got instances floating around in indeterminate state. I could take the result of `MakeData()` and query these uninitialized quantities and nothing in the design would hinder me or warn me from doing that. So, you've got instances of the data type whose semantics differ based on what has been done to them and no simple way to detect that from the outside.

Again, I'm not marking this as a defect (as it is a) a stop gap and b) internal).
</blockquote></details>

BTW Of course if the next PR will delete this code anyway it doesn't matter (I haven't looked), but even for internal code I think it's worth _documenting_ that the constructor creates garbage memory on purpose for performance (instead of paying for the cost of `setConstant(nk, NaN)` for safety).  Even if the code is correct, it's also important to be clear for the next person to comes along to read it.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NV_usbD8waAMLxQ8meu:b-zboa7w) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*

No worries.
</blockquote></details>

BTW Drake's [ssize](https://drake.mit.edu/doxygen_cxx/namespacedrake.html#a65f207e2f4d5cda4d2bc9da7b8a54256) works on all platforms.  There is nothing related to Focal vs Jammy in play here. Sean was referring to `drake::ssize` not `std::ssize`.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWGcyJA5BfAs6ruV9R:-NVduzAi-aggkfXO4JGe:b-k5ydui) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L36)):*

BTW Of course if the next PR will delete this code anyway it doesn't matter (I haven't looked), but even for internal code I think it's worth _documenting_ that the constructor creates garbage memory on purpose for performance (instead of paying for the cost of `setConstant(nk, NaN)` for safety).  Even if the code is correct, it's also important to be clear for the next person to comes along to read it.
</blockquote></details>

Thanks. As Jeremy pointed out, there is no need to pay for the cost of initializing data here (even if small). I documented this. 
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWI3aS4KZ9nluzgb3h:-NVdwb9v5hBM7uTrZjZi:bs48ox9) ([raw file](https://github.com/RobotLocomotion/drake/blob/79f8f4f890f91ad9cd4dfbbf1a3b6be7089adf29/multibody/contact_solvers/sap/sap_constraint.h#L226)):*

To reiterate one of my points. Anyone implementing `DoMakeData()` should be able to rely on the fact that it is invoked that the parameters have been validated. 

1. That promise isn't documented anywhere.
2. It doesn't make sense to document on `MakeData` because `MakeData` is the one declaring the prerequisite.

> more useful to find all docs in a single place.

This works best when *all* of the docs *can* be kept in a single place. But when you have contradictory docs (this must have the right size, the size is guaranteed), then that model doesn't quite work.

Remember, you're not writing this documentation for *you*. You're writing it for a developer you've never met.

That said, I tagged it BTW, and I'll mark myself satisfied, deferring to author preference.
</blockquote></details>

I settled on an intermediate solution. I updated both docs here and also for the `DoFoo()` implementations. I believe your suggestions are covered.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 262 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKjvm8uEPkoLM7S4g:-NVdzLmu22tqrfjAPCjJ:belohig) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L262)):*

Here were my arguments in favor of NVI:
  - Drake typically prefers NVI.
  -  It's always better to have it up front than to try and add it later.
  -  It also makes your class more consistent -- all of the methods any derived class has to implement are of the `DoFoo()` variety. 
  - I can find  three other classes that declare a virtual `Clone()`, but nine classes that declare a virtual `DoClone()`.

Argument against:
  - The [drake styleguide](https://drake.mit.edu/styleguide/cppguide.html#Copyable_Movable_Types) explicitly says, "If your base class needs to be copyable, provide a public virtual Clone() method." So, as legal arguments go, that wins.

So, for all those reasons, I'm marking myself as satisfied. This design is, de jure, not a defect.
</blockquote></details>

I actually like the NVI better for consistency. After I'm done, all constraints will have to implement NVIs. Nice to have them in a single place.
I therefore made the change to make Clone() an NVI as well.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NVdzveq5HltCCTl3ZDt:b-pynclm) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*

Move semantics in derived classes. That *sounds* odd to me. Is that an explicit intent? Because that implies that you're not accessing your constraints via pointers/references on the base class, but by accessing them via their derived type explicitly. Will that be happening a lot?
</blockquote></details>

I need to play with this a lil bit. I thought we wanted move semantics for at least things like `std::make_unique`?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVe-JaY8cRb_FibCJaZ:bs60kv1) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

Two call `DRAKE_UNREACHABLE`. Those are the land mines. If they get called, things abort.
</blockquote></details>

oh, I see. When I was prototyping I only added the unreachables wherever the compiler barked about a non-return. It does make sense all three have it.
___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 87 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NV_Tl311In2DDp-KB2k:-NVe-gzL-BI3cIx_Eyb5:b1xf40s) ([raw file](https://github.com/RobotLocomotion/drake/blob/5383a7b233056ffc5b122da1a17689923b10866d/multibody/contact_solvers/sap/sap_constraint.cc#L87)):*

D'oh! I glossed over the `const`.  
</blockquote></details>

nice catch! thanks!
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NVe1Es0E0ZwLCiI6NMc:b-1p5qoo) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*

BTW Drake's [ssize](https://drake.mit.edu/doxygen_cxx/namespacedrake.html#a65f207e2f4d5cda4d2bc9da7b8a54256) works on all platforms.  There is nothing related to Focal vs Jammy in play here. Sean was referring to `drake::ssize` not `std::ssize`.
</blockquote></details>

ah... thanks! Done

SeanCurtis-TRI (commented):
A couple of issues left -- largely based on how this PR fits into a larger ecosystem -- I don't have insight into that bigger plan, so we're playing twenty questions. :)


One last thought (that I'm not blocking on):

You can make a more aggressive use of inlined functions if you want.  For example:

```C++
class Foo {
 public:
  std::unique_ptr<Foo> Clone() const {
    return DoClone();
  }
  ...
};

class FooChild : public Foo {
 private:
  std::unique_ptr<Foo> DoClone() const override {
    return std::make_unique<FooChild>(*this);
  }
};
```

You're not obliged to, but these examples of in-lining one-line functions are not defective and they do have the advantage of locality. 

> Reviewed 10 of 13 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVeTxbJ2ihoImrx61Ih:b-h0ru0r)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NVeOYCW0W6AeeJ3ANFG:b-yhu8ui) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*
`std::make_unique` *constructs* the object in place. It doesn't construct it and then move it. So, all `make_unique` needs is a constructor (available and consistent with the parameters provided). If you have a *move constructor* then you *can* do `make_unique<Foo>(move(my_foo))`; but that's not a property of `make_unique`. It's a property of `Foo`. (In fact, that same thing will work if `Foo` lacks a move constructor but has a copy constructor.)

The [styleguide link](https://drake.mit.edu/styleguide/cppguide.html#Copyable_Movable_Types) I provided above actually goes on to provide some specific guidance in this regard:

> If your base class needs to be copyable, provide a public virtual Clone() method, and a protected copy constructor that derived classes can use to implement it.

Related to, but not directly specified by that guidance, the standard practice is to delete everything but the copy constructor and then make that protected so derived classes can use it in their own implementation of `DoClone()`.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVePmGP2_1BDbxtVnxY:bz52y54) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*
So, now, the claim that the implementations are "no-op" is the lie.

But that leads us to a bigger issue. If they all blow up, it's not clear how that "eases the migration". Generally, the comment/code don't seem to present a coherent problem-solution.

If I had to make an argument of what this code *actually* achieves, it would be:

> The default implementations all abort. This will guarantee that all current (and future) derived classes have overridden these methods; failure to do so would lead to a program abort.

However, what's the point of that? If things aren't blowing up *now* then it seems that the condition "every derived class has implemented" is already satisfied. And if you want to make sure that happy state of affairs continues into the future (during the transitional period), the *best* thing to do would be to make the methods purely virtual so the *compiler* complains about it.

The possible exception to this logic is that there exists implementations that *don't* implement these methods, but they don't get exercised. Swapping to purely virtual would make those derived classes angry.

Perhaps I'm missing something (either in what effect is trying to be achieved, or in what further effects these bombs have). That's certainly been known to happen. But, right now, this is defective in that the documentation and implementation don't make sense together.
___
*[`multibody/contact_solvers/sap/sap_constraint_bundle.cc` line 147 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWTBMe6Itkpg14u0zm:-NVeRDY2Ax0s7vQzvuGp:b-p6r544) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint_bundle.cc#L147)):*

ah... thanks! Done
</blockquote></details>

That's so pretty.

amcastro-tri (commented):
> largely based on how this PR fits into a larger ecosystem

The full ecosystem is implemented in #1939, though not sure it'd help much, at this point I think you already got a good picture.

> You can make a more aggressive use of inlined functions if you want.

I like it! 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVekIVc0mUW08oQPoDK:b-ur08rq)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NVensQm0yY0spAEnM-t:b-8pe7wn) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*

`std::make_unique` *constructs* the object in place. It doesn't construct it and then move it. So, all `make_unique` needs is a constructor (available and consistent with the parameters provided). If you have a *move constructor* then you *can* do `make_unique<Foo>(move(my_foo))`; but that's not a property of `make_unique`. It's a property of `Foo`. (In fact, that same thing will work if `Foo` lacks a move constructor but has a copy constructor.)

The [styleguide link](https://drake.mit.edu/styleguide/cppguide.html#Copyable_Movable_Types) I provided above actually goes on to provide some specific guidance in this regard:

> If your base class needs to be copyable, provide a public virtual Clone() method, and a protected copy constructor that derived classes can use to implement it.

Related to, but not directly specified by that guidance, the standard practice is to delete everything but the copy constructor and then make that protected so derived classes can use it in their own implementation of `DoClone()`.
</blockquote></details>

Thanks for the info. I removed the move semantics and allow the copy semantics.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVenqCY-51qLqucjQ7n:b-xb4ppn) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

So, now, the claim that the implementations are "no-op" is the lie.

But that leads us to a bigger issue. If they all blow up, it's not clear how that "eases the migration". Generally, the comment/code don't seem to present a coherent problem-solution.

If I had to make an argument of what this code *actually* achieves, it would be:

> The default implementations all abort. This will guarantee that all current (and future) derived classes have overridden these methods; failure to do so would lead to a program abort.

However, what's the point of that? If things aren't blowing up *now* then it seems that the condition "every derived class has implemented" is already satisfied. And if you want to make sure that happy state of affairs continues into the future (during the transitional period), the *best* thing to do would be to make the methods purely virtual so the *compiler* complains about it.

The possible exception to this logic is that there exists implementations that *don't* implement these methods, but they don't get exercised. Swapping to purely virtual would make those derived classes angry.

Perhaps I'm missing something (either in what effect is trying to be achieved, or in what further effects these bombs have). That's certainly been known to happen. But, right now, this is defective in that the documentation and implementation don't make sense together.
</blockquote></details>

You got it right. I updated the docs.

SeanCurtis-TRI (commented):
Getting close. Two steps forward, one step sideways.

> Reviewed 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVf8z-Y0cgM8tlzX2XW:b-ksgt9g)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 265 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWKoXi9sgA07wh0zI-:-NVf-y421BUBWcd0GLwF:b-v9fkop) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L265)):*

Thanks for the info. I removed the move semantics and allow the copy semantics.
</blockquote></details>

A common practice is to leave a stanza in the public API (as we're accustomed to looking there). Something like this (stolen from another example):

```c++
  /* \name Does not allow copy, move, or assignment generally.

   Protected copy construction is enabled for sub-classes to use in their
   implementation of DoClone(). */
  //@{
  SapConstraint& operator=(const SapConstraint&) = delete;
  SapConstraint(SapConstraint&&) = delete;
  SapConstraint& operator=(SapConstraint&&) = delete;
  //@}
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVf0NBLAx-OX6HS6NAk:b5srthf) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

You got it right. I updated the docs.
</blockquote></details>

Only one of my points got across.

How about my second point? The second point is, for this purpose, there's a better way to do it. Simply remove the implementations completely and make them purely virtual. Why wait for runtime to find out something that the compiler can tell you about?
___
*[`multibody/contact_solvers/sap/sap_friction_cone_constraint.h` line 67 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVf6OTvESDPy_tluGft:-NVf6OTvESDPy_tluGfu:b-5r1x0q) ([raw file](https://github.com/RobotLocomotion/drake/blob/12760d5b8f5ecf9d110bfe22cb0c7ca9adae8fc4/multibody/contact_solvers/sap/sap_friction_cone_constraint.h#L67)):*
> ```objc
> class SapFrictionConeConstraint final : public SapConstraint<T> {
>  public:
>   // Copyable but not moveable.
> ```

I imagine right now you are contemplating making a voodoo doll of me. But...

The previous state of this code declared copy and move semantics. That was erroneous.

The current state in R6 is technically valid, but semantically mucky. To clean it up:

  1. Apply the same "not generlally movable or copyable" boilerplate to this class as well (see `sap_constraint.h`).
  2. The copy constructor should be protected.
  3. The spelling of `DoClone()` can no longer use `make_unique` (it doesn't have access). Instead, it has to use:

```c++
return std::unique_ptr<SapFrictionConeConstraint<T>>(
        new SapFrictionConeConstraint<T>(*this));
```

Here's the reasoning. This class is part of a polymorphic family. The basic assumption is that after instantiating one, you only ever deal with them via pointers/references to the base class. Copy (and move) constructors on polymorphic families can easily lead to slicing. One of the reasons we eliminate all of the public copy/move is to completely eliminate the possibility of slicing. We should continue that through to here. It's part of the practice of presenting a polymorphic family.

N.B. I declared this to be "technically valid" because this class is declared `final`. Therefore, nothing can derive from this, therefore slicing, from this level of the class hierarchy, is no longer possible. So, at this point, it comes down to convention and expectation. We're better with a consistent story about, "If you want a copy of this polymorphic class, clone it.". Plus, it's more future safe in case someone says, "I'm removing `final` from this class, because I need to make two different implementation."

Sorry to keep thrashing your code... The same applies to the other derived classes.
___
*[`multibody/contact_solvers/sap/sap_friction_cone_constraint.h` line 161 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVf0zSs6Cildas1tZZO:-NVf0zSs6Cildas1tZZP:b2th8rj) ([raw file](https://github.com/RobotLocomotion/drake/blob/12760d5b8f5ecf9d110bfe22cb0c7ca9adae8fc4/multibody/contact_solvers/sap/sap_friction_cone_constraint.h#L161)):*
> ```objc
> 
>  private:
>   std::unique_ptr<SapConstraint<T>> DoClone() const override {
> ```

nit: swap `override` for `final`. The class is declared `final`, so let's keep a consistent story.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVfJQsk3kEMuRBidqAP:b2xkdws) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

Only one of my points got across.

How about my second point? The second point is, for this purpose, there's a better way to do it. Simply remove the implementations completely and make them purely virtual. Why wait for runtime to find out something that the compiler can tell you about?
</blockquote></details>

I cannot make them pure virtual, as we migrate towards #1939, some constraints will implement the older family of methods and migrated constraints will implement the new family. All this, within a week of landing this PR. So I believe there's no reason on beautifying this further. As soon as this land, I will push a separate PR for each constraint, at the same time to be reviewed in parallel.
___
*[`multibody/contact_solvers/sap/sap_friction_cone_constraint.h` line 67 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVf6OTvESDPy_tluGft:-NVfPuOV6CBN-M5ZxShT:b-4ewjbx) ([raw file](https://github.com/RobotLocomotion/drake/blob/12760d5b8f5ecf9d110bfe22cb0c7ca9adae8fc4/multibody/contact_solvers/sap/sap_friction_cone_constraint.h#L67)):*

I imagine right now you are contemplating making a voodoo doll of me. But...

The previous state of this code declared copy and move semantics. That was erroneous.

The current state in R6 is technically valid, but semantically mucky. To clean it up:

  1. Apply the same "not generlally movable or copyable" boilerplate to this class as well (see `sap_constraint.h`).
  2. The copy constructor should be protected.
  3. The spelling of `DoClone()` can no longer use `make_unique` (it doesn't have access). Instead, it has to use:

```c++
return std::unique_ptr<SapFrictionConeConstraint<T>>(
        new SapFrictionConeConstraint<T>(*this));
```

Here's the reasoning. This class is part of a polymorphic family. The basic assumption is that after instantiating one, you only ever deal with them via pointers/references to the base class. Copy (and move) constructors on polymorphic families can easily lead to slicing. One of the reasons we eliminate all of the public copy/move is to completely eliminate the possibility of slicing. We should continue that through to here. It's part of the practice of presenting a polymorphic family.

N.B. I declared this to be "technically valid" because this class is declared `final`. Therefore, nothing can derive from this, therefore slicing, from this level of the class hierarchy, is no longer possible. So, at this point, it comes down to convention and expectation. We're better with a consistent story about, "If you want a copy of this polymorphic class, clone it.". Plus, it's more future safe in case someone says, "I'm removing `final` from this class, because I need to make two different implementation."

Sorry to keep thrashing your code... The same applies to the other derived classes.
</blockquote></details>

no... I just regret not packing one this morning from the stash of "Sean's voodoo dolls" I accumulated over the years :laughing:

amcastro-tri (commented):
Are we ready to merge @SeanCurtis-TRI? this is starting to get on the critical path to my progress

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVim8TTChiFbOBCyfR9:bpr2z1y)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM:

> Reviewed 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-:-NVj42oCDdqc29V8CQKQ:bmxilld)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),joemasterjohn

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVWLr2f5bSs-KCBWe_I:-NVj3LDfBIca4wsfpymX:bwgac64) ([raw file](https://github.com/RobotLocomotion/drake/blob/bd6b99c6290bdacd4d6d4b77446301089e158559/multibody/contact_solvers/sap/sap_constraint.h#L284)):*

I cannot make them pure virtual, as we migrate towards #1939, some constraints will implement the older family of methods and migrated constraints will implement the new family. All this, within a week of landing this PR. So I believe there's no reason on beautifying this further. As soon as this land, I will push a separate PR for each constraint, at the same time to be reviewed in parallel.
</blockquote></details>

Ah....I see. So, the *true* purpose is that as individual constraints move from old to new APIs, they will *remove* their implementations of these functions. So, these bombs are proof that none of the *constraint-using* infrastructure is attempting to evaluate the deprecated interface. That's a different purpose.

I'm not going to press you to change the documentation here to reflect that. As you say, it's a week before it goes away correctly and that is well characterized as "facilitating transition".
___
*[`multibody/contact_solvers/sap/sap_friction_cone_constraint.h` line 67 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19402#-NVf6OTvESDPy_tluGft:-NVj3DynE8HpInD_wt9T:bxq8u5h) ([raw file](https://github.com/RobotLocomotion/drake/blob/12760d5b8f5ecf9d110bfe22cb0c7ca9adae8fc4/multibody/contact_solvers/sap/sap_friction_cone_constraint.h#L67)):*

no... I just regret not packing one this morning from the stash of "Sean's voodoo dolls" I accumulated over the years :laughing: 
</blockquote></details>

Well, at least you have a supply. Best to keep one or two on hand at all times.

amcastro-tri:
Thank you for the review @SeanCurtis-TRI!

PR #19405, opened 2023-05-14
Implements ExpressionCost

### LINKED ISSUE #17897: Add support for generic (nonlinear) ExpressionCost

We currently offer a generic `MathematicalProgram::AddConstraint(Expression, lb, ub)` function that's useful for prototyping:

https://github.com/RobotLocomotion/drake/blob/7233a94a7eb96306422f75cd2f12d426a57a9160/solvers/mathematical_program.h#L1280-L1287

The user can pass any symbolic expression -- if it's of a functional form that we can easily recognize and handle efficiently, we parse it into that form (e.g, a `Binding<LinearConstraint>`).  If we can't parse it, we fall back to a `Binding<ExpressionConstraint>` that uses `Expression::Evaluate` and `symbolic::Jacobian` to evaluate the constraint.

We also have an analogous `MathematicalProgram::AddCost(Expression)` function:

https://github.com/RobotLocomotion/drake/blob/7233a94a7eb96306422f75cd2f12d426a57a9160/solvers/mathematical_program.h#L1165-L1171

It has the same flavor, but has a _much stronger_ pre-condition -- only linear or quadratic costs are supported.  If the user passes a non-linear cost, they get an exception about "Expression ... is not polynomial ...".

This is inconsistent and confusing.

This `AddCost` overload should _only_ throw when the cost is trivial (i.e., a constant like "5").

This probably means adding an `ExpressionCost` class as the most generic fallback.

RussTedrake:
Most importantly, now AddCost(Expression) will succeed, even if it is passed in an Expression that is non-Polynomial.

Resolves #17897.

+@jwnimmer-tri for feature review, please?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405)

### DIFF
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -822,9 +822,7 @@ void BindMathematicalProgram(py::module m) {
       .def("AddCost",
           static_cast<Binding<Cost> (MathematicalProgram::*)(
               const Expression&)>(&MathematicalProgram::AddCost),
-          // N.B. There is no corresponding C++ method, so the docstring here
-          // is a literal, not a reference to documentation_pybind.h
-          "Adds a cost expression.")
+          py::arg("e"), doc.MathematicalProgram.AddCost.doc_1args_e)
       .def(
           "AddCost",
           [](MathematicalProgram* self, const std::shared_ptr<Cost>& obj,
--- a/solvers/BUILD.bazel
+++ b/solvers/BUILD.bazel
@@ -203,9 +203,11 @@ drake_cc_library(
     srcs = ["cost.cc"],
     hdrs = ["cost.h"],
     interface_deps = [
+        ":decision_variable",
         ":evaluator_base",
     ],
     deps = [
+        ":constraint",
         "//math:gradient",
     ],
 )
--- a/solvers/cost.cc
+++ b/solvers/cost.cc
@@ -4,6 +4,7 @@
 
 #include "drake/math/autodiff_gradient.h"
 #include "drake/math/differentiable_norm.h"
+#include "drake/solvers/constraint.h"
 
 using Eigen::MatrixXd;
 using Eigen::VectorXd;
@@ -304,6 +305,45 @@ void PerspectiveQuadraticCost::DoEval(
 std::ostream& PerspectiveQuadraticCost::DoDisplay(
     std::ostream& os, const VectorX<symbolic::Variable>& vars) const {
   return DisplayCost(*this, os, "PerspectiveQuadraticCost", vars);
+}
+
+ExpressionCost::ExpressionCost(const symbolic::Expression& e)
+    : Cost(e.GetVariables().size()),
+      /* We reuse the Constraint evaluator's implementation. */
+      evaluator_(std::make_unique<ExpressionConstraint>(
+          Vector1<symbolic::Expression>{e},
+          /* The ub, lb are unused but still required. */
+          Vector1d(0.0), Vector1d(0.0))) {}
+
+const VectorXDecisionVariable& ExpressionCost::vars() const {
+  return dynamic_cast<const ExpressionConstraint&>(*evaluator_).vars();
+}
+
+const symbolic::Expression& ExpressionCost::expression() const {
+  return dynamic_cast<const ExpressionConstraint&>(*evaluator_)
+      .expressions()
+      .coeffRef(0);
+}
+
+void ExpressionCost::DoEval(const Eigen::Ref<const Eigen::VectorXd>& x,
+                            Eigen::VectorXd* y) const {
+  evaluator_->Eval(x, y);
+}
+
+void ExpressionCost::DoEval(const Eigen::Ref<const AutoDiffVecXd>& x,
+                            AutoDiffVecXd* y) const {
+  evaluator_->Eval(x, y);
+}
+
+void ExpressionCost::DoEval(
+    const Eigen::Ref<const VectorX<symbolic::Variable>>& x,
+    VectorX<symbolic::Expression>* y) const {
+  evaluator_->Eval(x, y);
+}
+
+std::ostream& ExpressionCost::DoDisplay(
+    std::ostream& os, const VectorX<symbolic::Variable>& vars) const {
+  return DisplayCost(*this, os, "ExpressionCost", vars);
 }
 
 }  // namespace solvers
--- a/solvers/cost.h
+++ b/solvers/cost.h
@@ -10,6 +10,7 @@
 
 #include <Eigen/SparseCore>
 
+#include "drake/solvers/decision_variable.h"
 #include "drake/solvers/evaluator_base.h"
 
 namespace drake {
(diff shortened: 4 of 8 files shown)


jwnimmer-tri (commented):
See https://github.com/RussTedrake/drake/compare/expression_cost...jwnimmer-tri:drake:expression_cost for a suggested change.  (No need to copy-paste the implementation.)

> Reviewed 2 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-:-NVUgLtWAFIjAa5hnVyQ:bm0dpw0)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVUfUi5BRZP8g2CMhXR:-NVUfUi63zYfNTi9F-0b:b-ytnrui) (no related file):*
The docs for this function on `mathematical_program.h` are out of date now:

```c++
  /**
   * Adds a cost in the symbolic form.
   * Note that the constant part of the cost is ignored. So if you set
   * `e = x + 2`, then only the cost on `x` is added, the constant term 2 is
   * ignored.
   * @param e The linear or quadratic expression of the cost.
   * @pre `e` is linear or `e` is quadratic. Otherwise throws a runtime error.
   * @return The newly created cost, together with the bound variables.
   *
   * @exclude_from_pydrake_mkdoc{Not bound in pydrake.}
   */
  Binding<Cost> AddCost(const symbolic::Expression& e);
```

Both in terms of the preconditions, and the ignoring of the constant factor.

When resolving it, having the constant factor be sometimes-ignored and sometimes-obeyed seems too confusing on first impression, so we will need to figure out how to do it all one way or the other, and the test cases should reflect whichever contract we choose.

___
___
*[`solvers/BUILD.bazel` line 210 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVUeQnKCjjtjRIoN6J_:-NVUeQnLAb0sTG1QVxyS:b-b4xfnr) ([raw file](https://github.com/RobotLocomotion/drake/blob/45084b974e29acaaeab9d6df73e3fa221a823573/solvers/BUILD.bazel#L210)):*
> ```Starlark
>     ],
>     deps = [
>         "//common/symbolic:polynomial",
> ```

nit Unused new deps `symbolic:polynomial`?

jwnimmer-tri:
Oops, I pasted the wrong URL.  I meant to link to https://github.com/RussTedrake/drake/pull/49.

RussTedrake (commented):
i considered doing it this way, but didn't think we would want cost to depend on constraint at the root implementation level (plus it was only a few lines).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-:-NVZC61D795drOfupPHK:b-l229aq)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVUfUi5BRZP8g2CMhXR:-NVZHXTq3DcaiIQahahv:b-f4pfyp) (no related file):*

The docs for this function on `mathematical_program.h` are out of date now:

```c++
  /**
   * Adds a cost in the symbolic form.
   * Note that the constant part of the cost is ignored. So if you set
   * `e = x + 2`, then only the cost on `x` is added, the constant term 2 is
   * ignored.
   * @param e The linear or quadratic expression of the cost.
   * @pre `e` is linear or `e` is quadratic. Otherwise throws a runtime error.
   * @return The newly created cost, together with the bound variables.
   *
   * @exclude_from_pydrake_mkdoc{Not bound in pydrake.}
   */
  Binding<Cost> AddCost(const symbolic::Expression& e);
```

Both in terms of the preconditions, and the ignoring of the constant factor.

When resolving it, having the constant factor be sometimes-ignored and sometimes-obeyed seems too confusing on first impression, so we will need to figure out how to do it all one way or the other, and the test cases should reflect whichever contract we choose.
</blockquote></details>

Done.  Good catch.  It turns out the the AddQuadraticCost docstring was a lie, too.  (The existing tests proved it, but I've added a narrower test to confirm).  And the python binding was bizarre, but is now fixed. 

I don't see any other places where it claims that the constant is dropped.

fyi @hongkai-dai 

___
___
*[`solvers/BUILD.bazel` line 210 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVUeQnKCjjtjRIoN6J_:-NVZCRLGD4BJEfX3hn2w:b8e33y3) ([raw file](https://github.com/RobotLocomotion/drake/blob/45084b974e29acaaeab9d6df73e3fa221a823573/solvers/BUILD.bazel#L210)):*

nit Unused new deps `symbolic:polynomial`?
</blockquote></details>

Done.  (it was providing the decompose.h, but that was also not needed anymore)

jwnimmer-tri (commented):
I should have also clarified -- my main motivation in de-duplication was to avoiding duplicating the non-threadsafe `environment_` member field into a second place.

:lgtm: feature.

+@xuchenhan-tri for platform review per schedule, please.  (I'm on-call today.)

> Reviewed 1 of 6 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-:-NVZmWI32ggNXWji5mzk:b-ppcy6o)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`solvers/cost.h` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVZp1ugARuxI5SzGLY1:-NVZp1ugARuxI5SzGLY2:b-8zefbu) ([raw file](https://github.com/RobotLocomotion/drake/blob/e339f52be75e8565a39d094574ddc19bc28f5fed/solvers/cost.h#L8)):*
> ```objc
> #include <optional>
> #include <string>
> #include <unordered_map>
> ```

nit Unused `<unordered_map>`.  (I thought I'd removed this line in my PR but maybe I forgot to push that edit to github.)
___
*[`solvers/test/cost_test.cc` line 772 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-NVZqIzS1ljHzAlVRqfC:-NVZqIzT7nfyDx376YnB:bx0dg4v) ([raw file](https://github.com/RobotLocomotion/drake/blob/e339f52be75e8565a39d094574ddc19bc28f5fed/solvers/test/cost_test.cc#L772)):*
> ```cpp
>   std::ostringstream os;
>   cost.Display(os, x_e);
>   EXPECT_EQ(fmt::format("{}", os.str()), "ExpressionCost (x * sin(y))");
> ```

nit No need for format, as best I can tell.



_Suggestion:_
```cpp
  EXPECT_EQ(os.str(), "ExpressionCost (x * sin(y))");
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405)* status: LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

xuchenhan-tri (commented):
Platform :lgtm: 

> Reviewed 1 of 6 files at r1, 5 of 6 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19405#-:-NV_zzia-Ypx9atBwR-h:b-cxwtpw)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform)

jwnimmer-tri:
Oops.  I meant to remove the issue number from the commit message on merge.  I should have left a `nit` to remind us.  Oh well.

Maybe at some point we should configure reviewable to help us remember.

PR #19065, opened 2023-03-27
[WIP] Implements SpaceXYZFloatingJoint

amcastro-tri:
Closes #14949

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19065)

PR #19406, opened 2023-05-14
Improve error message for xyz mobilizers in singularity.

RussTedrake:
Previously, users saw `DRAKE_DEMAND(abs(cp) > 1.0e-3)` and had no idea where it was coming from. Now they should get a more helpful error message.

+@xuchenhan-tri for feature review, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19406)

### DIFF
--- a/multibody/tree/BUILD.bazel
+++ b/multibody/tree/BUILD.bazel
@@ -488,6 +488,7 @@ drake_cc_googletest(
         ":mobilizer_tester",
         ":tree",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
@@ -497,6 +498,7 @@ drake_cc_googletest(
         ":mobilizer_tester",
         ":tree",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/multibody/tree/space_xyz_floating_mobilizer.cc
+++ b/multibody/tree/space_xyz_floating_mobilizer.cc
@@ -7,6 +7,7 @@
 #include "drake/common/eigen_types.h"
 #include "drake/math/roll_pitch_yaw.h"
 #include "drake/math/rotation_matrix.h"
+#include "drake/multibody/tree/body.h"
 #include "drake/multibody/tree/multibody_tree.h"
 
 namespace drake {
@@ -153,7 +154,16 @@ void SpaceXYZFloatingMobilizer<T>::DoCalcNMatrix(
   const T cp = cos(angles[1]);
   // Demand for the computation to be away from a state for which Einv_F is
   // singular.
-  DRAKE_DEMAND(abs(cp) > 1.0e-3);
+  if (abs(cp) < 1.0e-3) {
+    // TODO(russt): Add a "likely associated with SpaceXYZFloatingJoint" to
+    // match the error in space_xyz_mobilizer pending #19065.
+    throw std::runtime_error(fmt::format(
+        "The SpaceXYZFloatingMobilizer between body {} and body {} has reached "
+        "a singularity. This occurs when the pitch angle takes values near π/2 "
+        "+ kπ, ∀ k ∈ ℤ. At the current configuration, we have pitch = {}. "
+        "Consider using a QuaternionFloatingJoint instead.",
+        this->inboard_body().name(), this->outboard_body().name(), angles[1]));
+  }
 
   const T sp = sin(angles[1]);
   const T sy = sin(angles[2]);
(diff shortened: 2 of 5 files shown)


xuchenhan-tri (commented):
:lgtm:
+@SeanCurtis-TRI for platform review per rotation please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-:-NVVJsIU6mhzcgUQKHk4:b-rt7w0i)* status: LGTM missing from assignee SeanCurtis-TRI(platform) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-:-NVVO2Nf5omjf7XTbpvv:bmxilld)* status: 2 unresolved discussions

___
*[`multibody/tree/space_xyz_floating_mobilizer.cc` line 159 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-NVVMTZR5iQcWEeF0xnb:-NVVMTZR5iQcWEeF0xnc:brg1dek) ([raw file](https://github.com/RobotLocomotion/drake/blob/c59d1aec5c591fb0e1492b0378b5800ea5cc9a81/multibody/tree/space_xyz_floating_mobilizer.cc#L159)):*
> ```cpp
>   if (abs(cp) < 1.0e-3) {
>         "The SpaceXYZFloatingMobilizer between body {} and body {} has reached "
> ```

BTW We could use a TODO on this to make this error message match that in `space_xyz_mobilizer` in that it references a *joint* type -- once we have a joint type that references it.

Although, I'm curious as to why it's not referenced. Is this dead end code or code that will eventually be plumbed through to a joint?
___
*[`multibody/tree/space_xyz_floating_mobilizer.cc` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-NVVLVxlAtddxuPdPA-O:-NVVLVxlAtddxuPdPA-P:b-9dj4v8) ([raw file](https://github.com/RobotLocomotion/drake/blob/c59d1aec5c591fb0e1492b0378b5800ea5cc9a81/multibody/tree/space_xyz_floating_mobilizer.cc#L161)):*
> ```cpp
>         "The SpaceXYZFloatingMobilizer between body {} and body {} has reached "
>         "a singularity. This occurs when the pitch angle takes values near π/2 "
>         "+ kπ, ∀ k ∈ ℤ. At the current configuration, we have cos(pitch) = {}. "
> ```

BTW The problem is characterized as the value of pitch being close to π/2. Is there any reason to favor printing cos(pitch) instead of simply pitch in the error message? It seems that the characterization and printed value should more obviously go hand in hand. (Yes, a reader can infer the one from the other, but why force them to?) Seems to me that *reading* error messages is not the time to be doing math.

As the BTW indicates, I *can* be content with one of a number of options (but as the prioritized list below suggests, I'm most biased against the current state).

- Express the error condition as cos(pitch) getting too near to zero (and printing that value).
- Do both pitch can't be be "at 90 degrees", with cos(pitch) near zero and then print either of the values.
- Simply printing pitch (and allowing the user to handle periodicity to recognize that they're at the bad rotation).
- Leave it as is.

(Same for the other instances.)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19406)* status: 2 unresolved discussions

___
*[`multibody/tree/space_xyz_floating_mobilizer.cc` line 159 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-NVVMTZR5iQcWEeF0xnb:-NVWX-LF1NGaFgXmmqkE:b8zcz4o) ([raw file](https://github.com/RobotLocomotion/drake/blob/c59d1aec5c591fb0e1492b0378b5800ea5cc9a81/multibody/tree/space_xyz_floating_mobilizer.cc#L159)):*

There's a WIP PR that towards an accompanying joint: https://github.com/RobotLocomotion/drake/pull/19065
</blockquote></details>

That'd be it. The world isn't broken. Or better yet, let's name it in anticipation of the PR merging and not have to revisit this at all.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19406)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform)

___
*[`multibody/tree/space_xyz_floating_mobilizer.cc` line 159 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-NVVMTZR5iQcWEeF0xnb:-NVZ9CHx4sa19ccMtH6n:bup2qgk) ([raw file](https://github.com/RobotLocomotion/drake/blob/c59d1aec5c591fb0e1492b0378b5800ea5cc9a81/multibody/tree/space_xyz_floating_mobilizer.cc#L159)):*

That'd be it. The world isn't broken. Or better yet, let's name it in anticipation of the PR merging and not have to revisit this at all.
</blockquote></details>

Done. it's been open since march, though...
___
*[`multibody/tree/space_xyz_floating_mobilizer.cc` line 161 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19406#-NVVLVxlAtddxuPdPA-O:-NVZAXnz7tMHtHZULELh:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/c59d1aec5c591fb0e1492b0378b5800ea5cc9a81/multibody/tree/space_xyz_floating_mobilizer.cc#L161)):*

BTW The problem is characterized as the value of pitch being close to π/2. Is there any reason to favor printing cos(pitch) instead of simply pitch in the error message? It seems that the characterization and printed value should more obviously go hand in hand. (Yes, a reader can infer the one from the other, but why force them to?) Seems to me that *reading* error messages is not the time to be doing math.

As the BTW indicates, I *can* be content with one of a number of options (but as the prioritized list below suggests, I'm most biased against the current state).

- Express the error condition as cos(pitch) getting too near to zero (and printing that value).
- Do both pitch can't be be "at 90 degrees", with cos(pitch) near zero and then print either of the values.
- Simply printing pitch (and allowing the user to handle periodicity to recognize that they're at the bad rotation).
- Leave it as is.

(Same for the other instances.)
</blockquote></details>

Done.

PR #19407, opened 2023-05-14
[multibody] Improve error message for !(Delta > 0)

### LINKED ISSUE #14590: py mbp: Getting hard exit from "Failure at external/drake/multibody/plant/tamsi_solver.cc:217 in SolveQuadraticForTheSmallestPositiveRoot(): condition 'Delta > 0' failed."?
In some Python code in Anzu, I'm running MBP with some crazy neural network whatever visuomotor stuff, and at some point I get the following *hard* error, using 6ee5e9325821277a62bd5cd5456ccf02ca25dab7
```
Failure at external/drake/multibody/plant/tamsi_solver.cc:217 in SolveQuadraticForTheSmallestPositiveRoot(): condition 'Delta > 0' failed.
```

This should at least be an `Exception` in Python-land? (e.g. #5268).
I'm not sure if I can make a min-repro, but I'll at least see if I can convert this to a softer exception.

May re-assign if there's more nuance here.

FYI @amcastro-tri @sherm1

jwnimmer-tri:
Closes #14590.
Closes #19416.

I suspect someone is feeding a NaN into this arithmetic and it just happens to get all the way to the `sqrt` before we notice.  It's likely that diagnosing the NaN earlier would be better, but anyway this at least somewhat improves the message in this case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407)

### DIFF
--- a/multibody/plant/tamsi_solver.cc
+++ b/multibody/plant/tamsi_solver.cc
@@ -13,7 +13,7 @@ namespace drake {
 namespace multibody {
 namespace internal {
 template <typename T>
-T TalsLimiter<T>::CalcAlpha(
+std::optional<T> TalsLimiter<T>::CalcAlpha(
     const Eigen::Ref<const Vector2<T>>& v,
     const Eigen::Ref<const Vector2<T>>& dv,
     double cos_theta_max, double v_stiction, double relative_tolerance) {
@@ -147,7 +147,13 @@ T TalsLimiter<T>::CalcAlpha(
       // quadratic equation (Δ = b² - 4ac) must be positive.
       // We use a very specialized quadratic solver for this case where we know
       // there must exist a positive (i.e. real) root.
-      alpha = SolveQuadraticForTheSmallestPositiveRoot(a, b, c);
+      // However, in case of a diverging simulation it might not hold true.
+      std::optional<T> maybe_alpha =
+          SolveQuadraticForTheSmallestPositiveRoot(a, b, c);
+      if (!maybe_alpha.has_value()) {
+        return std::nullopt;
+      }
+      alpha = *maybe_alpha;
 
       // The geometry of the problem tells us that α ≤ 1.0
       DRAKE_ASSERT(alpha <= 1.0);
@@ -195,7 +201,7 @@ bool TalsLimiter<T>::CrossesTheStictionRegion(
 }
 
 template <typename T>
-T TalsLimiter<T>::SolveQuadraticForTheSmallestPositiveRoot(
+std::optional<T> TalsLimiter<T>::SolveQuadraticForTheSmallestPositiveRoot(
     const T& a, const T& b, const T& c) {
   using std::abs;
   using std::max;
@@ -214,8 +220,13 @@ T TalsLimiter<T>::SolveQuadraticForTheSmallestPositiveRoot(
   } else {
     // The determinant, Δ = b² - 4ac, of the quadratic equation.
     const T Delta = b * b - 4 * a * c;  // Uppercase, as in Δ.
-    // Geometry tell us that a real solution does exist i.e. Delta > 0.
-    DRAKE_THROW_UNLESS(Delta > 0);
+    // Geometry tell us that a real solution does exist, i.e., Delta > 0.
+    // However, in case of a diverging simulation it might not hold true.
+    if constexpr (scalar_predicate<T>::is_bool) {
+      if (!(Delta > 0)) {
+        return std::nullopt;
+      }
+    }
     const T sqrt_Delta = sqrt(Delta);
 
     // To avoid loss of significance, when 4ac is relatively small compared
@@ -583,7 +594,7 @@ void TamsiSolver<T>::CalcJacobian(
 }
 
 template <typename T>
-T TamsiSolver<T>::CalcAlpha(
+std::optional<T> TamsiSolver<T>::CalcAlpha(
     const Eigen::Ref<const VectorX<T>>& vt,
     const Eigen::Ref<const VectorX<T>>& Delta_vt) const {
   using std::min;
@@ -593,11 +604,13 @@ T TamsiSolver<T>::CalcAlpha(
     const int ik = 2 * ic;  // Index ik scans contact vector quantities.
     auto vt_ic = vt.template segment<2>(ik);
     const auto dvt_ic = Delta_vt.template segment<2>(ik);
-    alpha = min(
-        alpha,
-        internal::TalsLimiter<T>::CalcAlpha(
-            vt_ic, dvt_ic,
-            cos_theta_max_, v_stiction, parameters_.relative_tolerance));
+    const std::optional<T> tals_alpha = internal::TalsLimiter<T>::CalcAlpha(
+        vt_ic, dvt_ic, cos_theta_max_, v_stiction,
+        parameters_.relative_tolerance);
+    if (!tals_alpha.has_value()) {
+      return std::nullopt;
+    }
+    alpha = min(alpha, *tals_alpha);
   }
   DRAKE_DEMAND(0 < alpha && alpha <= 1.0);
   return alpha;
@@ -747,10 +760,13 @@ TamsiSolverResult TamsiSolver<T>::SolveWithGuess(
     vt_error = ExtractDoubleOrThrow(Delta_vt.norm());
 
     // Limit the angle change between vₜᵏ⁺¹ and vₜᵏ for all contact points.
-    T alpha = CalcAlpha(vt, Delta_vt);
+    std::optional<T> alpha = CalcAlpha(vt, Delta_vt);
+    if (!alpha.has_value()) {
+      return TamsiSolverResult::kAlphaSolverFailed;
+    }
 
     // Update generalized velocity vector.
-    v = v + alpha * Delta_v;
+    v = v + alpha.value() * Delta_v;
 
     // Save iteration statistics.
     statistics_.Update(vt_error);
--- a/multibody/plant/tamsi_solver.h
+++ b/multibody/plant/tamsi_solver.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 #include <vector>
 
 #include "drake/common/default_scalars.h"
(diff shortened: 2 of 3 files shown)


jwnimmer-tri:

jwnimmer-tri:
The other solution path would be to remove the `DRAKE_THROW_UNLESS` entirely, and allow the NaN to be returned instead of nullopt.  Presumably that would also show up as a nice "convergence failure" message.

jwnimmer-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-:-NVWBTj9ECIPqOY2eZxH:bfqkacy)* status: 3 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWCBf17gKNBYdJr73J:-NVWCBf17gKNBYdJr73K:bl7inyz) (no related file):*
@amcastro-tri I also just tweaked the exception text in tamsi_driver.  Let me know if it looks okay.

___
___
*[`multibody/plant/tamsi_solver.cc` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUudkc37_gq3AlS-5Y:-NVV_9M81l9IcQJkRPZP:bxy8da1) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/tamsi_solver.cc#L226)):*

is this relying on `NaN > 0` being false? should we place a note?
</blockquote></details>

Phrasing error checking as "if not victory_condition then error" is par for the course when writing code for floating point, so it's not worthy of a comment.  (Anyone who wrote something the manually-negated `if Delta <= 0` if their if-check has formed bad habits and needs retraining.)

However, it is probably worth a more specific regression test, which is added now.  (The existing test did already cause a NaN here already, but that wasn't necessarily obvious when reading it.)
___
*[`multibody/plant/test/tamsi_solver_test.cc` line 370 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUuytS3KKWd2ch_cMX:-NVWC8OSCCoYy8rP2zzS:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/test/tamsi_solver_test.cc#L370)):*

nice! should we have a test when something is NaN?
</blockquote></details>

Done

xuchenhan-tri (commented):
:lgtm: pending CI.

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-:-NVWM8cy52tA6QC3TFNz:bwo9p68)* status: 4 unresolved discussions

___
*[`multibody/plant/tamsi_driver.cc` line 285 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWEHjoBfiJovAfXc7E:-NVWEHjoBfiJovAfXc7F:b-yu5ys4) ([raw file](https://github.com/RobotLocomotion/drake/blob/a7a32cb100ec821f2c0ab529d98b74454ab47098/multibody/plant/tamsi_driver.cc#L285)):*
> ```cpp
>         "  2. decrease the high gains in your controller whenever possible,\n"
>         "  3. switch to a continuous model (discrete update period is zero), "
>         "     though this might affect the simulation run time,",
> ```

nit, do we want a line break here?

amcastro-tri (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407)* status: 3 unresolved discussions

___
*[`multibody/plant/tamsi_driver.cc` line 286 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWLheq5ExNI8FdFteb:-NVWLher9Fx8OP8m1kjU:b1tnhud) ([raw file](https://github.com/RobotLocomotion/drake/blob/a7a32cb100ec821f2c0ab529d98b74454ab47098/multibody/plant/tamsi_driver.cc#L286)):*
> ```cpp
>         "  3. switch to a continuous model (discrete update period is zero), "
>         "     though this might affect the simulation run time,",
>         "  4. avoid terminal bodies with zero mass.",
> ```

I think this might actually be common? This is not a problem for instance if you weld a zero-mass body at the end effector of your robot simply to track an additional frame (overkill IMO, but people do this).
Why do you want to add this here? I believe the messages that Paul put in place already catch cases like this? and do a better job at figuring out whether the body is welded or not, etc?

Wait, did he only do it for continuous?
Maybe then rephrase to something like: "If using zero mass bodies, ensure that their motion implies the motion of another body with non-zero mass". Not sure how useful that is though
___
*[`multibody/plant/tamsi_solver.cc` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUudkc37_gq3AlS-5Y:-NVWL-EM-KXoC5uE9dBA:bx1lp3u) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/tamsi_solver.cc#L226)):*

Phrasing error checking as "if not victory_condition then error" is par for the course when writing code for floating point, so it's not worthy of a comment.  (Anyone who wrote something the manually-negated `if Delta <= 0` if their if-check has formed bad habits and needs retraining.)

However, it is probably worth a more specific regression test, which is added now.  (The existing test did already cause a NaN here already, but that wasn't necessarily obvious when reading it.)
</blockquote></details>

Maybe too obvious, but I think worth adding a one-liner for dummies "N.B. Deta > 0 == false when Delta = NAN.". That way you can train people the right way when reading the comment ;-)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407)* status: 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWCBf17gKNBYdJr73J:-NVWRGpI-p6ZvdNRBGsw:b-kp6z4d) (no related file):*

@amcastro-tri I also just tweaked the exception text in tamsi_driver.  Let me know if it looks okay.
</blockquote></details>

Retracted.

___
___
*[`multibody/plant/tamsi_solver.cc` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUudkc37_gq3AlS-5Y:-NVWQr8PACLcrWjWSFTT:b-o1egur) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/tamsi_solver.cc#L226)):*

Maybe too obvious, but I think worth adding a one-liner for dummies "N.B. Deta > 0 == false when Delta = NAN.". That way you can train people the right way when reading the comment ;-)
</blockquote></details>

No; some deep sinkhole in the middle of the plant solver logic is not the right place to teach developers how floating point works.
___
*[`multibody/plant/tamsi_driver.cc` line 286 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWLheq5ExNI8FdFteb:-NVWQ3dGALfPN5JgwAVd:b-r3qj7d) ([raw file](https://github.com/RobotLocomotion/drake/blob/a7a32cb100ec821f2c0ab529d98b74454ab47098/multibody/plant/tamsi_driver.cc#L286)):*
> I believe the messages that Paul put in place already catch cases like this?

My recollection was that there has been no progress on that front.  If you can cite a place in the code where there is such a check already in place, I would be curious to see it.  I thought the only check was in a unit test, not in runtime code.

For now, I'll just retract this change.  We can add it later once we have sample inputs from users that hit this exception message due to zero-mass problems rather than gain-stiffness problems.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407)* status: 1 unresolved discussion

___
*[`multibody/plant/tamsi_solver.cc` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUudkc37_gq3AlS-5Y:-NVWWInO5nn4LqBJtLfx:boowro6) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/tamsi_solver.cc#L226)):*

No; some deep sinkhole in the middle of the plant solver logic is not the right place to teach developers how floating point works.
</blockquote></details>

FWIW I disagree -- I'm very aware of how floating point works but think the odd NaN behavior is worth a reminder every time. It is very easy when reading the code to think "math" rather than "computer quirk".

jwnimmer-tri (commented):
See also #19416 for an alternative approach that leans into the NaNs instead using optional.  I am OK with either way.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-:-NV_27WO2-OiEVjDApBS:b-9tw66u)* status: 1 unresolved discussion

___
*[`multibody/plant/tamsi_solver.cc` line 226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVUudkc37_gq3AlS-5Y:-NV_2JsS45R39gKnUW6X:bvv2mlz) ([raw file](https://github.com/RobotLocomotion/drake/blob/7dc0ea19d7723914c2d970e5a0fbe5a42abfce88/multibody/plant/tamsi_solver.cc#L226)):*

FWIW I disagree -- I'm very aware of how floating point works but think the odd NaN behavior is worth a reminder every time. It is very easy when reading the code to think "math" rather than "computer quirk".
</blockquote></details>

See #19416 for what that looks like.

amcastro-tri (commented):
I think I like the optional version better. I am not a big fan of using special values like NaN to signal failure. But obviously, that's just a preference. I'll let you choose.

Now, what I do like about #19416, is how carefully you went through the logic and flow of NANs and made meticulous comments where appropriate. If you like `optional` better, I'd also place those comments in this PR. non-blocking though.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-:-NV_DBnS5nqDRuaJTPZ_:b-r7ko2q)* status: :shipit: complete! all discussions resolved, LGTM from assignees amcastro-tri,xuchenhan-tri(platform)

___
*[`multibody/plant/tamsi_driver.cc` line 286 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19407#-NVWLheq5ExNI8FdFteb:-NV_EFJX5Z8Faw7UAlNo:baigujj) ([raw file](https://github.com/RobotLocomotion/drake/blob/a7a32cb100ec821f2c0ab529d98b74454ab47098/multibody/plant/tamsi_driver.cc#L286)):*

> I believe the messages that Paul put in place already catch cases like this?

My recollection was that there has been no progress on that front.  If you can cite a place in the code where there is such a check already in place, I would be curious to see it.  I thought the only check was in a unit test, not in runtime code.

For now, I'll just retract this change.  We can add it later once we have sample inputs from users that hit this exception message due to zero-mass problems rather than gain-stiffness problems.
</blockquote></details>

SGTM, thank you. I believe Paul's code only works in continuous mode, so TAMSI does not exercise it.

sherm1:
In my experience with scientific software, most f.p. if statements are written with no thought to NaN behavior (and I believe that is true for the vast majority of conditionals in Drake). If someone has actually thought that through and crafted the sense of that condition to work properly with NaNs I think it is worth saying so to prevent a future teenage programmer from doing "clean up". 

It is very natural to look at a conditional in mathematical software with a "math" mindset rather than the more appropriate "bizarre computational quirks" mindset. Worth a prod to the right mindset, IMO.

jwnimmer-tri:
I still think the best guard against teenage programmers is unit tests, not comment pasta.

sherm1:
I agree that NaN-injected unit tests would be sufficient. I'd still comment any NaN-aware conditionals or (better) use std::isfinite() to make it unnecessary to think of the conditionals in any way other than the apparent math.

jwnimmer-tri:
Yes, I would support adding `isfinite` in preference over text comments.

PR #19422, opened 2023-05-17
Overload Trajectory::vector_values to accept Eigen::VectorXd

RussTedrake:
This makes it easier (in c++) to, for instance, using Eigen::LinSpaced to sample at regular intervals from a trajectory.

+@xuchenhan-tri for both reviews, please (if it's not too late).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19422)

### DIFF
--- a/bindings/pydrake/trajectories_py.cc
+++ b/bindings/pydrake/trajectories_py.cc
@@ -240,8 +240,10 @@ struct Impl {
       cls  // BR
           .def(py::init<>())
           .def("value", &Class::value, py::arg("t"), cls_doc.value.doc)
-          .def(
-              "vector_values", &Class::vector_values, cls_doc.vector_values.doc)
+          .def("vector_values",
+              overload_cast_explicit<MatrixX<T>, const std::vector<T>&>(
+                  &Class::vector_values),
+              py::arg("t"), cls_doc.vector_values.doc)
           .def("has_derivative", &Class::has_derivative,
               cls_doc.has_derivative.doc)
           .def("EvalDerivative", &Class::EvalDerivative, py::arg("t"),
--- a/common/trajectories/test/piecewise_polynomial_test.cc
+++ b/common/trajectories/test/piecewise_polynomial_test.cc
@@ -287,13 +287,20 @@ GTEST_TEST(testPiecewisePolynomial, VectorValueTest) {
   // Note: Keep one negative time to confirm that negative values can work for
   // constant trajectories.
   const std::vector<double> times = {-1.5, 0, .5, 1, 1.5};
+  const Eigen::Map<const Eigen::VectorXd> etimes(times.data(), times.size());
   const Eigen::Vector3d value(1, 2, 3);
 
   const PiecewisePolynomial<double> col(value);
   Eigen::MatrixXd out = col.vector_values(times);
   EXPECT_EQ(out.rows(), 3);
   EXPECT_EQ(out.cols(), 5);
-  for (int i = 0; i < 4; i++) {
+  for (int i = 0; i < 4; ++i) {
+    EXPECT_TRUE(CompareMatrices(out.col(i), value, 0));
+  }
+  out = col.vector_values(etimes);
+  EXPECT_EQ(out.rows(), 3);
+  EXPECT_EQ(out.cols(), 5);
+  for (int i = 0; i < 4; ++i) {
     EXPECT_TRUE(CompareMatrices(out.col(i), value, 0));
   }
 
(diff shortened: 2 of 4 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19422)* status: 2 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),xuchenhan-tri(platform) (waiting on @xuchenhan-tri)

___
*[`common/trajectories/trajectory.h` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfW71-CksFjW0102SW:-NVfW71-CksFjW0102SX:b-vgcsc1) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.h#L56)):*
> ```objc
>    * @throws std::exception if both cols and rows are not equal to 1.
>    */
>   MatrixX<T> vector_values(const Eigen::Ref<const Eigen::VectorXd>& t) const;
> ```

BTW Why is this overload limited to double-valued times, while the other one allows for autodiff times (etc)?



_Suggestion:_
```objc
VectorX<T>
```
___
*[`common/trajectories/trajectory.cc` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfWFCS7sv84T7z7Vix:-NVfWFCS7sv84T7z7Viy:bmg6685) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.cc#L9)):*
> ```cpp
> 
> template <typename T>
> MatrixX<T> Trajectory<T>::vector_values(const std::vector<T>& t) const {
> ```

BTW If the other overload accepts `<T>` instead of `double`, then this function should switch to calling the other one using an `Eigen::Map` to avoid the duplication.

xuchenhan-tri (commented):
:lgtm: pending Jeremy's comment.

BTW should we include this in the release notes? It'd be something new for C++ users.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-:-NVfegV5269dui9Sph7g:b9frm8t)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform) (waiting on @jwnimmer-tri)

___
*[`common/trajectories/trajectory.cc` line 37 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfcpvdB38BhdCB0OwZ:-NVfcpvdB38BhdCB0Ow_:bgucvaz) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.cc#L37)):*
> ```cpp
>   if (cols() == 1) {
>     MatrixX<T> values(rows(), t.size());
>     for (int i = 0; i < static_cast<int>(t.size()); i++) {
> ```

nit, preincrement here and throughout the PR



_Suggestion:_
```cpp
++i
```

jwnimmer-tri (commented):
Yes +(release notes: fix) -(release notes: none) this is a new API so we should mention it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-:-NVgE-4755a58O9cGboN:bqdobkg)* status: 3 unresolved discussions

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19422)* status: :shipit: complete! all discussions resolved, LGTM from assignee xuchenhan-tri(platform)

___
*[`common/trajectories/trajectory.h` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfW71-CksFjW0102SW:-NW27U6LC8gaur6xTvNZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.h#L56)):*

BTW Why is this overload limited to double-valued times, while the other one allows for autodiff times (etc)?
</blockquote></details>

Done.
___
*[`common/trajectories/trajectory.cc` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfWFCS7sv84T7z7Vix:-NW27Uuj5YLXjwHlQGAd:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.cc#L9)):*

BTW If the other overload accepts `<T>` instead of `double`, then this function should switch to calling the other one using an `Eigen::Map` to avoid the duplication.
</blockquote></details>

Done.
___
*[`common/trajectories/trajectory.cc` line 37 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19422#-NVfcpvdB38BhdCB0OwZ:-NW27_6DEALSTcBZvknX:b3iievr) ([raw file](https://github.com/RobotLocomotion/drake/blob/b50b072e48225468c79eb03706dc17b239668ef6/common/trajectories/trajectory.cc#L37)):*

nit, preincrement here and throughout the PR
</blockquote></details>

Done.  (old habits die hard)

PR #19426, opened 2023-05-18
Make RenderEngineGl intelligently threadsafe

SeanCurtis-TRI:
This allows a single RenderEngineGL to be cloned so that each clone can safely be used in different threads.

1. The OpenGlContext for each RenderEngineGl is thread-independent; binding one in one thread, doesn't preclude binding another in another thread.
2. OpenGlContexts that get cloned from each other share the same OpenGl artifacts (display lists, etc.) on the GPU.

Introduces a unit test that failed under the old non-threadsafe regime.

TODO:
 - Delete thread_test_2.cc
 - Change OpenGlContext so the test passes.
 - Chnage the test to make sure OpenGlContext can be tested directly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426)

### DIFF
--- a/geometry/render_gl/BUILD.bazel
+++ b/geometry/render_gl/BUILD.bazel
@@ -313,8 +313,13 @@ drake_cc_googletest_ubuntu_only(
     ],
     tags = vtk_test_tags() + ["cpu:4"],
     deps = [
+        ":factory",
+        ":internal_render_engine_gl",
         ":internal_texture_library",
         "//common:find_resource",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//systems/sensors:image",
+        "//systems/sensors:image_writer",
     ],
 )
 
--- a/geometry/render_gl/factory.h
+++ b/geometry/render_gl/factory.h
@@ -25,9 +25,23 @@ extern const bool kHasRenderEngineGl;
  Drake ecosystem, only the visualization will be upside down. This has been
  documented in https://github.com/RobotLocomotion/drake/issues/14254.
 
- @warning %RenderEngineGl is not threadsafe. If a SceneGraph is instantiated
- with a RenderEngineGl and there are multiple Context instances for that
- SceneGraph, rendering in multiple threads may exhibit issues.
+ <b> Using RenderEngineGl in multiple threads </b>
+
+ Most importantly, a single %RenderEngineGl should *not* be exercised in
+ multiple threads. One thread, one %RenderEngineGl instance.
+
+ A %RenderEngineGl instance and its *clones* can be used in different threads
+ simultaneously, but *only* the rendering APIs are threadsafe. Do not mutate the
+ contents of the engine (e.g., adding/removing geometries, etc.) in parallel.
+
+ Two independently constructed %RenderEngineGl instances can be freely used in
+ different threads -- all APIs are available.
+
+ The expected workflow is to add a %RenderEngineGl instance a SceneGraph
+ instance (see SceneGraph::AddRenderer()) and then to populate SceneGraph with
+ the desired geometry. Each systems::Context allocated for that SceneGraph will
+ receive a clone of the original %RenderEngineGl. One systems::Context can be
+ used per thread to create rendered images in parallel.
 
  @throws std::exception if kHasRenderEngineGl is false. */
 std::unique_ptr<render::RenderEngine> MakeRenderEngineGl(
--- a/geometry/render_gl/internal_opengl_context.cc
+++ b/geometry/render_gl/internal_opengl_context.cc
@@ -79,8 +79,12 @@ static Display* display() {
   // program exits, but it seems not so evil to skip that.
   // (https://linux.die.net/man/3/xclosedisplay)
   // If problems crop up in the future, this can/should be investigated.
-  static Display* g_display = (XInitThreads(), XOpenDisplay(0));
-  DRAKE_THROW_UNLESS(g_display != nullptr);
+  static Display* g_display = []() {
+    XInitThreads();
+    Display* display = XOpenDisplay(0);
+    DRAKE_THROW_UNLESS(display != nullptr);
+    return display;
+  }();
   return g_display;
 }
 
--- a/geometry/render_gl/internal_opengl_context.cc
+++ b/geometry/render_gl/internal_opengl_context.cc
@@ -72,13 +72,25 @@ void GlDebugCallback(GLenum, GLenum type, GLuint, GLenum severity, GLsizei,
   }
 }
 
+static Display* display() {
+  // Turn Display into a singleton to make CI happy, since when we close and
+  // reopen the display on CI, we can't request a new OpenGL context.
+  // This pattern won't call the corresponding `XCloseDisplay()` when the
+  // program exits, but it seems not so evil to skip that.
+  // (https://linux.die.net/man/3/xclosedisplay)
+  // If problems crop up in the future, this can/should be investigated.
+  static Display* g_display = (XInitThreads(), XOpenDisplay(0));
+  DRAKE_THROW_UNLESS(g_display != nullptr);
+  return g_display;
+}
+
 }  // namespace
 
 class OpenGlContext::Impl {
  public:
   // Open an X display and initialize an OpenGL context. The display will be
   // open and ready for offscreen rendering, but no window is visible.
-  explicit Impl(bool debug) {
+  explicit Impl(bool debug, GLXContext source_context = NULL) : debug_(debug) {
     // See Offscreen Rendering section here:
     // https://sidvind.com/index.php?title=Opengl/windowless
 
@@ -102,6 +114,13 @@ class OpenGlContext::Impl {
                                   None};
     int fb_count = 0;
     const int screen_id = DefaultScreen(display());
+
+    // No matter what, we want to make sure that the context is not current
+    // at the conclusion of construction.
+    ScopeExit unbind([]() {
+      OpenGlContext::ClearCurrent();
+    });
+
     GLXFBConfig* fb_configs =
         glXChooseFBConfig(display(), screen_id, kVisualAttribs, &fb_count);
     ScopeExit guard([fb_configs]() { XFree(fb_configs); });
@@ -157,8 +176,8 @@ class OpenGlContext::Impl {
     // NOTE: The consts True and False come from gl/glx.h (indirectly), but
     // ultimately from X11/Xlib.h.
     // This requires a call to glXDestroyContext in the destructor.
-    context_ = glXCreateContextAttribsARB(display(), fb_configs[0], 0, True,
-                                          kContextAttribs);
+    context_ = glXCreateContextAttribsARB(
+        display(), fb_configs[0], source_context, True, kContextAttribs);
     if (context_ == nullptr) {
       throw std::runtime_error(
           "Error initializing OpenGL Context for RenderEngineGL; failed to "
@@ -171,17 +190,20 @@ class OpenGlContext::Impl {
 
     XSync(display(), False);
 
-    // Make it the current context.
-    MakeCurrent();
-
     // Enable debug.
     if (debug) {
+      MakeCurrent();
       drake::log()->info("Vendor: {}", GetGlVendor());
       glEnable(GL_DEBUG_OUTPUT);
+      glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
       glDebugMessageCallback(GlDebugCallback, 0);
     }
     is_complete = true;
   }
+
+  // Constructs a copy which shares the OpenGl objects stored on the GPU with
+  // `other`.
+  Impl(const Impl& other) : Impl(other.debug_, other.context_) {}
 
   ~Impl() {
     glXDestroyContext(display(), context_);
@@ -196,6 +218,10 @@ class OpenGlContext::Impl {
         !glXMakeCurrent(display(), window_, context_)) {
       throw std::runtime_error("Error making an OpenGL context current");
     }
+  }
+
+  bool IsCurrent() const {
+    return glXGetCurrentContext() == context_;
   }
 
   void DisplayWindow(const int width, const int height) {
@@ -268,19 +294,6 @@ class OpenGlContext::Impl {
   }
 
  private:
-  static Display* display() {
-    // Turn Display into a singleton to make CI happy, since when we close and
-    // reopen the display on CI, we can't request a new OpenGL context.
-    // This pattern won't call the corresponding `XCloseDisplay()` when the
-    // program exits, but it seems not so evil to skip that.
-    // (https://linux.die.net/man/3/xclosedisplay)
-    // TODO(duy): If problems crop up in the future, this can/should be
-    // investigated.
-    static Display* g_display = (XInitThreads(), XOpenDisplay(0));
-    DRAKE_THROW_UNLESS(g_display != nullptr);
-    return g_display;
-  }
-
   GLXContext context_{nullptr};
 
   // The associated window to support display of rendering results.
@@ -292,14 +305,23 @@ class OpenGlContext::Impl {
   // not yet understood.
   int window_width_{640};
   int window_height_{480};
+
+  const bool debug_{};
 };
 
 OpenGlContext::OpenGlContext(bool debug)
     : impl_(new OpenGlContext::Impl(debug)) {}
 
+OpenGlContext::OpenGlContext(const OpenGlContext& other)
+    : impl_(std::make_unique<OpenGlContext::Impl>(*other.impl_)) {}
+
 OpenGlContext::~OpenGlContext() = default;
 
 void OpenGlContext::MakeCurrent() const { impl_->MakeCurrent(); }
+
+void OpenGlContext::ClearCurrent() { glXMakeCurrent(display(), None, NULL); }
+
+bool OpenGlContext::IsCurrent() const { return impl_->IsCurrent(); }
 
 void OpenGlContext::DisplayWindow(const int width, const int height) {
   impl_->DisplayWindow(width, height);
--- a/geometry/render_gl/internal_opengl_context.h
+++ b/geometry/render_gl/internal_opengl_context.h
@@ -16,18 +16,43 @@ namespace internal {
  */
 class OpenGlContext {
  public:
-  /* Constructor. Initializes an OpenGL context.
+  /* Constructor. Initializes a brand new OpenGL context without any objects.
+
    @param debug  If debug is true, the OpenGl context will be a "debug" context,
    in that the OpenGl implementation's errors will be written to the Drake log.
    See https://www.khronos.org/opengl/wiki/Debug_Output for more information.
    */
   explicit OpenGlContext(bool debug = false);
 
+  /* Copy constructs a context that shares OpenGl objects with `other`.
+
+   - All %OpenGlContext instances share a common display.
+   - Each instance has a unique window/OpenGL objects for connecting an OpenGl
+     context to an X-windows display.
+   - Copying an %OpenGlContext is distinct from creating a new instance in that
+     all of the OpenGl objects (buffers, textures, display lists, etc.)
+     available to `other` will also be available to the copy. New instances have
+     no OpenGl objects. */
+  OpenGlContext(const OpenGlContext& other);
+
+  /* All other copy and move semantics are simply deleted. */
+  OpenGlContext& operator=(const OpenGlContext& other) = delete;
+  OpenGlContext(OpenGlContext&&) = delete;
+  OpenGlContext& operator=(OpenGlContext&&) = delete;
+
   ~OpenGlContext();
 
   /* Makes this context current.
    @throws std::exception if not successful.  */
   void MakeCurrent() const;
+
+  /* Clears the "current" context from the current thread. Note: if this context
+   has been bound in one thread and this called in another thread, the context
+   will still be bound in the first thread. */
+  static void ClearCurrent();
+
+  /* Reports if this context is bound to the current thread. */
+  bool IsCurrent() const;
 
   /* Displays the window at the given dimensions. Calling this redundantly (on
    an already visible window of the given size) has no effect.  */
--- a/geometry/render_gl/internal_render_engine_gl.cc
+++ b/geometry/render_gl/internal_render_engine_gl.cc
@@ -7,6 +7,7 @@
 
 #include <fmt/format.h>
 
+#include "drake/common/scope_exit.h"
 #include "drake/common/ssize.h"
 #include "drake/common/text_logging.h"
 #include "drake/common/unused.h"
@@ -64,9 +65,6 @@ class DefaultRgbaColorShader final : public ShaderProgram {
   explicit DefaultRgbaColorShader(const Rgba& default_diffuse)
       : ShaderProgram(), default_diffuse_(default_diffuse) {
     LoadFromSources(kVertexShader, kFragmentShader);
-    diffuse_color_loc_ = GetUniformLocation("diffuse_color");
-    normal_mat_loc_ = GetUniformLocation("normal_mat");
-    light_dir_loc_ = GetUniformLocation("light_dir_C");
   }
 
   void SetInstanceParameters(const ShaderProgramData& data) const final {
@@ -79,6 +77,12 @@ class DefaultRgbaColorShader final : public ShaderProgram {
   }
 
  private:
+  void DoConfigureUniforms() final {
+    diffuse_color_loc_ = GetUniformLocation("diffuse_color");
+    normal_mat_loc_ = GetUniformLocation("normal_mat");
+    light_dir_loc_ = GetUniformLocation("light_dir_C");
+  }
+
   std::unique_ptr<ShaderProgram> DoClone() const final {
     return make_unique<DefaultRgbaColorShader>(*this);
   }
@@ -179,10 +183,6 @@ class DefaultTextureColorShader final : public ShaderProgram {
       : ShaderProgram(), library_(std::move(library)) {
     DRAKE_DEMAND(library_ != nullptr);
     LoadFromSources(kVertexShader, kFragmentShader);
-    diffuse_map_loc_ = GetUniformLocation("diffuse_map");
-    diffuse_scale_loc_ = GetUniformLocation("diffuse_map_scale");
-    normal_mat_loc_ = GetUniformLocation("normal_mat");
-    light_dir_loc_ = GetUniformLocation("light_dir_C");
   }
 
   void SetInstanceParameters(const ShaderProgramData& data) const final {
@@ -198,6 +198,13 @@ class DefaultTextureColorShader final : public ShaderProgram {
   }
 
  private:
+  void DoConfigureUniforms() final {
+    diffuse_map_loc_ = GetUniformLocation("diffuse_map");
+    diffuse_scale_loc_ = GetUniformLocation("diffuse_map_scale");
+    normal_mat_loc_ = GetUniformLocation("normal_mat");
+    light_dir_loc_ = GetUniformLocation("light_dir_C");
+  }
+
   std::unique_ptr<ShaderProgram> DoClone() const final {
     return make_unique<DefaultTextureColorShader>(*this);
   }
@@ -323,8 +330,6 @@ class DefaultDepthShader final : public ShaderProgram {
 
   DefaultDepthShader() : ShaderProgram() {
     LoadFromSources(kVertexShader, kFragmentShader);
-    depth_z_near_loc_ = GetUniformLocation("depth_z_near");
-    depth_z_far_loc_ = GetUniformLocation("depth_z_far");
   }
 
   void SetDepthCameraParameters(
@@ -334,6 +339,11 @@ class DefaultDepthShader final : public ShaderProgram {
   }
 
  private:
+  void DoConfigureUniforms() final {
+    depth_z_near_loc_ = GetUniformLocation("depth_z_near");
+    depth_z_far_loc_ = GetUniformLocation("depth_z_far");
+  }
+
   std::unique_ptr<ShaderProgram> DoClone() const final {
     return make_unique<DefaultDepthShader>(*this);
   }
@@ -418,7 +428,6 @@ class DefaultLabelShader final : public ShaderProgram {
       std::function<Vector4<float>(const PerceptionProperties&)> label_encoder)
       : ShaderProgram(), label_encoder_(std::move(label_encoder)) {
     LoadFromSources(kVertexShader, kFragmentShader);
-    encoded_label_loc_ = GetUniformLocation("encoded_label");
   }
 
   void SetInstanceParameters(const ShaderProgramData& data) const final {
@@ -427,6 +436,10 @@ class DefaultLabelShader final : public ShaderProgram {
   }
 
  private:
+  void DoConfigureUniforms() final {
+    encoded_label_loc_ = GetUniformLocation("encoded_label");
+  }
+
   std::unique_ptr<ShaderProgram> DoClone() const final {
     return make_unique<DefaultLabelShader>(*this);
   }
@@ -471,22 +484,13 @@ void main() {
 
 RenderEngineGl::RenderEngineGl(RenderEngineGlParams params)
     : RenderEngine(params.default_label),
-      opengl_context_(make_shared<OpenGlContext>()),
+      opengl_context_(make_unique<OpenGlContext>()),
       texture_library_(make_shared<TextureLibrary>()),
       parameters_(std::move(params)) {
   // Configuration of basic OpenGl state.
   opengl_context_->MakeCurrent();
-  glEnable(GL_CULL_FACE);
-  glCullFace(GL_BACK);
-  glClipControl(GL_UPPER_LEFT, GL_NEGATIVE_ONE_TO_ONE);
-  glClearDepth(1.0);
-  glEnable(GL_DEPTH_TEST);
-  // Generally, there should be no blending for depth and label images. We'll
-  // selectively enable blending for color images.
-  glDisable(GL_BLEND);
-  // We blend the rgb values (the first two parameters), but simply accumulate
-  // transparency (the last two parameters).
-  glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE);
+
+  InitGlState();
 
   // Color shaders. See documentation on GetShaderProgram. We want color from
   // texture to be "more preferred" than color from rgba, so we add the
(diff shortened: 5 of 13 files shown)


SeanCurtis-TRI (commented):
@jwnimmer-tri I'd like you to look at this.

R2 is (more or less) the finished, correct fix. RenderEngineGl and OpenGlContext are well behaved. However, the copy constructor i incomplete and `DoClone()` does the extra work.
R3 is the change necessary to make the copy constructor self sufficient. A large portion of it is the movement of geometry management out of `RenderEngineGl` and into the new `GeometryLibrary`.

I'd like you to particularly look at R3 (although I'll welcome comments more generally).

(Not all the unit tests pass -- this is known. I'll be patching that in the morning. One of the failed unit tests (multithread_safety_test) is due to the distinction between CI and my puget - CI and my laptop don't mind binding contexts in multiple threads.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW4crlmEebOLS_08n-j:b-tvta2t)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
Checkpoint (prior to meetings) with some aside comments.  I'll return to the meat of r3 stuff a bit later on.

> Reviewed 3 of 5 files at r1, 1 of 10 files at r2, 1 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW8V9tUBFjhIo6s08zb:b57bpup)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_texture_library.h` line 16 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8SnKc8_cJxs-5bC8o:-NW8SnKc8_cJxs-5bC8p:b-sn9kga) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_texture_library.h#L16)):*
> ```objc
> namespace internal {
> 
> /* Stores a set of OpenGl textures objects, keyed by their in-filesystem name.
> ```

BTW Possibly worth a TODO and/or issue:

Especially with Bazel, many filesystem paths might be symlinks to the same underlying image.  The cache should be keyed on the sha256 of the file contents, not the file's path.

This is also true because images might change on disk, and we shouldn't be caching the old image contents if the file changed but its path did not.
___
*[`geometry/render_gl/internal_texture_library.h` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8SVqFBy5ZL1dfijV9:-NW8SVqFBy5ZL1dfijVA:b-u0odyo) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_texture_library.h#L42)):*
> ```objc
>    @returns A valid OpenGl identifier if the texture has been successfully
>             added to the library, std::nullopt if not.
>    @pre The OpenGl context associated with this library has been bound. */
> ```

nit It would probably help clarify here to document that "associated with" is implicit in the libglx per-thread singleton-ness -- there no Drake C++ pointer or something like that which associates the two concepts.  It's all in the background.

jwnimmer-tri (commented):
Okay, I've read all of the code (but not tests).  Any particular feedback you'd like?

I think maybe the biggest missing piece is an overview paragraph somewhere in this subsystem (maybe in internal_render_engine_gl.h) that explains a mental model of the state that is not apparent in our own C++ data structures.  (Basically, the GLX state.)  Most of the code is dealing with synchronizing and sharing our C++ state with that GL / GPU state, so it would help to clearly articulate and summarize (1) the ontology of the GL / GPU state, and (2) our invariants / premise about how we sync our state with the GL state, in terms of activation (currency), locking, reuse, and cloning.

> Reviewed 4 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW8vbipELS167G60tk6:b-xtamjk)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_opengl_context.h` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8qBWg9RkoFI9itrQg:-NW8qBWhEtIljzrkIxhC:bka4hto) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_opengl_context.h#L19)):*
> ```objc
> class OpenGlContext {
>  public:
>   /* Copy constructor. Provided solely so RenderEngineGl can maintain a
> ```

BTW https://drake.mit.edu/styleguide/cppguide.html#Declaration_Order says that when using the DRAKE_...COPY... macro the macro must come first, but other than that constructor order is up to us.  For my taste, laying out the functions in the order they would be called (default ctor, followed by copy ctor) makes the documentation easier to read when in source code form.

That would allow us to say "default ctor: makes a context (and display, the first time), doesn't have any objects" vs "copy ctor: retains any objects from 'other'".

(It would also better match the cc file implementation order.)
___
*[`geometry/render_gl/internal_opengl_context.h` line 20 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8pRDF4j1qUSPKZqN_:-NW8pRDG9LK7X6m5FfUM:b1d1zwu) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_opengl_context.h#L20)):*
> ```objc
>  public:
>   /* Copy constructor. Provided solely so RenderEngineGl can maintain a
>    copyable_unique_ptr for this class.
> ```

BTW It seems to me like defining `Clone` on this might be less hazardous than defining copying.  The copyable_unique_ptr would be indifferent, and other random code couldn't accidentally invoke the coping.

SeanCurtis-TRI (commented):
The first thing I wanted to do was to make sure I hadn't misinterpreted what you'd suggested vis a vis automating the copying. As you didn't say anything, then I'm assuming the mechanism appears sound.

For me, I wasn't persuaded the R3 was particularly worth it. It's all in support of one purpose: make the *private* copy constructor of `RenderEngineGl` "complete". It's no more *correct* vis a vis calling the `Clone()` method.

This comes with some of its own weirdness:
- Playing with OpenGl state is further distributed (outside of `internal_render_engine_gl.cc`.
- There are restrictions on the members that future editors have to be aware of -- comments on the wrapper and unbinder alone may not be enough.
- The mechanisms that rely initializaiton/construction order are sufficiently esoteric, that being familiar with other corners of Drake will not prepare a reader for this class.
- There are a lot of masked effects - it's not clear how much voodoo takes place in the constructor/copy constructor.

On the hand, R2 reads more obviously and directly. The only real cost is that the private constructor has documentation that says , "Only `DoClone()` should call this."

So, is it worth it?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW9B5491Lf5VTOzlHHK:bg48t40)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
I totally did not grok that r3 was only the C++ control flow refactor.  I thought it was fixing meaningful sharing/race problems.  It's certainly not worth it just for what we were discussing on slack.

Maybe I can start from r2 and provide a patch for what I was thinking about.  It should be really just like 10 lines.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW9D9Rl00teFqLBR8zB:bwwk07n)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
The `GeometryLibrary` is necessary so that all of the vertex array objects get handled as part of copying.
THe changes to the header (`OpenGlContextWrapper`, `GlInitializer`, and `ContextUnbinder`) are part of that machinery.

THe key is for every unique "blob" of information that needs to be handled, I need a type that does the work upon construction/copying.

The majority of the lines of change are *moving* functionality *out* of `RenderEngineGl` and into the new `GeometryLibrary`. I had a plan for how I would PR to separate the code movement from actual changes. So, you can more or less *ignore* `GeometryLibrary` and ignore the deletions of whole functions from `RenderEngineGl` and largely limit your consideration to the classes introduced in the header.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NW9ERZi-zpftUsh0_b_:b-4pe91o)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @jwnimmer-tri)

jwnimmer-tri:
Okay, in terms of the copy ctor `RenderEngineGl(const RenderEngineGl&) = default` being a hazard because it doesn't finish the post-copying tasks, I think it would be sufficient to mark it private and call it a day.  Having the post-facto adjustments live inside `DoClone` all in one place seems just fine.

Separately, if it's _clearer on its own_ to carve out any helper classes to make responsibilities clear, I think that's fine.  (The DoClone to call functions on helper classes, for example.)  But I won't care them up just to get magic from `RenderEngineGl(const RenderEngineGl&) = default`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_opengl_context.h` line 20 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8pRDF4j1qUSPKZqN_:-NX6AKJH65LNIXrJYWtp:b-drktop) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_opengl_context.h#L20)):*

While I can see that argument, I'm choosing to avoid it for ergonomic reasons:

1. To meaningfully do that, I have to make the copy constructor private. This means either a) I can't use `make_unique` or I have to friend it in.
2. It introduces unique pointers to all of my test code which is like wise an inconvenience.
3. This is an internal class -- if people are running around copying this, then they are running around where they shouldn't be.

The overhead for `Clone()` seems heavier than I want for something buried so deeply.
</blockquote></details>

(1) This is not very important.  Using the raw pointer constructor for unique_ptr is typical in cases like this, and not meaningfully worse.

(2) This is a fair reason to leave it alone (leaving it as copyable).

(3) The hazard isn't about users -- it's about internal code within Drake that is accidentally making a copy when we didn't intend for it to do so.  Since the copy semantics are a little bit special, having a reminder to the reader at call sites that something special is happening (via the extra syntax) could be an advantage.  The question would be the tradeoff of that possible benefit vs the downsides of 2.

SeanCurtis-TRI (commented):
Ok, assuming CI passes, we'll be ready for feature review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXChfTt4yIDD0ltIEez:b-5g6njn)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_opengl_context.h` line 20 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NW8pRDF4j1qUSPKZqN_:-NXCh3ts-EGBscULgCpN:bx906o0) ([raw file](https://github.com/RobotLocomotion/drake/blob/5676f1d9a504b2214d2f664a7e5bfd1eb5078cb3/geometry/render_gl/internal_opengl_context.h#L20)):*

(1) This is not very important.  Using the raw pointer constructor for unique_ptr is typical in cases like this, and not meaningfully worse.

(2) This is a fair reason to leave it alone (leaving it as copyable).

(3) The hazard isn't about users -- it's about internal code within Drake that is accidentally making a copy when we didn't intend for it to do so.  Since the copy semantics are a little bit special, having a reminder to the reader at call sites that something special is happening (via the extra syntax) could be an advantage.  The question would be the tradeoff of that possible benefit vs the downsides of 2.
</blockquote></details>

(3) i wasn't thinking about users. Even developers aren't going to be running around and using this. It's not a utility class. It has zero value outside of `render_gl`. I feel like we can count the number of people who will trip over this on the fingers of one foot. (Assuming we always have at least one feature reviewer that understands OpenGl.) 

Still, assuming the current version passes CI, if you feel strongly about `Clone()` only, I'll change it. You're the feature reviewer.  :)  I agree that some ergonomic ugliness isn't the conclusive argument against it.

jwnimmer-tri (commented):
Checkpoint.

> Reviewed 8 of 20 files at r4, 9 of 12 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXGvT_eEk-T4DQRGkok:by96gdd)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/factory.h` line 33 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGszo4Ch7JqPUWe1Kg:-NXGszo5-oYKQ0WJR2Zg:bvl3xxe) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/factory.h#L33)):*
> ```objc
>  then simply providing a unique systems::Context for each thread will be
>  sufficient. Mutating the *contents* of the %RenderEngineGl (e.g., adding or
>  removing geometries) is *not* threadsafe.
> ```

BTW Just to make sure I completely understand -- let me check this sequence:

(1) Add one GL engine to the scene graph.
(2) Add the scene graph to the diagram.
(3) Launch several worker threads, each with their own dedicated diagram context.
(4) The worker thread removes some perception geometry (by deleting some perception properties) using the context-specific functions on SceneGraph.

From the comment, I believe that step (4) is unsafe.  Is that right?

(Assuming yes, there is no need to change any wording, just making sure.)
___
*[`geometry/render_gl/internal_opengl_context.cc` line 309 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGq6_v8MzmRnS0diOf:-NXGq6_w27aT0636cXqg:b-l60qqn) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_opengl_context.cc#L309)):*
> ```cpp
>   int window_height_{480};
> 
>   bool debug_{};
> ```

BTW consider marking `const debug_;`, I _think_ we never reassign it?
___
*[`geometry/render_gl/internal_shader_program.h` line 68 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGu2H5AjuS3xXBsq6H:-NXGu2H5AjuS3xXBsq6I:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_shader_program.h#L68)):*
> ```objc
>  should call Relink() on each of its ShaderProgram instances with its own
>  context bound during the clone's construction -- this guarantees an independent
>  namespace for its uniforms (at the cost of increased GPU memory. Relink() is
> ```

typo



_Suggestion:_
```objc
for its uniforms (at the cost of increased GPU memory).
```
___
*[`geometry/render_gl/internal_shader_program.h` line 197 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGuLRVEOHQuRm-nSLC:-NXGuLRVEOHQuRm-nSLD:b-dvtybx) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_shader_program.h#L197)):*
> This should be a precursor to destroying the %ShaderProgram.

Naively, this sounds like something we could enforce in the `ShaderProgram::~ShaderProgram()` code?  If that's not possible, possibly worth explaining why so the next reader doesn't also wonder why not.

Also this comment seems to contradict the comment that's already on our destructor.  Possibly that comment should cite this function as a reminder?

jwnimmer-tri (commented):
Checkpoint 2/3.

> Reviewed 1 of 12 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXHXoJJ9ZBlQReNWfqN:bgrr7e3)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 89 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHSKlJ0IQRWxwhAI4Q:-NXHSKlJ0IQRWxwhAI4R:b1y1xb0) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L89)):*
> ```cpp
>   // the actual asynchronous work has better odds of having the library
>   // invocations overlap.
>   contexts.reserve(kNumThreads);  // Presize; we can't move contexts.
> ```

BTW There is a one-liner for this, if you like:

```patch
--- a/geometry/render_gl/test/multithread_safety_test.cc
+++ b/geometry/render_gl/test/multithread_safety_test.cc
@@ -86,11 +86,8 @@ GTEST_TEST(ThreadSafetyTest, TextureLibrary) {
   // context so they share GPU memory. Do this cloning in its own loop so that
   // the actual asynchronous work has better odds of having the library
   // invocations overlap.
-  contexts.reserve(kNumThreads);  // Presize; we can't move contexts.
   OpenGlContext source;
-  for (int i = 0; i < kNumThreads; ++i) {
-    contexts.push_back(source);
-  }
+  contexts = std::vector<OpenGlContext>(kNumThreads, source);
 
   std::vector<std::future<void>> futures;
   for (int i = 0; i < kNumThreads; ++i) {
```
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 97 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHStcy7GGqFjBKB68i:-NXHStcy7GGqFjBKB68j:b-3gwskh) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L97)):*
> ```cpp
>   std::vector<std::future<void>> futures;
>   for (int i = 0; i < kNumThreads; ++i) {
>     // Each thread gets a copy of source, so they all share GPU state.
> ```

nit Given the similarly comment 10 lines prior, I'm not sure what more this comment is trying to tell me.
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 141 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHTSOe6-cfHlQlYm2V:-NXHTSOe6-cfHlQlYm2W:bl63jq0) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L141)):*
> ```cpp
>         log()->warn(
>             "Worker 0 should not be able to bind the source context in a "
>             "worker thread.");
> ```

BTW It might help clarify the situation if the warning message had a second sentence like.  "This indicates a deficiency in the underlying libGL implementation from your operating system." Someone just skimming the test output might not have all of the context of the implementation comment nearby, and misconstrue that some corrective action was required.
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 192 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHV36MBtr7w7OcQXd7:-NXHV36MBtr7w7OcQXd8:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L192)):*
> ```cpp
> }
> 
> /* We use Vector4<it> to compare the pixel values in an image with an expected
> ```

typo



_Suggestion:_
```cpp
 Vector4<int>
```
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 201 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHVGDf62XAhP0BB6QB:-NXHVGDf62XAhP0BB6QC:bdq4k) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L201)):*
> ```cpp
>   };
>   return Vector4<int>(to_T(rgba.r()), to_T(rgba.g()), to_T(rgba.b()),
>                       to_T(rgba.a()));
> ```

BTW I think there's a shorter (more canonical?) spelling available:

```c++
return rgba.rgba().unaryExpr([](double c) {
  return static_cast<int>(c * 255 + 0.5);
});
```
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 212 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHW3uAAc40kKFfL1pV:-NXHW3uAAc40kKFfL1pW:by7zxow) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L212)):*
> ```cpp
>  clean up work.
> 
>  This test confirms that clones of RenderEngineGl does enough clean up. We
> ```

typo grammar



_Suggestion:_
```cpp
clones of RenderEngineGl do enough
```
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 240 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHWI93EI0tFSm676Z9:-NXHWI941idThWDt0nit:b-le7ylb) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L240)):*
> ```cpp
>  render_gl/factory.h).
> 
>  To test the correctness of the rendered engine, we'll sample the color on
> ```

typo either "render engine" or "rendered image"



_Code quote:_
```cpp
rendered engine
```
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 252 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHWkxN8oHbK0RvqZQu:-NXHWkxN8oHbK0RvqZQv:b94xr5r) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L252)):*
> ```cpp
>      the dynamic geometries doesn't depend on the OpenGlContext, so other tests
>      on correct posing of dynamic geometries are sufficient. */
> GTEST_TEST(ThreadSafetyTest, RenderEngineGl) {
> ```

BTW Does this subsume `thread_test.cc` now?  It seems like they both call the engine directly, and inspect the image outputs.

I guess this test case doesn't have `AsyncMode` yet, so that's one difference.  If that's the only difference, then I'd be in the market for a TODO to add the `AsyncMode` param onto this test case and then delete the other test program.
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 342 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHXYo82ghuKdJJlaot:-NXHXYo92dfp1WrckEDM:b5nlyfb) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L342)):*
> ```cpp
>       } catch (std::exception& e) {
>         log()->error("Worker {} exception: {}", i, e.what());
>         rendered = false;
> ```

nit This bool is kind of a weird duck for me.  I would have expected instead ...

```c++
++num_errors;
return;
```

... and then lose the extra branching below.

jwnimmer-tri (commented):
:lgtm:

> Reviewed 2 of 12 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXHa3Ne3Twz7O0jotws:bnfp4nl)* status: 15 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/factory.h` line 33 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGszo4Ch7JqPUWe1Kg:-NXHZFW1-btwJ5cl8K21:b-4y6sd8) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/factory.h#L33)):*

BTW Just to make sure I completely understand -- let me check this sequence:

(1) Add one GL engine to the scene graph.
(2) Add the scene graph to the diagram.
(3) Launch several worker threads, each with their own dedicated diagram context.
(4) The worker thread removes some perception geometry (by deleting some perception properties) using the context-specific functions on SceneGraph.

From the comment, I believe that step (4) is unsafe.  Is that right?

(Assuming yes, there is no need to change any wording, just making sure.)
</blockquote></details>

Ah, the text at "A discussion on using RenderEngineGl in multiple threads." was much more clear, to me at least.  Possibly some of that could be copied here.
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 71 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHZWVpBEAFHrF-UM6T:-NXHZWVpBEAFHrF-UM6U:bipu7l4) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_render_engine_gl.h#L71)):*
> ```objc
> 
>  However, in sharing data across multiple OpenGl contexts not all data is
>  shareable, or *should* be.
> ```

nit clarity



_Suggestion:_
```objc
shareable, nor *should* be.
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 696 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXH_T5621y7oLC6KGRr:-NXH_T5621y7oLC6KGRs:b5xi06s) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_render_engine_gl.cc#L696)):*
> ```cpp
>     OpenGlContext::ClearCurrent();
>   });
>   clone->opengl_context_->MakeCurrent();
> ```

BTW https://en.cppreference.com/w/cpp/thread/scoped_lock or https://en.cppreference.com/w/cpp/thread/lock_guard is a thing.

If the functions on `OpenGlContext` were named `lock()` and `unlock()`, we could use a scoped_lock to make it current.  I'm not sure if the syntax would be better or worse.

SeanCurtis-TRI (commented):
Comments addressed.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXLVOTi5r77J1ZGE0eY:b83h6f9)* status: LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 696 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXH_T5621y7oLC6KGRr:-NXLPOHl-EGhW_xTqfoM:betekry) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_render_engine_gl.cc#L696)):*

BTW https://en.cppreference.com/w/cpp/thread/scoped_lock or https://en.cppreference.com/w/cpp/thread/lock_guard is a thing.

If the functions on `OpenGlContext` were named `lock()` and `unlock()`, we could use a scoped_lock to make it current.  I'm not sure if the syntax would be better or worse.
</blockquote></details>

I see what you mean. I'm not sure it would be better either. My mental model is:

1. The only time we require the context to be auto-unbound is upon construction.
2. In all other cases, leaving it bound is a good thing.
3. In those other cases, invocations like `opengl_context_.lock()` would not be helpful.

I could certainly double up the APIs to satisfy the `BasicLockable` spec as well as the other cases, but it's not clear if that has value. So, for all of the above, I'm deferring this one.
___
*[`geometry/render_gl/internal_shader_program.h` line 197 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGuLRVEOHQuRm-nSLC:-NXLRNu8ErbZ5VgM5stN:b-nenpol) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_shader_program.h#L197)):*

> This should be a precursor to destroying the %ShaderProgram.

Naively, this sounds like something we could enforce in the `ShaderProgram::~ShaderProgram()` code?  If that's not possible, possibly worth explaining why so the next reader doesn't also wonder why not.

Also this comment seems to contradict the comment that's already on our destructor.  Possibly that comment should cite this function as a reminder?
</blockquote></details>

I really wanted to do clean up in the destructor. But, because it can only be called with the appropriate context bound, I didn't feel the infrastructure is sufficiently wired to confidently rely on that. Maybe in the future.

I've massaged the language (here and above) to paint what is hopefully a more coherent and complete picture.
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 252 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXHWkxN8oHbK0RvqZQu:-NXLTg8F0y3fZitH_7dg:b-w73nf6) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/test/multithread_safety_test.cc#L252)):*

BTW Does this subsume `thread_test.cc` now?  It seems like they both call the engine directly, and inspect the image outputs.

I guess this test case doesn't have `AsyncMode` yet, so that's one difference.  If that's the only difference, then I'd be in the market for a TODO to add the `AsyncMode` param onto this test case and then delete the other test program.
</blockquote></details>

That was my belief as well. I've added a TODO.

jwnimmer-tri (commented):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gl/internal_shader_program.h` line 197 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXGuLRVEOHQuRm-nSLC:-NXLXJQOBPgsD_oXRjnY:b-q9qqdf) ([raw file](https://github.com/RobotLocomotion/drake/blob/5532299721879eb41e66cb27f5ecfaf984eaa855/geometry/render_gl/internal_shader_program.h#L197)):*

I really wanted to do clean up in the destructor. But, because it can only be called with the appropriate context bound, I didn't feel the infrastructure is sufficiently wired to confidently rely on that. Maybe in the future.

I've massaged the language (here and above) to paint what is hopefully a more coherent and complete picture.
</blockquote></details>

`:thumbsup:`
___
*[`geometry/render_gl/internal_shader_program.h` line 196 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXLXN4i0TDTOGnQa3CW:-NXLXN4jAUItyVqy7ZoB:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/06522f728639326a91cf5424811c39daa5e0fc8a/geometry/render_gl/internal_shader_program.h#L196)):*
> ```objc
>    be a precursor to destroying the %ShaderProgram.
> 
>    Ideally, this would be done in the constructor. However, the RenderEngineGl
> ```

typo



_Suggestion:_
```objc
destructor
```

SeanCurtis-TRI (commented):
+(release notes: feature) 

(I'll clean up the typo along with platform review comments.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXLckefEl2NufUnz9dA:bkm59lz)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Whew -- nice! Platform :lgtm_strong: 

> Reviewed 7 of 20 files at r4, 8 of 12 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXMy5sMDRg2hjfj2JC7:b-da77e1)* status: 2 unresolved discussions

___
*[`geometry/render_gl/internal_opengl_context.cc` line 83 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-NXMpXxdEp4Re9Mt1OKt:-NXMpXxdEp4Re9Mt1OKu:bl7mqb0) ([raw file](https://github.com/RobotLocomotion/drake/blob/06522f728639326a91cf5424811c39daa5e0fc8a/geometry/render_gl/internal_opengl_context.cc#L83)):*
> ```cpp
>   // If problems crop up in the future, this can/should be investigated.
>   static Display* g_display = (XInitThreads(), XOpenDisplay(0));
>   DRAKE_THROW_UNLESS(g_display != nullptr);
> ```

BTW did you want this null check every time display() is called? If you only want to check once at static initialization you could initialize g_display with a lambda that does the test before returning the `Display*`

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
Thanks for the fixup, Jeremy.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19426#-:-NXR5R0Y-20MyflvapBY:b-s9ixdk)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

PR #19436, opened 2023-05-19
[sensors] Add depth image conversion helpers

jwnimmer-tri:
Towards #19437.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)

### DIFF
--- a/bindings/pydrake/systems/sensors_py.cc
+++ b/bindings/pydrake/systems/sensors_py.cc
@@ -173,6 +173,13 @@ PYBIND11_MODULE(sensors, m) {
     type_visit(instantiation_visitor, PixelTypeList{});
   }
 
+  // Image conversion functions.
+  m  // BR
+      .def("ConvertDepth32FTo16U", &ConvertDepth32FTo16U, py::arg("input"),
+          py::arg("output"))
+      .def("ConvertDepth16UTo32F", &ConvertDepth16UTo32F, py::arg("input"),
+          py::arg("output"));
+
   using T = double;
 
   // Systems.
--- a/bindings/pydrake/systems/test/sensors_test.py
+++ b/bindings/pydrake/systems/test/sensors_test.py
@@ -185,6 +185,19 @@ class TestSensors(unittest.TestCase):
             np.testing.assert_array_equal(data, channel_default)
             np.testing.assert_array_equal(mutable_data, channel_default)
 
+    def test_depth_image_conversion(self):
+        foo = mut.ImageDepth32F(width=3, height=4)
+        bar = mut.ImageDepth16U()
+        mut.ConvertDepth32FTo16U(input=foo, output=bar)
+        self.assertEqual(bar.width(), 3)
+        self.assertEqual(bar.height(), 4)
+
+        foo = mut.ImageDepth16U(width=3, height=4)
+        bar = mut.ImageDepth32F()
+        mut.ConvertDepth16UTo32F(input=foo, output=bar)
+        self.assertEqual(bar.width(), 3)
+        self.assertEqual(bar.height(), 4)
+
     def test_camera_config(self):
         mut.CameraConfig()
         config = mut.CameraConfig(
--- a/geometry/render_gltf_client/internal_render_client.cc
+++ b/geometry/render_gltf_client/internal_render_client.cc
@@ -419,16 +419,7 @@ void RenderClient::LoadDepthImage(const std::string& path,
   } else if (image_data->GetScalarType() == VTK_TYPE_UINT16) {
     ImageDepth16U u16(width, height);
     image_exporter->Export(u16.at(0, 0));
-    // Convert from millimeters to meters.
-    for (int y = 0; y < height; ++y) {
-      for (int x = 0; x < width; ++x) {
-        const std::uint16_t millimeters = u16.at(x, y)[0];
-        const float meters = (millimeters == ImageDepth16U::Traits::kTooFar)
-                                 ? ImageDepth32F::Traits::kTooFar
-                                 : millimeters * 1e-3;
-        *depth_image_out->at(x, y) = meters;
-      }
-    }
+    ConvertDepth16UTo32F(u16, depth_image_out);
   } else {
     /* no cover */
     throw std::runtime_error("RenderClient: unsupported channel type");
--- a/systems/sensors/image.cc
+++ b/systems/sensors/image.cc
@@ -1,4 +1,68 @@
 #include "drake/systems/sensors/image.h"
 
-// For now, this is an empty .cc file that only serves to confirm
-// image.h is a stand-alone header.
+#include <algorithm>
+
+namespace drake {
+namespace systems {
+namespace sensors {
+
+namespace {
+/* Converts an input depth image to an output depth image using the given
+per-pixel conversion function. */
+template <typename InputImage, typename OutputImage, typename Func>
+void ConvertDepthPixelwise(const InputImage& input, OutputImage* output,
+                           Func func) {
+  const int width = input.width();
+  const int height = input.height();
+  if (!(output->width() == width && output->height() == height)) {
+    output->resize(width, height);
+  }
+  const int size = input.size();
+  if (size == 0) {
+    return;
+  }
+  const typename InputImage::T* const in = input.at(0, 0);
+  typename OutputImage::T* const out = output->at(0, 0);
+  for (int i = 0; i < size; ++i) {
+    out[i] = func(in[i]);
+  }
+}
+}  // namespace
+
+// N.B. The `output` images below are phrased as an output argument rather than
+// a return value to allow the caller to reuse the same storage from one Convert
+// call to the next.
+
+void ConvertDepth32FTo16U(const ImageDepth32F& input, ImageDepth16U* output) {
+  DRAKE_THROW_UNLESS(output != nullptr);
+  ConvertDepthPixelwise(
+      input, output, [](const ImageDepth32F::T pixel) -> ImageDepth16U::T {
+        // Start by scaling from meters to millimeters, keeping all math as
+        // 32-bit floats to mitigate roundoff errors.
+        float result = pixel * 1000.0f;
+        // Clamp too-far values (including +Inf and NaN) to the 16-bit constant.
+        result = std::min<float>(ImageDepth16U::Traits::kTooFar, result);
+        // Clamp too-close values (including -Inf) to the 16-bit constant.
+        result = std::max<float>(ImageDepth16U::Traits::kTooClose, result);
+        // Truncate towards zero.
+        return result;
+      });
+}
+
+void ConvertDepth16UTo32F(const ImageDepth16U& input, ImageDepth32F* output) {
+  DRAKE_THROW_UNLESS(output != nullptr);
+  ConvertDepthPixelwise(
+      input, output, [](const ImageDepth16U::T pixel) -> ImageDepth32F::T {
+        using InPixel = ImageDepth16U::Traits;
+        using OutPixel = ImageDepth32F::Traits;
+        // No special case is necessary for kTooClose.
+        static_assert(InPixel::kTooClose == 0 && OutPixel::kTooClose == 0);
+        // Scale from millimeters to meters, with kTooFar as a special case,
+        // using intermediate 64-bit doubles to mitigate roundoff errors.
+        return (pixel == InPixel::kTooFar) ? OutPixel::kTooFar : 0.001 * pixel;
+      });
+}
+
+}  // namespace sensors
+}  // namespace systems
+}  // namespace drake
--- a/systems/sensors/image.h
+++ b/systems/sensors/image.h
@@ -160,6 +160,18 @@ class Image {
   std::vector<T> data_;
 };
 
+/// Converts a single channel 32-bit float depth image with depths in meters to
+/// a single channel 16-bit unsigned int depth image with depths in millimeters.
+///
+/// Note that pixels that are not kTooFar in the input image may saturate to be
+/// kTooFar in the output image due to the smaller representable range.
+void ConvertDepth32FTo16U(const ImageDepth32F& input, ImageDepth16U* output);
+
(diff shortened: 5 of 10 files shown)


jwnimmer-tri:

SeanCurtis-TRI (commented):
:LGTM: With a grab bag of design and implementation questions.

> Reviewed 8 of 10 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-:-NW2z2s7-S5tYR-88t3O:bjazedn)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`-- commits` line 6 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2_C3T51dQlmqreOPX):*
nit typo



_Suggestion:_
```
it's
```
___
*[`systems/sensors/image.cc` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2aFxP7tYoX7y21NOj:-NW2aFxP7tYoX7y21NOk:b-p9e1xo) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L24)):*
> ```cpp
>     return;
>   }
>   const typename InputImage::T* in = input.at(0, 0);
> ```

BTW The contrast between `in` and `out` would read a bit better if the differences between them were only the significant ones. 



_Suggestion:_
```cpp
const typename InputImage::T* const in = input.at(0, 0);
```
___
*[`systems/sensors/image.cc` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2yHX8EcYDpH-aUlFb:-NW2yHX8EcYDpH-aUlFc:b-3yw3ah) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L32)):*
> ```cpp
> }  // namespace
> 
> void ConvertDepth32FTo16U(const ImageDepth32F& input, ImageDepth16U* output) {
> ```

BTW Is there a strong reason to prefer `void` with an output parameter instead of simply returning the other type? Images are very amenable to move semantics.
___
*[`systems/sensors/image.cc` line 38 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2kJ-N1xLQ_1D4ZtSj:-NW2kJ-N1xLQ_1D4ZtSk:b-i6wp8i) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L38)):*
> ```cpp
>         // Start by scaling from meters to millimeters, using 64-bit doubles to
>         // mitigate roundoff errors.
>         double result = pixel * 1000.0;
> ```

BTW I've been trying to craft some value of `pixel` for which this matters and I have failed. In principle, I get it. However, my mental model of how single-precision math gets done made me curious to see if this really mattered. My mental model consists of two principles:

1. On modern 64-bit processors, single-precision math is done with double-precision math and the answer is reduced in precision subsequently.
2. 1000.0 is perfectly represented by single-precision float.
3. 32-bit float has 24 bits of mantissa which is eventually reduced to 16 bits. It has more than enough bits to handle any precision errors out at what would be the 25th bit.

So, with those thoughts in mind, it's difficult to see how this `double` detour matters. And in the limited tests I did, I couldn't get it to matter.

Am I missing something?
___
*[`systems/sensors/image.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2oKD3BN0P-g7hkMQL:-NW2oKD3BN0P-g7hkMQM:b-3yap4r) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L41)):*
> ```cpp
>         // Clamp too-far values (including +Inf and NaN) to the 16-bit constant.
>         result = std::min<double>(ImageDepth16U::Traits::kTooFar, result);
>         // Clamp too-close values (including -Inf) to the 16-bit constant.
> ```

BTW Several questions:

1. Do we have a formal definition of what a "valid" depth image is?
    - The 16-bit depth image suggests that valid returns are strictly positive.
    - Presumably we would say the same thing about the 32-bit depth image.
2. Are these functions operating under the assumption that the input depth images are "valid" (in the sense of question 1)?
3. Do they get to have knowledge of how 'too close" is implemented (as the value zero)?

I'm sure you can see where this is going, if the answers to those questions are yes, then clamping at "too close" (i.e., `max(too_close, result)`) is redundant. If not, we should probably put some notes to future developers about this being isolated from those other concepts by design.

I note your conversion in the other direction does nail down the implementation of too close.
___
*[`systems/sensors/test/image_test.cc` line 222 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2xZcA9SDJxYXP1EOy:-NW2xZcBBhyOk9oUgjYk:b-8s2b13) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/test/image_test.cc#L222)):*
> ```cpp
>   EXPECT_EQ(image_out.width(), 0);
>   EXPECT_EQ(image_out.height(), 0);
> }
> ```

BTW To be pedantic, these tests would still pass if, instead of a for loop iterating over all pixels, you only ever converted the first pixel.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`systems/sensors/image.cc` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2yHX8EcYDpH-aUlFb:-NW37HEM58dL_ITJNH2c:b-esrpwi) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L32)):*

Yes.  In the systems framework, outputs are pre-allocated and we want to scribble into them.  Even with move semantics, we'd still be allocating here, copying into the context buffer elsewhere, and then freeing the allocation.

In my next push I'll add a comment somewhere explaining the thought process.
</blockquote></details>

As I saw it, the allocation was purely in the vector that contained the pxiel data. That would simply be passed around. But, meh.
___
*[`systems/sensors/image.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2oKD3BN0P-g7hkMQL:-NW37V6Z8J9YiOXkorvY:b-5qyw27) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L41)):*

> Do we have a formal definition of what a "valid" depth image is?

Yes, that's a correct definition.  It's not valid for a float image to have negatives, infinities, or nans.

> Are these functions operating under the assumption that the input depth images are "valid" (in the sense of question 1)?

No.  We should be behave sensibly in case the user provides us with an invalid image. Converting negative values to valid-range integer depths would not be doing the user any favors.
</blockquote></details>

And given the promotion to a public API, being more defensive seems good.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`systems/sensors/image.cc` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2yHX8EcYDpH-aUlFb:-NW38ZyY6qzo0n7bGCOL:b-fxkr78) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L32)):*

As I saw it, the allocation was purely in the vector that contained the pxiel data. That would simply be passed around. But, meh.
</blockquote></details>

Also...do we do conversion as part of the system framework? Or are we anticipating that conversion might be done in such a system and are leaving the door open? (And, I'm particularly thinking of python bindings here...that is the tricky domain, isn't it?)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)* status: 4 unresolved discussions, needs at least two assigned reviewers

___
*[`systems/sensors/image.cc` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2yHX8EcYDpH-aUlFb:-NW384WQ1pSXsZOx69fV:b-ji89qz) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L32)):*
> As I saw it, the allocation was purely in the vector that contained the pixel data. That would simply be passed around. But, meh.

Yes, it would be the vector where things allocated.

Still, as written there will be zero allocations (very cpu-cache-friendly), but with a return value there would always be one allocation (less friendly).

> Do we do conversion as part of the system framework?

Yes, in `RgbdSensor::CalcDepthImage16U`.

> (And, I'm particularly thinking of python bindings here...that is the tricky domain, isn't it?)

I don't understand what you're getting at.  The python bindings here are pretty vanilla and not particularly hazardous.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`systems/sensors/image.cc` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2yHX8EcYDpH-aUlFb:-NW3DoYvE4xy0LJIRb-C:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L32)):*

> As I saw it, the allocation was purely in the vector that contained the pixel data. That would simply be passed around. But, meh.

Yes, it would be the vector where things allocated.

Still, as written there will be zero allocations (very cpu-cache-friendly), but with a return value there would always be one allocation (less friendly).

> Do we do conversion as part of the system framework?

Yes, in `RgbdSensor::CalcDepthImage16U`.

> (And, I'm particularly thinking of python bindings here...that is the tricky domain, isn't it?)

I don't understand what you're getting at.  The python bindings here are pretty vanilla and not particularly hazardous.
</blockquote></details>

Done.
___
*[`systems/sensors/image.cc` line 38 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2kJ-N1xLQ_1D4ZtSj:-NW3JJVqAtI5iOVMZX5f:b-jb6su4) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L38)):*
Done.

The space of 32-bit floats in the 0m-66m range are few enough that it's plausible to just brute force search them.  I did that, and added a test case to cover an edge case I found.

As it turns out, to avoid roundoff errors in this direction it's actually better to stick with **floats**.

Here's the python reproducer for where the multiplication varies:

```
>>> np.float64(65.531)*np.float64(1000)
65531.00000000001
>>> np.float32(65.531)*np.float32(1000)
65531.0
>>> np.float32(65.531)*np.float64(1000)
65530.99822998047
```

In the last case, truncating to 65530 is unwanted.

jwnimmer-tri:
+@ggould-tri for platform review per schedule, please.

ggould-tri (commented):
:lgtm:

> Reviewed 7 of 10 files at r1, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-:-NW3m2S82I4KOKa2qWnl:bnfp4nl)* status: 2 unresolved discussions

___
*[`systems/sensors/test/image_test.cc` line 214 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW3lDbK2drkjtzxgJXJ:-NW3lDbK2drkjtzxgJXK:b-do2v7a) ([raw file](https://github.com/RobotLocomotion/drake/blob/446dc2a81a499386a508ed5f10c0c5378f02d12b/systems/sensors/test/image_test.cc#L214)):*
> ```cpp
>   using InPixel = ImageDepth32F::Traits;
>   using OutPixel = ImageDepth16U::Traits;
>   const std::vector<std::pair<float, uint16_t>> test_cases{
> ```

BTW:  Yay!

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19436)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),SeanCurtis-TRI(platform)

___
*[`systems/sensors/image.cc` line 38 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19436#-NW2kJ-N1xLQ_1D4ZtSj:-NW3p7dJAzwfhLrH_14I:bw0ango) ([raw file](https://github.com/RobotLocomotion/drake/blob/1a2df6dd063c57baac901d7918e04c6af4d35b31/systems/sensors/image.cc#L38)):*

Done.

The space of 32-bit floats in the 0m-66m range are few enough that it's plausible to just brute force search them.  I did that, and added a test case to cover an edge case I found.

As it turns out, to avoid roundoff errors in this direction it's actually better to stick with **floats**.

Here's the python reproducer for where the multiplication varies:

```
>>> np.float64(65.531)*np.float64(1000)
65531.00000000001
>>> np.float32(65.531)*np.float32(1000)
65531.0
>>> np.float32(65.531)*np.float64(1000)
65530.99822998047
```

In the last case, truncating to 65530 is unwanted.
</blockquote></details>

That's kind of cool. And disappointing.

PR #19444, opened 2023-05-21
[planning] Removed unnecessary variables in constraints

wrangelvid:
Many constraints, like the path continuity constraints or velocity bounds are enforced on just one control point. Thus it's the constraint matrices end up sparse, which can be mitigated by dropping the unnecssary columns and variables in those constraints.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)

### DIFF
--- a/planning/trajectory_optimization/gcs_trajectory_optimization.cc
+++ b/planning/trajectory_optimization/gcs_trajectory_optimization.cc
@@ -1,6 +1,7 @@
 #include "drake/planning/trajectory_optimization/gcs_trajectory_optimization.h"
 
 #include <limits>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 
@@ -30,6 +31,7 @@ using geometry::optimization::GraphOfConvexSetsOptions;
 using geometry::optimization::HPolyhedron;
 using geometry::optimization::Point;
 using solvers::Binding;
+using solvers::ConcatenateVariableRefList;
 using solvers::Constraint;
 using solvers::Cost;
 using solvers::L2NormCost;
@@ -49,6 +51,60 @@ using VertexId = GraphOfConvexSets::VertexId;
 using EdgeId = GraphOfConvexSets::EdgeId;
 
 const double kInf = std::numeric_limits<double>::infinity();
+
+namespace {
+using VectorXb = Eigen::Matrix<bool, 1, Eigen::Dynamic>;
+
+// Given a list of matrices, return the matrices with every column where all
+// of the matrices are zero in that column, along with a boolean vector
+// indicating which columns were preserved (true) or removed (false).
+std::tuple<std::vector<MatrixXd>, VectorXb> CondenseToNonzeroColumns(
+    std::vector<MatrixXd> matrices) {
+  // Validate inputs.
+  DRAKE_DEMAND(matrices.size() > 0);
+  const int num_cols = matrices[0].cols();
+  for (const MatrixXd& matrix : matrices) {
+    DRAKE_DEMAND(matrix.cols() == num_cols);
+  }
+
+  // Find non-zero columns.
+  VectorXb nonzero_cols_mask = VectorXb::Constant(num_cols, false);
+  for (const MatrixXd& matrix : matrices) {
+    nonzero_cols_mask += matrix.cast<bool>().colwise().any();
+  }
+  const int nonzero_cols_count = nonzero_cols_mask.count();
+
+  // Create the output, copying only the non-zero columns.
+  std::vector<MatrixXd> condensed_matrices;
+  for (const MatrixXd& matrix : matrices) {
+    MatrixXd& condensed_matrix =
+        condensed_matrices.emplace_back(matrix.rows(), nonzero_cols_count);
+    int condensed_col = 0;
+    for (int orig_col = 0; orig_col < matrix.cols(); ++orig_col) {
+      if (nonzero_cols_mask(orig_col)) {
+        condensed_matrix.col(condensed_col) = matrix.col(orig_col);
+        condensed_col++;
+      }
+    }
+  }
+  return std::make_tuple(condensed_matrices, nonzero_cols_mask);
+}
+
+// Filters variables given a vector of variables along with a boolean vector
+// indicating which rows were preserved (true) or removed (false).
+VectorX<symbolic::Variable> FilterVariables(
+    const VectorX<symbolic::Variable>& vars,
+    const VectorXb& nonzero_cols_mask) {
+  VectorX<symbolic::Variable> vars_dense(nonzero_cols_mask.count());
+  int row = 0;
+  for (int i = 0; i < vars.size(); ++i) {
+    if (nonzero_cols_mask(i)) {
+      vars_dense(row++) = vars(i);
+    }
+  }
+  return vars_dense;
+}
+}  // namespace
 
 Subgraph::Subgraph(
     const ConvexSets& regions,
@@ -109,12 +165,14 @@ Subgraph::Subgraph(
       u_r_trajectory_.control_points().col(order);
   MatrixXd M(num_positions(), edge_vars.size());
   DecomposeLinearExpressions(path_continuity_error, edge_vars, &M);
-  // TODO(wrangelvid) The matrix M might be sparse, so we could drop columns
-  // here.
+  // Condense M to only keep non-zero columns.
+  const auto& [condensed_matrices, nonzero_cols_mask] =
+      CondenseToNonzeroColumns({M});
+  MatrixXd M_dense = condensed_matrices[0];
 
   const auto path_continuity_constraint =
       std::make_shared<LinearEqualityConstraint>(
-          M, VectorXd::Zero(num_positions()));
+          M_dense, VectorXd::Zero(num_positions()));
 
   // Add Regions with time scaling set.
   for (size_t i = 0; i < regions_.size(); ++i) {
@@ -142,8 +200,10 @@ Subgraph::Subgraph(
     edges_.emplace_back(uv_edge);
 
     // Add path continuity constraints.
-    uv_edge->AddConstraint(
-        Binding<Constraint>(path_continuity_constraint, {u.x(), v.x()}));
+    uv_edge->AddConstraint(Binding<Constraint>(
+        path_continuity_constraint,
+        FilterVariables(ConcatenateVariableRefList({u.x(), v.x()}),
+                        nonzero_cols_mask)));
   }
 }
 
@@ -187,15 +247,18 @@ void Subgraph::AddPathLengthCost(const MatrixXd& weight_matrix) {
   for (int i = 0; i < u_rdot_control.cols(); ++i) {
     MatrixXd M(num_positions(), u_vars_.size());
     DecomposeLinearExpressions(u_rdot_control.col(i) / order(), u_vars_, &M);
-    // TODO(wrangelvid) The matrix M might be sparse, so we could drop columns
-    // here.
+    // Condense M to only keep non-zero columns.
+    const auto& [condensed_matrices, nonzero_cols_mask] =
+        CondenseToNonzeroColumns({M});
+    MatrixXd M_dense = condensed_matrices[0];
 
     const auto path_length_cost = std::make_shared<L2NormCost>(
-        weight_matrix * M, VectorXd::Zero(num_positions()));
+        weight_matrix * M_dense, VectorXd::Zero(num_positions()));
 
     for (Vertex* v : vertices_) {
       // The duration variable is the last element of the vertex.
-      v->AddCost(Binding<L2NormCost>(path_length_cost, v->x()));
+      v->AddCost(Binding<L2NormCost>(
+          path_length_cost, FilterVariables(v->x(), nonzero_cols_mask)));
     }
   }
 }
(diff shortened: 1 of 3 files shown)


RussTedrake (commented):
:lgtm:  nice!

With that much repeated (almost identical) code, it really begs for pulling the basic idea out into a subroutine.  But I won't block on it.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NW246t50p7QnwuMnZLP:b-mot1m7)* status: LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @ggould-tri and @wrangelvid)

wrangelvid (commented):
I agree. I am open to feedback/suggestions on how to make it nicer.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NW2Dzj29lT9kXPDHB9d:b-ikf5y7)* status: LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @ggould-tri and @wrangelvid)

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @wrangelvid)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW2WTHr2dlTjTTtnKtJ:-NW2WTHr2dlTjTTtnKtK:bx5qysl) (no related file):*
Okay, as Russ said, we're way over Rule Of Three on the idiom here.
I made a quick pass at factoring to a common function; here's what I put at the top:

```
namespace {

using VectorXb = Eigen::Matrix<bool, 1, Eigen::Dynamic>;

// Given a list of matrices, return the matrices with every column where all
// of the matrices are zero in that column, along with a boolean vector
// indicating which columns were preserved (true) or removed (false).
std::tuple<std::vector<MatrixXd>, VectorXb> CondenseToNonzeroColumns(
    std::vector<MatrixXd> matrices) {
  // Validate inputs.
  DRAKE_DEMAND(matrices.size() > 0);
  const int num_cols = matrices[0].cols();
  for (const MatrixXd& matrix : matrices) {
    DRAKE_DEMAND(matrix.cols() == num_cols);
  }

  // Find non-zero columns.
  VectorXb nonzero_cols_mask = VectorXb::Constant(num_cols, false);
  for (const MatrixXd& matrix : matrices) {
    nonzero_cols_mask += matrix.cast<bool>().colwise().any();
  }
  const int nonzero_cols_count = nonzero_cols_mask.count();

  // Create the output, copying only the non-zero columns.
  std::vector<MatrixXd> condensed_matrices;
  for (const MatrixXd& matrix : matrices) {
    MatrixXd& condensed_matrix =
        condensed_matrices.emplace_back(matrix.rows(), nonzero_cols_count);
    int condensed_col_idx = 0;
    for (int orig_col_idx = 0; orig_col_idx < matrix.cols(); ++orig_col_idx) {
      if (nonzero_cols_mask(orig_col_idx)) {
        condensed_matrix.col(condensed_col_idx) = matrix.col(orig_col_idx);
        condensed_col_idx++;
      }
    }
  }
  return std::make_tuple(condensed_matrices, nonzero_cols_mask);
}

}  // namespace
```

and then each callsite became something like:

```
    const auto& [condensed_matrices, nonzero_cols_mask] =
        CondenseToNonzeroColumns({M, b});
    MatrixXd M_dense = condensed_matrices[0];
    MatrixXd b_dense = condensed_matrices[1];
```

I think an approach along these lines will improve readability and save a few lines.

___

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @ggould-tri and @wrangelvid)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW2WTHr2dlTjTTtnKtJ:-NW4cfhNBlJtCb61dv3b:bj3m6wv) (no related file):*

(Obviously that's not quite polished code, but I think it makes the point.  You'd probably want to replace the `orig_col_idx` loop there with a range-for and if efficiency is crucial maybe combine a couple of the serial fors and check that it's RVO'ing instead of copying everything twice.)
</blockquote></details>

Great suggestion! I did something similar for filtering the variables. 

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform), missing label for release notes (waiting on @ggould-tri and @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW4dVMn2s26FnU3z-Wc:-NW4dVMn2s26FnU3z-Wd:b-psdhm7) ([raw file](https://github.com/RobotLocomotion/drake/blob/23eab2429eb12cded8ad2c797555f30fe10d4b57/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L81)):*
nit  From [#General_Naming_Rules](https://drake.mit.edu/styleguide/cppguide.html#General_Naming_Rules) we have:

> Do not abbreviate by deleting letters within a word.

Ditto anywhere `idx` is used.  Another alternative might be just `condensed_col` without `_idx` nor `_index` suffix.



_Suggestion:_
```cpp
    int condensed_col_index = 0;
```

ggould-tri (commented):
:lgtm_strong: this is much better.  One trivial syntax item.

I don't know if you have merge permissions for this repository; if I am inattentive and this sits in mergeable state for too long kick me and I'll punch the button.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NW7mKlJ2ozbq19sqRDO:bevz04x)* status: 2 unresolved discussions, missing label for release notes (waiting on @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 57 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW7lrxg2mWTYful0273:-NW7lrxg2mWTYful0274:b-2vyhc5) ([raw file](https://github.com/RobotLocomotion/drake/blob/74465427de2d1cc10a7601427f2349dc7e15849d/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L57)):*
> ```cpp
> namespace {
> using VectorXb = Eigen::Matrix<bool, 1, Eigen::Dynamic>;
> // Given a list of matrices, return the matrices with every column where all
> ```

nit:  We generally put a newline before an interface comment.



_Suggestion:_
```cpp
using VectorXb = Eigen::Matrix<bool, 1, Eigen::Dynamic>;

// Given a list of matrices, return the matrices with every column where all
```
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 94 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW7mBdY7pkAVfprPjh2:-NW7mBdY7pkAVfprPjh3:b9n14ja) ([raw file](https://github.com/RobotLocomotion/drake/blob/74465427de2d1cc10a7601427f2349dc7e15849d/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L94)):*
> ```cpp
> // Filters variables given a vector of variables along with a boolean vector
> // indicating which rows were preserved (true) or removed (false).
> VectorX<symbolic::Variable> FilterVariables(
> ```

BTW:  Nice, I hadn't thought to factor that idiom as well :+1:

ggould-tri:
It's an implementation-only change; do you think `gcs_trajectory_optimization.h` is inadequately covered?  It looked pretty well-covered to me, but possibly I am missing your area of concern.  Or are you looking for a test of the helper function itself?

jwnimmer-tri:
One prong of code correctness here is whether the planned trajectories are still the optimal paths.  There's a fair chance the tests already cover that. I haven't looked, but it sounds like you have and I believe you that it's already well-covered.

The second prong of code correctness here is whether the implementation is fast.  (This seems to me like a performance improvement PR?)  Ideally, that would somehow be guarded by a regression test.  How do we know that this is *actually* reducing the size of the programs we're solving?  Maybe there's a typo that makes this all just a fancy no-op.  It should be possible to instrument whether this change is reducing the facet(s) of the MathematicalProgram that we care about.

jwnimmer-tri:
In any case my main point is that testing should be *discussed*.  Adding new code that doesn't require a unit test is unusual, so if it's truly the case that we don't need a test then something in the discussion timeline should explain that reasoning, for posterity.

wrangelvid (commented):
Since the mathematical program is not exposed, it will be hard to check for the reduction in program size. A test comparing runtimes between PR's would be interesting, although I don't know how to implement that.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NW8Tzyw2YO78nhD21YG:bol138v)* status: missing label for release notes (waiting on @wrangelvid)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 57 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NW7lrxg2mWTYful0273:-NW8UdQDBMuCa5j_9LWH:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/74465427de2d1cc10a7601427f2349dc7e15849d/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L57)):*

nit:  We generally put a newline before an interface comment.
</blockquote></details>

Done.

jwnimmer-tri:
The main question is -- do we know this code is correct, and/or do we need to guard against regressions? If no, then it's moot. If yes, we need to find a way to refactor for testing. In the extreme, we pass SolverOptions with kPrintFile and scrape the log file to find out about the program's attributes. I'll bet there is a much easier refactor for testing than that, though. Grant should be able help. First decide how possible bugs or regressions are as we have a dozen other people editing this file over the next few years.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: missing label for release notes (waiting on @wrangelvid)

wrangelvid (commented):
The test cases check the returned trajectory and would fail if the this change would break anything. We could test the condense and variable filter function to check if it does anything.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NWD56pSANZo4aYiB-NG:bpsp41k)* status: missing label for release notes (waiting on @wrangelvid)

jwnimmer-tri:
I agree that the current tests guard against correctness regressions.  The question is whether we want to guard against performance regressions.  Here's the testing I suggest:

After a call to each of the public functions that is improved in this PR (e.g., `Subgraph::AddPathLengthCost`), have the unit test check an upper bound (e.g., `EXPECT_LE`) on the total number of variables in the costs and/or constraints of the GCS program.

I am imagining a new function like `int GcsTrajectoryOptimization::GetBindingComplexity() const` to return that information.  It would calculate something like:

```c++
int result = 0;
for (v : gcs.Vertices()) {
  for (c : v.GetCosts()) { result += c.GetNumElements(); }
  for (c : v.GetConstraints()) { result += c.GetNumElements(); }
}
for (e : gcs.Edges()) {
  for (c : e.GetCosts()) { result += c.GetNumElements(); }
  for (c : e.GetConstraints()) { result += c.GetNumElements(); }
}
return result;
```

As I understand this PR, the goal is to reduce the width of some of those costs and/or constraints.  I would expect that the result of that tally should be starkly different between the current code on master and the code in this PR.

If the function would not be interested to user, its documentation could say "(Internal use only)" or some other phrase to dissuade users from calling it.  I could maybe see it being useful in certain cases though, in case users want to get a feel for how sparse their program is.

WDYT?

jwnimmer-tri:
Or maybe the function could return `pair<int, int>` with the aggregate cost and aggregate constraint complexity, respectively.  That would be somewhat more precise of an answer, for both unit testing and users poking at it.

ggould-tri (commented):
Hi, apologies for not getting back to you more quickly on this -- I've been busy this week on non-Drake stuff.  I think Jeremy's answer (bounding some naive measure of complexity in the test suite) is great and probably fairly straightforward to implement, so I'd recommend it as well.

I'd go so far as to say it probably should be a public function (albeit documented as advanced usage) since I expect some of the academic users might be interested in it, e.g. for lecture materials for Russ's manipulation course.  But that's just a superficial opinion.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NWIrY5g7bujyRCydaUv:b-sa2vz3)* status: missing label for release notes (waiting on @wrangelvid)

wrangelvid (commented):
Thank you guys! I'll get to it by the end of Monday. Graduation has been busy.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NX885n_5ggJKOjArSoL:b5rjkhe)* status: missing label for release notes (waiting on @wrangelvid)

wrangelvid (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NXCqa5c1srQAJ8YKm2R:bnfp4nl)* status: missing label for release notes (waiting on @wrangelvid)

ggould-tri (commented):
+(release notes: none) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-:-NXGMjUI-R_PZNyafoXf:b-safpmt)* status: 2 unresolved discussions

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: 2 unresolved discussions

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.h` line 388 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NXFISiaBZhrWrfwMuBA:-NXHixsE7A-a3j4m-26U:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f833b82df1c9230af27fb4dfd56feef038318ab0/planning/trajectory_optimization/gcs_trajectory_optimization.h#L388)):*

This comment is insufficient for the public API.  I would have no idea what this means.  Even the name of the function is off-putting to me.  Why not name it total number of variables in costs and constraints, or something more specific?
</blockquote></details>

Done.
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 813 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19444#-NXFIcFe47WBRGCxMhDA:-NXHiwhnCQJycHbDDc20:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f833b82df1c9230af27fb4dfd56feef038318ab0/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L813)):*

consider moving this to the public documentaiton.
</blockquote></details>

Done.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: 2 unresolved discussions

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19444)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),RussTedrake(platform)

ggould-tri:
Thank you to @wrangelvid for bearing with this long and difficult process, and to @RussTedrake for pushing this over the finish line.

PR #18156, opened 2022-10-21
[solvers] Deprecate dReal and IBEX for removal

jwnimmer-tri:
Deprecate the classes DrealSolver, IbexSolver, and their nested helper classes.
Deprecate the workspace externals cds, dreal, ibex, and picosat.

----

This pull request is a proposal of an idea bouncing around for a few months.  I'll leave it open for comments until **2022-10-31** at which point we'll plan to merge it (unless we hear any objections).  Feel free to reply in this issue, or Drake slack, or send me an email.

There are two thoughts behind this proposal.

(1) It is becoming increasingly difficult to keep up with the build and release tasks involved with distributing dReal and IBEX (#15872, #15970, #16882, #17231, #17869, etc.).

The code and its dependencies offers a few unique challenges -- LGPL licensing, x86 assembly code, and a very finicky version compatibility matrix across patch-level releases for multiple packages.  As we broaden our support for ARM64-compatible binary releases, and fully-linked libraries (e.g., for PyPI), those all have been posing challenges for the build team.  We can probably overcome all of those challenges by throwing more time and money the the problem.  However, given the next point below, my current hypothesis is that it's not worth it.

(2) Talking to users who have tried to use Drake's IBEX and/or dReal support, the general consensus is that it's implausibly slow to run for anything larger than toy-sized problems. (Sometimes, even toy-sized problems were unreasonably slow.)  We don't think that anyone has successfully used Drake's wrappers of those tools for any meaningful research problem.

I think there's probably still a great frontier for research here, but I think that work can happen external to the Drake source tree.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/18156)

PR #19450, opened 2023-05-22
Rename SolutionResult::kUnknownError to kSolverSpecificError

RussTedrake:
I had a student trip on this; it was incredibly frustrating to be working on a mathematical program only to have the solver apparently return "kUnknownError".  What they needed, instead, was a hint that they should look at the solver details, etc, to figure out the error.

+@hongkai-dai for feature review, please (and confirmation of the renaming idea).

fyi @jwnimmer-tri .  I spent some time trying to deprecate this more properly (both in c++ and python), but I couldn't find any of the proper affordances for deprecating an attribute of an enum.  I'm inclined to just make the breaking change (I suspect that it won't actually break anyone, but of course can't be sure).  wdyt?

### DIFF
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -1564,8 +1564,8 @@ for every column of ``prog_var_vals``. )""")
           doc.SolutionResult.kInfeasibleConstraints.doc)
       .value("kUnbounded", SolutionResult::kUnbounded,
           doc.SolutionResult.kUnbounded.doc)
-      .value("kUnknownError", SolutionResult::kUnknownError,
-          doc.SolutionResult.kUnknownError.doc)
+      .value("kSolverSpecificError", SolutionResult::kSolverSpecificError,
+          doc.SolutionResult.kSolverSpecificError.doc)
       .value("kInfeasibleOrUnbounded", SolutionResult::kInfeasibleOrUnbounded,
           doc.SolutionResult.kInfeasibleOrUnbounded.doc)
       .value("kIterationLimit", SolutionResult::kIterationLimit,
--- a/bindings/pydrake/solvers/test/snopt_solver_test.py
+++ b/bindings/pydrake/solvers/test/snopt_solver_test.py
@@ -2,10 +2,12 @@ import unittest
 
 import numpy as np
 
+from pydrake.autodiffutils import AutoDiffXd
 from pydrake.common.test_utilities import numpy_compare
 from pydrake.solvers import (
     MathematicalProgram,
     SnoptSolver,
+    SolutionResult,
     SolverType,
 )
 
@@ -35,3 +37,17 @@ class TestSnoptSolver(unittest.TestCase):
                 result.get_solver_details().F, np.array([0, 1.]))
             np.testing.assert_allclose(
                 result.get_solver_details().Fmul, np.array([0, 1.]))
+
+    def test_solver_specific_error(self):
+        # Intentionally write a constraint with incorrect gradients.
+        def my_constraint(x):
+            return [AutoDiffXd(np.sin(x[0].value()), np.ones(1))]
+        prog = MathematicalProgram()
+        x = prog.NewContinuousVariables(1, "x")
+        prog.AddConstraint(my_constraint, [1], [1], x)
+        solver = SnoptSolver()
+        result = solver.Solve(prog)
+        self.assertEqual(result.is_success(), False)
+        self.assertEqual(result.get_solution_result(),
+                         SolutionResult.kSolverSpecificError)
+        self.assertEqual(result.get_solver_details().info, 41)
--- a/solvers/branch_and_bound.cc
+++ b/solvers/branch_and_bound.cc
@@ -37,7 +37,7 @@ MixedIntegerBranchAndBoundNode::MixedIntegerBranchAndBoundNode(
       fixed_binary_variable_{},
       fixed_binary_value_{-1},
       remaining_binary_variables_{binary_variables},
-      solution_result_{SolutionResult::kUnknownError},
+      solution_result_{SolutionResult::kSolverSpecificError},
       optimal_solution_is_integral_{OptimalSolutionIsIntegral::kUnknown},
       solver_id_{solver_id} {
   // Check if there are still binary variables.
--- a/solvers/clp_solver.cc
+++ b/solvers/clp_solver.cc
@@ -182,10 +182,10 @@ void SetSolution(
   SetBoundingBoxConstraintDualSolution(column_dual_sol,
                                        bb_con_dual_variable_indices, result);
   solver_details.status = model.status();
-  SolutionResult solution_result{SolutionResult::kUnknownError};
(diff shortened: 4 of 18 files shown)


jwnimmer-tri:
For a deprecation example of an enum, see here:
https://github.com/RobotLocomotion/drake/pull/18156/files#diff-670f8e85acf497104f62087e96cbfbb6cfefa876d65b66378cf1e0b871974ad8

If it helps, I could push the deprecation markings atop this PR.

RussTedrake (commented):
That would be amazing.  Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-:-NW2aCZs07ZWA2F2qFgi:b-nkkw5z)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai and @RussTedrake)

jwnimmer-tri:
Pushed.

RussTedrake (commented):
@hongkai-dai did you mean to lgtm?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-:-NW3VSog4RLfQI3h8iQI:b1uh8u3)* status: 1 unresolved discussion, LGTM missing from assignees hongkai-dai,ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri, @hongkai-dai, and @RussTedrake)

___
*[`solvers/solution_result.h` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-NW2m_3x-RuPqB1BXJxQ:-NW3VQpKDcA4tzC5lR6R:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/ddcb1c3f86c8ffd2c49e287f6b0671bdd7ed7b2d/solvers/solution_result.h#L19)):*

Maybe more specific and say "MathematicalProgramResult::get_solver_details()"?
</blockquote></details>

Done.

hongkai-dai (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-:-NW3VudTAoM8hZuZvYSH:bg7igkv)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri and @RussTedrake)

ggould-tri (commented):
See below; while I agree that we do need `kSolverSpecificError` and it is the right thing to say 99% of the time that we have been saying `kUnknownError`, I think the two are distinct in the case of an error that prevents any solver from being used, and I would have retained the old value instead of deprecating it.
Is there a better way to capture the default case prior to solver selection?  Perhaps a new `kSolverSelectionError` value?

> Reviewed 16 of 18 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-:-NW3hu9tBoLR-1YpxNQB:b-4ii2ev)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri and @RussTedrake)

___
*[`solvers/mathematical_program_result.cc` line 18 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-NW3hwcx1B1DMxKou75p:-NW3hwcx1B1DMxKou75q:bd54jai) ([raw file](https://github.com/RobotLocomotion/drake/blob/e3a56059f76a3e4a843f90c4425635045bb35555/solvers/mathematical_program_result.cc#L18)):*
> ```cpp
> MathematicalProgramResult::MathematicalProgramResult()
>     : decision_variable_index_{},
>       solution_result_{SolutionResult::kSolverSpecificError},
> ```

I'm less sure of this one.  How can we have a solver-specific error before we even have a solver?
`kUnknownError` seems like a reasonable spelling of "I failed before I could find a solver", but if it needs to go then possible `kInvalidInput` covers this better (the input did not admit any solver)?  Or perhaps a new error value `kNoSolver`?
Presumably this leaks through approximately zero percent of the time, so it doesn't matter very much.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri and @RussTedrake)

___
*[`solvers/mathematical_program_result.cc` line 18 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-NW3hwcx1B1DMxKou75p:-NW4Al-x6nmb25Z5nAm9:b-jd6jrk) ([raw file](https://github.com/RobotLocomotion/drake/blob/e3a56059f76a3e4a843f90c4425635045bb35555/solvers/mathematical_program_result.cc#L18)):*

Good point.  I agree that `kSolverSpecificError` isn't great here, but definitely don't think it should be `kUnknownError.` I would said `kInvalidInput` or `kSolutionResultNotSet` or something similar?  `kNoSolver` doesn't seem quite right to me, either.  
</blockquote></details>

BTW I like `kSolutionResultNotSet`.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @ggould-tri and @RussTedrake)

___
*[`solvers/mathematical_program_result.cc` line 18 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-NW3hwcx1B1DMxKou75p:-NW6rkBj2LK9x-XJVXDn:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/e3a56059f76a3e4a843f90c4425635045bb35555/solvers/mathematical_program_result.cc#L18)):*

BTW I like `kSolutionResultNotSet`.
</blockquote></details>

Done.

ggould-tri (commented):
:lgtm:; I expect we'll need to change the documentation to be a bit more user-centric rather than implementation-centric in the future, but that's something we'll discover if/when users hit this error and we figure out what the cases are in practice.

> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19450#-:-NW7k2zoCvZdwDVFXKkH:b-ggrd8f)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),hongkai-dai

jwnimmer-tri:
This PR necessitates some adjustments in Anzu.  When TRI people deprecate things, we ask that they either fix Anzu versus the deprecation or recruit a volunteer to do it who is more familiar with Anzu.  @ggould-tri would you like to take it?  Or as a backstop, you can volunteer me for it.

jwnimmer-tri:
I'll interpret that as falling onto my plate.  I'll PR to Anzu momentarily.

Issue #19446, opened 2023-05-21
Drake release against fmt v10.0.0 on macOS homebrew


jwnimmer-tri:
Now that https://github.com/Homebrew/homebrew-core/pull/130646 has merged, we need another stable release of Drake to keep up.

jwnimmer-tri:
Beyond the build errors, I also need to find a solution for https://github.com/fmtlib/fmt/issues/3456.

jwnimmer-tri:
- macOS users with fmt 9 should use stable release 1.16.0 or 20230520 (or anything prior).
- macOS users with fmt 10 should use stable release 1.17.0 or nightly build 20230523 (or anything newer).

This not apply to `pip` (i.e., wheel) users.  The wheels are insensitive to most details of Homebrew.

PR #19451, opened 2023-05-22
Fix build errors vs fmt 10

jwnimmer-tri:
Towards #19446.

This only fixes the build errors.  I'm going to work on the test failures separately.

+@ggould-tri for both reviews per schedule, please.

\CC @SeanCurtis-TRI FYI

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19451)

### DIFF
--- a/geometry/proximity/test/characterization_utilities.h
+++ b/geometry/proximity/test/characterization_utilities.h
@@ -20,6 +20,7 @@
 
 #include "drake/common/default_scalars.h"
 #include "drake/common/drake_export.h"
+#include "drake/common/fmt_ostream.h"
 #include "drake/common/unused.h"
 #include "drake/geometry/proximity/collision_filter.h"
 #include "drake/geometry/shape_specification.h"
@@ -440,6 +441,12 @@ class CharacterizeResultTest : public ::testing::Test {
 }  // namespace internal
 }  // namespace geometry
 }  // namespace drake
+
+namespace fmt {
+template <>
+struct formatter<drake::geometry::internal::GeometryType>
+    : drake::ostream_formatter {};
+}  // namespace fmt
 
 DRAKE_DECLARE_CLASS_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_NONSYMBOLIC_SCALARS(
   class ::drake::geometry::internal::CharacterizeResultTest)
--- a/geometry/render_gltf_client/internal_http_service_curl.cc
+++ b/geometry/render_gltf_client/internal_http_service_curl.cc
@@ -13,6 +13,7 @@
 #include <fmt/format.h>
 
 #include "drake/common/drake_assert.h"
+#include "drake/common/fmt_ostream.h"
 #include "drake/common/never_destroyed.h"
 #include "drake/common/text_logging.h"
 #include "drake/common/unused.h"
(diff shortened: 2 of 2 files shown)


ggould-tri (commented):
:lgtm:; one stupid standards-lawyer quibble.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19451#-:-NW2uSI8-ds4zyB52zDF:btf87ks)* status: 2 unresolved discussions

___
*[`geometry/render_gltf_client/internal_http_service_curl.cc` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19451#-NW2t6bq465joHqFhF16:-NW2t6bq465joHqFhF17:bk77xx0) ([raw file](https://github.com/RobotLocomotion/drake/blob/d29b1dcc576735d7ba67874d20c328d104f9f393/geometry/render_gltf_client/internal_http_service_curl.cc#L86)):*
> ```cpp
>   else if (type == CURLINFO_END)
>     return "CURLINFO_END";
>   return fmt::format("UNKNOWN_CURLINFO_TYPE={}", static_cast<uint64_t>(type));
> ```

BTW:
Is it really going to be necessary to do this any time we print a non-class enum?
___
*[`geometry/render_gltf_client/internal_http_service_curl.cc` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19451#-NW2uQ5b0SEUAgbthkco:-NW2uQ5b0SEUAgbthkcp:bdixj5h) ([raw file](https://github.com/RobotLocomotion/drake/blob/d29b1dcc576735d7ba67874d20c328d104f9f393/geometry/render_gltf_client/internal_http_service_curl.cc#L86)):*
> ```cpp
>   else if (type == CURLINFO_END)
>     return "CURLINFO_END";
>   return fmt::format("UNKNOWN_CURLINFO_TYPE={}", static_cast<uint64_t>(type));
> ```


minor:
Also, if we're going to be asserting a type bound for `curl_infotype`'s storage, why `uint64_t`?  It's going to `int` in practice, isn't it ("the value is convertible to the first type from the following list able to hold their entire value range:  int...")?  Bumping it to unsigned is more ABI-brittle than needed.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19451)* status: :shipit: complete! all discussions resolved, LGTM from assignee ggould-tri(platform)

___
*[`geometry/render_gltf_client/internal_http_service_curl.cc` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19451#-NW2t6bq465joHqFhF16:-NW2wXH58m2-jnwYB20H:b-j2jay4) ([raw file](https://github.com/RobotLocomotion/drake/blob/d29b1dcc576735d7ba67874d20c328d104f9f393/geometry/render_gltf_client/internal_http_service_curl.cc#L86)):*

BTW:
Is it really going to be necessary to do this any time we print a non-class enum?
</blockquote></details>

In writing my reply here, I realized that `fmt_streamed` is probably a nicer spelling.  For third-party enums, yes either `fmt_streamed` or casting will be required. (For Drake enums, we can define a formatter instead of streaming them.)
___
*[`geometry/render_gltf_client/internal_http_service_curl.cc` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19451#-NW2uQ5b0SEUAgbthkco:-NW2wdk28NWpwCojAm2W:b-ep0bc4) ([raw file](https://github.com/RobotLocomotion/drake/blob/d29b1dcc576735d7ba67874d20c328d104f9f393/geometry/render_gltf_client/internal_http_service_curl.cc#L86)):*


minor:
Also, if we're going to be asserting a type bound for `curl_infotype`'s storage, why `uint64_t`?  It's going to `int` in practice, isn't it ("the value is convertible to the first type from the following list able to hold their entire value range:  int...")?  Bumping it to unsigned is more ABI-brittle than needed.
</blockquote></details>

This is moot now, but to clarify r1's idea anyway: the `uint64` is actually less brittle than `int`, because it's future-proof -- it's correct no matter what curl decides to change its integer type to be.  As of today when the enum is smaller than 64 bits, casting to 64 bits will still have the same value as 32 bits.

PR #19455, opened 2023-05-22
Fix default arguments for GetStateNames() and friends to match C++.

RussTedrake:
The python bindings for MultibodyPlant::GetStateNames() as well as GetPositionNames, GetVelocityNames, and GetActuatorNames had slightly different default values than in c++.  This update makes them consistent.

+@ggould-tri for both reviews, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19455)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -1158,7 +1158,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def("GetPositionNames",
             overload_cast_explicit<std::vector<std::string>, bool, bool>(
                 &Class::GetPositionNames),
-            py::arg("add_model_instance_prefix") = false,
+            py::arg("add_model_instance_prefix") = true,
             py::arg("always_add_suffix") = true,
             cls_doc.GetPositionNames.doc_2args)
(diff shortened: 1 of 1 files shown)


ggould-tri (commented):
+(release notes: breaking change) This will break some possible existing calls.  It doesn't violate our deprecation policy ("Those defaults may change as we find better heuristics, without prior notice") but should be flagged in the exciting and dangerous section of the release notes in case anyone was doing string surgery on their dof names.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19455#-:-NW3PzYoCH1nmN9xLqQR:b-8vpk7p)* status: :shipit: complete! all discussions resolved, LGTM from assignee ggould-tri(platform)

jwnimmer-tri:
To my taste, changing this default value is not so likely to cause trouble that we need to scream about it.  All bug fixes or changes to defaults might impact users.  That's what the entire release notes document is for.  Unless there is a real and acute possibility of a difficult-to-diagnose user confusion resulting from a change, I don't think we need to call any special attention to it atop the document.  So, in #19456 I've omitted this from "breaking changes" and just kept it inline in the pydrake section.

PR #19457, opened 2023-05-22
Add binary operator overloads for Polynomial bop Expression

### LINKED ISSUE #19266: Automatic Type Casting Between Polynomial and Expression
When constructing polynomial optimization programs in Python, polynomial expressions often get cast into expressions, even though they're still polynomials. This requires frequent recasting back to polynomials, as there's no automatic type casting when adding a polynomial and an expression.

For example, the following (in my opinion, intuitive) code:
```
from pydrake.solvers import MathematicalProgram, Solve
import pydrake.symbolic as sym

deg = 4
prog = MathematicalProgram()
x, y = prog.NewIndeterminates(2)
a = prog.NewSosPolynomial(sym.Variables([x, y]), deg)[0]
b = prog.NewFreePolynomial(sym.Variables([x, y]), deg)
c = prog.NewFreePolynomial(sym.Variables([x, y]), deg)
poly = a + b * (x+y**3-2) + c * (x**2+y**2-1)
```
Throws the error message
```
TypeError: unsupported operand type(s) for +: 'pydrake.symbolic.Polynomial' and 'pydrake.symbolic.Expression'
```
To make it work correctly, the last line must be replaced with
```
poly = a + b * sym.Polynomial(x+y**3-2) + c * sym.Polynomial(x**2+y**2-1)
```
(which is much less readable!)

It would be helpful to avoid this, by either

- Supporting mathematical operations between polynomial and expression, leaving the output as an expression. (Ideally, this would also come with automatic casting support in functions such as `AddSosConstraint`, `AddEqualityConstraintBetweenPolynomials`, etc.)
- When adding indeterminates (or multiplying, exponentiating, etc.), treat the output as a polynomial instead of an expression.

AlexandreAmice:
Closes #19266.

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19457)


AlexandreAmice (commented):
+@hongkai-dai for feature review! Feel free to reassign.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19457#-:-NW3mPWb9wdgllPnjfwc:bf7vuqp)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

AlexandreAmice (commented):
Closes #19266. FYI @cohnt

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19457#-:-NW3mgnx3J4DwlFZFEV_:bc4ub9h)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @AlexandreAmice and @hongkai-dai)

hongkai-dai (commented):
:lgtm: thanks for working on this.

+@jwnimmer-tri could you do a platform review on this?

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19457#-:-NW4AD1L581ZJXz-5TOB:beoglro)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @AlexandreAmice and @jwnimmer-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19457)* status: 1 unresolved discussion

___
*[`common/symbolic/test/polynomial_test.cc` line 338 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19457#-NW7nK8sBQcZM3UhGB34:-NW83QU_AmBa5FcrTdVH:bjqgnqt) ([raw file](https://github.com/RobotLocomotion/drake/blob/c3cb39a6b950d8a40813a6569e5aa7db101c5704/common/symbolic/test/polynomial_test.cc#L338)):*

It tests that the return of a `Polynomial` and `Expression` that is a polynomial continues to be a polynomial after the operation as should be the case. I prefer to leave this test.
</blockquote></details>

Sure, it makes sense to keep the test.
 
We still should clarify for the _next_ person who comes along here and is confused, by regurgitating your answer here into a comment at this point in the test code.  We don't want a naive programmer to come along here and think it's redundant and then remove it accidentally.

PR #19463, opened 2023-05-22
Xml inertia rewriter

rpoyner-tri:
Towards #18773.

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)

### DIFF
--- a/tmp/BUILD.bazel
+++ b/tmp/BUILD.bazel
@@ -7,16 +7,12 @@ load(
 )
 load(
     "@drake//tools/skylark:drake_py.bzl",
-    "drake_py_library",
+    "drake_py_binary",
     "drake_py_unittest",
 )
 load("@drake//tools/install:install.bzl", "install_files")
 load("//tools/lint:lint.bzl", "add_lint_tests")
 load("@drake//tools/workspace:forward_files.bzl", "forward_files")
-load(
-    "//tools/workspace/dm_control_internal:files.bzl",
-    "dm_control_mujoco_files",
-)
 
 drake_cc_binary(
     name = "inertia_fixer_cxx",
@@ -26,6 +22,14 @@ drake_cc_binary(
         "//multibody/parsing:parser",
         "@gflags",
         "@tinyxml2_internal//:tinyxml2",
+    ],
+)
+
+drake_py_binary(
+    name = "inertia_fixer_py",
+    srcs = ["inertia_fixer.py"],
+    deps = [
+        "//bindings/pydrake",
     ],
 )
 
--- a/tmp/BUILD.bazel
+++ b/tmp/BUILD.bazel
@@ -22,11 +22,11 @@ drake_cc_binary(
     name = "inertia_fixer_cxx",
     srcs = ["inertia_fixer.cc"],
     deps = [
-        "//multibody/parsing:parser",
         "//common:add_text_logging_gflags",
+        "//multibody/parsing:parser",
         "@gflags",
+        "@tinyxml2_internal//:tinyxml2",
     ],
 )
-
 
 add_lint_tests()
--- a/tmp/BUILD.bazel
+++ b/tmp/BUILD.bazel
@@ -28,6 +28,7 @@ drake_cc_binary(
 drake_py_binary(
     name = "inertia_fixer_py",
     srcs = ["inertia_fixer.py"],
+    data = ["@drake_models"],
     deps = [
         "//bindings/pydrake",
     ],
--- a/tmp/BUILD.bazel
+++ b/tmp/BUILD.bazel
@@ -1,0 +1,32 @@
+load(
+    "@drake//tools/skylark:drake_cc.bzl",
+    "drake_cc_binary",
+    "drake_cc_googletest",
+    "drake_cc_library",
+    "drake_cc_package_library",
+)
+load(
+    "@drake//tools/skylark:drake_py.bzl",
+    "drake_py_library",
+    "drake_py_unittest",
+)
+load("@drake//tools/install:install.bzl", "install_files")
+load("//tools/lint:lint.bzl", "add_lint_tests")
+load("@drake//tools/workspace:forward_files.bzl", "forward_files")
+load(
+    "//tools/workspace/dm_control_internal:files.bzl",
+    "dm_control_mujoco_files",
+)
+
+drake_cc_binary(
+    name = "inertia_fixer_cxx",
+    srcs = ["inertia_fixer.cc"],
+    deps = [
+        "//multibody/parsing:parser",
+        "//common:add_text_logging_gflags",
+        "@gflags",
+    ],
+)
+
+
+add_lint_tests()
--- a/tmp/inertia_fixer.cc
+++ b/tmp/inertia_fixer.cc
@@ -21,10 +21,10 @@ using tinyxml2::XMLDocument;
 using tinyxml2::XMLElement;
 using tinyxml2::XMLPrinter;
 
-// XXX Does tinyxml2 quote-swapping ('' to "") break any strings?
-
-// Nope: it does entity substitution. So things may look worse in source, but
-// information will be preserved.
+// Does tinyxml2 quote-swapping ('' to "") break any strings?  Nope: it does
+// entity substitution. So things may look worse in source, but information
+// will be preserved. Furthermore, pretty much every other xml stack I have
+// tried does something similar.
 /*
 For Input:
   <material name='silver-"ish"'>
@@ -248,42 +248,22 @@ class InertiaProcessor {
       return {};
     }
 
-    // We only get one mass to describe potentially multiple geometries;
-    // proceed carefully to apportion masses correctly.
-    const double mass = old_inertia.get_mass();
-
     // Collect some density==1 inertias for all geometries.
-    std::vector<SpatialInertia<double>> M_GG_G_ones;
-    M_GG_G_ones.reserve(geoms.size());
+    SpatialInertia<double> M_BBo_B_one(0, {0, 0, 0}, {0, 0, 0});
     for (const auto& geom : geoms) {
-      M_GG_G_ones.push_back(CalcSpatialInertia(inspector.GetShape(geom), 1.0));
+      const auto& M_GG_G_one =
+          CalcSpatialInertia(inspector.GetShape(geom), 1.0);
+      const auto& X_BG = inspector.GetPoseInFrame(geom);
+      const auto M_GBo_B_one = M_GG_G_one.ReExpress(X_BG.rotation())
+                               .Shift(-X_BG.translation());
+      M_BBo_B_one += M_GBo_B_one;
     }
 
-    // Work out the total volume, and save the individual volumes.
-    double total_volume{};
-    std::vector<double> volumes;
-    volumes.reserve(M_GG_G_ones.size());
-    for (const auto& M_GG_G_one : M_GG_G_ones) {
-      // density = mass / volume; volume = mass / density ( 1.0! );
-      // volume = mass!
-      const double volume = M_GG_G_one.get_mass();
-      volumes.push_back(volume);
-      total_volume += volume;
-    }
+    // Rebuild the final inertia using the mass found in the input.
+    const double mass = old_inertia.get_mass();
+    SpatialInertia<double> M_BBo_B(mass, M_BBo_B_one.get_com(),
+                                M_BBo_B_one.get_unit_inertia());
 
-    // Make the individual mass-corrected inertias, express them in a common
-    // frame, and combine them.
-    SpatialInertia<double> M_BBo_B(0, {0, 0, 0}, {0, 0, 0});
-    for (int k = 0; k < ssize(M_GG_G_ones); ++k) {
-      const double portion = volumes[k] / total_volume;
-      const auto& M_GG_G_one = M_GG_G_ones[k];
-      SpatialInertia<double> M_GG_G(mass * portion, M_GG_G_one.get_com(),
-                                    M_GG_G_one.get_unit_inertia());
-      const auto& X_BG = inspector.GetPoseInFrame(geoms[k]);
-      const auto M_GBo_B = M_GG_G.ReExpress(X_BG.rotation())
-                           .Shift(-X_BG.translation());
-      M_BBo_B += M_GBo_B;
-    }
     return M_BBo_B;
   }
 
--- a/tmp/inertia_fixer.cc
+++ b/tmp/inertia_fixer.cc
@@ -1,0 +1,76 @@
+#include <regex>
+
+#include <drake_vendor/tinyxml2.h>
+#include <gflags/gflags.h>
+
+#include "drake/multibody/parsing/parser.h"
+
+namespace drake {
+namespace multibody {
+namespace {
+
+int do_main(int argc, char* argv[]) {
+  gflags::SetUsageMessage("[INPUT-FILE-OR-URL]\n"
+                          "Rewrite URDF/SDFormat with fixed-up inertias");
+  gflags::ParseCommandLineFlags(&argc, &argv, true);
+  if (argc < 2) {
+    drake::log()->error("missing input filename");
+    return 1;
+  }
+
+  // Hardcode a log pattern that gives us un-decorated error messages.
+  // This defeats the `-spdlog_pattern` command line option; oh well.
+  drake::logging::set_log_pattern("%v");
+
+  MultibodyPlant<double> plant{0.0};
+  drake::geometry::SceneGraph<double> scene_graph;
+  if (FLAGS_scene_graph) {
+    plant.RegisterAsSourceForSceneGraph(&scene_graph);
+  }
+  Parser parser{&plant};
+  parser.package_map().PopulateFromRosPackagePath();
+  if (FLAGS_strict) {
+    parser.SetStrictParsing();
+  }
+
+  const bool is_url = std::regex_search(
+      argv[1], std::regex("^[A-Za-z0-9+.-]+://"));
+
+  drake::log()->info("parsing {}", argv[1]);
+
+  // Chances are good that parsing may end in an error (implemented as throw).
+  // For purposes of this tool, we should make its message usable. To do that,
+  // catch it and display it via logging. Yes, this violates the coding
+  // standard; in this case it makes the tool significantly more useful.
+  std::vector<ModelInstanceIndex> models;
+  try {
+    if (is_url) {
+      models = parser.AddModelsFromUrl(argv[1]);
+    } else {
+      models = parser.AddModels(argv[1]);
+    }
+  } catch (const std::exception& e) {
+    drake::log()->error(e.what());
+    ::exit(EXIT_FAILURE);
+  }
+
+  drake::log()->info("Parsed {} models.", models.size());
+
+
+  drake::log()->info("\n\n\nXML Text\n\n\n");
+
+  XMLDocument xml_doc;
+  xml_doc.LoadFile(argv[1]);
+  xml_doc.SaveFile(argv[2]);
+
+
+  return models.empty();
+}
+
+}  // namespace
+}  // namespace multibody
+}  // namespace drake
+
+int main(int argc, char* argv[]) {
+  return drake::multibody::do_main(argc, argv);
+}
--- a/tmp/inertia_fixer.cc
+++ b/tmp/inertia_fixer.cc
@@ -5,9 +5,15 @@
 
 #include "drake/multibody/parsing/parser.h"
 
+DEFINE_bool(scene_graph, true, "include/exclude scene graph");
+
 namespace drake {
 namespace multibody {
 namespace {
+
+using tinyxml2::XMLNode;
+using tinyxml2::XMLDocument;
+using tinyxml2::XMLElement;
 
 int do_main(int argc, char* argv[]) {
   gflags::SetUsageMessage("[INPUT-FILE-OR-URL]\n"
@@ -17,6 +23,7 @@ int do_main(int argc, char* argv[]) {
     drake::log()->error("missing input filename");
     return 1;
   }
+  const char* outfile = (argc > 2) ? argv[2] : "/dev/stdout";
 
   // Hardcode a log pattern that gives us un-decorated error messages.
   // This defeats the `-spdlog_pattern` command line option; oh well.
@@ -29,9 +36,6 @@ int do_main(int argc, char* argv[]) {
   }
   Parser parser{&plant};
   parser.package_map().PopulateFromRosPackagePath();
-  if (FLAGS_strict) {
-    parser.SetStrictParsing();
-  }
 
   const bool is_url = std::regex_search(
       argv[1], std::regex("^[A-Za-z0-9+.-]+://"));
@@ -61,7 +65,12 @@ int do_main(int argc, char* argv[]) {
 
   XMLDocument xml_doc;
   xml_doc.LoadFile(argv[1]);
-  xml_doc.SaveFile(argv[2]);
+
+  // TODO(rpoyner-tri): Somehow, in here, inertia-fixing magic happens.
+
+  // tinyxml2 preserves the input text almost exactly. Probably good enough for
+  // this purpose.
+  xml_doc.SaveFile(outfile);
 
 
   return models.empty();
--- a/tmp/inertia_fixer.cc
+++ b/tmp/inertia_fixer.cc
@@ -1,3 +1,4 @@
+#include <fstream>
 #include <regex>
 
 #include <drake_vendor/tinyxml2.h>
@@ -15,6 +16,19 @@ using geometry::SceneGraph;
 using tinyxml2::XMLNode;
 using tinyxml2::XMLDocument;
 using tinyxml2::XMLElement;
+using tinyxml2::XMLPrinter;
+
+
+// We jump through hoops here, just to get 2-space indented output.
+class XmlPrinter final : public XMLPrinter {
+ public:
+  XmlPrinter() : XMLPrinter() {}
+  void PrintSpace(int depth) final {
+    for (int i = 0; i < depth; ++i) {
+      Write("  ");
+    }
+  }
+};
 
 
 std::optional<SpatialInertia<double>> MaybeFixBodyInertia(
@@ -225,10 +239,11 @@ int do_main(int argc, char* argv[]) {
     }
   }
 
-  // tinyxml2 preserves the input text almost exactly. Probably good enough for
-  // this purpose.
-  xml_doc.SaveFile(outfile);
-
+  // Use our custom XmlPrinter to "print to memory", then dump that to file.
+  XmlPrinter printer;
+  xml_doc.Print(&printer);
+  std::ofstream out(outfile);
+  out << printer.CStr();
 
   return models.empty();
 }
--- a/tmp/inertia_fixer.cc
+++ b/tmp/inertia_fixer.cc
@@ -23,6 +23,16 @@ using tinyxml2::XMLPrinter;
 
 // XXX Does tinyxml2 quote-swapping ('' to "") break any strings?
 
+// Nope: it does entity substitution. So things may look worse in source, but
+// information will be preserved.
+/*
+For Input:
+  <material name='silver-"ish"'>
+
+We get Output:
+  <material name="silver-&quot;ish&quot;">
+*/
+
 // We jump through hoops here, just to get 2-space indented output.
 class XmlPrinter final : public XMLPrinter {
  public:
@@ -34,6 +44,7 @@ class XmlPrinter final : public XMLPrinter {
   }
 };
 
+// Create a child element if it does not already exist.
 XMLElement* EnsureChildElement(XMLElement* el, const char* child) {
   XMLElement* kid = el->FirstChildElement(child);
   if (!kid) {
@@ -42,10 +53,13 @@ XMLElement* EnsureChildElement(XMLElement* el, const char* child) {
   return kid;
 }
 
+// Format a double in shortest round-trip representation.
 std::string roundtrip(double x) {
   return fmt::format("{}", x);
 }
 
+// Recursively find all of the descendant elements of 'el` named "link", and
+// accumulate them in an externally supplied vector.
 void FindLinks(XMLElement* el, std::vector<XMLElement*>* links) {
   if (std::string(el->Name()) == "link") {
     links->push_back(el);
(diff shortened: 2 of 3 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NW4o80_BqpXDEdigsqw:bl232f9) (no related file):*
None of the application logic will be much different in C++ vs Python.  So the question is which XML library does the best job of surgical edits to a human-readable XML document -- https://docs.python.org/3/library/xml.html or tinyxml2?

In the worst case, I'm pretty sure we can do what we need with either parser.  (Identify the XML inertia element start and end characters and text-replace it directly within the file buffer as a string.)  The question is whether either of the two parsers supports round-trip operation with minimal disruption to the file.

Our goal should be a minimal XML diff, where the only changes are to the `<inertia>...</inertia>` stanza, and nothing else.  Unrelated whitespace and comments need to be unchanged.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NW86T3j3BBqs-5WeZRP:b-7cd8ss) (no related file):*

None of the application logic will be much different in C++ vs Python.  So the question is which XML library does the best job of surgical edits to a human-readable XML document -- https://docs.python.org/3/library/xml.html or tinyxml2?

In the worst case, I'm pretty sure we can do what we need with either parser.  (Identify the XML inertia element start and end characters and text-replace it directly within the file buffer as a string.)  The question is whether either of the two parsers supports round-trip operation with minimal disruption to the file.

Our goal should be a minimal XML diff, where the only changes are to the `<inertia>...</inertia>` stanza, and nothing else.  Unrelated whitespace and comments need to be unchanged.
</blockquote></details>

FWIW, etree didn't come even close to round-trip. I haven't researched all of the python std library alternatives; however `lxml` (external dependency we just recently deleted) does a good job.

Believe it or not (=P), I opted for the least-dependency-depth solution.

The only corner-cases of non-round-trip I've seen with tinyxml2 are white-space and quote-entity fiddling. I'm somewhat doubtful that any rewriter will not stumble over those (possible exception being lxml).

In any case, I don't much mind transcribing if we find a more precise round-trip library.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NW88RsjDk9qvc8dNNoo:b7sku6m) (no related file):*
> The only corner-cases of non-round-trip I've seen with tinyxml2 are white-space and quote-entity fiddling. I'm somewhat doubtful that any rewriter will not stumble over those (possible exception being lxml).

In that case it sounds like we should use some library for xml parsing, but not use it for reserialization?  Rather, we ask the library for the start+end character spans for the elements we intent to replace, edit the file-as-string buffer directly, and then write the string to disk instead of reserializing.

Since we do not distribute C++ binaries in releases, the `def main()` we tell users to call needs to be in Python.  It would be plausible to write the implementation logic in C++ and bind it, so that `main()` is just a thin wrapper.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NW8BXXB0oV5C_XnBjFP:b-q38fwx) (no related file):*

> The only corner-cases of non-round-trip I've seen with tinyxml2 are white-space and quote-entity fiddling. I'm somewhat doubtful that any rewriter will not stumble over those (possible exception being lxml).

In that case it sounds like we should use some library for xml parsing, but not use it for reserialization?  Rather, we ask the library for the start+end character spans for the elements we intent to replace, edit the file-as-string buffer directly, and then write the string to disk instead of reserializing.

Since we do not distribute C++ binaries in releases, the `def main()` we tell users to call needs to be in Python.  It would be plausible to write the implementation logic in C++ and bind it, so that `main()` is just a thin wrapper.
</blockquote></details>

Ah, good call on the distro issue. I was pondering splitting main/library anyway, just for testing access.

As for bit-perfect round-trip, I'll look into it more. 

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWDUNJL7lBwvfGV0Gmh:bf7h4js) (no related file):*

Ah, good call on the distro issue. I was pondering splitting main/library anyway, just for testing access.

As for bit-perfect round-trip, I'll look into it more. 
</blockquote></details>

A few more experiments have convinced me that if we want bit-exact round-trip, it's `lxml` or the highway.

___

rpoyner-tri:

A few more experiments have convinced me that if we want bit-exact round-trip, it's `lxml` or the highway.
</blockquote></details>

Hm. Even `lxml` can't deal with whitespace adjacent to comments, or at EOF:
```
rico@torg:~/checkout/drake$ diff -u  ~/third/example-robot-data/robots/a1_description/urdf/a1.urdf foo.urdf
--- /home/rico/third/example-robot-data/robots/a1_description/urdf/a1.urdf	2023-05-24 15:16:17.503293194 -0400
+++ foo.urdf	2023-05-24 15:16:00.607024271 -0400
@@ -1,8 +1,5 @@
-<?xml version="1.0" ?>
-<!-- ==================================================================================================== -->
-<!-- |    This document was autogenerated by xacro, then had the Gazebo and transmission tags removed   | -->
-<!-- ==================================================================================================== -->
-<robot name="a1">
+<?xml version='1.0' encoding='ASCII'?>
+<!-- ==================================================================================================== --><!-- |    This document was autogenerated by xacro, then had the Gazebo and transmission tags removed   | --><!-- ==================================================================================================== --><robot name="a1">
   <material name="black">
     <color rgba="0.0 0.0 0.0 1.0"/>
   </material>
@@ -15,7 +12,7 @@
   <material name="grey">
     <color rgba="0.2 0.2 0.2 1.0"/>
   </material>
-  <material name='silver-"ish"'>
+  <material name="silver-&quot;ish&quot;">
     <color rgba="0.9137254901960784 0.9137254901960784 0.8470588235294118 1.0"/>
   </material>
   <material name="orange">
@@ -596,4 +593,4 @@
       <inertia ixx="9.600000000000001e-06" ixy="0.0" ixz="0.0" iyy="9.600000000000001e-06" iyz="0.0" izz="9.600000000000001e-06"/>
     </inertial>
   </link>
-</robot>
+</robot>
\ No newline at end of file
```

The entity-conversion thing seems pretty consistent across libraries I have tried.

rpoyner-tri (commented):
`r2` implements Sean's suggested simplified math, and updates some other commentary.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-:-NWEPw1b8GHA8bodbpSL:b-4j3adt)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWhc7Q9B5FtoeHJ8-Yo:b97zwiq) (no related file):*
> Rather, we ask the library for the start+end character spans for the elements we intent to replace

Looks like C (not C++!) expat can do this; I'm still looking at whether the std python expat bindings also expose enough to do the same thing.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWiIzvkDLFJdqS_T8Lv:b-7asdjc) (no related file):*

> Rather, we ask the library for the start+end character spans for the elements we intent to replace

Looks like C (not C++!) expat can do this; I'm still looking at whether the std python expat bindings also expose enough to do the same thing.
</blockquote></details>

So, let's really think through what "preserve bit-exact input" really entails:

* using expat, I can write a call-back-based parse that builds up an index of the input file. It will still be rife with corner cases, owing to `<el></el>` vs `<el/>`, and other nonsense.
* I still have to infer something about indent style of the input file, since I will need to write out nested elements. If I see literal tabs in the input file, all bets are pretty much off.
* Even having succeeded at the above, there will still be difficulties preserving attribute order, etc.

Quoth @ggould-tri : "This is a terrible plan and you shouldn't do it."

What could/should we do?

* Guarantee preservation of semantic content, aside from the parts we edit
* Guarantee that any transformation is indempotent

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWifSKsDIEv6OhaAZK_:b-gcjfn1) (no related file):*

So, let's really think through what "preserve bit-exact input" really entails:

* using expat, I can write a call-back-based parse that builds up an index of the input file. It will still be rife with corner cases, owing to `<el></el>` vs `<el/>`, and other nonsense.
* I still have to infer something about indent style of the input file, since I will need to write out nested elements. If I see literal tabs in the input file, all bets are pretty much off.
* Even having succeeded at the above, there will still be difficulties preserving attribute order, etc.

Quoth @ggould-tri : "This is a terrible plan and you shouldn't do it."

What could/should we do?

* Guarantee preservation of semantic content, aside from the parts we edit
* Guarantee that any transformation is indempotent
</blockquote></details>

I don't see where the difficulty comes from.  Maybe I've not correctly followed your explanation somehow.

I expect that trying to _reserialize_ the file is impractical, and madness.  I think we should use a real xml parser to find the file offsets that we need to edit (as well as the collision geometries to infer from), but then use string processing to do the edit and re-save -- no re-serialization.

For each `<link>`, if the file already has `<inertia>...</inertia>` then we presumably can find the starting and ending position of that inertia block (either as character offsets from the beginning of the file, or line + column offsets; they are equivalent).  If there is no inertia stanza already, we can choose to put it as the first or last element within the enclosing `<link>`.  Either way, we know exactly where it should go.

Taking the whole file as a single string, we can then erase the existing stanza (if any) and insert the new stanza charater-wise.  There's no need to infer the tab-layout of the existing file.  We can use a fixed indentation (e.g., +2 space per line).  If the user wants to repave the file whitespace, they are invited to do so after we're finished.

Problem solved?

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWimvZ09QpHbLon8DJG:b-cmm0ft) (no related file):*

I don't see where the difficulty comes from.  Maybe I've not correctly followed your explanation somehow.

I expect that trying to _reserialize_ the file is impractical, and madness.  I think we should use a real xml parser to find the file offsets that we need to edit (as well as the collision geometries to infer from), but then use string processing to do the edit and re-save -- no re-serialization.

For each `<link>`, if the file already has `<inertia>...</inertia>` then we presumably can find the starting and ending position of that inertia block (either as character offsets from the beginning of the file, or line + column offsets; they are equivalent).  If there is no inertia stanza already, we can choose to put it as the first or last element within the enclosing `<link>`.  Either way, we know exactly where it should go.

Taking the whole file as a single string, we can then erase the existing stanza (if any) and insert the new stanza charater-wise.  There's no need to infer the tab-layout of the existing file.  We can use a fixed indentation (e.g., +2 space per line).  If the user wants to repave the file whitespace, they are invited to do so after we're finished.

Problem solved?
</blockquote></details>

What, exactly, about the already-implemented solution is madness? I don't see that there *is* a problem to be solved.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWjB0iy8l_lGjv0pAyJ:b7ukoxe) (no related file):*

What, exactly, about the already-implemented solution is madness? I don't see that there *is* a problem to be solved.
</blockquote></details>

I hadn't looked at the code yet.  I was only going by the above thread that was wondering about `<el></el>` idempotence and such.  It seemed to me like any solution where that was in doubt was already barking up the wrong tree.

Does the current code meet the invariant that _nothing_ except the text inside an `inertia` element is changed?  Comments are intact, nothing is reflowed, no whitespace or indentation, changes, no `&quot` vs `"` swaps, etc. -- in other words, the `diff -u` between old and new file *only* mentions lines where inertia was edited (or inserted)?

If it already does that, then yes it was already solved.  I can try it tomorrow on some sample files and see what happens.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWm13MNDli9aY2TKLmG:b-t3wqxb) (no related file):*

I hadn't looked at the code yet.  I was only going by the above thread that was wondering about `<el></el>` idempotence and such.  It seemed to me like any solution where that was in doubt was already barking up the wrong tree.

Does the current code meet the invariant that _nothing_ except the text inside an `inertia` element is changed?  Comments are intact, nothing is reflowed, no whitespace or indentation, changes, no `&quot` vs `"` swaps, etc. -- in other words, the `diff -u` between old and new file *only* mentions lines where inertia was edited (or inserted)?

If it already does that, then yes it was already solved.  I can try it tomorrow on some sample files and see what happens.
</blockquote></details>

Please take it for a spin. We can spend as much effort as you deem necessary preserving non-semantic whitespace, but my estimation is that we are in a forest of only wrong trees (XML), and we need to get out alive.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWmPqKW2YCp6q7yrc0y:b-31phn2) (no related file):*

Please take it for a spin. We can spend as much effort as you deem necessary preserving non-semantic whitespace, but my estimation is that we are in a forest of only wrong trees (XML), and we need to get out alive.
</blockquote></details>

While that is stewing, I'll see if I can prototype a `pyexpat` version that tries to preserve source text.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWo2zoG6vcGoGBeE4ff:b-p13xmq) (no related file):*

While that is stewing, I'll see if I can prototype a `pyexpat` version that tries to preserve source text.
</blockquote></details>

`r4` has the promised `pyexpat` prototype for side-by-side comparision. There is much still to fix there, but it sorta works, and gives a flavor of how to do it. The "infer indentation" code is madness; don't let that distract you.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NW4o80_BqpXDEdigsqv:-NWsffpGDfGVY_4eeoEz:b-luqtjg) (no related file):*

`r4` has the promised `pyexpat` prototype for side-by-side comparision. There is much still to fix there, but it sorta works, and gives a flavor of how to do it. The "infer indentation" code is madness; don't let that distract you.
</blockquote></details>

`r5` improves the prototype py version in various ways. I'd say at this point I *mostly* like it better than the .cc.

___
___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NWsfyxaANtNtdkU90Xs:b-lsuq2x) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*
> ```Python
>         return mapping
> 
>     def format_inertia(self, inertial_facts, fixed_inertia, indentation):
> ```

This formatter and the sdf one below will currently erase comments that were contained within the original `inertial` tag. Bug? Feature? So far I'm leaning feature.
___
*[`tmp/inertia_fixer.py` line 85 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsgbsc2BZKhMCXsedn:-NWsgbsc2BZKhMCXsedo:bfnok6b) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L85)):*
> ```Python
>         assert len(models) >= 1
>         # XXX This assertion is likely violated by files that use include tags.
>         assert len(inertials) == plant.num_bodies() - 1, (
> ```

This sdf   associator is all  wrong and frequently breaks on probably-fine files found in drake. More work needed here.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWxBgylE9oRhLX7uC_8:-NWxBgylE9oRhLX7uC_9:b7ukgll) (no related file):*
A corner case worth pondering:
```
bazel-bin/tmp/inertia_fixer_py --in_place examples/hydroelastic/python_nonconvex_mesh/pepper.sdf
inertial facts: ElementFacts(name='inertial', attributes={}, depth=3, start=SourceLocation(index=196, line=6, column=6), end=SourceLocation(index=515, line=17, column=6), parent=ElementFacts(name='link', attributes={'name': 'yellow_bell_pepper_no_stem'}, depth=2, start=SourceLocation(index=118, line=4, column=4), end=None, parent=None, serial_number=0), serial_number=None)
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 376, in <module>
    main()
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 362, in main
    output_text = processor.process()
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 312, in process
    maybe_inertia = self._maybe_fix_inertia(body_index)
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 295, in _maybe_fix_inertia
    M_GG_G_one = CalcSpatialInertia(inspector.GetShape(geom), 1.0)
RuntimeError: CalcMassProperties currently only supports .obj files for mesh geometries but was given '/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake_models/veggies/yellow_bell_pepper_no_stem_low.vtk'.

```

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWxBgylE9oRhLX7uC_8:-NWxCRC5A-xQwQtB4zXd:blia1jb) (no related file):*

A corner case worth pondering:
```
bazel-bin/tmp/inertia_fixer_py --in_place examples/hydroelastic/python_nonconvex_mesh/pepper.sdf
inertial facts: ElementFacts(name='inertial', attributes={}, depth=3, start=SourceLocation(index=196, line=6, column=6), end=SourceLocation(index=515, line=17, column=6), parent=ElementFacts(name='link', attributes={'name': 'yellow_bell_pepper_no_stem'}, depth=2, start=SourceLocation(index=118, line=4, column=4), end=None, parent=None, serial_number=0), serial_number=None)
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 376, in <module>
    main()
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 362, in main
    output_text = processor.process()
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 312, in process
    maybe_inertia = self._maybe_fix_inertia(body_index)
  File "/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake/tmp/inertia_fixer.py", line 295, in _maybe_fix_inertia
    M_GG_G_one = CalcSpatialInertia(inspector.GetShape(geom), 1.0)
RuntimeError: CalcMassProperties currently only supports .obj files for mesh geometries but was given '/home/rico/checkout/drake/bazel-bin/tmp/inertia_fixer_py.runfiles/drake_models/veggies/yellow_bell_pepper_no_stem_low.vtk'.

```
</blockquote></details>

For my part, I am OK with an error in this case, i.e., declaring that to be an unsupported input file.  The inertia fixer is primarily aimed at stale model files from the internet past.   If it bombs out on already-highly-drake-specific model files, I think it's still an improvement on the status quo.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NXBbp578CCQTZmgCQFV:b-6ervby) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*

I think it's a feature to repave `<origin>` and `<inertia>`, but maybe not `<inertial>`.

If the pre-existing `<inertial>` has a `<mass>` then I don't think we should repave the entire `<inertial>`, since we want to keep the existing mass intact and the existing mass element might have useful comments nearby.
</blockquote></details>

Mass is a bit tricky, because the mbp mass may not necessarily be the file mass. I'm sure there are things we can do.

FWIW, the only comments I have seen removed anywhere were these (below are just excerpts of longer paragraphs):

```
examples/kuka_iiwa_arm/models/desk/transcendesk55inch.sdf-        TODO(Lucy-tri): Compute a more accurate spatial inertia of this table.
examples/kuka_iiwa_arm/models/table/extra_heavy_duty_table.sdf-        TODO(lucy-tri): Compute a more accurate spatial inertia of this table.
examples/kuka_iiwa_arm/models/table/extra_heavy_duty_table_surface_only_collision.sdf-        TODO(lucy-tri): Compute a more accurate spatial inertia of this table.

```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NXBpj9UEy7uD1lQEIkE:bkx6am2) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*

Mass is a bit tricky, because the mbp mass may not necessarily be the file mass. I'm sure there are things we can do.

FWIW, the only comments I have seen removed anywhere were these (below are just excerpts of longer paragraphs):

```
examples/kuka_iiwa_arm/models/desk/transcendesk55inch.sdf-        TODO(Lucy-tri): Compute a more accurate spatial inertia of this table.
examples/kuka_iiwa_arm/models/table/extra_heavy_duty_table.sdf-        TODO(lucy-tri): Compute a more accurate spatial inertia of this table.
examples/kuka_iiwa_arm/models/table/extra_heavy_duty_table_surface_only_collision.sdf-        TODO(lucy-tri): Compute a more accurate spatial inertia of this table.

```

</blockquote></details>

Ah.  I forgot that sometimes we ignore the mass because its malformed.  In those cases, it would be fine to repave `<mass>` as well.

I do think that we should still avoid repaving `<inertial>` (and probably `<mass>`) by default.  I ran this on `drake/manipulation/models/iiwa_description/urdf/iiwa14_primitive_collision.urdf` and it deleted some probably-still-useful comments about the mass.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

___
*[`tmp/inertia_fixer.py` line 183 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NXBZiTo6011DiDDwtxT:-NXC-5UK27zSoGMd7MXR:b-wfct9h) ([raw file](https://github.com/RobotLocomotion/drake/blob/a6e6f59a9c1cfc3270d834fdb6ccb8750f9efa06/tmp/inertia_fixer.py#L183)):*

BTW For both SDFormat and URDF, we will need to extract the canonical Bcm pose and emit it.
</blockquote></details>

This will involve a line or two of math: the current summed-up inertia lives at body frame (so the individual geoms' inertias can be summed at all). It needs to be shifted back to the resulting center of mass. 

Not on the table for the first PR is rotation to zero out products of inertia.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NXWsBC56SRiUh9Uifgn:bgem06z) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*

Ah.  I forgot that sometimes we ignore the mass because its malformed.  In those cases, it would be fine to repave `<mass>` as well.

I do think that we should still avoid repaving `<inertial>` (and probably `<mass>`) by default.  I ran this on `drake/manipulation/models/iiwa_description/urdf/iiwa14_primitive_collision.urdf` and it deleted some probably-still-useful comments about the mass.
</blockquote></details>

So here (at some complexity cost) is a version that avoids murdering comments between {(pose|origin), mass, inertia}, but still murders comments within `inertia` (sdf only).

We can carry on this sort of pedantry to the n-th degree. I draw the line at attribute order.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NXX1xuh8QP8iCIn1mlD:bo32jui) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*

So here (at some complexity cost) is a version that avoids murdering comments between {(pose|origin), mass, inertia}, but still murders comments within `inertia` (sdf only).

We can carry on this sort of pedantry to the n-th degree. I draw the line at attribute order.
</blockquote></details>

`r8` relieves some mistaken complexity, and opens the way for another layer of comment preservation. I'm still not a fan of trying to preserve attribute order.

Still to be addressed is the new bug -- input elements with any of the child tags missing will not get any repaired values written.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NXwBuWcDf6A_pGRQSoH:bu2i0ub) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*
> I'm still not a fan of trying to preserve attribute order.

Yes, I think its fair to repave any attributes within an element we're already repaving.

rpoyner-tri (commented):
`r10` is better at synthesizing missing elements, and the association algorithms are a bit better. Looks like some new multibody parser errors might have turned up with all the recent SpatialInertia hackery; will investigate when I am back from vacation.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-:-NY4yryC6r0edGlRk4w6:b-itof9n)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
I've re-organized the files for testing and (eventual) distribution; these will turn up in a new PR, to avoid revision fatigue. Some final notes are added below to support continuity.

New PR is #19731.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-:-NZgfNgP1uZhuccvh2MB:bubtzz4)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`tmp/inertia_fixer.py` line 58 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NWsfyxaANtNtdkU90Xr:-NZgfjutCWVrNndPUcP9:b-xvwmad) ([raw file](https://github.com/RobotLocomotion/drake/blob/2c7eb3319d410e2425b9f13acd42a4c2fd3b3ab1/tmp/inertia_fixer.py#L58)):*

> I'm still not a fan of trying to preserve attribute order.

Yes, I think its fair to repave any attributes within an element we're already repaving.
</blockquote></details>

`r8-r11` mostly take care of the missing child tags problem. The future PR should have tests to help sharpen things up further. 
___
*[`tmp/inertia_fixer.py` line 183 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19463#-NXBZiTo6011DiDDwtxT:-NZgfz0P2s3BGXmOrjHH:b-x2vjxg) ([raw file](https://github.com/RobotLocomotion/drake/blob/a6e6f59a9c1cfc3270d834fdb6ccb8750f9efa06/tmp/inertia_fixer.py#L183)):*

This will involve a line or two of math: the current summed-up inertia lives at body frame (so the individual geoms' inertias can be summed at all). It needs to be shifted back to the resulting center of mass. 

Not on the table for the first PR is rotation to zero out products of inertia.
</blockquote></details>

Not addressed here; the new PR will have code and tests to achieve fully refined Bcm pose (shift to center of mass; rotate to zero out products of inertia).

PR #19468, opened 2023-05-23
Fix caching for autodiff in direct collocation

### LINKED ISSUE #19396: Revisit caching in DirectCollocationConstraint
In #19019, I implemented caching for DirectCollocation (via DirectCollocationConstraint).  Upon further reflection, the implementation I provided is not avoiding the computation that I hoped it was.  The test that I added in that PR is not quite strong enough to be representative of the way that the solvers call the constraint.

The problem is that we only get a cache hit when both the ExtractValue and ExtractGradient match for the autodiff types.  When I use the same context for evaluating x[k] and u[k] in both constraint k and k+1, the values will match, but the gradients will not... simply because the order of the variables passed into the Eval function is different (therefore the derivatives will be different, because they were set via InitializeAutoDiff as subsets of different tuples).

Of course it's possible to be a little more clever about manually handling the autodiff variables in the implementation, but it will take more work to see the benefits.

cc @hongkai-dai @sammy-tri

RussTedrake:
My previous attempt at caching failed to address the fact that the autodiff gradients would be different when the status/inputs were passed in different places inside the variable list for the constraint. Now we implement the chain rule locally so that the cache is kept warm with gradients only with respect to the current variables.

Resolves #19396.

+@sammy-tri for feature review, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468)

### DIFF
--- a/planning/trajectory_optimization/direct_collocation.cc
+++ b/planning/trajectory_optimization/direct_collocation.cc
@@ -12,7 +12,13 @@ namespace drake {
 namespace planning {
 namespace trajectory_optimization {
 
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
 using math::AreAutoDiffVecXdEqual;
+using math::ExtractGradient;
+using math::ExtractValue;
+using math::InitializeAutoDiff;
+using math::InitializeAutoDiffTuple;
 using solvers::Binding;
 using solvers::Constraint;
 using solvers::MathematicalProgram;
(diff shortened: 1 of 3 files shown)


RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468)* status: 1 unresolved discussion, LGTM missing from assignee sammy-tri(platform), needs at least two assigned reviewers (waiting on @RussTedrake and @sammy-tri)

___
*[`planning/trajectory_optimization/direct_collocation.cc` line 138 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-NW9MjNy7UFIGgDHjIbq:-NWATHegEIxYUMVLBpCv:bbd6b1n) ([raw file](https://github.com/RobotLocomotion/drake/blob/992129d28ce291cd9ce94a008c6c4210d1369535/planning/trajectory_optimization/direct_collocation.cc#L138)):*

This test doesn't seem to be testing that the incoming derivatives match.  I can change the derivatives in `TestCollocationConstraintCaching` without triggering a recompute.  (if that's not supposed to happen, it should probably have a test too)
</blockquote></details>

Yes.  That's the intended feature of this approach. I've added to the test to cover this case.

sammy-tri (commented):
> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform) (waiting on @sherm1)

RussTedrake:
*[`planning/trajectory_optimization/direct_collocation.cc` line 138 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-NW9MjNy7UFIGgDHjIbq:-NWDcRGX2udrEuT4afEZ:b-z97kvg) ([raw file](https://github.com/RobotLocomotion/drake/blob/992129d28ce291cd9ce94a008c6c4210d1369535/planning/trajectory_optimization/direct_collocation.cc#L138)):*

nit (changing to nit) ok, thanks.  I think I was mislead by the comment "To have cache hits, we must match not only the x and u values, but also the derivatives." into thinking that the derivatives were supposed to match.
</blockquote></details>

The derivatives have to match in the arguments we pass into the context below.  This is achieved by calling InitializeAutoDiffTuple here (so the derivatives are always the identity matrix w.r.t. this x and u).  That makes us immune the the derivatives that are passed in.

Is that clear now?  Does that wording help?

sammy-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468)* status: LGTM missing from assignee sherm1(platform) (waiting on @sherm1)

___
*[`planning/trajectory_optimization/direct_collocation.cc` line 138 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-NW9MjNy7UFIGgDHjIbq:-NWDfhih0TI_M1xdSn5s:bgsgg94) ([raw file](https://github.com/RobotLocomotion/drake/blob/992129d28ce291cd9ce94a008c6c4210d1369535/planning/trajectory_optimization/direct_collocation.cc#L138)):*

The derivatives have to match in the arguments we pass into the context below.  This is achieved by calling InitializeAutoDiffTuple here (so the derivatives are always the identity matrix w.r.t. this x and u).  That makes us immune the the derivatives that are passed in.

Is that clear now?  Does that wording help?
</blockquote></details>

lgtm yeah, that makes sense

sherm1 (commented):
Sneaky! Platform :lgtm: with one my-edification-only question.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-:-NWEZo2O8n-ZZgNhJwNc:b-cpntcx)* status: 1 unresolved discussion

___
*[`planning/trajectory_optimization/test/direct_collocation_test.cc` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-NWEYjMOCZcV-iWUJZMp:-NWEYjMPErpXY7gLEUsJ:b131dh4) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bac655bbc2e9669eeb36057da3a31b398524bbc/planning/trajectory_optimization/test/direct_collocation_test.cc#L224)):*
> ```cpp
>   DirectCollocationConstraint constraint(system, sample_context[0].get(),
>                                          sample_context[1].get(),
>                                          collocation_context.get());
> ```

BTW (unrelated to this PR) it seems odd that these three contexts are supplied as non-const. Are they actually modified by the Constraint? If only their caches get updated I would have expected const. Does something else get written?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19468)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),sherm1(platform)

___
*[`planning/trajectory_optimization/test/direct_collocation_test.cc` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19468#-NWEYjMOCZcV-iWUJZMp:-NWF9rfx9VdxfPspuPxv:bu9b1il) ([raw file](https://github.com/RobotLocomotion/drake/blob/8bac655bbc2e9669eeb36057da3a31b398524bbc/planning/trajectory_optimization/test/direct_collocation_test.cc#L224)):*

BTW (unrelated to this PR) it seems odd that these three contexts are supplied as non-const. Are they actually modified by the Constraint? If only their caches get updated I would have expected const. Does something else get written?
</blockquote></details>

Yup.  The state gets set (via SetContinuousState) and the input ports get fixed.

Issue #9661, opened 2018-10-12
fix inconsistent use of `timestep` and `time_step` throughout the codebase


RussTedrake:
I think different authors have used different variants.  I don't have a strong opinion, but would like things to be consistent.

jamiesnape:
Almost raised this in https://github.com/RobotLocomotion/drake/pull/9200. `timestep` is not in the dictionary, of course.

jwnimmer-tri:
The two-word is used approx 900 times, the one-word approx 300 times.  The 900-times is mostly dynamics code (multibody, simulator), the 300-times is mostly Russ code (traj opt).

Shall we settle on the two-word version for new code?

Is it worth fixing the one-word version in old code (possibly only non-API-changing locations, to avoid churn)?

RussTedrake:
I'm happy to go with the two-word version, and even with changing the API at some point if we decide that's best.  I agree that there is significant value in consistency.

sherm1:
Grammarhow favors "time step": https://grammarhow.com/timestep-time-step-or-time-step/

PR #19476, opened 2023-05-24
Rename timestep => time_step in systems/controllers

RussTedrake:
Towards #9661.

+@sherm1 for both reviews, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19476)


sherm1 (commented):
:lgtm: x 2

BTW two functions here with `timestep` parameters didn't require pydrake deprecation (which is nice) but that's because they were wrapped without giving arg names (not as nice). Consider whether that's worth an upgrade (elsewhere).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19476#-:-NWDZ7Cd2BiRpE1gre8v:bwt9rzm)* status: :shipit: complete! all discussions resolved, LGTM from assignee sherm1(platform)

Issue #18452, opened 2022-12-14
Top usability issues from Fall 2022 manipulation projects


RussTedrake:
I am combing back through the _long_ list of discussions with students this project season, and wanted to call out the hottest hot spots all in one place.  I will reference existing issues and am happy to triage these out into new issues, too.  But I wanted to provide a wholistic view at least once.

- Getting object models into Drake. 
  - [ ] Meshcat's support seemed sufficient (I believe it successfully loaded all objs... including even one that was semi-transparent! and it supports stl/dae as well), but since Drake only officially supprts obj, users had to convert to that format.  => #19479
  -  [ ] **RgbdSensor was a real bottleneck**.  One student said "i downloaded like a hundred objs from the internet, but less than 10% of them worked!".  I'll follow up to get specific failure cases.  Related to #11949 .  
  - [ ] We need to make it easier to run convex decomposition (https://github.com/gizatt/convex_decomp_to_sdf helped in many cases), if we cannot run it automatically. => #19479
  - [x] parser package paths were mysterious and loading error messages could be more helpful.  "Since our project requires importing a scooper object, we were surprised by the difficulty of setting up the package.xml and other boilerplate necessary to import it into the simulation. It ended up being simple to put in, but we couldn’t figure it out on our own before asking for help." Update: the parser interface has come a long way now, with url support, etc.
- [ ] Lots of people authored SDF files with centers of mass outside the collision geometry.  the results was that they thought the contact simulation was terrible, and started to lose faith in the simulator.  But actually, there was e.g. a swinging mass far below the ground.  (https://github.com/RobotLocomotion/drake/issues/17662)
- [x] ycb meshes not in the pip wheels really frustrated some folks. (#15774)
- [x] People still saw `SystemExit: condition 'Delta >0' failed`.  (https://github.com/RobotLocomotion/drake/issues/14590) . I thought #17557 was supposed to help, but at least in RL examples it still appeared (and gives absolutely no help to the user). Update: hopefully resolved in #19407.
- [x] Lots of people got confused by system wiring.  #18446 is an attempt to fix. We should look for more places where the messages could be more helpful. The most common errors were trying to connect directly to a port in the manipulation station subdiagram from the parent diagram, or trying to connect to the controller_plant (which is built and passed to the InverseDynamicsController, etc, but never registered directly with a builder).  I think https://github.com/RobotLocomotion/drake/issues/13874 could make a huge difference.
[...]

PR #19481, opened 2023-05-24
Improve error message for integrator smaller step size

RussTedrake:
I've found that even a little nudge to say "your system is probably unstable" and "no, the discrete-time multibodyplant is not (directly) to blame, look at the continuous-state parts of your system" can help them understand/resolve.

Related to #18452.

+@sherm1 for both reviews, please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19481)


RussTedrake:
CI failure looks like a flake.

PR #19488, opened 2023-05-24
[multibody] Reduce SapContactProblem to selected/excluded dofs.

joemasterjohn:
Towards #18983.

Adds functionality to reduce a `SapContactProblem` via `SapContactProblem::Reduce()` by providing a list of "locked" or non-participating dofs. This results in a permutation of the original cliques and constraints of the problem as well as the associated linear dynamics matrices, v_star, and constraint jacobians. The returned data structure `ReducedMapping` contains the information to map a `SapSolverResults` associated with the reduced dof problem to a corresponding `SapSolverResults` for the original problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488)

### DIFF
--- a/multibody/contact_solvers/sap/BUILD.bazel
+++ b/multibody/contact_solvers/sap/BUILD.bazel
@@ -60,6 +60,8 @@ drake_cc_library(
         "//common:unused",
         "//common:value",
         "//multibody/contact_solvers:matrix_block",
+        "//multibody/contact_solvers/sap:partial_permutation",
+        "//multibody/plant:slicing_and_indexing",
     ],
 )
 
@@ -96,6 +98,7 @@ drake_cc_library(
         ":sap_constraint",
         "//common:default_scalars",
         "//common:essential",
+        "//multibody/plant:slicing_and_indexing",
     ],
 )
 
--- a/multibody/contact_solvers/sap/sap_constraint.cc
+++ b/multibody/contact_solvers/sap/sap_constraint.cc
@@ -4,6 +4,8 @@
 
 #include "drake/common/default_scalars.h"
 #include "drake/common/eigen_types.h"
+#include "drake/common/ssize.h"
+#include "drake/multibody/plant/slicing_and_indexing.h"
 
 namespace drake {
 namespace multibody {
@@ -50,6 +52,52 @@ void SapConstraint<T>::CalcCostHessian(const AbstractValue& data,
   const int ne = num_constraint_equations();
   G->resize(ne, ne);
   DoCalcCostHessian(data, G);
+}
+
+template <typename T>
+std::unique_ptr<SapConstraint<T>> SapConstraint<T>::MakeReduced(
+    const PartialPermutation& clique_permutation,
+    const std::vector<std::vector<int>>& per_clique_known_dofs) const {
+  DRAKE_DEMAND(clique_permutation.domain_size() ==
+               ssize(per_clique_known_dofs));
+  DRAKE_DEMAND(first_clique() < clique_permutation.domain_size());
+  DRAKE_DEMAND(num_cliques() <= 1 ||
+               second_clique() < clique_permutation.domain_size());
+
+  const bool first_participates =
+      clique_permutation.participates(first_clique());
+  const bool second_participates =
+      num_cliques() > 1 && clique_permutation.participates(second_clique());
+
+  // Neither clique participates, no constraint made.
+  if (!first_participates && !second_participates) return nullptr;
+
+  std::unique_ptr<SapConstraint<T>> c = this->Clone();
+
+  if (first_participates && second_participates) {
+    // Permute both cliques.
+    c->J_ = SapConstraintJacobian<T>(
+        clique_permutation.permuted_index(first_clique()),
+        drake::multibody::internal::ExcludeCols(
+            first_clique_jacobian(), per_clique_known_dofs[first_clique()]),
+        clique_permutation.permuted_index(second_clique()),
+        drake::multibody::internal::ExcludeCols(
+            second_clique_jacobian(), per_clique_known_dofs[second_clique()]));
+  } else if (first_participates) {
+    // Single clique, permute the first clique.
+    c->J_ = SapConstraintJacobian<T>(
+        clique_permutation.permuted_index(first_clique()),
+        drake::multibody::internal::ExcludeCols(
+            first_clique_jacobian(), per_clique_known_dofs[first_clique()]));
+  } else {
+    // Single clique, permute the second clique.
+    c->J_ = SapConstraintJacobian<T>(
+        clique_permutation.permuted_index(second_clique()),
+        drake::multibody::internal::ExcludeCols(
+            second_clique_jacobian(), per_clique_known_dofs[second_clique()]));
+  }
+
+  return c;
 }
 
 }  // namespace internal
--- a/multibody/contact_solvers/sap/sap_constraint.h
+++ b/multibody/contact_solvers/sap/sap_constraint.h
@@ -2,12 +2,14 @@
 
 #include <memory>
 #include <utility>
+#include <vector>
 
 #include "drake/common/drake_copyable.h"
 #include "drake/common/eigen_types.h"
 #include "drake/common/unused.h"
 #include "drake/common/value.h"
 #include "drake/multibody/contact_solvers/matrix_block.h"
+#include "drake/multibody/contact_solvers/sap/partial_permutation.h"
 #include "drake/multibody/contact_solvers/sap/sap_constraint_jacobian.h"
 
 namespace drake {
@@ -156,6 +158,33 @@ class SapConstraint {
 
   /* Polymorphic deep-copy into a new instance. */
   std::unique_ptr<SapConstraint<T>> Clone() const { return DoClone(); }
+
+  /* Creates a "reduced" clone of this constraint by removing known DoFs from
+   the constraint's Jacobian. That is, the newly reduced constraint will have
+   this constraint's Jacobian excluding columns for known DoFs. The following
+   table breaks down how the returned constraint relates to this constraint
+   where:
+     participates(clique_index) = clique_permutation.participates(clique_index)
+     permuted(clique_index) = clique_permutation.permuted_index(clique_index)
+   For convenience: c.first_clique  = i   c.second_clique  = j
+                    cᵣ.first_clique = iᵣ  cᵣ.second_clique = jᵣ
+     ________________________________________________________________________
+    | participates(i) | participates(j) |        iᵣ       |         jᵣ       |
+    |------------------------------------------------------------------------|
+    |     FALSE       |     FALSE       |       N/A       |       N/A        |
+    |     TRUE        |     FALSE       |   permuted(i)   |       N/A        |
+    |     FALSE       |     TRUE        |   permuted(j)   |       N/A        |
+    |     TRUE        |     TRUE        |   permuted(i)   |    permuted(j)   |
+    |------------------------------------------------------------------------|
+    Note: if both cliques do not participate, this function returns `nullptr`.
+    @pre clique_permutation.domain_size() == per_clique_known_dofs.size().
+    @pre first_clique() < clique_permutation.domain_size().
+    @pre if num_cliques() > 1,
+      second_clique() < clique_permutation.domain_size().
+  */
+  std::unique_ptr<SapConstraint<T>> MakeReduced(
+      const PartialPermutation& clique_permutation,
+      const std::vector<std::vector<int>>& per_clique_known_dofs) const;
 
  protected:
   /* Protected copy construction is enabled for sub-classes to use in their
--- a/multibody/contact_solvers/sap/sap_contact_problem.cc
+++ b/multibody/contact_solvers/sap/sap_contact_problem.cc
@@ -6,6 +6,7 @@
 #include "drake/common/drake_throw.h"
 #include "drake/common/eigen_types.h"
 #include "drake/multibody/contact_solvers/sap/contact_problem_graph.h"
+#include "drake/multibody/plant/slicing_and_indexing.h"
 
 namespace drake {
 namespace multibody {
@@ -59,6 +60,57 @@ std::unique_ptr<SapContactProblem<T>> SapContactProblem<T>::Clone() const {
     clone->AddConstraint(c.Clone());
   }
   return clone;
+}
+
+template <typename T>
+std::unique_ptr<SapContactProblem<T>> SapContactProblem<T>::MakeReduced(
+    const std::vector<int>& known_free_motion_dofs,
+    const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
+    ReducedMapping* mapping) const {
+  DRAKE_ASSERT_VOID(drake::multibody::internal::DemandIndicesValid(
+      known_free_motion_dofs, num_velocities()));
+  DRAKE_DEMAND(ssize(per_clique_known_free_motion_dofs) == num_cliques());
+  for (int i = 0; i < num_cliques(); ++i) {
+    DRAKE_ASSERT_VOID(drake::multibody::internal::DemandIndicesValid(
+        per_clique_known_free_motion_dofs[i], num_velocities(i)));
+  }
+  DRAKE_DEMAND(mapping != nullptr);
+
+  mapping->clique_permutation = PartialPermutation(num_cliques());
+  mapping->constraint_permutation = PartialPermutation(num_constraints());
+
+  // Project v_star and A matrices to reduced DOFs.
+  VectorX<T> v_star_reduced =
+      drake::multibody::internal::ExcludeRows(v_star_, known_free_motion_dofs);
+
+  std::vector<MatrixX<T>> A_reduced;
+  A_reduced.reserve(A_.size());
+  for (int i = 0; i < num_cliques(); ++i) {
+    // Clique participates if at least one of its dofs is not locked.
+    if (ssize(per_clique_known_free_motion_dofs[i]) < num_velocities(i)) {
+      A_reduced.push_back(drake::multibody::internal::ExcludeRowsCols(
+          A_[i], per_clique_known_free_motion_dofs[i]));
+      mapping->clique_permutation.push(i);
+    }
+  }
+
+  // Construct a problem with reduced parameters.
+  std::unique_ptr<SapContactProblem> problem =
+      std::make_unique<SapContactProblem<T>>(time_step(), std::move(A_reduced),
+                                             std::move(v_star_reduced));
+
+  // Make reduced constraints.
+  for (int i = 0; i < num_constraints(); ++i) {
+    std::unique_ptr<SapConstraint<T>> c = get_constraint(i).MakeReduced(
+        mapping->clique_permutation, per_clique_known_free_motion_dofs);
+
+    if (c) {
+      mapping->constraint_permutation.push(i);
+      problem->AddConstraint(std::move(c));
+    }
+  }
+
+  return problem;
 }
 
 template <typename T>
--- a/multibody/contact_solvers/sap/sap_contact_problem.h
+++ b/multibody/contact_solvers/sap/sap_contact_problem.h
@@ -13,6 +13,13 @@ namespace multibody {
 namespace contact_solvers {
 namespace internal {
 
+/* Struct returned by SapContactProblem::MakeReduced() which stores the mapping
+   between the original and reduced problems.*/
+struct ReducedMapping {
+  PartialPermutation clique_permutation;
+  PartialPermutation constraint_permutation;
+};
+
 /* In the SAP formulation of contact the state of a mechanical system is
 advanced from the previous state x₀ = [q₀,v₀] to the next state x = [q,v]. The
 SAP formulation linearizes the discrete time dynamics of the mechanical system
@@ -116,6 +123,39 @@ class SapContactProblem {
 
   /* Returns a deep-copy of `this` instance. */
   std::unique_ptr<SapContactProblem<T>> Clone() const;
+
+  /* Makes a "reduced" contact problem given the DOFs specified in
+    `known_free_motion_dofs` are known to equal the free-motion velocities.
+    That is, for an i-th DOF in `known_free_motion_dofs`, we know that vᵢ = vᵢ*.
+    `per_clique_known_free_motion_dofs` contains the same information stored in
+    `known_free_motion_dofs`, but expressed per-clique and in clique local
+    indices. e.g. If clique 0 corresponds to global velocity indices {5, 6, 7},
+    then its local velocity indices are {0, 1, 2}. If `known_free_motion_dofs` =
+    {6}, i.e. we know v₆ = v₆*, then `per_clique_known_free_motion_dofs[0] =
+    {1}` containing the known clique local index 1, corresponding to the known
+    global velocity index 6.
+
+    @param[in] known_free_motion_dofs Specifies known DOFs to be eliminated.
+    i ∈ known_free_motion_dofs specifies the i-th DOF.
+    @param[in] per_clique_known_free_motion_dofs Specifies known DOFs to be
+    eliminated, per clique. i ∈ per_clique_known_free_motion_dofs[c] specifies
+    the i-th DOF of the c-th clique. This parameter should contain the same
+    information in `known_free_motion_dofs`, but transformed to clique local
+    indices.
+    @param[out] mapping On output it will store information to map DOFs and
+    constraints between the original and reduced problems in a ReducedMapping.
+
+    @pre known_free_motion_dofs is a strict ordered subset of
+         [0, ..., num_velocities()-1].
+    @pre per_clique_known_free_motion_dofs.size() == num_cliques().
+    @pre per_clique_known_free_motion_dofs[c] is a strict ordered subset of
+         [0, ..., num_velocities(c)-1].
+    @pre mapping != nullptr.
+  */
+  std::unique_ptr<SapContactProblem<T>> MakeReduced(
+      const std::vector<int>& known_free_motion_dofs,
+      const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
+      ReducedMapping* mapping) const;
 
   /* TODO(amcastro-tri): consider constructor API taking std::vector<VectorX<T>>
    for v_star. It could be useful for deformables. */
--- a/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc
+++ b/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc
@@ -259,6 +259,174 @@ GTEST_TEST(ContactProblem, Clone) {
   EXPECT_EQ(graph.num_constraint_equations(), 17);
 }
 
+/* We test reducing the SapContactProblem. The graph setup sketched below
+(having the same semantics as the graph described in AddConstraints())
+corresponds to the graph of the contact problem that results from locking DoFs
+{0,1,2,3,4,9} in the contact problem produced by AddConstraints(). The indices
+in the graph below are the new mapped indices of the reduced problem. The
+following tables explain the mapping between original clique and constraint
+indices and their corresponding indices in the reduced problem:
+
+┌────────────┬────────────────────┐┌─────────────────────────────────────────┐
+│ clique idx │ reduced clique idx ││ constraint idx │ reduced constraint idx │
+├────────────│────────────────────┤├────────────────│────────────────────────┤
+│     0      │         N/A        ││       0        │           0            │
+│     1      │         N/A        ││       1        │          N/A           │
+│     2      │          0         ││       2        │           1            │
+│     3      │          1         ││       3        │           2            │
+└────────────┴────────────────────┘│       4        │           3            │
+                                   └────────────────┴────────────────────────┘
+
+Locking these DoFs has the effect of eliminating clique 0 and 1 from the
+problem. One constraint from the original problem will be completely removed.
+Other constraints that involved one of the eliminated clique now become single
+clique constraints:
+
+    0[0, 1, 2, 3](1, 6, 2, 5)
+             ┌─┐
+             │ │
+            ┌┴─┴┐     ┌───┐
+            │ 1 │     │ 0 │
+            └───┘     └───┘
+
+TODO(joemasterjohn): Make a fixture to consolidate the creation of this problem
+and its constraints.
+*/
+GTEST_TEST(ContactProblem, MakeReduced) {
+  const double time_step = 0.01;
+  const std::vector<MatrixXd> A{S22, S33, S44, S22};
+  const VectorXd v_star = VectorXd::LinSpaced(11, 1.0, 11.0);
+  SapContactProblem<double> problem(time_step, std::move(A), std::move(v_star));
+  AddConstraints(&problem);
+
+  // Lock all the dofs of clique 0 and 1. Lock velocity index 9, local index 1
+  // of clique 3.
+  const std::vector<int> locked_indices{0, 1, 2, 3, 4, 9};
+  const std::vector<std::vector<int>> clique_locked_indices{
+      {0, 1}, {0, 1, 2}, {}, {1}};
+  ReducedMapping mapping;
+  std::unique_ptr<SapContactProblem<double>> reduced_problem =
+      problem.MakeReduced(locked_indices, clique_locked_indices, &mapping);
+
+  EXPECT_EQ(reduced_problem->num_cliques(), 2);
+  EXPECT_EQ(reduced_problem->num_constraints(), 4);
+  EXPECT_EQ(reduced_problem->num_constraint_equations(), 14);
+  // Verify some expected data.
+  EXPECT_EQ(reduced_problem->v_star().size(), 5);
+  // Two cliques eliminated, so two remain.
+  EXPECT_EQ(reduced_problem->dynamics_matrix().size(), 2);
+  // Original clique 2 (reduced clique 0) is unaffected.
+  EXPECT_EQ(reduced_problem->dynamics_matrix()[0], S44);
+  // Original clique 3 (reduced clique 1) has only 1 dof now.
+  EXPECT_EQ(reduced_problem->dynamics_matrix()[1].size(), 1);
+  EXPECT_EQ(reduced_problem->dynamics_matrix()[1](0, 0), 2);
+
+  const ContactProblemGraph& graph = reduced_problem->graph();
+  EXPECT_EQ(graph.num_cliques(), 2);
+  EXPECT_EQ(graph.num_constraints(), 4);
+  EXPECT_EQ(graph.num_clusters(), 1);
+  EXPECT_EQ(graph.num_constraint_equations(), 14);
+
+  /* Clique permutation. We expect the first two cliques to be eliminated. */
+  EXPECT_EQ(mapping.clique_permutation.domain_size(), problem.num_cliques());
+  EXPECT_EQ(mapping.clique_permutation.permuted_domain_size(), 2);
+  EXPECT_FALSE(mapping.clique_permutation.participates(0));
+  EXPECT_FALSE(mapping.clique_permutation.participates(1));
+  EXPECT_TRUE(mapping.clique_permutation.participates(2));
+  EXPECT_TRUE(mapping.clique_permutation.participates(3));
+
+  EXPECT_EQ(mapping.clique_permutation.permuted_index(2), 0);
+  EXPECT_EQ(mapping.clique_permutation.permuted_index(3), 1);
+
+  /* Constraint permutation. We expect the second constraint between the two
+     eliminated cliques (0 and 1) to be eliminated. */
+  EXPECT_EQ(mapping.constraint_permutation.domain_size(),
+            problem.num_constraints());
+  EXPECT_EQ(mapping.constraint_permutation.permuted_domain_size(), 4);
+  EXPECT_TRUE(mapping.constraint_permutation.participates(0));
+  EXPECT_FALSE(mapping.constraint_permutation.participates(1));
+  EXPECT_TRUE(mapping.constraint_permutation.participates(2));
+  EXPECT_TRUE(mapping.constraint_permutation.participates(3));
+  EXPECT_TRUE(mapping.constraint_permutation.participates(4));
+
+  EXPECT_EQ(mapping.constraint_permutation.permuted_index(0), 0);
+  EXPECT_EQ(mapping.constraint_permutation.permuted_index(2), 1);
+  EXPECT_EQ(mapping.constraint_permutation.permuted_index(3), 2);
+  EXPECT_EQ(mapping.constraint_permutation.permuted_index(4), 3);
+
+  /* Verify that the each constraint's index mapping is as expected and
+     described by the graph. For each constraint, the table lists the original
+     problem constraint properties and the reduced problem properties.*/
+
+  /*
+    ┌──────────┬───────┬───────────────┬────────────────┬─────────────────┐
+    │ problem  │ index │ num_equations │ first_clique() │ second_clique() │
+    ├──────────│───────│───────────────│────────────────│─────────────────┤
+    │ original │   0   │      1        │       3        │      N/A        │
+    │ reduced  │   0   │      1        │       1        │      N/A        │
+    └──────────┴───────┴───────────────┴────────────────┴─────────────────┘
+  */
+  {
+    const SapConstraint<double>& c = reduced_problem->get_constraint(0);
+    EXPECT_EQ(c.num_cliques(), 1);
+    EXPECT_EQ(c.first_clique(), mapping.clique_permutation.permuted_index(3));
+    EXPECT_EQ(c.num_constraint_equations(), 1);
+    EXPECT_EQ(c.first_clique_jacobian().cols(),
+              reduced_problem->num_velocities(c.first_clique()));
+  }
+
+  /*
+    ┌──────────┬───────┬───────────────┬────────────────┬─────────────────┐
+    │ problem  │ index │ num_equations │ first_clique() │ second_clique() │
+    ├──────────│───────│───────────────│────────────────│─────────────────┤
+    │ original │   2   │      6        │       0        │       3         │
+    │ reduced  │   1   │      6        │       1        │      N/A        │
+    └──────────┴───────┴───────────────┴────────────────┴─────────────────┘
+  */
+  {
+    const SapConstraint<double>& c = reduced_problem->get_constraint(1);
+    EXPECT_EQ(c.num_cliques(), 1);
+    EXPECT_EQ(c.first_clique(), mapping.clique_permutation.permuted_index(3));
+    EXPECT_EQ(c.num_constraint_equations(), 6);
+    EXPECT_EQ(c.first_clique_jacobian().cols(),
+              reduced_problem->num_velocities(c.first_clique()));
+  }
+
+  /*
+    ┌──────────┬───────┬───────────────┬────────────────┬─────────────────┐
+    │ problem  │ index │ num_equations │ first_clique() │ second_clique() │
+    ├──────────│───────│───────────────│────────────────│─────────────────┤
+    │ original │   3   │      2        │       1        │       3         │
+    │ reduced  │   2   │      2        │       1        │      N/A        │
+    └──────────┴───────┴───────────────┴────────────────┴─────────────────┘
+  */
+  {
+    const SapConstraint<double>& c = reduced_problem->get_constraint(2);
+    EXPECT_EQ(c.num_cliques(), 1);
+    EXPECT_EQ(c.first_clique(), mapping.clique_permutation.permuted_index(3));
+    EXPECT_EQ(c.num_constraint_equations(), 2);
+    EXPECT_EQ(c.first_clique_jacobian().cols(),
+              reduced_problem->num_velocities(c.first_clique()));
+  }
+
+  /*
+    ┌──────────┬───────┬───────────────┬────────────────┬─────────────────┐
+    │ problem  │ index │ num_equations │ first_clique() │ second_clique() │
+    ├──────────│───────│───────────────│────────────────│─────────────────┤
+    │ original │   4   │      5        │       0        │       3         │
+    │ reduced  │   3   │      5        │       1        │      N/A        │
+    └──────────┴───────┴───────────────┴────────────────┴─────────────────┘
+  */
+  {
+    const SapConstraint<double>& c = reduced_problem->get_constraint(3);
+    EXPECT_EQ(c.num_cliques(), 1);
+    EXPECT_EQ(c.first_clique(), mapping.clique_permutation.permuted_index(3));
+    EXPECT_EQ(c.num_constraint_equations(), 5);
+    EXPECT_EQ(c.first_clique_jacobian().cols(),
+              reduced_problem->num_velocities(c.first_clique()));
+  }
+}
+
 }  // namespace
 }  // namespace internal
 }  // namespace contact_solvers
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -394,6 +394,7 @@ drake_cc_library(
     deps = [
         "//common:default_scalars",
         "//common:essential",
+        "//multibody/contact_solvers:matrix_block",
     ],
 )
 
@@ -538,6 +539,7 @@ drake_cc_googletest(
     deps = [
         ":slicing_and_indexing",
         "//common:essential",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/multibody/plant/slicing_and_indexing.cc
+++ b/multibody/plant/slicing_and_indexing.cc
@@ -1,9 +1,11 @@
 #include "drake/multibody/plant/slicing_and_indexing.h"
 
 #include <set>
+#include <utility>
 
 #include "drake/common/default_scalars.h"
 #include "drake/common/eigen_types.h"
+#include "drake/common/ssize.h"
 
 // TODO(rpoyner-tri): the Select*() and Expand*() functions below can be
 // removed and replaced with arbitrary indexing once Eigen 3.4 is our minimum
@@ -47,6 +49,50 @@ MatrixX<T> SelectRowsCols(const MatrixX<T>& M,
 }
 
 template <typename T>
+MatrixX<T> ExcludeRowsCols(const MatrixX<T>& M,
+                           const std::vector<int>& indices) {
+  DRAKE_DEMAND(M.rows() == M.cols());
+  DRAKE_ASSERT_VOID(DemandIndicesValid(indices, M.rows()));
+  if (indices.size() == 0) {
+    return M;
+  }
+
+  MatrixX<T> result(M.rows() - indices.size(), M.cols() - indices.size());
+
+  int r_index = 0;          // Keeps track of the row of result being filled.
+  int r_exclude_index = 0;  // Index into indices of the next row to exclude.
+  for (int i = 0; i < M.rows(); ++i) {
+    // If we still have rows to exclude and are on the next row to exclude,
+    // skip this row.
+    if (r_exclude_index < ssize(indices) && i == indices[r_exclude_index]) {
+      ++r_exclude_index;
+      continue;
+    }
+    // Keeps track of the column of result being filled.
+    int c_index = 0;
+    // Index into indices of the next column to exclude.
+    int c_exclude_index = 0;
+    for (int j = 0; j < M.cols(); ++j) {
+      // If we still have columns to exclude and are on the next columns to
+      // exclude, skip this column.
+      if (c_exclude_index < ssize(indices) && j == indices[c_exclude_index]) {
+        ++c_exclude_index;
+        continue;
+      }
+      result(r_index, c_index) = M(i, j);
+      ++c_index;
+    }
+    DRAKE_ASSERT(c_index == result.cols());
+    DRAKE_ASSERT(c_exclude_index == ssize(indices));
+    ++r_index;
+  }
+
+  DRAKE_ASSERT(r_index == result.rows());
+  DRAKE_ASSERT(r_exclude_index == ssize(indices));
+  return result;
+}
+
+template <typename T>
 MatrixX<T> SelectCols(const MatrixX<T>& M, const std::vector<int>& indices) {
   DRAKE_ASSERT_VOID(DemandIndicesValid(indices, M.cols()));
   const int selected_count = indices.size();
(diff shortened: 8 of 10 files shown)


joemasterjohn (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NWF4I8W5Ush73qDvCpX:b5xll3r)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri and @joemasterjohn)

amcastro-tri (commented):
Mostly to big picture comments:
 1. Documentation/semantics. It has to be cristal clear this "reduction" is only true when vᵢ = vᵢ* and we can eliminate DOFs (our original sin)
 2. Why mutation? what about `MakeReduced()` that makes an entirely new problem without ugly mutations. After all, that's exactly what you are doing, with the same amount of work and dynamic allocations.

> Reviewed 2 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NWhKZrZ9nQmAVPgi6DZ:b-uokqk3)* status: 10 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 127 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhCost2kH5bJ8vl9zY:-NWhCost2kH5bJ8vl9zZ:bsdmozd) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L127)):*
> ```objc
>   std::unique_ptr<SapContactProblem<T>> Clone() const;
> 
>   // Project the contact problem to a reduced DOF contact
> ```

TBH I do not like the "project" terminology, it's just wrong. 
What we are doing really is "eliminating known dofs", that's it. Reading these docs stand alone really say nothing. We should simply state the math.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhCk8Q5xk26gPDjhyi:-NWhCk8Q5xk26gPDjhyj:bje2vg1) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L129)):*
> ```objc
>   // Project the contact problem to a reduced DOF contact
>   // problem according to the set of contributing/non-contributing DOFs included
>   // in `locked_indices` and `clique_locked_indices` coming from joint locking.
> ```

minor, use `/* */` within this file for consistency
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhD7gD4NPPOyLiCS6P:-NWhD7gD4NPPOyLiCS6Q:bjq7kio) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*
> ```objc
>   // in `locked_indices` and `clique_locked_indices` coming from joint locking.
>   ReducedMapping Reduce(std::vector<int> locked_indices,
>                         std::vector<std::vector<int>> clique_locked_indices);
> ```

The docs are not clear at all on what these two arguments are and why they are different.  Sizes? preconditions? 
Don't we have all the information we need in `clique_locked_indices`?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhDc956wzSwFD8l9Nh:-NWhDc956wzSwFD8l9Ni:b-1tyt5a) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*
> ```objc
>   // in `locked_indices` and `clique_locked_indices` coming from joint locking.
>   ReducedMapping Reduce(std::vector<int> locked_indices,
>                         std::vector<std::vector<int>> clique_locked_indices);
> ```

I'm not sure about the semantics of this. I don't think we should mutate the problem but rather create a new one?
I am thinking something like this:

```c++
/* Makes a "reduced" contact problem given the per-clique DOFs specified in
 `known_free_motion_dofs` are known to equal the free-motion velocities.
 That is, for an i-th DOF in `known_free_motion_dofs`, we know that vᵢ = vᵢ*.

  @param[in] known_free_motion_dofs Specifies known DOFs to be eliminated. i =
  known_free_motion_dofs[c] specifies the i-th DOFs for the c-th clique.
  @param[out] mapping On output it will store information to map DOFs and
  constraints between the original and reduced problems. @see ReducedMapping for
  further details.

  @pre known_free_motion_dofs.size() equals num_cliques().  
  @pre known_free_motion_dofs[c].size() equals num_velocities(c).
*/
std::unique_ptr<SapContactProblem<T>> MakeReduced(const std::vector<std::vector<int>>& known_free_motion_dofs, ReducedMapping* mapping) const;
```
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGVeCAjiQs8QE_FcB:-NWhGVeCAjiQs8QE_FcC:bm49zrh) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L68)):*
> ```cpp
> ReducedMapping SapContactProblem<T>::Reduce(
>     std::vector<int> locked_indices,
>     std::vector<std::vector<int>> clique_locked_indices) {
> ```

why passing by copy?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGQpw15y2A88UmmYI:-NWhGQpw15y2A88UmmYJ:b-wp0zzf) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L69)):*
> ```cpp
>     std::vector<int> locked_indices,
>     std::vector<std::vector<int>> clique_locked_indices) {
>   const int num_cliques = static_cast<int>(clique_locked_indices.size());
> ```

nit, we can use `ssize`.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGo0i8DZfSnbuGnFD:-NWhGo0i8DZfSnbuGnFE:b-udmcyk) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L83)):*
> ```cpp
>   for (int i = 0; i < num_cliques; ++i) {
>     // Clique participates if at least one of its dofs is not locked.
>     if (static_cast<int>(clique_locked_indices[i].size()) < A_[i].rows()) {
> ```

nit, `ssize`
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhHNqV8r7TSgTU6T61:-NWhHNqV8r7TSgTU6T62:b-75ywm7) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L90)):*
> ```cpp
>   }
> 
>   A_reduced.resize(mapping.clique_permutation.permuted_domain_size());
> ```

why is this needed?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhJPX750CJNKYRRsSV:-NWhJPX750CJNKYRRsSW:b-u47pjc) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L96)):*
> ```cpp
>       std::move(this->constraints_);
> 
>   this->Reset(A_reduced, v_star_reduced);
> ```

for a new SapContactProblem, you could move into it.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 137 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhJfgv-f9sKjrJOy_g:-NWhJfgv-f9sKjrJOy_h:bxeok05) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L137)):*
> ```cpp
>       c->set_first_clique_jacobian(jacobians[0]);
>       if (idx > 1) {
>         c->set_second_clique(cliques[1]);
> ```

I gave this some thought over the weekend. 
For the `MakeReduced` semantics. What if the constraints also had a `MakeReduced()` or `CloneReduced()` method that made a new constraint without mutating the original one?
A signature like:
```c++
std::unique_ptr<Constraint<T>> Constraint::MakeReduced(const std::vector<int>& known_zero_dofs) const;
```

joemasterjohn (commented):
Very valid criticisims, thanks. I've incorporated your suggestions in this revision, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NWrsSW44VwHeo5pZw14:bom7423)* status: 7 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @amcastro-tri and @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 127 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhCost2kH5bJ8vl9zY:-NWsBmpO7NQ5v0cq7ZKx:b-ra0fdy) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L127)):*

TBH I do not like the "project" terminology, it's just wrong. 
What we are doing really is "eliminating known dofs", that's it. Reading these docs stand alone really say nothing. We should simply state the math.



</blockquote></details>

Done. Took out the wrong terminology and worded it with "known free motion velocities".
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhCk8Q5xk26gPDjhyi:-NWsBwWLCTXF2pm8Te7d:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L129)):*

minor, use `/* */` within this file for consistency
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhDc956wzSwFD8l9Nh:-NWrrS3e08Ea85gN7zmW:b-q09aq3) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*

I'm not sure about the semantics of this. I don't think we should mutate the problem but rather create a new one?
I am thinking something like this:

```c++
/* Makes a "reduced" contact problem given the per-clique DOFs specified in
 `known_free_motion_dofs` are known to equal the free-motion velocities.
 That is, for an i-th DOF in `known_free_motion_dofs`, we know that vᵢ = vᵢ*.

  @param[in] known_free_motion_dofs Specifies known DOFs to be eliminated. i =
  known_free_motion_dofs[c] specifies the i-th DOFs for the c-th clique.
  @param[out] mapping On output it will store information to map DOFs and
  constraints between the original and reduced problems. @see ReducedMapping for
  further details.

  @pre known_free_motion_dofs.size() equals num_cliques().  
  @pre known_free_motion_dofs[c].size() equals num_velocities(c).
*/
std::unique_ptr<SapContactProblem<T>> MakeReduced(const std::vector<std::vector<int>>& known_free_motion_dofs, ReducedMapping* mapping) const;
```
</blockquote></details>

Done. I like these semantics better.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhD7gD4NPPOyLiCS6P:-NWsBy6ZCH-aCv0gKFtl:b6gtsuc) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*

The docs are not clear at all on what these two arguments are and why they are different.  Sizes? preconditions? 
Don't we have all the information we need in `clique_locked_indices`?
</blockquote></details>

Done. Added docs for the arguments. We need both the dense locked indices and the per clique versions because v* is dense and A and the constraints are sparse.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGVeCAjiQs8QE_FcB:-NWsCOoSEpxsQvkZMftQ:bkr3rt2) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L68)):*

why passing by copy?
</blockquote></details>

Done. Just lazy typing.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGQpw15y2A88UmmYI:-NWsCuwX4rtpJtq4QwJM:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L69)):*

nit, we can use `ssize`.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhGo0i8DZfSnbuGnFD:-NWsD0ReAXcHXKJm8YzG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L83)):*

nit, `ssize`
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhHNqV8r7TSgTU6T61:-NWsD6VZ9l0uWHVVBXKt:bwqnrw3) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L90)):*

why is this needed?
</blockquote></details>

Done. Not needed.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhJPX750CJNKYRRsSV:-NWsDPI04uWzD58iv74o:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L96)):*

for a new SapContactProblem, you could move into it.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 137 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NWhJfgv-f9sKjrJOy_g:-NWrmgtt4xUA97l_KRVF:bmcmtjm) ([raw file](https://github.com/RobotLocomotion/drake/blob/610707943dc9873b509d8b050072e04e3987f579/multibody/contact_solvers/sap/sap_contact_problem.cc#L137)):*

I gave this some thought over the weekend. 
For the `MakeReduced` semantics. What if the constraints also had a `MakeReduced()` or `CloneReduced()` method that made a new constraint without mutating the original one?
A signature like:
```c++
std::unique_ptr<Constraint<T>> Constraint::MakeReduced(const std::vector<int>& known_zero_dofs) const;
```
</blockquote></details>

Done. This logic now lives in `SapConstraint::MakeReduced`.

amcastro-tri (commented):
Excellent!
Mostly nits on docs and pre-conditions.

> Reviewed 2 of 13 files at r2, 2 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NX6ezAs1SzWCy2ejsHF:byq5pu7)* status: 17 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 162 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6ZF2u0dYOY3gR_nTC:-NX6ZF2u0dYOY3gR_nTD:bryybzi) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L162)):*
> ```objc
>   std::unique_ptr<SapConstraint<T>> Clone() const { return DoClone(); }
> 
>   /* Creates a clone, cᵣ, of this constraint, c, modified by the clique and dof
> ```

nit, maybe the intro to this text should be more like:
```
Creates a "reduced" clone of this constraint by removing known DoFs from the constraint's Jacobian. That is, the newly reduced constraint will have this constraint's Jacobian excluding columns for known DoFs.
```
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 165 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6Y1cfE_1hVpa7P48K:-NX6Y1cg5ABDg-TYdtQa:b-yp7q5s) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L165)):*
> ```objc
>     permutation given by `clique_permutation` and `per_clique_locked_dofs`. The
>     following table breaks down how the returned constraint relates to this
>     constraint where participates(x) is true iff index x participates in the
> ```

it took me a bit to parse this and even go to the source.
Maybe say "where particapates(c) = clique_permuation.participates(c)`"
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 173 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6YUel91EO_MqOF-RN:-NX6YUel91EO_MqOF-RO:bxu89v9) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L173)):*
> ```objc
>     |------------------------------------------------------------------------|
>     |     FALSE       |     FALSE       |         No constraint made.        |
>     |     TRUE        |     FALSE       |   permuted(i)   |        -1        |
> ```

nit, instead of `-1`, which is not strictly true, maybe just use `N/A`?
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 180 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6YmgMDL1UW4BH73Ra:-NX6YmgMDL1UW4BH73Rb:bcm5xmo) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L180)):*
> ```objc
>     The constraint Jacobian for a participating clique is modified by removing
>     all columns that correspond to dofs included in
>     `per_clicked_locked_dofs[clique_index]`.
> ```

nit, maybe "The constraint Jacobian for the new reduced constraint corresponds to the Jacobian for this constraint excluding columns indexed by `per_clicked_locked_dofs`."
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 185 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6_za55mrNI16houJY:-NX6_za55mrNI16houJZ:b-yhyggv) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L185)):*
> ```objc
>   */
>   std::unique_ptr<SapConstraint<T>> MakeReduced(
>       const PartialPermutation& clique_permutation,
> ```

A few preconditions are missing:
 1. `clique_permutation.domain_size() == per_clique_known_dofs.size()`
 2.  first_clique() < clique_permutation.domain_size() and second_clique() < clique_permutation.domain_size(). Consider adding DRAKE_DEMAND for both.
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 186 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6_b6H-u9aLyLJ2VWu:-NX6_b6H-u9aLyLJ2VWv:bsvbrxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.h#L186)):*
> ```objc
>   std::unique_ptr<SapConstraint<T>> MakeReduced(
>       const PartialPermutation& clique_permutation,
>       const std::vector<std::vector<int>>& per_clique_locked_dofs) const;
> ```

The "locked" notation does not show in the `SapContactProblem::MakeReduced()`.
To keep things consistent, I'd probably just name vector `per_clique_known_dofs`.
___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 78 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6Xo8F7NWubGHxhKNy:-NX6Xo8F7NWubGHxhKNz:bgzrpms) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_constraint.cc#L78)):*
> ```cpp
>         drake::multibody::internal::ExcludeCols(
>             second_clique_jacobian(), per_clique_locked_dofs[second_clique()]));
> ```

minor, blank line
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 16 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX5rV6l804DUZAnn9Bo:-NX5rV6l804DUZAnn9Bp:b-94vz3g) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.h#L16)):*
> ```objc
> namespace internal {
> 
> /* Simple struct to keep the mapping from a SapContactProblem to a reduced dof
> ```

"simple" is very subjective. I'd just say: "Struct returned by SapContactProblem::MakeReduced() store the mapping between the original and reduced problems"
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 137 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX5s7BJDS2CtIL0dGy6:-NX5s7BJDS2CtIL0dGy7:b-eoxef6) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.h#L137)):*
> ```objc
>     the i-th DOF of the c-th clique.
>     @param[out] mapping On output it will store information to map DOFs and
>     constraints between the original and reduced problems. @see ReducedMapping
> ```

nit right now this `@see` creates an infinite loop. That is, if I don't find something here, I go to ReducedMappina and I find nothing. I'd simply remove the `@see`. The reason you have it here is simply to reference the struct, but C++ already does that for you.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 143 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX5siH23JjBFHm4pIH7:-NX5siH23JjBFHm4pIH8:b2ponx5) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.h#L143)):*
> ```objc
>   */
>   std::unique_ptr<SapContactProblem<T>> MakeReduced(
>       const std::vector<int>& known_free_motion_dofs,
> ```

nit, what about `knonw_free_motion_dofs` pre conditions?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 144 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX5swkI4a_yIDglW5qU:-NX5swkI4a_yIDglW5qV:b-rxmcpm) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.h#L144)):*
> ```objc
>   std::unique_ptr<SapContactProblem<T>> MakeReduced(
>       const std::vector<int>& known_free_motion_dofs,
>       const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
> ```

clearly `known_free_motion_dofs` and `per_clique_known_free_motion_dofs` are related. You could even build `known_free_motion_dofs` from `per_clique_known_free_motion_dofs`. I'd at least state that relationship in the docs.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 145 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX5smfv5a2m3muJZ8aD:-NX5smfv5a2m3muJZ8aE:bfj43r) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.h#L145)):*
> ```objc
>       const std::vector<int>& known_free_motion_dofs,
>       const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
>       ReducedMapping* mapping) const;
> ```

can `mapping` equal nullptr?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 67 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6bbly7hxhOIK4Tn7d:-NX6bbly7hxhOIK4Tn7e:b-597b05) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.cc#L67)):*
> ```cpp
> template <typename T>
> std::unique_ptr<SapContactProblem<T>> SapContactProblem<T>::MakeReduced(
>     const std::vector<int>& known_free_motion_dofs,
> ```

DRAKE_DEMAND(known_free_motions_dofs.size() <= num_velocities());
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 70 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6aMmQ9eMsqKIvZ-tD:-NX6aMmQ9eMsqKIvZ-tE:bivs0po) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.cc#L70)):*
> ```cpp
>     const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
>     ReducedMapping* mapping) const {
>   const int num_cliques = ssize(per_clique_known_free_motion_dofs);
> ```

the number of cliques is known.
`DRAKE_DEMAND(ssize(per_clique_known_dofs) == num_cliques())`
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 84 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6b3Yo0KPn67G5UigN:-NX6b3Yo0KPn67G5UigO:bryszd0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.cc#L84)):*
> ```cpp
>   for (int i = 0; i < num_cliques; ++i) {
>     // Clique participates if at least one of its dofs is not locked.
>     if (ssize(per_clique_known_free_motion_dofs[i]) < A_[i].rows()) {
> ```

consider DRAKE_DEMAND(ssize(per_clique_known_free_motion_dofs[i]) <= num_velocities(i));
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 84 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6bPPj2bwuCNUIOHdq:-NX6bPPj2bwuCNUIOHdr:bzeq2ia) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/sap_contact_problem.cc#L84)):*
> ```cpp
>   for (int i = 0; i < num_cliques; ++i) {
>     // Clique participates if at least one of its dofs is not locked.
>     if (ssize(per_clique_known_free_motion_dofs[i]) < A_[i].rows()) {
> ```

nit, consider `num_velocities(i)` instead of A_[i] for readability.
___
*[`multibody/plant/slicing_and_indexing.cc` line 59 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NX6eg3l0MbhRmqpO95-:-NX6eg3l0MbhRmqpO950:br3qmcy) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/plant/slicing_and_indexing.cc#L59)):*
> ```cpp
>     return M;
>   }
>   std::vector<int> selected_indices;
> ```

these select require an additional allocation of a vector of ints.
Per this PR you need the "ExcludeFoo", not the "SelectFoo". Is there a way to write the "ExcludeFoo" methods without resorting to the "SelctFoo" counterparts and avoid these extra heap allocations?

amcastro-tri (commented):
Feature review pass done.

> Reviewed 1 of 9 files at r1, 2 of 8 files at r3, 2 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXBhGr05OwBsjUQ8nE9:b75uegm)* status: 25 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 262 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXB_MX4EHZtj9QscknG:-NXB_MX4EHZtj9QscknH:b-t36ebe) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L262)):*
> ```cpp
> }
> 
> /* We test reducing the ContactProblem. The graph setup sketched below
> ```

nit, SapContactProblem?
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 263 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXB_Qom3qs8cGc4r9Y5:-NXB_Qom3qs8cGc4r9Y6:brl91eh) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L263)):*
> ```cpp
> 
> /* We test reducing the ContactProblem. The graph setup sketched below
> (with the same semantics as the graph described in `AddConstraints`) is the
> ```

this description is confusing. I think what you really want to say is that the sketch below corresponds to the graph of the contact problem that results from locking DoFs {0, 1, 2, 3, 4, 9} in the the contact problem produced by AddConstraints().

Also, consider making this a fixture, that way the the particulars of the problem, AddConstraints() and the locking are all connected in the same fixture and hopefully it's easier to documetnt
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 265 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBaIAhCZZyEPJ_s53z:-NXBaIAhCZZyEPJ_s54-:bf6hpf3) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L265)):*
> ```cpp
> (with the same semantics as the graph described in `AddConstraints`) is the
> resulting graph from locking dofs {0, 1, 2, 3, 4, 9}. This has the effect of
> eliminating clique 0 and 1 from the problem. One constraint from the original
> ```

Explain that the clique indexes below are the new indexes in the reduced problem.
I'd probably add a lil table to describe the mapping of cliques and constraints between the original and reduced problem.



_Code quote:_
```cpp
eliminating clique 0 and 1 
```
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 275 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBbaPXDpPyYv9Nh8sD:-NXBbaPXDpPyYv9Nh8sE:bx4o0t5) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L275)):*
> ```cpp
>          ┌──┤ 1 │     │ 0 │
>   0[2](2)└──└┬─┬┘     └───┘
>              | │
> ```

If we use the same labeling as in AddConsraints(), there should be a single edge for the one and only single cluster in the new reduced problem.
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 315 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBeWNr7kQMtS3VUIX1:-NXBeWNr7kQMtS3VUIX2:b-y1mi4a) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L315)):*
> ```cpp
> 
>   /* Clique permutation. We expect the first two cliques to be eliminated. */
>   EXPECT_EQ(mapping.clique_permutation.permuted_domain_size(), 2);
> ```

what about `domain_size()`?
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 325 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBf85w3jEBKX40c0Jx:-NXBf85w3jEBKX40c0Jy:bq9j0qa) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L325)):*
> ```cpp
> 
>   /* Constraint permutation. We expect the second constraint between the two
>    * eliminated cliques (0 and 1) to be eliminated. */
> ```

minor, remove extra "*"
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 326 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBfCx_821sSy3yeDWU:-NXBfCx_821sSy3yeDWV:bibqnb4) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L326)):*
> ```cpp
>   /* Constraint permutation. We expect the second constraint between the two
>    * eliminated cliques (0 and 1) to be eliminated. */
>   EXPECT_EQ(mapping.constraint_permutation.permuted_domain_size(), 4);
> ```

nit, ditto on `domain_size()`
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 349 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBgt3pBTtb5eS43QDF:-NXBgt3pBTtb5eS43QDG:b-9hl76y) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L349)):*
> ```cpp
>   }
> 
>   /* Original constraint: num_equations(6)  first_clique(0)  first_clique_nv(2)
> ```

I don't understand these comments.
I think you might want to say "Original constraint = 0", "Reduced constraint = xx", similarly "first_clique = 0"?
Maybe I got it wrong, but bottom line is that I couldn't parse it.

amcastro-tri (commented):
Thanks @joemasterjohn. :lgtm_strong: 
@rpoyner-tri for platform review please.

> Reviewed 1 of 9 files at r1, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXGq0pj4vISwaqKCvpM:bwvjy4u)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 349 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXBgt3pBTtb5eS43QDF:-NXGn9GI5X0-I4gyZyi3:b-7m74ox) ([raw file](https://github.com/RobotLocomotion/drake/blob/f643f5855e49116fa0ad5bb1541e44e9304a9248/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L349)):*

I reworked this labelling into a nicer table and explained in a comment. Let me know if it's still confusing.
</blockquote></details>

nice!
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 263 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXGjk5A7S1ELSF6W9Jb:-NXGjk5A7S1ELSF6W9Jc:b24216y) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L263)):*
> ```cpp
> 
> /* We test reducing the SapContactProblem. The graph setup sketched below
> (having the same semantics as the graph described in `AddConstraints`)
> ```

minor, you use`'AddConstraints'` (in single quotes) here but then `AddConstraints()` (without quotes, with parenthesis).
Personally I prefer the latter, but whichever it is choose one for consistency.
___
*[`multibody/plant/slicing_and_indexing.h` line 49 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXGoCpq9hOMO62NOFl_:-NXGoCpq9hOMO62NOFla:by1bnv0) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.h#L49)):*
> ```objc
> // @pre indices argument is valid according to DemandIndicesValid().
> template <typename T>
> contact_solvers::internal::MatrixBlock<T> SelectCols(
> ```

nit, do we still want this new select method? beyond the unit test, I believe it's no longer used.

joemasterjohn (commented):
Missed the assignment, but it's a bit late for @rpoyner-tri. +@sherm1 for platform review tomorrow, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXHQ_lcBniL-5fdpQ_0:b2iv5ss)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @sherm1)

sherm1 (commented):
Platform checkpoint.

> Reviewed 2 of 9 files at r1, 7 of 8 files at r3, 1 of 4 files at r4, 3 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXLtuA03ZO4Ops9nwfC:bjxpau6)* status: 8 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 167 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLlIl55vnq8HVOMpBK:-NXLlIl62Ho3xYmxjnX9:b-hvao54) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.h#L167)):*
> ```objc
>    table breaks down how the returned constraint relates to this constraint
>    where participates(c) = clique_permutation.participates(c) and
>    permuted(x) = clique_permutation.permuted_index(c). For convenience
> ```

minor: not clear what the `x` and `c` notation on this line means
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 177 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLm0r4D0Ht2RdNRhH0:-NXLm0r4D0Ht2RdNRhH1:b-mx7x4f) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.h#L177)):*
> ```objc
>     |     FALSE       |     TRUE        |   permuted(j)   |       N/A        |
>     |     TRUE        |     TRUE        |   permuted(i)   |    permuted(j)   |
>     |------------------------------------------------------------------------|
> ```

This table is very helpful documentation, thanks!
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 179 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLlwMlAd1NpD85UNWV:-NXLlwMlAd1NpD85UNWW:b-f6stre) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.h#L179)):*
> ```objc
>     |------------------------------------------------------------------------|
> 
>     Note: if both cliques do not participate, this function returns `nullptr`.
> ```

nit: extra leading space here and below
___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 74 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLm_BU-3LtEkpzJBay:-NXLm_BU-3LtEkpzJBaz:b-bl1xnl) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.cc#L74)):*
> ```cpp
> 
>   // Neither clique participates, no constraint made.
>   if (!first_participates && !second_participates) return nullptr;
> ```

BTW consider not cloning the constraint until after you've done this test
___
*[`multibody/plant/slicing_and_indexing.h` line 79 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLoFDOCvFaWmEPzWXy:-NXLoFDOCvFaWmEPzWXz:b-3hmeo1) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.h#L79)):*
> ```objc
> 
> // Make a new, possibly smaller vector from vector @p v, by selecting the rows
> // NOT indexed by @p indices.
> ```

BTW the wording here is different from the similarly-named ExcludeCols() functions above. I thought that meant this function did something different with the indices at first. Consider using the same wording for the same functionality to avoid implying a change when there is none.
___
*[`multibody/plant/slicing_and_indexing.cc` line 88 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLqCmL5na2Rn7ZlpI6:-NXLqCmL5na2Rn7ZlpI7:b7ijaho) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.cc#L88)):*
> ```cpp
>       ++r_exclude_index;
>     }
>   }
> ```

FYI fine as is, but you could make this code easier to read by using `continue` for the excluded case to get rid of much of the indenting. E.g.
```c++
if (r_exclude_index < ssize(indices) 
    && i == indices[r_exclude_index]) {
  ++r_exclude_index;
  continue;
}
int c_index = 0;
...
```
___
*[`multibody/plant/slicing_and_indexing.cc` line 159 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLreY-BVeOElecM2N0:-NXLreY-BVeOElecM2N1:b-51l415) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.cc#L159)):*
> ```cpp
>   } else {
>         "ExcludeColsCols only supports dense MatrixBlock arguments.");
> ```

typo: ColsCols -> Cols
___
*[`multibody/plant/test/slicing_and_indexing_test.cc` line 89 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLtXAv5oRR-cf1ceaw:-NXLtXAv5oRR-cf1ceax:b-onwkgi) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/test/slicing_and_indexing_test.cc#L89)):*
> ```cpp
>   // clang-format on
>   EXPECT_EQ(S, S_expected);
> }
> ```

minor: should have tests for the new Block methods also

sherm1 (commented):
Platform :lgtm_strong: pending a few minor things

> Reviewed 3 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXMF4tYAHxYvt3LTmiU:babq992)* status: 14 unresolved discussions, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 128 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM4DKT-ffF2JIodNle:-NXM4DKT-ffF2JIodNlf:b-3i1fuo) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.h#L128)):*
> ```objc
> 
>   /* Makes a "reduced" contact problem given the per-clique DOFs specified in
>     `known_free_motion_dofs` are known to equal the free-motion velocities.
> ```

minor: I found this confusing. Below there are both known_free_motion_dofs and per_clique_known_free_motion_dofs. The comment here seems to mix those up.

Also from the comment here and below I can't figure out why there are both of these parameters. Is it because not all dofs are included in cliques? Can you clarify the description here for the uninitiated?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM3_6u3GCfPQTTC0XE:-NXM3_6u3GCfPQTTC0XF:boqwvp1) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*
> ```objc
>     @param[in] per_clique_known_free_motion_dofs Specifies known DOFs to be
>     eliminated, per clique. i ∈ per_clique_known_free_motion_dofs[c] specifies
>     the i-th DOF of the c-th clique. This parameter should contain the same
> ```

minor: the rest of this sentence got lost
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 142 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM3rEt-SVbt--SA1Af:-NXM3rEt-SVbt--SA1Ag:b-o5s9xa) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.h#L142)):*
> ```objc
>          [0, ..., num_velocities()-1].
>     @pre per_clique_known_free_motion_dofs.size() == num_cliques().
>     @pre per_clique_known_free_motion_dofs[i] is a strict ordered subset of
> ```

BTW consider `c` here rather than `i` to match the comments above
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 131 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM5ozd5Lio9PQ4LzAa:-NXM5ozd5Lio9PQ4LzAb:brf35l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.cc#L131)):*
> ```cpp
>         "Jacobian does not match the number of velocities in this problem "
>         "for "
>         "the first clique.");
> ```

nit: combine lines
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 139 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM5qjKAItHWt9Fac16:-NXM5qjKAItHWt9Fac17:brf35l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.cc#L139)):*
> ```cpp
>         "Jacobian does not match the number of velocities in this problem "
>         "for "
>         "the second clique.");
> ```

nit: combine lines
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 278 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM6MNC3xhuVmh_s9O-:-NXM6MNC3xhuVmh_s9O0:berdh17) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L278)):*
> ```cpp
> |     3      |          1         ||       3        |           2            |
> └────────────┴────────────────────┘|       4        |           3            |
>                                    └────────────────┴────────────────────────┘
> ```

BTW this table didn't format well in Reviewable. Please check that it is OK in the code (also the other tables below):

___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 418 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXMF0VRDwO4FC1R-qpz:-NXMF0VRDwO4FC1R-qq-:b-awqw5w) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L418)):*
> ```cpp
>     | original |   4   |      5        |       0        |       3         |
>     | reduced  |   3   |      5        |       1        |      N/A        |
>     └──────────┴───────┴───────────────┴────────────────┴─────────────────┘
> ```

FYI Sweet! Love the tables.

joemasterjohn (commented):
Thanks Sherm! Fixing it up now.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-:-NXMHdNwBKj-rLQ8PIpP:bi1i2rn)* status: 14 unresolved discussions, missing label for release notes

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19488)* status: missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 167 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLlIl55vnq8HVOMpBK:-NXMLpb10k1HxZIJyJVM:bfu1gn) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.h#L167)):*

minor: not clear what the `x` and `c` notation on this line means
</blockquote></details>

Done. Placeholders for clique indices. Beautified it a bit more with some clarity added (hopefully).
___
*[`multibody/contact_solvers/sap/sap_constraint.h` line 177 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLm0r4D0Ht2RdNRhH0:-NXMLwAcCgDZM-eNqL65:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.h#L177)):*

This table is very helpful documentation, thanks!
</blockquote></details>

Thanks!
___
*[`multibody/contact_solvers/sap/sap_constraint.cc` line 74 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLm_BU-3LtEkpzJBay:-NXMMJeh2dgnPA5BHoyn:b35wj2g) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_constraint.cc#L74)):*

BTW consider not cloning the constraint until after you've done this test
</blockquote></details>

Done. Good point.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 128 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM4DKT-ffF2JIodNle:-NXMZstcE5XyxzVneVSP:bkj80ch) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.h#L128)):*

minor: I found this confusing. Below there are both known_free_motion_dofs and per_clique_known_free_motion_dofs. The comment here seems to mix those up.

Also from the comment here and below I can't figure out why there are both of these parameters. Is it because not all dofs are included in cliques? Can you clarify the description here for the uninitiated?
</blockquote></details>

Done. I added some docs to clarify. It has to do with the global velocity indices (stored in `known_free_motion_dofs`) vs. the cliques' local velocity indices (stored in `per_clique_known_free_motion_dofs`). They both contain the same information and you can transform between the two, but this is precomputed with the joint locking cache is evaluated. The reason I need both here is because some things in the `SapContactProblem` are stored dense with global indices (i.e. `v*`) and other things sparse with clique local indices (linear dynamics matrices and constraint jacobians).
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM3_6u3GCfPQTTC0XE:-NXMOVSu1boQmyh10NAJ:bu9gah8) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*

minor: the rest of this sentence got lost
</blockquote></details>

Done. Thanks for catching this.
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 278 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXM6MNC3xhuVmh_s9O-:-NXMPNhF3svMWEQfM0CO:blew1zn) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L278)):*

BTW this table didn't format well in Reviewable. Please check that it is OK in the code (also the other tables below):


</blockquote></details>

Weird it looks good in both Reviewable and in vscode for me. This is from github's code browser:



And even better in vim:



Maybe a windows vs. linux difference?
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 418 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXMF0VRDwO4FC1R-qpz:-NXMPC181BfhxEieFatX:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L418)):*

FYI Sweet! Love the tables.
</blockquote></details>

Thanks!
___
*[`multibody/plant/slicing_and_indexing.h` line 49 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXGoCpq9hOMO62NOFl_:-NXMR_7KCWu6AAqZXa21:b-8gij6t) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.h#L49)):*

nit, do we still want this new select method? beyond the unit test, I believe it's no longer used.
</blockquote></details>

Done. No longer used, plus I didn't have test coverage for it. Removed.
___
*[`multibody/plant/slicing_and_indexing.h` line 79 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLoFDOCvFaWmEPzWXy:-NXMS_4dCT3hWprrxI4r:b1bcodz) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.h#L79)):*

BTW the wording here is different from the similarly-named ExcludeCols() functions above. I thought that meant this function did something different with the indices at first. Consider using the same wording for the same functionality to avoid implying a change when there is none.
</blockquote></details>

Done. Thanks I at some point changed the language and missed this one.
___
*[`multibody/plant/slicing_and_indexing.cc` line 88 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLqCmL5na2Rn7ZlpI6:-NXMStF5-BJxLQjAHYl0:b9dfn12) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/slicing_and_indexing.cc#L88)):*

FYI fine as is, but you could make this code easier to read by using `continue` for the excluded case to get rid of much of the indenting. E.g.
```c++
if (r_exclude_index < ssize(indices) 
    && i == indices[r_exclude_index]) {
  ++r_exclude_index;
  continue;
}
int c_index = 0;
...
```
</blockquote></details>

Thanks! That is a much more readable way of writing it.
___
*[`multibody/plant/test/slicing_and_indexing_test.cc` line 89 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19488#-NXLtXAv5oRR-cf1ceaw:-NXMXysd2XCEyLNXcvpG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fb37885e0a297a21e54c707ade8244125cdf1620/multibody/plant/test/slicing_and_indexing_test.cc#L89)):*

minor: should have tests for the new Block methods also
</blockquote></details>

Done.

Issue #18659, opened 2023-01-26
Rewrite/refesh of push_release tooling


jwnimmer-tri:
See https://github.com/RobotLocomotion/drake/issues/18619#issuecomment-1396216666 for some background and details.

There's enough going on there that we think rewriting in Python will probably be to our advantage.

The basic requirements are:

(1) Copy all attachments from the GitHub release page into S3 for backup purposes (in case GitHub loses them somehow).

(2) Inject the `*.deb` files downloaded from the GitHub release page into the aptly site.

(3) Push the version-numbered docker tags.

At the time of the rewrite, we should support support macOS for this tool; it should target Ubuntu Jammy (only).

---

Also a note in case it helps for planning -- for (3) at the moment we just re-tag the nightly docker images as stable.  In the future, the `*.tar.gz` files associated with a stable release will embed a stable version number, not a nightly version stamp, at which point we'll want to create new docker images using those tarballs, rather than re-tagging nightly tag.  If it makes sense, the new script could work that way from the get-go (mimicing [drake-ci](step-build-docker-image.cmake), where it takes the downloaded `*.tar.gz` file from GitHub, unpacks it into a local Docker, runs `install_prereqs`, and drops a tag.

jwnimmer-tri:
Addendum for clarity: currently, the script is GPG-signing (`*.asc`) the https://drake.mit.edu/from_binary.html tarballs.  It should stop doing that.  We don't advertise that signature in our docs, and it introduces needless complexity.  Only the requirements listed in the top post should remain intact in the rewrite.

BTW related: https://blog.pypi.org/posts/2023-05-23-removing-pgp/

BetsyMcPhail:
When we eventually get to implementing a job in Jenkins, it is possible to "Enable project-based security" and create a security matrix that does not inherit from the global security settings. This way, we can specify users or, more likely, a Github group that is allowed access to the job.

jwnimmer-tri:
For some updates, see https://github.com/RobotLocomotion/drake/pull/19496#pullrequestreview-1444773578.

> This means that the *.deb.asc file needs to appear on the GitHub releases page one way or another.

Note that (4) is gone now.  We will no longer be signing individual release files, rather only the aptly Releases index.

jwnimmer-tri:
With the new scripts, we've improved and automated this a lot more.  There's more we could do, but we might as well defer that until we switch more automation over to Jenkins instead of manual scripts.

[...]

PR #19496, opened 2023-05-25
[releng] Don't sign individual files

jwnimmer-tri:
Towards #18659.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19496)

### DIFF
--- a/tools/release_engineering/dev/push_release
+++ b/tools/release_engineering/dev/push_release
@@ -86,7 +86,7 @@ if [[ "${push_docker}" -ne 0 ]] && ! command -v docker &>/dev/null; then
   exit 6
 fi
 
-if [[ "${push_apt}" -ne 0 || ("${push_tar}" -ne 0 && "${sign}" -ne 0) ]] && ! command -v gpg &>/dev/null; then
+if [[ "${push_apt}" -ne 0 ]] && ! command -v gpg &>/dev/null; then
   echo 'ERROR: gpg(1) was NOT found. Fix with apt-get install gnupg or brew install gnupg.' >&2
   exit 6
 fi
@@ -131,7 +131,7 @@ if ! [[ "${gpg_key}" =~ ^[a-fA-F0-9]{40}$ ]]; then
   exit 1
 fi
 
-if [[ "${push_apt}" -ne 0 || ("${push_tar}" -ne 0 && "${sign}" -ne 0) ]] && ! gpg --list-secret-keys "${gpg_key}" &>/dev/null; then
+if [[ "${push_apt}" -ne 0 ]] && ! gpg --list-secret-keys "${gpg_key}" &>/dev/null; then
   echo "ERROR: GPG secret key ${gpg_key} does NOT exist." >&2
   exit 10
 fi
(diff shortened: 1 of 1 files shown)


jwnimmer-tri (commented):
+@BetsyMcPhail to kick off review, please.  \CC @mwoehlke-kitware

There are two commits.

The first commit is definitely wanted.  It removes tgz signing.

The second commit is probably wanted, as best I understand.  I don't think uploading to aptly requires each deb to be signed individually.  Rather, only the releases files is signed.  So, this removes the sign/no-sign option entirely, by turning off all signing.  (You could test the equivalent thing by running the current push_release with the --no-sign flag.)

Note that the GPG key is still required.  The information we were using in the meeting to base our decisions on was wrong, I think.  The crux of the matter was the key passed to aptly, not gpg.

I propose the new goal is these three, separate, scripts:

(1) `release_aptly` does the aptly reindexing and release signing.  It should fetch the current S3 aptly data, download the new debs from the GitHub release page, merge them into the site ("switch").  This is very much like the `push_apt` stanza of the current script, but not GPG-signing each individual deb file.

(2) `release_docker` does the image push + pull tagging.  This is very much like the `push_docker` stanza of the current script.

(3) `release_mirror` syncs the GitHub attachments into S3 storage.

I think trying to bite off all three at once into a single unified program rewritten in big-bang style is slowing us down.  We should be able to create each new script one by one and start using it.  Once all three items are rewritten, we can delete the old script.

From the sound of it, the `release_aptly` is the most painful step currently so that should be the focus of the next tool to be written.

Given the above, if this PR is still useful then feel free to review it to be merged.  If the only thing we needed was clearer requirements, and we're going to write new programs entirely from scratch, maybe just reading this PR is enough and we can close is unmerged.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19496#-:-NWKBxjf-z_kG08yjbxe:bmfbwri)* status: LGTM missing from assignee BetsyMcPhail, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @BetsyMcPhail and @jwnimmer-tri)

mwoehlke-kitware (commented):
:lgtm:

I feel sort of weird about removing signing. Signing, done properly, is good, as it ensures (so long as no attacker gains access to the private signing key) that the files we distribute are as built and haven't been replaced with maliciously altered copies.

That said, long term that logic should be happening when the packages are built, and the artifacts should land in GitHub along with the packages themselves. (And the signature files should be properly documented.) So, while this may be worth revisiting, I think it's acceptable for the sake of being able to make progress to remove those actions from `push_release`.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19496#-:-NWiJj8cA2kr6BbQXzYp:b-to7rrf)* status: LGTM missing from assignee BetsyMcPhail, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @BetsyMcPhail and @jwnimmer-tri)

BetsyMcPhail (commented):
:lgtm: I'm happy for both of these changes to go in. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19496#-:-NWiNV__22aaed6SEnyt:b-m5pq8r)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

xuchenhan-tri (commented):
I don't have knowledge on the pros and cons of requiring a signature, but with two feature reviewers I feel I'm in good hands.
:lgtm: platform.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19496#-:-NWmiT3u9UBDo4UrKbut:bcgd7du)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

jwnimmer-tri:
> Signing, done properly, is good, as it ensures (so long as no attacker gains access to the private signing key) that the files we distribute are as built and haven't been replaced with maliciously altered copies.

This https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/ says it better than I could.

PR #19507, opened 2023-05-31
[rendering] TextureLibrary upgraded to prepare for cloning

SeanCurtis-TRI:
1. `TextureLibrary` no longer stores a pointer to `OpenGlContext`. It simply declares that the `OpenGlContext` must be bound. This will aid in cloning contexts across render engine families.
2. It handles look up of texture id in a thread-safe way. This was missing before and has now been corrected.
3. Tests have been added.
4. The texture library now handles symlink files so that a single file, aliased in different ways, doesn't lead to multiple textures on the GPU.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507)

### DIFF
--- a/geometry/render_gl/BUILD.bazel
+++ b/geometry/render_gl/BUILD.bazel
@@ -306,6 +306,31 @@ drake_cc_googletest_ubuntu_only(
     ],
 )
 
+drake_cc_googletest_ubuntu_only(
+    name = "multithread_safety_test",
+    data = [
+        "//geometry/render:test_models",
+    ],
+    tags = vtk_test_tags() + ["cpu:4"],
+    deps = [
+        ":internal_texture_library",
+        "//common:find_resource",
+    ],
+)
+
+drake_cc_googletest_ubuntu_only(
+    name = "internal_texture_library_test",
+    data = [
+        "//geometry/render:test_models",
+    ],
+    tags = vtk_test_tags(),
+    deps = [
+        ":internal_texture_library",
+        "//common:find_resource",
+        "//common:temp_directory",
+    ],
+)
+
 add_lint_tests(
     cpplint_extra_srcs = [
         "no_factory.cc",
--- a/geometry/render_gl/internal_render_engine_gl.cc
+++ b/geometry/render_gl/internal_render_engine_gl.cc
@@ -22,6 +22,7 @@ using geometry::internal::LoadRenderMeshFromObj;
 using math::RigidTransformd;
 using std::make_shared;
 using std::make_unique;
+using std::shared_ptr;
 using std::string;
 using std::unique_ptr;
 using std::unordered_map;
@@ -164,10 +165,16 @@ class DefaultTextureColorShader final : public ShaderProgram {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(DefaultTextureColorShader)
 
   /* Constructs the texture shader with the given library. The library will be
-   used to access OpenGl textures.  */
-  explicit DefaultTextureColorShader(TextureLibrary* library)
-      : ShaderProgram(), library_(library) {
-    DRAKE_DEMAND(library != nullptr);
+   used to access OpenGl textures.
+
+   When the RenderEngineGl is cloned, instances of this shader program are
+   likewise cloned, each with a shared ptr to the *same* texture library. This
+   is alright, because the owning RenderEngineGl instances share that library
+   as well -- so the shader program instances are consistent with the render
+   engine instances. */
+  explicit DefaultTextureColorShader(shared_ptr<TextureLibrary> library)
+      : ShaderProgram(), library_(std::move(library)) {
+    DRAKE_DEMAND(library_ != nullptr);
     LoadFromSources(kVertexShader, kFragmentShader);
     diffuse_map_loc_ = GetUniformLocation("diffuse_map");
     diffuse_scale_loc_ = GetUniformLocation("diffuse_map_scale");
@@ -241,7 +248,7 @@ class DefaultTextureColorShader final : public ShaderProgram {
     glUniformMatrix3fv(normal_mat_loc_, 1, GL_FALSE, normal_mat.data());
   }
 
-  TextureLibrary* library_{};
+  std::shared_ptr<TextureLibrary> library_{};
 
   // The location of the "diffuse_map" uniform in the shader.
   GLint diffuse_map_loc_{};
@@ -462,7 +469,7 @@ void main() {
 RenderEngineGl::RenderEngineGl(RenderEngineGlParams params)
     : RenderEngine(params.default_label),
       opengl_context_(make_shared<OpenGlContext>()),
-      texture_library_(make_shared<TextureLibrary>(opengl_context_.get())),
+      texture_library_(make_shared<TextureLibrary>()),
       parameters_(std::move(params)) {
   // Configuration of basic OpenGl state.
   opengl_context_->MakeCurrent();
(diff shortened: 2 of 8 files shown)


SeanCurtis-TRI (commented):
This is one of three steps to facilitate parallel-rendering using `RenderEngineGl`. To see how it fits into the whole, look at #19426.

+(release notes: none) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NWmUqYq8kxjpegncfVK:bhoyqon)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NWmYUuP5vVW24cmn6g1:b7dve2f)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI and @sherm1)

sherm1 (commented):
Feature :lgtm:

> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NWmzvap00K4h4C4-beo:b-n0m1lx)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

sherm1 (commented):
(Pending CI fix)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NWmzyhj8rtFRB-LELzr:bw5ntk2)* status: 6 unresolved discussions, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NXB-sAq5SXzu1SXhwXH:bcav8r0)* status: LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`geometry/render_gl/internal_texture_library.h` line 16 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NWmjf5_DxTnifarPj0_:-NXAtJXZC2agaFzjZ8rx:b9bw0f6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f6ae19a72d3b7990c9bc8436f99cd3da5a00188/geometry/render_gl/internal_texture_library.h#L16)):*

BTW is this really a TODO? Could also just be phrased as a warning and documented as a caution to users not to expect texture updates after loading.
To my naïve eyes, this doesn't look like much of a real-life problem! 
</blockquote></details>

I agree. This was born out of a conversation with Jeremy. He'd broached two topics (as fodder for a TODO):

1. Handling symlinks.
2. On-disk changes.

I thought his symlink issue was big enough (and simple enough to resolve) to just implement it. 

However, I think you're right. What started out as a "todo" (handle on-disk changes) evolved to analysis whose conclusion was a big, "meh". So, I've excised the "todo" and replaced it with a note so that future developers can see that it was considered and dismissed. But if they detect a change that suggests the assumptions are no longer ture, they can feel free to change things up.
___
*[`geometry/render_gl/internal_texture_library.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NWmkWIB09Mhb1SJA8b8:-NXAujKh76lPSedcGtd2:b-6hfqu) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f6ae19a72d3b7990c9bc8436f99cd3da5a00188/geometry/render_gl/internal_texture_library.h#L71)):*

nit: nuke this blank line?
</blockquote></details>

I tend to like vertical white space more than others. But I'm not married to it.
___
*[`geometry/render_gl/internal_texture_library.cc` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NWmlGCNAtEoHqr884Jx:-NXAup4AEFfruL72ddKk:brfgqqd) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f6ae19a72d3b7990c9bc8436f99cd3da5a00188/geometry/render_gl/internal_texture_library.cc#L45)):*

BTW from the comment this assert() doesn't mean there is a bug in your code but some kind of obscure user screw up. Would it be worth a throw? Or at least a DRAKE_DEMAND?
</blockquote></details>

Two things:

 ---------------------------------------

I've pulled the test from *this* PR because it's truly handled better in the failing test when `OpenGlContexts` are properly cloneable. That's the follow up PR.

-----------------------------------------

On the subject of `ASSERT` vs `DEMAND`:

Throughout rendering, I'm advocating for `ASSERT` over `DEMAND`.

1. These call sites are representative of drake bugs and not user errors. (So, `DRAKE_SOMETHING` is definitely the right spelling over `throw`.)
2. Our rendering is not fast enough. We are jumping through big hoops trying to improve performance. Therefore, I balk at *anything* that isn't strictly necessary to produce the images in release code.
3. `DRAKE_ASSERT` is clearly sufficient for the purposes because we do debug builds in CI; the initial revision of this PR failed in CI. Tada.
___
*[`geometry/render_gl/internal_texture_library.cc` line 91 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NWmlyCM2op65Cy3cFUc:-NXAvg116YqMeyqwOJYC:bnhelg6) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f6ae19a72d3b7990c9bc8436f99cd3da5a00188/geometry/render_gl/internal_texture_library.cc#L91)):*

nit: DEMAND seems more appropriate here since most users will not be running in Debug and there is plenty else going on so the cost should be affordable
</blockquote></details>

See previous note on why I'm advocating `DRAKE_ASSERT` over `DRAKE_DEMAND` in rendering code.
___
*[`geometry/render_gl/internal_texture_library.cc` line 136 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NWmmciED9f7XTD5JPmw:-NXAvxPV0a6autpRwrn9:b-w8upu9) ([raw file](https://github.com/RobotLocomotion/drake/blob/6f6ae19a72d3b7990c9bc8436f99cd3da5a00188/geometry/render_gl/internal_texture_library.cc#L136)):*

BTW I would think so -- I presume the user made a typo or forgot to install a file. Would be very helpful to say so. Otherwise they'll be looking at an image that seems wrong with little hint as to why.
</blockquote></details>

It's easy to make that argument, of course. However, as this code has existed for years without every actually doing so, there's clearly no *urgent* need for it.

I attribute this to the fact that we've done a good job publicizing the fact that Drake is stupid and only knows about pngs. Once we grow the set of supported image types, that message becomes less clear and the potential for error increases.

I've moved this TODO to the top of this function to more generally log errors for any of the reasons that would make the image name "unsupported".

ggould-tri (commented):
:lgtm:

> Reviewed 5 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-:-NXBHltd5Ar_m4U6dErz:bnfp4nl)* status: 1 unresolved discussion

___
*[`geometry/render_gl/internal_texture_library.cc` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19507#-NXBEgqGBXsRvJ5R_uEp:-NXBEgqGBXsRvJ5R_uEq:bf49506) ([raw file](https://github.com/RobotLocomotion/drake/blob/95e3f91971e1d96bb515b4246d203e2625ce294f/geometry/render_gl/internal_texture_library.cc#L41)):*
> ```cpp
>   }
> 
>   if (!IsSupportedImage(file_name)) {
> ```

BTW:  This stanza could reasonably move outside of the lock:  If the file name is an unsupported type, the `find` wasn't going to succeed anyway.

ggould-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19507)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),sherm1(platform)

PR #19508, opened 2023-05-31
OpenGlInstance now references OpenGlGeometry in preparation for cloning

SeanCurtis-TRI:
Previously, each `OpenGlInstance` contained a *copy* of the `OpenGlGeometry` that it instantiated. While the cost of a few ints isn't overly burdensome, this will be an obstacle in future efforts to clone the OpenGl context.

`RenderEngineGl` now maintains a vector of `OpenGlGeometry` and each instance merely gets an index into that vector. These indices are amenable to copying without requiring changing the references.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508)

### DIFF
--- a/geometry/render_gl/internal_opengl_geometry.h
+++ b/geometry/render_gl/internal_opengl_geometry.h
@@ -47,15 +47,22 @@ struct OpenGlGeometry {
    @param index_buffer_in       The handle to the OpenGl index buffer defining a
                                 set of triangles.
    @param index_buffer_size_in  The number of indices in the index buffer.
+   @param has_tex_coord_in      If true, the vertex buffer contains *meaningful*
+                                texture coordinates.
+   @param v_count_in            The number of vertices in this mesh (and, by
+                                implication, the number of normals and texture
+                                coordinates).
    @pre `index_buffer_size_in >= 0`.  */
-  OpenGlGeometry(GLuint vertex_array_in, GLuint vertex_buffer_in,
-                 GLuint index_buffer_in, int index_buffer_size_in,
-                 bool has_tex_coord_in)
+  OpenGlGeometry(GLuint vertex_array_in,
+                 GLuint vertex_buffer_in, GLuint index_buffer_in,
+                 int index_buffer_size_in, bool has_tex_coord_in,
+                 int v_count_in)
       : vertex_array{vertex_array_in},
         vertex_buffer{vertex_buffer_in},
         index_buffer{index_buffer_in},
         index_buffer_size{index_buffer_size_in},
-        has_tex_coord{has_tex_coord_in} {
+        has_tex_coord{has_tex_coord_in},
+        v_count(v_count_in) {
     if (index_buffer_size < 0) {
       throw std::logic_error("Index buffer size must be non-negative");
     }
@@ -77,33 +84,42 @@ struct OpenGlGeometry {
     if (!is_defined()) throw std::logic_error(message);
   }
 
+  // TODO(SeanCurtis-TRI): This can't really be a struct; there are invariants
+  // that need to be maintained: vertex_buffer is sized according to v_count,
+  // vertex_array depends on vertex_buffer, has_tex_coord needs to reflect
+  // the uv data in vertex_buffer, and index_buffer_size needs to be the actual
+  // size of index_buffer (in triangles).
   GLuint vertex_array{kInvalid};
   GLuint vertex_buffer{kInvalid};
   GLuint index_buffer{kInvalid};
   int index_buffer_size{0};
+
   /* True indicates that this has texture coordinates to support texture
    maps. See MeshData::has_tex_coord for detail.  */
   bool has_tex_coord{};
+
+  // The number of vertices encoded in `vertex_buffer`.
+  int v_count{};
 
   /* The value of an object (array, buffer) that should be considered invalid.
    */
   static constexpr GLuint kInvalid = std::numeric_limits<GLuint>::max();
 };
 
-/* An instance of a geometry in the renderer - the underlying OpenGl geometry
- definition in Frame G, its pose in the world frame W, and scale factors. The
- scale factors are not required to be uniform. They _can_ be negative, but that
- is not recommended; in addition to mirroring the geometry it will also turn
- the geometry "inside out".
+/* An instance of a geometry in the renderer - a reference to the underlying
+ OpenGl geometry definition in Frame G, its pose in the world frame W, and scale
+ factors. The scale factors are not required to be uniform. They _can_ be
+ negative, but that is not recommended; in addition to mirroring the geometry it
+ will also turn the geometry "inside out".
 
  When rendering, the visual geometry will be scaled around G's origin and
  subsequently posed relative to W.  */
 struct OpenGlInstance {
   /* Constructs an instance from a geometry definition, a pose, a scale factor
    and the instance's shader data for depth and label shaders.
-   @pre `g_in.is_defined()` reports `true`.
+   @pre g_in references a defined OpenGlGeometry.
    @pre The shader program data has valid shader ids.  */
-  OpenGlInstance(const OpenGlGeometry& g_in,
+  OpenGlInstance(int g_in,
                  const math::RigidTransformd& pose_in,
                  const Vector3<double>& scale_in, ShaderProgramData color_data,
                  ShaderProgramData depth_data, ShaderProgramData label_data)
@@ -114,12 +130,13 @@ struct OpenGlInstance {
     shader_data[RenderType::kColor] = std::move(color_data);
     shader_data[RenderType::kDepth] = std::move(depth_data);
     shader_data[RenderType::kLabel] = std::move(label_data);
-    DRAKE_DEMAND(geometry.is_defined());
+    DRAKE_DEMAND(geometry >= 0);
   }
 
+  // This is the index to the OpenGlGeometry stored by RenderEngineGl.
+  int geometry{};
   // TODO(SeanCurtis-TRI) Change these quantities to be float-valued so they
   //  can go directly into the shader without casting.
-  OpenGlGeometry geometry;
   math::RigidTransformd X_WG;
   Vector3<double> scale;
   std::array<ShaderProgramData, RenderType::kTypeCount> shader_data;
--- a/geometry/render_gl/internal_render_engine_gl.cc
+++ b/geometry/render_gl/internal_render_engine_gl.cc
@@ -7,6 +7,7 @@
 
 #include <fmt/format.h>
 
+#include "drake/common/ssize.h"
 #include "drake/common/text_logging.h"
 #include "drake/common/unused.h"
 
@@ -40,6 +41,8 @@ using systems::sensors::ImageTraits;
 using systems::sensors::PixelType;
 
 namespace {
+
+namespace fs = std::filesystem;
 
 constexpr char kInternalGroup[] = "render_engine_gl_internal";
 constexpr char kHasTexCoordProperty[] = "has_tex_coord";
@@ -508,9 +511,9 @@ void RenderEngineGl::UpdateViewpoint(const RigidTransformd& X_WR) {
 }
 
 void RenderEngineGl::ImplementGeometry(const Box& box, void* user_data) {
-  OpenGlGeometry geometry = GetBox();
-  ImplementGeometry(geometry, user_data,
-                    Vector3d(box.width(), box.depth(), box.height()));
+  const int geometry = GetBox();
+  AddGeometryInstance(geometry, user_data,
+                      Vector3d(box.width(), box.depth(), box.height()));
 }
 
 void RenderEngineGl::ImplementGeometry(const Capsule& capsule,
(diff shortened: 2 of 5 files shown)


SeanCurtis-TRI (commented):
This is one of three steps to facilitate parallel-rendering using `RenderEngineGl`. To see how it fits into the whole, look at #19426.

+(release notes: none) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-:-NWmUz_PEq5mCbSLtfXe:bhoyqon)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
+@xuchenhan-tri for feature review, please. I know this isn't *particularly* your wheelhouse, but you just have to convince yourself that the new indices are equivalent to the old copies. Thanks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-:-NWmYgZbDA_6Di0PUxOv:bcnc3yf)* status: LGTM missing from assignee xuchenhan-tri(platform), needs at least two assigned reviewers (waiting on @SeanCurtis-TRI and @xuchenhan-tri)

xuchenhan-tri (commented):
Feature :lgtm: with a few clarifying question due to my lack of familiarity in this area.
+@ggould-tri for platform review on Thursday per rotation please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-:-NWmsO1h103qo_cYXuCD:b-7abnoz)* status: 8 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmp1pZ80_ZdFUMm2Lr):*
I'm not quite seeing why this poses an obstacle. Could you elaborate?



_Code quote:_
```
  burdensome, this will be an obstacle in future efforts to clone the
  OpenGl context.
```
___
*[`geometry/render_gl/internal_opengl_geometry.h` line 56 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmjcJZDX7M3o3MTmGB:-NWmjcJZDX7M3o3MTmGC:b5ws7ay) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_opengl_geometry.h#L56)):*
> ```objc
>   OpenGlGeometry(GLuint vertex_array_in,
>                  GLuint vertex_buffer_in, GLuint index_buffer_in,
>                  int index_buffer_size_in, bool has_tex_coord_in,
> ```

BTW, this parameter is not documented.  Consider adding a note for completion if the omission is accidental and not intentional.



_Code quote:_
```objc
bool has_tex_coord_in,
```
___
*[`geometry/render_gl/internal_opengl_geometry.h` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmkT9M0gyhWQW0c1KL:-NWmkT9M0gyhWQW0c1KM:b-4p1uk5) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_opengl_geometry.h#L118)):*
> ```objc
>   /* Constructs an instance from a geometry definition, a pose, a scale factor
>    and the instance's shader data for depth and label shaders.
>    @pre `g_in.is_defined()` reports `true`.
> ```

nit this needs an update.



_Code quote:_
```objc
@pre `g_in.is_defined()` reports `true`.
```
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmmEzbA8697h43haaT:-NWmmEzbA8697h43haaU:blmiw7e) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.h#L75)):*
> ```objc
>   // texture, if found. When we eliminate that behavior, we can eliminate this
>   // method.
>   void ImplementMesh(int geometry_index, void* user_data,
> ```

BTW, I'm assuming all `geometry_index` parameters, here and introduced below, are indexing into data member `geometries_` of this object. Consider adding a note if that's not right.



_Code quote:_
```objc
geometry_index
```
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 251 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmnFRQ9ReKAbG2VhR7:-NWmnFRQ9ReKAbG2VhR8:b-xny95e) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.h#L251)):*
> ```objc
>   // cannot be realized from another capsule via a linear transformation (i.e.,
>   // translation, rotation, and non-uniform scale).
>   int sphere_{-1};
> ```

minor, I think it would be beneficial to add a note that these ints are indices into `geometries_` (IIUC).
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 1068 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmpv3t5BOnBzDJB5g3:-NWmpv3t5BOnBzDJB5g4:b96tp8d) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.cc#L1068)):*
> ```cpp
>   // canonical mesh. We keep all canonical meshes alive for the lifetime of the
>   // OpenGL context for convenient reuse.
>   const int index = static_cast<int>(geometries_.size());
> ```

BTW, ssize(geometries_)?



_Code quote:_
```cpp
static_cast<int>(geometries_.size())
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 1116 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmrhwmA7N1uQInRPB-:-NWmrhwn7d-ot-WL8YmD:biov8hl) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.cc#L1116)):*
> ```cpp
>   DRAKE_DEMAND(vbo_offset == float_count * sizeof(GLfloat));
> 
>   // Bind IBO with the VAO.
> ```

minor, it'd be nice to keep the definition of `VAO` from the previous version.



_Code quote:_
```cpp
 VAO
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 1158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWms7Z03HHjpVTs3uhZ:-NWms7Z03HHjpVTs3uh_:b-p9jun2) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.cc#L1158)):*
> ```cpp
> 
>     // We prioritize the shader by id; higher ids will always win.
>     if (data.has_value() && program.shader_id() < data->shader_id()) continue;
> ```

BTW, I'm assuming this diff has no behavior change (i.e. `program.shader_id() == candidate_data->shader_id()`)



_Code quote:_
```cpp
    // We prioritize the shader by id; higher ids will always win.
    if (data.has_value() && program.shader_id() < data->shader_id()) continue;
```

ggould-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-:-NWrRU7j1GtjUqj1NzFH:bnfp4nl)* status: 10 unresolved discussions

___
*[`geometry/render_gl/internal_opengl_geometry.h` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWrTFi73nA65pxB-VoA:-NWrTFi73nA65pxB-VoB:b2jorv0) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_opengl_geometry.h#L115)):*
> ```objc
>  When rendering, the visual geometry will be scaled around G's origin and
>  subsequently posed relative to W.  */
> struct OpenGlInstance {
> ```

minor; author's discretion:
This type name seems a bit misleading (it's a posed entity + geometry reference, not an instance of any of the other nine billion OpenGL-related entities).
But I know that's a heavy item for a relatively minor PR, so feel free to dismiss or TODO this.



_Code quote:_
```objc
OpenGlInstance
```
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 260 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWrQA_e31MeaMARH43-:-NWrQA_e31MeaMARH430:bd82l94) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.h#L260)):*
> ```objc
>   // and allowed deviation within a small tolerance, then we could reuse them.
> 
>   // Mapping from obj filename to the mesh loaded into an OpenGlGeometry.
> ```

minor:  "to an index into..."  would be clearer now that the type has changed.



_Suggestion:_
```objc
 to the `geometries_` index of the mesh
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),xuchenhan-tri(platform)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmp1pZ80_ZdFUMm2Lr):*

I'm not quite seeing why this poses an obstacle. Could you elaborate?
</blockquote></details>

It's incredibly subtle:

1. OpenGl identifiers are getting allocated in geometry registration order. So, the id value associated with a particular geometry depends on sdf files and the like.
2. During cloning, we don't know the ordering. We know which geometries have been created. So, we recreate them but (probably) in a different order. We probably end up with the same *set* of identifiers, but their meaning can permute among the set.
3. By simply copying the `OpenGlGeometry` we persist the *old* id with its old meaning. To have it be correct w.r.t. the cloned context, we'd have to visit each instance and update its identifiers.

Rather than update every copy of a particular `OpenGlGeometry`, we simply update the one and allow it to naturally trickle down.

Make sense?
___
*[`geometry/render_gl/internal_opengl_geometry.h` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWrTFi73nA65pxB-VoA:-NXB17kmDWJ4hMho9MqA:blcqxpy) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_opengl_geometry.h#L115)):*

minor; author's discretion:
This type name seems a bit misleading (it's a posed entity + geometry reference, not an instance of any of the other nine billion OpenGL-related entities).
But I know that's a heavy item for a relatively minor PR, so feel free to dismiss or TODO this.
</blockquote></details>

It's a term of art.

In graphics, instances often refer to the combination of "geometry", "material", and "pose". My narcissistic assumption is that for people who do a lot of computer graphics, the name will be meaningful.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 1158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWms7Z03HHjpVTs3uhZ:-NXB2d_YED6-3nL61JC0:ba98l65) ([raw file](https://github.com/RobotLocomotion/drake/blob/fc26d5a0df6bde03b7db69976fc23e7123f74ec2/geometry/render_gl/internal_render_engine_gl.cc#L1158)):*

BTW, I'm assuming this diff has no behavior change (i.e. `program.shader_id() == candidate_data->shader_id()`)
</blockquote></details>

Correct. Originally I put a todo that says, "Why am I evaluating candidate data if I'm going to throw it out on this basis *anwyays*?

ANd then I thought, "Really? A TODO? For such a small thing? Might as well sneak it in now."

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19508)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),xuchenhan-tri(platform)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19508#-NWmp1pZ80_ZdFUMm2Lr):*

It's incredibly subtle:

1. OpenGl identifiers are getting allocated in geometry registration order. So, the id value associated with a particular geometry depends on sdf files and the like.
2. During cloning, we don't know the ordering. We know which geometries have been created. So, we recreate them but (probably) in a different order. We probably end up with the same *set* of identifiers, but their meaning can permute among the set.
3. By simply copying the `OpenGlGeometry` we persist the *old* id with its old meaning. To have it be correct w.r.t. the cloned context, we'd have to visit each instance and update its identifiers.

Rather than update every copy of a particular `OpenGlGeometry`, we simply update the one and allow it to naturally trickle down.

Make sense?
</blockquote></details>

Yes, well explained. Thank you.

Issue #16923, opened 2022-04-11
proposal: `ConcatenateStdVector : public LeafSystem`


RussTedrake:
Pain point:  If I have multiple systems that produces one or more `ExternallyAppliedSpatialForce`, then (I believe) there is no ergonomic way to combine them into a single `std::vector<ExternallyAppliedSpatialForce>` to be passed to the appropriate input port of `MultibodyPlant`.  We have other `std::vector<Type>` input ports in the framework that could similarly benefit from some form of accumulator.

Proposal:  Rather than a narrow implementation for this one type, I think it would be more useful to write a new primitive that looks something like

```
template <typename T, template <typename> typename ValueType>
class ConcatenateStdVector : public LeafSystem<T> {
```
which will offer some number of input ports which accept `std::vector<ValueType<T>>` and some that accept `ValueType<T>`, and a single output port which concatenates them into a single `std::vector<ValueType<T>>`. 

Although I have not tried yet, I believe that writing a method that produces the python bindings for each `ValueType` in much the way we bind e.g. `solvers::Binding`, should be straightforward.

WDYT?

jwnimmer-tri:
This might be duplicate of #13139?

In terms of a mux system, my thoughts were https://github.com/RobotLocomotion/drake/issues/12401#issuecomment-617168844.  I still think we should do it type-erased, not with `typename ValueType` statically.

However, stepping back, I don't like the mux system idea much in the first place.

Instead of adding more awkward boilerplate systems that people need to manually wire up with more connection boilerplate, I'd much rather that we add the ability in DiagramBuilder to connect multiple output ports to a single input port, natively.

The developer (in this case, MbP) would need to opt-in input port declaration as allowing multiple connections. Once they do so, then the user can just connect all of the (e.g.) propeller force outputs to the sole MbP input, and call it a day.

RussTedrake:
Excellent.  Thank you for finding the old discussion!

I agree that having native support in the system framework would be much better.  But that's a bigger mountain for me to climb this week.  I think that offering the mux does no harm, and it would unblock these workflows right now.  I can write in this idea of the system framework native support into that file.  Sound ok?

jwnimmer-tri:
I think a specific ExternalAppliedSpatialForceMultiplexer (in the plant folder) makes a lot of sense as an expedient stopgap.  I think a generic (templated) mux has the problems noted in #13139 and should be avoided.

RussTedrake:
[...]

PR #19520, opened 2023-06-03
Implements multibody::RigidTransformSelector

RussTedrake:
The MultibodyPlant body_poses output port provides a vector of all body poses, but often a downstream system just wants to take a single pose on the input port. This system provides that capability.

While RigidTransform lives in drake::math, I felt that this system should live in drake::multibody since it is likely to have SpatialVector and SpatialAcceleration equivalents alongside it soon.

+@jwnimmer-tri for both reviews(?), please.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19520)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -26,6 +26,7 @@
 #include "drake/multibody/plant/multibody_plant_config_functions.h"
 #include "drake/multibody/plant/point_pair_contact_info.h"
 #include "drake/multibody/plant/propeller.h"
+#include "drake/multibody/plant/rigid_transform_selector.h"
 #include "drake/multibody/plant/wing.h"
 #include "drake/multibody/tree/spatial_inertia.h"
 
@@ -1310,6 +1311,17 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def("get_spatial_forces_output_port",
             &Class::get_spatial_forces_output_port, py_rvp::reference_internal,
             cls_doc.get_spatial_forces_output_port.doc);
+  }
+
+  // RigidTransformSelector
+  {
+    using Class = RigidTransformSelector<T>;
+    constexpr auto& cls_doc = doc.RigidTransformSelector;
+    auto cls = DefineTemplateClassWithDefault<Class, systems::LeafSystem<T>>(
+        m, "RigidTransformSelector", param, cls_doc.doc);
+    cls  // BR
+        .def(py::init<int>(), py::arg("index"), cls_doc.ctor.doc)
+        .def("index", &Class::index, cls_doc.index.doc);
   }
 
   // Wing
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -83,6 +83,7 @@ from pydrake.multibody.plant import (
     PointPairContactInfo_,
     PropellerInfo,
     Propeller_,
+    RigidTransformSelector_,
     Wing,
 )
 from pydrake.multibody.parsing import Parser
@@ -2662,6 +2663,11 @@ class TestPlant(unittest.TestCase):
 
         prop2 = Propeller_[float]([info, info])
         self.assertEqual(prop2.num_propellers(), 2)
+
+    @numpy_compare.check_all_types
+    def test_rigid_transform_selector(self, T):
+        system = RigidTransformSelector_[T](index=2)
+        self.assertIsInstance(system, LeafSystem_[T])
 
     def test_wing(self):
         builder = DiagramBuilder()
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -33,6 +33,7 @@ drake_cc_package_library(
         ":multibody_plant_core",
         ":point_pair_contact_info",
         ":propeller",
+        ":rigid_transform_selector",
         ":slicing_and_indexing",
         ":tamsi_solver",
         ":wing",
@@ -355,6 +356,17 @@ drake_cc_library(
     hdrs = ["propeller.h"],
     deps = [
         ":multibody_plant_core",
+    ],
+)
+
+drake_cc_library(
+    name = "rigid_transform_selector",
+    srcs = ["rigid_transform_selector.cc"],
+    hdrs = ["rigid_transform_selector.h"],
+    deps = [
+        "//common:default_scalars",
+        "//math:geometric_transform",
+        "//systems/framework:leaf_system",
     ],
 )
 
@@ -1052,6 +1064,15 @@ drake_cc_googletest(
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
         "//systems/framework",
+    ],
+)
+
+drake_cc_googletest(
+    name = "rigid_transform_selector_test",
+    deps = [
+        ":rigid_transform_selector",
+        "//common/test_utilities:expect_throws_message",
+        "//systems/framework/test_utilities",
     ],
 )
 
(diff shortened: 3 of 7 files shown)


jwnimmer-tri (commented):
> While RigidTransform lives in drake::math, I felt that this system should live in drake::multibody since it is likely to have SpatialVector and SpatialAcceleration equivalents alongside it soon.

I agree, also because our `//math` package should not depend on `//systems/framework` (even though it already erroneously does as of a couple weeks ago).

> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19520#-:-NXAXdLQAIHFDs_AUrUV:b-1lr5a0)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19520#-NXAWt2Z17xXDWbEgGyc:-NXAWt2Z17xXDWbEgGyd:bpqdv4q) (no related file):*
It seems to me that it would be strictly easier for the user if MbP provided i'th-body ports directly.

```patch
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -170,6 +170,9 @@ output_ports:
 - body_poses
 - body_spatial_velocities
 - body_spatial_accelerations
+- body_pose_{i}
+- body_spatial_velocity_{i}
+- body_spatial_acceleration_{i}
 - generalized_acceleration
 - reaction_forces
 - contact_results
@@ -634,6 +634,8 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   /// @throws std::exception if called pre-finalize.
   const systems::OutputPort<T>& get_body_poses_output_port() const;
 
+  const systems::OutputPort<T>& get_body_pose_output_port(BodyIndex body_index) const;
+
   /// Returns the output port of all body spatial velocities in the world frame.
   /// You can obtain the spatial velocity `V_WB` of a body B in the world frame
   /// W with:
```

etc.

Then the user can find the information they want without the extra work of discovering and then wiring up an additional bespoke system.  (It would also save us the cost of writing and binding several different bespoke selector systems.)  Additionally, it would allow the user to wire up a couple different bodies solely by port index, instead of adding multiple selector systems to the diagram.

Could you discuss the trade-offs of that approach vs what's here currently, and why dedicated system is the best choice for our component library?

---

If we think that "connect a couple of bodies, not just one" is a probably use case, then another design alternative would be to keep this feature as a separate System, but make it a demultiplexer instead of a selector.  We should also discuss whether "demultipliexer" or "selector" is the best design choice.

___
___
*[`multibody/plant/rigid_transform_selector.h` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19520#-NXB6lRU4jldihTRiwQT:-NXB6lRU4jldihTRiwQU:b-uxrc1o) ([raw file](https://github.com/RobotLocomotion/drake/blob/61e621b099c994ef41ef92997a7699c1e3fd6fed/multibody/plant/rigid_transform_selector.h#L18)):*
> ```objc
> // https://github.com/RobotLocomotion/drake/issues/16923 .
> 
> /**
> ```

nit If we keep this as a separate class, it will be difficult for users to discover it.  At minimum, we should to update the API documentation of MultibodyPlant to cite this helper class.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19520)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @RussTedrake)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19520#-NXAWt2Z17xXDWbEgGyc:-NXBtFjQEwyYtS7M8Y_R:batvcot) (no related file):*

It seems to me that it would be strictly easier for the user if MbP provided i'th-body ports directly.

```patch
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -170,6 +170,9 @@ output_ports:
 - body_poses
 - body_spatial_velocities
 - body_spatial_accelerations
+- body_pose_{i}
+- body_spatial_velocity_{i}
+- body_spatial_acceleration_{i}
 - generalized_acceleration
 - reaction_forces
 - contact_results
@@ -634,6 +634,8 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   /// @throws std::exception if called pre-finalize.
   const systems::OutputPort<T>& get_body_poses_output_port() const;
 
+  const systems::OutputPort<T>& get_body_pose_output_port(BodyIndex body_index) const;
+
   /// Returns the output port of all body spatial velocities in the world frame.
   /// You can obtain the spatial velocity `V_WB` of a body B in the world frame
   /// W with:
```

etc.

Then the user can find the information they want without the extra work of discovering and then wiring up an additional bespoke system.  (It would also save us the cost of writing and binding several different bespoke selector systems.)  Additionally, it would allow the user to wire up a couple different bodies solely by port index, instead of adding multiple selector systems to the diagram.

Could you discuss the trade-offs of that approach vs what's here currently, and why dedicated system is the best choice for our component library?

---

If we think that "connect a couple of bodies, not just one" is a probably use case, then another design alternative would be to keep this feature as a separate System, but make it a demultiplexer instead of a selector.  We should also discuss whether "demultipliexer" or "selector" is the best design choice.
</blockquote></details>

Russ just came out of the office and asked my opinion on this idea of having a port per body pose. 
Russ seemed to be worried about how cumbersome the wiring diamgram would look if we exposed 1000 ports for each body. However, after some thought, I believe that would not be that much of a problem if when drawing diagrams we only draw connected ports? It might actually tidy up things. 
The other option Russ came up with is to have some sort of "port registration API", where the user would register the one and only required port on the MBP before wiring it up to the rest of the diagram.

Another thought. @joemasterjohn is starting to work on adding/removing free bodies from an MBP. There will be immediate consequences for these `std::vector<SpatialQuantity>` ports given they are indexed by BodyIndex. We haven't come up with a design and/or solution. But maybe worth to consider this as a constraint in these new APIs? maybe body indexed APIs should no longer be added? Any thoughts on this @joemasterjohn?

___

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19520)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19520#-NXAWt2Z17xXDWbEgGyc:-NXHRPmX3DAxnYq4oBMw:b-80buts) (no related file):*

> Russ seemed to be worried about how cumbersome the wiring diagram would look if we exposed 1000 ports for each body. However, after some thought, I believe that would not be that much of a problem if when drawing diagrams we only draw connected ports? It might actually tidy up things.

Yes, I had all of the same thoughts.

> The other option Russ came up with is to have some sort of "port registration API", where the user would register the one and only required port on the MbP before wiring it up to the rest of the diagram.

Yes, you could image the function `OutputPort& MbP::DeclareBodyPoseOutputPort(BodyIndex)` that the user could call, instead of `get_body_pose_output_port(BodyIndex)`.  I would be fine with that as well.  It would create the requested port on-demand, instead of pre-creating all of them up front.

> There will be immediate consequences for these `std::vector<SpatialQuantity>` ports given they are indexed by BodyIndex.

Any new functions like `get_body_pose_output_port(BodyIndex)` or `DeclareBodyPoseOutputPort(BodyIndex)` will be no worse than all of the other MbP API that assume a dense `BodyIndex`, and porting them to use `BodyId` will be equally easy/hard boilerplate as the other API that already speaks of `BodyIndex` (if that's the route you take).

This is actually an argument against a demultiplexor or selector that is cascaded in the diagram, and a vote for body-specific output ports.
</blockquote></details>

I'm going to move forward with the `DeclareBodyPoseOutputPort` idea.  Will close this PR.

___

PR #19522, opened 2023-06-04
Add support for num_positions != num_velocities in DifferentialIK

RussTedrake:
Resolves a question from StackOverflow:
https://stackoverflow.com/questions/76391333/differential-inverse-kinematics-integrator-does-not-work-with-a-floating-plant

+@sherm1 for feature review, please?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)

### DIFF
--- a/bindings/pydrake/multibody/inverse_kinematics_py_differential.cc
+++ b/bindings/pydrake/multibody/inverse_kinematics_py_differential.cc
@@ -114,14 +114,25 @@ void DefineIkDifferential(py::module m) {
       "DoDifferentialInverseKinematics",
       [](const Eigen::VectorXd& q_current, const Eigen::VectorXd& v_current,
           const Eigen::VectorXd& V, const Eigen::MatrixXd& J,
-          const DifferentialInverseKinematicsParameters& parameters) {
+          const DifferentialInverseKinematicsParameters& parameters,
+          const std::optional<Eigen::Ref<const Eigen::MatrixXd>>& N,
+          const std::optional<Eigen::Ref<const Eigen::MatrixXd>>& Nplus) {
+        std::optional<Eigen::SparseMatrix<double>> N_sparse{std::nullopt};
+        std::optional<Eigen::SparseMatrix<double>> Nplus_sparse{std::nullopt};
+        if (N) {
+          N_sparse = N->sparseView();
+        }
+        if (Nplus) {
+          Nplus_sparse = Nplus->sparseView();
+        }
         return DoDifferentialInverseKinematics(
-            q_current, v_current, V, J, parameters);
+            q_current, v_current, V, J, parameters, N_sparse, Nplus_sparse);
       },
       py::arg("q_current"), py::arg("v_current"), py::arg("V"), py::arg("J"),
-      py::arg("parameters"),
+      py::arg("parameters"), py::arg("N") = std::nullopt,
+      py::arg("Nplus") = std::nullopt,
       doc.DoDifferentialInverseKinematics
-          .doc_5args_q_current_v_current_V_J_parameters);
+          .doc_7args_q_current_v_current_V_J_parameters_N_Nplus);
 
   m.def(
       "DoDifferentialInverseKinematics",
--- a/bindings/pydrake/multibody/test/inverse_kinematics_differential_test.py
+++ b/bindings/pydrake/multibody/test/inverse_kinematics_differential_test.py
@@ -75,6 +75,19 @@ class TestPlanner(unittest.TestCase):
         self.assertTrue(np.allclose(
             result.joint_velocities, [1, 0], atol=1e-8, rtol=0))
         self.assertEqual(result.status, enum.kSolutionFound)
+        result = mut.DoDifferentialInverseKinematics(q_current=[0, 1],
+                                                     v_current=[2, 3],
+                                                     V=[0, 1, 2, 3, 4, 5],
+                                                     J=np.array([
+                                                         [0, 1, 2, 3, 4, 5],
+                                                         [6, 7, 8, 9, 10, 11],
+                                                     ]).T,
+                                                     parameters=params,
+                                                     N=np.eye(2),
+                                                     Nplus=np.eye(2))
+        self.assertTrue(np.allclose(
+            result.joint_velocities, [1, 0], atol=1e-8, rtol=0))
+        self.assertEqual(result.status, enum.kSolutionFound)
 
     def test_mbp_overloads(self):
         file_name = FindResourceOrThrow(
--- a/multibody/inverse_kinematics/BUILD.bazel
+++ b/multibody/inverse_kinematics/BUILD.bazel
@@ -168,6 +168,7 @@ drake_cc_googletest(
     deps = [
         ":differential_inverse_kinematics",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//manipulation/kuka_iiwa:iiwa_constants",
         "//multibody/parsing",
     ],
--- a/multibody/inverse_kinematics/differential_inverse_kinematics.cc
+++ b/multibody/inverse_kinematics/differential_inverse_kinematics.cc
@@ -12,6 +12,8 @@
 namespace drake {
 namespace multibody {
 
+using Eigen::SparseMatrix;
+
 namespace internal {
 DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
     const Eigen::Ref<const VectorX<double>>& q_current,
@@ -19,7 +21,9 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
     const math::RigidTransform<double>& X_WE,
     const Eigen::Ref<const Matrix6X<double>>& J_WE_W,
     const SpatialVelocity<double>& V_WE_desired,
-    const DifferentialInverseKinematicsParameters& parameters) {
+    const DifferentialInverseKinematicsParameters& parameters,
+    const std::optional<Eigen::Ref<const SparseMatrix<double>>>& N,
+    const std::optional<Eigen::Ref<const SparseMatrix<double>>>& Nplus) {
   const math::RotationMatrix<double> R_EW = X_WE.rotation().transpose();
   const SpatialVelocity<double> V_WE_E = R_EW * V_WE_desired;
 
@@ -44,8 +48,9 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
 
   return DoDifferentialInverseKinematics(
       q_current, v_current, V_WE_E_with_flags.head(num_cart_constraints),
-      J_WE_E_with_flags.topRows(num_cart_constraints), parameters);
+      J_WE_E_with_flags.topRows(num_cart_constraints), parameters, N, Nplus);
 }
+
 }  // namespace internal
 
 std::ostream& operator<<(std::ostream& os,
@@ -105,7 +110,7 @@ DifferentialInverseKinematicsParameters::
       num_velocities_(num_velocities.value_or(num_positions)),
       nominal_joint_position_(VectorX<double>::Zero(num_positions)),
       joint_centering_gain_(
-          MatrixX<double>::Zero(num_velocities_, num_positions)) {
+          MatrixX<double>::Zero(num_positions_, num_positions)) {
   DRAKE_DEMAND(num_positions > 0);
   DRAKE_DEMAND(num_velocities > 0);
 }
@@ -115,7 +120,9 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
     const Eigen::Ref<const VectorX<double>>& v_current,
     const Eigen::Ref<const VectorX<double>>& V,
     const Eigen::Ref<const MatrixX<double>>& J,
-    const DifferentialInverseKinematicsParameters& parameters) {
+    const DifferentialInverseKinematicsParameters& parameters,
+    const std::optional<Eigen::Ref<const SparseMatrix<double>>>& N,
+    const std::optional<Eigen::Ref<const SparseMatrix<double>>>& Nplus) {
   const int num_positions = parameters.get_num_positions();
   const int num_velocities = parameters.get_num_velocities();
   const double dt = parameters.get_time_step();
@@ -124,9 +131,6 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
   DRAKE_DEMAND(v_current.size() == num_velocities);
   DRAKE_DEMAND(J.rows() == num_cart_constraints);
   DRAKE_DEMAND(J.cols() == num_velocities);
-  // A bunch of the operations below assume num_positions == num_velocities.
-  DRAKE_DEMAND(num_positions == num_velocities);
-
   const auto identity_num_positions =
       MatrixX<double>::Identity(num_positions, num_positions);
 
@@ -141,15 +145,30 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
   const double kPrimaryObjectiveGain = 100;
   prog.AddLinearCost(-Vector1d{kPrimaryObjectiveGain}, 0.0, alpha);
 
-  // | P * (v_next - K * (q_nominal - q_current)) |^2
+  // |P⋅(v_next - N⁺(q)⋅K⋅(q_nominal - q_current))|²
   const Eigen::FullPivLU<MatrixX<double>> lu(J);
   if (lu.rank() < num_velocities) {
     const Eigen::MatrixXd P = lu.kernel().transpose();
-    prog.Add2NormSquaredCost(
-        P,
-        P * parameters.get_joint_centering_gain() *
-            (parameters.get_nominal_joint_position() - q_current),
-        v_next);
+    if (Nplus) {
+      DRAKE_DEMAND(Nplus->rows() == num_velocities);
+      DRAKE_DEMAND(Nplus->cols() == num_positions);
+      prog.Add2NormSquaredCost(
+          P,
+          P * (*Nplus) * parameters.get_joint_centering_gain() *
+              (parameters.get_nominal_joint_position() - q_current),
+          v_next);
+    } else if (num_positions != num_velocities) {
+      throw std::runtime_error(
+          "You must pass the Nplus matrix to DoDifferentialInverseKinematics "
+          "when J is not full column rank and num_positions != "
+          "num_velocities.");
+    } else {
+      prog.Add2NormSquaredCost(
+          P,
+          P * parameters.get_joint_centering_gain() *
+              (parameters.get_nominal_joint_position() - q_current),
+          v_next);
+    }
     quadratic_cost = true;
   }
 
@@ -165,12 +184,26 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
   // 0 <= alpha <= 1
   prog.AddBoundingBoxConstraint(0, 1, alpha);
 
-  // joint_lim_min <= q_current + v_next*dt <= joint_lim_max
+  // joint_lim_min <= q_current + N⋅v_next⋅dt <= joint_lim_max
   if (parameters.get_joint_position_limits()) {
-    prog.AddBoundingBoxConstraint(
-        (parameters.get_joint_position_limits()->first - q_current) / dt,
-        (parameters.get_joint_position_limits()->second - q_current) / dt,
-        v_next);
+    const Eigen::VectorXd lb =
+        (parameters.get_joint_position_limits()->first - q_current) / dt;
+    const Eigen::VectorXd ub =
+        (parameters.get_joint_position_limits()->second - q_current) / dt;
+    if (N) {
+      DRAKE_DEMAND(N->rows() == num_positions);
+      DRAKE_DEMAND(N->cols() == num_velocities);
+      // We cannot currently use prog.AddLinearConstraint with a SparseMatrix.
+      prog.AddConstraint(
+          std::make_shared<solvers::LinearConstraint>(*N, lb, ub), v_next);
+    } else if (num_positions != num_velocities) {
+      throw std::runtime_error(
+          "You must pass the N matrix to DoDifferentialInverseKinematics "
+          "when you have joint position limits and num_positions != "
+          "num_velocities.");
+    } else {
+      prog.AddBoundingBoxConstraint(lb, ub, v_next);
+    }
   }
 
   // joint_vel_lim_min <= v_next <= joint_vel_lim_max
@@ -237,9 +270,17 @@ DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
                                     frame_E, Vector3<double>::Zero(),
                                     frame_W, frame_W, &J_WE);
 
+  std::optional<Eigen::SparseMatrix<double>> N = std::nullopt;
+  std::optional<Eigen::SparseMatrix<double>> Nplus = std::nullopt;
+  if (!plant.IsVelocityEqualToQDot()) {
+    if (parameters.get_joint_position_limits()) {
+      N = plant.MakeVelocityToQDotMap(context);
+    }
+    Nplus = plant.MakeQDotToVelocityMap(context);
+  }
   return internal::DoDifferentialInverseKinematics(
-      plant.GetPositions(context), plant.GetVelocities(context),
-      X_WE, J_WE, SpatialVelocity<double>(V_WE_desired), parameters);
+      plant.GetPositions(context), plant.GetVelocities(context), X_WE, J_WE,
+      SpatialVelocity<double>(V_WE_desired), parameters, N, Nplus);
 }
 
 DifferentialInverseKinematicsResult DoDifferentialInverseKinematics(
--- a/multibody/inverse_kinematics/differential_inverse_kinematics.h
+++ b/multibody/inverse_kinematics/differential_inverse_kinematics.h
@@ -64,9 +64,6 @@ class DifferentialInverseKinematicsParameters {
    * @param num_positions Number of generalized positions.
    * @param num_velocities Number of generalized velocities (by default it will
    * be set to num_positions).
-   *
-   * @note Currently DoDifferentialInverseKinematics requires that
-   * num_positions == num_velocities.
    */
   DifferentialInverseKinematicsParameters(
       int num_positions, std::optional<int> num_velocities = std::nullopt);
@@ -157,11 +154,11 @@ class DifferentialInverseKinematicsParameters {
 
   /**
    * Sets the joint centering gain, K, so that the joint centering command is
-   * v_next = K * (q_nominal - q_current).
-   * @pre K must be num_velocities x num_positions.
+   * attempting to achieve v_next = N⁺(q) * K * (q_nominal - q_current).
+   * @pre K must be num_positions x num_positions.
    */
   void set_joint_centering_gain(const MatrixX<double>& K) {
-    DRAKE_DEMAND(K.rows() == num_velocities_);
+    DRAKE_DEMAND(K.rows() == num_positions_);
     DRAKE_DEMAND(K.cols() == num_positions_);
     joint_centering_gain_ = K;
   }
@@ -287,13 +284,13 @@ class DifferentialInverseKinematicsParameters {
  * MathematicalProgram:
  *
  * ```
- *   min_{v_next,alpha}  -100 * alpha
- *                         + | P * (v_next - K * (q_nominal - q_current)) |^2
+ *   min_{v_next,alpha}
+ *     -100 * alpha + |P⋅(v_next - N⁺(q)⋅K⋅(q_nominal - q_current))|²
  *
  *   s.t.
- *     J * v_next = alpha * V, // J * v_next has the same direction as V
+ *     J⋅v_next = alpha⋅V, // J⋅v_next has the same direction as V
  *     0 <= alpha <= 1,        // Never go faster than V
- *     joint_lim_min <= q_current + v_next*dt <= joint_lim_max,
+ *     joint_lim_min <= q_current + N⋅v_next⋅dt <= joint_lim_max,
  *     joint_vel_lim_min <= v_next <= joint_vel_lim_max,
  *     joint_accel_lim_min <= (v_next - v_current)/dt <= joint_accel_lim_max,
  *     and additional linear velocity constraints,
@@ -301,7 +298,7 @@ class DifferentialInverseKinematicsParameters {
  * where:
  *   - The rows of P form an orthonormal basis for the nullspace of J,
  *   - J.rows() == V.size(),
- *   - J.cols() == v_current.size() == q_current.size() == v_next.size(),
+ *   - J.cols() == v_current.size() == v_next.size(),
  *   - V can have any size, with each element representing a constraint on the
  *     solution (6 constraints specifying an end-effector spatial velocity is
  *     typical, but not required),
@@ -312,8 +309,8 @@ class DifferentialInverseKinematicsParameters {
  * Intuitively, this finds a v_next such that J*v_next is in the same direction
  * as V, and the difference between |V| and |J * v_next| is minimized while all
  * constraints in @p parameters are satisfied as well. In the nullspace of this
- * objective, we have a secondary objective to minimize |v_next - K *
- * (q_nominal - q_current)|.
+ * objective, we have a secondary objective to minimize |v_next -
+ * N⁺(q)⋅K⋅(q_nominal - q_current)|².
  *
  * For more details, see
  * https://manipulation.csail.mit.edu/pick.html#diff_ik_w_constraints .
@@ -330,10 +327,18 @@ class DifferentialInverseKinematicsParameters {
  * @param V Desired spatial velocity. It must have the same number of rows as
  * @p J.
  * @param J Jacobian with respect to generalized velocities v. It must have the
- * same number of rows as @p V. J * v need to represent the same spatial
+ * same number of rows as @p V. J * v needs to represent the same spatial
  * velocity as @p V.
  * @param parameters Collection of various problem specific constraints and
  * constants.
+ * @param N (optional) matrix which maps q̇ = N(q)⋅v. See
+ * MultibodyPlant::MakeVelocityToQDotMap(). By default, it is taken to be the
+ * identity matrix. If dim(q) != dim(v) and any joint position limits are set in
+ * `parameters`, then you *must* provide N.
+ * @param Nplus (optional) matrix which maps q̇ = N⁺(q)⋅v. See
+ * MultibodyPlant::MakeQDotToVelocityMap(). By default, it is taken to be the
+ * identity matrix. If dim(q) != dim(v) and J is not full column rank, then you
+ * *must* provide Nplus.
  * @return If the solver successfully finds a solution, joint_velocities will
  * be set to v, otherwise it will be nullopt.
  *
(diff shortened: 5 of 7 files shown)


sherm1 (commented):
Feature :lgtm: with a few minor comments

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NX6i06O0W5VuH7CGlVr:br5gxmj)* status: 4 unresolved discussions, needs at least two assigned reviewers (waiting on @RussTedrake)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.h` line 337 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6hhBNEEibFWVAM_Av:-NX6hhBNEEibFWVAM_Aw:b-30gauy) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.h#L337)):*
> ```objc
>  * MultibodyPlant::MapQDotToVelocity() for more details. By default, it is
>  * taken to be the identity matrix. If dim(q) != dim(v), then you *must*
>  * provide Nplus.
> ```

BTW consider whether the user might need some help constructing N+. You could expose the function you wrote and mention it here, or just document the approach of making multiple calls to MapQDotToVelocity(). I'm not sure that will be obvious to users.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 54 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6e0SQ7wHnvURteIRW:-NX6e0SQ7wHnvURteIRX:b-dt4skp) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L54)):*
> ```cpp
> // Makes the matrix Nplus(q) by calling MapQDotToVelocity repeatedly.
> Eigen::MatrixXd MakeQDotToVelocityMap(const MultibodyPlant<double>& plant,
>                                const systems::Context<double>& context) {
> ```

nit: odd indent
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NX6fQSz8DpDgDl9qrqX:b-jh4oej) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*
> ```cpp
>                                     frame_W, frame_W, &J_WE);
> 
>   const Eigen::MatrixXd Nplus = MakeQDotToVelocityMap(plant, context);
> ```

BTW should this be conditional on npositions != nvelocities?
___
*[`multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc` line 280 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6gQNhB2ZBW3Hrbr3-:-NX6gQNhB2ZBW3Hrbr30:b-s340c) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc#L280)):*
> ```cpp
>   DifferentialInverseKinematicsResult result =
>       DoDiffIKForSpatialVelocity(V_WE_W);
>   drake::log()->info("result.status = {}", result.status);
> ```

BTW do you still want this line in the code?

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NXBHJq62nGISPs88HIJ:bcav8r0)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.h` line 337 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6hhBNEEibFWVAM_Av:-NXBGoipEhCKe5UxE8hr:b-hlwc6h) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.h#L337)):*

BTW consider whether the user might need some help constructing N+. You could expose the function you wrote and mention it here, or just document the approach of making multiple calls to MapQDotToVelocity(). I'm not sure that will be obvious to users.
</blockquote></details>

I've made a bigger breadcrumb.  I think that's enough for now; most people who have MultibodyPlant will use the MultibodyPlant overloads... not hand in Nplus themselves.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXBH4bNEN692W32TPbT:bj7z63m) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

BTW should this be conditional on npositions != nvelocities?
</blockquote></details>

I thought about it, but that's not actually safe.  We have some joints where num_positions == num_velocities, but Nplus is not the identity.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NX6nTot5-3CgRb-dlh8:bed45nn) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

I thought about it, but that's not actually safe.  We have some joints where num_positions == num_velocities, but Nplus is not the identity.
</blockquote></details>

That makes sense, but in that case we should assign someone from Dex Manip (perhaps Eric or Calder) to empirically measure whether the performance regression here (for robots that only use revolute joints) will break their realtime robots.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXBOZmA4XHr653DDz0Q:b-kews6u) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

That makes sense, but in that case we should assign someone from Dex Manip (perhaps Eric or Calder) to empirically measure whether the performance regression here (for robots that only use revolute joints) will break their realtime robots.
</blockquote></details>

Ah, right I withdraw that comment. We would need instead a plant check on whether N is identity. 

If performance is an issue the handling of N should be done via the plant operators (MapVelocityToQDot() etc.) rather than by forming explicit matrices. The full-size N is quite big but is block diagonal -- the operators exploit that.

ggould-tri (commented):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXBdwJe7QbFTC4FtKq0:b-58uxgv) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

Ah, right I withdraw that comment. We would need instead a plant check on whether N is identity. 

If performance is an issue the handling of N should be done via the plant operators (MapVelocityToQDot() etc.) rather than by forming explicit matrices. The full-size N is quite big but is block diagonal -- the operators exploit that.
</blockquote></details>

FYI Calder is out this week.  Pinging @EricCousineau-TRI .
___
*[`multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc` line 75 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXBcTgCBgzNB5HoDYXc:-NXBcTgDDh9tUeN6h3Is:b-rz755d) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc#L75)):*
> ```cpp
>     // Set the initial plant state.
>     VectorXd q = plant_->GetPositions(*context_);
>     // Set the revolute joints. These values were randomly generated.
> ```

minor:  If it's useful to set random state on the revolute joints, it's probably also useful to set it on the floating joint.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXDHe6f7dL8xGEzwP_S:b-8wuzms) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

FYI Calder is out this week.  Pinging @EricCousineau-TRI .
</blockquote></details>

A few thoughts.
1) we can't easily use MapVelocityToQDot directly, because this (admittedly old) API has separated the diffIK implementation from the multibodyplant.
2) the current implementation would give wrong answers, since the num_positions == num_velocities check is insufficient, so I feel that we must either land a version of this change, or provide a way to guard against unsupported usage.

A few possible solutions:
1) leave the implementation as I have it here. for 7 dof robots I suspect the change is in the noise.
2) change the entry point I've proposed for Nplus to use sparse matrices, though I think the new cost in in calling MapQDotToVelocity num_positions times more than multiplying by the identity matrix.
3) change the entry point for Nplus to instead take a function handle to a MapQDotToVelocity implementation, which would passing in a lambda to MultibodyPlant.
4) refactor the implementation so that the multibody plant variants don't go through the current overload which takes J directly.

wdyt?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXDNCGA0P3XqyHTb72F:b-fxepi0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*
> I think the new cost is in calling MapQDotToVelocity `num_positions` times, more than multiplying by the identity matrix.

Yes, I'd also bet on that.

If I understand correctly there are more options:

(5) Change this local variable `Nplus` to be optional and only assign `MakeQDotToVelocityMap` to it when there are are nontrivial joints.  This would probably require extra accessor(s) on MbP to report whether this is the case (i.e., whether Nplus is really just the identity).  Maybe that MbP function would be generally useful, too. We have joint characterization functions like `can_rotate` and `can_translate` already; another function `bool Joint::is_qdot_identically_v` seems in tune with existing practice.

(6) Reduce the penalty here from O(N^2) to O(N) by setting `Nplus` using the mobilizer API instead of the MbP API.  That would mean calling MbT code that lives inside of `namespace internal`, but `internal` primarily means that users should avoid it; to me, our Diff IK code that needs to run at 1 kHz is fair game for calling into the "internal" tree.

For all I know, option (1) is just fine.  My point is only that Sherm does not seem like a sufficient feature reviewer for this aspect of the change, rather that we should have someone who uses this code regularly weigh in on how much of a downside this would be in practice.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXGen1mAzEYMka82GgS:b-yx89db) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

> I think the new cost is in calling MapQDotToVelocity `num_positions` times, more than multiplying by the identity matrix.

Yes, I'd also bet on that.

If I understand correctly there are more options:

(5) Change this local variable `Nplus` to be optional and only assign `MakeQDotToVelocityMap` to it when there are are nontrivial joints.  This would probably require extra accessor(s) on MbP to report whether this is the case (i.e., whether Nplus is really just the identity).  Maybe that MbP function would be generally useful, too. We have joint characterization functions like `can_rotate` and `can_translate` already; another function `bool Joint::is_qdot_identically_v` seems in tune with existing practice.

(6) Reduce the penalty here from O(N^2) to O(N) by setting `Nplus` using the mobilizer API instead of the MbP API.  That would mean calling MbT code that lives inside of `namespace internal`, but `internal` primarily means that users should avoid it; to me, our Diff IK code that needs to run at 1 kHz is fair game for calling into the "internal" tree.

For all I know, option (1) is just fine.  My point is only that Sherm does not seem like a sufficient feature reviewer for this aspect of the change, rather that we should have someone who uses this code regularly weigh in on how much of a downside this would be in practice.
</blockquote></details>

The code I am using actually does not use Diff IK directly, so it will not affect that portion.
However, code in the Continual side of things may still be using `MoveTool` type primitives.

I will check w/ that team to see if someone can run the corresponding code.
(My gut feeling is that these changes shouldn't matter too much, though)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXGgT0s8XYic67mjwuy:b-9duvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

The code I am using actually does not use Diff IK directly, so it will not affect that portion.
However, code in the Continual side of things may still be using `MoveTool` type primitives.

I will check w/ that team to see if someone can run the corresponding code.
(My gut feeling is that these changes shouldn't matter too much, though)
</blockquote></details>

Sorry, "the code I am using" => "the code the Intuitive Team is using"

Some additional minor notes:
- For option (1) - the Continual code using this can be up to 14-DoF.
- There could also be a Option (6) - expose API to compute N and Nplus directly (perhaps block sparse, if helpful)
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXGhe0Q4vr7RQ3BNCcr:buzlwwn) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*
> ```cpp
>     if (Nplus) {
>       // We use sparse matrix multiplication here to handle limits with inf.
>       // (Multiplying the dense matrix results in nans, because 0*inf => nan).
> ```

BTW Could this be sidestepped by also passing `N`? e.g.

```
struct VelocityMapping {
  Ref<MatrixXd> N;
  Ref<MatrixXd> Nplus;
};
```

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXGrCWj3TWFlyt6XFw2:b2j2ebf) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

Sorry, "the code I am using" => "the code the Intuitive Team is using"

Some additional minor notes:
- For option (1) - the Continual code using this can be up to 14-DoF.
- There could also be a Option (6) - expose API to compute N and Nplus directly (perhaps block sparse, if helpful)
</blockquote></details>

I'm going to implement an MbP::is_qdot_identically_v, not only because of this PR, but because I suspect we have number of instances of num_positions==num_velocities being an insufficient check for assuming that qdot==v.  Let me land that first, then we can check the performance.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: LGTM missing from assignee ggould-tri(platform)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NX6fQSz8DpDgDl9qrqW:-NXWvnN24P_zK_NWwEb0:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f7c7f43686ab81f77a4503359ebfce8e91ca2afc/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L279)):*

I'm going to implement an MbP::is_qdot_identically_v, not only because of this PR, but because I suspect we have number of instances of num_positions==num_velocities being an insufficient check for assuming that qdot==v.  Let me land that first, then we can check the performance.
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXMZ8pBCdH9Nz_5xd_3:bfyo0py) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

BTW Could this be sidestepped by also passing `N`? e.g.

```
struct VelocityMapping {
  Ref<MatrixXd> N;
  Ref<MatrixXd> Nplus;
};
```
</blockquote></details>

That would be definitely worse, I think.  It would change a bounding box constraint into a linear constraint.
___
*[`multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc` line 75 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXBcTgCBgzNB5HoDYXc:-NXX1tyIEHGtdwFMaa-d:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/test/differential_inverse_kinematics_test.cc#L75)):*

minor:  If it's useful to set random state on the revolute joints, it's probably also useful to set it on the floating joint.
</blockquote></details>

Done.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.h` line 157 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXjtVIu52xQp0hKjkFk:-NXk0zlM7N2U7796R6d4:b-xoe7hb) ([raw file](https://github.com/RobotLocomotion/drake/blob/22fabeac641a5306c49a370ce629b4b0214b5ad8/multibody/inverse_kinematics/differential_inverse_kinematics.h#L157)):*

nit Usage of `v_next` for position-sized values is a bit confusing.

Should this instead be `qd_next`?
Or should it instead read `v_next =  N⁺*K*(q_nominal - q_current)` as you have below?
</blockquote></details>

Done. Yikes. I missed that one.  Good catch.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.h` line 161 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXjtSOl6ieem19O0t6-:-NXk11vJ1n4u2SfkZIlb:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/22fabeac641a5306c49a370ce629b4b0214b5ad8/multibody/inverse_kinematics/differential_inverse_kinematics.h#L161)):*

nit This is now out of sync with precondition?
</blockquote></details>

Done.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXk-_6qBFdoUHI33rhS:b7vy2yu) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

Ah, that brings up a new but related question - what if the mapping of N⁺ implies not just scaling + translation, but rotation as well?
At that point, I think we'd need a linear constraint, not just bounding box?
</blockquote></details>

No.  ?

The constraint is 
```
joint_lim_min <= q_current + N*v_next*dt <= joint_lim_max,
```
which I'm implementing as
```
Nplus*(joint_limit_min - q_current)/dt <= v_next
```
etc.  That's always a bounding box constraints on v_next.  In diffIK, the point is *always* that q is fixed, so the nonlinear terms become constants.  ?

jwnimmer-tri (commented):
I'm assuming -(status: do not review) now that its rebased.

> Reviewed 1 of 4 files at r3, 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NXk3BedCnmdkxOG5SHu:b-ttzbe0)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform)

EricCousineau-TRI (commented):
Yup, can do!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NXkADHl6HZoEYPHxS37:b-s9xy4d)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXk6phI02eeMoHzp19R:b-gh54sr) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

No.  ?

The constraint is 
```
joint_lim_min <= q_current + N*v_next*dt <= joint_lim_max,
```
which I'm implementing as
```
Nplus*(joint_limit_min - q_current)/dt <= v_next
```
etc.  That's always a bounding box constraints on v_next.  In diffIK, the point is *always* that q is fixed, so the nonlinear terms become constants.  ?
</blockquote></details>

Aye, I get that q is fixed + constant. But multiplying all sides of the linear constraint by Nplus and expecting a bounding box with limits in the same order doesn't "smell" right.
e.g., if Nplus implies rotation or mirroring, then either the bounding box is misaligned or may even be flipped from what it needs.

I can dismiss myself, though, and follow up with a more concrete example, if you'd like!

EricCousineau-TRI (commented):
:lgtm: platform

I'll work on creating a concrete example of rotation / bounding box, but feel free to dismiss me to move forward!

> Reviewed 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NXkIhnb-YJrFQVFl7PW:b-3740dm)* status: 1 unresolved discussion

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: :shipit: complete! all discussions resolved, LGTM from assignees EricCousineau-TRI(platform),sherm1(platform)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXkfIlk730lNeHA2OpD:b-ui5vp9) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

Aye, I get that q is fixed + constant. But multiplying all sides of the linear constraint by Nplus and expecting a bounding box with limits in the same order doesn't "smell" right.
e.g., if Nplus implies rotation or mirroring, then either the bounding box is misaligned or may even be flipped from what it needs.

I can dismiss myself, though, and follow up with a more concrete example, if you'd like!
</blockquote></details>

Here's a toy example: https://nbviewer.org/github/EricCousineau-TRI/repro/blob/58c9ea9/drake_stuff/notebooks/rotation_bounding_box.ipynb
In this case, I think we just need to handle sign flips (if our `N` Jacobians are event capable of it), so doing something like this could help:

```
lb_fixed = lb.array().min(ub.array())
ub_fixed = ub.array().max(lb.array())
```

But we can also hold off until (if) we actually run into this edge case.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 1 unresolved discussion

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXktTx_0dFUGmSSxt4d:b-2wv1c4) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

Here's a toy example: https://nbviewer.org/github/EricCousineau-TRI/repro/blob/58c9ea9/drake_stuff/notebooks/rotation_bounding_box.ipynb
In this case, I think we just need to handle sign flips (if our `N` Jacobians are event capable of it), so doing something like this could help:

```
lb_fixed = lb.array().min(ub.array())
ub_fixed = ub.array().max(lb.array())
```

But we can also hold off until (if) we actually run into this edge case.
</blockquote></details>

Working: Ah, made a mistake, and with fix it makes me more concerned about this approach. Will post.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 1 unresolved discussion

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXkvnkyAmGagrPlKzPp:b45za1t) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

Working: Ah, made a mistake, and with fix it makes me more concerned about this approach. Will post.
</blockquote></details>

Fixed the mistake, now it shows what I was trying to describe:
https://nbviewer.org/github/EricCousineau-TRI/repro/blob/4346589/drake_stuff/notebooks/rotation_bounding_box.ipynb

Basically, if the Jacobian `N` involves rotation (or perhaps even skew), then doing algebra against the bounding box bounds vs. keeping it a linear constraint may lose the geometry of the intended constraint.
Does this make sense / seem like a more sensible defect now?

RussTedrake (commented):
once more for the win.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NXmQrkZ1RDnph1mbauF:b-cm0npz)* status: 1 unresolved discussion

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXmQLsf-8ygY3NC9RfB:bpeal7t) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

Fixed the mistake, now it shows what I was trying to describe:
https://nbviewer.org/github/EricCousineau-TRI/repro/blob/4346589/drake_stuff/notebooks/rotation_bounding_box.ipynb

Basically, if the Jacobian `N` involves rotation (or perhaps even skew), then doing algebra against the bounding box bounds vs. keeping it a linear constraint may lose the geometry of the intended constraint.
Does this make sense / seem like a more sensible defect now?
</blockquote></details>

I don't agree with your rotation example, but I agree that I was not carefully checking for the sign flips (that N is monotone).  Rather than check it on every call, I've gone ahead and settled for the linear constraint.  I do think that's better.

e.g.
https://math.stackexchange.com/questions/368260/monotone-matrix
https://math.stackexchange.com/questions/1294088/inversion-of-a-matrix-in-a-system-of-linear-inequalities

EricCousineau-TRI (commented):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXp50SE8BK_7GogZEhL:-NXp50SE8BK_7GogZEhM:bm113lg) (no related file):*
FYI CI is failing with `realloc(): invalid next size`
https://drake-cdash.csail.mit.edu/test/1085707607

___
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXGhe0Q4vr7RQ3BNCcq:-NXp3Bed11flxynyLPfH:b-jib7jh) ([raw file](https://github.com/RobotLocomotion/drake/blob/4926a4e13cff6ce684b4873c251b6982f556ec84/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L198)):*

I don't agree with your rotation example, but I agree that I was not carefully checking for the sign flips (that N is monotone).  Rather than check it on every call, I've gone ahead and settled for the linear constraint.  I do think that's better.

e.g.
https://math.stackexchange.com/questions/368260/monotone-matrix
https://math.stackexchange.com/questions/1294088/inversion-of-a-matrix-in-a-system-of-linear-inequalities
</blockquote></details>

OK Gotcha, thank you for the info!

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522)* status: 2 unresolved discussions

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 63 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXpWUXy2Pcz6tCgpJwG:-NXpWUXy2Pcz6tCgpJwH:b-c91x74) ([raw file](https://github.com/RobotLocomotion/drake/blob/14eab9f07b11f90a9862585d3bf2342ea8f8c827/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L63)):*
> ```cpp
>   Eigen::VectorXd qdot(plant.num_positions());
>   Eigen::VectorXd v = Eigen::VectorXd::Zero(plant.num_velocities());
>   for (int i = 0; i < plant.num_positions(); ++i) {
> ```

nit I believe this should be `i < plant.num_velocities()`
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 180 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXpWSFG2LvUqr6Si0y0:-NXpWSFG2LvUqr6Si0y1:bydkym5) ([raw file](https://github.com/RobotLocomotion/drake/blob/14eab9f07b11f90a9862585d3bf2342ea8f8c827/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L180)):*
> ```cpp
>     const Eigen::MatrixXd P = lu.kernel().transpose();
>     prog.Add2NormSquaredCost(
>         (N ? P * (*N) : P),
> ```

nit I think CI failure is because `P` is for kernel of `J_v`, not `J_qdot`.

I can't immediately think of a way to do this without requiring `Nplus` / `J_qdot`, or doing a pseudoinverse on `N` / adding `qdot` and constraining `qdot = N * v`

RussTedrake (commented):
PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NZ0eHqu6XYwYh0-fr2W:b-83330m)* status: :shipit: complete! all discussions resolved, LGTM from assignees EricCousineau-TRI(platform),sherm1(platform)

___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 63 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXpWUXy2Pcz6tCgpJwG:-NXrVJ18AYvYCna6ij35:b-7elk7u) ([raw file](https://github.com/RobotLocomotion/drake/blob/14eab9f07b11f90a9862585d3bf2342ea8f8c827/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L63)):*

nit I believe this should be `i < plant.num_velocities()`
</blockquote></details>

Done.  :man_facepalming:  Thanks.
___
*[`multibody/inverse_kinematics/differential_inverse_kinematics.cc` line 180 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-NXpWSFG2LvUqr6Si0y0:-NZ0XUwi5Ndl52LBBBak:b-czos6x) ([raw file](https://github.com/RobotLocomotion/drake/blob/14eab9f07b11f90a9862585d3bf2342ea8f8c827/multibody/inverse_kinematics/differential_inverse_kinematics.cc#L180)):*

nit I think CI failure is because `P` is for kernel of `J_v`, not `J_qdot`.

I can't immediately think of a way to do this without requiring `Nplus` / `J_qdot`, or doing a pseudoinverse on `N` / adding `qdot` and constraining `qdot = N * v`
</blockquote></details>

Done.  Thanks.  I needed to step back and be more deliberate.  Hopefully good now.

EricCousineau-TRI (commented):
Looks great!

> Reviewed 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19522#-:-NZ1OAjAEcR2ORBzAEnU:b9bkkp1)* status: :shipit: complete! all discussions resolved, LGTM from assignees EricCousineau-TRI(platform),sherm1(platform)

Issue #19437, opened 2023-05-19
Multi-camera simulation performance using triggered async computation


jwnimmer-tri:
This builds on some of the discussions in #10387.  In that ticket, we noticed that simulations with 1 color camera and a lot of depth cameras were pretty slow (in terms of simulation throughput, i.e., realtime rate).  We fixed that by implementing `RenderEngineGl` which was roughly 2x faster at depth rendering than `RenderEngineVtk`.

In the brainstorming discussions there, I proposed the following idea (which we didn't implement at the time):

https://github.com/RobotLocomotion/drake/issues/10387#issuecomment-466042957 (pasted here for convenience):

> Another thing to keep in mind: 8-12 cameras with a global, sync'd shutter are not exactly the physically realistic model, though possibly a reasonable compromise for performance for now. But in reality, each camera in the scene has its own timer. Also, the time from a camera shutter exposure to the LCM image message actually being received by the controller for processing is non-zero, and I wonder if your current benchmark is accounting for that -- i.e., do you have a delay line (or equivalent) between the camera snapshot and the code that uses it (probably, the LCM publisher)? Another way to achieve better image throughput might be to have a timed event on a (single) camera system kick off the render, but don't publish the result of that render until simulation time has advanced by some N micro- or milliseconds. If in reality if we have an, e.g., 2ms delay line for shutter-to-transmit, then your current 5ms wallclock render delay with the VTK camera is cut down to just 1ms wallclock (5ms wallclock - 2ms simclock / 50% realtime vs simtime rate). That leaves the door open to having desync'd cameras as part of a simulation, is more physically meaningful, and isn't that far away from reaching realtime if we can then nudge the single-camera rendering delay down just a little bit more.

Subsequent discussions in that issue clarified and detailed exactly what that all means.  Refer to the link to learn more.

We've again reached a point where we'd like our camera simulations to run closer to realtime, with higher camera image sizes and fps rates.

My plan is to prototype this approach and see if it can deliver the anticipated benefits.

jwnimmer-tri:
For the record -- I was initially aiming this at `RenderEngineGl` async, but most immediately I'll be targeting ``RenderEngineGltfClient`.  The latter should already be thread-safe, and some RPC servers are very slow where the async will still help a lot.

PR #19525, opened 2023-06-04
[geometry] Add SceneGraphInspector::GetAllSourceIds

jwnimmer-tri:
This enables downstream code to know which input ports and frame kinematics to provide to (or expect from) a SceneGraph.

Towards #19437.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_scene_graph.cc
+++ b/bindings/pydrake/geometry/geometry_py_scene_graph.cc
@@ -52,9 +52,9 @@ void DoScalarDependentDefinitions(py::module m, T) {
     cls  // BR
          // Scene-graph wide data.
         .def("num_sources", &Class::num_sources, cls_doc.num_sources.doc)
-        .def("num_frames", &Class::num_frames, cls_doc.num_frames.doc);
-
-    cls  // BR
+        .def("num_frames", &Class::num_frames, cls_doc.num_frames.doc)
+        .def("GetAllSourceIds", &Class::GetAllSourceIds,
+            cls_doc.GetAllSourceIds.doc)
         .def("GetAllFrameIds", &Class::GetAllFrameIds,
             cls_doc.GetAllFrameIds.doc)
         .def("world_frame_id", &Class::world_frame_id,
--- a/bindings/pydrake/geometry/test/scene_graph_test.py
+++ b/bindings/pydrake/geometry/test/scene_graph_test.py
@@ -85,6 +85,7 @@ class TestGeometrySceneGraph(unittest.TestCase):
         inspector = scene_graph.model_inspector()
         self.assertEqual(inspector.num_sources(), 2)
         self.assertEqual(inspector.num_frames(), 3)
+        self.assertEqual(len(inspector.GetAllSourceIds()), 2)
         self.assertEqual(len(inspector.GetAllFrameIds()), 3)
         self.assertTrue(inspector.world_frame_id()
                         in inspector.GetAllFrameIds())
--- a/geometry/geometry_state.cc
+++ b/geometry/geometry_state.cc
@@ -320,6 +320,20 @@ GeometryState<T>::GetCollisionCandidates() const {
 }
 
 template <typename T>
+std::vector<SourceId> GeometryState<T>::GetAllSourceIds() const {
+  std::vector<SourceId> result;
+  result.reserve(source_frame_id_map_.size());
+  result.push_back(self_source_);
+  for (const auto& [source_id, _] : source_names_) {
+    if (source_id != self_source_) {
+      result.push_back(source_id);
+    }
+  }
+  std::sort(result.begin() + 1, result.end());
+  return result;
+}
+
+template <typename T>
 bool GeometryState<T>::SourceIsRegistered(SourceId source_id) const {
   return source_frame_id_map_.find(source_id) != source_frame_id_map_.end();
 }
--- a/geometry/geometry_state.h
+++ b/geometry/geometry_state.h
@@ -195,6 +195,11 @@ class GeometryState {
   /** @name          Sources and source-related data  */
   //@{
 
+  /** Returns all of the source ids in the scene graph. The order is guaranteed
+   to be stable and consistent. The first element is the SceneGraph-internal
+   source.  */
+  std::vector<SourceId> GetAllSourceIds() const;
+
   /** Implementation of SceneGraphInspector::SourceIsRegistered().  */
   bool SourceIsRegistered(SourceId source_id) const;
 
--- a/geometry/scene_graph.h
+++ b/geometry/scene_graph.h
@@ -143,17 +143,15 @@ class QueryObject;
  registered with %SceneGraph as producers (a.k.a. _geometry sources_). They
  do this by acquiring a SourceId (via SceneGraph::RegisterSource()). The
  SourceId serves as a unique handle through which the producer's identity is
- validated and its ownership of its registered geometry is maintained.
+ validated.
 
  _Registering Geometry_
 
  %SceneGraph cannot know what geometry _should_ be part of the shared world.
  Other systems are responsible for introducing geometry into the world. This
  process (defining geometry and informing %SceneGraph) is called
- _registering_ the geometry. The source that registers the geometry "owns" the
- geometry; the source's unique SourceId is required to perform any operations
- on the geometry registered with that SourceId. Geometry can be registered as
- _anchored_ or _dynamic_.
+ _registering_ the geometry. Geometry can be registered as  _anchored_ or
+ _dynamic_, and is always registered to (associated with) a SourceId.
 
  Dynamic geometry can move; more specifically, its kinematics (e.g., pose)
  depends on a system's Context. Particularly, a non-deformable dynamic geometry
@@ -396,19 +394,12 @@ class SceneGraph final : public systems::LeafSystem<T> {
      geometries can be registered via calls to the (experimental)
      RegisterDeformableGeometry() methods.
 
-   %SceneGraph has a concept of "ownership" that is separate from the C++
-   notion of ownership. In this case, %SceneGraph protects geometry and frames
-   registered by one source from being modified by another source. All methods
-   that change the world are associated with the SourceId of the geometry source
-   requesting the change. One source cannot "hang" geometry onto a frame (or
-   geometry) that belongs to another source. However, all sources have read
-   access to all geometries in the world. For example, queries will return
-   GeometryId values that span all sources and the properties of the associated
-   geometries can be queried by arbitrary sources.
-
-   That said, if one source _chooses_ to share its SourceId externally, then
-   arbitrary code can use that SourceId to modify the geometry resources that
-   are associated with that SourceId.
+   %SceneGraph has a concept of "ownership" that is separate from the C++ notion
+   of ownership. All methods that *change* the world require passing the
+   SourceId that was originally used to register that geometry. However,
+   read-only operations on geometries do not require the SourceId, e.g., queries
+   will return GeometryId values that span all sources and the properties of the
+   associated geometries can be queried by arbitrary sources.
 
    @note There are no Context-modifying variants for source or frame
    registration yet, as these methods modify the port semantics.  */
@@ -455,6 +446,8 @@ class SceneGraph final : public systems::LeafSystem<T> {
   FrameId RegisterFrame(SourceId source_id, FrameId parent_id,
                         const GeometryFrame& frame);
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument, instead using
+  // the source_id associated with the given `frame_id`.
   /** Registers a new rigid geometry G for this source. This hangs geometry G on
    a previously registered frame F (indicated by `frame_id`). The pose of the
    geometry is defined in a fixed pose relative to F (i.e., `X_FG`).
@@ -481,6 +474,8 @@ class SceneGraph final : public systems::LeafSystem<T> {
   GeometryId RegisterGeometry(SourceId source_id, FrameId frame_id,
                               std::unique_ptr<GeometryInstance> geometry);
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument, instead using
+  // the source_id associated with the given `frame_id`.
   /** systems::Context-modifying variant of RegisterGeometry(). Rather than
    modifying %SceneGraph's model, it modifies the copy of the model stored in
    the provided context.  */
@@ -488,6 +483,7 @@ class SceneGraph final : public systems::LeafSystem<T> {
                               FrameId frame_id,
                               std::unique_ptr<GeometryInstance> geometry) const;
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
   /** Registers a new _anchored_ geometry G for this source. This hangs geometry
    G from the world frame (W). Its pose is defined in that frame (i.e., `X_WG`).
    Returns the corresponding unique geometry id.
@@ -510,6 +506,7 @@ class SceneGraph final : public systems::LeafSystem<T> {
   GeometryId RegisterAnchoredGeometry(
       SourceId source_id, std::unique_ptr<GeometryInstance> geometry);
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
   // TODO(xuchenhan-tri): Consider allowing registering deformable geometries to
   // non-world frames.
   /** Registers a new deformable geometry G for this source. This registers
@@ -547,6 +544,7 @@ class SceneGraph final : public systems::LeafSystem<T> {
       SourceId source_id, FrameId frame_id,
       std::unique_ptr<GeometryInstance> geometry, double resolution_hint);
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
   /** systems::Context-modifying variant of RegisterDeformableGeometry(). Rather
    than modifying %SceneGraph's model, it modifies the copy of the model stored
    in the provided context.
@@ -555,6 +553,7 @@ class SceneGraph final : public systems::LeafSystem<T> {
       systems::Context<T>* context, SourceId source_id, FrameId frame_id,
       std::unique_ptr<GeometryInstance> geometry, double resolution_hint) const;
 
+  // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
   /** Changes the `shape` of the geometry indicated by the given `geometry_id`.
 
    The geometry is otherwise unchanged -- same geometry_id, same assigned roles,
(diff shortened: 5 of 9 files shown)


jwnimmer-tri:

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXB3z5779WEh0yA4Gjq:b-j8s5kg) (no related file):*

> Taking this as a one-off, merely leaves us in an incoherent state.

Please explain specifically what is incoherent (perhaps some documentation somewhere?) so that I can make it cohere with this API.
</blockquote></details>

I was able to fine a few docs inconsistencies, and pushed another update.

BTW I was also under the impression that [this slack thread](https://app.slack.com/client/T0JNB2DS4/C2WBPQDB7/thread/C2WBPQDB7-1681406418.741039) already retired this design discussion.  That is, the premise of removing the "password" concept is already okay, and the only question is finding more places in the code that I've forgotten to update.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXB5gSX8IIHZUbWfKGP:b-5ld993) (no related file):*

I was able to fine a few docs inconsistencies, and pushed another update.

BTW I was also under the impression that [this slack thread](https://app.slack.com/client/T0JNB2DS4/C2WBPQDB7/thread/C2WBPQDB7-1681406418.741039) already retired this design discussion.  That is, the premise of removing the "password" concept is already okay, and the only question is finding more places in the code that I've forgotten to update.
</blockquote></details>

Yeah. I saw the slack thread after my diatribe. My last comment on it is the more succinct version of what I wrote above. "Leaking source ids may be a viable stop gap, but I'd feel a lot happier if we did it with a more concrete end in mind."

I did a rough glance through the documentation (as you did) and find that the documentation sparingly hints at the contract. I've tagged a couple of locations below that are contradicted by leaking `SourceIds`.

The incoherency, if such exists, is the general justification of `SourceId` implied by the API design. This speaks to my point about the down-the-road plan. Simply leaking them and never thinking about it again doesn't seem like a good idea. Anyone who tries to infer intent will be foiled.

But for now, I'll stand by what I wrote in that slack conversation. If this unblocks you, we can make a couple more tweaks to the documentation (as flagged below) and we can move forward.

___
___
*[`geometry/scene_graph.h` line 399 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXB6Bb12ZQYwsbVchFV:-NXB6Bb12ZQYwsbVchFW:bghr0bd) ([raw file](https://github.com/RobotLocomotion/drake/blob/56f8c8e2e0ba58076f6264b6e6338334f1e09e35/geometry/scene_graph.h#L399)):*
> ```objc
>    %SceneGraph has a concept of "ownership" that is separate from the C++
>    notion of ownership. In this case, %SceneGraph protects geometry and frames
>    registered by one source from being modified by another source. All methods
> ```

nit: `SceneGraph` no longer "protects" anything if it broadcasts `SourceId`s. 



_Code quote:_
```objc
   notion of ownership. In this case, %SceneGraph protects geometry and frames
   registered by one source from being modified by another source. All methods
```
___
*[`geometry/scene_graph.h` line 402 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXB6UU8BUeykbuaKKsr:-NXB6UU956SaRdl-mqQt:b6lli4l) ([raw file](https://github.com/RobotLocomotion/drake/blob/56f8c8e2e0ba58076f6264b6e6338334f1e09e35/geometry/scene_graph.h#L402)):*
> ```objc
>    that change the world are associated with the SourceId of the geometry source
>    requesting the change. One source cannot "hang" geometry onto a frame (or
>    geometry) that belongs to another source. However, all sources have read
> ```

nit: "cannot....unless...."



_Suggestion:_
```objc
   requesting the change. One source cannot "hang" geometry onto a frame (or
   geometry) that belongs to another source, unless that second source provides
a  access to its source id. However, all sources have read
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXBMV-_2pSvCWRHqXiQ:b-4t31o3) (no related file):*
> Anyone who tries to infer intent will be foiled.

After this PR, the purpose of SourceId will be to partition registered geometry by ID to provide cohesive kinematic input ports -- all kinematics for a given source appear on the same input port, and the parent-child kinematic tree must homogeneously live within each single source.  If the docs don't make that clear, we can improve upon that here.  I don't see any lingering uncertainly about the intention beyond that?

___
___
*[`geometry/scene_graph.h` line 399 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXB6Bb12ZQYwsbVchFV:-NXBO9iC7yS6FLZQjojQ:b-896fig) ([raw file](https://github.com/RobotLocomotion/drake/blob/56f8c8e2e0ba58076f6264b6e6338334f1e09e35/geometry/scene_graph.h#L399)):*

nit: `SceneGraph` no longer "protects" anything if it broadcasts `SourceId`s. 
</blockquote></details>

Done?
___
*[`geometry/scene_graph.h` line 402 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXB6UU8BUeykbuaKKsr:-NXBOByr969CkUi7whbg:bxsjfw8) ([raw file](https://github.com/RobotLocomotion/drake/blob/56f8c8e2e0ba58076f6264b6e6338334f1e09e35/geometry/scene_graph.h#L402)):*

nit: "cannot....unless...."
</blockquote></details>

I think it's still true that the parent frame of newly added geometry must have the same source_id as the new geometry.  I haven't changed that in this PR.

SeanCurtis-TRI (commented):
:LGTM:

I still feel that we probably need to take another pass on `SceneGraph`'s documentation, but it can be deferred. Just in this last revision, I found an anachronism (adding a geometry to another geometry). So, clean up would be called for.

> Reviewed 8 of 8 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-:-NXBn1TR1oVeUpjlGryL:b-3klveh)* status: 3 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXBj7wiA4jsPHI1glxo:b-8w1tdz) (no related file):*

> Anyone who tries to infer intent will be foiled.

After this PR, the purpose of SourceId will be to partition registered geometry by ID to provide cohesive kinematic input ports -- all kinematics for a given source appear on the same input port, and the parent-child kinematic tree must homogeneously live within each single source.  If the docs don't make that clear, we can improve upon that here.  I don't see any lingering uncertainly about the intention beyond that?
</blockquote></details>

That's a port-centric view. And if source ids only mattered to that, it would be easy to summarize. Your use case of, someone *else* wants to change the visual color is not consistent with that. So, everything vis a vis *geometry* doesn't fit (as the geometries don't affect the port interface, only the frames).

Also, we're still stuck in a world in which `MbP` can throw (in limited cases) if someone were to modify proximity geometries in `SceneGraph`. So, even independent of the port interface question, there is still some brittleness.

___
___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXBjXz-6-weylNpwjaQ):*
btw I'm not entirely sure what "expect from" means in this context. Possible attachment ambiguity here?



_Code quote:_
```
expect from
```
___
*[`geometry/scene_graph.h` line 402 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXB6UU8BUeykbuaKKsr:-NXBkJtEBzfYYNMD5Ws4:b-gvlg1h) ([raw file](https://github.com/RobotLocomotion/drake/blob/56f8c8e2e0ba58076f6264b6e6338334f1e09e35/geometry/scene_graph.h#L402)):*
BTW This is a subtle issue of semantics. I believe we're reading the text in two different ways:

1. You're equating a "source" with its "source id".
2. For me "source" means some code that requested a source id. So, such code can hang a frame on a frame registered by another source, if it has access to that other source's id. The undocumented implication is that the newly hung frame would belong to the *other* source.

Obviously, this language suffers from ambiguity about what a "source" means.  Perhaps, the correct language would be more like:

> One source cannot "hang" its own geometry (or frame) onto a frame that belongs to another source. However, if the source has access to that other source's id, it can use that borrowed id to register the geometry, but the geometry will be owned by that other source.

However, I'm marking myself as satisfied. 
___
*[`geometry/scene_graph.h` line 155 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXBjrUT8feC6XEwVCbj:-NXBjrUT8feC6XEwVCbk:b-tq998e) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d0f467c18a373ee3b69507fb9bed611334432ab/geometry/scene_graph.h#L155)):*
nit: I'm not sure what went into this edit. The following:

>  the source's unique SourceId is required to perform any operations on the geometry registered with that SourceId.

Is still perfectly factual. Why delete it? It can probably be made more succinct.



_Code quote:_
```objc
 geometry; the source's unique SourceId is required to perform any operations
 on the geometry registered with that SourceId. Geometry can be registered as
```

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-:-NXBnMMK2wFxG3fkUa19:bcav8r0)* status: 3 unresolved discussions, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

jwnimmer-tri:
I'm going to remove -@ggould-tri for the moment.  I think its worth one(?) more pass to clean things up before calling it ready.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 2 unresolved discussions, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXBuopoBHrkP1VA1U3I:b-2obruj) (no related file):*

That's a port-centric view. And if source ids only mattered to that, it would be easy to summarize. Your use case of, someone *else* wants to change the visual color is not consistent with that. So, everything vis a vis *geometry* doesn't fit (as the geometries don't affect the port interface, only the frames).

Also, we're still stuck in a world in which `MbP` can throw (in limited cases) if someone were to modify proximity geometries in `SceneGraph`. So, even independent of the port interface question, there is still some brittleness.
</blockquote></details>

Ah, I see now.  You're referring to registration or mutation functions that take a `FrameId` or `GeometryId` as a unique key, but then _also_ require a `SourceId` as a passphrase, and we should remove the passphrase.

I've added the TODOs to make sure we agree on the planned future work.  In short -- keep `SourceId` in the frame registration functions (because the source kinematics needs to populate that frame in the input).  Otherwise, get rid of it as a function argument.

If it helps us align, I think I could probably add the new functions and deprecations in this same PR without too much trouble?

___
___
*[`geometry/scene_graph.h` line 155 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXBjrUT8feC6XEwVCbj:-NXCANP31BhXB-1QI595:b35v44k) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d0f467c18a373ee3b69507fb9bed611334432ab/geometry/scene_graph.h#L155)):*

nit: I'm not sure what went into this edit. The following:

>  the source's unique SourceId is required to perform any operations on the geometry registered with that SourceId.

Is still perfectly factual. Why delete it? It can probably be made more succinct.

</blockquote></details>

If we're moving away from this premise, I think it makes sense to cull the text?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXAl_bL7Te-wMiZ77y4:-NXCSB7ABLKn5_Ze06HJ:b-y7lwu9) (no related file):*

Ah, I see now.  You're referring to registration or mutation functions that take a `FrameId` or `GeometryId` as a unique key, but then _also_ require a `SourceId` as a passphrase, and we should remove the passphrase.

I've added the TODOs to make sure we agree on the planned future work.  In short -- keep `SourceId` in the frame registration functions (because the source kinematics needs to populate that frame in the input).  Otherwise, get rid of it as a function argument.

If it helps us align, I think I could probably add the new functions and deprecations in this same PR without too much trouble?
</blockquote></details>

I think we're on the same page (I just have one question below).

I suspect that there will be a fair amount of word-smithing in the documentation of the new APIs. Specifically, the overview documentation. As such, I'm content to let this go through with the understanding that we'll resolve these TODOs in a reasonable timeline.

___
___
*[`geometry/scene_graph.h` line 155 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXBjrUT8feC6XEwVCbj:-NXCQy4iBBbIyJnN7RY5:b-s4rfe9) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d0f467c18a373ee3b69507fb9bed611334432ab/geometry/scene_graph.h#L155)):*

If we're moving away from this premise, I think it makes sense to cull the text?
</blockquote></details>

Ah...I see. I erred on the side of "text should represent what is", you're going with "text might as well represent what will be soon". However, we're not doing that consistently. But, meh.
___
*[`geometry/scene_graph.h` line 486 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXCRUaA6RUgf4UYAz_-:-NXCRUaA6RUgf4UYAz_0:b-p2sn2u) ([raw file](https://github.com/RobotLocomotion/drake/blob/d8c3aa02440d0018912f048fe58e7bc88f3850de/geometry/scene_graph.h#L486)):*
> ```objc
>                               std::unique_ptr<GeometryInstance> geometry) const;
> 
>   // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
> ```

What's your plan vis a vis ownership? Have `SceneGraph` "own" all anchored geometries?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/scene_graph.h` line 486 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXCRUaA6RUgf4UYAz_-:-NXCTraI4QhzeopzOZhX:blyuk5b) ([raw file](https://github.com/RobotLocomotion/drake/blob/d8c3aa02440d0018912f048fe58e7bc88f3850de/geometry/scene_graph.h#L486)):*

What's your plan vis a vis ownership? Have `SceneGraph` "own" all anchored geometries?
</blockquote></details>

Yes.  I think it already does:

https://github.com/SeanCurtis-TRI/drake/blob/2d8aa72ac50ebcac8434d7cc33d6f32830f85877/geometry/geometry_state.cc#L1382

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: 1 unresolved discussion, needs at least two assigned reviewers (waiting on @jwnimmer-tri)

___
*[`geometry/scene_graph.h` line 486 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXCRUaA6RUgf4UYAz_-:-NXCdRpL9Xja60XCOE4a:b2rnbak) ([raw file](https://github.com/RobotLocomotion/drake/blob/d8c3aa02440d0018912f048fe58e7bc88f3850de/geometry/scene_graph.h#L486)):*

Yes.  I think it already does:

https://github.com/SeanCurtis-TRI/drake/blob/2d8aa72ac50ebcac8434d7cc33d6f32830f85877/geometry/geometry_state.cc#L1382
</blockquote></details>

That's not quite doing what you think. It is a trick to allow *anyone* to hang a geometry on the world frame. But the geometry still belongs to the original source. See [here](https://github.com/SeanCurtis-TRI/drake/blob/2d8aa72ac50ebcac8434d7cc33d6f32830f85877/geometry/geometry_state.cc#L794-L795).

Nonetheless, since we're getting rid of the passkey, it'll be fine. Anyone will be able to mutate (and remove) the geometry without a passkey, so the fact that `SG` owns it will be an invisible detail....with one caveat. You won't be able to tell who *added* it. If someone asks for the geometries from a source, it will never return the anchored geometries.

We'll have to document that.

rpoyner-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-:-NXGLQjRDq1zwhmtYPAo:bnfp4nl)* status: 1 unresolved discussion

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19525)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),SeanCurtis-TRI(platform)

___
*[`geometry/scene_graph.h` line 486 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19525#-NXCRUaA6RUgf4UYAz_-:-NXGOCpM9baAcQ3j4K98:b1rwad3) ([raw file](https://github.com/RobotLocomotion/drake/blob/d8c3aa02440d0018912f048fe58e7bc88f3850de/geometry/scene_graph.h#L486)):*

@SeanCurtis-TRI OK to resolve this discussion?
</blockquote></details>

Oops. Meant to mark myself as satisified.

(And, I assure you, the link is correct. Both anchored and non-anchored geometry registration pass through the same method.)

PR #19533, opened 2023-06-05
MultibodyPlant no longer assigned perception color

SeanCurtis-TRI:
Geometry with perception roles, if they have no definition of color, should defer to the `RenderEngine`'s logic and default values to pick up a color. MultibodyPlant shouldn't be playing a role in that regard.

In this case, MultibodyPlant is assuming that if the illustration properties have a color defined, the perception role should pick up the same. This is correct so far as our parsing cannot yet distinguish between illustration and perception roles (they're all just "visual").

This *can* affect an end user who never specified a color -- i.e., the default may change from MultibodyPlant's default color to their render engine or visualizer's default color.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19533)

### DIFF
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -715,10 +715,11 @@ geometry::GeometryId MultibodyPlant<T>::RegisterVisualGeometry(
   //  and illustration in favor of a protocol that allows definition.
   geometry::PerceptionProperties perception_props;
   perception_props.AddProperty("label", "id", RenderLabel(body.index()));
-  perception_props.AddProperty(
-      "phong", "diffuse",
-      properties.GetPropertyOrDefault(
-          "phong", "diffuse", Vector4<double>(0.9, 0.9, 0.9, 1.0)));
+  if (properties.HasProperty("phong", "diffuse")) {
+    perception_props.AddProperty(
+        "phong", "diffuse",
+        properties.GetProperty<geometry::Rgba>("phong", "diffuse"));
+  }
   if (properties.HasProperty("phong", "diffuse_map")) {
     perception_props.AddProperty(
         "phong", "diffuse_map",


SeanCurtis-TRI (commented):
+@jwnimmer-tri for both reviews.

Thoughts on release notes tag?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-:-NXBV64388SaEpiAusqm:byte0m8)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @jwnimmer-tri and @SeanCurtis-TRI)

jwnimmer-tri (commented):
I was wondering about unit testing, but I'll make this argument: since this PR is only _deleting_ code, we don't need a regression test.  There is already a regression test in place that confirms properties are _preserved_.  We don't need a test for "does not add default values" since it's obvious by inspection that the code isn't doing so.

So :lgtm: feature but let's check with +@ggould-tri for platform to validate the thought process.

---

> Thoughts on release notes tag?

This is +(release notes: fix).

It's not a breaking change because it's only a change to default values, which is explicitly allowed: _"Sometimes our software will choose a default value when nothing more specific is provided. Those defaults may change as we find better heuristics, without prior notice."_

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-:-NXBWroD1w2X7CM5gbi8:br68k3s)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform) (waiting on @ggould-tri)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-NXBXBEv0swzJVEroogr):*
typo?



_Suggestion:_
```
assigns
```

ggould-tri (commented):
I agree with `fix` -- while I'm not doctrinaire about the "defaults aren't breaking" rule, this is not likely to ruin anybody's day.
I agree about unit testing:  The existing tests cover it just fine for the complexity of code involved; if anything here merited testing it would be the magic number, and that's gone now.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-:-NXBkUoa0cx-Vf2Z_MWP:b-i72p7n)* status: 2 unresolved discussions

___
*[`multibody/plant/multibody_plant.cc` line 718 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-NXBjr1UBoeYHQ_Qioyj:-NXBjr1UBoeYHQ_Qioyk:b-rnbtvz) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d68af85b75b65ebd2eef77d57c77174e54bcd1/multibody/plant/multibody_plant.cc#L718)):*
> ```cpp
>   geometry::PerceptionProperties perception_props;
>   perception_props.AddProperty("label", "id", RenderLabel(body.index()));
>   if (properties.HasProperty("phong", "diffuse")) {
> ```

BTW:  I have a misgiving about the rule-of-three-with-small-differences here -- it's hard to read for the differences and invites nearly-correct copypasta.
However with a couple of quick attempts I couldn't improve the readability, so no defect.

jwnimmer-tri:
*[`multibody/plant/multibody_plant.cc` line 718 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19533#-NXBjr1UBoeYHQ_Qioyj:-NXBpKBh5XZNNR-x3VRJ:bvr8k2n) ([raw file](https://github.com/RobotLocomotion/drake/blob/64d68af85b75b65ebd2eef77d57c77174e54bcd1/multibody/plant/multibody_plant.cc#L718)):*

BTW:  I have a misgiving about the rule-of-three-with-small-differences here -- it's hard to read for the differences and invites nearly-correct copypasta.
However with a couple of quick attempts I couldn't improve the readability, so no defect.
</blockquote></details>

FYI I thought about this also.  I think the solution would be a "bulk copy all properties" member function from one GeometryProperties object onto another GeometryProperties object, or a `.cast<>` function with the same effect.  However, per the TODO forking the properties here in the first place is unwanted in the first place, I decided we don't need to polish the nearly-dead code.

PR #19465, opened 2023-05-23
Add BlockSparseLowerTriangularOrSymmetricMatrix

xuchenhan-tri:
Introduce data structure used for bespoke block sparse cholesky solver for SapSolver as well as linear solves in FEM.

Towards #19466.

This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19465)

PR #19545, opened 2023-06-06
Disallow partial Dirichlet BC for an FEM node

xuchenhan-tri:
Instead, we require the full state (3 DoFs) of a vertex is specified when a vertex is put under a Dirichlet boundary condition.

The flexibility of putting individual DoFs of a vertex under BC turns out to be rather useless in practice as our public APIs only allow users to describes BCs at the level of vertices.

This also paves the way for transitioning to the new data structure of storing FEM tangent matrix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545)

### DIFF
--- a/multibody/fem/dirichlet_boundary_condition.cc
+++ b/multibody/fem/dirichlet_boundary_condition.cc
@@ -7,8 +7,8 @@ namespace internal {
 
 template <typename T>
 void DirichletBoundaryCondition<T>::AddBoundaryCondition(
-    int dof_index, const Eigen::Ref<const Vector3<T>>& boundary_state) {
-  index_to_boundary_state_[dof_index] = boundary_state;
+    FemNodeIndex index, const NodeState<T>& boundary_state) {
+  index_to_boundary_state_[index] = boundary_state;
 }
 
 template <typename T>
@@ -16,16 +16,16 @@ void DirichletBoundaryCondition<T>::ApplyBoundaryConditionToState(
     FemState<T>* fem_state) const {
   DRAKE_DEMAND(fem_state != nullptr);
   if (index_to_boundary_state_.empty()) return;
-  VerifyIndexes(fem_state->num_dofs());
+  VerifyIndices(fem_state->num_nodes());
 
   // TODO(xuchenhan-tri): Consider avoiding this copy.
   VectorX<T> q = fem_state->GetPositions();
   VectorX<T> v = fem_state->GetVelocities();
   VectorX<T> a = fem_state->GetAccelerations();
-  for (const auto& [dof_index, boundary_state] : index_to_boundary_state_) {
-    q(dof_index) = boundary_state(0);
-    v(dof_index) = boundary_state(1);
-    a(dof_index) = boundary_state(2);
+  for (const auto& [node_index, boundary_state] : index_to_boundary_state_) {
+    q.template segment<3>(3 * node_index) = boundary_state.q;
+    v.template segment<3>(3 * node_index) = boundary_state.v;
+    a.template segment<3>(3 * node_index) = boundary_state.a;
   }
   fem_state->SetPositions(q);
   fem_state->SetVelocities(v);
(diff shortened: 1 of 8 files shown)


xuchenhan-tri (commented):
+@joemasterjohn for feature review please. This is a little bit of a yak shave to switch to the new data structure introduced in #19465 to store tangent matrices for FEM.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-:-NXHVF5kDx3UA0Y-7tzX:b469h6y)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @joemasterjohn and @xuchenhan-tri)

joemasterjohn (commented):
:lgtm: with one little request for more docs. +@sherm1 for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-:-NXLgTuq2lC9QyOCw64c:b-eoluh0)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

___
*[`multibody/fem/dirichlet_boundary_condition.h` line 52 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-NXLZGMu8PetDuTSvN6L:-NXLZGMu8PetDuTSvN6M:b6nwfa0) ([raw file](https://github.com/RobotLocomotion/drake/blob/576e8d2805ef44f28b58d8199c825d8cfce8dce4/multibody/fem/dirichlet_boundary_condition.h#L52)):*
> ```objc
> 
>   /* Sets the node with the given `index` to be subject to the prescribed
>    `boundary_state`. */
> ```

nit: Are there any pre-conditions on `boundary_state` that need to be stated here? As a noob user I'd want to understand a bit more about the semantics of these boundary conditions. For example from my limited understanding I'd expect specifying a Dirichlet BC would only require me specifying `q`.

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

___
*[`multibody/fem/dirichlet_boundary_condition.h` line 52 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-NXLZGMu8PetDuTSvN6L:-NXLrKGP0EJSyzYIKPxR:bdrllm7) ([raw file](https://github.com/RobotLocomotion/drake/blob/576e8d2805ef44f28b58d8199c825d8cfce8dce4/multibody/fem/dirichlet_boundary_condition.h#L52)):*

A Dirichlet BC specifies the solution of the system at the boundary. Here our system has more than positional states and we have to specify all of them. 

In practice, we currently only support zero velocity BC in DeformableModel (i.e. q = q0, v = 0, a = 0). Other types of BC are trickier to set up because we'd need to make sure the q,v,a's are consistent in time, but that's not the concern of this class.
</blockquote></details>

Ok thanks, you answered my second question about whether they had to be consistent in time before I could ask.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545)* status: LGTM missing from assignee sherm1(platform)

___
*[`multibody/fem/dirichlet_boundary_condition.h` line 52 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-NXLZGMu8PetDuTSvN6L:-NXM11lJ-DCSTx_0_PtI:bpse78o) ([raw file](https://github.com/RobotLocomotion/drake/blob/576e8d2805ef44f28b58d8199c825d8cfce8dce4/multibody/fem/dirichlet_boundary_condition.h#L52)):*

Ok thanks, you answered my second question about whether they had to be consistent in time before I could ask. 
</blockquote></details>

Knowledge of the time stepping scheme is needed to enforce consistency, so it can't be this class.

sherm1 (commented):
Platform :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-:-NXMkNx1A66uaJ1EwTWQ:b8qx53o)* status: 1 unresolved discussion

___
*[`multibody/fem/dirichlet_boundary_condition.h` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19545#-NXMimxc1NPuxlRBFB_T:-NXMimxc1NPuxlRBFB_U:b5hv1ff) ([raw file](https://github.com/RobotLocomotion/drake/blob/047177283ff866c28a3fce7651349f4ea8e030a4/multibody/fem/dirichlet_boundary_condition.h#L35)):*
> ```objc
> 
>  We only allow specifying BCs at the node level. That means that either all
>  degrees of freedom (DoFs) of a node is subject to BCs or none of the DoFs of
> ```

nit: "all ... are" and "none ... is" (change is -> are above)

Issue #19544, opened 2023-06-06
Unable to parse thin rod via SDFormat


jwnimmer-tri:
### What happened?

The SDFormat parser fails to parse a valid link inertia.

This demo program:
```python
from pydrake.all import *

def main():
    model = """
<?xml version="1.0"?>
<sdf version="1.7">
  <model name="semiacrobot">
    <link name="link1">
      <inertial>
        <pose>0 20 0 0 0 0</pose>
        <mass>1</mass>
        <inertia>
          <ixx>0.083</ixx>
          <iyy>0.083</iyy>
          <izz>0</izz>
          <ixy>0</ixy>
          <ixz>0</ixz>
          <iyz>0</iyz>
        </inertia>
      </inertial>
    </link>
  </model>
</sdf>
    """
    plant = MultibodyPlant(0)
    Parser(plant).AddModelsFromString(model, ".sdf")

main()
```

... produces this output ...

```
WARNING:drake:<literal-string>.sdf:5: warning: Spatial inertia fails SpatialInertia::IsPhysicallyValid().
 mass = 1.0
 Center of mass = [0.0  20.0  0.0]
 Inertia about point P, I_BP =
[400.083        0        0]
[      0    0.083        0]
[      0        0      400]
 Inertia about center of mass, I_BBcm =
[0.083      0      0]
[    0  0.083      0]
[    0      0      0]
 Principal moments of inertia about Bcm (center of mass) =
[0.0  0.083  0.08300000000002683]
```

... and substitutes a stand-in spherical inertia for the actual thin rod inertia.

The desired outcome is no warnings, and the actual inertia remains intact as given by the user (within a few epsilon, I suppose).

I am not sure whether the parser code is misusing the Multibody API (i.e., a parser bug), or whether the Multibody API has precision-matching problems, or something else.

### Version

v1.17.0

### What operating system are you using?

Ubuntu 22.04

### What installation option are you using?

pip install drake

jwnimmer-tri:
For the record, we discovered this while developing the user-facing test cases for #19210.

mitiguy:
The following code reproduced the error.
Note: Currently this code is in spatial_inertia_test.cc for a possibly pending PR.
Below is my first guess as to what/why this is happening.
```
  // Create a spatial inertia for a thin rod B about its center of mass Bcm.
  const double mass = 10.0, length = 5.0;
  const SpatialInertia<double> M_BBcm_B =
[...]

PR #19551, opened 2023-06-07
[WIP] Improve tolerance for inertia triangle inequality check.

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)

### DIFF
--- a/bindings/pydrake/multibody/tree_py.cc
+++ b/bindings/pydrake/multibody/tree_py.cc
@@ -1096,6 +1096,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
             &Class::CalcPrincipalMomentsAndAxesOfInertia,
             cls_doc.CalcPrincipalMomentsAndAxesOfInertia.doc)
         .def("CouldBePhysicallyValid", &Class::CouldBePhysicallyValid,
+            py::arg("mass_dcm_squared") = 0.0,
             cls_doc.CouldBePhysicallyValid.doc)
         .def("ReExpress", &Class::ReExpress, py::arg("R_AE"),
             cls_doc.ReExpress.doc)
--- a/multibody/tree/rotational_inertia.h
+++ b/multibody/tree/rotational_inertia.h
@@ -587,13 +587,14 @@ class RotationalInertia {
   /// requires a rotational inertia to satisfy the above checks *after* `this`
   /// is shifted to the center of mass, i.e., the sufficient condition requires
   /// calling CouldBePhysicallyValid() when the about-point is Bcm (the body's
-  /// center of mass).  Note: this class does not know its about-point or its
-  /// center of mass location.
+  /// center of mass). Note: this class does not know its about-point or its
+  /// center of mass location. The parameter @p mass_dcm_squared is intended to
+  /// indirectly inform this function about mass and shift to center of mass.
   /// @return `true` for a plausible rotational inertia passing the above
-  ///          necessary but insufficient checks and `false` otherwise.
-  /// @throws std::exception if principal moments of inertia cannot be
-  ///         calculated (eigenvalue solver) or if scalar type T cannot be
-  ///         converted to a double.
+  /// necessary but insufficient checks and `false` otherwise.
+  /// @throws std::exception if principal moments of inertia cannot be computed
+  /// (eigenvalue solver), or if scalar type T cannot be converted to a double,
+  /// or is mass_dcm_squared < 0.
   boolean<T> CouldBePhysicallyValid(const T mass_dcm_squared = 0.0) const {
     DRAKE_ASSERT(mass_dcm_squared >= 0.0);
 
--- a/multibody/tree/rotational_inertia.h
+++ b/multibody/tree/rotational_inertia.h
@@ -566,16 +566,22 @@ class RotationalInertia {
     return std::pair(Imoment, R_EP);
   }
 
-  /// Performs several necessary checks to verify whether `this` rotational
-  /// inertia *could* be physically valid, including:
+  /// Performs several checks to verify whether `this` rotational inertia
+  /// *could* be physically valid, including:
   ///
   /// - No NaN moments or products of inertia.
   /// - Ixx, Iyy, Izz and principal moments are all non-negative.
   /// - Ixx, Iyy  Izz and principal moments satisfy the triangle inequality:
-  ///   - `Ixx + Iyy >= Izz`
-  ///   - `Ixx + Izz >= Iyy`
-  ///   - `Iyy + Izz >= Ixx`
+  ///   - `Ixx + Iyy + epsilon >= Izz`
+  ///   - `Ixx + Izz + epsilon >= Iyy`
+  ///   - `Iyy + Izz + epsilon >= Ixx`
   ///
+  /// @param[in] mass_dcm_squared non-negative real value that provides a
+  /// minimum guide for epsilon for the aforementioned triangle inequality.
+  /// If the calling function has information about the mass associated with
+  /// `this` and knows that `this` rotational inertia was shifted from an about-
+  /// point P to Bcm (center of mass), the calling function should pass
+  /// mass_dcm_squared = mass * p_PBcm.squaredNorm().
   /// @warning These checks are necessary (but NOT sufficient) conditions for a
   /// rotational inertia to be physically valid.  The sufficient condition
   /// requires a rotational inertia to satisfy the above checks *after* `this`
@@ -583,25 +589,29 @@ class RotationalInertia {
   /// calling CouldBePhysicallyValid() when the about-point is Bcm (the body's
   /// center of mass).  Note: this class does not know its about-point or its
   /// center of mass location.
-  ///
   /// @return `true` for a plausible rotational inertia passing the above
   ///          necessary but insufficient checks and `false` otherwise.
   /// @throws std::exception if principal moments of inertia cannot be
   ///         calculated (eigenvalue solver) or if scalar type T cannot be
   ///         converted to a double.
-  boolean<T> CouldBePhysicallyValid() const {
-    // To check the validity of rotational inertia use an epsilon value that is
+  boolean<T> CouldBePhysicallyValid(const T mass_dcm_squared = 0.0) const {
+    DRAKE_ASSERT(mass_dcm_squared >= 0.0);
+
+    // To check the validity of rotational inertia, use an epsilon value that is
     // a number related to machine precision multiplied by the largest possible
-    // element that can appear in a valid `this` rotational inertia.  Note: The
-    // largest product of inertia is at most half the largest moment of inertia.
+    // element that can appear in a valid `this` rotational inertia. Also use
+    // mass_dcm_squared to account for the loss of significant digits that can
+    // occur when shifting rotational inertia towards the center of mass.
     using std::max;
-    const double precision = 10 * std::numeric_limits<double>::epsilon();
+    const double precision = 16 * std::numeric_limits<double>::epsilon();
     const T max_possible_inertia_moment = CalcMaximumPossibleMomentOfInertia();
+    const T max_inertia_for_epsilon =
+        max(mass_dcm_squared, max_possible_inertia_moment);
 
     // In order to avoid false negatives for inertias close to zero we use, in
     // addition to a relative tolerance of "precision", an absolute tolerance
     // equal to "precision" as well.
-    const T epsilon = precision * max(1.0, max_possible_inertia_moment);
+    const T epsilon = precision * max(1.0, max_inertia_for_epsilon);
 
     // Calculate principal moments of inertia p and then test these principal
     // moments to be mostly non-negative and also satisfy triangle inequality.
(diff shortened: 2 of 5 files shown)


mitiguy (commented):
PR #19551 helps resolves issue #19544.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXIeP1tD2NfGyScEBp3:b3u669f)* status: needs platform reviewer assigned, needs at least two assigned reviewers (waiting on @mitiguy)

mitiguy (commented):
Feature review @sherm1 
Notes: 
1. The comments in spatial_inertia_test.cc may be repetitious.  Let me know.
2. Change Python bindings due to new optional parameter to RotationalInertia::CouldBePhysicallyValid().
 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXIgqQa8w70ysswOuAi:b-amdf1x)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

mitiguy (commented):
@sherm1  Reduced repetitious comments in spatial_inertia_test.cc and added Python binding for new optional parameter.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXMltcJ4-mwoFqylI90:b-6ctvwt)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

sherm1 (commented):
Feature review +@sherm1

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXNG2vCAMQX9H38dPZv:bet2hx0)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

sherm1 (commented):
Looks great, thanks! One discussion point -- I think we should consider an absolute cutoff for the error test so that we don't accept something that has bad downstream consequences. PTAL at my comment below.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXQvdco-ZOwLJGr8Q51:b-u55jch)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/rotational_inertia.h` line 591 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQsH870Um-DKl4XNZ8:-NXQsH870Um-DKl4XNZ9:b7jhuc7) ([raw file](https://github.com/RobotLocomotion/drake/blob/b84ac3b9d5ea1ea40e5d29534d3631bc3d15e78b/multibody/tree/rotational_inertia.h#L591)):*
> ```objc
>   /// calling CouldBePhysicallyValid() when the about-point is Bcm (the body's
>   /// center of mass).  Note: this class does not know its about-point or its
>   /// center of mass location.
> ```

BTW maybe note here that the mass_dcm_squared parameter is intended to provide that information (effectively)
___
*[`multibody/tree/test/spatial_inertia_test.cc` line 953 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQuGew3wcVbOFkS9gP:-NXQuGew3wcVbOFkS9gQ:b2vb74e) ([raw file](https://github.com/RobotLocomotion/drake/blob/b84ac3b9d5ea1ea40e5d29534d3631bc3d15e78b/multibody/tree/test/spatial_inertia_test.cc#L953)):*
> ```cpp
>   // I_BBcm_B calculates principal moments of inertia of
>   // [0.0  20.833333730697632  20.83333373069763] which violates the triangle
>   // inequality by ≈ 1.19E-06. Note the ratio 1.19E-06 / 1.14E-12 ≈ 10³ = 1000.
> ```

BTW Hmmm. This seems like such a huge violation I'm thinking we should reject it because of the potential downstream consequences. WDYT? Should we set a maximum limit on the error we'll tolerate? Maybe insist on 1e-10 or better? Your test above already has a 10 kg part shifted 300 meters!

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXQwc928kBqCU2oXRvJ:b-t9p0sp) (no related file):*
Naively, my expectation was that `MakeFromCentralInertia` could be improved to _maintain_ the triangle inequality (strictly, without any tolerancing), so that anyone who was _relying_ on the triangle inequality could continue to do so.  Here now, we have the possible hazard where downstream code that was assuming that the inequality held with no tolerance would be broken by this change.  Could you help me clarify my thinking around that, or point out where I went wrong?

___

mitiguy (commented):
Hmmm -- still pondering whether we should implement an absolute cutoff.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-:-NXRYxew8iWFz1QzpVtH:bc95hie)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXRJD6QAzoHwYCe1gKR:b-ebziml) (no related file):*

Naively, my expectation was that `MakeFromCentralInertia` could be improved to _maintain_ the triangle inequality (strictly, without any tolerancing), so that anyone who was _relying_ on the triangle inequality could continue to do so.  Here now, we have the possible hazard where downstream code that was assuming that the inequality held with no tolerance would be broken by this change.  Could you help me clarify my thinking around that, or point out where I went wrong?
</blockquote></details>

I think there is at least one more paths that can cause similar issues.  
I added another test in spatial_inertia_test.cc that may clarify similar issues with this constructor:
SpatialInertia(const T& mass, const Vector3<T>& p_PScm_E, const UnitInertia<T>& G_SP_E,
      const bool skip_validity_check = false).

___
___
*[`multibody/tree/rotational_inertia.h` line 591 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQsH870Um-DKl4XNZ8:-NXRXHZeBlmxdesoRppJ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b84ac3b9d5ea1ea40e5d29534d3631bc3d15e78b/multibody/tree/rotational_inertia.h#L591)):*

BTW maybe note here that the mass_dcm_squared parameter is intended to provide that information (effectively)
</blockquote></details>

Done

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXRZ_ySEEMQ8qtIs-0L:b5h3449) (no related file):*

I think there is at least one more paths that can cause similar issues.  
I added another test in spatial_inertia_test.cc that may clarify similar issues with this constructor:
SpatialInertia(const T& mass, const Vector3<T>& p_PScm_E, const UnitInertia<T>& G_SP_E,
      const bool skip_validity_check = false).
</blockquote></details>

That's not responsive to my question though.  Having to fix a few functions to correct the numbers isn't any worse than fixing just that one place.

Perhaps an analogy would help: We have other math functions that take a unit vector as an argument.  If the argument is too far from the unit norm, we will throw.  However, if the argument is within some tolerance of the norm, then we'll normalize it and proceed.  My question here is for our inertia constructors and factory methods, so long as the arguments are within this small tolerance of the inequality criterion, why don't we automatically normalize the numbers and then proceed?  Why do we leave the invalid number intact and hand a newly-constructed object back to the user that violates its invariants?

___

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXRa4g899TFPtJuSh09:b-5u9vfd) (no related file):*

That's not responsive to my question though.  Having to fix a few functions to correct the numbers isn't any worse than fixing just that one place.

Perhaps an analogy would help: We have other math functions that take a unit vector as an argument.  If the argument is too far from the unit norm, we will throw.  However, if the argument is within some tolerance of the norm, then we'll normalize it and proceed.  My question here is for our inertia constructors and factory methods, so long as the arguments are within this small tolerance of the inequality criterion, why don't we automatically normalize the numbers and then proceed?  Why do we leave the invalid number intact and hand a newly-constructed object back to the user that violates its invariants?
</blockquote></details>

I was also hoping for some kind of symmetry-preserving transformation. Maybe that could be done for some special cases? I'm not sure how though. What's happening here is that the inertia was provided as a beautiful central inertia with pattern (a, a, 0). We store that as an inertia about the body origin, requiring a large shift and producing a general inertia with no obvious pattern. Then when asked to do the check, we shift to CoM. Now the original pattern is damaged by roundoff caused by the two shifts. WDYT @mitiguy?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXRgbdc-AiJMmuw0tWc:bpltqnm) (no related file):*

I was also hoping for some kind of symmetry-preserving transformation. Maybe that could be done for some special cases? I'm not sure how though. What's happening here is that the inertia was provided as a beautiful central inertia with pattern (a, a, 0). We store that as an inertia about the body origin, requiring a large shift and producing a general inertia with no obvious pattern. Then when asked to do the check, we shift to CoM. Now the original pattern is damaged by roundoff caused by the two shifts. WDYT @mitiguy?
</blockquote></details>

I was imagining something like solving for the adjustments `d,e,f` that minimize `norm([d,e,f])` subject to "`[ixx+d,iyy+e,izz+f]` satisfies the triangle inequality" -- projecting the barely-invalid moments to the closest nearby moments.  (That would be a slightly more refined version of just "whatever the epsilon violation was, add it to one of the moments".)

In other words, maybe it would be OK if the second shift (back to the CoM) ended up with something within some epsilon of `[a, a, 0]` instead of exactly `[a, a, 0]`, as a tradeoff for the body-origin inertia being strictly physically valid.

___

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19551)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19551#-NXQwc928kBqCU2oXRvI:-NXS7lDo27-wr-g119GD:b-572yjm) (no related file):*

I was imagining something like solving for the adjustments `d,e,f` that minimize `norm([d,e,f])` subject to "`[ixx+d,iyy+e,izz+f]` satisfies the triangle inequality" -- projecting the barely-invalid moments to the closest nearby moments.  (That would be a slightly more refined version of just "whatever the epsilon violation was, add it to one of the moments".)

In other words, maybe it would be OK if the second shift (back to the CoM) ended up with something within some epsilon of `[a, a, 0]` instead of exactly `[a, a, 0]`, as a tradeoff for the body-origin inertia being strictly physically valid.
</blockquote></details>

Paul and I discussed this and decided we should abandon this approach. It has a fatal flaw: imagine someone gives us a slightly bad central inertia, say (1e-8, 0, 0), but claims the COM is far from the body origin, say (10, 20, 100). The current algorithm will use the shift vector to loosen the tolerance check so will then declare the inertia valid. That's not good!

Instead, here is a better approach: if a user provides a central inertia, do the validity check immediately on the central inertia rather than by first shifting to the body origin and then shifting back to COM to do the check. While that may not solve all possible problems, it should be a very satisfactory solution to inertias that come from sdf and urdf (and most other places) where central inertias are the coin of the realm.

___

PR #19526, opened 2023-06-04
fix copy slicing in CSpaceSeparatingPlane

AlexandreAmice:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19526)

PR #19552, opened 2023-06-07
Create CspaceFreeBox class constructor.

hongkai-dai:
Factor out CspaceFreePolytopeBase for the common code in both CspaceFreePolytope and CspaceFreeBox.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -19,8 +19,10 @@ drake_cc_package_library(
         ":c_iris_collision_geometry",
         ":c_iris_separating_plane",
         ":convex_set",
+        ":cspace_free_box",
         ":cspace_free_internal",
         ":cspace_free_polytope",
+        ":cspace_free_polytope_base",
         ":cspace_free_structs",
         ":cspace_separating_plane",
         ":graph_of_convex_sets",
@@ -105,11 +107,29 @@ drake_cc_library(
 )
 
 drake_cc_library(
+    name = "cspace_free_box",
+    srcs = ["cspace_free_box.cc"],
+    hdrs = ["cspace_free_box.h"],
+    deps = [
+        ":cspace_free_polytope_base",
+    ],
+)
+
+drake_cc_library(
     name = "cspace_free_polytope",
     srcs = [
         "cspace_free_polytope.cc",
     ],
     hdrs = ["cspace_free_polytope.h"],
+    deps = [
+        ":cspace_free_polytope_base",
+    ],
+)
+
+drake_cc_library(
+    name = "cspace_free_polytope_base",
+    srcs = ["cspace_free_polytope_base.cc"],
+    hdrs = ["cspace_free_polytope_base.h"],
     deps = [
         "//common/symbolic:monomial_util",
         "//geometry/optimization:c_iris_collision_geometry",
@@ -216,6 +236,29 @@ drake_cc_googletest(
 )
 
 drake_cc_googletest(
+    name = "cspace_free_polytope_base_test",
+    deps = [
+        ":c_iris_test_utilities",
+        ":cspace_free_polytope_base",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:symbolic_test_util",
+        "//geometry/optimization:cspace_free_internal",
+    ],
+)
+
+drake_cc_googletest(
+    name = "cspace_free_box_test",
+    deps = [
+        ":c_iris_test_utilities",
+        ":cspace_free_box",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+        "//common/test_utilities:symbolic_test_util",
+        "//geometry/optimization:cspace_free_internal",
+    ],
+)
+
+drake_cc_googletest(
     name = "cspace_free_polytope_test",
     deps = [
         ":c_iris_test_utilities",
@@ -265,6 +308,7 @@ drake_cc_library(
         "test/convex.obj",
     ],
     deps = [
+        ":cspace_free_box",
         ":cspace_free_polytope",
         "//geometry:scene_graph",
         "//multibody/plant",
--- a/geometry/optimization/cspace_free_box.cc
+++ b/geometry/optimization/cspace_free_box.cc
@@ -1,0 +1,96 @@
+#include "drake/geometry/optimization/cspace_free_box.h"
+
+#include <map>
+#include <memory>
+
+#include "drake/common/fmt_eigen.h"
+#include "drake/geometry/optimization/cspace_free_internal.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+CspaceFreeBox::CspaceFreeBox(const multibody::MultibodyPlant<double>* plant,
+                             const geometry::SceneGraph<double>* scene_graph,
+                             SeparatingPlaneOrder plane_order,
+                             const Options& options)
+    : CspaceFreePolytopeBase(plant, scene_graph, plane_order, options) {}
+
+CspaceFreeBox::~CspaceFreeBox() {}
+
+CspaceFreeBox::SeparatingPlaneLagrangians
+CspaceFreeBox::SeparatingPlaneLagrangians::GetSolution(
+    const solvers::MathematicalProgramResult& result) const {
+  CspaceFreeBox::SeparatingPlaneLagrangians ret(this->s_box_lower_.rows());
+  for (int i = 0; i < this->s_box_lower_.rows(); ++i) {
+    ret.s_box_lower_(i) = result.GetSolution(this->s_box_lower_(i));
+    ret.s_box_upper_(i) = result.GetSolution(this->s_box_upper_(i));
+  }
+  return ret;
+}
+
+void CspaceFreeBox::ComputeSBox(
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+    Eigen::VectorXd* s_box_lower, Eigen::VectorXd* s_box_upper,
+    Eigen::VectorXd* q_star) const {
+  if ((q_box_lower.array() > q_box_upper.array()).any()) {
+    throw std::runtime_error(
+        fmt::format("CspaceFreeBox: q_box_lower={} has some entries larger "
+                    "than q_box_upper={}.",
+                    fmt_eigen(q_box_lower.transpose()),
+                    fmt_eigen(q_box_upper.transpose())));
+  }
+  const auto& plant = this->rational_forward_kin().plant();
+  const Eigen::VectorXd q_position_lower = plant.GetPositionLowerLimits();
+  const Eigen::VectorXd q_position_upper = plant.GetPositionUpperLimits();
+  if ((q_box_lower.array() > q_position_upper.array()).any()) {
+    throw std::runtime_error(fmt::format(
+        "CspaceFreeBox: q_box_lower={} has some entries larger the the robot "
+        "position upper limit={}.",
+        fmt_eigen(q_box_lower.transpose()),
+        fmt_eigen(q_position_upper.transpose())));
+  }
+  if ((q_box_upper.array() < q_position_lower.array()).any()) {
+    throw std::runtime_error(fmt::format(
+        "CspaceFreeBox: q_box_upper={} has some entries smaller than the robot "
+        "position lower limit={}.",
+        fmt_eigen(q_box_upper.transpose()),
+        fmt_eigen(q_position_lower.transpose())));
+  }
+  const Eigen::VectorXd q_lower =
+      q_box_lower.array().max(q_position_lower.array()).matrix();
+  const Eigen::VectorXd q_upper =
+      q_box_upper.array().min(q_position_upper.array()).matrix();
+  *q_star = 0.5 * (q_lower + q_upper);
+  *s_box_lower = this->rational_forward_kin().ComputeSValue(q_lower, *q_star);
+  *s_box_upper = this->rational_forward_kin().ComputeSValue(q_upper, *q_star);
+}
+
+void CspaceFreeBox::GeneratePolynomialsToCertify(
+    const Eigen::Ref<const Eigen::VectorXd>& s_box_lower,
+    const Eigen::Ref<const Eigen::VectorXd>& s_box_upper,
+    const Eigen::Ref<const Eigen::VectorXd>& q_star,
+    const IgnoredCollisionPairs& ignored_collision_pairs,
+    PolynomialsToCertify* certify_polynomials) const {
+  this->CalcSBoundsPolynomial(s_box_lower, s_box_upper,
+                              &(certify_polynomials->s_minus_s_box_lower),
+                              &(certify_polynomials->s_box_upper_minus_s));
+
+  std::map<int, const CSpaceSeparatingPlane<symbolic::Variable>*>
+      separating_planes_map;
+  for (int i = 0; i < static_cast<int>(separating_planes().size()); ++i) {
+    const auto& plane = separating_planes()[i];
+    if (ignored_collision_pairs.count(SortedPair<geometry::GeometryId>(
+            plane.positive_side_geometry->id(),
+            plane.negative_side_geometry->id())) == 0) {
+      separating_planes_map.emplace(i, &plane);
+    }
+  }
+
+  internal::GenerateRationals(separating_planes_map, y_slack(), q_star,
+                              rational_forward_kin(),
+                              &(certify_polynomials->plane_geometries));
+}
+}  // namespace optimization
+}  // namespace geometry
+}  // namespace drake
--- a/geometry/optimization/cspace_free_box.h
+++ b/geometry/optimization/cspace_free_box.h
@@ -1,0 +1,138 @@
+#pragma once
+
+#include <vector>
+
+#include "drake/geometry/optimization/cspace_free_polytope_base.h"
+#include "drake/geometry/optimization/cspace_free_structs.h"
+#include "drake/solvers/mathematical_program_result.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+/**
+ This class tries to find large axis-aligned bounding boxes in the configuration
+ space, such that all configurations in the boxes are collision free.
+ Note that we don't guarantee to find the largest box.
+ */
+// CspaceFreeBox "is a" CspaceFreePolytopeBase because it can do anything inside
+// CspaceFreePolytopeBase. We factor out the common code in CspaceFreeBox and
+// CspaceFreePolytope to CspaceFreePolytopeBase, and also rely on the access
+// control (public/protected/private) in CspaceFreePolytopeBase.
+class CspaceFreeBox : public CspaceFreePolytopeBase {
+ public:
+  DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(CspaceFreeBox)
+
+  using CspaceFreePolytopeBase::IgnoredCollisionPairs;
+
+  ~CspaceFreeBox() override;
+
+  using CspaceFreePolytopeBase::Options;
+
+  /**
+   When searching for the separating plane, we want to certify that the
+   numerator of a rational is non-negative in the C-space box q_box_lower <= q
+   <= q_box_upper (or equivalently s_box_lower <= s <= s_box_upper). Hence for
+   each of the rational we will introduce Lagrangian multipliers for the
+   polytopic constraint s - s_box_lower >= 0, s_box_upper - s >= 0.
+   */
+  class SeparatingPlaneLagrangians {
+   public:
+    explicit SeparatingPlaneLagrangians(int s_size)
+        : s_box_lower_(s_size), s_box_upper_(s_size) {}
+
+    /** Substitutes the decision variables in each Lagrangians with its value in
+     * result, returns the substitution result.
+     */
+    [[nodiscard]] SeparatingPlaneLagrangians GetSolution(
+        const solvers::MathematicalProgramResult& result) const;
+
+    /// The Lagrangians for s - s_box_lower >= 0.
+    const VectorX<symbolic::Polynomial>& s_box_lower() const {
+      return s_box_lower_;
+    }
+
+    /// The Lagrangians for s - s_box_lower >= 0.
+    VectorX<symbolic::Polynomial>& mutable_s_box_lower() {
+      return s_box_lower_;
+    }
+
+    /// The Lagrangians for s_box_upper - s >= 0.
+    const VectorX<symbolic::Polynomial>& s_box_upper() const {
+      return s_box_upper_;
+    }
+
+    /// The Lagrangians for s_box_upper - s >= 0.
+    VectorX<symbolic::Polynomial>& mutable_s_box_upper() {
+      return s_box_upper_;
+    }
+
+   private:
+    // The Lagrangians for s - s_box_lower >= 0.
+    VectorX<symbolic::Polynomial> s_box_lower_;
+    // The Lagrangians for s_box_upper - s >= 0.
+    VectorX<symbolic::Polynomial> s_box_upper_;
+  };
+
+  /**
+   @param plant The plant for which we compute the C-space free boxes. It
+   must outlive this CspaceFreeBox object.
+   @param scene_graph The scene graph that has been connected with `plant`. It
+   must outlive this CspaceFreeBox object.
+   @param plane_order The order of the polynomials in the plane to separate a
+   pair of collision geometries.
+
+   @note CspaceFreeBox knows nothing about contexts. The plant and
+   scene_graph must be fully configured before instantiating this class.
+   */
+  CspaceFreeBox(const multibody::MultibodyPlant<double>* plant,
+                const geometry::SceneGraph<double>* scene_graph,
+                SeparatingPlaneOrder plane_order,
+                const Options& options = Options{});
+
+ private:
+  // Forward declare the tester class that will test the private members.
+  friend class CspaceFreeBoxTester;
+
+  /*
+   Computes the range of s from the box q_box_lower <= q <= q_box_upper. We also
+   set q_star = 0.5(q_box_lower + q_box_upper).
+   If q_box_lower is smaller than the robot position lower limit (or q_box_upper
+   is larger than the robot position upper limit), then we clamp q_box_lower (or
+   q_box_upper) within the robot position limits.
+   @throws error if any q_box_lower is larger than q_box_upper or the robot
+   position upper limit; similarly throws an error if any q_box_upper is smaller
+   than the robot position lower limit.
+   */
+  void ComputeSBox(const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+                   const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+                   Eigen::VectorXd* s_box_lower, Eigen::VectorXd* s_box_upper,
+                   Eigen::VectorXd* q_star) const;
+
+  /*
+   This class contains the polynomials that we wish to certify are non-negative.
+   */
+  struct PolynomialsToCertify {
+    // We have the invariant plane_geometries_[i].plane_index == i.
+    std::vector<PlaneSeparatesGeometries> plane_geometries;
+    VectorX<symbolic::Polynomial> s_minus_s_box_lower;
+    VectorX<symbolic::Polynomial> s_box_upper_minus_s;
+  };
+
+  // Generates the polynomials used for certifying the box s_box_lower <= s <=
+  // s_box_upper is collision free.
+  // @note The box [s_box_lower, s_box_upper] is already inside the
+  // tangent-configuration space box computed from the robot position
+  // lower/upper limits.
+  // TODO(hongkai.dai): after we finish implementing this class, consider to
+  // change the input argument to q_box_lower and q_box_upper, if ComputeSBox is
+  // only used with this function.
+  void GeneratePolynomialsToCertify(
+      const Eigen::Ref<const Eigen::VectorXd>& s_box_lower,
+      const Eigen::Ref<const Eigen::VectorXd>& s_box_upper,
+      const Eigen::Ref<const Eigen::VectorXd>& q_star,
+      const IgnoredCollisionPairs& ignored_collision_pairs,
+      PolynomialsToCertify* certify_polynomials) const;
+};
+}  // namespace optimization
+}  // namespace geometry
+}  // namespace drake
--- a/geometry/optimization/cspace_free_internal.cc
+++ b/geometry/optimization/cspace_free_internal.cc
@@ -156,8 +156,8 @@ int GenerateCollisionPairs(
 }
 
 void GenerateRationals(
-    const std::vector<std::unique_ptr<
-        CSpaceSeparatingPlane<symbolic::Variable>>>& separating_planes,
+    const std::map<int, const CSpaceSeparatingPlane<symbolic::Variable>*>&
+        separating_planes,
     const Vector3<symbolic::Variable>& y_slack,
     const Eigen::Ref<const Eigen::VectorXd>& q_star,
     const multibody::RationalForwardKinematics& rational_forward_kin,
@@ -170,11 +170,9 @@ void GenerateRationals(
       multibody::RationalForwardKinematics::Pose<symbolic::Polynomial>,
       BodyPairHash>
       body_pair_to_X_AB_multilinear;
-  for (int plane_index = 0;
-       plane_index < static_cast<int>(separating_planes.size());
-       ++plane_index) {
-    DRAKE_ASSERT(separating_planes.at(plane_index) != nullptr);
-    const auto& separating_plane = *(separating_planes.at(plane_index));
+  for (const auto& [plane_index, plane_ptr] : separating_planes) {
+    DRAKE_ASSERT(plane_ptr != nullptr);
+    const auto& separating_plane = *plane_ptr;
     // Compute X_AB for both side of the geometries.
     std::vector<symbolic::RationalFunction> positive_side_rationals;
     std::vector<symbolic::RationalFunction> negative_side_rationals;
@@ -262,8 +260,24 @@ void GenerateRationals(
       plane_geometries->emplace_back(
           positive_side_rationals, negative_side_rationals_with_y, plane_index);
     }
-    DRAKE_DEMAND(plane_geometries->at(plane_index).plane_index == plane_index);
+    DRAKE_DEMAND(plane_geometries->back().plane_index == plane_index);
   }
+}
+
+void GenerateRationals(
+    const std::vector<std::unique_ptr<
+        CSpaceSeparatingPlane<symbolic::Variable>>>& separating_planes,
+    const Vector3<symbolic::Variable>& y_slack,
+    const Eigen::Ref<const Eigen::VectorXd>& q_star,
+    const multibody::RationalForwardKinematics& rational_forward_kin,
+    std::vector<PlaneSeparatesGeometries>* plane_geometries) {
+  std::map<int, const CSpaceSeparatingPlane<symbolic::Variable>*>
+      separating_planes_map;
+  for (int i = 0; i < static_cast<int>(separating_planes.size()); ++i) {
+    separating_planes_map.emplace(i, separating_planes.at(i).get());
+  }
+  GenerateRationals(separating_planes_map, y_slack, q_star,
+                    rational_forward_kin, plane_geometries);
 }
 
 void SolveSeparationCertificateProgramBase(
--- a/geometry/optimization/cspace_free_internal.h
+++ b/geometry/optimization/cspace_free_internal.h
@@ -41,6 +41,18 @@ void GenerateRationals(
     std::vector<PlaneSeparatesGeometries>* plane_geometries);
 
 /*
+ Overloads GenerateRationals.
+ Use separating_planes as the map of the plane_index to the separating plane.
+ */
+void GenerateRationals(
+    const std::map<int, const CSpaceSeparatingPlane<symbolic::Variable>*>&
+        separating_planes,
+    const Vector3<symbolic::Variable>& y_slack,
+    const Eigen::Ref<const Eigen::VectorXd>& q_star,
+    const multibody::RationalForwardKinematics& rational_forward_kin,
+    std::vector<PlaneSeparatesGeometries>* plane_geometries);
+
+/*
  Returns the number of y_slack variables in `rational`.
  Not all y_slack necessarily appear in `rational`.
  */
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -17,78 +17,6 @@ namespace geometry {
 namespace optimization {
 namespace {
 const double kInf = std::numeric_limits<double>::infinity();
-/*
- The monomials in the numerator of body point position has the form ∏ᵢ pow(tᵢ,
- dᵢ) * ∏ⱼ pow(xⱼ, cⱼ), where tᵢ is the configuration variable tᵢ = tan(Δθᵢ/2)
- for the revolute joint, dᵢ = 0, 1, or 2, on the kinematic chain between the
- pair of bodies; xⱼ is the prismatic joint displacement on the same kinematic
- chain, cⱼ = 0 or 1. When we use SeparatingPlaneOrder = kAffine, the monomial
- basis would include all the monomials in the form ∏ᵢ pow(sᵢ, nᵢ) with nᵢ = 0
- or 1, s ∈ {tᵢ, xⱼ}. Let's denote this monomial basis as m(s).
-
- When we have non-polytopic collision geometries, we will also impose a matrix
- of rational functions being positive semidefinite. This requires a matrix-sos
- constraint with slack variable y. We will need the monomial basis in the form
- yᵢ*m(s) with i=0,1,2.
-
- @param[in/out] map_body_to_monomial_basis_array stores all the monomials basis
- already computed.
- @param[out] monomial_basis_array The monomial basis array for this pair of
- body, monomial_basis_array = [m(s), y₀*m(s), y₁*m(s), y₂*m(s)]
- */
-void FindMonomialBasisArray(
-    const multibody::RationalForwardKinematics& rational_forward_kin,
-    const Vector3<symbolic::Variable>& y_slack,
-    const SortedPair<multibody::BodyIndex>& body_pair,
-    std::unordered_map<SortedPair<multibody::BodyIndex>,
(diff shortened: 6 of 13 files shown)


AlexandreAmice (commented):
Checkpoint.

I'd like to push back on the design a bit here. A TC-space box `is-a` TC-space polytope and so I would expect that TC-space box would simply inherit from `CspaceFreePolytope`. On the other hand, I understand that `CspaceFreeBox` is intended to certify boxes in `q-space` by using the `CspaceFreePolytope`. In this situation, I would argue for `CspaceFreeBox` to `have-a` `CspaceFreePolytope` as in this case, I want to think of `CspaceFreeBox` as an object that knows nothing about the TC-space, it is simply using the certifier as a black box.

Perhaps unrelated to this PR, but perhaps we should come up with a coherent plan for the naming of these classes. If `CspaceFreePolytope` is in TC-space, but `CspaceFreeBox` is in C-space I think this could get confusing.

> Reviewed 5 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXQ3wdL-K-CdQptxpwj:btlfy7y)* status: 5 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_box.h` line 31 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ2ZwIBJg9y_XwXGLt:-NXQ2ZwIBJg9y_XwXGLu:bjb7tgp) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.h#L31)):*
> ```objc
> 
>   /**
>    @param plant The plant for which we compute the C-space free polytopes. It
> ```

nit: boxes not polytopes
___
*[`geometry/optimization/cspace_free_box.cc` line 35 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ2w1r9Cn_pRIbxmQY:-NXQ2w1r9Cn_pRIbxmQZ:bglm3xt) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L35)):*
> ```cpp
>   if ((q_box_lower.array() > q_box_upper.array()).any()) {
>         "CspaceFreeBox: q_box_lower has some entry larger than q_box_upper.");
> ```

nit grammar: "has some entries larger than"

Also consider printing out the associate vectors.
___
*[`geometry/optimization/cspace_free_box.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ33ncB1tQVJqbkXMY:-NXQ33ncB1tQVJqbkXMZ:bbdqzdm) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L42)):*
> ```cpp
>   if ((q_box_lower.array() > q_position_upper.array()).any()) {
>         "CspaceFreeBox: q_box_lower has some entry larger the the robot "
> ```

ditto on printing out vector
___
*[`geometry/optimization/cspace_free_box.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ35RQ3qS709QC6n6k:-NXQ35RQ3qS709QC6n6l:b-2bbt4c) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L47)):*
> ```cpp
>   if ((q_box_upper.array() < q_position_lower.array()).any()) {
>         "CspaceFreeBox: q_box_upper has some entry smaller than the robot "
> ```

ditto on printing out vector.
___
*[`geometry/optimization/cspace_free_box.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ3Th52s2vWWYiejVj:-NXQ3Th52s2vWWYiejVk:b7bbuhi) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L50)):*
> ```cpp
>         "positoin lower limit.");
>   }
>   const Eigen::VectorXd q_lower =
> ```

You may want to document this clamping behavior.

hongkai-dai (commented):
To decide whether we use the "has-a" idea, I think we can consider that if we have `CspaceFreePolytope` as a data member inside `CspaceFreeBox`, how many functions can we call in `CspaceFreePolytope` for certifying the box case, and I think we can't call that many functions. There are two reasons
1. CspaceFreePolytope assumes q_star = 0. It then generates all the rational expressions with this q_star=0 assumption, and then formulate the SOS program. On the other hand, CspaceFreeBox doesn't assume q_star=0; it is better to choose q_star = (q_box_lower + q_box_upper)/2, as I explained in that overleaf file. Hence I cannot directly call functions in `CspaceFreePolytope` for `CspaceFreeBox`.
2. In CspaceFreeBox, I hope to support the separating plane order = 0. While for CspaceFreePolytope, the math is almost identical with the plane order = 0 and 1, for CspaceFreeBox, the math would be slightly different between these two plane degrees. So while I can use some of the math in CspaceFreePolytope for CspaceFreeBox with plane_order = 1, I will need to modify the math for plane_order = 0.

Therefore, I think it is better to factor out the common code of CspaceFreePolytope and CspaceFreeBox into a common parent class.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXRM03F5Cq4z3k_KCOz:bvt2o3m)* status: 3 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_box.cc` line 42 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ33ncB1tQVJqbkXMY:-NXRX8xl7UqIhi0p2g_v:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L42)):*

ditto on printing out vector
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_box.cc` line 47 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ35RQ3qS709QC6n6k:-NXRX9Y68PjUV7OQK4Wn:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L47)):*

ditto on printing out vector.
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_box.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXQ3Th52s2vWWYiejVj:-NXRXQuY4H3FbRE5LxCJ:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/5fbd92883fb5d6b6b5f93fd9f7f3d3b8bf61b565/geometry/optimization/cspace_free_box.cc#L50)):*

You may want to document this clamping behavior.
</blockquote></details>

Done. Good call.

AlexandreAmice (commented):
I can see the argument there. In that case, I suggest moving `CspaceFreePolytopeBase` into an internal class and making it abstract. I understand the latter suggestion may make testing a little more obnoxious.

> Reviewed 8 of 13 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXSiwwN58cLQztG72bt:b-xvajxs)* status: 9 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_box.h` line 108 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSlYC337hsUOtjuGoJ:-NXSlYC337hsUOtjuGoK:bphmgsw) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L108)):*
> ```objc
> 
>   /**
>    This class contains the polynomials that we want to certify, that the
> ```

I think here you can drop the second clause.

"This class contains the (either polynomials or numerators of certain rationals) that we wish to certify are non-negative..."
___
*[`geometry/optimization/cspace_free_box.h` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSm4fJ32D4LgOdquYJ:-NXSm4fJ32D4LgOdquYK:bgyqlii) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L112)):*
> ```objc
>    non-negative on s_box_lower <= s <= s_box_upper.
>    */
>   struct CertifyPolynomials {
> ```

Would change this name to 'PolynomialsToCertify`. Currently this reads oddly and my first guess was that these were polynomails that had already been certified.
___
*[`geometry/optimization/cspace_free_box.h` line 124 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSmGVd4alIW3xWGPLQ:-NXSmGVd4alIW3xWGPLR:bg0yde9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L124)):*
> ```objc
>   // tangent-configuration space box computed from the robot position
>   // lower/upper limits.
>   void GenerateCertifyPolynomials(
> ```

Ditto on name
___
*[`geometry/optimization/cspace_free_box.h` line 125 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSmddO9rujJT6c0pcB:-NXSmddO9rujJT6c0pcC:bmpitx9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L125)):*
> ```objc
>   // lower/upper limits.
>   void GenerateCertifyPolynomials(
>       const Eigen::Ref<const Eigen::VectorXd>& s_box_lower,
> ```

I am not sure about the downstream code. But are you sure you don't want to make the arguments here `q_box_upper` and `q_box_lower` and call `ComputeSBox` in this method? It seems like this is the only place `ComputeSBox` would ever be needed so punting this responsibility to future code seems odd. Happy to mark myself as satisfied if you disagree.
___
*[`geometry/optimization/cspace_free_box.cc` line 54 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSmzdM-fvesWh5MIMI:-NXSmzdM-fvesWh5MIMJ:b-daop48) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.cc#L54)):*
> ```cpp
>         "CspaceFreeBox: q_box_upper={} has some entries smaller than the robot "
>         "positoin lower limit={}.",
> ```

typo: position
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSnVm_--kzNXEmCVWp:-NXSnVm_--kzNXEmCVWq:b4bczik) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.h#L102)):*
> ```objc
>   }
> 
>   [[nodiscard]] const Vector3<symbolic::Variable>& y_slack() const {
> ```

This might need documentation unless you think it is covered by the `Options` documentation.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 142 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSnNmw0Jmg6tsziEMm:-NXSnNmw0Jmg6tsziEMn:b-kshxzl) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.h#L142)):*
> ```objc
>   // Forward declare the tester class to test the private members.
>   friend class CspaceFreePolytopeBaseTester;
>   /**
> ```

Should this private comment have Doxygen comments?
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 17 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSkTlm9TWmiOVWOF9j:-NXSkTlm9TWmiOVWOF9k:bu0nqnb) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.cc#L17)):*
> ```cpp
>  for the revolute joint, dᵢ = 0, 1, or 2, on the kinematic chain between the
>  pair of bodies; xⱼ is the prismatic joint displacement on the same kinematic
>  chain, cⱼ = 0 or 1. When we use SeparatingPlaneOrder = kAffine, the monomial
> ```

Do you still want to keep this documentation about `kAffine`?
___
*[`geometry/optimization/test/cspace_free_box_test.cc` line 64 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSkr6u3lCEl9YzBiJN:-NXSkr6u3lCEl9YzBiJO:b-d7mibu) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/test/cspace_free_box_test.cc#L64)):*
> ```cpp
>       s_box_upper,
>       tester.cspace_free_box().rational_forward_kin().ComputeSValue(q_box_upper,
>                                                                     q_star)));
> ```

Can you add a test for the argument that `q_lower` is more than `q_upper` in at least one argument?

hongkai-dai (commented):
I changed it to the internal namespace.

I also tried to make the `CspaceFreePolytopeBase` class an abstract class, by changing its destructor as
```cc
virtual ~CspaceFreePolytopeBase() = 0;
```
But then I got the linking error
```
bazel-out/k8-opt/bin/geometry/optimization/_objs/cspace_free_polytope_test/cspace_free_polytope_test.pic.o:cspace_free_polytope_test.cc:function drake::geometry::optimization::CspaceFreePolytope::~CspaceFreePolytope(): error: undefined reference to 'drake::geometry::optimization::internal::CspaceFreePolytopeBase::~CspaceFreePolytopeBase()'
bazel-out/k8-opt/bin/geometry/optimization/_objs/cspace_free_polytope/cspace_free_polytope.pic.o:cspace_free_polytope.cc:function drake::geometry::optimization::CspaceFreePolytope::CspaceFreePolytope(drake::multibody::MultibodyPlant<double> const*, drake::geometry::SceneGraph<double> const*, drake::geometry::optimization::SeparatingPlaneOrder, Eigen::Ref<Eigen::Matrix<double, -1, 1, 0, -1, 1> const, 0, Eigen::InnerStride<1> > const&, drake::geometry::optimization::internal::CspaceFreePolytopeBase::Options const&) [clone .cold]: error: undefined reference to 'drake::geometry::optimization::internal::CspaceFreePolytopeBase::~CspaceFreePolytopeBase()'
```
@jwnimmer-tri is this the right way to make the class an abstract class by removing its destructor? Also, is this a good idea to make the class abstract, if I don't want to instantiate a concrete instance of this class?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXYTyqs9g9EliBH6H2L:bdqaiaz)* status: 8 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/optimization/cspace_free_box.h` line 108 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSlYC337hsUOtjuGoJ:-NXYUs6I4fp_dtE5BNqj:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L108)):*

I think here you can drop the second clause.

"This class contains the (either polynomials or numerators of certain rationals) that we wish to certify are non-negative..."
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_box.h` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSm4fJ32D4LgOdquYJ:-NXYV3EABrHwBKgGM4oZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L112)):*

Would change this name to 'PolynomialsToCertify`. Currently this reads oddly and my first guess was that these were polynomails that had already been certified.
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_box.h` line 124 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSmGVd4alIW3xWGPLQ:-NXYV4E29WEp-0mj_39M:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L124)):*

Ditto on name
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_box.h` line 125 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSmddO9rujJT6c0pcB:-NXYWlNXDlRos9d_6vrh:bnouby8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_box.h#L125)):*

I am not sure about the downstream code. But are you sure you don't want to make the arguments here `q_box_upper` and `q_box_lower` and call `ComputeSBox` in this method? It seems like this is the only place `ComputeSBox` would ever be needed so punting this responsibility to future code seems odd. Happy to mark myself as satisfied if you disagree.
</blockquote></details>

Since this is a private function, I didn't aim for ease of use (if this were a public function I would definitely use `q_box_lower` and `q_box_upper` as the input). 

If in the end we find that ComputeSBox is only used with this function, then I think we can put it inside this function. At the moment I would like to keep it in the current form, in case ComputeSBox is used elsewhere. I added a TODO here.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSnVm_--kzNXEmCVWp:-NXYXqsi-TKdliMkWXsW:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.h#L102)):*

This might need documentation unless you think it is covered by the `Options` documentation.
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 142 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSnNmw0Jmg6tsziEMm:-NXYXvEj3otUGUgh5mxZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.h#L142)):*

Should this private comment have Doxygen comments?
</blockquote></details>

Done.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 17 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSkTlm9TWmiOVWOF9j:-NXYY1_o8lUnug5mc9h9:bqtng5c) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/cspace_free_polytope_base.cc#L17)):*

Do you still want to keep this documentation about `kAffine`?
</blockquote></details>

I think when #19526 is merged I will change this documentation to use plane_order = 1. But for the moment I will use this `SeparatingPlaneOrder` as that is in the master.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)* status: 8 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/test/cspace_free_box_test.cc` line 64 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXSkr6u3lCEl9YzBiJN:-NXYa-Dj7nOTd6a4F7ir:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b7735495c139f26365662b6db34bf2518917c98/geometry/optimization/test/cspace_free_box_test.cc#L64)):*

Can you add a test for the argument that `q_lower` is more than `q_upper` in at least one argument? 
</blockquote></details>

Done.

jwnimmer-tri:
You could switch the constructor of `CspaceFreePolytopeBase` from `public:` to `protected:`.  Would that accomplish what you're looking for?

AlexandreAmice (commented):
One last nit on the name of the base class. Can we change the name to `CspaceFreeRegion` instead of `CspaceFreePolytopeBase`. I think if someone were to come along and decide that they wanted to `CspaceFreeEllipse` they would still want to inherit from this new class and so then it would be a misnomer.

> Reviewed 9 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXaD8AJE3aeMM_DH_2X:b-ddcbo)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
> You could switch the constructor of `CspaceFreePolytopeBase` from `public:` to `protected:`. Would that accomplish what you're looking for?

That is perfect, I forgot this trick. Thanks Jeremy

> One last nit on the name of the base class. Can we change the name to `CspaceFreeRegion` instead of `CspaceFreePolytopeBase`. I think if someone were to come along and decide that they wanted to `CspaceFreeEllipse` they would still want to inherit from this new class and so then it would be a misnomer.

I think if we want to support more generic type of regions, then we probably need to change several functions inside `CspaceFreePolytopeBase` class, such as `CalcSBoundsPolynomial` and `CalcMonomialBasis`. That seems quite a big change in `CspaceFreePolytopeBase` class.

If we do want to support CspaceFreeRegion for generic type, I think we can later add a new base class on top of CspaceFreePolytopeBase. What do you think?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXaLAC76Vat2KEWttAp:b2e3u5i)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

AlexandreAmice (commented):
Fair enough.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXaMOH76q03rwla0RCT:brb07kd)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
+@xuchenhan-tri for platform review please, thanks!

Sorry for the large PR. But a large portion of the PR is to move the code from `cspace_free_polytope.h/cc` to `cspace_free_polytope_base.h/cc`, so the actual code change is still below 700 lines. The code in `cspace_free_polytope_base.h/cc` has been previously platform reviewed.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXkVxv5CSgCTiravwU8:b-fxplhx)* status: LGTM missing from assignee xuchenhan-tri(platform)

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform)

___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkWpK5AynjMLH9shos:-NXkWpK5AynjMLH9shot:bc05axi) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/test/cspace_free_polytope_base_test.cc#L13)):*
> ```cpp
> namespace optimization {
> namespace internal {
> class CspaceFreePolytopeDummy : public CspaceFreePolytopeBase {
> ```

nit. To reviewer: this test is just moved from `cspace_free_polytope_test.cc`, no functionality change in this file.

xuchenhan-tri (commented):
Platform :lgtm: with a few comments.

BTW, why the release notes none tag? Is this a feature under development that we haven't released yet?

> Reviewed 3 of 13 files at r1, 8 of 10 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXkzw1UBINyjWPYes2p:b-vdpcno)* status: 15 unresolved discussions

___
*[`geometry/optimization/cspace_free_box.h` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl-aig8PF9p7hqs3J3:-NXl-aig8PF9p7hqs3J4:b-nck5df) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L13)):*
> ```objc
> namespace optimization {
> /**
>  This class tries to find large axis-aligned bounding boxes in the configuration
> ```

Btw, does this class search for "the largest" collision free box? or just a "large" one?



_Code quote:_
```objc
large
```
___
*[`geometry/optimization/cspace_free_box.h` line 26 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkxFv16L9d2rBqKW56:-NXkxFv28JX3LcjM3dvy:b74wcir) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L26)):*
> ```objc
>   using CspaceFreePolytopeBase::IgnoredCollisionPairs;
> 
>   ~CspaceFreeBox() override = default;
> ```

nit, this destructor shouldn't be inlined according to https://drake.mit.edu/styleguide/cppguide.html#Inline_Functions



_Code quote:_
```objc
~CspaceFreeBox() override = default;
```
___
*[`geometry/optimization/cspace_free_box.h` line 53 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkvq9r8VxqJpJoKIbo:-NXkvq9s5VT0UgDjZnu1:b-cij144) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L53)):*
> ```objc
>    polytopic constraint s - s_box_lower >= 0, s_box_upper - s >= 0.
>    */
>   class SeparatingPlaneLagrangians {
> ```

BTW, we usually declare nested classes before constructors. See https://drake.mit.edu/styleguide/cppguide.html#Declaration_Order
___
*[`geometry/optimization/cspace_free_box.h` line 95 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkwDcH-f90mTE4q6HU:-NXkwDcH-f90mTE4q6HV:b-1us0i2) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L95)):*
> ```objc
>   friend class CspaceFreeBoxTester;
> 
>   /**
> ```

nit odd doxygen comments in private sections. Ditto below



_Code quote:_
```objc
/**
```
___
*[`geometry/optimization/cspace_free_box.h` line 96 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkwKZ231HTx4X0BmFy:-NXkwKZ231HTx4X0BmFz:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L96)):*
> ```objc
> 
>   /**
>    Compute the range of s from the box q_box_lower <= q <= q_box_upper. We also
> ```

typo



_Suggestion:_
```objc
Computes
```
___
*[`geometry/optimization/cspace_free_box.h` line 117 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkxYpr6UmW8xHZZkpO:-NXkxYpr6UmW8xHZZkpP:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L117)):*
> ```objc
>   };
> 
>   // Generate the polynomials used for certifying the box s_box_lower <= s <=
> ```

typo



_Suggestion:_
```objc
 Generates
```
___
*[`geometry/optimization/cspace_free_box.h` line 121 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkwatg8BQjclqcXkVC:-NXkwatg8BQjclqcXkVD:b-glx362) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L121)):*
> ```objc
>   // @note The box [s_box_lower, s_box_upper] is already inside the
>   // tangent-configuration space box computed from the robot position
>   // lower/upper limits.
> ```

BTW, it looks like this function throws if this isn't satisfied. Probably worth documenting it as `@throws` along with a couple of other throw conditions.



_Code quote:_
```objc
  // @note The box [s_box_lower, s_box_upper] is already inside the
  // tangent-configuration space box computed from the robot position
  // lower/upper limits.
```
___
*[`geometry/optimization/cspace_free_box.cc` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkyA_C49t3nRds-31p:-NXkyA_C49t3nRds-31q:b-y8bdm3) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.cc#L13)):*
> ```cpp
> namespace optimization {
> CspaceFreeBox::CspaceFreeBox(const multibody::MultibodyPlant<double>* plant,
>                              const geometry::SceneGraph<double>* scene_graph,
> ```

nit, we should probably check for null pointers somewhere.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 17 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl0xlC65BLFLceyC_V:-NXl0xlC65BLFLceyC_W:b-e2zmai) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L17)):*
> ```objc
> namespace geometry {
> namespace optimization {
> namespace internal {
> ```

Minor, by putting this class in the internal namespace, all documentation you have in `/** ... */` in this file will not be visible from Doxygen. I want to make sure this is intentional. If so, we should change to `/* ... */` to prevent illusions that they do appear in Doxygen.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 30 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXko4Vn9yX1sK1UVwNs:-NXko4Vn9yX1sK1UVwNt:b74wcir) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L30)):*
> ```objc
>       std::unordered_set<SortedPair<geometry::GeometryId>>;
> 
>   virtual ~CspaceFreePolytopeBase() = default;
> ```

nit, this destructor shouldn't be inlined according to https://drake.mit.edu/styleguide/cppguide.html#Inline_Functions



_Code quote:_
```objc
virtual ~CspaceFreePolytopeBase() = default;
```
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 78 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkp5138nfF9giqlZ8s:-NXkp5138nfF9giqlZ8t:b6maji2) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L78)):*
> ```objc
>   };
> 
>   [[nodiscard]] const multibody::RationalForwardKinematics&
> ```

BTW, a few functions in this class are without documentation. Their meanings are not completely clear to me. I'm not sure if it is because of my lack of domain knowledge. Consider adding a few notes where you see appropriate. 
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 85 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkoSIM8DSOy8cxXFA3:-NXkoSIM8DSOy8cxXFA4:bhs4pxk) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L85)):*
> ```objc
>   /**
>    separating_planes()[map_geometries_to_separating_planes.at(geometry1_id,
>    geometry2_id)] is the separating plane that separates geometry1 and
> ```

nit  spelling consistency.



_Suggestion:_
```objc
geometry 1
```
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 109 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkxu9jELJMwV6iLwdK:-NXkxu9jELJMwV6iLwdL:bu76j54) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L109)):*
> ```objc
>   // We put the constructor in protected method to make sure that the user
>   // cannot instantiate a CspaceFreePolytopeBase instance.
>   CspaceFreePolytopeBase(const multibody::MultibodyPlant<double>* plant,
> ```

nit, preconditions of the constructor are missing from the documentation. As I understand it, plant and scene_graph needs to be non-null and wired up.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 187 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkpY4vCTstC91bkQj5:-NXkpY4vCTstC91bkQj6:b-e6b950) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L187)):*
> ```objc
> 
>   // See Options::with_cross_y for its meaning.
>   bool with_cross_y_;
> ```

BTW, why not store the options passed in at constructor instead?

hongkai-dai (commented):
Exactly, the feature is still under development and incomplete.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-:-NXlZ30X9SfYxO2LyHVt:b-27c63k)* status: 4 unresolved discussions

___
*[`geometry/optimization/cspace_free_box.h` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl-aig8PF9p7hqs3J3:-NXlZ9aXDdbSs36b5nNw:b49i70l) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L13)):*

Btw, does this class search for "the largest" collision free box? or just a "large" one?
</blockquote></details>

It finds a large one.

We are trying to maximize a function f(x), but this function is too hard to maximize, so we find an approximate function f_approx(x) ~ f(x) and maximize this f_approx(x). Furthermore, this maximizing this f_approx(x) is not a convex program hence we are not guaranteed to find the global maximal f_approx(x).
___
*[`geometry/optimization/cspace_free_box.h` line 121 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkwatg8BQjclqcXkVC:-NXldYYo3c6321abicnR:bvjervf) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L121)):*

BTW, it looks like this function throws if this isn't satisfied. Probably worth documenting it as `@throws` along with a couple of other throw conditions.
</blockquote></details>

It doesn't really throw in this function `GeneratePolynomialsToCertify`. It throws in a different function `ComputeSBox`,  I just added @throws there.
___
*[`geometry/optimization/cspace_free_box.cc` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkyA_C49t3nRds-31p:-NXleO1WBBwzuOj9DBW1:b-xcjkk6) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.cc#L13)):*

nit, we should probably check for null pointers somewhere.
</blockquote></details>

Done. I checked it in the parent class `CspaceFreePolytopeBase`.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 17 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl0xlC65BLFLceyC_V:-NXlezP19MzfXY3V70Hm:b-obtor5) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L17)):*

Minor, by putting this class in the internal namespace, all documentation you have in `/** ... */` in this file will not be visible from Doxygen. I want to make sure this is intentional. If so, we should change to `/* ... */` to prevent illusions that they do appear in Doxygen.
</blockquote></details>

Good call, I didn't know internal namespace ignores doxygen documentation. I do want to show these documentation in doxygen so I removed the internal namespace.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 187 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXkpY4vCTstC91bkQj5:-NXlgXPNDI7Kvw7k_x-f:b-p3f8mb) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L187)):*

BTW, why not store the options passed in at constructor instead?
</blockquote></details>

Just to save some typings, that we can type `this->with_cross_y_` instead of `this->options.with_cross_y_`.

xuchenhan-tri (commented):
> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)* status: 2 unresolved discussions

___
*[`geometry/optimization/cspace_free_box.h` line 13 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl-aig8PF9p7hqs3J3:-NXlt0mN-zYUrtHN7Pt0:b-jj45e9) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_box.h#L13)):*

It finds a large one.

We are trying to maximize a function f(x), but this function is too hard to maximize, so we find an approximate function f_approx(x) ~ f(x) and maximize this f_approx(x). Furthermore, this maximizing this f_approx(x) is not a convex program hence we are not guaranteed to find the global maximal f_approx(x).
</blockquote></details>

I see, that makes sense. Thanks for the explanation.

Personally I'd like to see this comment captured in the class doc in some shape. However, my main goal is to make sure it's not a typo :). Marking myself satisfied.
___
*[`geometry/optimization/cspace_free_box.h` line 102 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXlv2rt39Jn5Lza20QS:-NXlv2rt39Jn5Lza20QT:blfkwax) ([raw file](https://github.com/RobotLocomotion/drake/blob/6ff7818570d9098b1d5ab686e2a12aef4d864184/geometry/optimization/cspace_free_box.h#L102)):*
> ```objc
>    q_box_upper) within the robot position limits.
>    @throws error if any q_box_lower is larger than q_box_upper or the robot
>    position upper limit; similar throws an error if any q_box_upper is smaller
> ```

typo(?)



_Suggestion:_
```objc
similarly
```
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 17 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl0xlC65BLFLceyC_V:-NXluj_p78sBZX9aCoLg:b-r77bt9) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L17)):*

Good call, I didn't know internal namespace ignores doxygen documentation. I do want to show these documentation in doxygen so I removed the internal namespace.
</blockquote></details>

FYI, it's configured here: https://github.com/RobotLocomotion/drake/blob/51c7d22407fe1928ba12fc6cfcfa4264cf292b78/doc/doxygen_cxx/Doxyfile_CXX.in#LL136C1-L136C16

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)* status: :shipit: complete! all discussions resolved, LGTM from assignees AlexandreAmice,xuchenhan-tri(platform)

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 17 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19552#-NXl0xlC65BLFLceyC_V:-NXlxPLPAGy8VynMRE5v:b-kufkxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/14edb077b7b67aee5abe1d7fe0ed93dc927f07d5/geometry/optimization/cspace_free_polytope_base.h#L17)):*

FYI, it's configured here: https://github.com/RobotLocomotion/drake/blob/51c7d22407fe1928ba12fc6cfcfa4264cf292b78/doc/doxygen_cxx/Doxyfile_CXX.in#LL136C1-L136C16
</blockquote></details>

I see, thanks!

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19552)* status: :shipit: complete! all discussions resolved, LGTM from assignees AlexandreAmice,xuchenhan-tri(platform)

Issue #18830, opened 2023-02-20
IRIS using a CollisionChecker


jwnimmer-tri:
We currently offer these two functions:

```c++
ConvexSets MakeIrisObstacles(
    const QueryObject<double>& query_object,
    std::optional<FrameId> reference_frame = std::nullopt);

HPolyhedron IrisInConfigurationSpace(
    const multibody::MultibodyPlant<double>& plant,
    const systems::Context<double>& context,
    const IrisOptions& options = IrisOptions());
```

In both, the collision geometry comes from a query object, either directly (in the first) or the `plant` input port (in the second).

The `drake::planning::CollisionChecker` offers a more natural interface for collision queries used for motion planning, and most particularly the ability to configure collision padding on a per-body-pair basis.

It would make IRIS a lot more useful to more people if we could run it using padding, and therefore if it accepted a `CollisionChecker` as an alternative source of geometry and kinematics.

\CC @calderpg-tri @RussTedrake FYI

hongkai-dai:
To make sure we are on the same page, I think you are proposing these API

```cc
ConvexSets MakeIrisObstacles(const CollisionChecker& collision_checker, std::optional<FrameId> reference_frame = std::nullopt);

HPolyhedron IrisInConfigurationSpace(const CollisionChecker& collision_checker,  const IrisOptions& options = IrisOptions());
```

In the function `IrisInConfigurationSpace` with a CollisionChecker, I suppose we do not need to pass in a Context<double> object as the input argument? Because CollisionChecker already has a context `CollisionChecker::plant_context()` in it (but the documentation says the function might be tossed)? https://github.com/RobotLocomotion/drake/blob/6f1a88111c7b04217d461e906521a9ac7f6e58ed/planning/collision_checker.h#L239-L245.

jwnimmer-tri:
The documentation of the existing `IrisInConfigurationSpace` says that the `initialIRIS seed configuration` is passed via the Context.

In the new function, I think asking the user to call `CollisionChecker::UpdatePositions` to pass the seed would be awkward at best. Perhaps the new function can just take an Eigen `q` vector as argument for the seed.

sadraddini:
```

> ConvexSets MakeIrisObstacles(const CollisionChecker& collision_checker, std::optional<FrameId> reference_frame = std::nullopt);
> 
> HPolyhedron IrisInConfigurationSpace(const CollisionChecker& collision_checker,  const IrisOptions& options = IrisOptions());

```

[...]

PR #19569, opened 2023-06-10
Add InCollisionConstraint to be used for IrisInConfigurationSpace.

hongkai-dai:
Also move some internal functions in iris.cc to iris_internal.h/cc.

Towards #18830

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -171,6 +171,7 @@ drake_cc_library(
     deps = [
         ":convex_set",
         "//multibody/plant",
+        "//planning:collision_checker",
         "//solvers:mathematical_program",
         "//solvers:solve",
     ],
@@ -186,6 +187,7 @@ drake_cc_library(
         "//common:name_value",
         "//geometry:scene_graph",
         "//multibody/plant",
+        "//planning:collision_checker",
         "//solvers:choose_best_solver",
         "//solvers:ipopt_solver",
         "//solvers:snopt_solver",
@@ -195,6 +197,7 @@ drake_cc_library(
 drake_cc_library(
     name = "test_utilities",
     testonly = 1,
+    visibility = ["//visibility:private"],
     srcs = ["test_utilities.cc"],
     hdrs = ["test_utilities.h"],
     deps = [
@@ -382,6 +385,31 @@ drake_cc_googletest(
         "//multibody/inverse_kinematics",
         "//multibody/parsing:parser",
         "//systems/framework:diagram_builder",
+    ],
+)
+
+drake_cc_library(
+    name = "iris_test_utilities",
+    testonly = 1,
+    visibility = ["//visibility:private"],
+    srcs = ["test/iris_test_utilities.cc"],
+    hdrs = ["test/iris_test_utilities.h"],
+    deps = [
+        "//geometry:scene_graph",
+        "//multibody/plant",
+        "//planning:robot_diagram_builder",
+        "@gtest//:without_main",
+    ],
+)
+
+drake_cc_googletest(
+    name = "iris_internal_test",
+    deps = [
+        ":iris_internal",
+        ":iris_test_utilities",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+        "//planning:scene_graph_collision_checker",
     ],
 )
 
--- a/geometry/optimization/iris.h
+++ b/geometry/optimization/iris.h
@@ -8,6 +8,7 @@
 #include "drake/geometry/optimization/convex_set.h"
 #include "drake/geometry/optimization/hpolyhedron.h"
 #include "drake/multibody/plant/multibody_plant.h"
+#include "drake/planning/collision_checker.h"
 
 namespace drake {
 namespace geometry {
@@ -193,7 +194,6 @@ HPolyhedron IrisInConfigurationSpace(
     const multibody::MultibodyPlant<double>& plant,
     const systems::Context<double>& context,
     const IrisOptions& options = IrisOptions());
-
 }  // namespace optimization
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/optimization/iris_internal.cc
+++ b/geometry/optimization/iris_internal.cc
@@ -1,11 +1,16 @@
 #include "drake/geometry/optimization/iris_internal.h"
 
+#include <algorithm>
 #include <limits>
+#include <utility>
+
+#include <fmt/format.h>
 
 namespace drake {
 namespace geometry {
 namespace optimization {
 namespace internal {
+const double kInf = std::numeric_limits<double>::infinity();
 
 using Eigen::MatrixXd;
 using Eigen::VectorXd;
@@ -160,6 +165,83 @@ bool ClosestCollisionProgram::Solve(
     return true;
   }
   return false;
+}
+
+namespace {
+const planning::CollisionChecker& get_collision_checker(
+    const planning::CollisionChecker* ptr) {
+  if (ptr == nullptr) {
+    throw std::runtime_error("get_collision_checker(): got a nullptr");
+  }
+  return *ptr;
+}
+}  // namespace
+
+InCollisionConstraint::InCollisionConstraint(
+    SortedPair<multibody::BodyIndex> body_pair,
+    const planning::CollisionChecker* collision_checker,
+    double influence_distance)
+    : solvers::Constraint(
+          1, get_collision_checker(collision_checker).plant().num_positions(),
+          Vector1d(-kInf), Vector1d(0)),
+      body_pair_{std::move(body_pair)},
+      collision_checker_{collision_checker},
+      influence_distance_{influence_distance} {
+  if (collision_checker_->IsCollisionFilteredBetween(body_pair_.first(),
+                                                     body_pair_.second())) {
+    throw std::runtime_error(fmt::format(
+        "InCollisionConstraint(): collision between the body {} and {} is "
+        "filtered",
+        collision_checker_->plant().get_body(body_pair_.first()).name(),
+        collision_checker_->plant().get_body(body_pair_.second()).name()));
+  }
+}
+
+void InCollisionConstraint::DoEval(const Eigen::Ref<const Eigen::VectorXd>& x,
+                                   Eigen::VectorXd* y) const {
+  y->resize(1);
+  const planning::RobotClearance robot_clearance =
+      collision_checker_->CalcRobotClearance(x, influence_distance_);
+  // Loop through every entry in robot_clearance, find the smallest distance for
+  // that pair of bodies.
+  (*y)(0) = influence_distance_;
+  for (int i = 0; i < robot_clearance.size(); ++i) {
+    if (SortedPair<multibody::BodyIndex>(robot_clearance.robot_indices()[i],
+                                         robot_clearance.other_indices()[i]) ==
+        body_pair_) {
+      (*y)(0) = std::min((*y)(0), robot_clearance.distances()(i));
+    }
+  }
+}
+
+void InCollisionConstraint::DoEval(const Eigen::Ref<const AutoDiffVecXd>& x,
+                                   AutoDiffVecXd* y) const {
+  y->resize(1);
+  const Eigen::VectorXd q_val = math::ExtractValue(x);
+  const Eigen::MatrixXd q_grad = math::ExtractGradient(x);
+  const planning::RobotClearance robot_clearance =
+      collision_checker_->CalcRobotClearance(q_val, influence_distance_);
+  // Loop through every entry in robot_clearance, find the smallest distance for
+  // that pair of bodies.
+  (*y)(0).value() = influence_distance_;
+  for (int i = 0; i < robot_clearance.size(); ++i) {
+    if (SortedPair<multibody::BodyIndex>(robot_clearance.robot_indices()[i],
+                                         robot_clearance.other_indices()[i]) ==
+        body_pair_) {
+      if (robot_clearance.distances()(i) < (*y)(0).value()) {
+        (*y)(0).value() = robot_clearance.distances()(i);
+        (*y)(0).derivatives() =
+            (robot_clearance.jacobians().row(i) * q_grad).transpose();
+      }
+    }
+  }
+}
+
+void InCollisionConstraint::DoEval(
+    const Eigen::Ref<const VectorX<symbolic::Variable>>&,
+    VectorX<symbolic::Expression>*) const {
+  throw std::runtime_error(
+      "InCollisionConstraint doesn't support Eval with symbolic variable.");
 }
 }  // namespace internal
 }  // namespace optimization
--- a/geometry/optimization/iris_internal.h
+++ b/geometry/optimization/iris_internal.h
@@ -10,6 +10,7 @@
 #include "drake/geometry/optimization/convex_set.h"
 #include "drake/geometry/optimization/hyperellipsoid.h"
 #include "drake/multibody/plant/multibody_plant.h"
+#include "drake/planning/collision_checker.h"
 #include "drake/solvers/mathematical_program.h"
 #include "drake/solvers/solver_interface.h"
 
@@ -87,6 +88,37 @@ class ClosestCollisionProgram {
   solvers::MathematicalProgram prog_;
   solvers::VectorXDecisionVariable q_;
   std::optional<solvers::Binding<solvers::LinearConstraint>> P_constraint_{};
+};
+
+/*
+ The body pair (A, B) is in collision at this posture.
+ */
+class InCollisionConstraint final : public solvers::Constraint {
+ public:
+  DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(InCollisionConstraint)
+
+  /* Constructor.
+   Throws an error if the collision between body_pair is filtered.
+   @param collision_checker should remain alive for the life span of this
+   constraint.
+   */
+  InCollisionConstraint(SortedPair<multibody::BodyIndex> body_pair,
+                        const planning::CollisionChecker* collision_checker,
+                        double influence_distance);
+
+  ~InCollisionConstraint() override {}
+
+ private:
+  void DoEval(const Eigen::Ref<const Eigen::VectorXd>& x,
+              Eigen::VectorXd* y) const override;
+  void DoEval(const Eigen::Ref<const AutoDiffVecXd>& x,
+              AutoDiffVecXd* y) const override;
+  void DoEval(const Eigen::Ref<const VectorX<symbolic::Variable>>& x,
+              VectorX<symbolic::Expression>* y) const override;
+
+  const SortedPair<multibody::BodyIndex> body_pair_;
+  const planning::CollisionChecker* const collision_checker_;
+  double influence_distance_{};
 };
 }  // namespace internal
 }  // namespace optimization
--- a/geometry/optimization/test/iris_internal_test.cc
+++ b/geometry/optimization/test/iris_internal_test.cc
@@ -1,0 +1,87 @@
+#include "drake/geometry/optimization/iris_internal.h"
+
+#include <limits>
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
+#include "drake/geometry/optimization/test/iris_test_utilities.h"
+#include "drake/planning/scene_graph_collision_checker.h"
+
+namespace drake {
+namespace geometry {
+namespace optimization {
+namespace internal {
+
+const double kInf = std::numeric_limits<double>::infinity();
+
+TEST_F(Toy2DofRobotTest, InCollisionConstraint) {
+  planning::CollisionCheckerParams params;
+  const auto& plant = builder_.plant();
+  params.model = builder_.Build();
+  params.robot_model_instances.push_back(multibody::default_model_instance());
+  params.configuration_distance_function = [](const Eigen::VectorXd& q1,
+                                              const Eigen::VectorXd& q2) {
+    return (q1 - q2).norm();
+  };
+  params.edge_step_size = 0.05;
+  planning::SceneGraphCollisionChecker collision_checker(std::move(params));
+
+  const double influence_distance = 2;
+  const SortedPair<multibody::BodyIndex> body_pair(body_indices_[0],
+                                                   plant.world_body().index());
+  InCollisionConstraint dut(body_pair, &collision_checker, influence_distance);
+  EXPECT_EQ(dut.num_vars(), plant.num_positions());
+  EXPECT_EQ(dut.num_constraints(), 1);
+  EXPECT_TRUE(CompareMatrices(dut.upper_bound(), Vector1d(0)));
+  EXPECT_TRUE(CompareMatrices(dut.lower_bound(), Vector1d(-kInf)));
+
+  // Check Eval for double.
+  Eigen::VectorXd q(2);
+  q << 0.1, 0.2;
+  Eigen::VectorXd y;
+  dut.Eval(q, &y);
+  EXPECT_EQ(y.rows(), 1);
+  const planning::RobotClearance robot_clearance =
+      collision_checker.CalcRobotClearance(q, influence_distance);
+
+  double min_distance = kInf;
+  Eigen::RowVectorXd dphi_dq;
+  for (int i = 0; i < robot_clearance.size(); ++i) {
+    if (SortedPair<multibody::BodyIndex>(robot_clearance.robot_indices()[i],
+                                         robot_clearance.other_indices()[i]) ==
+            body_pair &&
+        robot_clearance.distances()(i) < min_distance) {
+      min_distance = robot_clearance.distances()(i);
+      dphi_dq = robot_clearance.jacobians().row(i);
+    }
+  }
+  const double kTol = 1E-12;
+  EXPECT_NEAR(min_distance, y(0), kTol);
+
+  // Check Eval for autodiff.
+  const Eigen::MatrixXd q_grad = Eigen::Vector2d(0.5, -2);
+  const AutoDiffVecXd x_ad = math::InitializeAutoDiff(q, q_grad);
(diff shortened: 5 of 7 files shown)


hongkai-dai (commented):
+@jwnimmer-tri for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-:-NXXhMRA8mRMs0T1_bRE:b-o4a2ka)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
To reviewers: I moved the class `SamePointConstraint` and `ClosestCollisionProgram` from `iris.cc` to `iris_internal.h/cc` because I think these classes are useful (and I might add some unit tests to these programs later). I didn't make any functional changes on these two classes. The only new class I added in `iris_internal.h/cc` is `InCollisionConstraint`.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-:-NXXhUc511j31Kd_ESbO:b-wlxr6r)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri:
To me, the code relocation of existing material is substantial enough that it should happen in its own separate pull request.  If you like, I can submit that separately myself.

hongkai-dai (commented):
I filed #19573 to move the code to `iris_internal.h/cc`.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-:-NXaBgna7mc0ZZNM8K8w:b-8rs787)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri (commented):
> Reviewed 2 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 15 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/BUILD.bazel` line 392 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbkPRI9dRDMJby6nIv:-NXbkPRI9dRDMJby6nIw:b-85332u) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/BUILD.bazel#L392)):*
> ```Starlark
> drake_cc_library(
>     name = "iris_test_utilities",
>     testonly = 1,
> ```

nit Missing `visibility = ["//visibility:private"]` here.
___
*[`geometry/optimization/iris.h` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbje6gCryAUxgJsfgU:-NXbje6hDIvej-bouABN:b-h6ozgu) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris.h#L198)):*
> ```objc
>     const IrisOptions& options = IrisOptions());
> 
> HPolyhedron IrisInConfigurationSpace(
> ```

nit Missing Doxygen docstring. 

... Oh, actually this is just a stray signature isn't it?  The function is not implemented, it seems.
___
*[`geometry/optimization/iris_internal.h` line 118 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbjXE32XBvTDXwmSh6:-NXbjXE32XBvTDXwmSh7:b2ni8l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.h#L118)):*
> ```objc
> 
>   SortedPair<multibody::BodyIndex> body_pair_;
>   const planning::CollisionChecker* collision_checker_;
> ```

nit



_Suggestion:_
```objc
  const SortedPair<multibody::BodyIndex> body_pair_;
  const planning::CollisionChecker* const collision_checker_;
```
___
*[`geometry/optimization/iris_internal.cc` line 177 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbispx5RfO034tRS7p:-NXbispx5RfO034tRS7q:b-8ntnkc) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L177)):*
> ```cpp
>                           Vector1d(-kInf), Vector1d(0)),
>       body_pair_{std::move(body_pair)},
>       collision_checker_{&collision_checker},
> ```

When a class retains a pointer to an argument passed into its constructor, we typically use a `const CollisionChecker*` pointer (instead of a reference) to help communicate the lifetime requirement to our callers.
___
*[`geometry/optimization/iris_internal.cc` line 178 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbhg7YAxdezmxOYvvp:-NXbhg7YAxdezmxOYvvq:b-v8y5sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L178)):*
> ```cpp
>       body_pair_{std::move(body_pair)},
>       collision_checker_{&collision_checker},
>       influence_distance_{influence_distance} {
> ```

If `influence_distance` is malformed, probably we should fail-fast here instead of later during Eval?
___
*[`geometry/optimization/iris_internal.cc` line 187 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbi4OH9gnB1iw7GdDv:-NXbi4OH9gnB1iw7GdDw:b-kii0j5) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L187)):*
> ```cpp
>         collision_checker_->plant().get_body(body_pair_.second()).name()));
>   }
> }
> ```

If the two indices are both non-robot ("other"), we need to fail-fast here.  The `CalcRobotClearance` does not report environment-environment clearance distances.  Needs a unit test.
___
*[`geometry/optimization/iris_internal.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbiJwVBzUXo2tHBcV_:-NXbiJwVBzUXo2tHBcVa:b-livg4n) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L195)):*
> ```cpp
>       collision_checker_->CalcRobotClearance(x, influence_distance_);
>   // Loop through every entry in robot_clearance, find the smallest distance for
>   // that pair of bodies.
> ```

We have `MinimumValueConstraint` available to blend multiple measurements into a single threshold with smooth gradients.  Why aren't we using it here?
___
*[`geometry/optimization/iris_internal.cc` line 215 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbjpLm3cH5Nnv2P0WC:-NXbjpLn5IUZQ-IlkYO1:b-pud28l) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L215)):*
> ```cpp
>   // Loop through every entry in robot_clearance, find the smallest distance for
>   // that pair of bodies.
>   (*y)(0).value() = influence_distance_;
> ```

When there are no collisions, it seems like this loop fails to zero out or delete the gradient of `y`.  Needs a unit test.
___
*[`geometry/optimization/test/iris_internal_test.cc` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbkbNa91N7JHDVTWVW:-NXbkbNa91N7JHDVTWVX:btby6bx) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_internal_test.cc#L19)):*
> ```cpp
> const double kInf = std::numeric_limits<double>::infinity();
> 
> class A {
> ```

nit A and B seem vestigial?
___
*[`geometry/optimization/test/iris_internal_test.cc` line 56 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbnl0XCzUqcDxQDRqC:-NXbnl0XCzUqcDxQDRqD:b-kdv9iy) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_internal_test.cc#L56)):*
> ```cpp
>   // Check Eval for double.
>   Eigen::VectorXd q(2);
>   q << 0.1, 0.2;
> ```

nit Given the difficult-to-read toy robot model, I can't tell whether or not this posture exercises the "multiple collisions, select the closest one" logic within the implementation.  If it does, a comment might help explain.

(See the test_utilities discussions re: readability of the toy model.)
___
*[`geometry/optimization/test/iris_internal_test.cc` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbluyrBPXuh0qV--Q5:-NXbluyrBPXuh0qV--Q6:b-77y8zf) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_internal_test.cc#L73)):*
> ```cpp
>       dphi_dq = robot_clearance.jacobians().row(i);
>     }
>   }
> ```

We've copy-pasted the implementation into the unit test?  Surely there must be a better way.  With carefully-chosen 
___
*[`geometry/optimization/test/iris_internal_test.cc` line 90 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXblX5_DdsIsg5OyqwB:-NXblX5_DdsIsg5OyqwC:bhyaucw) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_internal_test.cc#L90)):*
> ```cpp
>       InCollisionConstraint(SortedPair(body_indices_[0], body_indices_[1]),
>                             collision_checker, influence_distance),
>       fmt::format(
> ```

nit I don't see any reason why we need fmt complexity.  On first glance, it seems like a plain literal string would suffice?

The comparison could either ignore (`.*`) the specific body names, or if you want to test the names then hard-coding the same names as are in our test fixture does not seem too brittle.
___
*[`geometry/optimization/test/iris_test_utilities.h` line 3 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbl1nWA3bY6QGoyUqc:-NXbl1nWA3bY6QGoyUqd:bq87aof) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_test_utilities.h#L3)):*
> ```objc
> #pragma once
> 
> #include <memory>
> ```

nit Unused memory.
___
*[`geometry/optimization/test/iris_test_utilities.h` line 11 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbl2ofAeY115h9zN-q:-NXbl2ofAeY115h9zN-r:byn7uyt) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_test_utilities.h#L11)):*
> ```objc
> #include "drake/multibody/plant/multibody_plant.h"
> #include "drake/planning/robot_diagram_builder.h"
> #include "drake/systems/framework/diagram.h"
> ```

nit Unused scene_graph, multibody_plant, diagram.
___
*[`geometry/optimization/test/iris_test_utilities.h` line 16 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbmdg3AYB-fRvRITlC:-NXbmdg3AYB-fRvRITlD:b-2kbl4v) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/test/iris_test_utilities.h#L16)):*
> ```objc
> namespace geometry {
> namespace optimization {
> /* A toy robot with two revolute joints and some collision geometries.
> ```

This comment doesn't really communicate to the reader what kind of geometry or joint layout is embodied in the test fixture.

On top of that, the implementation sets up the scene using arcane C++ MbP API with tons of boilerplate getting in the way, instead a simpler and more natural SDFormat file (which enables easy visualization by running `model_visualizer` which has joint sliders to see different postures).

I suggest something more like this as a starting point for this fixture:
https://github.com/RobotLocomotion/drake/blob/af983351bd80c43d1a4203cc666020a850e708f0/planning/test/scene_graph_collision_checker_test.cc#L177

The important thing is the use of an ASCII figure and SDFormat text.  Whether the SDFormat model is in a literal string or a separate file is not a big deal either way.  It's easy for the reader to copy the text a separate file if they want to.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/optimization/BUILD.bazel` line 392 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbkPRI9dRDMJby6nIv:-NXctS3u6wtU5Rqzgq5S:bg8cizh) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/BUILD.bazel#L392)):*

nit Missing `visibility = ["//visibility:private"]` here.
</blockquote></details>

Done. I also added `visibility` tag to anoter `testonly` library `test_utilities"
___
*[`geometry/optimization/iris.h` line 198 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbje6gCryAUxgJsfgU:-NXctljm3SN5zLwsSAz6:b-obcopv) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris.h#L198)):*

nit Missing Doxygen docstring. 

... Oh, actually this is just a stray signature isn't it?  The function is not implemented, it seems.
</blockquote></details>

Done. This was a mistake, I removed the function.
___
*[`geometry/optimization/iris_internal.cc` line 177 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbispx5RfO034tRS7p:-NXcv1mL8h3oIRaptLJr:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L177)):*

When a class retains a pointer to an argument passed into its constructor, we typically use a `const CollisionChecker*` pointer (instead of a reference) to help communicate the lifetime requirement to our callers.
</blockquote></details>

Done.
___
*[`geometry/optimization/iris_internal.cc` line 178 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbhg7YAxdezmxOYvvp:-NXcv2n94L4CjqCl-yMc:b-u1813z) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L178)):*

If `influence_distance` is malformed, probably we should fail-fast here instead of later during Eval?
</blockquote></details>

Sorry, what is the requirement on influnce_distance? Is it just >= 0?

In the documentation of CollisionCheck::CalcRobotClearance, it says to compute the potential collision whose distance is less than influence_distance. So what happens if the distance is larger than influence_distance? Is the result not included in `CalcRobotClearance`?
___
*[`geometry/optimization/iris_internal.cc` line 187 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbi4OH9gnB1iw7GdDv:-NXcvj6DAGWZPLE-rZT_:bgz8ubx) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L187)):*

If the two indices are both non-robot ("other"), we need to fail-fast here.  The `CalcRobotClearance` does not report environment-environment clearance distances.  Needs a unit test.
</blockquote></details>

Sorry I don't quite follow, if both indices are non-robot, should IsCollisionFilteredBetween() report true?

jwnimmer-tri (commented):
> Reviewed 3 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/iris.h` line 11 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXerD6_7_CStVu15-3H:-NXerD6_7_CStVu15-3I:b-jqd3z2) ([raw file](https://github.com/RobotLocomotion/drake/blob/3763ce207f684738fc8b17fe2132d368a771d61b/geometry/optimization/iris.h#L11)):*
> ```objc
> #include "drake/geometry/optimization/hpolyhedron.h"
> #include "drake/multibody/plant/multibody_plant.h"
> #include "drake/planning/collision_checker.h"
> ```

nit Unused
___
*[`geometry/optimization/iris_internal.cc` line 178 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbhg7YAxdezmxOYvvp:-NXerb9x6vV4e_ZvdI7V:b5vgquc) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L178)):*
> Sorry, what is the requirement on influnce_distance? Is it just >= 0?

Here's what `collision_checker.cc` checks:

```c++
  DRAKE_THROW_UNLESS(influence_distance >= 0.0);
  DRAKE_THROW_UNLESS(std::isfinite(influence_distance));
```

> So what happens if the distance is larger than influence_distance? Is the result not included in `CalcRobotClearance`?

Yes, exactly.
___
*[`geometry/optimization/iris_internal.cc` line 187 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbi4OH9gnB1iw7GdDv:-NXes8aG9Re3kbscja4O:b8ittgs) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L187)):*

Sorry I don't quite follow, if both indices are non-robot, should IsCollisionFilteredBetween() report true?
</blockquote></details>

Ah, my mistake.  Comment withdrawn.

Env-Env collisions report `IsCollisionFilteredBetween` as `true` already, so the code will already throw.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/iris_internal.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbiJwVBzUXo2tHBcV_:-NXkZRn75Dh8wgJ_gP5i:b-wvkpl2) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L195)):*

I didn't use `MinimumValueConstraint` directly because `MinimumValueConstraint` imposes that the minimum value is <b> larger </b> than certain minimum value. Here we want the `MinimumValueConstraint` to be <b> smaller </b> than certain value.

I would propose that we factor out `MinimumValueConstraint` so that we support the minimum value constraint is bounded on both sides, and then use this modified MinimumValueConstraint. I can work on a separate PR.
</blockquote></details>

For expediency, we could negate the distances though and then "minimum" would apply just fine?

That would let us prove the tougher problem, which is that we might not know a reasonable upper bound on the number of distance (clearance) measurements that could be returned for a given body, to pass to `max_num_values` in the minimum value constraint constructor.  Is `collision_checker_->MaxNumDistances()` a tight enough bound?  In practice there will be perhaps 10 distances in practice, but MaxNumDistances can be in the thousands.

The first discussion point I'd like to agree on is whether we need to smoothly average multiple distances, or if the sharp switching using `std::min` is satisfactory.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/optimization/iris_internal.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbiJwVBzUXo2tHBcV_:-NXkc8_Q4O6kcBnkMiH5:bpcyqni) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L195)):*
> For expediency, we could negate the distances though and then "minimum" would apply just fine?

I would like to work on a separate PR to experiment this with MinimumValueConstraint, that if we have min(v) <= upper_bound instead of min(v) >=lower_bound, will it still work.

I don't think negating the distance would work here, min(-v) does not equal to - min(v), hence simply doing negation doesn't give use the desired result.

> The first discussion point I'd like to agree on is whether we need to smoothly average multiple distances, or if the sharp switching using `std::min` is satisfactory.

My hunch is that the sharp switching in `std::min` will cause problem in SNOPT, although IPOPT might work better.

So here is my plan:
1. Modify MinimumValueConstraint to support smoothed_min(v) <= upper_bound.
2. Use this modified MinimumValueConstraint in InCollisionConstraint.
3. Write down the MathematicalProgram that finds the posture in collision using InCollisionConstraint.
4. Now implement a separate InCollisionConstraintExperimental, which uses the sharp std::min function, and then compare the performance of the MathematicalProgram using InCollisionConstraint versus InCollisionConstraintExperimental.

Each step will have its own PR. What do you think?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/iris_internal.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbiJwVBzUXo2tHBcV_:-NXkeShL7noutxw42Y9v:b-4poarl) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L195)):*

> For expediency, we could negate the distances though and then "minimum" would apply just fine?

I would like to work on a separate PR to experiment this with MinimumValueConstraint, that if we have min(v) <= upper_bound instead of min(v) >=lower_bound, will it still work.

I don't think negating the distance would work here, min(-v) does not equal to - min(v), hence simply doing negation doesn't give use the desired result.

> The first discussion point I'd like to agree on is whether we need to smoothly average multiple distances, or if the sharp switching using `std::min` is satisfactory.

My hunch is that the sharp switching in `std::min` will cause problem in SNOPT, although IPOPT might work better.

So here is my plan:
1. Modify MinimumValueConstraint to support smoothed_min(v) <= upper_bound.
2. Use this modified MinimumValueConstraint in InCollisionConstraint.
3. Write down the MathematicalProgram that finds the posture in collision using InCollisionConstraint.
4. Now implement a separate InCollisionConstraintExperimental, which uses the sharp std::min function, and then compare the performance of the MathematicalProgram using InCollisionConstraint versus InCollisionConstraintExperimental.

Each step will have its own PR. What do you think?
</blockquote></details>

To help answer the question, I don't think we need the full PR process in each step.  A single draft PR with everything in one place (for side-by-side comparsion) seems like a better use of time.

You could even do a hacky `MaxiumumValueConstraint` copy of the other code, just to do the experiment.

The most important thing is that we figure out which math works (which means writing simple & clear code with peer review), not that we push to master right away.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/optimization/iris_internal.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19569#-NXbiJwVBzUXo2tHBcV_:-NXklHVz7jG_nftFj7QR:b-nu4sk) ([raw file](https://github.com/RobotLocomotion/drake/blob/da26f9d658fabeb7c991b20b8219a81198426a20/geometry/optimization/iris_internal.cc#L195)):*

To help answer the question, I don't think we need the full PR process in each step.  A single draft PR with everything in one place (for side-by-side comparsion) seems like a better use of time.

You could even do a hacky `MaxiumumValueConstraint` copy of the other code, just to do the experiment.

The most important thing is that we figure out which math works (which means writing simple & clear code with peer review), not that we push to master right away.
</blockquote></details>

Sounds good, I will file a single large PR then.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19569)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

RussTedrake:
I've only just realized that this PR is IrisInConfigurationSpace (not C-Iris).  I'm in the middle of substantial changes to that workflow -- i thought i'd ask you to wait @hongkai-dai ?

Please let me review tonight before merging anything.

RussTedrake:
Oops.  I see now this one is old.  Hopefully we're on the same page.  Apologies!

hongkai-dai:
I should have closed this PR a while ago. Sorry for the confusion.

PR #19575, opened 2023-06-12
Revise documentation for PathParameterizedTrajectory

RussTedrake:
The notation was a bit imprecise before. `q()` has the domain [t_0, t_f]. We need a different symbol for the path function, which has the domain [s_0, s_f].

This change makes the notation consistent with my notes: https://manipulation.csail.mit.edu/trajectories.html#topp .

+@xuchenhan-tri for both reviews, please.
cc @mpetersen94

### DIFF
--- a/common/trajectories/path_parameterized_trajectory.h
+++ b/common/trajectories/path_parameterized_trajectory.h
@@ -15,8 +15,8 @@ namespace trajectories {
 /**
  * A trajectory defined by a path and timing trajectory.
  *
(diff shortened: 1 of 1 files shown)


xuchenhan-tri (commented):
:lgtm: x2

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19575#-:-NXkJoI9ApAHXiDVZ8vj:bks4n6x)* status: :shipit: complete! all discussions resolved, LGTM from assignee xuchenhan-tri(platform)

PR #13760, opened 2020-07-28
systems: Add PublishInitializationEvents

EricCousineau-TRI:
Resolves #10015

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/13760)

Issue #10015, opened 2018-11-13
systems: Add `DispatchInitializeEvents(System)`


EricCousineau-TRI:
In two places ([show_model](https://github.com/RobotLocomotion/drake/blob/1e01440576e3efba6b6829dda45977f4208615b2/manipulation/util/show_model.py#L93) and [`pydrake/systems/test/test_util_py.cc`](https://github.com/RobotLocomotion/drake/blob/1e01440576e3efba6b6829dda45977f4208615b2/bindings/pydrake/systems/test/test_util_py.cc#L124-L125)), I've used `Simulator.Initializate()` to only dispatch initialization events (and nothing else from `Simulator`), which seems silly.

It would be nice to add this as a free function (or a method of `System`... but meh).

jwnimmer-tri:
Why is it silly?  It does what it says on the tin, and is what you wanted to happen?

EricCousineau-TRI:
Sorry: It's overkill, because I've called it when I don't really need any other functionality from `Simulator`.

(Updated text.)

jwnimmer-tri:
But you _do_ want other things from the simulator -- you want it to do initialization, and then dispatch the `t = t_0` publication event:
```
diagram = builder.Build()
simulator = Simulator(diagram)
simulator.StepTo(0.0)
```

I think its _worse_ to manually call both publishes, don't you?

EricCousineau-TRI:
Hm... Perhaps. But it still feels unclean that our semantics are so complex that you have to instantiate a `Simulator` just to call init + discrete publishes, and rely on the `set_publish_at_initialization` to work (which may or may not be sensitive to #9797).

sherm1:
FYI @edrumwri is improving and clarifying the relationship between Initialize() and Publish() in #9766; `set_publish_at_initializer()` won't be necessary in most cases after that.

I'm not sure why you want a special method for initialization events, especially if you really aren't initializing a simulation. "Events" are things that happen as time moves forward. Can you explain? And do you mean publish events only -- what if the initialization events are discrete or unrestricted updates?

You can see from
[Simulator::Initialize()](https://github.com/RobotLocomotion/drake/blob/master/systems/analysis/simulator.h#L491) that it takes quite a bit of machinery to process events correctly. Are you sure that "event" is the right abstraction for what you're trying to do?

There is `System::Publish()` which can force a call to DoPublish() whenever you want one. I presume that won't work here for some reason?

[...]

PR #19577, opened 2023-06-12
Implements System<T>::ExecuteInitializationEvents()

RussTedrake:
and uses it in MeshcatPoseSliders and JointSliders.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)

### DIFF
--- a/bindings/pydrake/systems/framework_py_systems.cc
+++ b/bindings/pydrake/systems/framework_py_systems.cc
@@ -410,6 +410,9 @@ struct Impl {
         .def("CalcForcedUnrestrictedUpdate",
             &System<T>::CalcForcedUnrestrictedUpdate, py::arg("context"),
             py::arg("state"), doc.System.CalcForcedUnrestrictedUpdate.doc)
+        .def("ExecuteInitializationEvents",
+            &System<T>::ExecuteInitializationEvents, py::arg("context"),
+            doc.System.ExecuteInitializationEvents.doc)
         .def("GetUniquePeriodicDiscreteUpdateAttribute",
             &System<T>::GetUniquePeriodicDiscreteUpdateAttribute,
             doc.System.GetUniquePeriodicDiscreteUpdateAttribute.doc)
--- a/bindings/pydrake/systems/test/custom_test.py
+++ b/bindings/pydrake/systems/test/custom_test.py
@@ -638,6 +638,27 @@ class TestCustom(unittest.TestCase):
         self.assertTrue(system.called_reset)
         self.assertTrue(system.called_system_reset)
 
+        # Test ExecuteInitializationEvents.
+        system = TrivialSystem()
+        context = system.CreateDefaultContext()
+        system.ExecuteInitializationEvents(context=context)
+        self.assertFalse(system.called_per_step)
+        self.assertFalse(system.called_periodic)
+        self.assertTrue(system.called_initialize_publish)
+        self.assertTrue(system.called_initialize_discrete)
+        self.assertTrue(system.called_initialize_unrestricted)
+        self.assertFalse(system.called_periodic_publish)
+        self.assertFalse(system.called_periodic_discrete)
+        self.assertFalse(system.called_periodic_unrestricted)
+        self.assertFalse(system.called_per_step_publish)
+        self.assertFalse(system.called_per_step_discrete)
+        self.assertFalse(system.called_per_step_unrestricted)
+        self.assertFalse(system.called_getwitness)
+        self.assertFalse(system.called_witness)
+        self.assertFalse(system.called_guard)
+        self.assertFalse(system.called_reset)
+        self.assertFalse(system.called_system_reset)
+
     def test_event_handler_returns_none(self):
         """Checks that a Python event handler callback function is allowed to
         (implicitly) return None, instead of an EventStatus. Because of all the
--- a/multibody/meshcat/BUILD.bazel
+++ b/multibody/meshcat/BUILD.bazel
@@ -104,6 +104,7 @@ drake_cc_googletest(
         "//geometry:meshcat_visualizer",
         "//geometry/test_utilities:meshcat_environment",
         "//multibody/parsing",
+        "//systems/framework/test_utilities",
     ],
 )
 
--- a/multibody/meshcat/joint_sliders.cc
+++ b/multibody/meshcat/joint_sliders.cc
@@ -279,6 +279,8 @@ Eigen::VectorXd JointSliders<T>::Run(const Diagram<T>& diagram,
   using Duration = std::chrono::duration<double>;
   const auto start_time = Clock::now();
 
+  diagram.ExecuteInitializationEvents(root_context.get());
+
   // Set the context to the initial slider values.
   plant_->SetPositions(&plant_context,
                        this->get_output_port().Eval(sliders_context));
--- a/multibody/meshcat/joint_sliders.h
+++ b/multibody/meshcat/joint_sliders.h
@@ -120,9 +120,12 @@ class JointSliders final : public systems::LeafSystem<T> {
   value of the plant Context.
 
   @pre `diagram` must be a top-level (i.e., "root") diagram.
-  @pre `diagram` must contain this JointSliders system.
   @pre `diagram` must contain the `plant` that was passed into this
   JointSliders system's constructor.
+  @pre `diagram` must contain this JointSliders system, however the output of
+  these sliders need not be connected (even indirectly) to any `plant` input
+  port. The positions of the `plant` will be updated directly using a call to
+  `plant.SetPositions(...)` when the slider values change.
   */
   Eigen::VectorXd Run(const systems::Diagram<T>& diagram,
                       std::optional<double> timeout = std::nullopt,
--- a/multibody/meshcat/test/joint_sliders_test.cc
+++ b/multibody/meshcat/test/joint_sliders_test.cc
@@ -10,6 +10,7 @@
 #include "drake/multibody/parsing/parser.h"
 #include "drake/multibody/plant/multibody_plant.h"
 #include "drake/systems/framework/diagram_builder.h"
+#include "drake/systems/framework/test_utilities/initialization_test_system.h"
 
 namespace drake {
 namespace multibody {
@@ -282,12 +283,20 @@ TEST_F(JointSlidersTest, Run) {
   MeshcatVisualizer<double>::AddToBuilder(&builder_, scene_graph_, meshcat_);
   auto* dut = builder_.AddSystem<JointSliders<double>>(meshcat_, &plant_,
                                                        initial_value);
+
+  auto init_system = builder_.AddSystem<systems::InitializationTestSystem>();
+
   auto diagram = builder_.Build();
 
   // Run for a while.
   const double timeout = 1.0;
   Eigen::VectorXd q = dut->Run(*diagram, timeout);
   EXPECT_TRUE(CompareMatrices(q, initial_value));
+
+  // Confirm that initialization events were triggered.
+  EXPECT_TRUE(init_system->get_pub_init());
+  EXPECT_TRUE(init_system->get_dis_update_init());
+  EXPECT_TRUE(init_system->get_unres_update_init());
 
   // Note: the stop button is deleted on timeout, so we cannot easily check
   // that it was created correctly here.
--- a/systems/framework/BUILD.bazel
+++ b/systems/framework/BUILD.bazel
@@ -734,7 +734,8 @@ drake_cc_googletest(
         "//common:essential",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -746,7 +747,8 @@ drake_cc_googletest(
         "//common:essential",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -803,7 +805,9 @@ drake_cc_googletest(
         "//common/test_utilities:is_dynamic_castable",
         "//examples/pendulum:pendulum_plant",
         "//systems/analysis/test_utilities:stateless_system",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:initialization_test_system",
+        "//systems/framework/test_utilities:pack_value",
+        "//systems/framework/test_utilities:scalar_conversion",
         "//systems/primitives:adder",
         "//systems/primitives:constant_value_source",
         "//systems/primitives:constant_vector_source",
@@ -822,7 +826,7 @@ drake_cc_googletest(
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
         "//common/test_utilities:is_dynamic_castable",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -863,7 +867,7 @@ drake_cc_googletest(
         ":leaf_context",
         "//common:essential",
         "//common/test_utilities",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -874,7 +878,9 @@ drake_cc_googletest(
         "//common:essential",
         "//common/test_utilities",
         "//common/test_utilities:limit_malloc",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:initialization_test_system",
+        "//systems/framework/test_utilities:my_vector",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -889,7 +895,7 @@ drake_cc_googletest(
     name = "model_values_test",
     deps = [
         ":model_values",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -898,7 +904,7 @@ drake_cc_googletest(
     deps = [
         ":abstract_values",
         "//common:essential",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -907,7 +913,7 @@ drake_cc_googletest(
     deps = [
         ":parameters",
         "//common:essential",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:pack_value",
     ],
 )
 
@@ -941,7 +947,8 @@ drake_cc_googletest(
     deps = [
         ":system_output",
         "//common:essential",
-        "//systems/framework/test_utilities",
+        "//common/test_utilities:is_dynamic_castable",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -972,7 +979,7 @@ drake_cc_googletest(
         ":input_port",
         ":leaf_system",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -984,7 +991,7 @@ drake_cc_googletest(
         "//common:essential",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
         "//systems/primitives:constant_vector_source",
     ],
 )
@@ -1011,7 +1018,7 @@ drake_cc_googletest(
         "//common:unused",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -1030,7 +1037,7 @@ drake_cc_googletest(
         "//common:essential",
         "//common:unused",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:my_vector",
     ],
 )
 
@@ -1048,7 +1055,7 @@ drake_cc_googletest(
     name = "single_output_vector_source_test",
     deps = [
         ":single_output_vector_source",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:scalar_conversion",
     ],
 )
 
@@ -1058,7 +1065,7 @@ drake_cc_googletest(
         ":vector_system",
         "//common/test_utilities:expect_no_throw",
         "//common/test_utilities:expect_throws_message",
-        "//systems/framework/test_utilities",
+        "//systems/framework/test_utilities:scalar_conversion",
         "//systems/primitives:integrator",
     ],
 )
--- a/systems/framework/system.cc
+++ b/systems/framework/system.cc
@@ -479,6 +479,37 @@ void System<T>::GetInitializationEvents(
 }
 
 template <typename T>
+void System<T>::ExecuteInitializationEvents(Context<T>* context) const {
+  auto discrete_updates = AllocateDiscreteVariables();
+  auto state = context->CloneState();
+  auto init_events = AllocateCompositeEventCollection();
+
+  // NOTE: The execution order here must match the code in
+  // Simulator::Initialize().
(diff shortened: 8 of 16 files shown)


jwnimmer-tri (commented):
FYI prior art:
* https://github.com/RobotLocomotion/drake/issues/10015#issuecomment-665282513
* https://github.com/RobotLocomotion/drake/pull/13760

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-:-NXkr9uB4nwNN-JaeRCC:b-xd7csb)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NXkrhcu2xTV12OfDdNR:b79p16p) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*
> ```cpp
> 
>   GetInitializationEvents(*context, init_events.get());
>   if (init_events->get_publish_events().HasEvents()) {
> ```

nit These events are executed in the wrong relative order.

See here:

https://github.com/RobotLocomotion/drake/blob/37d6c59a13431956d82123de6cda24cfa765c50d/systems/analysis/simulator.cc#L84

https://github.com/RobotLocomotion/drake/blob/37d6c59a13431956d82123de6cda24cfa765c50d/systems/analysis/simulator.cc#L233

jwnimmer-tri:
The links were not meant to imply that this API is inappropriate.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NXkyUBY2eOpw_G98ibi:b-t1mek7) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*

Done. Good catch.
</blockquote></details>

There were no test changes in the r2 push though, which means the unit tests are lacking (insufficient to flag this likely implementation defect).

I'll let Sherm weigh in on whether the Simulator implementation just call this new method as a subroutine, which might be a good way to help test-cover it thoroughly.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NXlOg9_8atrjkbybq-6:bwlxvtv) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*

There were no test changes in the r2 push though, which means the unit tests are lacking (insufficient to flag this likely implementation defect).

I'll let Sherm weigh in on whether the Simulator implementation just call this new method as a subroutine, which might be a good way to help test-cover it thoroughly.
</blockquote></details>

My 2 cents: It would be great to call this from Initialize() -- that would provide good testing, avoid duplication, and ensure perfectly matching behavior forever. However, the functionality is subtly different: the Simulator's Initialize has to publish early per-step and periodic events and track some statistics. So combining them probably requires pulling out some common functionality into a gawky private method that could be invoked from both places. Might not be worth it, but in that case a strong warning here that this must match the Initialize() code (esp. the unrestricted/discrete/publish order) would be worthwhile.

(P.S. I'm on vacation this week and will shortly be deprived of net access for a few days in the wilderness of Big Sur.)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NXwDhTVEYImROx5owCj:b-m0udu9) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*

Done.  If #19578 lands then I'm unblocked on my main project.  I don't mind waiting on this PR if need be. 
</blockquote></details>

I think the remaining question is whether we want to insist on a unit test of this function that would detect when the event handlers were processed in the wrong order (e.g., with publishes happening before updates).  I still lean towards "yes" but I'll defer to Sherm for the final word.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NYNwx_yEGSHxQwN-QdW:btxw5zz) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*

I think the remaining question is whether we want to insist on a unit test of this function that would detect when the event handlers were processed in the wrong order (e.g., with publishes happening before updates).  I still lean towards "yes" but I'll defer to Sherm for the final word.
</blockquote></details>

Perhaps the strong warning is enough?  @sherm1 -- welcome back.  waiting for your thoughts.

sherm1 (commented):
Feature :lgtm:
+@jwnimmer-tri for platform review please, since I'm tomorrow's

> Reviewed 7 of 9 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-:-NYObsFoEUJNebfkhEo5:b9wpskh)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`systems/framework/system.h` line 733 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOaARl8jj0WnEVVM0Y:-NYOaARl8jj0WnEVVM0Z:b-7pp1ik) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/system.h#L733)):*
> ```objc
>   GetInitializationEvents(). The method allocates temporary storage to perform
>   the updates, and is intended only as a convenience method for callers who do
>   not want to use the full Simulator workflow. */
> ```

minor: Please add something like: "Note that this is not fully equivalent to Simulator::Initialize() because _only_ initialization events are handled here while Simulator::Initialize() also processes time-0 periodic and per-step events."
___
*[`systems/framework/system.cc` line 488 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NXkrhcu2xTV12OfDdNQ:-NYObBUX-nVe3CZTjpa1:b-f5lng9) ([raw file](https://github.com/RobotLocomotion/drake/blob/5d839e855a692e9aae88cbe7d14d6f9af4109bd3/systems/framework/system.cc#L488)):*

Perhaps the strong warning is enough?  @sherm1 -- welcome back.  waiting for your thoughts.
</blockquote></details>

IMO the small function + warning is enough.

jwnimmer-tri (commented):
> Reviewed 7 of 9 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`multibody/meshcat/joint_sliders.cc` line 282 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOf9ZX1BKOQiZPoGC7:-NYOf9ZYDoAmrpoCgv29:bu062td) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/multibody/meshcat/joint_sliders.cc#L282)):*
> ```cpp
>   const auto start_time = Clock::now();
> 
>   diagram.ExecuteInitializationEvents(root_context.get());
> ```

If I delete this line of code, I think maybe no tests will fail?  If that's true, then we are missing a unit test case on this class.
___
*[`systems/framework/system.cc` line 494 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYP3H6REjJGxNYPG0uB:-NYP3H6REjJGxNYPG0uC:bdv3fq) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/system.cc#L494)):*
> ```cpp
>     CalcUnrestrictedUpdate(*context,
>                            init_events->get_unrestricted_update_events(),
>                            state.get());
> ```

It seems to me like the new (post-initialization) `state` and `discrete_updates` values are complete ignored?

The simulator applies the updates after calculating them, e.g.,

```c++
    system_.ApplyUnrestrictedUpdate(events, unrestricted_updates_.get(),
        context_.get());
```
___
*[`systems/framework/test/leaf_system_test.cc` line 2884 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOhmMN81sqWn41_hSv:-NYOhmMN81sqWn41_hSw:bfdi6n7) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/test/leaf_system_test.cc#L2884)):*
> ```cpp
>   EXPECT_TRUE(dut.get_unres_update_init());
> 
>   // Now again with the ExecuteInitializationEvents method.
> ```

Working

In test cases, having two copies of local variables `foo` vs `foo2` makes it way too easy to accidentally typo one name versus the other, and end up checking the wrong object(s) for their properties, leading to missing test coverage.  Spotting a one-character typo during review is too difficult.

I'll provide a patch that carves this out into a separate test case.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`systems/framework/test/leaf_system_test.cc` line 2884 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOhmMN81sqWn41_hSv:-NYP5B_gAWFXVHU6fF40:b-k3cgq1) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/test/leaf_system_test.cc#L2884)):*

Working

In test cases, having two copies of local variables `foo` vs `foo2` makes it way too easy to accidentally typo one name versus the other, and end up checking the wrong object(s) for their properties, leading to missing test coverage.  Spotting a one-character typo during review is too difficult.

I'll provide a patch that carves this out into a separate test case.
</blockquote></details>

See https://gist.github.com/jwnimmer-tri/b6b7bb9cd3479d59352d1f5909698545.

It just extracts the helper class to be outside of the test function, and splits into two gtests.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`multibody/meshcat/joint_sliders.cc` line 282 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOf9ZX1BKOQiZPoGC7:-NYnyDQ05wYNc2VmIRF5:b-vhsrgn) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/multibody/meshcat/joint_sliders.cc#L282)):*

If I delete this line of code, I think maybe no tests will fail?  If that's true, then we are missing a unit test case on this class.
</blockquote></details>

Done.  I hit the rule of three, so had to touch a few files on this one.
___
*[`systems/framework/system.h` line 733 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOaARl8jj0WnEVVM0Y:-NYnXuTe0N2GmTZM_3et:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/system.h#L733)):*

minor: Please add something like: "Note that this is not fully equivalent to Simulator::Initialize() because _only_ initialization events are handled here while Simulator::Initialize() also processes time-0 periodic and per-step events."
</blockquote></details>

Done.
___
*[`systems/framework/system.cc` line 494 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYP3H6REjJGxNYPG0uB:-NYnz4ng2D7OzzQyyklH:buzttpp) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/system.cc#L494)):*

It seems to me like the new (post-initialization) `state` and `discrete_updates` values are complete ignored?

The simulator applies the updates after calculating them, e.g.,

```c++
    system_.ApplyUnrestrictedUpdate(events, unrestricted_updates_.get(),
        context_.get());
```
</blockquote></details>

Done. Yikes.  Good catch.  Thanks.
___
*[`systems/framework/test/leaf_system_test.cc` line 2884 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYOhmMN81sqWn41_hSv:-NYo-UHn8c7f-Fk848yH:bmfxq1z) ([raw file](https://github.com/RobotLocomotion/drake/blob/0fcaf9b8cb11cf8e944e3448913bafac1b84f3a2/systems/framework/test/leaf_system_test.cc#L2884)):*

See https://gist.github.com/jwnimmer-tri/b6b7bb9cd3479d59352d1f5909698545.

It just extracts the helper class to be outside of the test function, and splits into two gtests.
</blockquote></details>

Done.

sherm1 (commented):
> Reviewed 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`systems/framework/system.h` line 736 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYpDjoG83m1entAIMRu:-NYpDjoH1aiyF6Znnli4:b-h2grnl) ([raw file](https://github.com/RobotLocomotion/drake/blob/0a82dac6e9314e1fcdb2e77f697a811ab2f16ca2/systems/framework/system.h#L736)):*
> ```objc
> 
>   Note that this is not fully equivalent to Simulator::Initialize() because
>   *only* initialization events are handled here, while Simulator::Initialize()
> ```

minor: the `*` at beginning of line confuses Doxygen (it thinks that's one of those archaic comment formatting styles and eats the first `*`:


Use `_only_` instead. I would suggest always using underscores instead of asterisks to avoid this problem -- even if it starts out OK a little reflowing can push the asterisk to BoL later.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`systems/framework/system.h` line 736 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYpDjoG83m1entAIMRu:-NYrtaBMEqfwUSzC-rKI:b-f423z) ([raw file](https://github.com/RobotLocomotion/drake/blob/0a82dac6e9314e1fcdb2e77f697a811ab2f16ca2/systems/framework/system.h#L736)):*

minor: the `*` at beginning of line confuses Doxygen (it thinks that's one of those archaic comment formatting styles and eats the first `*`:


Use `_only_` instead. I would suggest always using underscores instead of asterisks to avoid this problem -- even if it starts out OK a little reflowing can push the asterisk to BoL later.
</blockquote></details>

Done.  Thanks for catching it.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri (commented):
:lgtm:

> Reviewed 11 of 12 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-:-NYt0KoQ2EJzEFZnk8xN:bnfp4nl)* status: 5 unresolved discussions

___
*[`systems/framework/system.h` line 738 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsx9AZ-DP9Uvq00q5t:-NYsx9AZ-DP9Uvq00q5u:bxon09w) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/system.h#L738)):*
> ```objc
>   _only_ initialization events are handled here, while Simulator::Initialize()
>   also processes other events associated with time zero.
>   */
> ```

nit This file places `*/` at the end of line, not on a fresh line.



_Suggestion:_
```objc
  also processes other events associated with time zero. */
```
___
*[`systems/framework/test/leaf_system_test.cc` line 2850 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYt-Phq6UYYbKDSGiFJ:-NYt-Phq6UYYbKDSGiFK:b9q77xp) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test/leaf_system_test.cc#L2850)):*
> ```cpp
>   EXPECT_TRUE(dut.get_dis_update_init());
>   EXPECT_TRUE(dut.get_unres_update_init());
>   EXPECT_TRUE(context->get_discrete_state_vector()[0]);
> ```

Checking the true/false truth value of the discrete state floating-point value does not clearly communicate what's happening.



_Suggestion:_
```cpp
  EXPECT_EQ(context->get_discrete_state_vector()[0], 1.23);
```
___
*[`systems/framework/test/leaf_system_test.cc` line 2851 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYt-ZNRCAIp_ecoHZ3-:-NYt-ZNRCAIp_ecoHZ30:b62jfu3) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test/leaf_system_test.cc#L2851)):*
> ```cpp
>   EXPECT_TRUE(dut.get_unres_update_init());
>   EXPECT_TRUE(context->get_discrete_state_vector()[0]);
>   EXPECT_NE(context->get_abstract_state<bool>(0), 0.0);
> ```

Comparing `true` to `!= 0.0` does not clearly communicate what's happening.



_Suggestion:_
```cpp
EXPECT_TRUE(context->get_abstract_state<bool>(0));
```
___
*[`systems/framework/test_utilities/BUILD.bazel` line 29 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsy7Kv3e_s2kDNRZ3R:-NYsy7Kw11Mfh8MWUXEu:b-6g5mkc) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test_utilities/BUILD.bazel#L29)):*
> ```Starlark
>     hdrs = ["initialization_test_system.h"],
>     deps = [
>         "//systems/framework",
> ```

For edit-test responsiveness, we shouldn't add unnecessary dependencies to our framework package's own unit tests.  Prior to this PR, the utilities library depended only practically no framework code.  Now it depends on all of it.

To resolve this, please edit `drake/systems/framework/BUILD.bazel` so that instead of using the roll-up `"//systems/framework/test_utilities"` anywhere, it only depends on the exact components it needs.

Note that the roll-up `"//systems/framework/test_utilities"` is fine for other packages to use (e.g., visualization) since they are already pulling in the whole framework.  It's only the framework's own BUILD file where this matters.
___
*[`systems/framework/test_utilities/initialization_test_system.h` line 10 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsxvZeAp54qN9C-oim:-NYsxvZeAp54qN9C-oin:b-h8gzq) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test_utilities/initialization_test_system.h#L10)):*
> ```objc
> namespace systems {
> 
> class InitializationTestSystem : public LeafSystem<double> {
> ```

nit Now that this is project-wide, it needs a class overview API doc summarizing what it accomplishes.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: 1 unresolved discussion

___
*[`systems/framework/system.h` line 738 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsx9AZ-DP9Uvq00q5t:-NYxrmMP3_Ol9E3gl5S5:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/system.h#L738)):*

nit This file places `*/` at the end of line, not on a fresh line.
</blockquote></details>

Done
___
*[`systems/framework/test/leaf_system_test.cc` line 2850 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYt-Phq6UYYbKDSGiFJ:-NYxs3Ti2WJOS1k8Uznz:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test/leaf_system_test.cc#L2850)):*

Checking the true/false truth value of the discrete state floating-point value does not clearly communicate what's happening.
</blockquote></details>

Done.
___
*[`systems/framework/test/leaf_system_test.cc` line 2851 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYt-ZNRCAIp_ecoHZ3-:-NYxs3vQ4vWxyfAA80Qu:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test/leaf_system_test.cc#L2851)):*

Comparing `true` to `!= 0.0` does not clearly communicate what's happening.
</blockquote></details>

Done.
___
*[`systems/framework/test_utilities/BUILD.bazel` line 29 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsy7Kv3e_s2kDNRZ3R:-NYxvpR40fbklV2AcsEm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test_utilities/BUILD.bazel#L29)):*

For edit-test responsiveness, we shouldn't add unnecessary dependencies to our framework package's own unit tests.  Prior to this PR, the utilities library depended only practically no framework code.  Now it depends on all of it.

To resolve this, please edit `drake/systems/framework/BUILD.bazel` so that instead of using the roll-up `"//systems/framework/test_utilities"` anywhere, it only depends on the exact components it needs.

Note that the roll-up `"//systems/framework/test_utilities"` is fine for other packages to use (e.g., visualization) since they are already pulling in the whole framework.  It's only the framework's own BUILD file where this matters.
</blockquote></details>

Done.
___
*[`systems/framework/test_utilities/initialization_test_system.h` line 10 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19577#-NYsxvZeAp54qN9C-oim:-NYxsvn11vZnsE0P-Pj2:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0b4c8e10ffe5d5d90e506758df7d5aca7c894aa1/systems/framework/test_utilities/initialization_test_system.h#L10)):*

nit Now that this is project-wide, it needs a class overview API doc summarizing what it accomplishes.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19577)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform)

PR #19591, opened 2023-06-14
[math] Add soft min and max function

hongkai-dai:
Support soft min or max being either the under or over approximation of min/max.

This will be used later when I impose the minimum distance constraint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)

### DIFF
--- a/math/BUILD.bazel
+++ b/math/BUILD.bazel
@@ -31,6 +31,7 @@ drake_cc_package_library(
         ":linear_solve",
         ":matrix_util",
         ":quadratic_form",
+        ":soft_min_max",
         ":vector3_util",
         ":wrap_to",
     ],
@@ -261,6 +262,17 @@ drake_cc_library(
         ":matrix_util",
         "//common:essential",
         "@eigen",
+    ],
+)
+
+drake_cc_library(
+    name = "soft_min_max",
+    srcs = ["soft_min_max.cc"],
+    hdrs = ["soft_min_max.h"],
+    deps = [
+        "//common:autodiff",
+        "//common:default_scalars",
+        "//common:essential",
     ],
 )
 
@@ -551,6 +563,15 @@ drake_cc_googletest(
         ":autodiff",
         ":geometric_transform",
         "//common/test_utilities",
+    ],
+)
+
+drake_cc_googletest(
+    name = "soft_min_max_test",
+    deps = [
+        ":soft_min_max",
+        "//common/test_utilities:eigen_matrix_compare",
+        "@eigen",
     ],
 )
 
--- a/math/soft_min_max.cc
+++ b/math/soft_min_max.cc
@@ -1,108 +1,88 @@
 #include "drake/math/soft_min_max.h"
 
+#include <algorithm>
+#include <cmath>
+
 #include "drake/common/default_scalars.h"
+#include "drake/common/drake_throw.h"
 
 namespace drake {
 namespace math {
 namespace {
-/** Computes log(∑exp(xᵢ)). Exploits the shift invariance of log-sum-exp to
-avoid overflow. */
+using std::exp;
+using std::isfinite;
+using std::log;
+
+/* We use tare_x to exploit the shift invariance of log-sum-exp to avoid
+overflow. It should be set to the extremum of x (i.e., max when alpha > 0,
+min when alpha < 0). */
 template <typename T>
-T LogSumExp(const std::vector<T>& x) {
-  DRAKE_ASSERT(x.size() > 0);
-  using std::exp;
-  using std::log;
-  const T x_max = *std::max_element(x.begin(), x.end());
-  T sum_exp{0.0};
+T SoftOverMaxImpl(const std::vector<T>& x, double tare_x, double alpha) {
+  T sum_exp{0};
   for (const T& xi : x) {
-    sum_exp += exp(xi - x_max);
+    sum_exp += exp(alpha * (xi - tare_x));
   }
-  return x_max + log(sum_exp);
+  return log(sum_exp) / alpha + tare_x;
+}
+
+/* We use tare_x to avoid overflow. It should be set to the extremum of x
+(i.e., max when alpha > 0, min when alpha < 0). */
+template <typename T>
+T SoftUnderMaxImpl(const std::vector<T>& x, double tare_x, double alpha) {
+  T soft_max{0};
+  T sum_exp{0};
+  for (const T& xi : x) {
+    T exp_xi = exp(alpha * (xi - tare_x));
+    sum_exp += exp_xi;
+    exp_xi *= xi;
+    soft_max += exp_xi;
+  }
+  return soft_max / sum_exp;
 }
 }  // namespace
 
-/**
- Computes a smooth over approximation of max function, namely SoftOverMax(x) >=
- max(x).
- Mathematically we compute this as (log (∑ᵢ exp(αxᵢ))) / α.
- @param x The vector we want to compute its soft max.
- @param alpha α in the documentation above. Larger α makes the soft max more
- similar to max, with a sharper corner.
- @throws std::exception if α <= 0.
- */
 template <typename T>
-[[nodiscard]] T SoftOverMax(const std::vector<T>& x, double alpha) {
-  DRAKE_ASSERT(x.size() > 0);
+T SoftOverMax(const std::vector<T>& x, double alpha) {
+  DRAKE_THROW_UNLESS(x.size() > 0);
   DRAKE_THROW_UNLESS(alpha > 0);
-  if (alpha == 1.0) {
-    return LogSumExp(x);
-  }
-  std::vector<T> x_scaled{x};
-  for (T& xi_scaled : x_scaled) {
-    xi_scaled *= alpha;
-  }
-  return LogSumExp(x_scaled) / alpha;
+  DRAKE_THROW_UNLESS(std::isfinite(alpha));
+  const double max_x =
+      ExtractDoubleOrThrow(*std::max_element(x.begin(), x.end()));
+  return SoftOverMaxImpl(x, max_x, alpha);
 }
 
-/**
- Computes a smooth under approximation of max function, namely SoftUnderMax(x)
- <= max(x). Mathematically we compute this as ∑ᵢ exp(αxᵢ)*xᵢ) / d, where d = ∑ⱼ
- exp(αxⱼ)
- @param x The vector we want to compute its soft max.
- @param alpha α in the documentation above. Larger α makes the soft max more
- similar to max, with a sharper corner.
- @throws std::exception if α <= 0.
- */
 template <typename T>
-[[nodiscard]] T SoftUnderMax(const std::vector<T>& x, double alpha) {
-  DRAKE_ASSERT(x.size() > 0);
+T SoftUnderMax(const std::vector<T>& x, double alpha) {
+  DRAKE_THROW_UNLESS(x.size() > 0);
   DRAKE_THROW_UNLESS(alpha > 0);
-  std::vector<T> x_scaled{x};
-  for (auto& xi_scaled : x_scaled) {
-    xi_scaled *= alpha;
-  }
-  // To avoid overflow, we subtract x_scaled_max.
-  const T x_scaled_max = *std::max_element(x_scaled.begin(), x_scaled.end());
-  T soft_max{0};
-  T d{0};
-  for (int i = 0; i < static_cast<int>(x.size()); ++i) {
-    using std::exp;
-    const T exp_xi = exp(x_scaled[i] - x_scaled_max);
-    soft_max += exp_xi * x[i];
-    d += exp_xi;
-  }
-  return soft_max / d;
-}
-
-/**
- Computes a smooth over approximation of max function, namely SoftOverMin(x) >=
- min(x).
- Mathematically we compute this as ∑ᵢ exp(-αxᵢ)*xᵢ) / d, where d = ∑ⱼ
- exp(-αxⱼ)
- @param x The vector we want to compute its soft min.
- @param alpha α in the documentation above. Larger α makes the soft min more
- similar to min, with a sharper corner.
- @throws std::exception if α <= 0.
- */
-template <typename T>
-[[nodiscard]] T SoftOverMin(const std::vector<T>& x, double alpha) {
-  // min(x) = -max(-x)
-  std::vector<T> x_negate{x};
-  for (auto& xi : x_negate) {
-    xi *= -1;
-  }
-  return -SoftUnderMax(x_negate, alpha);
+  DRAKE_THROW_UNLESS(std::isfinite(alpha));
+  const double max_x =
+      ExtractDoubleOrThrow(*std::max_element(x.begin(), x.end()));
+  return SoftUnderMaxImpl(x, max_x, alpha);
 }
 
 template <typename T>
-[[nodiscard]] T SoftUnderMin(const std::vector<T>& x, double alpha) {
+T SoftOverMin(const std::vector<T>& x, double alpha) {
+  DRAKE_THROW_UNLESS(x.size() > 0);
+  DRAKE_THROW_UNLESS(alpha > 0);
+  DRAKE_THROW_UNLESS(std::isfinite(alpha));
   // min(x) = -max(-x)
-  std::vector<T> x_negate{x};
-  for (auto& xi : x_negate) {
-    xi *= -1;
-  }
-  return -SoftOverMax(x_negate, alpha);
+  const double min_x =
+      ExtractDoubleOrThrow(*std::min_element(x.begin(), x.end()));
+  return SoftUnderMaxImpl(x, min_x, -alpha);
 }
+
+template <typename T>
+T SoftUnderMin(const std::vector<T>& x, double alpha) {
+  DRAKE_THROW_UNLESS(x.size() > 0);
+  DRAKE_THROW_UNLESS(alpha > 0);
+  DRAKE_THROW_UNLESS(std::isfinite(alpha));
+  // min(x) = -max(-x)
+  const double min_x =
+      ExtractDoubleOrThrow(*std::min_element(x.begin(), x.end()));
+  return SoftOverMaxImpl(x, min_x, -alpha);
+}
+
 // Explicit instantiation
 DRAKE_DEFINE_FUNCTION_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_NONSYMBOLIC_SCALARS(
     (&SoftOverMax<T>, &SoftUnderMax<T>, &SoftOverMin<T>, &SoftUnderMin<T>))
--- a/math/soft_min_max.cc
+++ b/math/soft_min_max.cc
@@ -1,0 +1,110 @@
+#include "drake/math/soft_min_max.h"
+
+#include "drake/common/default_scalars.h"
+
+namespace drake {
+namespace math {
+namespace {
+/** Computes log(∑exp(xᵢ)). Exploits the shift invariance of log-sum-exp to
+avoid overflow. */
+template <typename T>
+T LogSumExp(const std::vector<T>& x) {
+  DRAKE_ASSERT(x.size() > 0);
+  using std::exp;
+  using std::log;
+  const T x_max = *std::max_element(x.begin(), x.end());
+  T sum_exp{0.0};
+  for (const T& xi : x) {
+    sum_exp += exp(xi - x_max);
+  }
+  return x_max + log(sum_exp);
+}
+}  // namespace
+
+/**
+ Computes a smooth over approximation of max function, namely SoftOverMax(x) >=
+ max(x).
+ Mathematically we compute this as (log (∑ᵢ exp(αxᵢ))) / α.
+ @param x The vector we want to compute its soft max.
+ @param alpha α in the documentation above. Larger α makes the soft max more
+ similar to max, with a sharper corner.
+ @throws std::exception if α <= 0.
+ */
+template <typename T>
+[[nodiscard]] T SoftOverMax(const std::vector<T>& x, double alpha) {
+  DRAKE_ASSERT(x.size() > 0);
+  DRAKE_THROW_UNLESS(alpha > 0);
+  if (alpha == 1.0) {
+    return LogSumExp(x);
+  }
+  std::vector<T> x_scaled{x};
+  for (T& xi_scaled : x_scaled) {
+    xi_scaled *= alpha;
+  }
+  return LogSumExp(x_scaled) / alpha;
+}
+
+/**
+ Computes a smooth under approximation of max function, namely SoftUnderMax(x)
+ <= max(x). Mathematically we compute this as ∑ᵢ exp(αxᵢ)*xᵢ) / d, where d = ∑ⱼ
+ exp(αxⱼ)
+ @param x The vector we want to compute its soft max.
+ @param alpha α in the documentation above. Larger α makes the soft max more
+ similar to max, with a sharper corner.
+ @throws std::exception if α <= 0.
+ */
+template <typename T>
+[[nodiscard]] T SoftUnderMax(const std::vector<T>& x, double alpha) {
+  DRAKE_ASSERT(x.size() > 0);
+  DRAKE_THROW_UNLESS(alpha > 0);
+  std::vector<T> x_scaled{x};
+  for (auto& xi_scaled : x_scaled) {
+    xi_scaled *= alpha;
+  }
+  // To avoid overflow, we subtract x_scaled_max.
+  const T x_scaled_max = *std::max_element(x_scaled.begin(), x_scaled.end());
+  T soft_max{0};
+  T d{0};
+  for (int i = 0; i < static_cast<int>(x.size()); ++i) {
+    using std::exp;
+    const T exp_xi = exp(x_scaled[i] - x_scaled_max);
+    soft_max += exp_xi * x[i];
+    d += exp_xi;
+  }
+  return soft_max / d;
+}
+
+/**
+ Computes a smooth over approximation of max function, namely SoftOverMin(x) >=
+ min(x).
+ Mathematically we compute this as ∑ᵢ exp(-αxᵢ)*xᵢ) / d, where d = ∑ⱼ
+ exp(-αxⱼ)
+ @param x The vector we want to compute its soft min.
+ @param alpha α in the documentation above. Larger α makes the soft min more
+ similar to min, with a sharper corner.
+ @throws std::exception if α <= 0.
+ */
+template <typename T>
+[[nodiscard]] T SoftOverMin(const std::vector<T>& x, double alpha) {
+  // min(x) = -max(-x)
+  std::vector<T> x_negate{x};
+  for (auto& xi : x_negate) {
+    xi *= -1;
+  }
+  return -SoftUnderMax(x_negate, alpha);
+}
+
+template <typename T>
+[[nodiscard]] T SoftUnderMin(const std::vector<T>& x, double alpha) {
+  // min(x) = -max(-x)
+  std::vector<T> x_negate{x};
+  for (auto& xi : x_negate) {
+    xi *= -1;
+  }
+  return -SoftOverMax(x_negate, alpha);
+}
+// Explicit instantiation
+DRAKE_DEFINE_FUNCTION_TEMPLATE_INSTANTIATIONS_ON_DEFAULT_NONSYMBOLIC_SCALARS(
+    (&SoftOverMax<T>, &SoftUnderMax<T>, &SoftOverMin<T>, &SoftUnderMin<T>))
+}  // namespace math
+}  // namespace drake
--- a/math/soft_min_max.cc
+++ b/math/soft_min_max.cc
@@ -14,35 +14,41 @@ using std::isfinite;
 using std::log;
 
 /* We use tare_x to exploit the shift invariance of log-sum-exp to avoid
-overflow. It should be set to the extremum of x (i.e., max when alpha > 0,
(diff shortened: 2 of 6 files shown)


EricCousineau-TRI:
Howdy Hongkai! Per platform checklist: This PR does not yet have a reviewer assigned. Is it ready for review yet? If yes, then please assign a feature reviewer. If not, then please label it “status: do not review”.

jwnimmer-tri:
@hongkai-dai If you can get the math, docs, and test cases to the place where you like it, after that I can volunteer to join in and push fixes for C++ style details.

hongkai-dai:
@jwnimmer-tri do you mean that I finish all the implementations to support IrisInConfigurationSpace with CollisionChecker in one branch, and then we break the large change into small PRS? Or do you mean that for this small PR you can handle the C++ details?

jwnimmer-tri:
I just mean for this PR.  I would move the code to cc files with explicit template instantiations, adjust the types of the arguments to allow for more than just `std::vector` with extra copying, etc.

hongkai-dai (commented):
Thanks Jeremy.

I think I am satisfied with the math, docs and tests.

May I ask why you want to move the code to cc files with explicit template instantiation? Is it for compilation speed? If we do explicit instantiation, then we will only support double and AutoDiffXd, but we can't support other types like `AutoDiffScalar<Vector1d> `, `AutoDiffScalar<Vector2d>`, `AutoDiffScalar<Vector3d>` etc.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NXwOmLx3-w514p_MOwd:blxidtk)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai:
> Do we have any code that would call this with such types? Could you point me to it?

No we don't have these types, but I thought it would be nice to support them in the future, as we used to support them in the old version of Drake.

I can move the code to the cc file with explicit instantiation, and then maybe you can help replace the container type from `std::vector`? Thanks a lot!

hongkai-dai (commented):
@jwnimmer-tri I moved the code to the cc file with the explicit instantiation, could you help replace the container type? Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NXw_8BaBfXqZ7TJb1XP:b-ctv0rs)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 39 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-K-QS8BHDSCp5vu_x:-NY-K-QS8BHDSCp5vu_y:b-bj4jk4) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.h#L39)):*
> ```objc
>  Computes a smooth over approximation of max function, namely SoftOverMin(x) >=
>  min(x).
>  Mathematically we compute this as ∑ᵢ exp(-αxᵢ)*xᵢ) / d, where d = ∑ⱼ
> ```

nit I think this is missing a negative sign, compared to the implementation?



_Suggestion:_
```objc
∑ᵢ exp(-αxᵢ)*-xᵢ) / d
```
___
*[`math/soft_min_max.cc` line 15 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-amqABogK6yLEzw6l:-NY-amqABogK6yLEzw6m:b-n7x0a8) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.cc#L15)):*
> ```cpp
>   using std::exp;
>   using std::log;
>   const T x_max = *std::max_element(x.begin(), x.end());
> ```

Have you thought through whether the biasing term `x_max` should keep its gradient or not?  (That is, should x_max be a `double` or a `T`?)  I can try to pencil out the math, but maybe you already have done it?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 39 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-K-QS8BHDSCp5vu_x:-NY0MkFEDo9dOpDz9oOz:bjav23g) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.h#L39)):*

nit I think this is missing a negative sign, compared to the implementation?
</blockquote></details>

I think the documentation is right.

The soft-max function is `∑ᵢ exp(αxᵢ)*xᵢ) / d`. The soft min is -SoftMax(-x), hence it is ∑ᵢ exp(-αxᵢ)*xᵢ) / d
___
*[`math/soft_min_max.cc` line 15 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-amqABogK6yLEzw6l:-NY0N4CV39YHwOFCY655:b-2e48ic) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.cc#L15)):*

Have you thought through whether the biasing term `x_max` should keep its gradient or not?  (That is, should x_max be a `double` or a `T`?)  I can try to pencil out the math, but maybe you already have done it?
</blockquote></details>

I don't think it matters in this case. 

The shift invariant property `LogSumExp(x) = LogSumExp(x - c) + c` holds for any constant `c`. So we could drop the gradient in `x_max` and treat it as a constant. But the property also holds if we treat `c` as any function of `x`, you could prove that
```
∇ₓ( log ∑ᵢ exp(xᵢ − c(x))  + c(x)) = ∇ₓ( log ∑ᵢ exp(xᵢ))
```
for any function `c(x)` (which is max(x) in our case).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 39 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-K-QS8BHDSCp5vu_x:-NY0Pmzs5q0GMf7uPy2s:bamhl2) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.h#L39)):*

I think the documentation is right.

The soft-max function is `∑ᵢ exp(αxᵢ)*xᵢ) / d`. The soft min is -SoftMax(-x), hence it is ∑ᵢ exp(-αxᵢ)*xᵢ) / d
</blockquote></details>

Oops, you're right of course.  I'd forgotten about the `-` in `-SoftMax`.
___
*[`math/soft_min_max.cc` line 15 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-amqABogK6yLEzw6l:-NY0PvcF2qO9Qe-txCZV:bi5yba3) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.cc#L15)):*

I don't think it matters in this case. 

The shift invariant property `LogSumExp(x) = LogSumExp(x - c) + c` holds for any constant `c`. So we could drop the gradient in `x_max` and treat it as a constant. But the property also holds if we treat `c` as any function of `x`, you could prove that
```
∇ₓ( log ∑ᵢ exp(xᵢ − c(x))  + c(x)) = ∇ₓ( log ∑ᵢ exp(xᵢ))
```
for any function `c(x)` (which is max(x) in our case).
</blockquote></details>

Ah.  I wasn't sure if the `x_max` partials were helping smooth things out.  If not, I definitely think we should `ExtractDoubleOrThrow` so that `x_max` can be `double`.  That will be substantially faster as we calculate inside the looping.  I'll make that change as part of my edits.

hongkai-dai (commented):
Thanks @jwnimmer-tri for the fix.

+@calderpg-tri for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NY1MP1351ZYWmgAYs7w:brcmlgo)* status: 1 unresolved discussion, LGTM missing from assignees jwnimmer-tri(platform),calderpg-tri, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 2 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),calderpg-tri, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 14 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY3LI2b2VsOgnclvJLG:-NY3LI2c4ksSCfF0oar8:b-pfkjhd) ([raw file](https://github.com/RobotLocomotion/drake/blob/72a5542c73caa9419d57ae2b075ce05b382439dc/math/soft_min_max.h#L14)):*
> ```objc
>  @param alpha α in the documentation above. Larger α makes the soft max more
>  similar to max, with a sharper corner.
>  @throws std::exception if α <= 0.
> ```

nit Maybe we should explain that alpha must be finite.

To me, documenting bounds on argument docs is clearer than `@throws`, so here is how I would do it.

Ditto for all four functions.



_Suggestion:_
```objc
 @param alpha α in the documentation above. Larger α makes the soft max more
 similar to max, with a sharper corner. Must be strictly positive and finite.
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 3 unresolved discussions, LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 21 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY3VFNO1t_5YS_zmFJF:-NY3VFNO1t_5YS_zmFJG:bhjj6ka) ([raw file](https://github.com/RobotLocomotion/drake/blob/72a5542c73caa9419d57ae2b075ce05b382439dc/math/soft_min_max.h#L21)):*
> ```objc
> /**
>  Computes a smooth under approximation of max function, namely SoftUnderMax(x)
>  <= max(x). Mathematically we compute this as ∑ᵢ exp(αxᵢ)*xᵢ / d, where d = ∑ⱼ
> ```

nit To me at least, it's not obvious that the math in the comment here is sufficiently descriptive of what happens in the implementation.  In the implementation, we shift x so that all xᵢ are non-positive, so that when `exp` is used as a scaling factor it is always <= 1.  If we skip the shifting I'm not convinced that we're still "under", nor that the contour of the gradient is the same?

```
∑ᵢ exp(αxᵢ-m)*xᵢ / ∑ⱼ exp(αxⱼ-m)
where m = max(x)
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 2 unresolved discussions, LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 21 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY3VFNO1t_5YS_zmFJF:-NY5OxOKD1fiHNW82fux:b-jz5z7z) ([raw file](https://github.com/RobotLocomotion/drake/blob/72a5542c73caa9419d57ae2b075ce05b382439dc/math/soft_min_max.h#L21)):*

nit To me at least, it's not obvious that the math in the comment here is sufficiently descriptive of what happens in the implementation.  In the implementation, we shift x so that all xᵢ are non-positive, so that when `exp` is used as a scaling factor it is always <= 1.  If we skip the shifting I'm not convinced that we're still "under", nor that the contour of the gradient is the same?

```
∑ᵢ exp(αxᵢ-m)*xᵢ / ∑ⱼ exp(αxⱼ-m)
where m = max(x)
```
</blockquote></details>

The shifting is just to avoid numerical overflow. The operation we have is shift invariant.

We could prove that
```
∑ᵢ exp(αxᵢ-m)*xᵢ / ∑ⱼ exp(αxⱼ-m) =∑ᵢ exp(αxᵢ)*xᵢ / ∑ⱼ exp(αxⱼ)
```
for any `m`. It is equivalent to multiplying exp(m) simultaneously on both the numerator and denominator.
___
*[`math/soft_min_max.cc` line 15 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY-amqABogK6yLEzw6l:-NY5PQbOAkzcQasp4a1O:bo0c70i) ([raw file](https://github.com/RobotLocomotion/drake/blob/60456bc13cc6b6202b11f6160826b985ce992d17/math/soft_min_max.cc#L15)):*

Ah.  I wasn't sure if the `x_max` partials were helping smooth things out.  If not, I definitely think we should `ExtractDoubleOrThrow` so that `x_max` can be `double`.  That will be substantially faster as we calculate inside the looping.  I'll make that change as part of my edits.
</blockquote></details>

Thanks Jeremy.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 21 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NY3VFNO1t_5YS_zmFJF:-NY5Q8zj1VmQGbl82mz2:bjvzlhm) ([raw file](https://github.com/RobotLocomotion/drake/blob/72a5542c73caa9419d57ae2b075ce05b382439dc/math/soft_min_max.h#L21)):*

The shifting is just to avoid numerical overflow. The operation we have is shift invariant.

We could prove that
```
∑ᵢ exp(αxᵢ-m)*xᵢ / ∑ⱼ exp(αxⱼ-m) =∑ᵢ exp(αxᵢ)*xᵢ / ∑ⱼ exp(αxⱼ)
```
for any `m`. It is equivalent to multiplying exp(m) simultaneously on both the numerator and denominator.
</blockquote></details>

Ah, of course!  What made it click for me was _"It is equivalent to multiplying exp(m) simultaneously on both the numerator and denominator."_

Your full explanation above (all the sentences and the formula) would probably be helpful to add into the cc file comments to help future readers.  (I think the header file is okay as-is.)

calderpg-tri (commented):
> Reviewed 2 of 6 files at r1, 2 of 4 files at r3, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 5 unresolved discussions, LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.cc` line 22 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPcbD8A8v8bL4KMKJU:-NYPcbD8A8v8bL4KMKJV:b5c2uo1) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/soft_min_max.cc#L22)):*
> ```cpp
> T SoftOverMaxImpl(const std::vector<T>& x, double tare_x, double alpha) {
>   // Note that the function
>   // f(x) = log(∑ᵢ exp(αxᵢ) /α satisfies f(x) = f(x - m) + m for any arbitrary
> ```

minor missing a closing parenthesis, should be `log(∑ᵢ exp(αxᵢ)) / α`
___
*[`math/soft_min_max.cc` line 42 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPe1ElCWgO3dO0RWHX:-NYPe1ElCWgO3dO0RWHY:b-x1m5pw) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/soft_min_max.cc#L42)):*
> ```cpp
>   T sum_exp{0};
>   for (const T& xi : x) {
>     T exp_xi = exp(alpha * (xi - tare_x));
> ```

It seems to me that this would be clearer written as
```
const T exp_xi = exp(alpha * (xi - tare_x));
sum_exp += exp_xi;
soft_max += exp_xi * xi;
```
versus mutating `exp_xi`. Is there a performance reason to prefer mutation for autodiff?
___
*[`math/soft_min_max.cc` line 52 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPZjXuBT4fvoyb8T9D:-NYPZjXvEuMl182_I2k0:bfzdu0b) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/soft_min_max.cc#L52)):*
> ```cpp
> 
> template <typename T>
> T SoftOverMax(const std::vector<T>& x, double alpha) {
> ```

nit here and below,`double alpha` could be `const double alpha`
___
*[`math/test/soft_min_max_test.cc` line 10 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPgRj0DQml7vP2aMtn:-NYPgRj0DQml7vP2aMto:b5gvfkv) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/test/soft_min_max_test.cc#L10)):*
> ```cpp
> namespace drake {
> namespace math {
> GTEST_TEST(SoftOverMax, TestDouble) {
> ```

In general it seems a bit arbitrary which checks are contained in each test, and what tolerances are provided - e.g. some have `EXPECT_NEAR` (or equivalent via `CompareMatrices`), some don't
___
*[`math/test/soft_min_max_test.cc` line 38 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPfUyk-hWGp93sLMvl:-NYPfUyk-hWGp93sLMvm:bc29zsn) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/test/soft_min_max_test.cc#L38)):*
> ```cpp
>   EXPECT_NEAR(x_max.value(), 3, 0.01);
>   EXPECT_TRUE(
>       CompareMatrices(x_max.derivatives(), Eigen::Vector2d(0, -1), 1e-3));
> ```

minor there is a mix of styles when specifying tolerances (e.g. `1E-2`, `1e-3`, and `1E-5`), prefer using one or the other for consistency.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 2 unresolved discussions, LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.cc` line 42 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPe1ElCWgO3dO0RWHX:-NYRCSBx-60Zlfu9ry4T:btzgqzh) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/soft_min_max.cc#L42)):*

It seems to me that this would be clearer written as
```
const T exp_xi = exp(alpha * (xi - tare_x));
sum_exp += exp_xi;
soft_max += exp_xi * xi;
```
versus mutating `exp_xi`. Is there a performance reason to prefer mutation for autodiff?
</blockquote></details>

Done. Agreed.
___
*[`math/test/soft_min_max_test.cc` line 10 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPgRj0DQml7vP2aMtn:-NYRCnEqDfdwlEG9X55g:b-7sbuvr) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/test/soft_min_max_test.cc#L10)):*

In general it seems a bit arbitrary which checks are contained in each test, and what tolerances are provided - e.g. some have `EXPECT_NEAR` (or equivalent via `CompareMatrices`), some don't
</blockquote></details>

The pattern is that in the `TestDouble`, we test with different alpha, and make sure that as alpha increases the soft min/max gets closer to hard min/max. In TestAutoDiff we only test with a single alpha, and make sure that the soft min/max, together with its gradient are correct.

Some of the test uses `EXPECT_LT` or `EXPECT_GT` to make sure that the soft min/max is an over (or under) approximation. But we still want to make sure it is an approximation, hence use `EXPECT_NEAR`.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: 1 unresolved discussion, LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/test/soft_min_max_test.cc` line 10 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYPgRj0DQml7vP2aMtn:-NYTohRX5seRfDt55MVk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6089edca4775a3e4906b05f0d1ac2a57324690e7/math/test/soft_min_max_test.cc#L10)):*

Following that pattern, wouldn't it make sense to have all of the `alpha = 10` cases include an `EXPECT_NEAR`? Having it only for `SoftOverMax` and not in the others is odd.
</blockquote></details>

Done.

calderpg-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NYTtqSI14uAsSrsKU1N:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
+@sherm1 for platform review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NYTvbcJAVoCnrYr3cO2:buer3go)* status: LGTM missing from assignee sherm1(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

sherm1:
@drake-jenkins-bot linux-jammy-gcc-bazel-experimental-release please

sherm1 (commented):
Platform :lgtm: pending a few minor comments

> Reviewed 2 of 6 files at r1, 1 of 4 files at r3, 1 of 2 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-:-NYUydbs2hmOG0nTBhnP:bef1vyz)* status: 6 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

___
*[`math/soft_min_max.h` line 11 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUoYGdCPrmyI5Yt0uw:-NYUoYGdCPrmyI5Yt0ux:b-ncuofm) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.h#L11)):*
> ```objc
>  max(x).
>  Mathematically we compute this as (log (∑ᵢ exp(αxᵢ))) / α.
>  @param x The vector we want to compute its soft max.
> ```

nit: Consider: "The vector _for which_ we want to compute its soft max.", or "The vector whose soft max we want to compute."
___
*[`math/soft_min_max.h` line 13 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUqEWfDcmfWHAtBdpX:-NYUqEWfDcmfWHAtBdpY:bygblmb) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.h#L13)):*
> ```objc
>  @param x The vector we want to compute its soft max.
>  @param alpha α in the documentation above. Larger α makes the soft max more
>  similar to max, with a sharper corner. Must be strictly positive and finite.
> ```

BTW consider mentioning that α defaults to 1
___
*[`math/soft_min_max.h` line 42 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUq0-TDvyrefkEbNWk:-NYUq0-U5jgzZPKEAVVW:bgaj8ep) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.h#L42)):*
> ```objc
>  similar to min, with a sharper corner. Must be strictly positive and finite.
>  @throws std::exception if α <= 0.
>  @throws std::exception if α is infinite.
> ```

minor: C++ defines "finite" to mean "not +/- infinity and not NaN", so word this "if α is non-finite" or "if α is infinite or NaN"
___
*[`math/soft_min_max.h` line 54 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUpQhJ3CsClTp-l4hk:-NYUpQhJ3CsClTp-l4hl:b-2cfep4) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.h#L54)):*
> ```objc
>  similar to min, with a sharper corner. Must be strictly positive and finite.
>  @throws std::exception if α <= 0.
>  @throws std::exception if α is infinite.
> ```

minor: same comments here
___
*[`math/soft_min_max.cc` line 18 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUsmHSD8SBV2gl3N5r:-NYUqbQyECh4zFR_Sxtl:bqtzqsy) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.cc#L18)):*
> ```cpp
> /* We use tare_x to exploit the shift invariance of log-sum-exp to avoid
> overflow. It should be set to the extremum of x (i.e., max when alpha > 0,
> min when alpha < 0). */
> ```

BTW you have restricted alpha > 0 in the public API. Consider using a different name for the variable here, e.g. `factor` or `alpha_prime` or `beta` or whatever to distinguish this signed variable from the always-positive alpha.  Otherwise you should at least provide a comment warning of the terminology change (but I think it would be better not to use the same name for two different things!).
___
*[`math/test/soft_min_max_test.cc` line 11 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUvkAJET7dOEdqL8_a:-NYUvkAJET7dOEdqL8_b:b-kx5c1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/test/soft_min_max_test.cc#L11)):*
> ```cpp
> namespace math {
> GTEST_TEST(SoftOverMax, TestDouble) {
>   const std::vector<double> x{{1., 2., 3., 1.5, 2.9, 2.99}};
> ```

nit: per [GSG](https://drake.mit.edu/styleguide/cppguide.html#Floating_Literals) should be `1,2,3,` or `1.0,2.0,3.0`

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

___
*[`math/soft_min_max.cc` line 18 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19591#-NYUsmHSD8SBV2gl3N5r:-NYVBG29EHsL8GCQy9wt:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/2055aabd32a3cbe569fed28d32a50a1a5dd636f6/math/soft_min_max.cc#L18)):*

BTW you have restricted alpha > 0 in the public API. Consider using a different name for the variable here, e.g. `factor` or `alpha_prime` or `beta` or whatever to distinguish this signed variable from the always-positive alpha.  Otherwise you should at least provide a comment warning of the terminology change (but I think it would be better not to use the same name for two different things!).
</blockquote></details>

Done. Good call.

sherm1 (commented):
> Reviewed 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19591)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @hongkai-dai)

Issue #19514, opened 2023-06-02
IrisBuilder user interface


RussTedrake:
[IRIS](https://drake.mit.edu/doxygen_cxx/group__geometry__optimization.html#gafd8111a93cb3461e05bec3477ee736f6) and "IRIS in Configuration Space" are surprisingly effective and very useful tools for decomposing the (collision-free) configuration space into convex sets.  IRIS itself only creates one convex region, but we have a number of algorithms and tools for calling IRIS repeatedly to achieve an approximate convex cover of the space.

My goal is to release this as a notebook/tutorial which provides a user interface to e.g.
- load existing IRIS regions from yaml
- easily add new IRIS regions from a list of seed configurations
- provide a MeshCat teleop interface with visual cues (e.g. coloring the robot according to whether it's in a current IRIS region, in collision, or in an uncovered region); will make use of https://github.com/RobotLocomotion/drake/issues/17784
- provide sampling-based algorithms for efficient approximate covers
- serialize the regions back to yaml

The tool should be fairly robot agnostic.

RussTedrake:
I have an essentially feature-complete draft of the builder in my [iris_builder branch](https://github.com/RussTedrake/drake/blob/iris_builder):
https://github.com/RussTedrake/drake/blob/iris_builder/tutorials/iris_builder.ipynb

RussTedrake:
I've hosted a version of my WIP notebook on Deepnote:
https://deepnote.com/workspace/russtedrake-b58d8141-6691-43e8-ad76-0575b319c8b8/project/IRIS-Builder-c7dcbe06-efff-4f84-99d3-fca8c74872d2/notebook/iris_builder-49916403b579430cae744e3e3b1ebda4

Solve times are definitely slower there, but it all works.

RussTedrake:
See #19826 for a bigger picture of how IrisBuilder fits into the GCS Trajectory Optimization roadmap.

RussTedrake:
I've pushed the current draft of the IrisBuilder notebook to the manipulation repo ([here it is on deepnote](https://deepnote.com/workspace/Manipulation-ac8201a1-470a-4c77-afd0-2cc45bc229ff/project/06-Motion-Planning-0762b167-402a-4362-9702-7d559f0e73bb/notebook/iris_builder-3c25c10bc29d4c9493e48eaced475d03)) and I will continue to update/improve it there.

[...]

PR #19602, opened 2023-06-15
Implements MultibodyPlant::DeclareBodyPoseOutputPort

RussTedrake:
This allows a user of MultibodyPlant to declare a new output port that contains just a single body pose, instead of using the vector of poses from the existing body_poses output port.  This design was discussed in #19520.

Towards #19514.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -893,6 +893,16 @@ void DoScalarDependentDefinitions(py::module m, T) {
             overload_cast_explicit<const systems::OutputPort<T>&>(
                 &Class::get_body_poses_output_port),
             py_rvp::reference_internal, cls_doc.get_body_poses_output_port.doc)
+        .def("DeclareBodyPoseOutputPort",
+            overload_cast_explicit<const systems::OutputPort<T>&,
+                const Body<T>&>(&Class::DeclareBodyPoseOutputPort),
+            py::arg("body"), py_rvp::reference_internal,
+            cls_doc.DeclareBodyPoseOutputPort.doc)
+        .def("get_body_pose_output_port",
+            overload_cast_explicit<const systems::OutputPort<T>&, BodyIndex>(
+                &Class::get_body_pose_output_port),
+            py::arg("body_index"), py_rvp::reference_internal,
+            cls_doc.get_body_pose_output_port.doc)
         .def("get_body_spatial_velocities_output_port",
             overload_cast_explicit<const systems::OutputPort<T>&>(
                 &Class::get_body_spatial_velocities_output_port),
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -1322,6 +1322,9 @@ class TestPlant(unittest.TestCase):
         plant_f.Finalize()
         plant = to_type(plant_f, T)
         models = [iiwa_model, gripper_model]
+        gripper_body = plant.GetBodyByName("body")
+        self.assertIsInstance(
+            plant.DeclareBodyPoseOutputPort(body=gripper_body), OutputPort)
 
         # Fix inputs.
         context = plant.CreateDefaultContext()
@@ -1354,9 +1357,16 @@ class TestPlant(unittest.TestCase):
             self.assertGreater(len(value), 0)
             return value[0]
 
+        def extract_value(port):
+            self.assertIsInstance(port, OutputPort)
+            return port.Eval(context)
+
         self.assertIsInstance(
             extract_list_value(plant.get_body_poses_output_port()),
             RigidTransform_[T])
+        self.assertIsInstance(
+            extract_value(plant.get_body_pose_output_port(
+                gripper_body.index())), RigidTransform_[T])
         self.assertIsInstance(
             extract_list_value(
                 plant.get_body_spatial_velocities_output_port()),
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -802,6 +802,7 @@ drake_cc_googletest(
         ":plant",
         "//common:autodiff",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//math:geometric_transform",
         "//math:gradient",
         "//multibody/test_utilities:add_fixed_objects_to_plant",
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -350,6 +350,7 @@ MultibodyPlant<T>::MultibodyPlant(const MultibodyPlant<U>& other)
     //   -body_poses_port_
     //   -body_spatial_velocities_port_
     //   -body_spatial_acclerations_port_
+    //   Note: body_pose_ports_ is initialized after FinalizePlantOnly().
     //   -state_output_port_
     //   -instance_state_output_ports_
     //   -generalized_acceleration_output_port_
@@ -410,6 +411,12 @@ MultibodyPlant<T>::MultibodyPlant(const MultibodyPlant<U>& other)
   // on the new MultibodyTree on U. Therefore we only Finalize the plant's
   // internals (and not the MultibodyTree).
   FinalizePlantOnly();
+
+  // Now declare any body pose ports that were declared on `other`.
+  for (const auto& [body_index, output_index] : other.body_pose_ports_) {
+    unused(output_index);
+    DeclareBodyPoseOutputPort(get_body(body_index));
+  }
 
   // Note: The discrete update manager needs to be copied *after* the plant is
   // finalized.
@@ -3229,6 +3236,45 @@ const OutputPort<T>& MultibodyPlant<T>::get_body_poses_output_port()
 const {
   DRAKE_MBP_THROW_IF_NOT_FINALIZED();
   return systems::System<T>::get_output_port(body_poses_port_);
+}
+
+template <typename T>
+const OutputPort<T>& MultibodyPlant<T>::DeclareBodyPoseOutputPort(
+    const Body<T>& body) {
+  DRAKE_MBP_THROW_IF_NOT_FINALIZED();
+  if (body_pose_ports_.count(body.index()) > 0) {
+    throw std::runtime_error(fmt::format(
+        "A body pose output port for body {} has already been declared.",
+        body.name()));
+  }
+
+  const systems::OutputPort<T>& port = this->DeclareAbstractOutputPort(
+      fmt::format("{}_pose", body.name()),
+      []() {
+        return AbstractValue::Make<math::RigidTransform<T>>();
+      },
+      [this, &body](const Context<T>& context, AbstractValue* X_WB) {
+        DRAKE_MBP_THROW_IF_NOT_FINALIZED();
+        this->ValidateContext(context);
+        X_WB->set_value(EvalBodyPoseInWorld(context, body));
+      },
+      {this->configuration_ticket()});
+
+  body_pose_ports_[body.index()] = port.get_index();
+  return port;
+}
+
+template <typename T>
+const OutputPort<T>& MultibodyPlant<T>::get_body_pose_output_port(
+    BodyIndex body_index) const {
+  DRAKE_MBP_THROW_IF_NOT_FINALIZED();
+  if (body_pose_ports_.count(body_index) < 1) {
+    throw std::runtime_error(
+        fmt::format("You must call DeclareBodyPoseOutputPort() for the body "
+                    "named {} in order to use the output port..",
+                    get_body(body_index).name()));
+  }
+  return systems::System<T>::get_output_port(body_pose_ports_.at(body_index));
 }
 
 template <typename T>
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -179,6 +179,7 @@ output_ports:
 - '<em style="color:gray">
   model_instance_name[i]</em>_generalized_contact_forces'
 - <span style="color:green">geometry_pose</span>
+- <em style="color:orange">[body_name]</em>_pose
 @endsystem
 
 The ports whose names begin with <em style="color:gray">
@@ -192,6 +193,10 @@ zero-length vector. (Model instances `world_model_instance()` and
 The ports shown in <span style="color:green">
 green</span> are for communication with Drake's
 @ref geometry::SceneGraph "SceneGraph" system for dealing with geometry.
+
+The ports shown in <span style="color:orange">orange</span> are not provided by
+default, but can be declared individually (e.g. via
+DeclareBodyPoseOutputPort()).
 
 %MultibodyPlant provides a user-facing API for:
 
@@ -633,6 +638,30 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   /// equal to the identity at all times.
   /// @throws std::exception if called pre-finalize.
   const systems::OutputPort<T>& get_body_poses_output_port() const;
+
+  /// Declares an output port with the name "{body.name()}_pose" for obtaining
+  /// the pose of a single `body` in the world frame, X_WB, as a
+  /// RigidTransform<T>. Returns the declared port.  See also
+  /// get_body_poses_output_port() which is always available, but may be more
+  /// cumbersome when a downstream system needs only a single pose.
+  ///
+  /// Note: You must declare the output port before creating any Context that
+  /// will make use of this port.
+  ///
+  /// @throws std::exception if called pre-finalize.
+  /// @throws std::exception if a port for `body` has already been declared on
+  /// `this`.
+  const systems::OutputPort<T>& DeclareBodyPoseOutputPort(const Body<T>& body);
+
+  /// Returns the output port of a particular body in the world frame.
+  /// Individual body pose ports are not available by default, you must call
+  /// DeclareBodyPoseOututPort() to enable them.
+  ///
+  /// @throws std::exception if called pre-finalize.
+  /// @throws std::exception if a port for `body` has not been declared on
+  /// `this`.
+  const systems::OutputPort<T>& get_body_pose_output_port(
+      BodyIndex body_index) const;
 
   /// Returns the output port of all body spatial velocities in the world frame.
   /// You can obtain the spatial velocity `V_WB` of a body B in the world frame
@@ -5297,6 +5326,7 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
   systems::OutputPortIndex body_poses_port_;
   systems::OutputPortIndex body_spatial_velocities_port_;
   systems::OutputPortIndex body_spatial_accelerations_port_;
+  std::map<BodyIndex, systems::OutputPortIndex> body_pose_ports_{};
 
   // A port presenting state x=[q v] for the whole system, and a vector of
   // ports presenting state subsets xᵢ=[qᵢ vᵢ] ⊆ x for each model instance i,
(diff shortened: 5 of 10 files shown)


RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 1 unresolved discussion, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NXyejDY9zIo55ptlV3M:b-k5bqqd) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*
> ```cpp
>         "context is created *after* declaring any output ports.",
>         cache_entry().description()));
>   }
> ```

Note: without this check, users can easily segfault in release mode.  In debug mode, they would have failed the assertion here:
```
  const CacheEntryValue& get_cache_entry_value(CacheIndex index) const {
    DRAKE_ASSERT(has_cache_entry_value(index));
```
Protecting against this is a little subtle -- since it's in the caching system and a lot of the methods were marked as being "extremely fast".  PTAL.
cc @jwnimmer-tri in case you'd like to weigh in.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 2 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXzCURe8kWG2eOnoDu1:-NXzCURe8kWG2eOnoDu2:ba08ol1) (no related file):*
I was thinking about our discussion just yesterday actually.

What about having the plant's context itself in an output port? I remember this being an old discussion but I don't remember if we ever resolved it.
With that single port, then we could request anything we'd want using plant's APIs; poses, spatial velocities, contact forces, accelerations. 
`context` is the most glorified version of state we have after all. It not only exposes state nicely, but it is also efficient in that we can re-use already cached quantities.
What was the blocker? if any?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 2 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXzCURe8kWG2eOnoDu1:-NXzIJt_6oX2CdR2fNsf:b-c4gj6n) (no related file):*

I was thinking about our discussion just yesterday actually.

What about having the plant's context itself in an output port? I remember this being an old discussion but I don't remember if we ever resolved it.
With that single port, then we could request anything we'd want using plant's APIs; poses, spatial velocities, contact forces, accelerations. 
`context` is the most glorified version of state we have after all. It not only exposes state nicely, but it is also efficient in that we can re-use already cached quantities.
What was the blocker? if any?
</blockquote></details>

That's not a solution to the need this PR is meeting. The MbP already offers body poses output port (for all bodies, in a stacked vector). Russ could _already_ get the information he needs from that port. The purpose of this PR is to make selecting just one body _easy_. Outputting a Context does nothing to make it easy.  The user would still need to write a tiny throwaway system that takes a Context on the input port and produces a RigidTransform on the output port. The need here is to have a single body's RigidTransform on an output port, so that system can connect to somewhere else downstream.

Your proposal would make more sense for a use case like `CalcCenterOfMassTranslationalVelocityInWorld` that takes a set of model instances and returns the overall CoM.  There is no output port with that data, and it's probably not easy to add a port with that kind of data.  If the user needed that data on an output port, outputting the context (or a proxy for the context) might make sense.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 2 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXzCURe8kWG2eOnoDu1:-NXzp7-U8GkL-flZf1wB:b-r6gcvy) (no related file):*
> ... this PR falls into the pattern of "add every imaginable output port that you could think of and more".

I disagree.  There is nothing in this PR aimed at adding tons of new ports for things that are not currently possible, or things that are uncommon.

The premise of this PR is to add syntactic sugar to take a frequently-encountered use case that is already _possible_ and to make it _easier_.  If you don't think the operation is commonplace enough to deserve sugar, then you can debate that in the review here.  That would be a fair point!  I don't know how commonplace this operation is or isn't.

___
___
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NXzqFsm5oTjk7zn5BOB:bfu2qed) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*

Note: without this check, users can easily segfault in release mode.  In debug mode, they would have failed the assertion here:
```
  const CacheEntryValue& get_cache_entry_value(CacheIndex index) const {
    DRAKE_ASSERT(has_cache_entry_value(index));
```
Protecting against this is a little subtle -- since it's in the caching system and a lot of the methods were marked as being "extremely fast".  PTAL.
cc @jwnimmer-tri in case you'd like to weigh in.

</blockquote></details>

This probably not the best place to deal with this fracture.

My gut feeling more towards changing out the `SystemId` as new objects (ports, state, parameters, etc) are declared.  In any case, this kind of guard should be a separate PR from the MbP feature, to help focus each review on the relevant change and get the proper sets of feature reviewers assigned to each one.

You could open an issue against me to try to design this fail-fast feature here, if you'd like to carry it forward.

RussTedrake:
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NY0t96q0N889YJVqi3A:b-40q00d) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*

This probably not the best place to deal with this fracture.

My gut feeling more towards changing out the `SystemId` as new objects (ports, state, parameters, etc) are declared.  In any case, this kind of guard should be a separate PR from the MbP feature, to help focus each review on the relevant change and get the proper sets of feature reviewers assigned to each one.

You could open an issue against me to try to design this fail-fast feature here, if you'd like to carry it forward.
</blockquote></details>

My concern is that landing the main feature in this PR without the fail fast is leaving a landmine for users.  I hit it twice myself (once in c++ and again in python — it leads to an arbitrarily mysterious message and segfault).

Are you saying we can land it with the land mine but patch that quickly in a follow-up?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 3 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NY0xnc13eOyoofiAA0O:-NY0xnc2BKnUbOB4p6zt:b-m45d5q) (no related file):*
BTW Circling back analogously to a prior discussion re: externally applied spatial force mux -- another design idea here would be when we have a `std::vector<Foo>`-valued output port, to provide syntax to connect its i'th element to a `Foo`-valued input port, without adding any more ports or MbP-specific sugar or anything.

Imagine:

```c++
builder.ConnectSelect(
    plant.get_body_poses_output_port(),
    my_body.index(),
    thing_that_wants_a_pose.get_input_port();
```

This would solve the problem for all MbP ports (generalized acceleration, etc) and any other i-th value system ports, all at once.

___
___
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NY0wYENAXAmERfbSazL:b6leu8m) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*

My concern is that landing the main feature in this PR without the fail fast is leaving a landmine for users.  I hit it twice myself (once in c++ and again in python — it leads to an arbitrarily mysterious message and segfault).

Are you saying we can land it with the land mine but patch that quickly in a follow-up?
</blockquote></details>

My suggestion is that we add the fail-fast in another PR, with a suitable design phase and SMEs invited for review.  If this PR is too dangerous without a fail-fast (and I believe you that it is), then this PR will be blocked on that PR to land first.

Since this PR is just a little sugar, it doesn't seem like waiting is too much of a big deal.  We're anyway blocked here on Joe and/or Alejandro approving the idea that this sugar is worthwhile.  You might need to provide a link to some illustrations of sample use cases for the sugar here, to help demonstrate its value.  I realize this is a slightly different response that what I said in #19520.  My replies there were presuming the sugar was justified, and figuring out the best way to provide it.  I didn't actually stop to think whether the sugar has payoff.  I'm hoping that the reviewers here can speak to that point.

RussTedrake:

BTW Circling back analogously to a prior discussion re: externally applied spatial force mux -- another design idea here would be when we have a `std::vector<Foo>`-valued output port, to provide syntax to connect its i'th element to a `Foo`-valued input port, without adding any more ports or MbP-specific sugar or anything.

Imagine:

```c++
builder.ConnectSelect(
    plant.get_body_poses_output_port(),
    my_body.index(),
    thing_that_wants_a_pose.get_input_port();
```

This would solve the problem for all MbP ports (generalized acceleration, etc) and any other i-th value system ports, all at once.
</blockquote></details>

I don't know if that is the right design.  For instance, I think that MbP might be moving away from the std::vector which assumes all bodies are present to a std::map or something that will support removing bodies.  

@joemasterjohn - your thoughts, please?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 2 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NY0xnc13eOyoofiAA0O:-NYORaRL7STML-nhzOI8:bzij85s) (no related file):*

I don't know if that is the right design.  For instance, I think that MbP might be moving away from the std::vector which assumes all bodies are present to a std::map or something that will support removing bodies.  

@joemasterjohn - your thoughts, please?
</blockquote></details>

I didn't write down the full exposition (trying to avoid polluting the discussion) but maybe I should have.

There are three fundamental structures used for composition -- scalar, sequence, and mapping (c.f., yaml, json, libstdc++, Python module `collections` etc).  I think the framework should support all three kinds of structures.  We should have scalar-valued ports, sequence-valued ports, and mapping-valued ports, and the framework should know about and manage those details. The sequence and mapping are not type-erased containers, but rather full-fledged features, can be interrogated, etc).

If bodies change from BodyIndex to BodyId, then the output ports change from sequence-valued to mapping-valued, and the connection logic allows you to connect to a specific key, instead of a specific index.

___

joemasterjohn (commented):
Feature pass done, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-:-NYPDT5m7zn2Qq_JeH1I:b-7uzwif)* status: 6 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NY0xnc13eOyoofiAA0O:-NYPCUXc2ZBpwsO1auas:b-6iylfc) (no related file):*

I didn't write down the full exposition (trying to avoid polluting the discussion) but maybe I should have.

There are three fundamental structures used for composition -- scalar, sequence, and mapping (c.f., yaml, json, libstdc++, Python module `collections` etc).  I think the framework should support all three kinds of structures.  We should have scalar-valued ports, sequence-valued ports, and mapping-valued ports, and the framework should know about and manage those details. The sequence and mapping are not type-erased containers, but rather full-fledged features, can be interrogated, etc).

If bodies change from BodyIndex to BodyId, then the output ports change from sequence-valued to mapping-valued, and the connection logic allows you to connect to a specific key, instead of a specific index.
</blockquote></details>

I like the idea for the more general id/index based port selector, but I imagine that would take a bit to design implement. Also, I don't believe the sugar introduced here impedes that solution or becomes completely redundant afterwards. So I'd vote to merge this as designed.

___
___
*[`multibody/plant/multibody_plant.cc` line 416 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NYOzgvxDUTWUQput9ON:-NYOzgvyBNXnThtx_ZJa:bexdn8b) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/multibody/plant/multibody_plant.cc#L416)):*
> ```cpp
> 
>   // Now declare any body pose ports that were declared on `other`.
>   for (const auto& [body_index, output_index] : other.body_pose_ports_) {
> ```

This will iterate and create ports in sorted order by body index, but there is no guarantee that this was the order the body ports were declared in the plant we are scalar converting from. Because they're exposed, there's nothing stopping the user from storing all of the output port indices themselves and trying to use them to index into the output ports of the scalar converted plant. Ideally this iteration order should be by output port index so that output port indices remain consistent.

The other option is to not expose the output port indices in `DeclareBodyPoseOutputPort()` and guide the user to always use `get_body_pose_output_port()`. But this also doesn't completely stop the user from using output port indices obtained from the port itself.

See this patch for a small change to your unit test that fails: https://github.com/RobotLocomotion/drake/commit/94eb3aa84eacf891f0eb53fcc41c7d485c85beab
___
*[`multibody/plant/multibody_plant.cc` line 3247 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NYP1qR8B58k7K8jqpIi:-NYP1qR8B58k7K8jqpIj:b-vd5fr4) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/multibody/plant/multibody_plant.cc#L3247)):*
> ```cpp
>   if (body_pose_ports_.count(body.index()) > 0) {
>         "A body pose output port for body {} has already been declared.",
> ```

nit: might be helpful debugging info to include the body index in the exception message.
___
*[`multibody/plant/multibody_plant.cc` line 3261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NYP3XS60EmVXtWbs_61:-NYP3XS60EmVXtWbs_62:b-8279b8) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/multibody/plant/multibody_plant.cc#L3261)):*
> ```cpp
>         X_WB->set_value(EvalBodyPoseInWorld(context, body));
>       },
>       {this->configuration_ticket()});
> ```

This is making me think that we might have the dependencies of some of our ports incorrect. I'm thinking about parameters that don't directly affect configuration, but affect kinematic quantities, such as a fixed offset frame's pose in its parent frame, etc. I think these ports as well as the full body_poses_port need to depend on the parameters ticket as well.



_Code quote:_
```cpp
      {this->configuration_ticket()});
```
___
*[`multibody/plant/test/multibody_plant_scalar_conversion_test.cc` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NYPCAmcBKJ_KkS_5M_G:-NYPCAmcBKJ_KkS_5M_H:b-b6bdom) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/multibody/plant/test/multibody_plant_scalar_conversion_test.cc#L139)):*
> ```cpp
>                                 std::nullopt, Vector3<double>::UnitZ());
>   plant.Finalize();
>   plant.DeclareBodyPoseOutputPort(body);
> ```

See this patch for a modification to this unit test to make it more robust: https://github.com/RobotLocomotion/drake/commit/94eb3aa84eacf891f0eb53fcc41c7d485c85beab
___
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NYP7TNd-_o5QEaaoYEm:b-tav4qk) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*

My suggestion is that we add the fail-fast in another PR, with a suitable design phase and SMEs invited for review.  If this PR is too dangerous without a fail-fast (and I believe you that it is), then this PR will be blocked on that PR to land first.

Since this PR is just a little sugar, it doesn't seem like waiting is too much of a big deal.  We're anyway blocked here on Joe and/or Alejandro approving the idea that this sugar is worthwhile.  You might need to provide a link to some illustrations of sample use cases for the sugar here, to help demonstrate its value.  I realize this is a slightly different response that what I said in #19520.  My replies there were presuming the sugar was justified, and figuring out the best way to provide it.  I didn't actually stop to think whether the sugar has payoff.  I'm hoping that the reviewers here can speak to that point.
</blockquote></details>

I agree with Jeremy that changing out the `SystemId` for the more general fail-fast is going to be necessary. I'm thinking forward towards removing bodies (assuming the necessary feature of removing cache entries from a system has been implemented): Let's say I add a body pose port for an existing body, then remove that body (and consequently a body pose port that was added for it) and then add a new body pose port. Does the new body pose port have the same cache index as the old port and sneak it's way past this guard?

I believe the more general guard for "Does this context correspond to the current mutated version of this system?" is necessary. 

I don't feel like it's necessary to block this PR from going in with the check as you wrote it, with a note pointing towards an issue/PR for the more general change.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19602)* status: 6 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`systems/framework/leaf_output_port.cc` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19602#-NXyejDY9zIo55ptlV3L:-NYPETCO0OKGYwNDF49C:bmkl1dw) ([raw file](https://github.com/RobotLocomotion/drake/blob/2e1e43382c2a90cbf4d47a6ab1d18b23470c686b/systems/framework/leaf_output_port.cc#L29)):*

I agree with Jeremy that changing out the `SystemId` for the more general fail-fast is going to be necessary. I'm thinking forward towards removing bodies (assuming the necessary feature of removing cache entries from a system has been implemented): Let's say I add a body pose port for an existing body, then remove that body (and consequently a body pose port that was added for it) and then add a new body pose port. Does the new body pose port have the same cache index as the old port and sneak it's way past this guard?

I believe the more general guard for "Does this context correspond to the current mutated version of this system?" is necessary. 

I don't feel like it's necessary to block this PR from going in with the check as you wrote it, with a note pointing towards an issue/PR for the more general change.
</blockquote></details>

I think I know how to make the checking properly robust.  I can collaborate with Sherm to implement it, if we're planning to go with the dynamically-added ports design.  You can open an issue assigned to me if you'd like me to work on that.

I still think it needs to be a blocker for this PR.  In other words, we do need to land the robust framework guards before we land this PR.  I don't think this check is good enough as written.

RussTedrake:
Sorry that this has stalled. I will re-open it if/when I need it again.

PR #19607, opened 2023-06-15
[multibody] Expand SapSolver results from reduced problem.

joemasterjohn:
This is the final companion PR to #18983.

We introduce one new API `SapContactProblem::ExpandContactSolverResults()` that takes as input the results of a reduced version of the contact problem as well as the mapping between original problem and reduced, and fills a `SapSolverResults` with the dimensionality of the original problem. Values of the results for non-participating DoFs and constraints are set to zero.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)

### DIFF
--- a/multibody/contact_solvers/sap/BUILD.bazel
+++ b/multibody/contact_solvers/sap/BUILD.bazel
@@ -97,8 +97,10 @@ drake_cc_library(
     deps = [
         ":contact_problem_graph",
         ":sap_constraint",
+        ":sap_solver_results",
         "//common:default_scalars",
         "//common:essential",
+        "//multibody/contact_solvers:block_sparse_matrix",
         "//multibody/plant:slicing_and_indexing",
     ],
 )
--- a/multibody/contact_solvers/sap/sap_contact_problem.cc
+++ b/multibody/contact_solvers/sap/sap_contact_problem.cc
@@ -6,6 +6,7 @@
 #include "drake/common/drake_throw.h"
 #include "drake/common/eigen_types.h"
 #include "drake/common/ssize.h"
+#include "drake/multibody/contact_solvers/block_sparse_matrix.h"
 #include "drake/multibody/contact_solvers/sap/contact_problem_graph.h"
 #include "drake/multibody/plant/slicing_and_indexing.h"
 
@@ -66,8 +67,19 @@ std::unique_ptr<SapContactProblem<T>> SapContactProblem<T>::MakeReduced(
   }
   DRAKE_DEMAND(mapping != nullptr);
 
+  mapping->velocity_permutation = PartialPermutation(num_velocities());
   mapping->clique_permutation = PartialPermutation(num_cliques());
-  mapping->constraint_permutation = PartialPermutation(num_constraints());
+  mapping->constraint_equation_permutation =
+      PartialPermutation(num_constraint_equations());
+
+  auto it = known_free_motion_dofs.begin();
+  for (int i = 0; i < num_velocities(); ++i) {
+    if (it != known_free_motion_dofs.end() && i == *it) {
+      ++it;
+    } else {
+      mapping->velocity_permutation.push(i);
+    }
+  }
 
   // Project v_star and A matrices to reduced DOFs.
   VectorX<T> v_star_reduced =
@@ -91,17 +103,85 @@ std::unique_ptr<SapContactProblem<T>> SapContactProblem<T>::MakeReduced(
   problem->set_num_objects(num_objects());
 
   // Make reduced constraints.
-  for (int i = 0; i < num_constraints(); ++i) {
-    std::unique_ptr<SapConstraint<T>> c = get_constraint(i).MakeReduced(
+  int constraint_equation_index = 0;
+  for (const auto& c : constraints_) {
+    std::unique_ptr<SapConstraint<T>> c_reduced = c->MakeReduced(
         mapping->clique_permutation, per_clique_known_free_motion_dofs);
 
-    if (c) {
-      mapping->constraint_permutation.push(i);
-      problem->AddConstraint(std::move(c));
+    if (c_reduced) {
+      problem->AddConstraint(std::move(c_reduced));
+      for (int j = 0; j < c->num_constraint_equations(); ++j) {
+        mapping->constraint_equation_permutation.push(
+            constraint_equation_index + j);
+      }
     }
+    constraint_equation_index += c->num_constraint_equations();
   }
 
   return problem;
+}
+
+template <typename T>
+void SapContactProblem<T>::ExpandContactSolverResults(
+    const ReducedMapping& reduced_mapping,
+    const SapSolverResults<T>& reduced_results,
+    SapSolverResults<T>* results) const {
+  DRAKE_DEMAND(reduced_mapping.velocity_permutation.domain_size() ==
+               num_velocities());
+  DRAKE_DEMAND(reduced_mapping.clique_permutation.domain_size() ==
+               num_cliques());
+  DRAKE_DEMAND(reduced_mapping.constraint_equation_permutation.domain_size() ==
+               num_constraint_equations());
+  DRAKE_DEMAND(reduced_results.v.size() ==
+               reduced_mapping.velocity_permutation.permuted_domain_size());
+  DRAKE_DEMAND(
+      reduced_results.gamma.size() ==
+      reduced_mapping.constraint_equation_permutation.permuted_domain_size());
+  DRAKE_DEMAND(
+      reduced_results.vc.size() ==
+      reduced_mapping.constraint_equation_permutation.permuted_domain_size());
+  DRAKE_DEMAND(reduced_results.j.size() ==
+               reduced_mapping.velocity_permutation.permuted_domain_size());
+  DRAKE_DEMAND(results != nullptr);
+
+  // Zero out everything. Results data will be selectively filled in.
+  results->Resize(num_velocities(), num_constraint_equations());
+  results->v = v_star();
+  results->gamma.setZero();
+  results->vc.setZero();
+  results->j.setZero();
+
+  // Set vc to vc* for known DoFs. Unknown DoFs will be overwritten below.
+  for (int i = 0; i < num_constraints(); ++i) {
+    const SapConstraint<T>& c = get_constraint(i);
+
+    Eigen::VectorBlock<VectorX<T>> vc_segment = results->vc.segment(
+        constraint_equations_start(i), c.num_constraint_equations());
+
+    c.first_clique_jacobian().MultiplyAndAddTo(
+        results->v.segment(velocities_start(c.first_clique()),
+                           num_velocities(c.first_clique())),
+        &vc_segment);
+
+    if (c.num_cliques() > 1) {
+      c.second_clique_jacobian().MultiplyAndAddTo(
+          results->v.segment(velocities_start(c.second_clique()),
+                             num_velocities(c.second_clique())),
+          &vc_segment);
+    }
+  }
+
+  // Copy v and j for participating velocities.
+  reduced_mapping.velocity_permutation.ApplyInverse(reduced_results.v,
+                                                    &results->v);
+  reduced_mapping.velocity_permutation.ApplyInverse(reduced_results.j,
+                                                    &results->j);
+
+  // Copy gamma and vc for participating constraints.
+  reduced_mapping.constraint_equation_permutation.ApplyInverse(
+      reduced_results.gamma, &results->gamma);
+  reduced_mapping.constraint_equation_permutation.ApplyInverse(
+      reduced_results.vc, &results->vc);
 }
 
 template <typename T>
@@ -146,6 +226,10 @@ int SapContactProblem<T>::AddConstraint(std::unique_ptr<SapConstraint<T>> c) {
       c->num_cliques() == 1
           ? graph_.AddConstraint(c->first_clique(), ni)
           : graph_.AddConstraint(c->first_clique(), c->second_clique(), ni);
+
+  // Starting index for the next constraint's velocity.
+  constraint_equations_start_.push_back(constraint_equations_start_.back() +
+                                        c->num_constraint_equations());
 
   constraints_.push_back(std::move(c));
 
--- a/multibody/contact_solvers/sap/sap_contact_problem.h
+++ b/multibody/contact_solvers/sap/sap_contact_problem.h
@@ -8,6 +8,7 @@
 #include "drake/common/eigen_types.h"
 #include "drake/multibody/contact_solvers/sap/contact_problem_graph.h"
 #include "drake/multibody/contact_solvers/sap/sap_constraint.h"
+#include "drake/multibody/contact_solvers/sap/sap_solver_results.h"
 #include "drake/multibody/math/spatial_algebra.h"
 
 namespace drake {
@@ -18,8 +19,9 @@ namespace internal {
 /* Struct returned by SapContactProblem::MakeReduced() which stores the mapping
    between the original and reduced problems.*/
 struct ReducedMapping {
+  PartialPermutation velocity_permutation;
   PartialPermutation clique_permutation;
-  PartialPermutation constraint_permutation;
+  PartialPermutation constraint_equation_permutation;
 };
 
 /* In the SAP formulation of contact the state of a mechanical system is
@@ -110,7 +112,8 @@ class SapContactProblem {
     information in `known_free_motion_dofs`, but transformed to clique local
     indices.
     @param[out] mapping On output it will store information to map DOFs and
-    constraints between the original and reduced problems in a ReducedMapping.
+    constraint equations between the original and reduced problems in a
+    ReducedMapping.
 
     @pre known_free_motion_dofs is a strict ordered subset of
          [0, ..., num_velocities()-1].
@@ -123,6 +126,41 @@ class SapContactProblem {
       const std::vector<int>& known_free_motion_dofs,
       const std::vector<std::vector<int>>& per_clique_known_free_motion_dofs,
       ReducedMapping* mapping) const;
+
+  /* Maps solver results for a reduced version of this problem obtained with
+    MakeReduced() into solver results for this original problem. Known
+    velocities eliminated from the reduced problem are set to v* in `results`,
+    consistent with the documentation in MakeReduced(). Constraints eliminated
+    in the reduced problem do not participate and therefore their corresponding
+    impulses are set to zero.
+
+     @param[in] reduced_mapping Stores the mapping between this problem and a
+       reduced problem obtained with MakeReduced().
+     @param[in] reduced_results Solver results for a reduced version of this
+       problem consistent with `reduced_mapping`.
+     @param[out] results On output stores the solver results contained in
+       `reduced_results` mapped back to this original problem. Known velocities
+        are set to v* and impulses for constraints eliminated from the reduced
+        problem are set to zero.
+     @pre reduced_mapping.velocity_permutation.domain_size() ==
+          num_velocities().
+     @pre reduced_mapping.clique_permutation.domain_size() == num_cliques().
+     @pre reduced_mapping.constraint_equation_permutation.domain_size() ==
+       num_constraint_equations().
+     @pre reduced_results.v.size() ==
+          reduced_mapping.velocity_permutation.permuted_domain_size()
+     @pre reduced_results.j.size() ==
+          reduced_mapping.velocity_permutation.permuted_domain_size()
+     @pre reduced_results.gamma.size() ==
+          reduced_mapping.constraint_equation_permutation.permuted_domain_size()
+     @pre reduced_results.vc.size() ==
+          reduced_mapping.constraint_equation_permutation.permuted_domain_size()
+     @pre results != nullptr.
+     @see SapContactProblem::MakeReduced() for more information.
+  */
+  void ExpandContactSolverResults(const ReducedMapping& reduced_mapping,
+                                  const SapSolverResults<T>& reduced_results,
+                                  SapSolverResults<T>* results) const;
 
   /* TODO(amcastro-tri): consider constructor API taking std::vector<VectorX<T>>
    for v_star. It could be useful for deformables. */
@@ -181,6 +219,17 @@ class SapContactProblem {
     return graph_.num_constraint_equations();
   }
 
+  /* Returns the index to the first constraint velocity for a given constraint,
+   within the full vector of constraint velocities for the entire problem. That
+   is, with vc the full vector of constraint velocities for this problem,
+   vc.segment(constraint_equations_start(c), num_constraint_equations(c))
+   corresponds to the vector of constraint velocities for the c-th constraint.*/
+  int constraint_equations_start(int constraint_index) const {
+    DRAKE_THROW_UNLESS(0 <= constraint_index &&
+                       constraint_index < num_constraints());
+    return constraint_equations_start_[constraint_index];
+  }
+
   /* Accesses constraint with index `constraint_index` as assigned by the call
    to AddConstraint(). */
   const SapConstraint<T>& get_constraint(int constraint_index) const {
@@ -224,6 +273,10 @@ class SapContactProblem {
   T time_step_{0.0};    // Discrete time step.
   int num_objects_{0};  // Number of physical objects.
   std::vector<int> velocities_start_;
+  // Gives the index of the first constraint equation for each constraint.
+  // Has size = num_constraints() + 1 and at any time:
+  // constraint_equations_start_.back() == num_constraint_equations().
+  std::vector<int> constraint_equations_start_{0};
   std::vector<MatrixX<T>> A_;  // Linear dynamics matrix.
   VectorX<T> v_star_;          // Free-motion velocities.
   ContactProblemGraph graph_;  // Contact graph for this problem.
--- a/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc
+++ b/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc
@@ -43,7 +43,7 @@ class TestConstraint final : public SapConstraint<double> {
   // No objects are registered.
   TestConstraint(int num_constraint_equations, int clique, int clique_nv)
       : SapConstraint<double>(
-            {clique, MatrixXd::Zero(num_constraint_equations, clique_nv)}, {}) {
+            {clique, MatrixXd::Ones(num_constraint_equations, clique_nv)}, {}) {
   }
 
   // Constructor for a constraint between two cliques.
@@ -52,9 +52,9 @@ class TestConstraint final : public SapConstraint<double> {
                  int first_clique_nv, int second_clique, int second_clique_nv)
       : SapConstraint<double>(
             {first_clique,
-             MatrixXd::Zero(num_constraint_equations, first_clique_nv),
+             MatrixXd::Ones(num_constraint_equations, first_clique_nv),
              second_clique,
-             MatrixXd::Zero(num_constraint_equations, second_clique_nv)},
+             MatrixXd::Ones(num_constraint_equations, second_clique_nv)},
             {}) {}
 
   // N.B no-op overloads to allow us compile this testing constraint. These
@@ -231,6 +231,13 @@ GTEST_TEST(ContactProblem, AddConstraints) {
   EXPECT_EQ(problem.num_constraint_equations(), 17);
   EXPECT_EQ(problem.num_objects(), 4);
 
+  // Expected constraint equation indicies.
+  const std::vector<int> constraint_equations_start_expected{0, 1, 4, 10, 12};
+  for (int i = 0; i < problem.num_constraints(); ++i) {
+    EXPECT_EQ(problem.constraint_equations_start(i),
+              constraint_equations_start_expected[i]);
+  }
+
   // Verify graph for this problem.
   const ContactProblemGraph& graph = problem.graph();
   EXPECT_EQ(graph.num_cliques(), 4);
@@ -300,14 +307,15 @@ GTEST_TEST(ContactProblem, MakeReduced) {
   SapContactProblem<double> problem(time_step, std::move(A), std::move(v_star));
   AddConstraints(&problem);
 
-  // Lock all the dofs of clique 0 and 1. Lock velocity index 9, local index 1
+  // Lock all the dofs of clique 0 and 1. Lock velocity index 9, local index 0
   // of clique 3.
-  const std::vector<int> locked_indices{0, 1, 2, 3, 4, 9};
-  const std::vector<std::vector<int>> clique_locked_indices{
-      {0, 1}, {0, 1, 2}, {}, {1}};
+  const std::vector<int> unknown_indices{5, 6, 7, 8, 10};
+  const std::vector<int> known_indices{0, 1, 2, 3, 4, 9};
+  const std::vector<std::vector<int>> clique_known_indices{
+      {0, 1}, {0, 1, 2}, {}, {0}};
   ReducedMapping mapping;
   std::unique_ptr<SapContactProblem<double>> reduced_problem =
-      problem.MakeReduced(locked_indices, clique_locked_indices, &mapping);
+      problem.MakeReduced(known_indices, clique_known_indices, &mapping);
 
   EXPECT_EQ(reduced_problem->num_cliques(), 2);
   EXPECT_EQ(reduced_problem->num_constraints(), 4);
@@ -328,6 +336,21 @@ GTEST_TEST(ContactProblem, MakeReduced) {
   EXPECT_EQ(graph.num_clusters(), 1);
   EXPECT_EQ(graph.num_constraint_equations(), 14);
 
+  /* Velocity permutation. Expect velocities 5, 6, 7, 8, 10 to participate. */
+  EXPECT_EQ(mapping.velocity_permutation.domain_size(),
+            problem.num_velocities());
+  EXPECT_EQ(mapping.velocity_permutation.permuted_domain_size(), 5);
+  for (int i : unknown_indices) {
+    EXPECT_TRUE(mapping.velocity_permutation.participates(i));
+  }
+  for (int i : known_indices) {
+    EXPECT_FALSE(mapping.velocity_permutation.participates(i));
+  }
+  for (int i = 0; i < ssize(unknown_indices); ++i) {
+    EXPECT_EQ(mapping.velocity_permutation.permuted_index(unknown_indices[i]),
+              i);
+  }
+
   /* Clique permutation. We expect the first two cliques to be eliminated. */
   EXPECT_EQ(mapping.clique_permutation.domain_size(), problem.num_cliques());
   EXPECT_EQ(mapping.clique_permutation.permuted_domain_size(), 2);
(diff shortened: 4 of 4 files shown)


joemasterjohn (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NXzqGkm9LOFefd7bQXP:b5xll3r)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 2 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 153 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NXzqYK88S8-s6jjf84F:-NXzqYK88S8-s6jjf84G:bqs9hq2) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L153)):*
> ```objc
>      @see SapContactProblem::MakeReduced() for more information.
>   */
>   void ExpandContactSolverResults(
> ```

BTW I wavered between having this function in SapDriver vs here and landed on putting it here. To me it makes more sense living in the only class that knows about the semantics of `ReducedMapping` so far. It has the side effect of `SapContactProblem` know having a dependency on `SapSolverResults`, but I'm not too sure how bad (if at all) that is. Let me know if you think this should live somewhere else.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 69 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NXzqP0v-J3u0RH4ALXk:-NXzqP0v-J3u0RH4ALXl:b2gqmxo) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L69)):*
> ```cpp
>   DRAKE_DEMAND(mapping != nullptr);
> 
>   mapping->velocity_permutation = PartialPermutation(num_velocities());
> ```

BTW Here I had to add the velocity permutation to the mapping data structure in order to map v and j. Tests for this addition are added.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 16 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNs7Bs6JcvXdxkbyQx:-NYNs7Bs6JcvXdxkbyQy:b-7rgqal) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*
> ```objc
>   /* This method maps the SapSolverResults in `reduced_results` for a reduced
>      version of this SapContactProblem into results stored
>      `sap_results`. This mapping ensures that the dimensionality of the results
> ```

The intro to these docs should probably state up front the connection of this method to `MakeReduced()` and by doing so you'd avoid having to explain things already documented there. My try:

"Maps solver results for a reduced version of this problem obtained with MakeReduced() into solver results for this original problem. Known velocities eliminated from the reduced problem are set to v* in `results`,  consistently with the documentation in MakeReduced(). Constraints eliminated in the reduced problem do not participate and therefore their corresponding impulses are set to zero.



_Code quote:_
```objc
  /* This method maps the SapSolverResults in `reduced_results` for a reduced
     version of this SapContactProblem into results stored
     `sap_results`. This mapping ensures that the dimensionality of the results
```
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNrZGL3TSYWrj65qI4:-NYNrZGMAMXTYhZifD7O:bypn810) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*
> ```objc
>      are mapped back to their original indices. The ReducedMapping stored  in
>      `reduced_mapping` encodes the reduced problem and how to map
>      from permuted indices back to original indices. Velocities and impulses for
> ```

Per documentation in  `MakeReduced()` I believe we expect known velocities to equal v*, which could or not be zero.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNuUcv8z6wi9B5xsjy:-NYNuUcv8z6wi9B5xsjz:b-f6u9oz) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*
> ```objc
>      are mapped back to their original indices. The ReducedMapping stored  in
>      `reduced_mapping` encodes the reduced problem and how to map
>      from permuted indices back to original indices. Velocities and impulses for
> ```

you say this below in the docs for this variable. Too close in scope to do that IMO



_Code quote:_
```objc
     are mapped back to their original indices. The ReducedMapping stored  in
     `reduced_mapping` encodes the reduced problem and how to map
     from permuted indices back to original indices. Velocities and impulses for
```
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 144 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNv0jH9gLEVz3nL9Y7:-NYNv0jH9gLEVz3nL9Y8:b-ijksb0) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L144)):*
> ```objc
>      @param[out] sap_results On output stores the solver results contained in
>        `reduced_results` mapped back to original indices. Zero values are stored
>        for non-participating velocities and constraints.
> ```

We can exploit the relationships better when documenting this, I think



_Suggestion:_
```objc
     @param[in] reduced_mapping Stores the mapping between this problem and a
       reduced problem obtained with MakeReduced().
     @param[in] reduced_results Solver results for a reduced version of this
       problem consistent with `reduced_mapping`.
     @param[out] sap_results On output stores the solver results contained in
       `reduced_results` mapped back to this original problem. Known velocities 
        are set to v* and impulses for constraints eliminated from the reduced
        problem are set to zero.
```
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 148 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNvxUX32HP8ZO-mJKn:-NYNvxUX32HP8ZO-mJKo:bqbmgrb) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L148)):*
> ```objc
>      num_velocities().
>      @pre reduced_mapping.clique_permutation.domain_size() == num_cliques().
>      @pre reduced_mapping.constraint_permutation.domain_size() ==
> ```

shouldn't there be a preconditions for `reduced_results` as well?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 153 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NXzqYK88S8-s6jjf84F:-NYNw4cYBqZ16Q72KFlN:b-qthwkg) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L153)):*

BTW I wavered between having this function in SapDriver vs here and landed on putting it here. To me it makes more sense living in the only class that knows about the semantics of `ReducedMapping` so far. It has the side effect of `SapContactProblem` know having a dependency on `SapSolverResults`, but I'm not too sure how bad (if at all) that is. Let me know if you think this should live somewhere else.
</blockquote></details>

I don't think it bothers me.
Another option that makes sense is to place this in `SapSolver`, who does know about both the problem and the results.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO22cu0_6X9fBLv4aR:-NYO22cu0_6X9fBLv4aS:b-635946) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L156)):*
> ```objc
>       const contact_solvers::internal::ReducedMapping& reduced_mapping,
>       const contact_solvers::internal::SapSolverResults<T>& reduced_results,
>       contact_solvers::internal::SapSolverResults<T>* sap_results) const;
> ```

nit, for consistency consider `sap_reduced_results` and `sap_results`, or `reduced_results` and `results`, but not a mix.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNyMFHEwQNloJApdrL:-NYNyMFHEwQNloJApdrM:b-movolg) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L75)):*
> ```cpp
>   auto it = known_free_motion_dofs.begin();
>   for (int i = 0; i < num_velocities(); ++i) {
>     if (it != known_free_motion_dofs.end() && i == *it) {
> ```

I think here you can do:
```c++
for (int v : known_free_motion_dofs) mapping.velocity_permutation.push(v);
```
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 132 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNypKn9dvr_xqDLH5-:-NYNypKn9dvr_xqDLH50:bs49g8x) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L132)):*
> ```cpp
>   // Zero out everything. Results data will be selectively filled in.
>   sap_results->Resize(num_velocities(), num_constraint_equations());
>   sap_results->v.setZero();
> ```

This should be v*
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 134 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNz5li0SB7Rnzo7tP9:-NYNz5li0SB7Rnzo7tPA:b-w9ii2h) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L134)):*
> ```cpp
>   sap_results->v.setZero();
>   sap_results->gamma.setZero();
>   sap_results->vc.setZero();
> ```

to be consistent, I believe we should set constraint velocities to vc*. 
I don't think we'll need that in the near future, but I'd at least have a big warning in the docs for this method in case that matters in the future (consider prescribed motions for instance).
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO-50KCT1dwOmujNtl:-NYO-50KCT1dwOmujNtm:b-558ydc) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L140)):*
> ```cpp
>   for (int i = 0; i < num_velocities(); ++i) {
>     if (reduced_mapping.velocity_permutation.participates(i)) {
>       sap_results->v[i] =
> ```

why not to use `ApplyInverse()` here?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO-M5I6APRalAXEj5z:-NYO-M5I6APRalAXEj6-:brd7tlg) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L154)):*
> ```cpp
>   for (int i = 0; i < num_constraints(); ++i) {
>     const SapConstraint<T>& c = get_constraint(i);
>     if (reduced_mapping.constraint_permutation.participates(i)) {
> ```

had you saved the permutation on constraint equations instead, you could've used `ApplyInverse()`
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 305 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO-chVC52fsTZYkrQC:-NYO-chVC52fsTZYkrQD:b-vbjfjn) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L305)):*
> ```cpp
>   // Lock all the dofs of clique 0 and 1. Lock velocity index 9, local index 1
>   // of clique 3.
>   const std::vector<int> unlocked_indices{5, 6, 7, 8, 10};
> ```

nit, I still see the "lock" terminology in this file.
Within the context of this file it'd be more appropriate to say "known" or "unknown" 
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 459 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO1ZCp-s_1c7lTupuR:-NYO1ZCp-s_1c7lTupuS:b-ovm5n1) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L459)):*
> ```cpp
>   const std::vector<int> locked_indices{0, 1, 2, 3, 4, 9};
>   const std::vector<std::vector<int>> clique_locked_indices{
>       {0, 1}, {0, 1, 2}, {}, {0}};
> ```

comment says local index 1, here you use local index 0
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 501 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO1wUP3UxSttKJ_F-P:-NYO1wUP3UxSttKJ_F-Q:bqyute8) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L501)):*
> ```cpp
> 
>   int equation_index = 0;
>   int reduced_index = 0;
> ```

nit, for consistency, maybe name `reduced_equation_index`

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 12 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNs7Bs6JcvXdxkbyQx:-NYTFjq12BuSMVYD1oIw:b4dqo45) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L131)):*

The intro to these docs should probably state up front the connection of this method to `MakeReduced()` and by doing so you'd avoid having to explain things already documented there. My try:

"Maps solver results for a reduced version of this problem obtained with MakeReduced() into solver results for this original problem. Known velocities eliminated from the reduced problem are set to v* in `results`,  consistently with the documentation in MakeReduced(). Constraints eliminated in the reduced problem do not participate and therefore their corresponding impulses are set to zero.
</blockquote></details>

Done. Thanks for the snippet.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNrZGL3TSYWrj65qI4:-NYTFqH36jW0fm7-5Jz5:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*

Per documentation in  `MakeReduced()` I believe we expect known velocities to equal v*, which could or not be zero.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNuUcv8z6wi9B5xsjy:-NYTGBvlEm4oe-yPUa-L:bhv1u4h) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L135)):*

you say this below in the docs for this variable. Too close in scope to do that IMO
</blockquote></details>

Done. Removed in the re-written docs.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 144 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNv0jH9gLEVz3nL9Y7:-NYTGZGW3Ot0GI5fEPKH:b-rkerv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L144)):*

We can exploit the relationships better when documenting this, I think
</blockquote></details>

Done. Thanks this looks better. 
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 148 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNvxUX32HP8ZO-mJKn:-NYTWsp22T_1uXmakM_L:be5yh18) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L148)):*

shouldn't there be a preconditions for `reduced_results` as well?
</blockquote></details>

Done. Yes the size of the reduced results should match the size of the permuted domains of the different partial permutations.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 153 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NXzqYK88S8-s6jjf84F:-NYTXTfsAWpZl7QzOx-Q:b7tsgdj) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L153)):*

I don't think it bothers me.
Another option that makes sense is to place this in `SapSolver`, who does know about both the problem and the results.
</blockquote></details>

I think it makes more sense to have it in `SapSolver` actually, because I need to be able to evaluate `vc_star` easily for the expanded results.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNyMFHEwQNloJApdrL:-NYTYMI0-mtCruWKLjYO:be7gm3g) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L75)):*

I think here you can do:
```c++
for (int v : known_free_motion_dofs) mapping.velocity_permutation.push(v);
```
</blockquote></details>

It's actually the opposite, the known velocities do not participate in the permutation.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 132 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNypKn9dvr_xqDLH5-:-NYTYz_o33d9kOn4ROme:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L132)):*

This should be v*
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 134 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNz5li0SB7Rnzo7tP9:-NYTZ-ZgEGwQCi4v0xYM:b-asssa2) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L134)):*

to be consistent, I believe we should set constraint velocities to vc*. 
I don't think we'll need that in the near future, but I'd at least have a big warning in the docs for this method in case that matters in the future (consider prescribed motions for instance).
</blockquote></details>

Ah that is true. I think I'll have to move this function to `SapSolver` to have easier access to computing `vc_star`. But from a quick look, it seems like `SapModel` (which is constructed on a call to SolveWithGuess) is the class with the capability of easily getting a cached vc_star. But a `SapModel` won't be constructed for the original problem, only the reduced problem. It seems inefficient to construct a whole `SapModel` just to evaluate `vc_star` once for these results. How do you reccommend getting `vc_star`?
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO-50KCT1dwOmujNtl:-NYTaCTG7AOgqaj7Bx8N:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L140)):*

why not to use `ApplyInverse()` here?
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO-M5I6APRalAXEj5z:-NYTaG-jBvg9DsUyROko:bs8b6qw) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L154)):*

had you saved the permutation on constraint equations instead, you could've used `ApplyInverse()`
</blockquote></details>

Done. I modified `MakeReduced` to use a permutation on the constraint equations and it makes this function so much cleaner.
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 459 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYO1ZCp-s_1c7lTupuR:-NYTaOjW64GgdutXq3Ds:b-a46hav) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L459)):*

comment says local index 1, here you use local index 0
</blockquote></details>

Done. Updated the inaccurate comment.

amcastro-tri (commented):
Thanks @joemasterjohn. :lgtm: 

+@sherm1 for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYTzKC38IfLRcgxh4GU:bayc4eo)* status: 4 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNyMFHEwQNloJApdrL:-NYTxDBtBVF5O7rQFjOm:b-3wc6oa) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L75)):*

It's actually the opposite, the known velocities do not participate in the permutation.
</blockquote></details>

oh, I see. This requires `knonw_free_motions_dofs` to be sorted right?
We should at least document that pre-condition, or better if the code didn't need it.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 134 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNz5li0SB7Rnzo7tP9:-NYTyt7Y38qc26Vfg7qJ:bs6ia29) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L134)):*

Ah that is true. I think I'll have to move this function to `SapSolver` to have easier access to computing `vc_star`. But from a quick look, it seems like `SapModel` (which is constructed on a call to SolveWithGuess) is the class with the capability of easily getting a cached vc_star. But a `SapModel` won't be constructed for the original problem, only the reduced problem. It seems inefficient to construct a whole `SapModel` just to evaluate `vc_star` once for these results. How do you reccommend getting `vc_star`?
</blockquote></details>

per comment above, I only recommended a big warning for now. Constraints have the Jacobians, so we should be able to get vc* just from the problem.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 106 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYTyRR_BjJ7Cq2m9LcY:-NYTyRRa87f9G5CDEFyb:bhjshcz) ([raw file](https://github.com/RobotLocomotion/drake/blob/ac8f9a1be6a5a4da8d46641182b3d60a34dfa9e1/multibody/contact_solvers/sap/sap_contact_problem.cc#L106)):*
> ```cpp
>   // Make reduced constraints.
>   int constraint_equation_index = 0;
>   for (auto&& c : constraints_) {
> ```

why not `const auto&`?

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNyMFHEwQNloJApdrL:-NYUdJY21wCfPC2FJboz:b-i2tep2) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L75)):*

oh, I see. This requires `knonw_free_motions_dofs` to be sorted right?
We should at least document that pre-condition, or better if the code didn't need it.
</blockquote></details>

I said "strictly ordered" in the doc, but changed it to "sorted". 
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 134 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNz5li0SB7Rnzo7tP9:-NYUdt9v34U1wKKrcK6g:b-fnhqgk) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L134)):*

per comment above, I only recommended a big warning for now. Constraints have the Jacobians, so we should be able to get vc* just from the problem.
</blockquote></details>

Ah true, I was thinking I'd need to build the whole dense Jacobian. I took a crack at it, lemme know what you think or if I should just go with a warning for now.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 106 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYTyRR_BjJ7Cq2m9LcY:-NYUdT9e6va7odTQ2BqJ:b-4rjnem) ([raw file](https://github.com/RobotLocomotion/drake/blob/ac8f9a1be6a5a4da8d46641182b3d60a34dfa9e1/multibody/contact_solvers/sap/sap_contact_problem.cc#L106)):*

why not `const auto&`?
</blockquote></details>

Yeah `const auto&` is stricter and clearer, I just wrote that fast.

sherm1 (commented):
FYI sap_contact_problem_test is failing on three of the CI platforms

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYV5Eco2jvqOg7sec6F:b-iph2t2)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

amcastro-tri (commented):
> Reviewed 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 153 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NXzqYK88S8-s6jjf84F:-NYV63Fm5kAyXBgBKghb:b-d2i4s7) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.h#L153)):*

I think it makes more sense to have it in `SapSolver` actually, because I need to be able to evaluate `vc_star` easily for the expanded results.
</blockquote></details>

Up to you. I'm satisfied with this as is.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 228 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV7tYs-TwE-QDVsZnW:-NYV7tYs-TwE-QDVsZnX:bgb2z0m) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.h#L228)):*
> ```objc
>   vc.segment(constraint_velocity_start(c), num_constraint_equations(c))
>   corresponds to the vector of constraint velocities for the c-th constraint.*/
>   int constraint_velocity_start(int constraint_index) const {
> ```

nit, you might want to add unit test.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 75 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNyMFHEwQNloJApdrL:-NYV6MZr3tTrFNJWXddt:b-at1kxp) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L75)):*

I said "strictly ordered" in the doc, but changed it to "sorted". 
</blockquote></details>

I simply missed it. Either wording is find for me. Just use the same wording in all pre-conditions for consistency. Now there's a mix.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 134 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYNz5li0SB7Rnzo7tP9:-NYV8qfABhI-g9WijZGQ:b-286c2f) ([raw file](https://github.com/RobotLocomotion/drake/blob/93a3bb6bc7faab79b2c7b8521c9d115461a96182/multibody/contact_solvers/sap/sap_contact_problem.cc#L134)):*

Ah true, I was thinking I'd need to build the whole dense Jacobian. I took a crack at it, lemme know what you think or if I should just go with a warning for now.
</blockquote></details>

It looks good. You should probably add a test to make sure the new code path is exercised. 
___
*[`multibody/contact_solvers/sap/test/sap_contact_problem_test.cc` line 509 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV8mhx2jecFZYMrGMZ:-NYV8mhx2jecFZYMrGM_:bgrczyb) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/test/sap_contact_problem_test.cc#L509)):*
> ```cpp
>   }
> 
>   EXPECT_TRUE(CompareMatrices(results.v, v_expected));
> ```

test for vc_star needed.

sherm1 (commented):
Platform pass done. I'll hold off on lgtm until CI is clean., lmk.

> Reviewed 1 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYVAE0Z7YHzkGwLyqUi:bo5brip)* status: 7 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 146 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV6rAg6jY_FxKRTjFf:-NYV6rAg6jY_FxKRTjFg:b3o6rkb) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.h#L146)):*
> ```objc
>         problem are set to zero.
>      @pre reduced_mapping.velocity_permutation.domain_size() ==
>      num_velocities().
> ```

nit: indent to match below
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 227 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV6ZxaET3AVq89ko1y:-NYV6ZxaET3AVq89ko1z:b-z9167m) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.h#L227)):*
> ```objc
>   is, with vc the full vector of constraint velocities for this problem,
>   vc.segment(constraint_velocity_start(c), num_constraint_equations(c))
>   corresponds to the vector of constraint velocities for the c-th constraint.*/
> ```

nit: indenting for this block is different than the others
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 128 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV84vX-aEaNu0DSlff:-NYV84vX-aEaNu0DSlfg:b-g11544) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.cc#L128)):*
> ```cpp
>     const contact_solvers::internal::ReducedMapping& reduced_mapping,
>     const contact_solvers::internal::SapSolverResults<T>& reduced_results,
>     contact_solvers::internal::SapSolverResults<T>* results) const {
> ```

nit: you're in contact_solvers::internal:: already so shouldn't have to write this out
___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 234 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV96w9BM6YfY2jt4ZO:-NYV96w9BM6YfY2jt4ZP:bjw1tde) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.cc#L234)):*
> ```cpp
>   } else {
>     constraint_velocity_start_.push_back(constraint_velocity_start_.back() +
>                                          c->num_constraint_equations());
> ```

BTW is this right? I would have expected the start of _this_ constraint c to depend on the start of the previous constraint plus the number of equations in the _previous_ constraint.

amcastro-tri (commented):
> Reviewed 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 227 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYY_etpDI9VR5ok_s4y:-NYY_etpDI9VR5ok_s4z:b-9smqhq) ([raw file](https://github.com/RobotLocomotion/drake/blob/952c46a90382d75b75910838340dcfceff073d24/multibody/contact_solvers/sap/sap_contact_problem.h#L227)):*
> ```objc
>    vc.segment(constraint_velocities_start(c), num_constraint_equations(c))
>    corresponds to the vector of constraint velocities for the c-th constraint.*/
>   int constraint_velocities_start(int constraint_index) const {
> ```

nit, In retrospect, this might get confused with "generalized velocities".
Consider instead `constraint_equations_start()`, which would apply to both constraint velocities and impulses.
___
*[`multibody/contact_solvers/sap/sap_contact_problem.h` line 276 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYY_OxmAVtVJf4kcNo8:-NYY_OxmAVtVJf4kcNo9:bbpj3wj) ([raw file](https://github.com/RobotLocomotion/drake/blob/952c46a90382d75b75910838340dcfceff073d24/multibody/contact_solvers/sap/sap_contact_problem.h#L276)):*
> ```objc
>   int num_objects_{0};  // Number of physical objects.
>   std::vector<int> velocities_start_;
>   std::vector<int> constraint_velocities_start_{0};
> ```

nice. 
Consider adding a comment that `constraint_velocities_start_.size()` equals `num_constraints()+1`, so that `constraint_velocities_start_.back()` equals `num_constraint_equations()`.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @joemasterjohn)

___
*[`multibody/contact_solvers/sap/sap_contact_problem.cc` line 234 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-NYV96w9BM6YfY2jt4ZO:-NYYl8OBC9tKKXIUy05A:bihyj) ([raw file](https://github.com/RobotLocomotion/drake/blob/a4b93b73318762e61c677fd33cb567d28c933929/multibody/contact_solvers/sap/sap_contact_problem.cc#L234)):*

Nope, it's totally off by one. Good catch thanks. Now I initialize `constraint_velocity_start` with `{0}` and each constraint pushes the starting index of the next constraint to be added.
</blockquote></details>

Please make sure there is a unit test that would have failed with the off-by-one error but passes now.

sherm1 (commented):
Great, thanks Joe! Platform :lgtm:
I'm assuming +(release notes: none) for this PR; please change if that's wrong.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYZ0Bp85D7jdlXd4-x9:b-2cqzug)* status: 2 unresolved discussions

sherm1 (commented):
@amcastro-tri you are still blocking but I think Joe addressed your request

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYZ0grA4f1zLnOTE6Ak:bnwmika)* status: 1 unresolved discussion

amcastro-tri (commented):
nice! thanks @joemasterjohn 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19607#-:-NYZQgYO-oqzsqVIZNv_:b-zcbvlv)* status: :shipit: complete! all discussions resolved, LGTM from assignees amcastro-tri,sherm1(platform)

PR #19609, opened 2023-06-15
[C-IRIS] Compute s on the kinematics chain between each body.

hongkai-dai:
For CspaceFreeBox, for the separating plane between a pair of geometries, the a(s) and b(s) of that plane only need to depend on the subset of s on the kinematics chain between the pair of geometries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609)

### DIFF
--- a/geometry/optimization/cspace_free_box.cc
+++ b/geometry/optimization/cspace_free_box.cc
@@ -13,7 +13,9 @@ CspaceFreeBox::CspaceFreeBox(const multibody::MultibodyPlant<double>* plant,
                              const geometry::SceneGraph<double>* scene_graph,
                              SeparatingPlaneOrder plane_order,
                              const Options& options)
-    : CspaceFreePolytopeBase(plant, scene_graph, plane_order, options) {}
+    : CspaceFreePolytopeBase(plant, scene_graph, plane_order,
+                             CspaceFreePolytopeBase::SForPlane::kOnChain,
+                             options) {}
 
 CspaceFreeBox::~CspaceFreeBox() {}
 
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -243,7 +243,8 @@ CspaceFreePolytope::CspaceFreePolytope(
     SeparatingPlaneOrder plane_order,
     const Eigen::Ref<const Eigen::VectorXd>& q_star,
     const CspaceFreePolytope::Options& options)
-    : CspaceFreePolytopeBase(plant, scene_graph, plane_order, options),
+    : CspaceFreePolytopeBase(plant, scene_graph, plane_order,
+                             CspaceFreePolytopeBase::SForPlane::kAll, options),
       q_star_{q_star} {
   s_lower_ = rational_forward_kin().ComputeSValue(
       rational_forward_kin().plant().GetPositionLowerLimits(), q_star_);
--- a/geometry/optimization/cspace_free_polytope_base.cc
+++ b/geometry/optimization/cspace_free_polytope_base.cc
@@ -32,33 +32,19 @@ void FindMonomialBasisArray(
     const multibody::RationalForwardKinematics& rational_forward_kin,
     const Vector3<symbolic::Variable>& y_slack,
     const SortedPair<multibody::BodyIndex>& body_pair,
+    const std::unordered_map<SortedPair<multibody::BodyIndex>,
+                             std::vector<int>>& map_body_pair_to_s_on_chain,
     std::unordered_map<SortedPair<multibody::BodyIndex>,
                        std::array<VectorX<symbolic::Monomial>, 4>>*
         map_body_to_monomial_basis_array,
     std::array<VectorX<symbolic::Monomial>, 4>* monomial_basis_array) {
   auto body_pair_it = map_body_to_monomial_basis_array->find(body_pair);
   if (body_pair_it == map_body_to_monomial_basis_array->end()) {
-    const std::vector<multibody::internal::MobilizerIndex> mobilizer_indices =
-        multibody::internal::FindMobilizersOnPath(rational_forward_kin.plant(),
-                                                  body_pair.first(),
-                                                  body_pair.second());
-    const auto& tree =
-        multibody::internal::GetInternalTree(rational_forward_kin.plant());
     symbolic::Variables s_set;
-    for (const auto& mobilizer_index : mobilizer_indices) {
-      const auto& mobilizer = tree.get_mobilizer(mobilizer_index);
-      if ((mobilizer.can_rotate() && !mobilizer.can_translate()) ||
-          (mobilizer.can_translate() && !mobilizer.can_rotate())) {
-        // This is a revolute or prismatic joint.
-        s_set.insert(
-            rational_forward_kin
-                .s()[rational_forward_kin
-                         .map_mobilizer_to_s_index()[mobilizer_index]]);
-      } else if (mobilizer.num_velocities() > 0) {
-        throw std::runtime_error(
-            "FindMonomialBasis: we only support revolute, prismatic or weld "
-            "mobilizers.");
-      }
+    const std::vector<int>& s_indices =
+        map_body_pair_to_s_on_chain.at(body_pair);
+    for (const int s_index : s_indices) {
+      s_set.insert(rational_forward_kin.s()[s_index]);
     }
     if (s_set.empty()) {
       // No s variable. The monomial basis is just [1].
@@ -81,12 +67,13 @@ void FindMonomialBasisArray(
     *monomial_basis_array = body_pair_it->second;
   }
 }
-
 }  // namespace
+
 CspaceFreePolytopeBase::CspaceFreePolytopeBase(
     const multibody::MultibodyPlant<double>* plant,
     const geometry::SceneGraph<double>* scene_graph,
-    SeparatingPlaneOrder plane_order, const Options& options)
+    SeparatingPlaneOrder plane_order, SForPlane s_for_plane_enum,
+    const Options& options)
     : rational_forward_kin_(plant),
       scene_graph_{scene_graph},
       link_geometries_{internal::GetCollisionGeometries(*plant, *scene_graph)},
@@ -105,6 +92,11 @@ CspaceFreePolytopeBase::CspaceFreePolytopeBase(
       rational_forward_kin_.plant(), *scene_graph_, link_geometries_,
       &collision_pairs);
 
+  // Initialize map_body_pair_to_s_on_chain.
+  for (const auto& [body_pair, collisions] : collision_pairs) {
+    SetIndicesOfSOnChainForBodyPair(body_pair);
+  }
+
   separating_planes_.reserve(num_collision_pairs);
   const symbolic::Monomial monomial_one{};
   for (const auto& [link_pair, geometry_pairs] : collision_pairs) {
@@ -115,7 +107,7 @@ CspaceFreePolytopeBase::CspaceFreePolytopeBase(
       switch (plane_order) {
         case SeparatingPlaneOrder::kAffine: {
           const VectorX<symbolic::Variable> s_for_plane =
-              rational_forward_kin_.s();
+              GetSForPlane(link_pair, s_for_plane_enum);
           plane_decision_vars.resize(4 * s_for_plane.rows() + 4);
           for (int i = 0; i < plane_decision_vars.rows(); ++i) {
             plane_decision_vars(i) =
@@ -138,6 +130,14 @@ CspaceFreePolytopeBase::CspaceFreePolytopeBase(
           SortedPair<geometry::GeometryId>(geometry_pair.first->id(),
                                            geometry_pair.second->id()),
           static_cast<int>(separating_planes_.size()) - 1);
+
+      // Later we will also need the s variables on the kinematics chain from
+      // the expressed body to either ends of link_pair (for example, when we
+      // compute the monomial basis). Hence we compute it here.
+      SetIndicesOfSOnChainForBodyPair(
+          SortedPair<multibody::BodyIndex>(link_pair.first(), expressed_body));
+      SetIndicesOfSOnChainForBodyPair(
+          SortedPair<multibody::BodyIndex>(link_pair.second(), expressed_body));
     }
   }
 
@@ -161,6 +161,7 @@ void CspaceFreePolytopeBase::CalcMonomialBasis() {
           plane.expressed_body, collision_geometry->body_index());
       std::array<VectorX<symbolic::Monomial>, 4> monomial_basis_array;
       FindMonomialBasisArray(rational_forward_kin_, y_slack_, body_pair,
+                             map_body_pair_to_s_on_chain_,
                              &map_body_to_monomial_basis_array_,
                              &monomial_basis_array);
     }
@@ -182,6 +183,54 @@ void CspaceFreePolytopeBase::CalcSBoundsPolynomial(
     (*s_upper_minus_s)(i) = symbolic::Polynomial(symbolic::Polynomial::MapType(
         {{monomial_one, s_upper(i)}, {symbolic::Monomial(s(i)), -1}}));
   }
+}
+
+void CspaceFreePolytopeBase::SetIndicesOfSOnChainForBodyPair(
+    const SortedPair<multibody::BodyIndex>& body_pair) {
+  if (map_body_pair_to_s_on_chain_.count(body_pair) == 0) {
+    const std::vector<multibody::internal::MobilizerIndex> mobilizer_indices =
+        multibody::internal::FindMobilizersOnPath(rational_forward_kin_.plant(),
+                                                  body_pair.first(),
+                                                  body_pair.second());
+    const auto& tree =
+        multibody::internal::GetInternalTree(rational_forward_kin_.plant());
+    std::vector<int> s_indices;
+    for (const auto& mobilizer_index : mobilizer_indices) {
+      const auto& mobilizer = tree.get_mobilizer(mobilizer_index);
+      if ((mobilizer.num_positions() == 1 && mobilizer.num_velocities() == 1) &&
+          ((mobilizer.can_rotate() && !mobilizer.can_translate()) ||
+           (mobilizer.can_translate() && !mobilizer.can_rotate()))) {
+        // This is a revolute or prismatic joint.
+        s_indices.push_back(
+            rational_forward_kin_.map_mobilizer_to_s_index()[mobilizer_index]);
+      } else if (mobilizer.num_velocities() > 0) {
+        throw std::runtime_error(
+            "FindMonomialBasis: we only support revolute, prismatic or weld "
+            "mobilizers.");
+      }
+    }
+    map_body_pair_to_s_on_chain_.emplace(body_pair, s_indices);
+  }
+}
+
+VectorX<symbolic::Variable> CspaceFreePolytopeBase::GetSForPlane(
+    const SortedPair<multibody::BodyIndex>& body_pair,
+    SForPlane s_for_plane_enum) const {
+  switch (s_for_plane_enum) {
+    case SForPlane::kAll: {
+      return rational_forward_kin_.s();
+    }
+    case SForPlane::kOnChain: {
+      const std::vector<int>& s_indices =
+          map_body_pair_to_s_on_chain_.at(body_pair);
+      VectorX<symbolic::Variable> s_for_plane(s_indices.size());
+      for (int i = 0; i < s_for_plane.rows(); ++i) {
+        s_for_plane(i) = rational_forward_kin_.s()(s_indices[i]);
+      }
+      return s_for_plane;
+    }
+  }
+  DRAKE_UNREACHABLE();
 }
 }  // namespace optimization
 }  // namespace geometry
--- a/geometry/optimization/cspace_free_polytope_base.h
+++ b/geometry/optimization/cspace_free_polytope_base.h
@@ -105,16 +105,27 @@ class CspaceFreePolytopeBase {
   }
 
  protected:
-  // Constructor.
-  // We put the constructor in protected method to make sure that the user
-  // cannot instantiate a CspaceFreePolytopeBase instance.
-  // @pre plant and scene_graph should be non-null pointers.
+  /** When we set up the separating plane {x | a(s)ᵀx + b(s) = 0} between a pair
+   of geometries, we need to determine which s are used in a(s) and b(s).
+   */
+  enum class SForPlane {
+    kAll,      ///< Use all s in the robot tangent-configuration space.
+    kOnChain,  ///< Use s on the kinematics chain between the pair of
+               ///< geometries.
+  };
+
+  /** Constructor.
+   We put the constructor in protected method to make sure that the user
+   cannot instantiate a CspaceFreePolytopeBase instance.
+   @pre plant and scene_graph should be non-null pointers.
+   */
   CspaceFreePolytopeBase(const multibody::MultibodyPlant<double>* plant,
                          const geometry::SceneGraph<double>* scene_graph,
                          SeparatingPlaneOrder plane_order,
+                         SForPlane s_for_plane_enum,
                          const Options& options = Options{});
 
-  // Computes s-s_lower and s_upper - s as polynomials of s.
+  /** Computes s-s_lower and s_upper - s as polynomials of s. */
   void CalcSBoundsPolynomial(
       const Eigen::VectorXd& s_lower, const Eigen::VectorXd& s_upper,
       VectorX<symbolic::Polynomial>* s_minus_s_lower,
@@ -137,6 +148,12 @@ class CspaceFreePolytopeBase {
     return plane_order_;
   }
 
+  /** Maps a pair of body (body1, body2) to an array of monomial basis
+   `monomial_basis_array`. monomial_basis_array[0] contains all the monomials
+   of form ∏ᵢ pow(sᵢ, dᵢ), dᵢ=0 or 1, sᵢ correspond to the revolute/prismatic
+   joint on the kinematic chain between body1 and body2.
+   monomial_basis_array[i+1] = y_slack_[i] * monomial_basis_array[0]
+   */
   [[nodiscard]] const std::unordered_map<
       SortedPair<multibody::BodyIndex>,
       std::array<VectorX<symbolic::Monomial>, 4>>&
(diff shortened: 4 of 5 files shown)


AlexandreAmice (commented):
I am confused. I thought we were already doing this in `CspaceFreePolytope`?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NY3L1Ds-dKJEDqd7BtM:bxnfvnu)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
This is a subtle issue

For CspaceFreePolytope, we want to certify the separating plane {x | a(s)ᵀ x + b(s) = 0} for the TC-space polytope {s | C*s <= d, s_lower <= s <= s_upper}. Since the TC-space polytope involves every entry of `s` in C*s <= d, the separating plane parameters a(s) and b(s) also depends on every entry of `s`.

This ends up certifying a polynomial
```
numerator(s) - λ(s) * (d - C*s) - λ_lower(s) * (s - s_lower) - λ_upper(s) * (s_upper - s) is SOS
```
This polynomial has some structure here, that we can choose the monomial basis to be ∏ᵢ power(sᵢ, dᵢ), where sᵢ are all the s on the half-chain, from the middle of the geometries (A, B) to either geometry A or B.

For CspaceFreeBox, we still want to certify the separating plane {x | a(s)ᵀ x + b(s) = 0}, but for a box s_lower <= s <= s_upper. Now for the geometry pair (A, B), we only need to consider a low-dimensional box on a subset of s (denoted as s_AB, the set of s on the kinematics chain between the geometry A and B), namely s_lower_AB <= s_on_AB <= s_upper_AB. Hence now the separating plane a(s) and b(s) also only need to depend on the subset s_AB, instead of depending on all s.

The SOS polynomial `numerator(s) - λ_lower(s) * (s - s_lower) - λ_upper(s) * (s_upper - s) is SOS` still has the same monomial basis. But we know that some λ_lower(s) and λ_upper(s) should be 0, namely the Lagrangian multipliers corresponding to the s that is not in s_AB.

Does this make sense?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NY3pbTaCIkRCtfogCTV:b2z74o8)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

AlexandreAmice (commented):
Hmm i see. So in previous code, it was only the forward kinematics where we were expressing using the 's' on the kinematic chains between the bodies. 

I believe that in general you could do the exact same trick for the certification step in `CspaceFreePolytope`. The problem would come in when you attempt to grow the regions and you lose some power for the expansion step. I just want to double check that I fully understand the subtlety.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NYEhFS98r9GS6EBJQUz:b-rzyw72)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 184 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYJoFNQ4TGY_m5-psOq:-NYJoFNRCcZPf_ARYp4g:bdv4ql2) ([raw file](https://github.com/RobotLocomotion/drake/blob/290b98fd3ff99ac1f9126468ce7601790d0d2012/geometry/optimization/cspace_free_polytope_base.h#L184)):*
> ```objc
>   void CalcMonomialBasis();
> 
>   /* Set map_body_pair_to_s_on_chain;
> ```

I think this needs a more descriptive name and doc string.

Can we change it to `ComputeIndicesOfSOnChainForBodyPair`. If this is too long, then a longer comment would be very helpful. 
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYJoVOmBmUkU1O__aCw:-NYJoVOmBmUkU1O__aCx:bx6m3ea) ([raw file](https://github.com/RobotLocomotion/drake/blob/290b98fd3ff99ac1f9126468ce7601790d0d2012/geometry/optimization/cspace_free_polytope_base.cc#L94)):*
> ```cpp
>       rational_forward_kin_.plant(), *scene_graph_, link_geometries_,
>       &collision_pairs);
> ```

I would include a comment here
```
// initializing map_geometries_to_pair_to_s_on_chain
```

hongkai-dai (commented):
> I believe that in general you could do the exact same trick for the certification step in `CspaceFreePolytope`

I don't think it would work nicely with CspaceFreePolytope. Long story short, because the polytope `C*s <= d` involves every `s`, hence the separating plane also need to involve every `s`.

To make it concrete, let's consider the following example:
Say we have a 3-joint robot, `body0 <-> joint0 <-> body1 <-> joint1 <-> body2 <-> joint2 <-> body3`. Now we consider the separation between body0 and body 2. We compute the pose of the body0 in the expressed frame body 1 (in the middle between body0 and body2). This pose `X_10` is a rational function of `s0`. Specifically, the numerator of the pose will contain monomials of the form `1, s0, s0²`.

Let's consider a TC-space polytope `C*s <= d`. Note that here in each row of `C*s`, it will always include s1 and s2, which do not show up in the pose`X_10`. Now we need to design the separating plane`{x | a(s)ᵀx + b(s) = 0}`, such that the numerator of a(s)ᵀ* p(s) + b(s), which we denote as `numerator(s)`,  is non-negative in the polytope `C*s<=d`. Namely
```
numerator(s) - λ(s)ᵀ(d−Cs) is sos
λ(s) is sos
```
The question is which `s` should be included in `a(s), b(s)`?

If we only use the `s` on the kinematics chain from body 0 to body 2  (namely `s0` and `s1`), then in `numerator(s)`, it contains monomials of `1, s0, s0², s0³, s1, s0*s1, s0²*s1`.  As we have discussed before, the Lagrangian multiplier λ(s) will contain the monomials of s on the "half kinematics chain"; in this case λ(s) contains `1, s0, s0²`. As a result, λ(s)ᵀ(d−Cs) would contain the term `1, s0, s0², s0³,s1, s0*s1, s0²*s1,s2, s0*s2, s0²*s2`. Notice that the monomials `s2, s0*s2, s0²*s2` show up in λ(s)ᵀ(d−Cs) but not show up in `numerator(s)`. This mismatch on the monomials might cause the SOS problem to fail.

On the other hand, for a C-space box`s_lower <= s <= s_upper`, to prove the separation between body0 and body 2, we only need to consider the lower-dimensional box `s_lower(0) <= s(0) <= s_upper(0), s_lower(1) <= s(1) <= s_upper(1)`. We don't need to consider the dimension s(2) which doesn't affect the kinematic pose `X_10`. As a result, we don't need the separating plane a(s) and b(s) to depend on s(2).

I am sorry that this is quite subtle. Does this make sense to you?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NYKn6P81ca_5yUlUTQk:brfiaxn)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 184 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYJoFNQ4TGY_m5-psOq:-NYKspjg2G6nwYFcuzQw:b-3g1qx6) ([raw file](https://github.com/RobotLocomotion/drake/blob/290b98fd3ff99ac1f9126468ce7601790d0d2012/geometry/optimization/cspace_free_polytope_base.h#L184)):*

I think this needs a more descriptive name and doc string.

Can we change it to `ComputeIndicesOfSOnChainForBodyPair`. If this is too long, then a longer comment would be very helpful. 
</blockquote></details>

Done. I changed the name and expand the documentation.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 94 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYJoVOmBmUkU1O__aCw:-NYKt7mg7kZHxvwDUuK_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/290b98fd3ff99ac1f9126468ce7601790d0d2012/geometry/optimization/cspace_free_polytope_base.cc#L94)):*

I would include a comment here
```
// initializing map_geometries_to_pair_to_s_on_chain
```
</blockquote></details>

Done.

AlexandreAmice (commented):
Yes it makes sense to me. What I am suggesting is that in the general case, for every pair you could compute a separation certificate for the projection of `Cs <= d` onto the half-kinematic chain and not attempt to certify all of `Cs <= d`. But it gets complicated to do so. 

Either way, its a very nice catch on your part!

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NYLThIi3k99a9z3qBvY:b29ry3c)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
I see, that is a good point to project `C*s<=d` to the lower dimensional space.

But as you said, it makes the later bilinear alternation really hard, as now the Lagrangian multiplier λ(s) is for the projected `C_project * s_project <= d_project`, not for the original C-space polytope, hence we cannot search the original C-space polytope with this Lagrangian multiplier λ(s).

+@jwnimmer-tri for Tuesday's platform review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NYLjd3k5ET-B0Gfwcp3:bht5hl1)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri (commented):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPreKu1a7taWFxV3Rw:-NYPreKu1a7taWFxV3Rx:bkdenx5) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L21)):*
> ```objc
>  the common functionality between these concrete derived class to this shared
>  parent class.
>  */
> ```

BTW It's a bit weird that the `protected:` functions have non-Doxygen documentation, yet the class allows end-users to subclass it.  That's somewhat inconsistent.  Usually we would document protected functions for end-users, since they could call them from their own subclasses.

If the design intention is that _only_ Drake code is allowed to subclass this class, then the class overview doc here should document that restriction for users.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 123 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPsDNHEFx0mq14XcTB:-NYPsDNHEFx0mq14XcTC:b-tiylzc) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L123)):*
> ```objc
>                          const geometry::SceneGraph<double>* scene_graph,
>                          SeparatingPlaneOrder plane_order,
>                          SForPlane s_for_plane_enum,
> ```

BTW If you think it helps readability you can keep it as-is, but I'll point out that putting `_enum` in a variable name is a little bit weird (throughout).



_Suggestion:_
```objc
SForPlane s_for_plane
```
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 158 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPsfq6E4dseTlE0Btx:-NYPsfq6E4dseTlE0Bty:bc9n0bb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L158)):*
> ```objc
>   [[nodiscard]] bool with_cross_y() const { return with_cross_y_; }
> 
>   [[nodiscard]] const std::unordered_map<SortedPair<multibody::BodyIndex>,
> ```

nit It's probably worth explaining what this map denotes in an API comment, to make it easier for developers of subclass codes to understand it?

Possibly relocating the member field comment to here would suffice.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 222 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPt4qq7jINgXaCN9yz:-NYPt4qq7jINgXaCN9z-:b-s7emhs) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L222)):*
> ```objc
> 
>   // For a pair of bodies body_pair, returns the indices of all s on the
>   // kinematics chain from body_pair.first() to body_pair.second().
> ```

nit From this comment, its unclear to me which body pairs are keys in this map.  Is it the full `N*(N-1)/2` set of all bodies?  I think probably not, given the other code nearby.  In practice, which body pairs will live here?  Obviously its whatever `SetIndicesOfSOnChainForBodyPair` got called on, but which pairs is that called on?  Help the reader put together all of the pieces.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPtgVx6Na_WvezKoKQ:-NYPtgVx6Na_WvezKoKR:b-wvqb9a) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.cc#L35)):*
> ```cpp
>     const Vector3<symbolic::Variable>& y_slack,
>     const SortedPair<multibody::BodyIndex>& body_pair,
>     const std::unordered_map<SortedPair<multibody::BodyIndex>, std::vector<int>>
> ```

nit This is missing a `&` so its making a copy.  That's probably not what you wanted to do?
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPubpwC-7L3c0LGqs2:-NYPubpx3WkhbokPcj80:boep016) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.cc#L202)):*
> ```cpp
>       if ((mobilizer.can_rotate() && !mobilizer.can_translate()) ||
>           (mobilizer.can_translate() && !mobilizer.can_rotate())) {
>         // This is a revolute or prismatic joint.
> ```

I would have thought that we also need to check `mobilizer.num_positions() == 1 && mobilizer.num_velocities() == 1` like we did in `rational_forward_kinematics.cc`.  Is that intentionally omitted here?
___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 118 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPvOhy0XvOFidIs5mg:-NYPvOhy0XvOFidIs5mh:b-bzm7qt) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L118)):*
> ```cpp
>     for (int i = 0; i < static_cast<int>(s_indices.size()); ++i) {
>       EXPECT_EQ(s_indices[i], s_indices_expected[i]);
>     }
> ```

Comparing vectors with a loop seems a bit verbose.  Doesn't `EXPECT_EQ(s_indices, s_indices_expected)` already do what we need?  It should print out fine, also, for any failures.

I wonder if we even need this `compare_s_indices` helper.  It might be most readable if it were written without the lambda.

The `testing::ElementsAre()` matcher (see http://google.github.io/googletest/reference/matchers.html#container-matchers) might be most readable.

```c++
EXPECT_THAT(
    map_body_pair_to_s_on_chain.at({world_index, body_indices_[1]}),
    testing::ElementsAre(0));
```
___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPwV4c9krJwSa4gy6S:-NYPwV4c9krJwSa4gy6T:bb3r0t4) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L157)):*
> ```cpp
>       EXPECT_EQ(s_in_a.size(), s_expected.size());
>       for (const auto& s : s_in_a) {
>         EXPECT_NE(s_expected.find(s), s_expected.end());
> ```

This is arcane enough that using `testing::UnorderedElementsAre` is probably a lot more readable?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPwV4c9krJwSa4gy6S:-NYPwvcl45XCcgQFZzKx:bd96wwd) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L157)):*

This is arcane enough that using `testing::UnorderedElementsAre` is probably a lot more readable?
</blockquote></details>

Well, `UnorderedElementsAreArray` to be specific.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPreKu1a7taWFxV3Rw:-NYVQJ4w3hWg8bIVkFSS:bbk7o6z) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L21)):*

BTW It's a bit weird that the `protected:` functions have non-Doxygen documentation, yet the class allows end-users to subclass it.  That's somewhat inconsistent.  Usually we would document protected functions for end-users, since they could call them from their own subclasses.

If the design intention is that _only_ Drake code is allowed to subclass this class, then the class overview doc here should document that restriction for users.
</blockquote></details>

Done. I changed the documentation to `/** ... */`. I didn't know that `protected` should also have Doxygen documentation.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 123 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPsDNHEFx0mq14XcTB:-NYVQVHUAJXM543lT3D6:b-tcotnf) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L123)):*

BTW If you think it helps readability you can keep it as-is, but I'll point out that putting `_enum` in a variable name is a little bit weird (throughout).
</blockquote></details>

I think if we call the variable name as `s_for_plane` it would be confusing, as it would mean the set of variable `s` for this plane, instead of the enum. We actually have `s_for_plane` in our code to mean the set of variable `s` for this plane.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 222 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPt4qq7jINgXaCN9yz:-NYVRho89ZblIVoiPHlM:bc88kz7) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L222)):*

nit From this comment, its unclear to me which body pairs are keys in this map.  Is it the full `N*(N-1)/2` set of all bodies?  I think probably not, given the other code nearby.  In practice, which body pairs will live here?  Obviously its whatever `SetIndicesOfSOnChainForBodyPair` got called on, but which pairs is that called on?  Help the reader put together all of the pieces.
</blockquote></details>

Done. I added the documentation.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPtgVx6Na_WvezKoKQ:-NYVS8TrEo6Ud_z15vHR:b-376vyb) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.cc#L35)):*

nit This is missing a `&` so its making a copy.  That's probably not what you wanted to do?
</blockquote></details>

Done. Good catch.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPubpwC-7L3c0LGqs2:-NYVSNBwC_wG7kfrYeSB:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.cc#L202)):*

I would have thought that we also need to check `mobilizer.num_positions() == 1 && mobilizer.num_velocities() == 1` like we did in `rational_forward_kinematics.cc`.  Is that intentionally omitted here?
</blockquote></details>

Done. Good call.
___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 118 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPvOhy0XvOFidIs5mg:-NYVSo3H5-Ss2mNqPgnv:b-gpvzgl) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L118)):*

Comparing vectors with a loop seems a bit verbose.  Doesn't `EXPECT_EQ(s_indices, s_indices_expected)` already do what we need?  It should print out fine, also, for any failures.

I wonder if we even need this `compare_s_indices` helper.  It might be most readable if it were written without the lambda.

The `testing::ElementsAre()` matcher (see http://google.github.io/googletest/reference/matchers.html#container-matchers) might be most readable.

```c++
EXPECT_THAT(
    map_body_pair_to_s_on_chain.at({world_index, body_indices_[1]}),
    testing::ElementsAre(0));
```
</blockquote></details>

Done!

This is cool! Thanks Jeremy
___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPwV4c9krJwSa4gy6S:-NYVWCQX74jloyw9r66v:b-baxok6) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L157)):*

Well, `UnorderedElementsAreArray` to be specific.
</blockquote></details>

`UnorderedElementsAre` doesn't work for symbolic::Variables.

I changed to `EXPECT_TRUE(s_expected.include(s))` which I think is more readable?

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-:-NYWROij8AkxHjHTPAie:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),AlexandreAmice

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPreKu1a7taWFxV3Rw:-NYWPxMO6TMvCJQ9lfue:bt99zzu) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L21)):*

Done. I changed the documentation to `/** ... */`. I didn't know that `protected` should also have Doxygen documentation.
</blockquote></details>

Yup, "protected" shows up on the website, e.g., see LeafSystem:
https://drake.mit.edu/doxygen_cxx/classdrake_1_1systems_1_1_leaf_system.html
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 123 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPsDNHEFx0mq14XcTB:-NYWQD4E1jRNzHCOOscp:b53dsna) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/cspace_free_polytope_base.h#L123)):*

I think if we call the variable name as `s_for_plane` it would be confusing, as it would mean the set of variable `s` for this plane, instead of the enum. We actually have `s_for_plane` in our code to mean the set of variable `s` for this plane.
</blockquote></details>

That's fine.  But maybe the typename `SForPlane` could be improved given that thought.
___
*[`geometry/optimization/test/cspace_free_polytope_base_test.cc` line 157 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYPwV4c9krJwSa4gy6S:-NYWR3XP5TCEyLpCww4o:bytqsbv) ([raw file](https://github.com/RobotLocomotion/drake/blob/4ef7d6946631d46a87c30958946d9ad3423307ac/geometry/optimization/test/cspace_free_polytope_base_test.cc#L157)):*

`UnorderedElementsAre` doesn't work for symbolic::Variables.

I changed to `EXPECT_TRUE(s_expected.include(s))` which I think is more readable?
</blockquote></details>

I agree.  I forgot about the overloaded `operator==` in symbolic code causing havoc with googletest.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19609)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),AlexandreAmice

___
*[`geometry/optimization/cspace_free_polytope_base.h` line 125 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19609#-NYWRhI-1MAtMewf15eW:-NYWRhI-1MAtMewf15eX:b-mtunho) ([raw file](https://github.com/RobotLocomotion/drake/blob/4fd339bd54791aabe83587b0a785d06fe2e3b192/geometry/optimization/cspace_free_polytope_base.h#L125)):*
> ```objc
>                          const geometry::SceneGraph<double>* scene_graph,
>                          SeparatingPlaneOrder plane_order,
>                          SForPlane s_for_plane_enum,
> ```

FYI For reference -- this is _not_ a breaking change because this constructor has not yet appeared in two stable releases.  It's only a week old and has only been in one release (1.18).

PR #19632, opened 2023-06-19
Implements MakeSemidefiniteRelaxation

RussTedrake:
Given a (typically nonconvex) MathematicalProgram, this method constructs a new MathematicalProgram which implements the SDP relaxation of the original program.

This initial implementation supports the case of quadratic objectives and constraints; it's an extremely important (and quite general) case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)

### DIFF
--- a/bindings/pydrake/solvers/BUILD.bazel
+++ b/bindings/pydrake/solvers/BUILD.bazel
@@ -68,6 +68,7 @@ drake_pybind_library(
         "solvers_py_osqp.cc",
         "solvers_py_scs.cc",
         "solvers_py_sdpa_free_format.cc",
+        "solvers_py_semidefinite_relaxation.cc",
         "solvers_py_snopt.cc",
         "solvers_py_unrevised_lemke.cc",
     ],
@@ -261,6 +262,14 @@ drake_py_unittest(
 
 drake_py_unittest(
     name = "sdpa_free_format_test",
+    deps = [
+        ":solvers",
+        "//bindings/pydrake/common/test_utilities",
+    ],
+)
+
+drake_py_unittest(
+    name = "semidefinite_relaxation_test",
     deps = [
         ":solvers",
         "//bindings/pydrake/common/test_utilities",
--- a/bindings/pydrake/solvers/solvers_py.cc
+++ b/bindings/pydrake/solvers/solvers_py.cc
@@ -24,6 +24,7 @@ top-level documentation for :py:mod:`pydrake.math`.
   internal::DefineSolversMixedIntegerOptimizationUtil(m);
   internal::DefineSolversMixedIntegerRotationConstraint(m);
   internal::DefineSolversSdpaFreeFormat(m);
+  internal::DefineSolversSemidefiniteRelaxation(m);
   internal::DefineSolversClp(m);
   internal::DefineSolversCsdp(m);
   internal::DefineSolversGurobi(m);
--- a/bindings/pydrake/solvers/solvers_py.h
+++ b/bindings/pydrake/solvers/solvers_py.h
@@ -60,6 +60,9 @@ void DefineSolversScs(py::module m);
 /* Defines bindings per solvers_py_sdpa_free_format.cc. */
 void DefineSolversSdpaFreeFormat(py::module m);
 
+/* Defines bindings per solvers_py_semidefinite_relaxation.cc. */
+void DefineSolversSemidefiniteRelaxation(py::module m);
+
 /* Defines the SNOPT bindings. See solvers_py_snopt.cc. */
 void DefineSolversSnopt(py::module m);
 
--- a/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
+++ b/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc
@@ -1471,6 +1471,14 @@ for every column of ``prog_var_vals``. )""")
           py::arg("binding"), py::arg("prog_var_vals"),
           doc.MathematicalProgram.EvalBinding.doc)
       .def(
+          "EvalBindingAtInitialGuess",
+          [](const MathematicalProgram& prog,
+              const Binding<EvaluatorBase>& binding) {
+            return prog.EvalBindingAtInitialGuess(binding);
+          },
+          py::arg("binding"),
+          doc.MathematicalProgram.EvalBindingAtInitialGuess.doc)
+      .def(
           "EvalBindings",
           [](const MathematicalProgram& prog,
               const std::vector<Binding<EvaluatorBase>>& binding,
--- a/bindings/pydrake/solvers/solvers_py_semidefinite_relaxation.cc
+++ b/bindings/pydrake/solvers/solvers_py_semidefinite_relaxation.cc
@@ -1,0 +1,21 @@
+#include "drake/bindings/pydrake/documentation_pybind.h"
+#include "drake/bindings/pydrake/pydrake_pybind.h"
+#include "drake/bindings/pydrake/solvers/solvers_py.h"
+#include "drake/solvers/semidefinite_relaxation.h"
+
+namespace drake {
+namespace pydrake {
+namespace internal {
+
+void DefineSolversSemidefiniteRelaxation(py::module m) {
+  // NOLINTNEXTLINE(build/namespaces): Emulate placement in namespace.
+  using namespace drake::solvers;
+  constexpr auto& doc = pydrake_doc.drake.solvers;
+
+  m.def("MakeSemidefiniteRelaxation", &solvers::MakeSemidefiniteRelaxation,
+      py::arg("prog"), doc.MakeSemidefiniteRelaxation.doc);
+}
+
+}  // namespace internal
+}  // namespace pydrake
+}  // namespace drake
--- a/bindings/pydrake/solvers/test/semidefinite_relaxation_test.py
+++ b/bindings/pydrake/solvers/test/semidefinite_relaxation_test.py
@@ -1,0 +1,24 @@
+import numpy as np
+import unittest
+
+from pydrake.solvers import (
+    MakeSemidefiniteRelaxation,
+    MathematicalProgram,
+)
+
+
+class TestSemidefiniteRelaxation(unittest.TestCase):
+    def test_MakeSemidefiniteRelaxation(self):
+        prog = MathematicalProgram()
+        y = prog.NewContinuousVariables(2, "y")
+        A = np.array([[0.5, 0.7], [-.2, 0.4], [-2.3, -4.5]])
+        lb = np.array([1.3, -.24, 0.25])
+        ub = np.array([5.6, 0.1, 1.4])
+        prog.AddLinearConstraint(A, lb, ub, y)
+        relaxation = MakeSemidefiniteRelaxation(prog)
+
+        self.assertEqual(relaxation.num_vars(), 6)
+        self.assertEqual(len(relaxation.positive_semidefinite_constraints()),
+                         1)
+        self.assertEqual(len(relaxation.bounding_box_constraints()), 1)
+        self.assertEqual(len(relaxation.linear_constraints()), 2)
--- a/geometry/optimization/spectrahedron.cc
+++ b/geometry/optimization/spectrahedron.cc
@@ -31,7 +31,7 @@ const double kInf = std::numeric_limits<double>::infinity();
 // TODO(russt): This can be replaced by vars(indices) once we have Eigen 3.4.
 VectorXDecisionVariable GetVariablesByIndex(
     const Eigen::Ref<const VectorXDecisionVariable>& vars,
-    std::vector<int> indices) {
+    const std::vector<int>& indices) {
   VectorXDecisionVariable new_vars(indices.size());
   for (int i = 0; i < ssize(indices); ++i) {
     new_vars[i] = vars[indices[i]];
@@ -126,7 +126,7 @@ Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
     VectorXDecisionVariable vars = stack_xt(binding.variables());
     MatrixXd Ab(binding.evaluator()->num_constraints(), vars.size());
     Ab.leftCols(binding.evaluator()->num_vars()) =
-        binding.evaluator()->GetDenseA();
+        binding.evaluator()->get_sparse_A();
     Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
     constraints.emplace_back(prog->AddLinearEqualityConstraint(Ab, 0, vars));
   }
@@ -143,7 +143,7 @@ Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
     VectorXDecisionVariable vars = stack_xt(binding.variables());
     MatrixXd Ab(binding.evaluator()->num_constraints(), vars.size());
     Ab.leftCols(binding.evaluator()->num_vars()) =
-        binding.evaluator()->GetDenseA();
+        binding.evaluator()->get_sparse_A();
     if (binding.evaluator()->lower_bound().array().isFinite().any()) {
       Ab.rightCols<1>() = -binding.evaluator()->lower_bound();
       constraints.emplace_back(prog->AddLinearConstraint(Ab, 0, kInf, vars));
@@ -203,11 +203,11 @@ Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
     const int num_constraints = binding.evaluator()->num_constraints();
     std::tie(this_A, this_b) = binding_Ab(binding.variables());
     MatrixXd A_bar(num_constraints, x.size() + t.size());
-    A_bar.leftCols(x.size()) = binding.evaluator()->GetDenseA() * this_A;
+    A_bar.leftCols(x.size()) = binding.evaluator()->get_sparse_A() * this_A;
     A_bar.rightCols(t.size()) =
         -binding.evaluator()->upper_bound() * c.transpose();
     VectorXd b_bar = d * binding.evaluator()->upper_bound() -
-                     binding.evaluator()->GetDenseA() * this_b;
+                     binding.evaluator()->get_sparse_A() * this_b;
     constraints.emplace_back(
         prog->AddLinearEqualityConstraint(A_bar, b_bar, {x, t}));
   }
@@ -225,9 +225,9 @@ Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
     const int num_constraints = binding.evaluator()->num_constraints();
     std::tie(this_A, this_b) = binding_Ab(binding.variables());
     MatrixXd A_bar(num_constraints, x.size() + t.size());
-    A_bar.leftCols(x.size()) = binding.evaluator()->GetDenseA() * this_A;
+    A_bar.leftCols(x.size()) = binding.evaluator()->get_sparse_A() * this_A;
     VectorXd b_bar = d * binding.evaluator()->lower_bound() -
-                     binding.evaluator()->GetDenseA() * this_b;
+                     binding.evaluator()->get_sparse_A() * this_b;
     if (b_bar.array().isFinite().any()) {
       A_bar.rightCols(t.size()) =
           -binding.evaluator()->lower_bound() * c.transpose();
@@ -239,7 +239,7 @@ Spectrahedron::DoAddPointInNonnegativeScalingConstraints(
     // [A_c * A, -b_c * c'][x;t] ≤ d * b_c - A_c * b
     // A̅ [x;t] ≤ b̅,
     b_bar = d * binding.evaluator()->upper_bound() -
-            binding.evaluator()->GetDenseA() * this_b;
+            binding.evaluator()->get_sparse_A() * this_b;
     if (b_bar.array().isFinite().any()) {
       A_bar.rightCols(t.size()) =
           -binding.evaluator()->upper_bound() * c.transpose();
--- a/solvers/BUILD.bazel
+++ b/solvers/BUILD.bazel
@@ -61,6 +61,7 @@ drake_cc_package_library(
         ":rotation_constraint",
         ":scs_solver",
         ":sdpa_free_format",
+        ":semidefinite_relaxation",
         ":snopt_solver",
         ":solution_result",
         ":solve",
@@ -961,6 +962,16 @@ drake_cc_library(
     deps = [],
 )
 
+drake_cc_library(
+    name = "semidefinite_relaxation",
+    srcs = ["semidefinite_relaxation.cc"],
+    hdrs = ["semidefinite_relaxation.h"],
+    interface_deps = [
+        ":mathematical_program",
+    ],
+    deps = [],
+)
+
 drake_cc_optional_library(
     name = "csdp_solver_error_handling",
     opt_out_condition = "//tools:no_csdp",
@@ -1137,6 +1148,15 @@ drake_cc_googletest(
         ":csdp_test_examples",
         ":sdpa_free_format",
         "//common:temp_directory",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
+    ],
+)
+
+drake_cc_googletest(
+    name = "semidefinite_relaxation_test",
+    deps = [
+        ":semidefinite_relaxation",
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
     ],
--- a/solvers/semidefinite_relaxation.cc
+++ b/solvers/semidefinite_relaxation.cc
@@ -1,0 +1,224 @@
+#include "drake/solvers/semidefinite_relaxation.h"
+
+#include <initializer_list>
+#include <limits>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "drake/common/ssize.h"
+#include "drake/common/text_logging.h"
+#include "drake/solvers/program_attribute.h"
+
+namespace drake {
+namespace solvers {
+
+using Eigen::MatrixXd;
+using Eigen::SparseMatrix;
+using Eigen::Triplet;
+using Eigen::VectorXd;
+using symbolic::Expression;
+using symbolic::Variable;
+
+namespace {
+
+const double kInf = std::numeric_limits<double>::infinity();
+
+}  // namespace
+
+std::unique_ptr<MathematicalProgram> MakeSemidefiniteRelaxation(
+    const MathematicalProgram& prog) {
+  std::string unsupported_message{};
+  const ProgramAttributes supported_attributes(
+      std::initializer_list<ProgramAttribute>{
+          ProgramAttribute::kLinearCost, ProgramAttribute::kQuadraticCost,
+          ProgramAttribute::kLinearConstraint,
+          ProgramAttribute::kLinearEqualityConstraint,
+          ProgramAttribute::kQuadraticConstraint});
+  if (!AreRequiredAttributesSupported(prog.required_capabilities(),
+                                      supported_attributes,
+                                      &unsupported_message)) {
+    throw std::runtime_error(fmt::format(
+        "MakeSemidefiniteRelaxation() does not (yet) support this program: {}.",
+        unsupported_message));
+  }
+
+  auto relaxation = std::make_unique<MathematicalProgram>();
+
+  // Build a symmetric matrix X of decision variables using the original
+  // program variables (so that GetSolution, etc, works using the original
+  // variables).
+  relaxation->AddDecisionVariables(prog.decision_variables());
+  MatrixX<Variable> X(prog.num_vars() + 1, prog.num_vars() + 1);
+  // X = xxᵀ; x = [prog.decision_vars(); 1].
+  X.topLeftCorner(prog.num_vars(), prog.num_vars()) =
+      relaxation->NewSymmetricContinuousVariables(prog.num_vars(), "Y");
+  X.topRightCorner(prog.num_vars(), 1) = prog.decision_variables();
+  X.bottomLeftCorner(1, prog.num_vars()) =
+      prog.decision_variables().transpose();
+  // X(-1,-1) = 1.
+  Variable one("one");
+  X(prog.num_vars(), prog.num_vars()) = one;
+  relaxation->AddDecisionVariables(Vector1<Variable>(one));
+  relaxation->AddBoundingBoxConstraint(1, 1,
+                                       X(prog.num_vars(), prog.num_vars()));
+  // X ≽ 0.
+  relaxation->AddPositiveSemidefiniteConstraint(X);
+
+  auto x = X.col(prog.num_vars());
+
+  // Returns the {a, vars} in relaxation, such that a' vars = 0.5*tr(QY). This
+  // assumes Q=Q', which is ensured by QuadraticCost and QuadraticConstraint.
+  auto half_trace_QY = [&X, &prog](const Eigen::MatrixXd& Q,
+                                   const VectorXDecisionVariable& prog_vars)
+      -> std::pair<VectorXd, VectorX<Variable>> {
+    const int N = prog_vars.size();
+    const int num_vars = N * (N + 1) / 2;
+    const std::vector<int> indices =
+        prog.FindDecisionVariableIndices(prog_vars);
+    VectorXd a = VectorXd::Zero(num_vars);
+    VectorX<Variable> y(num_vars);
+    int count = 0;
+    for (int i = 0; i < N; ++i) {
+      for (int j = 0; j <= i; ++j) {
+        // tr(QY) = ∑ᵢ ∑ⱼ Qᵢⱼ Yⱼᵢ.
+        a[count] = ((i == j) ? 0.5 : 1.0) * Q(i, j);
+        y[count] = X(indices[i], indices[j]);
+        ++count;
+      }
+    }
+    return {a, y};
+  };
+
+  // Linear costs => Linear costs.
+  for (const auto& binding : prog.linear_costs()) {
+    relaxation->AddCost(binding);
+  }
+  // Quadratic costs.
+  // 0.5 y'Qy + b'y + c => 0.5 tr(QY) + b'y + c
+  for (const auto& binding : prog.quadratic_costs()) {
+    const int N = binding.variables().size();
+    const int num_vars = N + (N * (N + 1) / 2);
+    std::pair<VectorXd, VectorX<Variable>> quadratic_terms =
+        half_trace_QY(binding.evaluator()->Q(), binding.variables());
+    VectorXd a(num_vars);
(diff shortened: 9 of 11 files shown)


hongkai-dai (commented):
I left some high-level comment. Maybe we can discuss the design in more details through a github issue?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NYMC_OC30JkdeafXvVK:b-xzq4sn)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMArAg5jTDHmS5iS_S:-NYMArAg5jTDHmS5iS_T:b-fzcscx) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.h#L21)):*
> ```objc
>  linear nor quadratic.
>  */
> std::unique_ptr<MathematicalProgram> MakeSemidefiniteRelaxation(
> ```

Do we lose the mapping from the decision variables in the old program to the new decision variables in the new program?

I think often I want to know that if I have a quadratic constraint, then how the decision variables in this quadratic constraint is mapped to the rank-1 PSD matrix in the new constraint.
___
*[`solvers/semidefinite_relaxation.cc` line 26 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBFfhD8tcEJ9LtS72:-NYMBFfhD8tcEJ9LtS73:b2xw47e) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L26)):*
> ```cpp
> VectorXDecisionVariable GetVariablesByIndex(
>     const Eigen::Ref<const VectorXDecisionVariable>& vars,
>     std::vector<int> indices) {
> ```

`indices` should be pass by reference?
___
*[`solvers/semidefinite_relaxation.cc` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBXQA07V7rKrKQ5B0:-NYMBXQA07V7rKrKQ5B1:b-gj1tyk) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L60)):*
> ```cpp
>   relaxation->AddDecisionVariables(prog.decision_variables());
>   MatrixX<Variable> X(prog.num_vars() + 1, prog.num_vars() + 1);
>   // X = xxᵀ; x = [prog.decision_vars(); 1].
> ```

Not all variables need this PSD matrix term, only the variables that appear in the quadratic terms would need this capital `X`.

It is actually more subtle than including all the variables in the quadratic term and create a single `X`. Often there is quite a lot of sparsity we can exploit. For example, if we have x² and y², but not the cross term xy in our constraint, then it is better to create two small PSD matrix
```
[x² x]
[x  1]
```
and
```
[y² y]
[y  1]
```
rather than a single large matrix
```
[x² xy x]
[xy y² y]
[x  y  1]
```

RussTedrake (commented):
Thanks.   I'd prefer to discuss it here rather than create a separate issue.

I got some helpful sideband comments from @bernhardpg that I will address now, too.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NYNn7EVCmvtBl7cSMZs:b-us5fs6)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMArAg5jTDHmS5iS_S:-NYNnHTG35Dm_715jYw4:buv85za) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.h#L21)):*

Do we lose the mapping from the decision variables in the old program to the new decision variables in the new program?

I think often I want to know that if I have a quadratic constraint, then how the decision variables in this quadratic constraint is mapped to the rank-1 PSD matrix in the new constraint.
</blockquote></details>

The variables in the original program are again variables in this program, so the solution can be retrieved easily.  Currently the new decision variables corresponding to the quadratic terms are all lumped into one slack matrix, Y.  

I thought about creating Y manually to create more descriptive variable names.  I had not thought of returning any form of mapping.  I'm not against the idea, but would need a proposal for a reasonable API.
___
*[`solvers/semidefinite_relaxation.cc` line 26 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBFfhD8tcEJ9LtS72:-NYNo0t8AiCBWime-udZ:bkz5ckv) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L26)):*

`indices` should be pass by reference?
</blockquote></details>

Done.  Good catch.
___
*[`solvers/semidefinite_relaxation.cc` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBXQA07V7rKrKQ5B0:-NYNo6Ob50JQQ6-NCnnn:b-uqr4rz) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L60)):*

Not all variables need this PSD matrix term, only the variables that appear in the quadratic terms would need this capital `X`.

It is actually more subtle than including all the variables in the quadratic term and create a single `X`. Often there is quite a lot of sparsity we can exploit. For example, if we have x² and y², but not the cross term xy in our constraint, then it is better to create two small PSD matrix
```
[x² x]
[x  1]
```
and
```
[y² y]
[y  1]
```
rather than a single large matrix
```
[x² xy x]
[xy y² y]
[x  y  1]
```
</blockquote></details>

If the variables are coupled even through a linear constraint Az<=b, z=[x;y], then the strengthening constraints (Az-b)(Az-b)' >= 0 already has the cross terms, and it adds strength to the relaxation.

If the variables are completely decoupled in the program in every way, then the author should write two separate programs.  I don't think we need to cater to that case?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: 2 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMArAg5jTDHmS5iS_S:-NYPS8LC-Xn6ijikNkRO:b-y6p5gd) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.h#L21)):*

In the future, the original program might already contain some PSD constraints, so it is hard to know in the new program, which PSD constraint is from the original program, and which is from the new program. I think it would be more clear to return both the new program, together with this new PSD matrix.
</blockquote></details>

So the proposal is to return Y in addition to the program?  I think returning the std::pair would make the most common call sights less clean.  The reason to potentially do this now would be if we don't want the return value to change type (since that's hard to deprecate) in the future.  But I would propose that if and when we need that ability, I would prefer to add the additional return value as an optional (pointer) argument.  Sound ok?
___
*[`solvers/semidefinite_relaxation.cc` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBXQA07V7rKrKQ5B0:-NYPQd4P5IQoimGGqwri:bp1a7eg) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L60)):*

I still don't think we have to create this large PSD matrix for all the decision variables, as this might lead to a too large PSD constraint, which really doesn't scale well. One example is that if the variables are coupled through PSD constraint, for example we have the BMI
```
f(x) - λ₁(x)g₁(x) − λ₂(x)g₂(x) is sos
```
where the polynomials `f(x),  λ₁, (x), g₁(x), λ₂(x), g₂(x)` are all decision variables. In this case the bilinear product only appears in `λ₁(x)g₁(x)` and `λ₂(x)g₂(x)`. It seems to me that it would be a waste of computation if we consider the cross product `λ₁(x)λ₂(x)`, as would happen if we create a big PSD matrix that includes all the decision variables. Note that in this case we don't have any linear constraint involving λ₁(x) or λ₂(x), hence we won't create the coupling between them through multiplying the linear constraint `(A*z-b) * (A*z-b)`.

This might be a future problem as here it seems we only want to handle the bilinear product shown up in quadratic constraints/costs, not in the PSD constraints, so it should be fine for the moment, but I would like to throw this out for discussion.
</blockquote></details>

Even in your specific example, we should probably be multiplying those constraints together to get the tightest quadratic approximation.  I'm going to stop short of working out the details, because...

I feel very strongly that we have underestimated the strength of the (first-level) SDP relaxation in the past by cutting corners like this.  I know you're right, and I agree that the SDPs can get big fast, but I want to establish the foundation to be the correct complete relaxation; we can add options to cut corners later -- and compare against this implementation to understand more precisely what we might be giving up.

hongkai-dai (commented):
> Reviewed 1 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: 7 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMArAg5jTDHmS5iS_S:-NYP_PBUAxjUcm2-RD7Z:b85ldz5) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.h#L21)):*

So the proposal is to return Y in addition to the program?  I think returning the std::pair would make the most common call sights less clean.  The reason to potentially do this now would be if we don't want the return value to change type (since that's hard to deprecate) in the future.  But I would propose that if and when we need that ability, I would prefer to add the additional return value as an optional (pointer) argument.  Sound ok?
</blockquote></details>

I see, yes adding additional return value as an optional argument sounds good to me.
___
*[`solvers/semidefinite_relaxation.cc` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBXQA07V7rKrKQ5B0:-NYP_mr0ENN1B104Vc_K:b-4o79zm) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L60)):*

Even in your specific example, we should probably be multiplying those constraints together to get the tightest quadratic approximation.  I'm going to stop short of working out the details, because...

I feel very strongly that we have underestimated the strength of the (first-level) SDP relaxation in the past by cutting corners like this.  I know you're right, and I agree that the SDPs can get big fast, but I want to establish the foundation to be the correct complete relaxation; we can add options to cut corners later -- and compare against this implementation to understand more precisely what we might be giving up.
</blockquote></details>

Sounds good!

BTW, we can multiply the linear equality constraint (`A*x-b = 0`) with any linear expression `C*x-d`, and requires the linear equality constraint on the bilinear product. Here we chose to multiply `A*x-b=0` with itself. I wonder if there is any guidelines to choose the other linear expression `C*x-d`.
___
*[`solvers/semidefinite_relaxation.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPbGZSBW7NvGTxPvTO:-NYPbGZSBW7NvGTxPvTP:bv792fn) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L68)):*
> ```cpp
>   // X(-1,-1) = 1.
>   Variable one = relaxation->NewContinuousVariables<1>("one")[0];
>   X(prog.num_vars(), prog.num_vars()) = one;
> ```

BTW, if we print `X`, then `X(-1, -1)` will be printed as `one(0)`. I think you maybe want to created it directly as
```cc
X(prog.num_vars(), prog.num_vars()) = symbolic::Variable("one");
relaxation->AddDecisionVariables(Vector1<symbolic::Variable>(X(prog.num_vars(), prog.num_vars())));
```
?
___
*[`solvers/semidefinite_relaxation.cc` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPhuBu4W5dbwMRu0N0:-NYPhuBu4W5dbwMRu0N1:buvdbaa) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L79)):*
> ```cpp
>   // y in prog.
>   auto linear_vars = [&x, &prog](const VectorXDecisionVariable& prog_vars) {
>     return GetVariablesByIndex(x, prog.FindDecisionVariableIndices(prog_vars));
> ```

Why not to return prog_vars directly? They should be the same variables as we just add all the variables in `prog` to `relaxation`
___
*[`solvers/semidefinite_relaxation.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPeHMj4ZMgJKMEH1JG:-NYPdYO38nuRo9ql3DZ8:bwhxolu) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L99)):*
> ```cpp
>   // Returns the {a, vars} in relaxation, such that a' vars = 0.5*tr(QY). This
>   // assumes Q=Q', which is ensured by QuadraticCost and QuadraticConstraint.
>   auto half_trace_QY = [&X, &prog](const Eigen::MatrixXd& Q,
> ```

BTW, it took me a while to convince myself that the code would work correctly even if one quadratic constraint contains duplicated variables. (which we do allow when we add constraints through `prog.AddQuadraticConstraint(Q, b, lb, ub, vars)`. Could you add a test case with duplicated variables in the constraint in the original program? Namely if we have the original program s
```cc
prog.AddQuaraticConstraint(Eigen::Matrix2d::Ones(), Eigen::Vector2d::Zero(), 0, 1, Vector2<symbolic::Variable>(x(0), x(0)));
auto relaxed_prog = MakeSemidefiniteRelaxation(prog);
```
this `relaxed_prog` imposes the right constraint `0 <= 2*X(0, 0) <= 1`?
___
*[`solvers/semidefinite_relaxation.cc` line 143 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYQ3xru8FZPmYPjfKm5:-NYQ3xru8FZPmYPjfKm6:b-rbqvgi) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L143)):*
> ```cpp
>   // 0 ≤ (Ay-b)(Ay-b)ᵀ, for both b=lb and b=ub.
>   // The second constraint is implemented as -bbᵀ ≤ AYAᵀ - b(Ay)ᵀ - Aybᵀ.
>   for (const auto& binding : prog.linear_constraints()) {
> ```

I wonder if we should first aggregate all linear inequality constraint together as `A*x-b` with a big `A` matrix including all the linear inequality constraints, and then add the strengthening constraint `(A*x-b) * (A*x-b)' >= 0`. Currently we add this strengthening separately for each `Binding<LinearConstraint>` object.
___
*[`solvers/semidefinite_relaxation.cc` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPt1SL6CjLX_JB4Sem:-NYPt1SL6CjLX_JB4Sen:b98jmrt) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L156)):*
> ```cpp
>         binding.evaluator()->GetDenseA() * Y *
>         binding.evaluator()->GetDenseA().transpose();
>     if (binding.evaluator()->lower_bound().array().isFinite().any()) {
> ```

Instead of checking if any of the lower bound is finite, I think we should loop through each row of the constraint, and pick out the rows that has finite lower bound? It is possible that people put (-2, -inf) as the lower bound, and then we would have infinity in the second row of the constraint.

hongkai-dai (commented):
A quick thought: we have a flag `is_convex()` for the quadratic cost and constraints. Should we keep the convex cost/constraints, and only relax the non-convex ones?

This can be left as a TODO as currently in MosekSolver and ScsSolver I haven't supported convex quadratic constraints yet. But I wonder if that will give us a tighter relaxation.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NYTshkP5gHUpDa0U4Qe:b-wtvq1m)* status: 7 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

hongkai-dai (commented):
I take it back. If the quadratic constraint is convex, then we consider the SDP relaxation, it should be the same as the original quardratic constraints.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NYnx-Lo1OxGat1s4gJx:bi5s7wm)* status: 7 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

RussTedrake (commented):
This is ready for review now.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZEXHLQ-9dengjSFk7U:b-lfgkic)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMArAg5jTDHmS5iS_S:-NZEPcCH9yYua_83rFKd:brn65xo) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.h#L21)):*

I see, yes adding additional return value as an optional argument sounds good to me.
</blockquote></details>

Done.  Added a TODO
___
*[`solvers/semidefinite_relaxation.cc` line 60 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYMBXQA07V7rKrKQ5B0:-NZEPjPmDlZycZUF2TCn:bf8o4ei) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L60)):*

Sounds good!

BTW, we can multiply the linear equality constraint (`A*x-b = 0`) with any linear expression `C*x-d`, and requires the linear equality constraint on the bilinear product. Here we chose to multiply `A*x-b=0` with itself. I wonder if there is any guidelines to choose the other linear expression `C*x-d`.
</blockquote></details>

Done.  As discussed f2f, Bernhard reminded me of the more general form (Ax+b)x' = 0.
___
*[`solvers/semidefinite_relaxation.cc` line 68 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPbGZSBW7NvGTxPvTO:-NZEQPO-1-GbWJNr9CNN:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L68)):*

BTW, if we print `X`, then `X(-1, -1)` will be printed as `one(0)`. I think you maybe want to created it directly as
```cc
X(prog.num_vars(), prog.num_vars()) = symbolic::Variable("one");
relaxation->AddDecisionVariables(Vector1<symbolic::Variable>(X(prog.num_vars(), prog.num_vars())));
```
?
</blockquote></details>

Done.
___
*[`solvers/semidefinite_relaxation.cc` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPhuBu4W5dbwMRu0N0:-NZEQjOf8HReFvyrOYyn:b-qf4gb8) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L79)):*

Why not to return prog_vars directly? They should be the same variables as we just add all the variables in `prog` to `relaxation`
</blockquote></details>

prog_vars is variables from a binding associated with prog; it may not be all of the variables.
___
*[`solvers/semidefinite_relaxation.cc` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPeHMj4ZMgJKMEH1JG:-NZEWagQ3sZgUEr6ECz1:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L99)):*

BTW, it took me a while to convince myself that the code would work correctly even if one quadratic constraint contains duplicated variables. (which we do allow when we add constraints through `prog.AddQuadraticConstraint(Q, b, lb, ub, vars)`. Could you add a test case with duplicated variables in the constraint in the original program? Namely if we have the original program s
```cc
prog.AddQuaraticConstraint(Eigen::Matrix2d::Ones(), Eigen::Vector2d::Zero(), 0, 1, Vector2<symbolic::Variable>(x(0), x(0)));
auto relaxed_prog = MakeSemidefiniteRelaxation(prog);
```
this `relaxed_prog` imposes the right constraint `0 <= 2*X(0, 0) <= 1`?
</blockquote></details>

Done.
___
*[`solvers/semidefinite_relaxation.cc` line 143 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYQ3xru8FZPmYPjfKm5:-NZER271-xpwjU4AntEN:bhot5yt) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L143)):*

I wonder if we should first aggregate all linear inequality constraint together as `A*x-b` with a big `A` matrix including all the linear inequality constraints, and then add the strengthening constraint `(A*x-b) * (A*x-b)' >= 0`. Currently we add this strengthening separately for each `Binding<LinearConstraint>` object.
</blockquote></details>

Done.  I loop through all of the pairs.
___
*[`solvers/semidefinite_relaxation.cc` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPt1SL6CjLX_JB4Sem:-NZER69v9sW3jv_irtuN:bnes380) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L156)):*

Instead of checking if any of the lower bound is finite, I think we should loop through each row of the constraint, and pick out the rows that has finite lower bound? It is possible that people put (-2, -inf) as the lower bound, and then we would have infinity in the second row of the constraint.
</blockquote></details>

This will be slightly inefficient, perhaps, but would make the code much messier.  I'd like to land this simpler version first...?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.cc` line 143 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYQ3xru8FZPmYPjfKm5:-NZGFKGB2ELM3s7SWV7g:bt0imoc) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L143)):*

Done.  I loop through all of the pairs.
</blockquote></details>

Done again.  I decided your way was cleaner.  ;-)
___
*[`solvers/semidefinite_relaxation.cc` line 156 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPt1SL6CjLX_JB4Sem:-NZGFNWk2AuV1sv95zZ2:b-s6zuk7) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L156)):*

This will be slightly inefficient, perhaps, but would make the code much messier.  I'd like to land this simpler version first...?
</blockquote></details>

Done.  I decided it was easy to handle in your version (making the big A).

hongkai-dai (commented):
:lgtm: 

> Reviewed 5 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZHa7mzCpuyQLz4x8L2:bg7igkv)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.cc` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPhuBu4W5dbwMRu0N0:-NZHWuZa7hB5TVyA5ON1:bkbpx43) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L79)):*

prog_vars is variables from a binding associated with prog; it may not be all of the variables.
</blockquote></details>

I am still not sure if I get it. `prog_vars` is the variables from a binding associated with `prog`. This function returns the variables in `x` that are `prog_vars`. Should that be `prog_vars`? From my reading on the code, we have the invariant `x(i) == prog.decision_variable(i)` (except the last entry of `x`). Hence `GetVariablesByIndex(x, prog.FindDecisionVariableIndices(prog_vars))` should be `prog_vars`.
___
*[`solvers/semidefinite_relaxation.cc` line 164 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZHYkEjA6zSf4HxIxq5:-NZHYkEjA6zSf4HxIxq6:b-av3cn0) ([raw file](https://github.com/RobotLocomotion/drake/blob/2cae39ccfe9f428872604a3b5b00ec7b165f839f/solvers/semidefinite_relaxation.cc#L164)):*
> ```cpp
>               triplet_list.push_back(Triplet<double>(
>                   constraint, indices[j],
>                   -binding.evaluator()->get_sparse_A().coeff(i, j)));
> ```

BTW, using `coeff` function for an Eigen sparse matrix can be slow, as mentioned in https://eigen.tuxfamily.org/dox/group__TutorialSparse.html#:~:text=Random%20access%20to,non%2Dzeros%20elements..

Would you mind adding a TODO here? We can change to iterating over the Eigen::SparseMatrix::InnerIterator later.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`solvers/semidefinite_relaxation.cc` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NYPhuBu4W5dbwMRu0N0:-NZLlAym8mgRdWG06Pjs:b89vetl) ([raw file](https://github.com/RobotLocomotion/drake/blob/a1e17d6550454c8d36140147fa7117dc82628b16/solvers/semidefinite_relaxation.cc#L79)):*

I am still not sure if I get it. `prog_vars` is the variables from a binding associated with `prog`. This function returns the variables in `x` that are `prog_vars`. Should that be `prog_vars`? From my reading on the code, we have the invariant `x(i) == prog.decision_variable(i)` (except the last entry of `x`). Hence `GetVariablesByIndex(x, prog.FindDecisionVariableIndices(prog_vars))` should be `prog_vars`.
</blockquote></details>

Done.  Now I understand your point. Thanks!
___
*[`solvers/semidefinite_relaxation.cc` line 164 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZHYkEjA6zSf4HxIxq5:-NZLlMGJ0ZsCzowTLxeb:b-7jctgu) ([raw file](https://github.com/RobotLocomotion/drake/blob/2cae39ccfe9f428872604a3b5b00ec7b165f839f/solvers/semidefinite_relaxation.cc#L164)):*

BTW, using `coeff` function for an Eigen sparse matrix can be slow, as mentioned in https://eigen.tuxfamily.org/dox/group__TutorialSparse.html#:~:text=Random%20access%20to,non%2Dzeros%20elements..

Would you mind adding a TODO here? We can change to iterating over the Eigen::SparseMatrix::InnerIterator later.
</blockquote></details>

Done. The reason I didn't is because I'm interating rowwise, and the sparse matrix iterators are columnwise; i thought it was more clear to keep this order.

RussTedrake (commented):
+@sherm1 for platform review, please (per rotation)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZLltAq8yuAA1H2MZ-5:bo51vas)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform)

hongkai-dai (commented):
> Reviewed 3 of 11 files at r1, 2 of 8 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Platform checkpoint (lunch)

> Reviewed 3 of 8 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZbRu-OD6Es3QX19Pi9:bypd00n)* status: 2 unresolved discussions, LGTM missing from assignee sherm1(platform)

___
*[`solvers/semidefinite_relaxation.cc` line 42 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbOXd06THO8d5fRiqU:-NZbOXd06THO8d5fRiqV:byqgclj) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/semidefinite_relaxation.cc#L42)):*
> ```cpp
>                                       &unsupported_message)) {
>         "MakeSemidefiniteRelaxation does not (yet) support this program: {}.",
> ```

BTW consider adding () `MakeSemidefiniteRelaxation()` to distinguish function name from class name
___
*[`solvers/semidefinite_relaxation.cc` line 119 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbQa_66rIIYl4ntac1:-NZbQa_66rIIYl4ntac2:b-t28puu) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/semidefinite_relaxation.cc#L119)):*
> ```cpp
>     // Now assemble one big Ay <= b matrix.
>     int num_constraints = 0;
>     int num_nnz = 0;
> ```

BTW nnz=num non zeros. Consider calling this just `nnz` or `num_non_zeroes`

sherm1 (commented):
This is amazing! Platform :lgtm_strong:

Are the changes to spectrahedron.cc (dense->sparse) related to the other changes here? I can't tell. If not, please remove or just mention them in the PR description & commit message.

> Reviewed 3 of 11 files at r1, 4 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZc0EiUDjJsW-e2Bwqx:b-w53uvn)* status: 3 unresolved discussions

___
*[`solvers/test/semidefinite_relaxation_test.cc` line 151 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbxIrzD51UfxeZuEa-:-NZbxIrzD51UfxeZuEa0:bi86mdo) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/test/semidefinite_relaxation_test.cc#L151)):*
> ```cpp
> 
>   // Third linear constraint is 0 ≤ (Ay-b)(Ay-b)ᵀ, where A and b represent all
>   // of the constraints stacked.
> ```

minor: I found this comment confusing because the "linear constraint" here isn't linear, while in the above two cases the original constraint is linear. Hongkai explained to me how you relax it. Please clarify.

RussTedrake (commented):
Done.  (mentioned it in the commit message)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-:-NZcEvzG8cMTJLlZqrq_:b-4jbgyq)* status: :shipit: complete! all discussions resolved, LGTM from assignees sherm1(platform),hongkai-dai

___
*[`solvers/semidefinite_relaxation.cc` line 42 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbOXd06THO8d5fRiqU:-NZcDuH4EpvRrst6n-n_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/semidefinite_relaxation.cc#L42)):*

BTW consider adding () `MakeSemidefiniteRelaxation()` to distinguish function name from class name
</blockquote></details>

Done.
___
*[`solvers/semidefinite_relaxation.cc` line 119 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbQa_66rIIYl4ntac1:-NZcE6Zl4NhEZ4StzODT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/semidefinite_relaxation.cc#L119)):*

BTW nnz=num non zeros. Consider calling this just `nnz` or `num_non_zeroes`
</blockquote></details>

Done.
___
*[`solvers/test/semidefinite_relaxation_test.cc` line 151 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19632#-NZbxIrzD51UfxeZuEa-:-NZcEOf23s3-5MR_a_cO:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/50209b2900c926c97430bb84114b5b0f7624c468/solvers/test/semidefinite_relaxation_test.cc#L151)):*

minor: I found this comment confusing because the "linear constraint" here isn't linear, while in the above two cases the original constraint is linear. Hongkai explained to me how you relax it. Please clarify.
</blockquote></details>

Done.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19632)* status: :shipit: complete! all discussions resolved, LGTM from assignees sherm1(platform),hongkai-dai

PR #19633, opened 2023-06-20
MathematicalProgram::AddConstraint parses QuadraticConstraint

RussTedrake:
Previously, a quadratic constraint passes to AddConstraint would default to an ExpressionConstraint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)

### DIFF
--- a/solvers/constraint.h
+++ b/solvers/constraint.h
@@ -543,8 +543,10 @@ class EvaluatorConstraint : public Constraint {
 /**
  * A constraint on the values of multivariate polynomials.
  *
- *  lb[i] <= P[i](x, y...) <= ub[i], where each P[i] is a multivariate
- *  polynomial in x, y...
+ * @verbatim
+ * lb[i] ≤ P[i](x, y...) ≤ ub[i],
+ * @endverbatim
+ * where each P[i] is a multivariate polynomial in x, y...
  *
  * The Polynomial class uses a different variable naming scheme; thus the
  * caller must provide a list of Polynomial::VarType variables that correspond
--- a/solvers/create_constraint.cc
+++ b/solvers/create_constraint.cc
@@ -38,6 +38,14 @@ Binding<Constraint> ParseConstraint(
   DRAKE_ASSERT(v.rows() == lb.rows() && v.rows() == ub.rows());
 
   if (!IsAffine(v)) {
+    // Quadratic constraints.
+    if (v.size() == 1 && v[0].is_polynomial()) {
+      const symbolic::Polynomial poly{v[0]};
+      if (poly.TotalDegree() == 2) {
+        return ParseQuadraticConstraint(v[0], lb[0], ub[0]);
+      }
+    }
+
     auto constraint = make_shared<ExpressionConstraint>(v, lb, ub);
     return CreateBinding(constraint, constraint->vars());
   }  // else, continue on to linear-specific version below.
--- a/solvers/mathematical_program.h
+++ b/solvers/mathematical_program.h
@@ -1305,9 +1305,9 @@ class MathematicalProgram {
    * @param ub A scalar, the upper bound.
    *
    * The resulting constraint may be a BoundingBoxConstraint, LinearConstraint,
-   * LinearEqualityConstraint, or ExpressionConstraint, depending on the
-   * arguments.  Constraints of the form x == 1 (which could be created as a
-   * BoundingBoxConstraint or LinearEqualityConstraint) will be
+   * LinearEqualityConstraint, QuadraticConstraint, or ExpressionConstraint,
+   * depending on the arguments.  Constraints of the form x == 1 (which could
+   * be created as a BoundingBoxConstraint or LinearEqualityConstraint) will be
    * constructed as a LinearEqualityConstraint.
    */
   Binding<Constraint> AddConstraint(const symbolic::Expression& e, double lb,
--- a/solvers/test/create_constraint_test.cc
+++ b/solvers/test/create_constraint_test.cc
@@ -15,6 +15,10 @@ namespace drake {
 namespace solvers {
 namespace {
 
+using Eigen::Vector2d;
+
+const double kInf = std::numeric_limits<double>::infinity();
+
 void CheckParseQuadraticConstraint(
     const Expression& e, double lb, double ub,
     std::optional<QuadraticConstraint::HessianType> hessian_type,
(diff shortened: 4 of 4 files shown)


hongkai-dai (commented):
:lgtm: 

+@jwnimmer-tri for platform review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-:-NYLmfVf5wSJLHaG3lui:b-puciyp)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`solvers/create_constraint.cc` line 40 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-NYLl439DuuUumx40j6E:-NYLl439DuuUumx40j6F:bvmypm) ([raw file](https://github.com/RobotLocomotion/drake/blob/15513f85b97b12533b18a800be47dfcb455cac02/solvers/create_constraint.cc#L40)):*
> ```cpp
>   DRAKE_ASSERT(v.rows() == lb.rows() && v.rows() == ub.rows());
> 
>   // Quadratic constraints.
> ```

Constructing a symbolic::Polynomial from a symbolic::Expression is usually the time bottle neck. Here if we have a linear expression, then we construct the symbolic::Polynomial first, and then check the polynomial degree. We discarded the newly constructed symbolic Polynomial here, and later re-construct the symbolic::Polynomial again when we check if the expression is affine or parsing the affine expression.

Could you put a TODO here?
```
TODO(hongkai.dai): If the expression is polynomial, then construct the symbolic::Polynomial and use the polynomial instead of the original expression afterwards.
```

I will work on this in a separate PR with benchmarks.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`solvers/create_constraint.cc` line 40 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-NYLl439DuuUumx40j6E:-NYNwUQ652bBhL0jkR7p:b-ocwwu3) ([raw file](https://github.com/RobotLocomotion/drake/blob/15513f85b97b12533b18a800be47dfcb455cac02/solvers/create_constraint.cc#L40)):*

Constructing a symbolic::Polynomial from a symbolic::Expression is usually the time bottle neck. Here if we have a linear expression, then we construct the symbolic::Polynomial first, and then check the polynomial degree. We discarded the newly constructed symbolic Polynomial here, and later re-construct the symbolic::Polynomial again when we check if the expression is affine or parsing the affine expression.

Could you put a TODO here?
```
TODO(hongkai.dai): If the expression is polynomial, then construct the symbolic::Polynomial and use the polynomial instead of the original expression afterwards.
```

I will work on this in a separate PR with benchmarks.
</blockquote></details>

I agree that we should prioritize the performance of the linear constraints -- they are the most common by far.  I've moved the quadratic constraints to live behind the affine check.

Do you still want the TODO, given this version?

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-:-NYOEEu2Bod4jX3rf2QN:b-vtj4vo)* status: 2 unresolved discussions

___
*[`solvers/test/create_constraint_test.cc` line 97 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-NYODrej1gV-CsCl8x5o:-NYODrej1gV-CsCl8x5p:b-fy3jji) ([raw file](https://github.com/RobotLocomotion/drake/blob/d24dacb144de4cdaea1cd55ebaf9ad038472c6ca/solvers/test/create_constraint_test.cc#L97)):*
> ```cpp
>   EXPECT_NE(dynamic_cast<QuadraticConstraint*>(binding.evaluator().get()),
>             nullptr);
> }
> ```

BTW Given the amount of if-checks in the implementation, it seems a little suspicious that there is only one new test case.

Are all of the the guards (size == 1; totaldegree == 2) covered with both true and false values?  In other words, if I delete any of those conditions from the cc file, does a unit test fail?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: 1 unresolved discussion

___
*[`solvers/create_constraint.cc` line 40 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-NYLl439DuuUumx40j6E:-NYOvzm982WlpND8Yiss:bhugnt6) ([raw file](https://github.com/RobotLocomotion/drake/blob/15513f85b97b12533b18a800be47dfcb455cac02/solvers/create_constraint.cc#L40)):*

I agree that we should prioritize the performance of the linear constraints -- they are the most common by far.  I've moved the quadratic constraints to live behind the affine check.

Do you still want the TODO, given this version?
</blockquote></details>

Thanks! I think this is better.

Still I think we have some redundancy in the implementation. `IsAffine` function will also try to parse the symbolic expression to a symbolic Polynomial. I don't think we need to add the TODO for now. I will create a benchmark first.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

___
*[`solvers/test/create_constraint_test.cc` line 97 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19633#-NYODrej1gV-CsCl8x5o:-NYPTV_s9rxUDCRutP9m:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/d24dacb144de4cdaea1cd55ebaf9ad038472c6ca/solvers/test/create_constraint_test.cc#L97)):*

BTW Given the amount of if-checks in the implementation, it seems a little suspicious that there is only one new test case.

Are all of the the guards (size == 1; totaldegree == 2) covered with both true and false values?  In other words, if I delete any of those conditions from the cc file, does a unit test fail?
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

hongkai-dai (commented):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19633)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

Issue #17634, opened 2022-07-26
discrete_lyapunov_equation.cc bug in solution of Equation 4.2


grantgib:
https://github.com/RobotLocomotion/drake/blob/4b98f4375f978a02182093b6dfabdcd3d499c236/math/discrete_lyapunov_equation.cc#L180

There is a possible bug in one of the steps for computing an intermediate term xbar. The factorization of the referenced equation 4.2 is missing an identity subtraction in the diagonal elements of MatrixXd lhs (line 199) 

It should in fact be 
MatrixXd lhs(2 * (m - 2), 2 * (m - 2));
    lhs << s_11(0, 0) * S_1.transpose() - **Identity**, s_11(1, 0) * S_1.transpose(),
        s_11(0, 1) * S_1.transpose(), s_11(1, 1) * S_1.transpose() - **Identity**;

jwnimmer-tri:
\CC FYI @RussTedrake @sherm1 @hongkai-dai @FischerGundlach @weiqiao (as participants in #10885).

grantgib:
[dlyap_solution_derivation.pdf](https://github.com/RobotLocomotion/drake/files/9212984/dlyap_solution_derivation.pdf)
I've attached a correct derivation. The A,Q matrices used in the [dlyap unit tests](https://github.com/RobotLocomotion/drake/blob/master/math/test/discrete_lyapunov_equation_test.cc) might be too sparse to illuminate the error. Comparing the corrected solution of Eq 4.2 with the original code returns the same result for all unit tests. Perhaps with more dense State and Cost matrices the error would become more clear.

Thanks,

hongkai-dai:
@grantgib Thanks a lot for finding this bug and deriving the math! May I ask if you have a test example that would fail with the current code on master (without subtracting the identity)? What triggered you to find this math bug?

grantgib:
@hongkai-dai I tried to generate some random examples that would show failure but so far both methods return the same result (6 significant digits).

As far as finding the bug, I was implementing my own solution for dlyap equations and was using this as reference.

ggould-tri:
Tagging as mathematical_program / hongkai for now; feel free to re-tag and re-assign as appopriate.

RussTedrake:
Amazing.  @grantgib -- thanks for catching and reporting it!  @hongkai -- thanks for fixing it.

PR #19634, opened 2023-06-20
[planning] ConvexSet::AddPointInSetConstraints returns the new variables

### LINKED ISSUE #19585: Document Constraints that May Add Decision Variables, and Return Those Decision Variables
When constructing mathematical programs, the `AddPointInSetConstraints` method of `ConvexSet` may add additional decision variables to the program. A concrete example of this is for the `CartesianProduct` class, as can be seen [here](https://github.com/RobotLocomotion/drake/blob/37d6c59a13431956d82123de6cda24cfa765c50d/geometry/optimization/cartesian_product.cc#L173). This can cause problems down the line if the user is relying on the structure of the variables they've constructed. In my case, this became an issue when using the `initial_guess` parameter for `Solve`. It would be helpful to document that such functions can add new decision variables, and also to return those variables to the user, so they can use the `SetInitialGuess` methods.

hongkai-dai:
Resolves #19585

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)

### DIFF
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -62,7 +62,9 @@ class TestGeometryOptimization(unittest.TestCase):
         np.testing.assert_array_equal(hpoly.b(), self.b)
         self.assertTrue(hpoly.PointInSet(x=[0, 0, 0], tol=0.0))
         self.assertFalse(hpoly.IsBounded())
-        hpoly.AddPointInSetConstraints(self.prog, self.x)
+        new_vars, new_constraints = hpoly.AddPointInSetConstraints(
+            self.prog, self.x)
+        self.assertEqual(new_vars.size, 0)
         constraints = hpoly.AddPointInNonnegativeScalingConstraints(
             prog=self.prog, x=self.x, t=self.t)
         self.assertGreaterEqual(len(constraints), 2)
@@ -167,7 +169,10 @@ class TestGeometryOptimization(unittest.TestCase):
         np.testing.assert_array_equal(ellipsoid.A(), self.A)
         np.testing.assert_array_equal(ellipsoid.center(), self.b)
         self.assertTrue(ellipsoid.PointInSet(x=self.b, tol=0.0))
-        ellipsoid.AddPointInSetConstraints(self.prog, self.x)
+        new_vars, new_constraints = ellipsoid.AddPointInSetConstraints(
+            self.prog, self.x)
+        self.assertEqual(new_vars.size, 0)
+        self.assertGreater(len(new_constraints), 0)
         constraints = ellipsoid.AddPointInNonnegativeScalingConstraints(
             prog=self.prog, x=self.x, t=self.t)
         self.assertGreaterEqual(len(constraints), 2)
@@ -227,7 +232,9 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertEqual(vpoly.ambient_dimension(), 2)
         np.testing.assert_array_equal(vpoly.vertices(), vertices)
         self.assertTrue(vpoly.PointInSet(x=[1.0, 5.0], tol=1e-8))
-        vpoly.AddPointInSetConstraints(self.prog, self.x[0:2])
+        new_vars, new_constraints = vpoly.AddPointInSetConstraints(
+            self.prog, self.x[0:2])
+        self.assertEqual(new_vars.size, vertices.shape[1])
         constraints = vpoly.AddPointInNonnegativeScalingConstraints(
             prog=self.prog, x=self.x[:2], t=self.t)
         self.assertGreaterEqual(len(constraints), 2)
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -222,6 +222,7 @@ drake_cc_library(
         "//geometry:scene_graph",
         "//solvers:mathematical_program",
         "//solvers:solve",
+        "@gtest//:without_main",
     ],
 )
 
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -164,25 +164,42 @@ bool CartesianProduct::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
   return true;
 }
 
-void CartesianProduct::DoAddPointInSetConstraints(
+std::pair<VectorX<symbolic::Variable>, std::vector<Binding<Constraint>>>
+CartesianProduct::DoAddPointInSetConstraints(
     MathematicalProgram* prog,
     const Eigen::Ref<const VectorXDecisionVariable>& x) const {
+  // Use std::vector which allocates heap memory logarithmically instead of
+  // linearly.
+  std::vector<Variable> new_vars;
   VectorXDecisionVariable y;
+  std::vector<Binding<Constraint>> new_constraints;
   if (A_) {
     // Note: The constructor enforces that A_ is full column rank.
     y = prog->NewContinuousVariables(A_->rows(), "y");
+    new_vars = std::vector<Variable>(y.data(), y.data() + y.rows() * y.cols());
     // y = Ax + b, or [I,-A]*[y;x] = b.
     MatrixXd Aeq = MatrixXd::Identity(A_->rows(), A_->rows() + A_->cols());
     Aeq.rightCols(A_->cols()) = -(*A_);
-    prog->AddLinearEqualityConstraint(Aeq, (*b_), {y, x});
+    new_constraints.push_back(
+        prog->AddLinearEqualityConstraint(Aeq, (*b_), {y, x}));
   } else {
     y = x;
   }
   int index = 0;
   for (const auto& s : sets_) {
-    s->AddPointInSetConstraints(prog, y.segment(index, s->ambient_dimension()));
+    const auto [new_var_in_s, new_constraints_in_s] =
+        s->AddPointInSetConstraints(prog,
+                                    y.segment(index, s->ambient_dimension()));
+    for (int i = 0; i < new_var_in_s.rows(); ++i) {
+      new_vars.push_back(new_var_in_s(i));
+    }
+    new_constraints.insert(new_constraints.end(), new_constraints_in_s.begin(),
+                           new_constraints_in_s.end());
     index += s->ambient_dimension();
   }
+  VectorX<Variable> new_vars_vec =
+      Eigen::Map<VectorX<Variable>>(new_vars.data(), new_vars.size());
+  return {std::move(new_vars_vec), std::move(new_constraints)};
 }
 
 std::vector<Binding<Constraint>>
@@ -215,7 +232,7 @@ CartesianProduct::DoAddPointInNonnegativeScalingConstraints(
   return constraints;
 }
 
-std::vector<solvers::Binding<solvers::Constraint>>
+std::vector<Binding<Constraint>>
 CartesianProduct::DoAddPointInNonnegativeScalingConstraints(
     solvers::MathematicalProgram* prog, const Eigen::Ref<const MatrixXd>& A_x,
     const Eigen::Ref<const VectorXd>& b_x, const Eigen::Ref<const VectorXd>& c,
--- a/geometry/optimization/cartesian_product.h
+++ b/geometry/optimization/cartesian_product.h
@@ -82,7 +82,9 @@ class CartesianProduct final : public ConvexSet {
   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
                     double tol) const final;
 
-  void DoAddPointInSetConstraints(
+  std::pair<VectorX<symbolic::Variable>,
+            std::vector<solvers::Binding<solvers::Constraint>>>
+  DoAddPointInSetConstraints(
       solvers::MathematicalProgram*,
       const Eigen::Ref<const solvers::VectorXDecisionVariable>&) const final;
 
--- a/geometry/optimization/convex_set.cc
+++ b/geometry/optimization/convex_set.cc
@@ -35,7 +35,9 @@ std::optional<Eigen::VectorXd> ConvexSet::DoMaybeGetPoint() const {
   return std::nullopt;
 }
 
-void ConvexSet::AddPointInSetConstraints(
+std::pair<VectorX<symbolic::Variable>,
+          std::vector<solvers::Binding<solvers::Constraint>>>
+ConvexSet::AddPointInSetConstraints(
     solvers::MathematicalProgram* prog,
     const Eigen::Ref<const solvers::VectorXDecisionVariable>& vars) const {
   DRAKE_THROW_UNLESS(vars.size() == ambient_dimension());
--- a/geometry/optimization/convex_set.h
+++ b/geometry/optimization/convex_set.h
@@ -114,12 +114,18 @@ class ConvexSet : public ShapeReifier {
     return DoPointInSet(x, tol);
   }
 
-  // Note: I would like to return the Binding, but the type is subclass
-  // dependent.
   /** Adds a constraint to an existing MathematicalProgram enforcing that the
   point defined by vars is inside the set.
+  @return (new_vars, new_constraints) Some of the derived class will add new
+  decision variables to enforce this constraint, we return all the newly added
+  decision variables as new_vars. The meaning of these new decision variables
+  differs in each subclass. If no new variables are added, then we return an
+  empty Eigen vector. Also we return all the newly added constraints to `prog`
+  through this function.
   @throws std::exception if ambient_dimension() == 0 */
-  void AddPointInSetConstraints(
+  std::pair<VectorX<symbolic::Variable>,
+            std::vector<solvers::Binding<solvers::Constraint>>>
+  AddPointInSetConstraints(
       solvers::MathematicalProgram* prog,
       const Eigen::Ref<const solvers::VectorXDecisionVariable>& vars) const;
 
@@ -137,6 +143,7 @@ class ConvexSet : public ShapeReifier {
   and ⊕ is the Minkowski sum.  For t > 0, this is equivalent to x ∈ t S, but for
   t = 0, we have only x ∈ rec(S).
   @throws std::exception if ambient_dimension() == 0 */
+  // TODO(hongkai.dai): return the new variables also.
   std::vector<solvers::Binding<solvers::Constraint>>
   AddPointInNonnegativeScalingConstraints(
       solvers::MathematicalProgram* prog,
@@ -221,7 +228,9 @@ class ConvexSet : public ShapeReifier {
   /** Non-virtual interface implementation for AddPointInSetConstraints().
   @pre vars.size() == ambient_dimension()
   @pre ambient_dimension() > 0 */
-  virtual void DoAddPointInSetConstraints(
+  virtual std::pair<VectorX<symbolic::Variable>,
+                    std::vector<solvers::Binding<solvers::Constraint>>>
+  DoAddPointInSetConstraints(
       solvers::MathematicalProgram* prog,
       const Eigen::Ref<const solvers::VectorXDecisionVariable>& vars) const = 0;
 
--- a/geometry/optimization/hpolyhedron.cc
+++ b/geometry/optimization/hpolyhedron.cc
@@ -536,13 +536,17 @@ bool HPolyhedron::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
   return ((A_ * x).array() <= b_.array() + tol).all();
 }
 
-void HPolyhedron::DoAddPointInSetConstraints(
+std::pair<VectorX<Variable>, std::vector<Binding<Constraint>>>
+HPolyhedron::DoAddPointInSetConstraints(
     MathematicalProgram* prog,
     const Eigen::Ref<const VectorXDecisionVariable>& vars) const {
-  prog->AddLinearConstraint(
+  VectorX<Variable> new_vars;
+  std::vector<Binding<Constraint>> new_constraints;
+  new_constraints.push_back(prog->AddLinearConstraint(
       A_,
       VectorXd::Constant(b_.size(), -std::numeric_limits<double>::infinity()),
-      b_, vars);
+      b_, vars));
+  return {std::move(new_vars), std::move(new_constraints)};
 }
 
 std::vector<Binding<Constraint>>
--- a/geometry/optimization/hpolyhedron.h
+++ b/geometry/optimization/hpolyhedron.h
@@ -217,7 +217,9 @@ class HPolyhedron final : public ConvexSet {
   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
                     double tol) const final;
 
-  void DoAddPointInSetConstraints(
+  std::pair<VectorX<symbolic::Variable>,
+            std::vector<solvers::Binding<solvers::Constraint>>>
+  DoAddPointInSetConstraints(
       solvers::MathematicalProgram* prog,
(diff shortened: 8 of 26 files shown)


hongkai-dai (commented):
@cohnt could you take a look at this PR and make sure it is want you want? Thanks a lot!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYMDQToBIft2158EZXt:bvhxyby)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

cohnt (commented):
This is exactly what I was looking for, thanks! I just have one minor comment on documentation.

> Reviewed 21 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYODwTVAZbEJ9Y_bs2d:bibxcqx)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/convex_set.h` line 123 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYODUiJClHpRpwhWabO:-NYODUiJClHpRpwhWabP:btv6drc) ([raw file](https://github.com/RobotLocomotion/drake/blob/ecf8618df56de72ee50158464771cb9861359fde/geometry/optimization/convex_set.h#L123)):*
> ```objc
>   @return new_vars Some of the derived class will add new decision variables to
>   enforce this constraint, we return all the newly added decision variables. The
>   meaning of these new decision variables differs in each subclass.
> ```

Small change, but I think it would help to mention that derived classes which don't create new decision variables will return an empty list.

hongkai-dai (commented):
+@cohnt for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYP8vKT6dpiZIrv3KVQ:bhsqy6j)* status: 1 unresolved discussion, LGTM missing from assignee cohnt, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/convex_set.h` line 123 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYODUiJClHpRpwhWabO:-NYP8zAT7Jf7Vd5mm833:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/ecf8618df56de72ee50158464771cb9861359fde/geometry/optimization/convex_set.h#L123)):*

Small change, but I think it would help to mention that derived classes which don't create new decision variables will return an empty list.
</blockquote></details>

Done.

cohnt (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYPJUm95LPtpUA_XdVD:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

hongkai-dai (commented):
+@jwnimmer-tri for a platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYPVFphD2tlS4bT12ts:b-qhld66)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

RussTedrake (commented):
> Reviewed 6 of 21 files at r1, 1 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/convex_set.h` line 126 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYPVbqGAVpFuylDeEhL:-NYPVbqGAVpFuylDeEhM:b-i63tv8) ([raw file](https://github.com/RobotLocomotion/drake/blob/e676967edc36562cec01c1280aaac5a5f93ee9b7/geometry/optimization/convex_set.h#L126)):*
> ```objc
>   variables are added, then we return an empty Eigen vector.
>   @throws std::exception if ambient_dimension() == 0 */
>   VectorX<symbolic::Variable> AddPointInSetConstraints(
> ```

For consistency with other Add___Constraint methods, I would expect the return type (if there is one) from this method to be the list of Bindings to the newly added constraints.  More thoroughly, it could be the std::pair with the added variable + the list of added constraints.  E.g. as done in https://drake.mit.edu/doxygen_cxx/classdrake_1_1solvers_1_1_mathematical_program.html#a6060c77125050d8c1b479a83d983a569 

I worry that adding only the variable return values would make the API inconsistent and annoying to change later?

jwnimmer-tri:
FYI I'm going to pause my review until Russ' question is resolved.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes

___
*[`geometry/optimization/convex_set.h` line 126 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYPVbqGAVpFuylDeEhL:-NYR5iiLDMzV_aXs3tLS:b-hjs3mm) ([raw file](https://github.com/RobotLocomotion/drake/blob/e676967edc36562cec01c1280aaac5a5f93ee9b7/geometry/optimization/convex_set.h#L126)):*

For consistency with other Add___Constraint methods, I would expect the return type (if there is one) from this method to be the list of Bindings to the newly added constraints.  More thoroughly, it could be the std::pair with the added variable + the list of added constraints.  E.g. as done in https://drake.mit.edu/doxygen_cxx/classdrake_1_1solvers_1_1_mathematical_program.html#a6060c77125050d8c1b479a83d983a569 

I worry that adding only the variable return values would make the API inconsistent and annoying to change later?
</blockquote></details>

Done. I changed the return type to a std::pair of added variable and constraints.

RussTedrake (commented):
> Reviewed 3 of 23 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/convex_set.h` line 126 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYPVbqGAVpFuylDeEhL:-NYSMPXr3q9s9fXBAUa7:b-c6eh91) ([raw file](https://github.com/RobotLocomotion/drake/blob/e676967edc36562cec01c1280aaac5a5f93ee9b7/geometry/optimization/convex_set.h#L126)):*

Done. I changed the return type to a std::pair of added variable and constraints.
</blockquote></details>

That was a lot of work.  Thank you!  I'll leave it to the reviewers to check more carefully for correctness / test coverage.

cohnt (commented):
Not sure if I'm still the assigned reviewer, but I'll throw in a 
:lgtm:

> Reviewed 23 of 23 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYThPec9tcvUuzM7vNw:b-v4zmze)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri (commented):
+(release notes: feature) to indicate the new feature.

+(release notes: breaking change) due to the return type change.  If a user had written their own ConvexSet subclass, this would break them.

I've started the platform review, but need a break for meetings.  I've posted one small comment in the meantime.

> Reviewed 1 of 21 files at r1, 12 of 23 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYTm4LU9TMnmBpqPYDq:btlujr7)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 65 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYTlNfyEHWHkWmTyCVB:-NYTlNfz8kio40d5LLL_:b-x0cya) ([raw file](https://github.com/RobotLocomotion/drake/blob/43354ca500251910a841e5babd0448d755ccfd22/bindings/pydrake/geometry/test/optimization_test.py#L65)):*
> ```Python
>         self.assertTrue(hpoly.PointInSet(x=[0, 0, 0], tol=0.0))
>         self.assertFalse(hpoly.IsBounded())
>         new_vars_and_constraints = hpoly.AddPointInSetConstraints(
> ```

nit This is non-idiomatic Python.  We should use structured names for the return values, not arcane `[0]` and `[1]` magic.



_Suggestion:_
```Python
new_vars, new_constraints = hpoly.AddPointInSetConstraints(
```

jwnimmer-tri (commented):
See https://github.com/hongkai-dai/drake/pull/31 for some cleanups.  Feel free to squash it in here, no need to stay separate other than for you to look it over.

> Reviewed 1 of 23 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYUAu1t6ZTRp8PMRQMv:b-603nin)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUBNxY07eF_D1cQ65W):*
nit Fit to length, remove stray capitaliztion, and update for recent changes:



_Suggestion:_
```
[planning] ConvexSet::AddPointInSetConstraints returns the new constraints
```
___
*[`geometry/optimization/convex_set.h` line 146 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYU10mg27c2JhJby0Uv:-NYU10mg27c2JhJby0Uw:b-hb2odi) ([raw file](https://github.com/RobotLocomotion/drake/blob/87d2e7048e87bc5f26c1e45a1bdf83d525e4de5e/geometry/optimization/convex_set.h#L146)):*
> ```objc
>   t = 0, we have only x ∈ rec(S).
>   @throws std::exception if ambient_dimension() == 0 */
>   std::vector<solvers::Binding<solvers::Constraint>>
> ```

BTW I assume we switch to `std::pair<VectorX<symbolic::Variable>,` to this function as well as the one below (to return the newly-added variables)?  I don't see a reason why these functions should be different that the one we're changing here.

Do you want to add a TODO (or issue) to that effect somewhere?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUBNxY07eF_D1cQ65W):*

nit Fit to length, remove stray capitaliztion, and update for recent changes:
</blockquote></details>

Done.

By Fit to length, you mean the first line is within 72 characters?
___
*[`geometry/optimization/convex_set.h` line 146 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYU10mg27c2JhJby0Uv:-NYUH5naCQIetaIBUI-6:b-wficnx) ([raw file](https://github.com/RobotLocomotion/drake/blob/87d2e7048e87bc5f26c1e45a1bdf83d525e4de5e/geometry/optimization/convex_set.h#L146)):*

BTW I assume we switch to `std::pair<VectorX<symbolic::Variable>,` to this function as well as the one below (to return the newly-added variables)?  I don't see a reason why these functions should be different that the one we're changing here.

Do you want to add a TODO (or issue) to that effect somewhere?
</blockquote></details>

I added a TODO here. It will be a breaking change.

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 3 of 23 files at r4, 8 of 8 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-:-NYUJ43k5nezgN1N-DMx:b-vtj4vo)* status: 5 unresolved discussions

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUBNxY07eF_D1cQ65W):*

Done.

By Fit to length, you mean the first line is within 72 characters?
</blockquote></details>

Yes, but after the ` (#17634)` gets tacked on automatically.
___
*[`geometry/optimization/convex_set.h` line 146 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYU10mg27c2JhJby0Uv:-NYUIvJzBK8r-k0SJqHI:b-8gqk6a) ([raw file](https://github.com/RobotLocomotion/drake/blob/87d2e7048e87bc5f26c1e45a1bdf83d525e4de5e/geometry/optimization/convex_set.h#L146)):*

I added a TODO here. It will be a breaking change.
</blockquote></details>

Yup.  This PR is already breaking though, so two breaks in one release cycle isn't too terrible, I hope.
___
*[`geometry/optimization/intersection.cc` line 113 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUI0aK1zrIT9JS_wCf:-NYUI0aK1zrIT9JS_wCg:b7vzre) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/intersection.cc#L113)):*
> ```cpp
>   VectorX<Variable> new_vars_vec =
>       Eigen::Map<VectorX<Variable>>(new_vars.data(), new_vars.size());
>   return {std::move(new_vars_vec), std::move(new_constraints)};
> ```

The return value logic here is non-trival.  I think we need a unit test to check it specifically?
___
*[`geometry/optimization/spectrahedron.cc` line 97 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUHjwK-QlqHVx1OiSc:-NYUHjwK-QlqHVx1OiSd:b2bufzi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/spectrahedron.cc#L97)):*
> ```cpp
>   std::vector<Binding<Constraint>> new_constraints;
>   for (const auto& binding : sdp_->GetAllConstraints()) {
>     new_constraints.push_back(prog->AddConstraint(
> ```

I think its important to have a spectrahedron unit test that counts the >1 number of `new_constraints` that get returned.
___
*[`geometry/optimization/test_utilities.cc` line 39 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUHHJR4QSY1a45sKTT:-NYUHHJR4QSY1a45sKTU:bdtoocz) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/test_utilities.cc#L39)):*
> ```cpp
>   const auto [new_vars, new_constraints] =
>       set.AddPointInSetConstraints(&prog, x);
>   EXPECT_EQ(prog.num_vars(), x.rows() + new_vars.rows());
> ```

nit A lot of the subclasses don't have unit tests of the `new_constraints` return value.  I think we could solve that with a simple change here:

```c++
EXPECT_FALSE(new_constraints.empty());
```
___
*[`geometry/optimization/test/minkowski_sum_test.cc` line 51 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIFKQ7CipZchY3uOt:-NYUIFKQ7CipZchY3uOu:b-ddtjse) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/test/minkowski_sum_test.cc#L51)):*
> ```cpp
>     auto [new_vars, new_constraints] = S.AddPointInSetConstraints(&prog, x);
>     EXPECT_EQ(new_vars.rows(), 4);
>     EXPECT_FALSE(new_constraints.empty());
> ```

nit I think its not too much of a maintenance burden to check an exact number here, and provides a real improvement to the level of implementation correctness we're checking for?



_Suggestion:_
```cpp
EXPECT_EQ(new_constraints.size(), ...);
```
___
*[`geometry/optimization/test/vpolytope_test.cc` line 132 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIWWRDXUgLlt1ppVz:-NYUIWWRDXUgLlt1ppW-:b-ddtjse) ([raw file](https://github.com/RobotLocomotion/drake/blob/25bd29508991724c353cf0ef988045367b6e6930/geometry/optimization/test/vpolytope_test.cc#L132)):*
> ```cpp
>     auto [new_vars, new_constraints] = V.AddPointInSetConstraints(&prog, x);
>     EXPECT_EQ(new_vars.rows(), V.vertices().cols());
>     EXPECT_FALSE(new_constraints.empty());
> ```

nit I think its not too much of a maintenance burden to check an exact number here, and provides a real improvement to the level of implementation correctness we're checking for?



_Suggestion:_
```cpp
 EXPECT_EQ(new_constraints.size(), ...);
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 5 unresolved discussions

___
*[`geometry/optimization/test/minkowski_sum_test.cc` line 51 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIFKQ7CipZchY3uOt:-NYUJkCCEqp9vpBQ5miL:b-tzdksk) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/test/minkowski_sum_test.cc#L51)):*

nit I think its not too much of a maintenance burden to check an exact number here, and provides a real improvement to the level of implementation correctness we're checking for?
</blockquote></details>

I actually initially checked `EXPECT_EQ(new_constraints.size(), ...)`, and then realized the check is too brittle. There are many ways to write the constraints, with different number of solvers::Binding objects. For example in https://github.com/RobotLocomotion/drake/blob/d1a5989a5f0f21d823c3743f44a6ff86ce41f6ed/geometry/optimization/minkowski_sum.cc#L146-L150, it adds the equality constraint for each row separately. It is not hard to imagine that we can combine this for loop into just one single constraint, that adds the equality constraint for all rows simultaneously. Hence I think checking the exact number would be a brittle test.
___
*[`geometry/optimization/test/vpolytope_test.cc` line 132 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIWWRDXUgLlt1ppVz:-NYUKKkgBxaxNVW8Oz1S:b-a2pn5t) ([raw file](https://github.com/RobotLocomotion/drake/blob/25bd29508991724c353cf0ef988045367b6e6930/geometry/optimization/test/vpolytope_test.cc#L132)):*

nit I think its not too much of a maintenance burden to check an exact number here, and provides a real improvement to the level of implementation correctness we're checking for?
</blockquote></details>

Same as in the `minkowski_sum_test` case, checking the number of new constraints will be very brittle. For example it adds these two constraints separately https://github.com/RobotLocomotion/drake/blob/d1a5989a5f0f21d823c3743f44a6ff86ce41f6ed/geometry/optimization/vpolytope.cc#L352-L358, but one can also imagine to write it as a single linear equality constraint.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 5 unresolved discussions

___
*[`geometry/optimization/test/minkowski_sum_test.cc` line 51 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIFKQ7CipZchY3uOt:-NYULNusDVtmG_0Qkc4m:b-14unnj) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/test/minkowski_sum_test.cc#L51)):*

I actually initially checked `EXPECT_EQ(new_constraints.size(), ...)`, and then realized the check is too brittle. There are many ways to write the constraints, with different number of solvers::Binding objects. For example in https://github.com/RobotLocomotion/drake/blob/d1a5989a5f0f21d823c3743f44a6ff86ce41f6ed/geometry/optimization/minkowski_sum.cc#L146-L150, it adds the equality constraint for each row separately. It is not hard to imagine that we can combine this for loop into just one single constraint, that adds the equality constraint for all rows simultaneously. Hence I think checking the exact number would be a brittle test.
</blockquote></details>

Perhaps is there a lower bound we can test for?  I'm worried that `>= 1` is too loose?  Maybe its fine.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: 3 unresolved discussions

___
*[`geometry/optimization/intersection.cc` line 113 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUI0aK1zrIT9JS_wCf:-NYUcvOc8fDiSDmilgrF:b-ki5kr2) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/intersection.cc#L113)):*

The return value logic here is non-trival.  I think we need a unit test to check it specifically?
</blockquote></details>

Done. I added a test on the size of new_constraints.
___
*[`geometry/optimization/spectrahedron.cc` line 97 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUHjwK-QlqHVx1OiSc:-NYUeMQQ7IFCh7jcQmTP:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/spectrahedron.cc#L97)):*

I think its important to have a spectrahedron unit test that counts the >1 number of `new_constraints` that get returned.
</blockquote></details>

Done.
___
*[`geometry/optimization/test/minkowski_sum_test.cc` line 51 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIFKQ7CipZchY3uOt:-NYUeUY7Cyym_1LmDBJC:b-oz8tym) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5cb33ce1f328ae70724d4dd2ad0dbf7912c3cb5/geometry/optimization/test/minkowski_sum_test.cc#L51)):*

Perhaps is there a lower bound we can test for?  I'm worried that `>= 1` is too loose?  Maybe its fine.
</blockquote></details>

Good call.
___
*[`geometry/optimization/test/vpolytope_test.cc` line 132 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19634#-NYUIWWRDXUgLlt1ppVz:-NYUfJSe470NuuRSoRgV:bmzkm0b) ([raw file](https://github.com/RobotLocomotion/drake/blob/25bd29508991724c353cf0ef988045367b6e6930/geometry/optimization/test/vpolytope_test.cc#L132)):*

Same as in the `minkowski_sum_test` case, checking the number of new constraints will be very brittle. For example it adds these two constraints separately https://github.com/RobotLocomotion/drake/blob/d1a5989a5f0f21d823c3743f44a6ff86ce41f6ed/geometry/optimization/vpolytope.cc#L352-L358, but one can also imagine to write it as a single linear equality constraint.
</blockquote></details>

Done, I added a test.

jwnimmer-tri (commented):
> Reviewed 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19634)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),cohnt

PR #19638, opened 2023-06-20
[py autodiff] Handle empty derivatives in pydrake.forwarddiff.jacobian.

### LINKED ISSUE #19584: AutoDiffXd Operations in Python May Not Preserve Derivative Shape
### What happened?

Certain operations in `pydrake.math`, including `max` and `min`, may end up changing the number of derivatives an `AutoDiffXd` variable has. This can become a major issue down the line -- for example, when `jacobian` on a function, it expects the number of derivatives to be the same for all output values, in order to construct the Jacobian matrix. Here's a simple example of code that exhibits such a problem:
```
import numpy as np
from pydrake.all import AutoDiffXd, jacobian
import pydrake.math

def func(q):
	# q should be of length 2
	return np.array([q[0], pydrake.math.min(q[1], 1)])

x = np.array([0, 0])
jacobian(func, x)

y = np.array([0, 2])
jacobian(func, y)
```
The first jacobian computation works as expected, whereas the second throws an error due to the derivative shapes not lining up.

Here's an example in general of derivative shapes changing when calling `min` and `max`, or not being checked:
```
import numpy as np
from pydrake.all import AutoDiffXd
import pydrake.math

x = AutoDiffXd(0, np.zeros(2))
y = AutoDiffXd(1, np.zeros(3))

print(pydrake.math.min(x, y))
print(pydrake.math.max(x, y))
```

### Version

_No response_

_No response_

_No response_

### Relevant log output

_No response_

cohnt:
Resolves #19584

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19638)

### DIFF
--- a/bindings/pydrake/BUILD.bazel
+++ b/bindings/pydrake/BUILD.bazel
@@ -441,6 +441,7 @@ drake_py_unittest(
     name = "forward_diff_test",
     deps = [
         ":autodiffutils_py",
+        ":math_py",
         "//bindings/pydrake/common/test_utilities",
     ],
 )
--- a/bindings/pydrake/forwarddiff.py
+++ b/bindings/pydrake/forwarddiff.py
@@ -53,8 +53,13 @@ def jacobian(function, x):
         der[i] = 1
         x_ad.flat[i] = AutoDiffXd(x.flat[i], der)
     y_ad = np.asarray(function(x_ad))
+    # An AutoDiffXd variable with empty derivatives should be treated as
+    # having all zero derivatives. Checking the length of the derivatives
+    # vector, and replacing it with all zeros ensures np.vstack doesn't
+    # throw an error when the shapes don't line up.
     return np.vstack(
-        [y.derivatives() for y in y_ad.flat]).reshape(y_ad.shape + (-1,))
+        [y.derivatives() if len(y.derivatives()) > 0 else np.zeros(x.size)
+            for y in y_ad.flat]).reshape(y_ad.shape + (-1,))
 
 
 # Method overloads:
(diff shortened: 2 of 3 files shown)


jwnimmer-tri:

+@jwnimmer-tri for platform review per schedule.

hongkai-dai (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19638#-:-NYPCoS7E7tBurb_fLgL:bg7igkv)* status: LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri (commented):
:lgtm: platform, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19638#-:-NYPqpI2Dv_hA9Mublmm:bhm0jc2)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

___
*[`bindings/pydrake/test/forward_diff_test.py` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19638#-NYPqrel-YKSLgpencwA:-NYPqremBwwWUuiHNDnc:b909oru) ([raw file](https://github.com/RobotLocomotion/drake/blob/52dce675978125d1a937d96204210eb4c2dcaa91/bindings/pydrake/test/forward_diff_test.py#L90)):*
> ```Python
>     def test_jacobian_derivative_shape(self):
>         # Test that jacobian can handle AutoDiffXd entries with
>         # empty derivative vectors.
> ```

FYI This comment style is consistent with the rest of this 6+ year old file, but in new (modern) code we would use docstring quoting (`"""`) instead of `#`.  That allows `unittest` to display the test summary docstring as part of the test output upon failure, which is a small convenience for developers.

PR #19663, opened 2023-06-22
[multibody] Select a subset of hydroelastic ContactResults

### LINKED ISSUE #19592: [multibody] Select only contact results of a specified geometry
Related to
- #19494 

For visualization, in #19494 we found a use case to both accelerate MeshCat and also help users to focus on contact patches of the actuator and to skip drawing contact patches among free bodies.  In that example, instead of drawing 16 table-cylinder contact patches, we would draw only a few paddle-cylinder contact patches. A [prototype](https://github.com/RobotLocomotion/drake/compare/master...DamrongGuoy:drake:clutter_sim_DG_2023-06-08) showed that Gamepad response time becomes almost real-time.

In [multibody/plant/contact_results.h](https://github.com/RobotLocomotion/drake/compare/master...DamrongGuoy:drake:clutter_sim_DG_2023-06-08#diff-c237799815d208d42eef1fe795c2981d011b7bc5edb30b167999c65d7801635b), provide this advanced API

```C++
  /** (Advanced) Experimental API to select all hydroelastic contacts with a
   specific collision geometry. It enables users to process only contact info
   of a manipulator and skip contact info between free bodies.
   See issue [19494](https://github.com/RobotLocomotion/drake/issues/19494). */
  ContactResults Select(geometry::GeometryId selector) const;
```
With that API, users can make a simple system to select desirable contact patches for visualization and also downstream processing (e.g., sim2real verification) [like this.](https://github.com/DamrongGuoy/drake/blob/b62403f4fceed9bda2a88f40e39ef4b469a0b7c1/examples/clutter/clutter_teleop_sim.py#L145-L179)

Expose the four [hidden APIs ](https://github.com/RobotLocomotion/drake/blob/9e49f35e858e372d7a5f4c58c7419febc2381599/multibody/plant/contact_results.h#L63-L81) of ContactResults to public.  However, I think that would be overkill.

**Beneficiary**

@rcory @rthomasson23 of TRI and Stanford.

**Additional context**
Add any other context or screenshots about the feature request here.  The following movie shows only the contact patches on the paddle as it moves around. Users can focus on the paddle contacts without distraction from other contacts between free bodies.

https://github.com/RobotLocomotion/drake/assets/42557859/6421c7bb-ab4d-480d-9397-2e5f78e45ed9

DamrongGuoy:
Fix #19592

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -115,7 +115,9 @@ void DoScalarDependentDefinitions(py::module m, T) {
             cls_doc.num_hydroelastic_contacts.doc)
         .def("hydroelastic_contact_info", &Class::hydroelastic_contact_info,
             py::arg("i"), cls_doc.hydroelastic_contact_info.doc)
-        .def("plant", &Class::plant, py_rvp::reference, cls_doc.plant.doc);
+        .def("plant", &Class::plant, py_rvp::reference, cls_doc.plant.doc)
+        .def("SelectHydroelastic", &Class::SelectHydroelastic,
+            py::arg("selector"), cls_doc.SelectHydroelastic.doc);
     DefCopyAndDeepCopy(&cls);
     AddValueInstantiation<Class>(m);
   }
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -2446,6 +2446,7 @@ class TestPlant(unittest.TestCase):
         self.assertTrue(contact_results.num_point_pair_contacts() == 0)
         self.assertIsNone(contact_results.plant())
         copy.copy(contact_results)
+        contact_results.SelectHydroelastic(selector=lambda _: True)
 
     def test_contact_model(self):
         plant = MultibodyPlant_[float](0.1)
--- a/multibody/plant/BUILD.bazel
+++ b/multibody/plant/BUILD.bazel
@@ -846,6 +846,14 @@ drake_cc_googletest(
 )
 
 drake_cc_googletest(
+    name = "contact_results_test",
+    deps = [
+        ":contact_results",
+        ":multibody_plant_core",
+    ],
+)
+
+drake_cc_googletest(
     name = "contact_results_to_lcm_test",
     deps = [
         ":contact_results_to_lcm",
--- a/multibody/plant/contact_results.cc
+++ b/multibody/plant/contact_results.cc
@@ -111,6 +111,29 @@ int ContactResults<T>::num_hydroelastic_contacts() const {
   }
 }
 
+template <typename T>
+ContactResults<T> ContactResults<T>::SelectHydroelastic(
+    std::function<bool(const HydroelasticContactInfo<T>&)> selector) const {
+  ContactResults<T> selected_alias_pointers;
+  if (this->plant() != nullptr) {
+    selected_alias_pointers.set_plant(this->plant());
+  }
+  int num_hydroelastic_contacts = this->num_hydroelastic_contacts();
+  for (int i = 0; i < num_hydroelastic_contacts; ++i) {
+    const HydroelasticContactInfo<T>& contact_info =
+        this->hydroelastic_contact_info(i);
+    if (selector(contact_info)) {
+      selected_alias_pointers.AddContactInfo(&contact_info);
+    }
+  }
+  // Deep copy of the selected hydroelastic contact info.
+  ContactResults<T> output_unique_pointers = selected_alias_pointers;
+  // Deep copy of the point pair contact info.
+  output_unique_pointers.point_pairs_info_ = this->point_pairs_info_;
+
+  return output_unique_pointers;
+}
+
 }  // namespace multibody
 }  // namespace drake
 
--- a/multibody/plant/contact_results.h
+++ b/multibody/plant/contact_results.h
@@ -58,6 +58,17 @@ class ContactResults {
   result will be non-null, but default-constructed results might have nulls. */
   const MultibodyPlant<T>* plant() const;
 
+  /** Returns ContactResults with only HydroelasticContactInfo instances
+   satisfying the selection criterion and with all PointPairContactInfo
+   instances.
+
+   @param selector Boolean predicate that returns true to select which
+                   HydroelasticContactInfo.
+
+   @note It uses deep copy. */
+  ContactResults<T> SelectHydroelastic(
+      std::function<bool(const HydroelasticContactInfo<T>&)> selector) const;
+
   // The following methods should only be called by MultibodyPlant and testing
   // fixtures and are undocumented rather than being made private with friends.
   #ifndef DRAKE_DOXYGEN_CXX
--- a/multibody/plant/test/contact_results_test.cc
+++ b/multibody/plant/test/contact_results_test.cc
@@ -1,0 +1,165 @@
+#include "drake/multibody/plant/contact_results.h"
+
+#include <gtest/gtest.h>
+
+#include "drake/multibody/plant/multibody_plant.h"
+
+namespace drake {
+namespace multibody {
+
+using Eigen::Vector3d;
+using geometry::ContactSurface;
+using geometry::GeometryId;
+using geometry::PolygonSurfaceMesh;
+using geometry::PolygonSurfaceMeshFieldLinear;
+using geometry::PenetrationAsPointPair;
+
+class ContactResultsTest : public ::testing::Test {
+ public:
+  void SetUp() override {
+    // Empty surface mesh for hydroelastic contact surface
+    auto surface_mesh = std::make_unique<PolygonSurfaceMesh<double>>();
+    // Empty pressure field for hydroelastic contact surface
+    auto field =
+        std::make_unique<PolygonSurfaceMeshFieldLinear<double, double>>(
+            std::vector<double>{}, surface_mesh.get(), false);
+    contact_surface_ = std::make_unique<ContactSurface<double>>(
+        GeometryId::get_new_id(), GeometryId::get_new_id(),
+        std::move(surface_mesh), std::move(field));
+    my_hydroelastic_contact_info_ =
+        std::make_unique<HydroelasticContactInfo<double>>(
+            contact_surface_.get(), SpatialForce<double>{},
+            std::vector<HydroelasticQuadraturePointData<double>>{});
+  }
+
+ protected:
+  const MultibodyPlant<double> plant_{/* time_step */ 0};
+  const PointPairContactInfo<double> point_pair_info_{
+      BodyIndex{0},
+      BodyIndex{1},
+      Vector3d::Zero(),
+      Vector3d::Zero(),
+      0,
+      0,
+      PenetrationAsPointPair<double>{}};
+
+  std::unique_ptr<ContactSurface<double>> contact_surface_;
+  std::unique_ptr<HydroelasticContactInfo<double>>
+      my_hydroelastic_contact_info_;
+};
+
+TEST_F(ContactResultsTest, Default) {
+  const ContactResults<double> default_contact_results;
+  EXPECT_EQ(default_contact_results.num_point_pair_contacts(), 0);
+  EXPECT_EQ(default_contact_results.num_hydroelastic_contacts(), 0);
+  EXPECT_EQ(default_contact_results.plant(), nullptr);
+}
+
+// Tests the set, access, and clear operations. For the access and clear
+// operations, the code has two branches that:
+//   1. use alias pointers for hydroealstic_contact_info_
+//   2. use unique pointers for hydroelastic_contact_info_ (happens after
+//      assignment or copy constructor).
+TEST_F(ContactResultsTest, SetAccessClear) {
+  // Set
+  // It will use alias pointer to my_hydroelastic_contact_info_.
+  ContactResults<double> contact_results;
+  contact_results.set_plant(&plant_);
+  contact_results.AddContactInfo(point_pair_info_);
+  contact_results.AddContactInfo(my_hydroelastic_contact_info_.get());
+
+  // Access
+  EXPECT_EQ(contact_results.plant(), &plant_);
+  EXPECT_EQ(contact_results.num_point_pair_contacts(), 1);
+  EXPECT_EQ(contact_results.num_hydroelastic_contacts(), 1);
+  EXPECT_EQ(contact_results.point_pair_contact_info(0).bodyA_index(),
+            point_pair_info_.bodyA_index());
+  EXPECT_EQ(&contact_results.hydroelastic_contact_info(0),
+            my_hydroelastic_contact_info_.get());
+  // `copy` will use unique pointers for hydroelastic_contact_info_.
+  ContactResults<double> copy(contact_results);
+  EXPECT_EQ(copy.num_hydroelastic_contacts(), 1);
+  EXPECT_NE(&copy.hydroelastic_contact_info(0),
+            my_hydroelastic_contact_info_.get());
+  // Clear
+  // Test the alias-pointer variant.
+  contact_results.Clear();
+  EXPECT_EQ(contact_results.plant(), nullptr);
+  EXPECT_EQ(contact_results.num_point_pair_contacts(), 0);
+  EXPECT_EQ(contact_results.num_hydroelastic_contacts(), 0);
+  // Test the unique-pointer variant.
+  copy.Clear();
+  EXPECT_EQ(copy.num_hydroelastic_contacts(), 0);
+}
+
+// This is just a coverage test of the assignment operator. It doesn't try to
+// verify the results of assignment operator rigorously. It only checks that
+// the three member variables are valid without looking at their values.
+//
+// The operator manages two variants of hydroelastic_contact_info_, which
+// can be a vector of aliased pointers or a vector of unique pointers.
+// We check the pointer values as returned by the
+// hydroelastic_contact_info(int) function to distinguish the two cases.
+TEST_F(ContactResultsTest, AssignHydroelasticContactInfo) {
+  // Assign empty RHS to LHS.
+  {
+    ContactResults<double> lhs;
+    const ContactResults<double> rhs_empty;
+    lhs = rhs_empty;
+    EXPECT_EQ(lhs.num_hydroelastic_contacts(), 0);
+    EXPECT_EQ(lhs.num_point_pair_contacts(), 0);
+    EXPECT_EQ(lhs.plant(), nullptr);
+  }
+  // Assign non-empty RHS to LHS.
+  // LHS will create unique pointers of copies of hydroelastic contacts.
+  {
+    ContactResults<double> lhs;
+    ContactResults<double> rhs_non_empty;
+    rhs_non_empty.set_plant(&plant_);
+    rhs_non_empty.AddContactInfo(point_pair_info_);
+    rhs_non_empty.AddContactInfo(my_hydroelastic_contact_info_.get());
+
+    lhs = rhs_non_empty;
+    EXPECT_EQ(lhs.num_hydroelastic_contacts(), 1);
+    EXPECT_EQ(lhs.num_point_pair_contacts(), 1);
+    EXPECT_EQ(lhs.plant(), &plant_);
+    // Confirm that the assignment operator copies the content to
+    // a different memory address.
+    EXPECT_NE(&lhs.hydroelastic_contact_info(0),
+              &rhs_non_empty.hydroelastic_contact_info(0));
+  }
+}
+
+TEST_F(ContactResultsTest, SelectHydroelastic) {
+  ContactResults<double> contact_results;
+  contact_results.set_plant(&plant_);
+  contact_results.AddContactInfo(point_pair_info_);
+  contact_results.AddContactInfo(my_hydroelastic_contact_info_.get());
+
+  const ContactResults<double> no_hydro_contacts =
+      contact_results.SelectHydroelastic(
+          [](const HydroelasticContactInfo<double>&) -> bool {
+            return false;
+          });
+  EXPECT_EQ(no_hydro_contacts.num_point_pair_contacts(), 1);
+  EXPECT_EQ(no_hydro_contacts.num_hydroelastic_contacts(), 0);
+
+  const ContactResults<double> one_hydro_contact =
+      contact_results.SelectHydroelastic(
+          [](const HydroelasticContactInfo<double>&) -> bool {
+            return true;
+          });
+  EXPECT_EQ(one_hydro_contact.num_point_pair_contacts(), 1);
+  EXPECT_EQ(one_hydro_contact.num_hydroelastic_contacts(), 1);
+  // Sanity check of hydroelastic contact info
+  EXPECT_EQ(
+      one_hydro_contact.hydroelastic_contact_info(0).contact_surface().id_M(),
+      my_hydroelastic_contact_info_->contact_surface().id_M());
+  // Verify the deep copy by checking for different memory address.
+  EXPECT_NE(
+      &one_hydro_contact.hydroelastic_contact_info(0),
+      my_hydroelastic_contact_info_.get());
+}
+
+}  // namespace multibody
+}  // namespace drake


DamrongGuoy (commented):

CC: @rcory @amcastro-tri 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYZit1O56CBe37NnrBC:b-5b2kx4)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

DamrongGuoy (commented):
For completion, I also set up unit tests of other functions in ContactResults.  They were covered by other system tests already but didn't have their own unit tests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYZl5-p4q8HnMfID_lA:b-cku2yz)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

jwnimmer-tri:
I'm pretty sure that aliasing the storage of a different `ContactResults` object will be unsafe in your desired use case.  I would bet money that you need to copy, not alias.  If you and Sherm want any help getting to the bottom of that, let me know.

It would help clarify if you could link to sample code for how to plan to use this feature.  If it's https://github.com/DamrongGuoy/drake/commit/e0a7f43bb9367f54d3d7a49b82b2a2c2e6e7f098, then you can see how `FilterContactResultsOutput` sets its output to be pointers into the input port's storage.  That's unsafe.

sherm1 (commented):
I looked this over with Damrong yesterday for the use case of visualizing only a subset of the contact results via a LeafSystem that takes the full ContactResults object on an input port and puts out a trimmed-down ContactResults object on its output port. I couldn't see anything that would be unsafe -- the output ContactResult object has pointers into the input ContactResult object (both of which exist in the same Context). Then the output object is used to construct messages to feed to Meshcat (presumably serialized without any pointers). The pointers are recalculated whenever there is a change; they don't persist across steps, for example. Do you see something unsafe that I missed? Or are you anticipating some case where the pointers persist past their lifetime?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NY_2Q2RCP0hXHSkoRwi:b-urqszc)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

jwnimmer-tri:
> Or are you anticipating some case where the pointers persist past their lifetime?

Why even take the risk?  This is a public function, we don't know how users will use it.  I would be surprised if copying the results has any performance impediment, given that we're already eating an indirect virtual function call for every contact pair (in the bool predicate).  I would think the slow part we care about is the visualization processing, not a little extra copying in the framework.

This implementation is only safe when used to alias an output port against an input port (so that caching keeps the two in sync).  Most other kinds of local manipulation of this value by a user would be subject to either dangling references or out-of-date values.

Type-erased values in the framework are supposed to have value semantics, not reference semantics.  Placing reference semantics at a user's fingertips is asking for trouble.

jwnimmer-tri:
(If the copying does profile as being too costly, then the better solution is to implement copy-on-write inside of ContactResults.)

sherm1 (commented):
It is conventional to avoid copying meshes when possible, so this makes sense to me. (Hydro results can be big meshes) I agree this shouldn't be misused though but would be fine with wrapping it in danger signals like "(Advanced)" "Don't use this if you don't understand the lifetime issues" or similar. IMO this function is well-fit for the intended purpose, and the ContactResults class was carefully engineered to support shallow references. We just need to make sure it is clear what the restrictions are. If you are very uncomfortable with it, how about "(Internal use only)"?

Pending adequate messaging and restriction, feature :lgtm:

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NY_7lfa7ouV0LhH24JA:b-em65ro)* status: 8 unresolved discussions, needs at least two assigned reviewers

___
*[`multibody/plant/contact_results.h` line 74 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_4bF36wYgn1yerPlE:-NY_4bF36wYgn1yerPlF:b-c7pq10) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/contact_results.h#L74)):*
> ```objc
> 
>    @note Assume the lifetime of `this` covers the lifetime of
>          the return value. */
> ```

BTW this is an important note to make this safe. Consider strengthening it some, maybe "This object _must_ outlive the returned view and the view _must_ be recalculated whenever there is _any_ change to `this` ContactResults object."
___
*[`multibody/plant/test/contact_results_test.cc` line 61 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_5w5l88J_Ad4LJogG:-NY_5w5l88J_Ad4LJogH:bhr172f) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L61)):*
> ```cpp
> // operations, the code has two branches that:
> //   1. use alias pointers for hydroealstic_contact_info_
> //   2. use unique pointers for hydroelastic_contact_info_ (happened after
> ```

nit: happened -> happens
___
*[`multibody/plant/test/contact_results_test.cc` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6GRWACol15NLnZXz:-NY_6GRWACol15NLnZY-:b-rb9qgl) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L88)):*
> ```cpp
> 
>   // Clear
>   // Test the alias-pointer variance.
> ```

nit: variance -> variant
___
*[`multibody/plant/test/contact_results_test.cc` line 93 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6L0b1TTj8QfszXK8:-NY_6L0b1TTj8QfszXK9:b-rb9qgl) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L93)):*
> ```cpp
>   EXPECT_EQ(contact_results.num_point_pair_contacts(), 0);
>   EXPECT_EQ(contact_results.num_hydroelastic_contacts(), 0);
>   // Test the unique-pointer variance.
> ```

nit: variance -> variant
___
*[`multibody/plant/test/contact_results_test.cc` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6ThTAp9dwpXqq4wM:-NY_6ThTAp9dwpXqq4wN:btdg7d1) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L102)):*
> ```cpp
> // the three member variables are valid without looking at their values.
> //
> // The operator manages two variances of hydroelastic_contact_info_, which
> ```

nit: variances -> variants
___
*[`multibody/plant/test/contact_results_test.cc` line 104 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6XRT4j5U5VC-2psT:-NY_6XRT4j5U5VC-2psU:b-rb9qgl) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L104)):*
> ```cpp
> // The operator manages two variances of hydroelastic_contact_info_, which
> // can be a vector of aliased pointers or a vector of unique pointers. Since
> // we can't check which variance the current hydroelastic_contact_info_ is,
> ```

nit: variance -> variant
___
*[`bindings/pydrake/multibody/plant_py.cc` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_7FoG5UDEWc_YCIHY:-NY_7FoG5UDEWc_YCIHZ:b-8ky95o) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/bindings/pydrake/multibody/plant_py.cc#L120)):*
> ```cpp
>         .def("plant", &Class::plant, py_rvp::reference, cls_doc.plant.doc)
>         .def("ViewHydroelastic", &Class::ViewHydroelastic,
>             cls_doc.ViewHydroelastic.doc);
> ```

minor: should include `py::arg("selector")`
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2452 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_7Tp70JPQH_SlY5w2:-NY_7Tp89NL-qIiFon3W:b-hpvso0) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/bindings/pydrake/multibody/test/plant_test.py#L2452)):*
> ```Python
>         def myselector(hydroelastic_contact_info):
>             return True
>         contact_results_view = contact_results.ViewHydroelastic(myselector)
> ```

minor: should test kwargs



_Suggestion:_
```Python
ViewHydroelastic(selector=myselector)
```

sherm1 (commented):
+@jwnimmer-tri for platform review per schedule, please

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NY_9B1s29-2Z9hXg-So:b-okvarn)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri:
@DamrongGuoy is your plan to tell users to call this function directly in their own code, or is there another PR coming where we provide some kind of ContactFilterSystem that wraps this feature?

DamrongGuoy (commented):
Thanks for the important discussion @jwnimmer-tri @sherm1.  Let me do another test similar to [DamrongGuoy@e0a7f43](https://github.com/DamrongGuoy/drake/commit/e0a7f43bb9367f54d3d7a49b82b2a2c2e6e7f098) with copying data instead of aliasing to be sure.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NY_QelJ0CjCI1-D_vkO:b-i2x4ep)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri:
> If you are very uncomfortable with it, how about "(Internal use only)"?

If the pointer-aliasing optimization was encapsulated within a Drake-provided `ContactFilterSystem` such that it was only used for output port(s), I would be OK with that.  In the current version, the hazard of calling the function in places other than output ports seems too great.

If I were running the show, I'd just implement COW here because it should only take ~30 minutes of work and then its solved forever, for all other use cases beyond this narrow one.  With my platform reviewer hat on, though, I can be satisfied with any narrower solution so long as it structurally or programmatically prevents dangling or mis-indexed pointers from being dereferenced.  (Documentation alone is not enough.)

DamrongGuoy (commented):
Switched to copying data instead of aliasing.  The clutter test went well.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYphsaBDxx8oc1oB5kP:b-kadzqj)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

___
*[`multibody/plant/contact_results.h` line 74 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_4bF36wYgn1yerPlE:-NYpigwV6UdgZfcD51sO:b-n5u5q0) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/contact_results.h#L74)):*

BTW this is an important note to make this safe. Consider strengthening it some, maybe "This object _must_ outlive the returned view and the view _must_ be recalculated whenever there is _any_ change to `this` ContactResults object."
</blockquote></details>

Done.  No requirement for the lifetime of `this` anymore.  Now I have a note that it is a deep copy.
___
*[`multibody/plant/test/contact_results_test.cc` line 61 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_5w5l88J_Ad4LJogG:-NYpmBWDEuKnDc9LTh0U:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L61)):*

nit: happened -> happens
</blockquote></details>

Done
___
*[`multibody/plant/test/contact_results_test.cc` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6GRWACol15NLnZXz:-NYpmCS268CstV5WZs3n:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L88)):*

nit: variance -> variant
</blockquote></details>

Done
___
*[`multibody/plant/test/contact_results_test.cc` line 93 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6L0b1TTj8QfszXK8:-NYpmE0_BYgHdUlP6IBO:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L93)):*

nit: variance -> variant
</blockquote></details>

Done
___
*[`multibody/plant/test/contact_results_test.cc` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6ThTAp9dwpXqq4wM:-NYpmF6R1W4HTHMVUMdM:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L102)):*

nit: variances -> variants
</blockquote></details>

Done
___
*[`multibody/plant/test/contact_results_test.cc` line 104 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_6XRT4j5U5VC-2psT:-NYpmGiaDKE6QWfTXaID:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/multibody/plant/test/contact_results_test.cc#L104)):*

nit: variance -> variant
</blockquote></details>

Done
___
*[`bindings/pydrake/multibody/plant_py.cc` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_7FoG5UDEWc_YCIHY:-NYpmOAj8WT1O9P-6Duh:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/bindings/pydrake/multibody/plant_py.cc#L120)):*

minor: should include `py::arg("selector")`
</blockquote></details>

Done
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2452 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NY_7Tp70JPQH_SlY5w2:-NYpmQCKCsBbrU-R94ko:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/04b6fc4720198595b2d1418dff95f0ab25f4b5e2/bindings/pydrake/multibody/test/plant_test.py#L2452)):*

minor: should test kwargs
</blockquote></details>

Done

DamrongGuoy (commented):
> (1) Is your plan to tell users to call this function directly in their own code, or is there another PR coming where we provide some kind of ContactFilterSystem that wraps this feature?
>

I'll tell users to call this function directly when they make their own systems. I don't have another PR for the ContactFilterSystem wrapper.
 
> (2) When you tried copying the data instead of borrowing pointers, did it run too slowly?

The test went well with copying.

In the future, we can do COW to clean it up (I only read about it but never did COW).  I agree that the existing `variant<alias_pointer, unique_pointer>` is less satisfying.  Thanks for the idea.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYppFm0BMg_mAqG6JWa:bevk0o6)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

jwnimmer-tri (commented):
Still some minor lack of clarity on the filtering (see below).

Also for the record -- I would have thought that binding `AddContactInfo` in Python would have been a much simpler solution. Then the filtering is totally flexible and up to the user.  (Container types without an "Add" function are always a huge headache.  We could fix that once and for all.)

> Reviewed 1 of 6 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYsnVWwEPypbVJAKG1N:b7n6hqz)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @DamrongGuoy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYsmwMO1PbGlsQmo9ZV:bmdxnxi) (no related file):*
> I'll tell users to call this function directly when they make their own systems. I don't have another PR for the ContactFilterSystem wrapper.

This new function discards all point contact information.  If the user's simulation has both point contacts and hydroelastic contacts, they won't see the point contacts at all ever.  Why do we think that's what users will want?

I thought the goal here was to remove some uninteresting-and-expensive-to-display hydro contacts, but not affect the point contact arrows at all.

___
___
*[`multibody/plant/contact_results.h` line 62 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYskDuqB7KuFuuhTkQT:-NYskDur0maQgy9kp2lT:bbl2j5v) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.h#L62)):*
> ```objc
> 
>   /** Returns ContactResults with only HydroelasticContactInfo instances
>    satisfying the selection criterion.
> ```

nit This documentation does not crisply enough state that all non-hydro contacts are discarded.  It's implied by this sentence, but I could easily see a user misreading it that the point contacts are copied unchanged.
___
*[`multibody/plant/contact_results.h` line 68 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsj3taBigA33jgd5is:-NYsj3taBigA33jgd5it:bm7jd5j) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.h#L68)):*
> ```objc
> 
>    @note It copies the selected HydroelasticContactInfo with deep copy. */
>   ContactResults<T> ViewHydroelastic(
> ```

Here and in the PR title, typically "View" implies that as the underlying object changes, the view will update.  We could try to work around that with documentation, but probably it's better just to use a more meaningful name in the first place.

Maybe `SelectHydroelastic`?  That matches `selector` argument name.
___
*[`multibody/plant/contact_results.cc` line 122 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYskmA3-BeijYVunckx:-NYskmA3-BeijYVuncky:b61dn62) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.cc#L122)):*
> ```cpp
>   }
>   int num_hydroelastic_contacts = this->num_hydroelastic_contacts();
>   for (int i=0; i < num_hydroelastic_contacts; ++i) {
> ```

nit

_"Assignment operators always have spaces around them."
-- https://drake.mit.edu/styleguide/cppguide.html#Horizontal_Whitespace



_Suggestion:_
```cpp
int i = 0
```
___
*[`multibody/plant/test/contact_results_test.cc` line 80 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYssEvgAEvkzsypbDdm:-NYssEvgAEvkzsypbDdn:b-qo2qfz) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L80)):*
> copy will use unique pointers for hydroelastic_contact_info_.

The test case doesn't actually check this, but it should.  Testing the value of `id_M` doesn't disambiguate whether anything was copied or not.

```c++
EXPECT_NE(
    &copy.hydroelastic_contact_info(0),
    my_hydroelastic_contact_info_.get());
```

For symmetry, we could also check `EXPECT_EQ` up above for the `Access` testing of the same pointer comparison.
___
*[`multibody/plant/test/contact_results_test.cc` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYst05s-saOXqgZMJ5M:-NYst05s-saOXqgZMJ5N:b-tf4is0) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L105)):*
> ```cpp
> // can be a vector of aliased pointers or a vector of unique pointers. Since
> // we can't check which variant the current hydroelastic_contact_info_ is,
> // we only check the number of records.
> ```

We can't check the variant exactly, but we can easily check the pointer values as returned by the `hydroelastic_contact_info(iint)` function, to see what is or isn't aliased.  That would vastly improve this test case to actually mean something.
___
*[`multibody/plant/test/contact_results_test.cc` line 152 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYstHHt7p4Sws_GJfAB:-NYstHHt7p4Sws_GJfAC:bl33cyz) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L152)):*
> ```cpp
> 
> // The copy constructor uses the assignment operator. This is just a coverage
> // test.
> ```

We already covered this up on line 81:

```c++
ContactResults<double> copy(contact_results);
```

Why do we also need this test case?
___
*[`multibody/plant/test/contact_results_test.cc` line 172 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYstU6l8Ehv4lfAD7RD:-NYstU6mBHUDIYa7Pdtn:b53lkjf) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L172)):*
> ```cpp
> 
>   const ContactResults<double> one_contact = contact_results.ViewHydroelastic(
>       [](const HydroelasticContactInfo<double>& contact) -> bool {
> ```

nit Be consistent (vs the above).



_Suggestion:_
```cpp
(const HydroelasticContactInfo<double>&)
```
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2451 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYslOCwBID7wUbWtgbR:-NYslOCwBID7wUbWtgbS:b-o4hjtt) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/bindings/pydrake/multibody/test/plant_test.py#L2451)):*
> ```Python
> 
>         def myselector(hydroelastic_contact_info):
>             return True
> ```

nit This testing idiom splits the difference between a fully-fledged correctness test versus a sanity-check compile stub.  It's not clear what we're actually trying to test and why.

I think there are two options for making it clear:

(1) Check the callback logic explicitly: test (assert) that the callback function is actually being called (i.e., that there are actually hydro contacts available); and test (assert) that the object passed to the callback argument is the right type.

(2)  Simplify the new testing to just this one line:

```py
contact_results.ViewHydroelastic(selector=lambda _: True)
```

DamrongGuoy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYsuuMyDyqBaDvC17j0:bgp2t8l) (no related file):*

> I'll tell users to call this function directly when they make their own systems. I don't have another PR for the ContactFilterSystem wrapper.

This new function discards all point contact information.  If the user's simulation has both point contacts and hydroelastic contacts, they won't see the point contacts at all ever.  Why do we think that's what users will want?

I thought the goal here was to remove some uninteresting-and-expensive-to-display hydro contacts, but not affect the point contact arrows at all.
</blockquote></details>

Thanks for asking.  In the target application, users are only interested in hydroelastic contact surfaces with the manipulator.  They want to compare the shape of contact patches with physical experiments. Is that right, @rcory? Or do you think we should provide the point contact info as well?

___
___
*[`multibody/plant/contact_results.h` line 62 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYskDuqB7KuFuuhTkQT:-NYt_azg-7GUiMf7U-XE:b8vzp3h) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.h#L62)):*

nit This documentation does not crisply enough state that all non-hydro contacts are discarded.  It's implied by this sentence, but I could easily see a user misreading it that the point contacts are copied unchanged.
</blockquote></details>

Thank you for pointing that out.  I added a sentence to clarify that.
___
*[`multibody/plant/contact_results.h` line 68 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsj3taBigA33jgd5is:-NYt_lO3DooQvBYZZBJT:beygd07) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.h#L68)):*

Here and in the PR title, typically "View" implies that as the underlying object changes, the view will update.  We could try to work around that with documentation, but probably it's better just to use a more meaningful name in the first place.

Maybe `SelectHydroelastic`?  That matches `selector` argument name.
</blockquote></details>

Thank you.  I renamed it to `SelectHydroelastic`.  It's a better name.
___
*[`multibody/plant/contact_results.cc` line 122 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYskmA3-BeijYVunckx:-NYt_yWA2ea_qEr6hTT9:boyexbg) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/contact_results.cc#L122)):*

nit

_"Assignment operators always have spaces around them."
-- https://drake.mit.edu/styleguide/cppguide.html#Horizontal_Whitespace
</blockquote></details>

Done. Thanks.
___
*[`multibody/plant/test/contact_results_test.cc` line 80 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYssEvgAEvkzsypbDdm:-NYt-oqV9zvhBAlSSjOc:b-nngr4f) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L80)):*

> copy will use unique pointers for hydroelastic_contact_info_.

The test case doesn't actually check this, but it should.  Testing the value of `id_M` doesn't disambiguate whether anything was copied or not.

```c++
EXPECT_NE(
    &copy.hydroelastic_contact_info(0),
    my_hydroelastic_contact_info_.get());
```

For symmetry, we could also check `EXPECT_EQ` up above for the `Access` testing of the same pointer comparison.
</blockquote></details>

Done. Thank you.  Checking memory addresses is better.  I added the tests for the alias pointers (EXPECT_EQ) and unique pointers (EXPECT_NE).
___
*[`multibody/plant/test/contact_results_test.cc` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYst05s-saOXqgZMJ5M:-NYth7JB8sFiC8pF7xLq:b3gm2of) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L105)):*

We can't check the variant exactly, but we can easily check the pointer values as returned by the `hydroelastic_contact_info(iint)` function, to see what is or isn't aliased.  That would vastly improve this test case to actually mean something.
</blockquote></details>

Thank you for pointing that out.  I added the comment and modified the test accordingly. I also shortened the tests.
___
*[`multibody/plant/test/contact_results_test.cc` line 152 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYstHHt7p4Sws_GJfAB:-NYthz5z8v13--JG4UPR:b6akvhg) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L152)):*

We already covered this up on line 81:

```c++
ContactResults<double> copy(contact_results);
```

Why do we also need this test case?
</blockquote></details>

I agree that this test is redundant.  I have removed it.
___
*[`multibody/plant/test/contact_results_test.cc` line 172 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYstU6l8Ehv4lfAD7RD:-NYtiE9V49TOJW_02NLQ:boyexbg) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/multibody/plant/test/contact_results_test.cc#L172)):*

nit Be consistent (vs the above).
</blockquote></details>

Done. Thanks.
___
*[`bindings/pydrake/multibody/test/plant_test.py` line 2451 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYslOCwBID7wUbWtgbR:-NYtiJRE-IHX4hAHqV_g:bp3c57) ([raw file](https://github.com/RobotLocomotion/drake/blob/778f9b6acb5e02398dc4cd0f2fd068594f6aa29d/bindings/pydrake/multibody/test/plant_test.py#L2451)):*

nit This testing idiom splits the difference between a fully-fledged correctness test versus a sanity-check compile stub.  It's not clear what we're actually trying to test and why.

I think there are two options for making it clear:

(1) Check the callback logic explicitly: test (assert) that the callback function is actually being called (i.e., that there are actually hydro contacts available); and test (assert) that the object passed to the callback argument is the right type.

(2)  Simplify the new testing to just this one line:

```py
contact_results.ViewHydroelastic(selector=lambda _: True)
```
</blockquote></details>

Thank you.  I appreciate the code. It's only a sanity-check compile stub.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYtmVBz6obeq31M_tIP:b-afkozx) (no related file):*
> They want to compare the shape of contact patches with physical experiments.

Oh.  I had thought this was a visualization performance problem.  You're saying they are logging the data?  Why can't their logger filter the contacts already?

To underline my thesis here: Our goal is to provide general tools that can satisfy many users' needs.  If our tool always nukes the point contacts, then it will be less useful for some users, even if your particular user doesn't care.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYtpK_90IoCpJ_5Ovt3:b-9bn0sc) (no related file):*

> They want to compare the shape of contact patches with physical experiments.

Oh.  I had thought this was a visualization performance problem.  You're saying they are logging the data?  Why can't their logger filter the contacts already?

To underline my thesis here: Our goal is to provide general tools that can satisfy many users' needs.  If our tool always nukes the point contacts, then it will be less useful for some users, even if your particular user doesn't care.
</blockquote></details>

In fact, the call to action from #19592 is to filter contacts related to the end effector.  Discarding point contacts on the end effector actively goes against the stated goal.

___

DamrongGuoy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYtrB-h6RfuWNDnWq1o:bot4lns) (no related file):*

In fact, the call to action from #19592 is to filter contacts related to the end effector.  Discarding point contacts on the end effector actively goes against the stated goal.
</blockquote></details>

I see your points.  I'll change it to preserve point contacts.

Set to working.

___

DamrongGuoy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-NYsmwMO1PbGlsQmo9ZU:-NYuXksPCghtBImzZniE:b-896fix) (no related file):*

I see your points.  I'll change it to preserve point contacts.

Set to working.
</blockquote></details>

Done.

___

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663#-:-NYuZ8DCDXDhTevnT7bq:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19663)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform)

jwnimmer-tri:

jwnimmer-tri:

jwnimmer-tri:
It finally passed.  I'm inclined to think it was just unlucky flakes, unrelated to this PR.  I'll merge now.

DamrongGuoy:
Thank you very much. I appreciate your help.

My local Jammy also passed: `$ bazel  run --config=gcc --compilation_mode=dbg //bindings/pydrake/visualization:py/model_visualizer_test`.  It's CI flakes.

Issue #19604, opened 2023-06-15
Bring RgbdSensorDiscrete up to parity with RgbdSensorAsync


jwnimmer-tri:
In #19553 we added `RgbdSensorAsync` and invented some useful features that should be backported to `RgbdSensorDiscrete`:
- Configure enable / disable of color and depth (to avoid capturing unnecessary images).
- Configure using fps instead of period.
- Maybe offer an immediate constructor (instead of a sensor-wrapper constructor) that takes the parent frame, camera configs, etc. directly.

There's also some related cleanups required:
- "One class per file."  Move into its own file.
- #19198

jwnimmer-tri:
I have a WIP here: https://github.com/jwnimmer-tri/drake/commits/rgbd-sensor-discrete-api.

It requires several prerequisite PRs first.

PR #19664, opened 2023-06-22
[sensors] RgbdSensor can be constructed from color properties only

jwnimmer-tri:
Towards #19662 and https://github.com/RobotLocomotion/drake/issues/19604 and https://github.com/RobotLocomotion/drake/issues/19437.

This centralizes the logic for how to infer the color properties from depth properties (or vice versa), which will be shared between both the discrete and async sensors. (It's also more convenient for users.  The `{Color|Depth}RenderCamera` constructors don't have enough defaults, and this is a step in a better direction.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19664)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_render.cc
+++ b/bindings/pydrake/geometry/geometry_py_render.cc
@@ -163,7 +163,9 @@ void DoScalarIndependentDefinitions(py::module m) {
     cls  // BR
         .def(py::init<Class const&>(), py::arg("other"), "Copy constructor")
         .def(py::init<RenderCameraCore, DepthRange>(), py::arg("core"),
-            py::arg("depth_range"), cls_doc.ctor.doc)
+            py::arg("depth_range"), cls_doc.ctor.doc_2args)
+        .def(py::init<RenderCameraCore>(), py::arg("core"),
+            cls_doc.ctor.doc_1args)
         .def("core",
             static_cast<RenderCameraCore const& (Class::*)() const>(
                 &Class::core),
--- a/bindings/pydrake/systems/sensors_py.cc
+++ b/bindings/pydrake/systems/sensors_py.cc
@@ -206,8 +206,9 @@ PYBIND11_MODULE(sensors, m) {
       m, "RgbdSensor", doc.RgbdSensor.doc);
 
   rgbd_sensor
-      .def(py::init<FrameId, const RigidTransformd&, ColorRenderCamera,
-               DepthRenderCamera>(),
+      .def(py::init<FrameId, const RigidTransformd&,
+               const std::optional<ColorRenderCamera>&,
+               const std::optional<DepthRenderCamera>&>(),
           py::arg("parent_id"), py::arg("X_PB"), py::arg("color_camera"),
           py::arg("depth_camera"),
           doc.RgbdSensor.ctor.doc_individual_intrinsics)
@@ -250,8 +251,8 @@ PYBIND11_MODULE(sensors, m) {
     py::class_<Class, LeafSystem<T>>(m, "RgbdSensorAsync", cls_doc.doc)
         .def(py::init<const geometry::SceneGraph<double>*, FrameId,
                  const math::RigidTransformd&, double, double, double,
-                 std::optional<ColorRenderCamera>,
-                 std::optional<DepthRenderCamera>, bool>(),
+                 const std::optional<ColorRenderCamera>&,
+                 const std::optional<DepthRenderCamera>&, bool>(),
             py::arg("scene_graph"), py::arg("parent_id"), py::arg("X_PB"),
             py::arg("fps"), py::arg("capture_offset"), py::arg("output_delay"),
             py::arg("color_camera"), py::arg("depth_camera") = std::nullopt,
--- a/geometry/render/render_camera.cc
+++ b/geometry/render/render_camera.cc
@@ -1,9 +1,12 @@
 #include "drake/geometry/render/render_camera.h"
 
 #include <cmath>
+#include <limits>
 #include <utility>
 
 #include <fmt/format.h>
+
+#include "drake/common/text_logging.h"
 
 namespace drake {
 namespace geometry {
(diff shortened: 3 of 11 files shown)


jwnimmer-tri:
I still have some tests to shore up, but @SeanCurtis-TRI this is ready for a design once-over in case you're up for it.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19664)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

___
*[`geometry/render/render_camera.cc` line 86 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19664#-NY_a88-6jfYlNKLt0jn:-NY_a88-6jfYlNKLt0jo:b-xosw98) ([raw file](https://github.com/RobotLocomotion/drake/blob/aaf53e6d81bf447a62131ad170a44d457c9a96ab/geometry/render/render_camera.cc#L86)):*
> ```cpp
> 
> namespace {
> /* Returns a suitable DepthRange to assume for the given camera's clipping.  */
> ```

BTW I'm not attached to this _particular_ heuristic.  But I do think we need _some_ default value.  Happy to take suggestions on tweaks.
___
*[`geometry/render/render_camera.cc` line 101 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19664#-NY_aLJS3rY4B6YVdq8P:-NY_aLJS3rY4B6YVdq8Q:b9mqm04) ([raw file](https://github.com/RobotLocomotion/drake/blob/aaf53e6d81bf447a62131ad170a44d457c9a96ab/geometry/render/render_camera.cc#L101)):*
> ```cpp
>   constexpr float kMaxValidDepth16UInMeters =
>       (std::numeric_limits<uint16_t>::max() - 1) / 1000.0;
>   log()->error("{} {} {}", near, far, kMaxValidDepth16UInMeters);
> ```

Working

Remove this before merge (and the include).

SeanCurtis-TRI:
I've taken a quick glance and I think there's one thing that needs to be hashed out.

The approach of inferring depth range, I believe, is a non-starter. For the same reason we don't provide default values for camera width, height, or field of view, we shouldn't provide "automatic" values for depth range. That is a property of a physical camera. The clipping range will encompass everything we want to see in color and label images and is likely to enclose a span that has nothing to do with the depth range of an actual camera.

In my own tests, I tend to derive color from depth (as that is trivial). But depth from color seems fraught.

Some alternate ideas:

1. Provide a sensor constructor that only takes a depth sensor. That way the sensor, in any meaningful way, is fully specified with the minimum number of inputs.
2. Define the ports on the sensor based on the cameras provided and make them optional.
3. Go with what you're doing, but a sensor that has created `DepthRange` from `ClippingRange` should emit a warning if it is ever asked to render a depth image: "The depth range has been created by heuristic and may not model any physical camera."

SeanCurtis-TRI:
(I just recognized that we already have (1).)

jwnimmer-tri:
> The approach of inferring depth range, I believe, is a non-starter. For the same reason we don't provide default values for camera width, height, or field of view, we shouldn't provide "automatic" values for depth range. That is a property of a physical camera.

To me, this is relevant only if we are trying to model an _actual_ physical camera based on the real world.  It doesn't apply in the case of "I'm new to Drake, setting up fisher price my first simulation, and I just want some basic sensor images without typing in a dozen random numbers into my program".

With a strong prejudice, I think all of the setup code in Drake should have default values that provide a reasonable guess.  We're not taking away the ability of the user to customize those values, we're just making it easier to get up and running.

We *already do* provide default values for width, height, etc. in `CameraConfig`.  I do think we should propagate those defaults to `RenderCameraCore` etc. as well.  Those settings have obvious effects on the output.  To help with onramp, I think it's a fair tradeoff to allow them to be omitted, versus the potential confusion of having "wrong" values or not knowing where to look.  It's not like anyone would assume that Drake only supports 640x480. If they want a different resolution, they'll go look up how to change it, and the tutorials and such anyway should how to do it.

> The clipping range will encompass everything we want to see in color and label images and is likely to enclose a span that has nothing to do with the depth range of an actual camera.

Ditto.  We're not bound to match actual physical cameras; we're a simulator and scenarios can be imaginary.

There has be _some_ guess at a default that least inoffensive.

> Define the ports on the sensor based on the cameras provided and make them optional.

I already wrote up a draft of this, in my first attempt.  It's a _lot_ of churn, and particularly confusing when paired with the continuous-time sensor overload for (1).

SeanCurtis-TRI:
I feel like several topics have become mixed together.

1. Creating sensors more easily.
2. Things not having defaults that should.
3. The difference between default and derived.

It seems what you really want is default-initialized `RenderCamera`s. And, as near as I can tell, `RenderCamera`s that can be initialized with an arbitrary subset of values (such that unspecified values go to defaults).

If you had that, then you could have a sensor initialization that takes *no* render camera at all, just a renderer name (that seems the only thing that doesn't handily admit defaulting). If they wanted to specify more, then they would specify those parts of a `RenderCamera` they care about. But, in that case, all other values would go to the default value. In the case, where the default values are incompatible with the specified values, the user would have to look up those values and recognize that they need to change them to be consistent. But, at the very least, the unspecified values would always be the same and not a moving target.

If we defaulted intrinsics, depth ranges and the like, we could then characterize what a default camera describes. Not a particular physical camera, but a generic camera that could physically exist..

jwnimmer-tri:
The missing use case I want to have us achieve is:
- I have a `ColorRenderCamera` somehow.  Maybe some of my lab-mate's code has already created this object.
- I want to create `RgbdSensor` from it and read from all output ports, even the depth port.
- I don't especially care how the depths are clipped, just that the color image is mostly well-related to the depth image.

One way or another, I want us to be able to choose _default_ values for depth_min and depth_max, starting from only `ColorRenderCamera` or `RenderCameraCore`.

It's true that this is a mix of user setup as well as async/discrete refactor, but I think they are inter-related.

jwnimmer-tri:
I'm going to put this one on the back burner for a few days.  I think the incremental step is probably not clear enough on its own; I'll circle back with a full branch of all the sensors at once and we can take it from there.

SeanCurtis-TRI (commented):
For me, the problematic mixing wasn't in the code. It was in the rhetoric. It felt like the justification for one point was being dragged into service for another point.

> I have a `ColorRenderCamera` somehow....

This seems to be the kind of argument that you're least likely to accept from another person. A code change based on a hypothetical use case. It's difficult for me to imagine that someone is instantiating a `RenderSensor` (or one of its kin) that happens to have a `ColorRenderCamera` of unknown provenance. If that lab mate decided to create a `ColorRenderCamera`, then they should've been encouraged to create a `DepthRenderCamera` and just kept that around.

> choose *default* values for `depth_min and `depth_max...

This is a linguistic point I think we aren't seeing eye to eye on. I don't consider it to be a "default" value, if it's derived from other values. For me defaults aren't derived. For you, those terms seem to be more interchangable. Can you confirm this? If this isn't the case, then I fear I'm still not having the same conversation as you are.

> Reviewed 1 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19664#-:-NYb7fH106BOy36y3uJN:b-gqfh93)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

PR #19665, opened 2023-06-23
[solvers] Support upper bound in MinimumValueConstraint.

hongkai-dai:
Now we support lower <= min(f(x)) <= upper.

Towards #18830

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)

### DIFF
--- a/multibody/inverse_kinematics/minimum_distance_constraint.h
+++ b/multibody/inverse_kinematics/minimum_distance_constraint.h
@@ -111,7 +111,7 @@ class MinimumDistanceConstraint final : public solvers::Constraint {
 
   /** Getter for the minimum distance. */
   double minimum_distance() const {
-    return minimum_value_constraint_->minimum_value();
+    return minimum_value_constraint_->minimum_value_lower();
   }
 
   /** Getter for the influence distance. */
--- a/solvers/minimum_value_constraint.cc
+++ b/solvers/minimum_value_constraint.cc
@@ -1,7 +1,6 @@
 #include "drake/solvers/minimum_value_constraint.h"
 
 #include <limits>
-#include <vector>
 
 #include "drake/math/autodiff_gradient.h"
 #include "drake/math/soft_min_max.h"
@@ -10,14 +9,24 @@ namespace drake {
 namespace solvers {
 
 namespace {
+constexpr double kInf = std::numeric_limits<double>::infinity();
+
 /** Computes a smooth over approximation of max(x). */
 template <typename T>
-T SmoothMax(const std::vector<T>& x) {
+T SmoothOverMax(const std::vector<T>& x) {
   // We compute the smooth max of x as smoothmax(x) = log(∑ᵢ exp(αxᵢ)) / α.
   // This smooth max approaches max(x) as α increases. We choose α = 100, as
   // that gives a qualitatively good fit for xᵢ ∈ [0, 1], which is the range of
   // potential penalty values when the MinimumValueConstraint is feasible.
   return math::SoftOverMax(x, 100 /* alpha */);
+}
+
+/** Computes a smooth under approximation of max(x). */
+template <typename T>
+T SmoothUnderMax(const std::vector<T>& x) {
+  // This SoftUnderMax approaches max(x) as α increases. We choose α = 100, as
+  // that gives a qualitatively good fit for xᵢ around 1.
+  return math::SoftUnderMax(x, 100 /* alpha */);
 }
 
 template <typename T>
@@ -26,18 +35,19 @@ T ScaleValue(T value, double minimum_value, double influence_value) {
 }
 
 void InitializeY(const Eigen::Ref<const Eigen::VectorXd>&, Eigen::VectorXd* y,
-                 double y_value) {
-  (*y)(0) = y_value;
+                 int y_index, double y_value) {
+  (*y)(y_index) = y_value;
 }
 
 void InitializeY(const Eigen::Ref<const AutoDiffVecXd>& x, AutoDiffVecXd* y,
-                 double y_value) {
-  (*y) = math::InitializeAutoDiff(
-      Vector1d(y_value), Eigen::RowVectorXd::Zero(x(0).derivatives().size()));
+                 int y_index, double y_value) {
+  (*y)(y_index).value() = y_value;
+  (*y)(y_index).derivatives() =
+      Eigen::RowVectorXd::Zero(x(0).derivatives().size());
 }
 
 void Penalty(const double& value, double minimum_value, double influence_value,
-             MinimumValuePenaltyFunction penalty_function, double* y) {
+             const MinimumValuePenaltyFunction& penalty_function, double* y) {
   double penalty;
   const double x = ScaleValue(value, minimum_value, influence_value);
   penalty_function(x, &penalty, nullptr);
@@ -46,9 +56,11 @@ void Penalty(const double& value, double minimum_value, double influence_value,
 
 void Penalty(const AutoDiffXd& value, double minimum_value,
              double influence_value,
-             MinimumValuePenaltyFunction penalty_function, AutoDiffXd* y) {
+             const MinimumValuePenaltyFunction& penalty_function,
+             AutoDiffXd* y) {
   const AutoDiffXd scaled_value_autodiff =
       ScaleValue(value, minimum_value, influence_value);
+
   double penalty, dpenalty_dscaled_value;
   penalty_function(scaled_value_autodiff.value(), &penalty,
                    &dpenalty_dscaled_value);
@@ -108,26 +120,89 @@ MinimumValueConstraint::MinimumValueConstraint(
     std::function<VectorX<double>(const Eigen::Ref<const VectorX<double>>&,
                                   double)>
         value_function_double)
-    : solvers::Constraint(1, num_vars,
-                          Vector1d(-std::numeric_limits<double>::infinity()),
-                          Vector1d(1)),
-      value_function_{value_function},
-      value_function_double_{value_function_double},
-      minimum_value_{minimum_value},
-      influence_value_{minimum_value + influence_value_offset},
+    : MinimumValueConstraint(num_vars, minimum_value, kInf,
+                             minimum_value + influence_value_offset,
+                             max_num_values, std::move(value_function),
+                             std::move(value_function_double)) {}
+
+namespace {
+int NumConstraints(double minimum_value_lower, double minimum_value_upper) {
+  if (std::isfinite(minimum_value_lower) ||
+      std::isfinite(minimum_value_upper)) {
+    return static_cast<int>(std::isfinite(minimum_value_lower)) +
+           static_cast<int>(std::isfinite(minimum_value_upper));
+  } else {
+    throw std::runtime_error(fmt::format(
+        "MinimumValueConstraint: minimum_value_lower={}, "
+        "minimum_value_upper={}. At least one of them should be finite.",
+        minimum_value_lower, minimum_value_upper));
+  }
+}
+
+Eigen::VectorXd LowerBounds(double minimum_value_lower,
+                            double minimum_value_upper) {
+  // We assume at least one of minimum_value_lower and minimum_value_upper is
+  // finite, as checked already in NumConstraints().
+  if (std::isfinite(minimum_value_lower) &&
+      std::isfinite(minimum_value_upper)) {
+    return Eigen::Vector2d(-kInf, 1);
+  } else if (std::isfinite(minimum_value_lower)) {
+    return Vector1d(-kInf);
+  } else if (std::isfinite(minimum_value_upper)) {
+    return Vector1d(1);
+  }
+  DRAKE_UNREACHABLE();
+}
+
+Eigen::VectorXd UpperBounds(double minimum_value_lower,
+                            double minimum_value_upper) {
+  // We assume at least one of minimum_value_lower and minimum_value_upper is
+  // finite, as checked already in NumConstraints().
+  if (std::isfinite(minimum_value_lower) &&
+      std::isfinite(minimum_value_upper)) {
+    return Eigen::Vector2d(1, kInf);
+  } else if (std::isfinite(minimum_value_lower)) {
+    return Vector1d(1);
+  } else if (std::isfinite(minimum_value_upper)) {
+    return Vector1d(kInf);
+  }
+  DRAKE_UNREACHABLE();
+}
+}  // namespace
+
+MinimumValueConstraint::MinimumValueConstraint(
+    int num_vars, double minimum_value_lower, double minimum_value_upper,
+    double influence_value, int max_num_values,
+    std::function<AutoDiffVecXd(const Eigen::Ref<const AutoDiffVecXd>&, double)>
+        value_function,
+    std::function<VectorX<double>(const Eigen::Ref<const VectorX<double>>&,
+                                  double)>
+        value_function_double)
+    : Constraint(NumConstraints(minimum_value_lower, minimum_value_upper),
+                 num_vars,
+                 LowerBounds(minimum_value_lower, minimum_value_upper),
+                 UpperBounds(minimum_value_lower, minimum_value_upper)),
+      value_function_{std::move(value_function)},
+      value_function_double_{std::move(value_function_double)},
+      minimum_value_lower_{minimum_value_lower},
+      minimum_value_upper_{minimum_value_upper},
+      influence_value_{influence_value},
       max_num_values_{max_num_values} {
-  DRAKE_DEMAND(influence_value_offset > 0);
-  DRAKE_DEMAND(std::isfinite(influence_value_offset));
+  DRAKE_DEMAND(!std::isnan(minimum_value_lower_));
+  DRAKE_DEMAND(!std::isnan(minimum_value_upper_));
+  DRAKE_DEMAND(influence_value_ > minimum_value_lower_);
+  if (std::isfinite(minimum_value_upper)) {
+    DRAKE_DEMAND(influence_value_ > minimum_value_upper_);
+  }
+  DRAKE_DEMAND(std::isfinite(influence_value_));
   set_penalty_function(QuadraticallySmoothedHingeLoss);
 }
 
 void MinimumValueConstraint::set_penalty_function(
     MinimumValuePenaltyFunction new_penalty_function) {
-  penalty_function_ = new_penalty_function;
+  penalty_function_ = std::move(new_penalty_function);
   double unscaled_penalty_at_minimum_value{};
-  penalty_function_(
-      ScaleValue(minimum_value_, minimum_value_, influence_value_),
-      &unscaled_penalty_at_minimum_value, nullptr);
+  penalty_function_(-1, &unscaled_penalty_at_minimum_value, nullptr);
   penalty_output_scaling_ = 1 / unscaled_penalty_at_minimum_value;
 }
 
@@ -146,40 +221,87 @@ AutoDiffVecXd MinimumValueConstraint::Values(
 }
 
 template <typename T>
+void MinimumValueConstraint::EvalMaxPenalty(
+    double bound_value, const VectorX<T>& values, int y_index,
+    const std::function<T(const std::vector<T>&)>& smooth_func,
+    VectorX<T>* y) const {
+  if (y_index >= 0) {
+    const int num_values = values.size();
+    std::vector<T> penalties;
+    penalties.reserve(max_num_values_);
+    for (int i = 0; i < num_values; ++i) {
+      const T& value = values(i);
+      if (value < influence_value_) {
+        penalties.emplace_back();
+        Penalty(value, bound_value, influence_value_, penalty_function_,
+                &(penalties.back()));
+        penalties.back() *= penalty_output_scaling_;
+      }
+    }
+
+    if (!penalties.empty()) {
+      // Pad penalties up to max_num_values_ so that the constraint
+      // function is actually smooth.
+      penalties.resize(max_num_values_, T{0.0});
+      (*y)(y_index) = smooth_func(penalties);
+    }
+  }
+}
+
+template <typename T>
 void MinimumValueConstraint::DoEvalGeneric(
     const Eigen::Ref<const VectorX<T>>& x, VectorX<T>* y) const {
-  y->resize(1);
+  y->resize(num_constraints());
+
+  // Indices for the one or two possible constraints. Infinite bound valuess map
+  // to invalid index values (-1).
+  // SoftOverMax(φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_lower)) / φ(-1)) ≤ 1.
+  const int y_lower_index = std::isfinite(minimum_value_lower_) ? 0 : -1;
+  // Index for the constraint
+  // SoftUnderMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_upper)) / φ(-1) ) ≥
+  // 1.
+  const int y_upper_index =
+      std::isfinite(minimum_value_upper_) ? y_lower_index + 1 : -1;
 
   // If we know that Values() will return at most zero values, then this
-  // is a non-constraint. Return zero in that case.
+  // is a non-constraint. Return zero for the lower bound constraint, and 2 for
+  // the upper bound constraint.
   if (max_num_values_ == 0) {
-    InitializeY(x, y, 0.0);
+    if (y_lower_index >= 0) {
+      InitializeY(x, y, y_lower_index, 0.0);
+    }
+    if (y_upper_index >= 0) {
+      InitializeY(x, y, y_upper_index, 2.0);
+    }
     return;
   }
 
-  // Initialize y to SmoothMax([0, 0, ..., 0]).
-  InitializeY(x, y, SmoothMax(std::vector<double>(max_num_values_, 0.0)));
+  // Initialize y(y_lower_index) and y(y_upper_index) to
+  // SmoothOverMax([0,..., 0]) and SmoothUnderMax([0, ..., 0]), respectively.
+  if (y_lower_index >= 0) {
+    InitializeY(x, y, y_lower_index,
+                SmoothOverMax(std::vector<double>(max_num_values_, 0.0)));
+  }
+  if (y_upper_index >= 0) {
+    InitializeY(x, y, y_upper_index,
+                SmoothUnderMax(std::vector<double>(max_num_values_, 0.0)));
+  }
 
   VectorX<T> values = Values(x);
-  std::vector<T> penalties{};
   const int num_values = static_cast<int>(values.size());
   DRAKE_ASSERT(num_values <= max_num_values_);
-  penalties.reserve(max_num_values_);
-  for (int i = 0; i < num_values; ++i) {
-    const T& value = values(i);
-    if (value < influence_value_) {
-      penalties.emplace_back();
-      Penalty(value, minimum_value_, influence_value_, penalty_function_,
-              &penalties.back());
-      penalties.back() *= penalty_output_scaling_;
-    }
-  }
-  if (!penalties.empty()) {
-    // Pad penalties up to max_num_values_ so that the constraint
-    // function is actually smooth.
-    penalties.resize(max_num_values_, T{0.0});
-    (*y)(0) = SmoothMax(penalties);
-  }
+  this->EvalMaxPenalty<T>(
+      minimum_value_lower_, values, y_lower_index,
+      [](const std::vector<T>& v) {
+        return SmoothOverMax<T>(v);
+      },
+      y);
+  this->EvalMaxPenalty<T>(
+      minimum_value_upper_, values, y_upper_index,
+      [](const std::vector<T>& v) {
+        return SmoothUnderMax<T>(v);
+      },
+      y);
 }
 
 void MinimumValueConstraint::DoEval(const Eigen::Ref<const Eigen::VectorXd>& x,
--- a/solvers/minimum_value_constraint.h
+++ b/solvers/minimum_value_constraint.h
@@ -1,7 +1,10 @@
 #pragma once
 
 #include <functional>
+#include <utility>
+#include <vector>
 
+#include "drake/common/drake_deprecated.h"
 #include "drake/solvers/constraint.h"
 
 namespace drake {
@@ -47,21 +50,29 @@ multidisciplinary workshop on Advances in Preference Handling. */
 void QuadraticallySmoothedHingeLoss(double x, double* penalty,
                                     double* dpenalty_dx);
 
-/** Constrain all elements of the vector returned by the user-provided function
-to be no smaller than a specified minimum value.
+/** Constrain lb <= min(v) <= ub. Namely all elements of the vector `v` returned
+by the user-provided function to be no smaller than a specified minimum value
+`lb`, and the minimal element of the vector is no larger than another specified
+value `ub`.
 
 The formulation of the constraint is
 
-SmoothMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ)) / φ(-1) ) ≤ 1
+<pre>
+SmoothOverMax( φ((vᵢ - v_influence)/(v_influence - lb)) / φ(-1) ) ≤ 1
+SmoothUnderMax( φ((vᵢ - v_influence)/(v_influence - ub)) / φ(-1) ) ≥ 1
+</pre>
 
-where vᵢ is the i-th value returned by the user-provided function, vₘᵢₙ is the
-minimum allowable value, v_influence is the "influence value" (the value below
-which an element influences the constraint or, conversely, the value above which
-an element is ignored), φ is a solvers::MinimumValuePenaltyFunction, and
-SmoothMax(v) is a smooth, conservative approximation of max(v) (i.e.
-SmoothMax(v) >= max(v), for all v). We require that vₘᵢₙ < v_influence. The
-input scaling (vᵢ - v_influence)/(v_influence - vₘᵢₙ) ensures that at the
-boundary of the feasible set (when vᵢ == vₘᵢₙ), we evaluate the penalty function
+where vᵢ is the i-th value returned by the user-provided function, `lb` is
+the minimum allowable value, `ub` is the upper bound for the min(v). (Note that
+`ub` is NOT the upper bound of `v`). v_influence is the "influence value" (the
+value below which an element influences the constraint or, conversely, the value
+above which an element is ignored), φ is a solvers::MinimumValuePenaltyFunction,
+and SmoothOverMax(v) is a smooth, over approximation of max(v) (i.e.
+SmoothOverMax(v) >= max(v), for all v). SmoothUnderMax(x) is a smooth, under
+approximation of max(v) (i.e. SmoothUnderMax(v) <= max(v) for all v). We require
+that lb < ub < v_influence unless ub is infinity, in which case ub is ignored.
+The input scaling (vᵢ - v_influence)/(v_influence - lb) ensures that at the
+boundary of the feasible set (when vᵢ == lb), we evaluate the penalty function
 at -1, where it is required to have a non-zero gradient. The user-provided
 function may return a vector with up to `max_num_values` elements. If it returns
 a vector with fewer than `max_num_values` elements, the remaining elements are
@@ -69,17 +80,20 @@ assumed to be greater than the "influence value". */
 class MinimumValueConstraint final : public solvers::Constraint {
  public:
   /** Constructs a MinimumValueConstraint.
+  min(v) >= lb
+  And we set ub to infinity in min(v) <= ub.
+
   @param num_vars The number of inputs to `value_function`
-  @param minimum_value The minimum allowed value, vₘᵢₙ, for all elements of the
-  vector returned by `value_function`.
+  @param minimum_value The minimum allowed value, lb, for all elements
+  of the vector returned by `value_function`.
   @param influence_value_offset The difference between the
-  influence value, v_influence, and the minimum value, vₘᵢₙ (see class
+  influence value, v_influence, and the minimum value, lb (see class
   documentation). This value must be finite and strictly positive, as it is used
   to scale the values returned by `value_function`. Smaller values may
   decrease the amount of computation required for each constraint evaluation if
   `value_function` can quickly determine that some elements will be
   larger than the influence value and skip the computation associated with those
-  elements. @default 1
+  elements.
   @param max_num_values The maximum number of elements in the vector returned by
   `value_function`.
   @param value_function User-provided function that takes a `num_vars`-element
@@ -106,10 +120,46 @@ class MinimumValueConstraint final : public solvers::Constraint {
                                     double)>
           value_function_double = {});
 
+  /**
+   Overloaded constructor.
+   This constructor allows to specify both the lower bound and upper bound of
+   the minimal element.
+   If you want to ignore the lower bound, then set `minimum_value_lower` to
+   -inf. Similary you can set `minimum_value_upper` to inf to ignore the upper
+   bound. But don't set both minimum_value_lower and minimum_value_upper to
+   infinity.
+   @param minimum_value_lower The minimum allowed value `lb`, for all
+   elements of the vector returned by `value_function`.
+   @param minimum_value_upper The upper bound on min(v), where `v` is the vector
+   returned by `value_function`.
+   @param influence_value This value must be finite and larger than both
+   `minimum_value_lower` and `minimum_value_upper`, as it is used to scale the
+   values returned by `value_function`. Smaller values may decrease the amount
+   of computation required for each constraint evaluation if `value_function`
+   can quickly determine that some elements will be larger than the influence
+   value and skip the computation associated with those elements.
+   */
+  MinimumValueConstraint(
+      int num_vars, double minimum_value_lower, double minimum_value_upper,
+      double influence_value, int max_num_values,
+      std::function<AutoDiffVecXd(const Eigen::Ref<const AutoDiffVecXd>&,
+                                  double)>
+          value_function,
+      std::function<VectorX<double>(const Eigen::Ref<const VectorX<double>>&,
+                                    double)>
+          value_function_double = {});
+
   ~MinimumValueConstraint() override {}
 
   /** Getter for the minimum value. */
-  double minimum_value() const { return minimum_value_; }
+  DRAKE_DEPRECATED("2023-11-01", "Use minimum_value_lower() instead.")
+  double minimum_value() const { return minimum_value_lower_; }
+
+  /** Getter for the lower bound on the minimum value. */
+  double minimum_value_lower() const { return minimum_value_lower_; }
+
+  /** Getter for the upper bound on the minimum value. */
+  double minimum_value_upper() const { return minimum_value_upper_; }
 
   /** Getter for the influence value. */
   double influence_value() const { return influence_value_; }
@@ -123,6 +173,14 @@ class MinimumValueConstraint final : public solvers::Constraint {
  private:
   template <typename T>
   VectorX<T> Values(const Eigen::Ref<const VectorX<T>>& x) const;
+
+  // Given v, evaluate
+  // smooth_fun( φ((vᵢ - v_influence)/(v_influence - bound_value)) / φ(-1) )
+  template <typename T>
+  void EvalMaxPenalty(
+      double bound_value, const VectorX<T>& values, int y_index,
+      const std::function<T(const std::vector<T>&)>& smooth_func,
+      VectorX<T>* y) const;
 
   template <typename T>
   void DoEvalGeneric(const Eigen::Ref<const VectorX<T>>& x,
@@ -144,13 +202,14 @@ class MinimumValueConstraint final : public solvers::Constraint {
   std::function<AutoDiffVecXd(const AutoDiffVecXd&, double)> value_function_;
   std::function<VectorX<double>(const VectorX<double>&, double)>
       value_function_double_;
-  const double minimum_value_;
+  const double minimum_value_lower_;
+  const double minimum_value_upper_;
   const double influence_value_;
   /** Stores the value of
   1 / φ((vₘᵢₙ - v_influence)/(v_influence - vₘᵢₙ)) = 1 / φ(-1). This is
   used to scale the output of the penalty function to be 1 when v == vₘᵢₙ. */
   double penalty_output_scaling_;
-  int max_num_values_{};
+  const int max_num_values_{};
   MinimumValuePenaltyFunction penalty_function_{};
 };
 
--- a/solvers/test/minimum_value_constraint_test.cc
+++ b/solvers/test/minimum_value_constraint_test.cc
@@ -1,14 +1,19 @@
 #include "drake/solvers/minimum_value_constraint.h"
 
+#include <limits>
+
 #include <gtest/gtest.h>
 
 #include "drake/common/symbolic/expression.h"
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
 #include "drake/math/autodiff_gradient.h"
 #include "drake/solvers/test_utilities/check_constraint_eval_nonsymbolic.h"
 
 namespace drake {
 namespace solvers {
 namespace {
+
+const double kInf = std::numeric_limits<double>::infinity();
 
 double kEps{std::numeric_limits<double>::epsilon()};
 
@@ -43,7 +48,8 @@ VectorX<T> ReturnNoValues(const Eigen::Ref<const VectorX<T>>&, double) {
 }
 
 // Verify that the constructor works as expected.
-GTEST_TEST(MinimumValueConstraintTests, ConstructorTest) {
+GTEST_TEST(MinimumValueConstraintTests, ConstructorTest1) {
+  // Constructor with only minimum_value_lower
   int expected_num_vars{5};
   int expected_max_num_values{3};
   double expected_minimum_value{0.1};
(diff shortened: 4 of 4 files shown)


hongkai-dai (commented):
+@calderpg-tri for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NYafnU5DutJRgPy2c_m:b-g27em7)* status: LGTM missing from assignee calderpg-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

RussTedrake (commented):
-@calderpg-tri for feature review, +@RussTedrake for feature review.  I'm keen to use this!
:lgtm: with a few comments and requests below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NZCauNC3Bh1Vtcn55ny:b10qup2)* status: 7 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`solvers/minimum_value_constraint.h` line 52 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCc8uQ5uqCdCeYA94t:-NZCc8uQ5uqCdCeYA94u:bi6lk0x) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L52)):*
> ```objc
> /** Constrain all elements of the vector returned by the user-provided function
> to be no smaller than a specified minimum value, the minimal element of the
> vector is no larger than another specified value.
> ```

nit: I think it would be much clearer to say that the minimum value of the vector is between lower_bound and upper_bound.  E.g. lb <= min(v) <= ub.  It took me three readings to make sure that's actually what these words said.
___
*[`solvers/minimum_value_constraint.h` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCcc7-6_iOCiny3Thj:-NZCcc70-9N8uq1texms:bn30nkp) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L59)):*
> ```objc
> SmoothOverMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_lower)) / φ(-1) ) ≤ 1
> SmoothUnderMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_upper)) / φ(-1) ) ≥ 1
> </pre>
> ```

nit: v_min_lower and upper could just be lower and upper?  or lb and ub?
___
*[`solvers/minimum_value_constraint.h` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCco7c5JXrf5Y6enVu:-NZCco7c5JXrf5Y6enVv:b6sv05c) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L62)):*
> ```objc
> 
> where vᵢ is the i-th value returned by the user-provided function, vₘᵢₙ_lower is
> the minimum allowable value, vₘᵢₙ_upper is the upper bound for the min(v).
> ```

nit: "the minimum allowable value" => is the lower bound.  (or just use lb/ub above)
___
*[`solvers/minimum_value_constraint.h` line 70 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCdv7H8GT9nDpwfVHU:-NZCdv7H8GT9nDpwfVHV:b1mglp) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L70)):*
> ```objc
> SmoothUnderMax(v) <= max(v) for all v). We require that vₘᵢₙ_lower < v_ₘᵢₙ_upper
> < v_influence. The input scaling (vᵢ - v_influence)/(v_influence - vₘᵢₙ_lower)
> ensures that at the boundary of the feasible set (when vᵢ == vₘᵢₙ), we evaluate
> ```

you either need to update this v_min to v_min lower, or change the line above (since this scaling idea is true for both lower and upper).
___
*[`solvers/minimum_value_constraint.h` line 77 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCeb9T6kO8oaq7HXCT:-NZCeb9T6kO8oaq7HXCU:b35pqz) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L77)):*
> ```objc
> class MinimumValueConstraint final : public solvers::Constraint {
>  public:
>   /** Constructs a MinimumValueConstraint.
> ```

this docstring should somewhere say what happens to v_min_upper, which is not specified.  presumably it's set to the influence distance?
___
*[`solvers/minimum_value_constraint.h` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCeT-LBzlGHfhmiTG6:-NZCeT-LBzlGHfhmiTG7:b-tl6555) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L88)):*
> ```objc
>   `value_function` can quickly determine that some elements will be
>   larger than the influence value and skip the computation associated with those
>   elements. @default 1
> ```

there is no default value?
___
*[`solvers/minimum_value_constraint.h` line 122 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCet3E16aqetuquJVa:-NZCet3E16aqetuquJVb:b-yjyp6y) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L122)):*
> ```objc
>    -inf. Similary you can set `minimum_value_upper` to inf to ignore the upper
>    bound. But don't set both minimum_value_lower and minimum_value_upper to
>    infinity.
> ```

btw: i guess that's true of every constraint?  but I guess it's useful to say explicitly that it is supported.
___
*[`solvers/minimum_value_constraint.h` line 147 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCfIatCgbT9fGTxM6T:-NZCfIatCgbT9fGTxM6U:b-m2c1au) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L147)):*
> ```objc
> 
>   /** Getter for the minimum value. */
>   double minimum_value() const { return minimum_value_lower_; }
> ```

i would think you should add a new method for minimum_value_lower().  even if you don't deprecate this (I'm fine with that), the docstring should be updated to say lower.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 4 unresolved discussions, needs at least two assigned reviewers, missing label for release notes

___
*[`solvers/minimum_value_constraint.h` line 52 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCc8uQ5uqCdCeYA94t:-NZHEIK2412vtUHiP1MH:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L52)):*

nit: I think it would be much clearer to say that the minimum value of the vector is between lower_bound and upper_bound.  E.g. lb <= min(v) <= ub.  It took me three readings to make sure that's actually what these words said.
</blockquote></details>

Done. Good call.
___
*[`solvers/minimum_value_constraint.h` line 70 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCdv7H8GT9nDpwfVHU:-NZHLyNpBQKyBe2C1fRS:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L70)):*

you either need to update this v_min to v_min lower, or change the line above (since this scaling idea is true for both lower and upper).
</blockquote></details>

Done.
___
*[`solvers/minimum_value_constraint.h` line 77 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCeb9T6kO8oaq7HXCT:-NZHM3dk7BufyA1Svg_H:blli50p) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L77)):*

this docstring should somewhere say what happens to v_min_upper, which is not specified.  presumably it's set to the influence distance?
</blockquote></details>

Done. I added some documentation. It's actually set to infinity and we ignore the constraint on `v_min_upper`.
___
*[`solvers/minimum_value_constraint.h` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCeT-LBzlGHfhmiTG6:-NZHMZ21E3YlDqNVTKEr:b-yh16l6) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L88)):*

there is no default value?
</blockquote></details>

Done. Good call.
___
*[`solvers/minimum_value_constraint.h` line 122 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCet3E16aqetuquJVa:-NZHMhHoAYmDAgDmm5pE:b-d02eov) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L122)):*

btw: i guess that's true of every constraint?  but I guess it's useful to say explicitly that it is supported.
</blockquote></details>

Currently in our implementation, if both `minimum_value_lower` and `minimum_value_upper` are set to 0, then this constraint has 0 entries (MinimumValueConstraint.num_constraints() == 0). This is slightly different from other cases (like setting the lower and upper bound of the LinearConstraint to infinity, which still has non-zero `num_constraints()`).
___
*[`solvers/minimum_value_constraint.h` line 147 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZCfIatCgbT9fGTxM6T:-NZHOH84AjNq-rJPOIIc:b63eylk) ([raw file](https://github.com/RobotLocomotion/drake/blob/748139945e7907bcacfec8e4dbee1d117bf7bbc0/solvers/minimum_value_constraint.h#L147)):*

i would think you should add a new method for minimum_value_lower().  even if you don't deprecate this (I'm fine with that), the docstring should be updated to say lower.
</blockquote></details>

Done. I deprecated the old method.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 2 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`solvers/minimum_value_constraint.h` line 71 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZHvkYN60b5Wa5zoXM1:-NZHvkYN60b5Wa5zoXM2:b-71d60s) ([raw file](https://github.com/RobotLocomotion/drake/blob/519cf800d4ca735da38c1f2906375c23a84376aa/solvers/minimum_value_constraint.h#L71)):*
> ```objc
> SmoothOverMax(v) >= max(v), for all v). SmoothUnderMax(x) is a smooth, under
> approximation of max(v) (i.e. SmoothUnderMax(v) <= max(v) for all v). We require
> that lb < ub < v_influence. The input scaling (vᵢ -
> ```

"We require that lb < ub < v_influence"... unless ub is infinity, in which case ub is ignored.
___
*[`solvers/minimum_value_constraint.h` line 83 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZHvVMJ1pnYSWtaog5B:-NZHvVMJ1pnYSWtaog5C:bb2hqbk) ([raw file](https://github.com/RobotLocomotion/drake/blob/519cf800d4ca735da38c1f2906375c23a84376aa/solvers/minimum_value_constraint.h#L83)):*
> ```objc
>   min(v) >= lb
>   Note that we don't enforce the constraint min(v) <= ub in the class
>   documentation.
> ```

this wording is a little strange.  can you just say that ub is set to infinity?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 2 unresolved discussions, needs at least two assigned reviewers, missing label for release notes

___
*[`solvers/minimum_value_constraint.h` line 71 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZHvkYN60b5Wa5zoXM1:-NZIWahrCuEGIoYLltWe:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/519cf800d4ca735da38c1f2906375c23a84376aa/solvers/minimum_value_constraint.h#L71)):*

"We require that lb < ub < v_influence"... unless ub is infinity, in which case ub is ignored.
</blockquote></details>

Done.
___
*[`solvers/minimum_value_constraint.h` line 83 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZHvVMJ1pnYSWtaog5B:-NZIWkTj5I4SNqSAf-dx:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/519cf800d4ca735da38c1f2906375c23a84376aa/solvers/minimum_value_constraint.h#L83)):*

this wording is a little strange.  can you just say that ub is set to infinity?
</blockquote></details>

Done.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NZNYo_m8-fxrW9wqPDB:b5r1yx7)* status: LGTM missing from assignee SeanCurtis-TRI(platform), missing label for release notes (waiting on @hongkai-dai)

SeanCurtis-TRI (commented):
+(release notes: feature) 
+(release notes: newly deprecated) 

First pass done. Mostly just some small stuff. One issue about handling the case where `lb` is finite and `ub` is `NaN`.

> Reviewed 1 of 3 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NZO7rwL5yTK7dDoYBxh:b-v5faze)* status: 15 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.h` line 159 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZO9sZQ0sMuiovBEO_h:-NZO9sZR1pAYNCbecAdM:b-c3d4eo) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.h#L159)):*
> ```objc
>   double minimum_value_lower() const { return minimum_value_lower_; }
> 
>   double minimum_value_upper() const { return minimum_value_upper_; }
> ```

nit: This is the only API that lacks documentation.
___
*[`solvers/minimum_value_constraint.cc` line 11 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOA52-DCTFIiKxv4aU:-NZOA52-DCTFIiKxv4aV:b-edsxll) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L11)):*
> ```cpp
> namespace drake {
> namespace solvers {
> const double kInf = std::numeric_limits<double>::infinity();
> ```

BTW You can make this `constexpr`.
___
*[`solvers/minimum_value_constraint.cc` line 24 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOAP0TAXjWGhOWcN2r:-NZOAP0TAXjWGhOWcN2s:b4z5rla) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L24)):*
> ```cpp
> }
> 
> /** Computes a smooth over approximation of max(x). */
> ```

nit Copypasta



_Suggestion:_
```cpp
under
```
___
*[`solvers/minimum_value_constraint.cc` line 124 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOBITG07_d5ja24m56:-NZOBITG07_d5ja24m57:bhmvqqi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L124)):*
> ```cpp
>     : MinimumValueConstraint(
>           num_vars, minimum_value, kInf, minimum_value + influence_value_offset,
>           max_num_values, value_function, value_function_double) {}
> ```

nit Because you're passing `value_function` and `value_function_double` by value, you should go ahead and `std::move` them when you pass them to the delegated constructor. Depending on the source of the `std::function`, it's probably not worth paying the copy cost.
___
*[`solvers/minimum_value_constraint.cc` line 143 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOC-WU48CT5Hsva-2P:-NZOC-WU48CT5Hsva-2Q:b-hl1od6) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L143)):*
> ```cpp
>   } else {
>         "MinimumValueConstraint: at least one of minimum_value_lower and "
> ```

nit: Here (and for `UpperBounds()`). As long as you're throwing, you should print out the values you were given that tripped the exception.

A further thought: you could remove the test from these two functions and into `NumConstraints()`. As that is always called before these other two functions, it'll catch the error condition at the earliest point. And then you'd only have to maintain one test.

I *would* put a comment on `LowerBounds()` and `UpperBounds()` that they have the precondition that at least one of the two values is finite. And `NumConstraints()` serves as the prerequisite validator.
___
*[`solvers/minimum_value_constraint.cc` line 177 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOCD-N1GQcoEnOcL35:-NZOCD-N1GQcoEnOcL36:b-3ppl0w) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L177)):*
> ```cpp
>                  LowerBounds(minimum_value_lower, minimum_value_upper),
>                  UpperBounds(minimum_value_lower, minimum_value_upper)),
>       value_function_{value_function},
> ```

btw: Continuing the theme, you should move these functions into the members. I  notice that there are a number of APIs that take functions by value. Generally, they'd all be better as moves into the final assignment, but general correction is outside the scope of this PR. So, feel free to defer/ignore.
___
*[`solvers/minimum_value_constraint.cc` line 218 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOF7YA9Ko1Ao9PORYP:-NZOF7YA9Ko1Ao9PORYQ:b-sxpsnr) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L218)):*
> ```cpp
>   y->resize(num_constraints());
> 
>   // Index for the constraint
> ```

nit: Some thoughts.

1. GSG: you're missing a terminal period.
2. There are actually "indices" (and not a single index).
3. It's probably worth noting here the -1 convention.
    - And then you can remove the somewhat inscrutable "Use -1 if value is inf` statements associated with the constraints.



_Suggestion:_
```cpp
// Indices for the one or two possible constraints. Infinite bound valuess map
// to invalid index values (-1).
```
___
*[`solvers/minimum_value_constraint.cc` line 219 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOExucEzU8gDpR0Q3y:-NZOExucEzU8gDpR0Q3z:b-hoytn9) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L219)):*
> ```cpp
> 
>   // Index for the constraint
>   // SoftOverMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_lower)) / φ(-1) ) ≤ 1
> ```

nit. Missing a period at the end of `≤ 1`.
___
*[`solvers/minimum_value_constraint.cc` line 221 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOE3mx-aNByfSxmzov:-NZOE3mx-aNByfSxmzow:b1uyzuc) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L221)):*
> ```cpp
>   // SoftOverMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_lower)) / φ(-1) ) ≤ 1
>   // Use -1 if minimum_value_lower_ is inf.
>   const int y_lower_index = std::isfinite(minimum_value_lower_) ? 0 : -1;
> ```

As I read just this new code, it seems to me that if I pass `NaN` as one of the values (and only one), it is *treated* as `inf`. Is that right? We can't rely on `Constraint`'s NaN detection, because we start treating `NaN` as `inf` in `LowerBounds` and `UpperBounds` (i.e., those functions guarantee that we don't pass `NaN` to `Constraint`).

Should initializing a value with `NaN` not cause an immediate problem?

I was able to confirm that I was able to happily construct a constraint with a finite lower bound and a `NaN` upper bound. (`NaN` for lower bound eventually fails in confirming that it is less than `influence_value_`.
___
*[`solvers/minimum_value_constraint.cc` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOF1iyCEGqCYHKNvCx:-NZOF1iyCEGqCYHKNvCy:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L224)):*
> ```cpp
>   // Index for the constraint
>   // SoftUnderMax( φ((vᵢ - v_influence)/(v_influence - vₘᵢₙ_upper)) / φ(-1) ) ≥
>   // 1 Use -1 if minimum_value_upper_ is inf.
> ```

nit: typo



_Suggestion:_
```cpp
// 1. Use -1 if 
```
___
*[`solvers/minimum_value_constraint.cc` line 242 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOGNN113XvYznH5jV_:-NZOGNN113XvYznH5jVa:b-ixtywa) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L242)):*
> ```cpp
> 
>   // Initialize y(y_lower_index) and y(y_upper_index) to
>   // SmoothOverMax([0, 0, ..., 0]).
> ```

nit: Incomplete.



_Suggestion:_
```cpp
// SmoothOverMax([0, ..., 0]) and SmoothUnderMax([0, ..., 0]), respectively.
```
___
*[`solvers/minimum_value_constraint.cc` line 253 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOJSwQ6LNhfLRxkn-0:-NZOJSwQ6LNhfLRxkn-1:b-ud4ze4) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L253)):*
> ```cpp
> 
>   VectorX<T> values = Values(x);
>   const int num_values = static_cast<int>(values.size());
> ```

nit: Surely Eigen already returns size as an int, yes?
___
*[`solvers/minimum_value_constraint.cc` line 255 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOKyRj-FgEyrQ6Jiq8:-NZOKyRj-FgEyrQ6Jiq9:b-nb9vlc) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L255)):*
> ```cpp
>   const int num_values = static_cast<int>(values.size());
>   DRAKE_ASSERT(num_values <= max_num_values_);
>   if (y_lower_index >= 0) {
> ```

BTW If you refactor this code slightly, it might be easier to read and maintain. This doesn't trip over the "rule of 3", but the logic between the two bounds is elaborate enough, I thought it might be worth being clear about what they should have in common and what they should have different.


```c++
  ...
  VectorX<T> values = Values(x);
  const int num_values = static_cast<int>(values.size());
  DRAKE_ASSERT(num_values <= max_num_values_);
  if (y_lower_index >= 0) {
    EvalBound(minimum_value_lower_, values, y_lower_index, SmoothOverMax<T>, y);
  }
  if (y_upper_index >= 0) {
    EvalBound(minimum_value_upper_, values, y_upper_index, SmoothUnderMax<T>,
              y);
  }
}

template <typename T>
void MinimumValueConstraint::EvalBound(
    double bound_value, const VectorX<T>& values, int index,
    const std::function<T(const std::vector<T>&)>& smooth_func,
    VectorX<T>* y) const {
  const int num_values = values.size();
  std::vector<T> penalties{};
  penalties.reserve(max_num_values_);
  for (int i = 0; i < num_values; ++i) {
    const T& value = values(i);
    if (value < influence_value_) {
      penalties.emplace_back();
      Penalty(value, bound_value, influence_value_, penalty_function_,
              &penalties.back());
      penalties.back() *= penalty_output_scaling_;
    }
  }
  if (!penalties.empty()) {
    // Pad penalties up to max_num_values_ so that the constraint
    // function is actually smooth.
    penalties.resize(max_num_values_, T{0.0});
    (*y)(index) = smooth_func(penalties);
  }
}
```
___
*[`solvers/minimum_value_constraint.cc` line 257 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOGuQ99cnh7jtPfyo0:-NZOGuQ99cnh7jtPfyo1:bqtaw09) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L257)):*
> ```cpp
>   if (y_lower_index >= 0) {
>     std::vector<T> penalties_lower{};
>     penalties_lower.reserve(max_num_values_);
> ```

BTW As I read this, `max_num_values_` is immutable for the lifespan of this constraint (as such, it'd be slick if it were marked `const` as the constraint is neither movable or copyable).

But, more particularly, it's a shame that every time we evaluate this constraint we've got heap allocations. It would be nice if that didn't happen.

Obviously, outside the scope of this PR, but it jumped out at me while reading this new code. I can imagine that the time cost of heap allocation is negligible to the cost of actually evaluating the constraint, but still.
___
*[`solvers/test/minimum_value_constraint_test.cc` line 64 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOLycd7CDkksdGveYy:-NZOLycd7CDkksdGveYz:bql3z3a) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/test/minimum_value_constraint_test.cc#L64)):*
> ```cpp
>   EXPECT_EQ(dut.max_num_values(), expected_max_num_values);
>   EXPECT_EQ(dut.minimum_value_lower(), expected_minimum_value);
>   EXPECT_FALSE(std::isfinite(dut.minimum_value_upper()));
> ```

NIT Is it enough to confirm it's not finite? Wouldn't it be a better test to confirm that it returns kInf? After all, returning `NaN` would pass the test but not be desired behavior.

Same for the `minimum_value_upper` test below.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.cc` line 11 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOA52-DCTFIiKxv4aU:-NZP-N342h0yifqQPRBi:bhsnkei) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L11)):*

BTW You can make this `constexpr`.
</blockquote></details>

Done.

For my own education, what is the advantage of making it as a `constexpr`? I thought that the major difference is that `constexpr` is can be used as a compile time constant, hence often used in the template instantiation. But I don't see that using `kInf` as a compile-time constant makes much difference?
___
*[`solvers/minimum_value_constraint.cc` line 143 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOC-WU48CT5Hsva-2P:-NZP298x4QOOq-htE2sc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L143)):*

nit: Here (and for `UpperBounds()`). As long as you're throwing, you should print out the values you were given that tripped the exception.

A further thought: you could remove the test from these two functions and into `NumConstraints()`. As that is always called before these other two functions, it'll catch the error condition at the earliest point. And then you'd only have to maintain one test.

I *would* put a comment on `LowerBounds()` and `UpperBounds()` that they have the precondition that at least one of the two values is finite. And `NumConstraints()` serves as the prerequisite validator.
</blockquote></details>

Done.
___
*[`solvers/minimum_value_constraint.cc` line 177 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOCD-N1GQcoEnOcL35:-NZP2E3w7fvGML6auI30:b-h3d9wh) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L177)):*

btw: Continuing the theme, you should move these functions into the members. I  notice that there are a number of APIs that take functions by value. Generally, they'd all be better as moves into the final assignment, but general correction is outside the scope of this PR. So, feel free to defer/ignore.
</blockquote></details>

Done.

For my own education, if the code doesn't do assignment, should I pass the function by value or by const reference? One example is in `Penalty` function
```cc
void Penalty(const double& value, double minimum_value, double influence_value,
             MinimumValuePenaltyFunction penalty_function, double* y) {
  double penalty;
  const double x = ScaleValue(value, minimum_value, influence_value);
  penalty_function(x, &penalty, nullptr);
  *y = penalty;
}

```
___
*[`solvers/minimum_value_constraint.cc` line 218 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOF7YA9Ko1Ao9PORYP:-NZP38gS01SjUFFAh-1d:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L218)):*

nit: Some thoughts.

1. GSG: you're missing a terminal period.
2. There are actually "indices" (and not a single index).
3. It's probably worth noting here the -1 convention.
    - And then you can remove the somewhat inscrutable "Use -1 if value is inf` statements associated with the constraints.
</blockquote></details>

Done.
___
*[`solvers/minimum_value_constraint.cc` line 221 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOE3mx-aNByfSxmzov:-NZP3Q3hAVEixPWwp_D3:b-x1kyr9) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L221)):*

As I read just this new code, it seems to me that if I pass `NaN` as one of the values (and only one), it is *treated* as `inf`. Is that right? We can't rely on `Constraint`'s NaN detection, because we start treating `NaN` as `inf` in `LowerBounds` and `UpperBounds` (i.e., those functions guarantee that we don't pass `NaN` to `Constraint`).

Should initializing a value with `NaN` not cause an immediate problem?

I was able to confirm that I was able to happily construct a constraint with a finite lower bound and a `NaN` upper bound. (`NaN` for lower bound eventually fails in confirming that it is less than `influence_value_`.


</blockquote></details>

I don't want `minimum_value_lower_` or `minimum_value_upper_` to be NaN. I added the check in the class constructor.
___
*[`solvers/minimum_value_constraint.cc` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOF1iyCEGqCYHKNvCx:-NZP3x3i8DFfcHSYupoL:by2sj3z) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L224)):*

nit: typo
</blockquote></details>

Done. I removed `-1` case as we have mentioned it a few lines above.
___
*[`solvers/minimum_value_constraint.cc` line 255 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOKyRj-FgEyrQ6Jiq8:-NZP7hQV3sZ4qTM7PmOx:b-9z0i5u) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L255)):*

BTW If you refactor this code slightly, it might be easier to read and maintain. This doesn't trip over the "rule of 3", but the logic between the two bounds is elaborate enough, I thought it might be worth being clear about what they should have in common and what they should have different.


```c++
  ...
  VectorX<T> values = Values(x);
  const int num_values = static_cast<int>(values.size());
  DRAKE_ASSERT(num_values <= max_num_values_);
  if (y_lower_index >= 0) {
    EvalBound(minimum_value_lower_, values, y_lower_index, SmoothOverMax<T>, y);
  }
  if (y_upper_index >= 0) {
    EvalBound(minimum_value_upper_, values, y_upper_index, SmoothUnderMax<T>,
              y);
  }
}

template <typename T>
void MinimumValueConstraint::EvalBound(
    double bound_value, const VectorX<T>& values, int index,
    const std::function<T(const std::vector<T>&)>& smooth_func,
    VectorX<T>* y) const {
  const int num_values = values.size();
  std::vector<T> penalties{};
  penalties.reserve(max_num_values_);
  for (int i = 0; i < num_values; ++i) {
    const T& value = values(i);
    if (value < influence_value_) {
      penalties.emplace_back();
      Penalty(value, bound_value, influence_value_, penalty_function_,
              &penalties.back());
      penalties.back() *= penalty_output_scaling_;
    }
  }
  if (!penalties.empty()) {
    // Pad penalties up to max_num_values_ so that the constraint
    // function is actually smooth.
    penalties.resize(max_num_values_, T{0.0});
    (*y)(index) = smooth_func(penalties);
  }
}
```
</blockquote></details>

Done. Great idea. I re-factored the code.
___
*[`solvers/minimum_value_constraint.cc` line 257 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOGuQ99cnh7jtPfyo0:-NZP80QlDl6iHqRMQWdx:buw3ho5) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L257)):*

BTW As I read this, `max_num_values_` is immutable for the lifespan of this constraint (as such, it'd be slick if it were marked `const` as the constraint is neither movable or copyable).

But, more particularly, it's a shame that every time we evaluate this constraint we've got heap allocations. It would be nice if that didn't happen.

Obviously, outside the scope of this PR, but it jumped out at me while reading this new code. I can imagine that the time cost of heap allocation is negligible to the cost of actually evaluating the constraint, but still.
</blockquote></details>

I made `max_num_values_` a const.

I also added the member variabe `penalties_double_` and `penalties_autodiff_`, and re-write the factored function EvalMaxPenalty. Could you help check the logic in this function?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 4 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.h` line 200 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZPKBdZ-1aeqmIf6k-u:-NZPKBdZ-1aeqmIf6k-v:b-mxogk1) ([raw file](https://github.com/RobotLocomotion/drake/blob/93b9620907f4770ea2a6d472998cbff1ac41f382/solvers/minimum_value_constraint.h#L200)):*
> ```objc
>   }
> 
>   template <typename T>
> ```

nit: Read the comment below on the newly added vectors first. If it turns out I'm mistaken on *that* comment, then you can come back here and read the rest of this. Otherwise, this comment will simply go away.

-----------------------------------------------------------------------------

This deserves some special attention.

1. It returns mutable reference from a `const` method.
2. At first glance, it's the mechanism by which we get access to the `mutable` vectors declared below.
3. In reality, it doesn't seem that you actually use it.

I'd probably recommend moving the function down with the vector declarations so they all from a common package.

Then you can mention that the function is there so that you can easily access the right type-dependent cache in t `EvalMaxPenalty()`.
___
*[`solvers/minimum_value_constraint.h` line 217 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZPKcYv2bQOQ_19mJYH:-NZPKcYv2bQOQ_19mJYI:b-c1fvq9) ([raw file](https://github.com/RobotLocomotion/drake/blob/93b9620907f4770ea2a6d472998cbff1ac41f382/solvers/minimum_value_constraint.h#L217)):*
> ```objc
>   // std::vectors at the class construction, and modify them in each Eval
>   // function.
>   mutable std::vector<double> penalties_double_;
> ```

This is ambitious. But now that you've done this, I have to ask another question:

Do you have a guarantee that one instance of this constraint will only ever be evaluated in one thread? Generally speaking, do you have thread-safety in `solvers`? Because *this* is not thread-safe. So, unless you have guarantees at a higher place, this is too ambitious.

So, in the absence of thread safety guarantees, you should probably simply shrug and say, "Yep. For now I'm allocating."

I kind of regret mentioning that passing thought. It is regrettable (the repetitive allocation), but it isn't clear there's much you can do right now.
___
*[`solvers/minimum_value_constraint.cc` line 11 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOA52-DCTFIiKxv4aU:-NZPGh1L4wO4dLq2yqz5:bkfwfvu) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L11)):*

Done.

For my own education, what is the advantage of making it as a `constexpr`? I thought that the major difference is that `constexpr` is can be used as a compile time constant, hence often used in the template instantiation. But I don't see that using `kInf` as a compile-time constant makes much difference?
</blockquote></details>

If I were to give a justification it would be to unblock anything downstream. `constexpr` can be used in both compile-time operations and run-time operations. `const` can only be used in run time.  So, by declaring things `constexpr` where you can, you're doing the *most* to empower the compiler.

Now, in this case, it's probable that within the scope of this single compilation unit, the compiler can take your global `const` and use it appropriately.

So, it's not clear that it's a definite win, it *feels* like it's a good habit to get into just so that you don't block the compiler when you *do* get burned.
___
*[`solvers/minimum_value_constraint.cc` line 177 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOCD-N1GQcoEnOcL35:-NZPITY80Di3m73VsAzV:b-ilqvi1) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L177)):*

Done.

For my own education, if the code doesn't do assignment, should I pass the function by value or by const reference? One example is in `Penalty` function
```cc
void Penalty(const double& value, double minimum_value, double influence_value,
             MinimumValuePenaltyFunction penalty_function, double* y) {
  double penalty;
  const double x = ScaleValue(value, minimum_value, influence_value);
  penalty_function(x, &penalty, nullptr);
  *y = penalty;
}

```
</blockquote></details>

Generally, prefer `const&` if you're not taking ownership. The [`operator()` method is declared `const`](https://en.cppreference.com/w/cpp/utility/functional/function/operator()) so you don't impair yourself in executing it by doing so.

This is most relevant if the function passed is a lambda with multiple captures (in my mind, I have a recollection that capturing a single word or so -- like a pointer -- doesn't allocate. But capturing more allocates.)

In contrast, if you're just passing a function pointer, there's no real difference. But by tailoring your parameter type appropriately, the caller doesn't have to worry.
___
*[`solvers/minimum_value_constraint.cc` line 257 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOGuQ99cnh7jtPfyo0:-NZPN2DvBw8-wVfog0lX:b-g1ci0z) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L257)):*

I made `max_num_values_` a const.

I also added the member variabe `penalties_double_` and `penalties_autodiff_`, and re-write the factored function EvalMaxPenalty. Could you help check the logic in this function?
</blockquote></details>

I'm closing this here. The `const`-ness you've added is good. But for the reasons outlined above, I think the cached vectors may have a problem.
___
*[`solvers/minimum_value_constraint.cc` line 11 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZPHip9CU5dgiHyH4ey:-NZPHip9CU5dgiHyH4ez:bsypgbj) ([raw file](https://github.com/RobotLocomotion/drake/blob/93b9620907f4770ea2a6d472998cbff1ac41f382/solvers/minimum_value_constraint.cc#L11)):*
> ```cpp
> namespace drake {
> namespace solvers {
> constexpr double kInf = std::numeric_limits<double>::infinity();
> ```

nit: Oops. Go ahead and move this *inside* the anonymous namespace so it can't conflict with any other compilation units.
___
*[`solvers/test/minimum_value_constraint_test.cc` line 64 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZPNLun6rjs2u4CRVZy:-NZPNLun6rjs2u4CRVZz:b-chmf2l) ([raw file](https://github.com/RobotLocomotion/drake/blob/93b9620907f4770ea2a6d472998cbff1ac41f382/solvers/test/minimum_value_constraint_test.cc#L64)):*
> ```cpp
>   EXPECT_EQ(dut.max_num_values(), expected_max_num_values);
>   EXPECT_EQ(dut.minimum_value_lower(), expected_minimum_value);
>   EXPECT_TRUE(std::isinf(dut.minimum_value_upper()));
> ```

BTW this is fine as is. However, I sort of expected something like:

```c++
EXPECT_EQ(dut.minimum_value_upper(), kInf);
EXPECT_EQ(dut.lower_bound(), -kInf);
```

It fits very nicely in with the other `EXPECT_EQ` you have and distinguishes the sign of the infinity (which is, of course, relevant). And you already have `kInf` defined in this context.  :)

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.h` line 217 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZPKcYv2bQOQ_19mJYH:-NZRXeFl1EDgo-KA8a9z:bdzs9np) ([raw file](https://github.com/RobotLocomotion/drake/blob/93b9620907f4770ea2a6d472998cbff1ac41f382/solvers/minimum_value_constraint.h#L217)):*

This is ambitious. But now that you've done this, I have to ask another question:

Do you have a guarantee that one instance of this constraint will only ever be evaluated in one thread? Generally speaking, do you have thread-safety in `solvers`? Because *this* is not thread-safe. So, unless you have guarantees at a higher place, this is too ambitious.

So, in the absence of thread safety guarantees, you should probably simply shrug and say, "Yep. For now I'm allocating."

I kind of regret mentioning that passing thought. It is regrettable (the repetitive allocation), but it isn't clear there's much you can do right now.
</blockquote></details>

Done. I removed these mutable vectors.
___
*[`solvers/minimum_value_constraint.cc` line 11 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOA52-DCTFIiKxv4aU:-NZRXjsNBlp9bFHeDZ26:bcr5s2o) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L11)):*

If I were to give a justification it would be to unblock anything downstream. `constexpr` can be used in both compile-time operations and run-time operations. `const` can only be used in run time.  So, by declaring things `constexpr` where you can, you're doing the *most* to empower the compiler.

Now, in this case, it's probable that within the scope of this single compilation unit, the compiler can take your global `const` and use it appropriately.

So, it's not clear that it's a definite win, it *feels* like it's a good habit to get into just so that you don't block the compiler when you *do* get burned.
</blockquote></details>

Got it, thanks!
___
*[`solvers/minimum_value_constraint.cc` line 177 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZOCD-N1GQcoEnOcL35:-NZRXBuC3KyFWmknDV3B:bcr5s2o) ([raw file](https://github.com/RobotLocomotion/drake/blob/b7278319fe04ddf77fa05b6e34045b4b5da22050/solvers/minimum_value_constraint.cc#L177)):*

Generally, prefer `const&` if you're not taking ownership. The [`operator()` method is declared `const`](https://en.cppreference.com/w/cpp/utility/functional/function/operator()) so you don't impair yourself in executing it by doing so.

This is most relevant if the function passed is a lambda with multiple captures (in my mind, I have a recollection that capturing a single word or so -- like a pointer -- doesn't allocate. But capturing more allocates.)

In contrast, if you're just passing a function pointer, there's no real difference. But by tailoring your parameter type appropriately, the caller doesn't have to worry.
</blockquote></details>

Got it, thanks!

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform)

SeanCurtis-TRI (commented):
A bit of lint, and one last fun corner case.

> Reviewed 1 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NZT17Y2EBaU6zCegskX:b-qylwmm)* status: 6 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.h` line 4 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZT03OH7lYqBmsV-Jso:-NZT03OH7lYqBmsV-Jsp:b-nc7mbc) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.h#L4)):*
> ```objc
> 
> #include <functional>
> ```

nit: angry lint:

`#include <vector>`
___
*[`solvers/minimum_value_constraint.h` line 201 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZT-zmA3lNmiRVCMX2v:-NZT-zmA3lNmiRVCMX2w:b-5pvl7a) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.h#L201)):*
> ```objc
> 
>   template <typename T>
>   std::vector<T>& penalties() const;
> ```

nit: Not defined. Not used. Left over from the caching efforts.
___
*[`solvers/minimum_value_constraint.cc` line 4 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZT0Ht96zRlkkyrSENd:-NZT0Ht96zRlkkyrSENe:b46v95j) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L4)):*
> ```cpp
> 
> #include <limits>
> #include <vector>
> ```

nit: Angry lint

`#include <utility>
___
*[`solvers/minimum_value_constraint.cc` line 233 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZSroDpB3zC25zjFFIS:-NZSroDpB3zC25zjFFIT:b-6d4a92) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L233)):*
> ```cpp
>     std::vector<T> penalties;
>     penalties.reserve(max_num_values_);
>     // Initialize penalty to 0.
> ```

nit: This comment is now a lie. We don't initialize anything.
___
*[`solvers/minimum_value_constraint.cc` line 248 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZStOssBkWInlsCr5h6:-NZStOssBkWInlsCr5h7:btdl4rj) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L248)):*
> ```cpp
>       // function is actually smooth.
>       penalties.resize(max_num_values_, T{0.0});
>       (*y)(y_index) = smooth_func(penalties);
> ```

Sorry, every time I look at this code, something *else* jumps out at me.

If `value >= influence_value_` *for all values*,  we add *no* entries to `penalties` and, therefore, we do not write to `(*y)(y_index)`. The output does not get changed. Is this right?

I looked at the `EvaluatorBase` documentation and it makes no statements one way or the other about whether `y` is guaranteed to be written to. Without warning that `y` *might* not get written to, it seems the default would be that `y` should always get written to (that's my own interpretative bias).

Recognizing the fact that `smooth_func({0, 0, 0, 0})` does *not* produce zero (but a value based on the *number* of zeros), it seems we should simply always pad it out with zeros (even if it was empty) and write out the `smooth_func(penalties)` to `y`.

Probably need a unit test for this as well.
___
*[`solvers/minimum_value_constraint.cc` line 306 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZSzuNh9HyfMGDWtEQy:-NZSzuNh9HyfMGDWtEQz:b6qmtz5) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L306)):*
> ```cpp
>         return SmoothUnderMax<T>(v);
>       },
>       y);
> ```

BTW The simpler, more obvious spelling works, I believe. (At least it did work for me locally.) I assume your *first* attempt you just named `SmoothOverMax` (omitting the template declaration) and *that* wouldn't work.

```c++
  this->EvalMaxPenalty<T>(minimum_value_lower_, values, y_lower_index,
                          SmoothOverMax<T>, y);
  this->EvalMaxPenalty<T>(minimum_value_upper_, values, y_upper_index,
                          SmoothUnderMax<T>, y);
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`solvers/minimum_value_constraint.cc` line 248 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZStOssBkWInlsCr5h6:-NZTJDiCBPwvyuwP81Kj:b5ix2by) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L248)):*

Sorry, every time I look at this code, something *else* jumps out at me.

If `value >= influence_value_` *for all values*,  we add *no* entries to `penalties` and, therefore, we do not write to `(*y)(y_index)`. The output does not get changed. Is this right?

I looked at the `EvaluatorBase` documentation and it makes no statements one way or the other about whether `y` is guaranteed to be written to. Without warning that `y` *might* not get written to, it seems the default would be that `y` should always get written to (that's my own interpretative bias).

Recognizing the fact that `smooth_func({0, 0, 0, 0})` does *not* produce zero (but a value based on the *number* of zeros), it seems we should simply always pad it out with zeros (even if it was empty) and write out the `smooth_func(penalties)` to `y`.

Probably need a unit test for this as well.
</blockquote></details>

`y` is written in the DoEvalGeneric method, when `value >= influence_value_` for all values, when we call `InitializeY` function.

```cc
  if (y_lower_index >= 0) {
    InitializeY(x, y, y_lower_index,
                SmoothOverMax(std::vector<double>(max_num_values_, 0.0)));
  }
  if (y_upper_index >= 0) {
    InitializeY(x, y, y_upper_index,
                SmoothUnderMax(std::vector<double>(max_num_values_, 0.0)));
  }
```

I added a test with all `value >= influence_value_`.
___
*[`solvers/minimum_value_constraint.cc` line 306 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZSzuNh9HyfMGDWtEQy:-NZTMOxH3snGH_80v9lY:b2ofzcs) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L306)):*

BTW The simpler, more obvious spelling works, I believe. (At least it did work for me locally.) I assume your *first* attempt you just named `SmoothOverMax` (omitting the template declaration) and *that* wouldn't work.

```c++
  this->EvalMaxPenalty<T>(minimum_value_lower_, values, y_lower_index,
                          SmoothOverMax<T>, y);
  this->EvalMaxPenalty<T>(minimum_value_upper_, values, y_upper_index,
                          SmoothUnderMax<T>, y);
```
</blockquote></details>

Right, I tried that simpler version. It works with gcc but not with clang.

Here is Chatgpt's explanation on why the compilation fails
```
Your code fails to compile because you are trying to convert a template function to an `std::function`. While this works in GCC due to a non-standard extension, it is not standard C++ and thus does not work in Clang.

You should instead pass a lambda function that captures the function template
```

SeanCurtis-TRI (commented):
:LGTM:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-:-NZTwRB81uYQCeisd7Iy:bmxilld)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),RussTedrake(platform)

___
*[`solvers/minimum_value_constraint.cc` line 248 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZStOssBkWInlsCr5h6:-NZTw8yxCwNYLiIAF8EY:bwtw4h) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L248)):*

`y` is written in the DoEvalGeneric method, when `value >= influence_value_` for all values, when we call `InitializeY` function.

```cc
  if (y_lower_index >= 0) {
    InitializeY(x, y, y_lower_index,
                SmoothOverMax(std::vector<double>(max_num_values_, 0.0)));
  }
  if (y_upper_index >= 0) {
    InitializeY(x, y, y_upper_index,
                SmoothUnderMax(std::vector<double>(max_num_values_, 0.0)));
  }
```

I added a test with all `value >= influence_value_`.
</blockquote></details>

OH, right. I needed to take a step back and look there again.  :)
___
*[`solvers/minimum_value_constraint.cc` line 306 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19665#-NZSzuNh9HyfMGDWtEQy:-NZTwHYDBPmransf3Qxo:b-1d3rl8) ([raw file](https://github.com/RobotLocomotion/drake/blob/5694a3871cab1e31586800233f70a93931df92a0/solvers/minimum_value_constraint.cc#L306)):*

Right, I tried that simpler version. It works with gcc but not with clang.

Here is Chatgpt's explanation on why the compilation fails
```
Your code fails to compile because you are trying to convert a template function to an `std::function`. While this works in GCC due to a non-standard extension, it is not standard C++ and thus does not work in Clang.

You should instead pass a lambda function that captures the function template
```
</blockquote></details>

There we go.  :) Sounds good to me.

Issue #15320, opened 2021-07-06
More granular control on ChooseBestSolver


hongkai-dai:
Currently `ChooseBestSolver` has a linear decision chain, it mains a list of solvers, checks if the first solver in the list is available and satisfies the program requirement. If yes it calls the first solver, otherwise it moves to the second solver, and so on.

The issue is that between two solvers solverA and solverB, the order of choosing these solvers should not be fixed, they should depend on mathematical program property. For example, IPOPT usually gives a more accurate solution than SCS for linear program, but for program with second order cone constraint, SCS should be preferable to IPOPT, as seen from #15316 .

I propose that at the beginning of `ChooseBestSolver`, we first determine the type of the optimization problem, and for some problem types, we use its own list of solvers. For example, if we determine the optimization is a linear program, then we use the list `Mosek, Gurobi, CLP, IPOPT, SCS`; if the optimization is second order cone program, then we use the list `Mosek, Gurobi, SCS, CSDP`. If the program doesn't fall into these special types (LP or SOCP) where we have designated list of solvers, then it falls back to the current implementation.

cc @jwnimmer-tri @RussTedrake

RussTedrake:
I agree with the need for more granular control.

In particular, I had the experience of writing a mathematical program which I knew was convex.  But after calling the `ChooseBestSolver` via `solvers::Solve`, the default solver got stuck in a local minima!  @hongkai-dai outlined the reasoning above, but let me just emphasize the detail:  

For conic constraints in particular, some solvers handle them explicitly (revealing their convexity), but when we hand them to a more general purpose solver like IPOPT, they see a slightly different formulation which can have zero-gradients / local minima.  Since IPOPT turned out to be faster/more accurate than the open-source convex solvers for some convex optimization problems, it moved itself up in the list of `ChooseBestSolver`.  But then I ended up with local minima in my convex optimization.

I feel that for conic constraints (and presumably SDP constraints, perhaps even others), where we fall back to a generic constraint implementation when the solver doesn't handle things explicitly, we have introduced this potential problem.  I think the proposed solution above definitely helps.  I still don't like that the user can be so surprised.

[...]

PR #19671, opened 2023-06-25
Hyperellipsoid::MinimumUniformScalingToTouch handles the empty case.

RussTedrake:
Previously, if the set `other` had no interior, the solver would fail the method would throw a confusing error message. Now it handles this case explicitly.

This PR also resolves a TODO to return to ChooseBestSolver in the MinimumUniformScalingToTouch now that #15320 is resolved.

+@cohnt for feature review, please?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -387,6 +387,7 @@ drake_cc_googletest(
         ":convex_set",
         ":test_utilities",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//common/yaml",
     ],
 )
--- a/geometry/optimization/hyperellipsoid.cc
+++ b/geometry/optimization/hyperellipsoid.cc
@@ -138,11 +138,20 @@ std::pair<double, VectorXd> Hyperellipsoid::MinimumUniformScalingToTouch(
   solvers::MathematicalProgramResult result;
   solver->Solve(prog, std::nullopt, std::nullopt, &result);
   if (!result.is_success()) {
-    throw std::runtime_error(fmt::format(
-        "Solver {} failed to solve the `minimum uniform scaling to touch' "
-        "problem; it terminated with SolutionResult {}). This should not "
-        "happen.",
-        result.get_solver_id().name(), result.get_solution_result()));
+    // Check if `other` is empty.
+    MathematicalProgram prog2;
+    auto x2 = prog2.NewContinuousVariables(ambient_dimension());
+    other.AddPointInSetConstraints(&prog2, x2);
+    auto result2 = Solve(prog2);
+    if (!result2.is_success()) {
+      throw std::runtime_error("The set `other` is empty.");
+    } else {
+      throw std::runtime_error(fmt::format(
+          "Solver {} failed to solve the `minimum uniform scaling to touch' "
+          "problem; it terminated with SolutionResult {}). The solver likely "
+          "ran into numerical issues.",
+          result.get_solver_id().name(), result.get_solution_result()));
+    }
   }
   return std::pair<double, VectorXd>(std::sqrt(result.get_optimal_cost()),
                                      result.GetSolution(x));
--- a/geometry/optimization/hyperellipsoid.h
+++ b/geometry/optimization/hyperellipsoid.h
@@ -63,6 +63,7 @@ class Hyperellipsoid final : public ConvexSet {
   precision).
   @pre `other` must have the same ambient_dimension as this.
   @returns the minimal scaling and the witness point, x, on other.
+  @throws std::exception if `other` is empty.
   @throws std::exception if ambient_dimension() == 0 */
   std::pair<double, Eigen::VectorXd> MinimumUniformScalingToTouch(
       const ConvexSet& other) const;
--- a/geometry/optimization/test/hyperellipsoid_test.cc
+++ b/geometry/optimization/test/hyperellipsoid_test.cc
@@ -6,6 +6,7 @@
 
 #include "drake/common/eigen_types.h"
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/common/yaml/yaml_io.h"
 #include "drake/geometry/geometry_frame.h"
 #include "drake/geometry/optimization/hpolyhedron.h"
@@ -19,6 +20,7 @@ namespace drake {
 namespace geometry {
 namespace optimization {
 
+using Eigen::Matrix2d;
 using Eigen::Matrix3d;
 using Eigen::MatrixXd;
 using Eigen::RowVector2d;
@@ -500,6 +502,20 @@ GTEST_TEST(HyperellipsoidTest, MinimumUniformScaling4) {
   auto [sigma, x] = E.MinimumUniformScalingToTouch(E2);
   EXPECT_NEAR(sigma, 3.0, kTol);
   EXPECT_TRUE(CompareMatrices(x, Vector2d{3.0, 0.0}, kTol));
+}
+
+// Check the case when `other` has no interior.
+GTEST_TEST(HyperellipsoidTest, MinimumUniformScaling5) {
+  const Hyperellipsoid E = Hyperellipsoid::MakeUnitBall(2);
+  // x₀ ≤ -1, x₀ ≥ 1.
+  Matrix2d A;
+  // clang-format off
+  A <<  1, 0,
+       -1, 0;
+  // clang-format on
+  const HPolyhedron H(A, Vector2d(-1, -1));
+  DRAKE_EXPECT_THROWS_MESSAGE(E.MinimumUniformScalingToTouch(H),
+                              ".*is empty.*");
 }
 
 GTEST_TEST(HyperellipsoidTest, Serialize) {


cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671)* status: 2 unresolved discussions, LGTM missing from assignee cohnt, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/hyperellipsoid.cc` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NYnoHDtEpCZyAgajjkI:-NYno3C395z4j4s_ADwG:b-miz237) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff6307c4e6b8eff0e4629fe15e955dece2fe276e/geometry/optimization/hyperellipsoid.cc#L131)):*
> ```cpp
>   }
>   auto result = Solve(prog);
>   if (!result.is_success()) {
> ```

Does it make more sense to check that `other` has an interior at the beginning of this function? I imagine it depends on runtime, but I don't have a sense of how long these actions take.
___
*[`geometry/optimization/hyperellipsoid.cc` line 136 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NYnmOxABvBW4KYcLsci:-NYnmOxABvBW4KYcLscj:b-i6gmu9) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff6307c4e6b8eff0e4629fe15e955dece2fe276e/geometry/optimization/hyperellipsoid.cc#L136)):*
> ```cpp
>     auto x2 = prog2.NewContinuousVariables(ambient_dimension());
>     other.AddPointInSetConstraints(&prog2, x2);
>     auto result2 = Solve(prog2);
> ```

I don't follow the logic for how this checks if `other` has an interior. Is it due to some numerical tolerance that `AddPointInSetConstraints` enforces?

Suppose `other` is a `Point`. Then the optimization problem will be successful, as shown by this python snippet:
```
from pydrake.all import Point, MathematicalProgram, Solve

point = Point([0,0])

prog = MathematicalProgram()
x = prog.NewContinuousVariables(2)
point.AddPointInSetConstraints(prog, x)

result = Solve(prog)
print(result.is_success())
```

I think one option to check this rigorously would be to set up an optimization problem which sticks a small ball or cube inside the set. Long-term, maybe it would make sense to create a method in the `ConvexSet` class to check the affine dimension (or at least check that the interior is nonempty). Then this function could just call that method.

Another thought, at the risk of rambling... if ConvexSet already stored its affine dimension (or perhaps computed it the first time its needed, and then stored it), then checking if the interior is nonempty would be zero cost, so the check could be moved to before the MathematicalProgram is solved. /ramble

RussTedrake (commented):
The CI failure is due to ChooseBestSolver picking CsdpSolver over ScsSolver.  I've just asked on the slack optimization channel, and @hongkai-dai agrees we can flip the order of those in `choose_best_solver.cc`.  This should pass CI once that lands.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-:-NYrxM3Z2HrunzGnvIcJ:btimpdv)* status: 2 unresolved discussions, LGTM missing from assignee cohnt, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/hyperellipsoid.cc` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NYnoHDtEpCZyAgajjkI:-NYruGQIALyE2JS4P9NT:bp8r3zf) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff6307c4e6b8eff0e4629fe15e955dece2fe276e/geometry/optimization/hyperellipsoid.cc#L131)):*

Does it make more sense to check that `other` has an interior at the beginning of this function? I imagine it depends on runtime, but I don't have a sense of how long these actions take.
</blockquote></details>

The interior check is only used here to give a better error message, and it does involve calling the solver, so I'd like to avoid except when it's absolutely necessary.
___
*[`geometry/optimization/hyperellipsoid.cc` line 136 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NYnmOxABvBW4KYcLsci:-NYrv4Gn1YYQndcY3ds3:bt8vwn) ([raw file](https://github.com/RobotLocomotion/drake/blob/ff6307c4e6b8eff0e4629fe15e955dece2fe276e/geometry/optimization/hyperellipsoid.cc#L136)):*

I don't follow the logic for how this checks if `other` has an interior. Is it due to some numerical tolerance that `AddPointInSetConstraints` enforces?

Suppose `other` is a `Point`. Then the optimization problem will be successful, as shown by this python snippet:
```
from pydrake.all import Point, MathematicalProgram, Solve

point = Point([0,0])

prog = MathematicalProgram()
x = prog.NewContinuousVariables(2)
point.AddPointInSetConstraints(prog, x)

result = Solve(prog)
print(result.is_success())
```

I think one option to check this rigorously would be to set up an optimization problem which sticks a small ball or cube inside the set. Long-term, maybe it would make sense to create a method in the `ConvexSet` class to check the affine dimension (or at least check that the interior is nonempty). Then this function could just call that method.

Another thought, at the risk of rambling... if ConvexSet already stored its affine dimension (or perhaps computed it the first time its needed, and then stored it), then checking if the interior is nonempty would be zero cost, so the check could be moved to before the MathematicalProgram is solved. /ramble
</blockquote></details>

Done.  You're absolutely right! I was careless with my words.  I meant only to check if the set was empty. Whether it has in interior is not actually relevant here.  Good catch.

RussTedrake (commented):
+@sammy-tri for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-:-NYxeCyECFh_l_rXp0Ri:bz1t39v)* status: LGTM missing from assignee sammy-tri(platform)

sammy-tri (commented):
I'm out of time today.  -@sammy-tri +@EricCousineau-TRI for tomorrow

> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-:-NYz2poUEEr9e1Y_vXY6:bociqk9)* status: LGTM missing from assignee EricCousineau-TRI(platform)

EricCousineau-TRI (commented):
:lgtm: platform pending CI fix

> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-:-NZ1PKVN9W6m_rQcGCa8:b-o7mqym)* status: :shipit: complete! all discussions resolved, LGTM from assignees cohnt,EricCousineau-TRI(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NZ1ObQ1DH4xOsRyrbyP:-NZ1ObQ1DH4xOsRyrbyQ:byuv2ev) (no related file):*
Test failure on `linux-focal-clang-bazel-experimental-release` - maybe due to CSDP now being used?

https://drake-cdash.csail.mit.edu/test/1101439505
```
[ RUN      ] HyperellipsoidTest.MinimumUniformScaling4
[2023-06-26 13:57:17.499] [console] [warning] The problem has free variables, and CSDP removes the free variables by computing the null space of linear constraint in the dual space. This step can be time consuming. Consider providing a lower and/or upper bound for each decision variable.
unknown file: Failure
C++ exception with description "Solver CSDP failed to solve the `minimum uniform scaling to touch' problem; it terminated with SolutionResult SolverSpecificError). The solver likely ran into numerical issues." thrown in the test body.
[  FAILED  ] HyperellipsoidTest.MinimumUniformScaling4 (2 ms)
```

___

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671)* status: :shipit: complete! all discussions resolved, LGTM from assignees cohnt,EricCousineau-TRI(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-NZ1ObQ1DH4xOsRyrbyP:-NZDbGhd0nBbaoXqObnA:bq9a6xz) (no related file):*

Test failure on `linux-focal-clang-bazel-experimental-release` - maybe due to CSDP now being used?

https://drake-cdash.csail.mit.edu/test/1101439505
```
[ RUN      ] HyperellipsoidTest.MinimumUniformScaling4
[2023-06-26 13:57:17.499] [console] [warning] The problem has free variables, and CSDP removes the free variables by computing the null space of linear constraint in the dual space. This step can be time consuming. Consider providing a lower and/or upper bound for each decision variable.
unknown file: Failure
C++ exception with description "Solver CSDP failed to solve the `minimum uniform scaling to touch' problem; it terminated with SolutionResult SolverSpecificError). The solver likely ran into numerical issues." thrown in the test body.
[  FAILED  ] HyperellipsoidTest.MinimumUniformScaling4 (2 ms)
```
</blockquote></details>

Thanks.  As I commented above, this is waiting for #19676 to land before I expect the tests to pass.  (if we decide to not complete #19676, then i'll update this PR accordingly)

___

RussTedrake (commented):
FWIW -- I've removed the ChooseBestSolver cleanup from this PR to separate it from #19676.  Should be ready to go.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19671#-:-NZcLUGvAsQhY9HmqVa2:b9y8kxj)* status: :shipit: complete! all discussions resolved, LGTM from assignees cohnt,EricCousineau-TRI(platform)

PR #19673, opened 2023-06-25
Adds some text logging to IrisInConfigurationSpace

RussTedrake:
For long-running calls to IRIS, this gives some much needed feedback. It also can give some insights on if/where IRIS is getting bogged down.

This PR also clarifies the usage of the term "seed" vs "samples" vs "searches", which got even more confusing when we start reporting excessive numbers of searches.

The code does not change the computation (and therefore does not change the tests).

+@cohnt for feature review, please?
cc @rhjiang

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673)

### DIFF
--- a/geometry/optimization/iris.cc
+++ b/geometry/optimization/iris.cc
@@ -428,7 +428,7 @@ HPolyhedron IrisInConfigurationSpace(const MultibodyPlant<double>& plant,
   // Check the inputs.
   plant.ValidateContext(context);
   const int nq = plant.num_positions();
-  const Eigen::VectorXd sample = plant.GetPositions(context);
+  const Eigen::VectorXd seed = plant.GetPositions(context);
   const int nc = static_cast<int>(options.configuration_obstacles.size());
   // Note: We require finite joint limits to define the bounding box for the
   // IRIS algorithm.
@@ -437,7 +437,7 @@ HPolyhedron IrisInConfigurationSpace(const MultibodyPlant<double>& plant,
   DRAKE_DEMAND(options.num_collision_infeasible_samples >= 0);
   for (int i = 0; i < nc; ++i) {
     DRAKE_DEMAND(options.configuration_obstacles[i]->ambient_dimension() == nq);
-    if (options.configuration_obstacles[i]->PointInSet(sample)) {
+    if (options.configuration_obstacles[i]->PointInSet(seed)) {
       throw std::runtime_error(
           fmt::format("The seed point is in configuration obstacle {}", i));
     }
@@ -454,7 +454,7 @@ HPolyhedron IrisInConfigurationSpace(const MultibodyPlant<double>& plant,
   DRAKE_DEMAND(P.A().rows() == 2 * nq);
   const double kEpsilonEllipsoid = 1e-2;
   Hyperellipsoid E = options.starting_ellipse.value_or(
-      Hyperellipsoid::MakeHypersphere(kEpsilonEllipsoid, sample));
+      Hyperellipsoid::MakeHypersphere(kEpsilonEllipsoid, seed));
 
   // Make all of the convex sets and supporting quantities.
   auto query_object =
@@ -482,7 +482,7 @@ HPolyhedron IrisInConfigurationSpace(const MultibodyPlant<double>& plant,
       std::make_shared<internal::SamePointConstraint>(&plant, context);
 
   // As a surrogate for the true objective, the pairs are sorted by the distance
-  // between each collision pair from the sample point configuration. This could
+  // between each collision pair from the seed point configuration. This could
   // improve computation times and produce regions with fewer faces.
   std::vector<GeometryPairWithDistance> sorted_pairs;
   for (const auto& [geomA, geomB] : pairs) {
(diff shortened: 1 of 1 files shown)


RussTedrake:
*[`geometry/optimization/iris.cc` line 635 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-NYmGeBbCiI4Xq3-08fo:-NYmGeBc4a_P4peNwGrJ:b-a0m4sx) ([raw file](https://github.com/RobotLocomotion/drake/blob/2dbd7e6fa98b8b5a1f9c3dbc8ee0ac1fb1bd04e1/geometry/optimization/iris.cc#L635)):*
> ```cpp
>           *sets.at(pair.geomA), *sets.at(pair.geomB), E,
>           A.topRows(num_constraints), b.head(num_constraints));
>       while (sample_point_requirement &&
> ```

fyi - this check moved the new if (!seed_point_requirement) break below, which also helped readability and avoids extra no-op work if the seed_point_requirement becomes false.

cohnt (commented):
Variable name changes all make sense, and print logic seems helpful. Just one pending question from me

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-:-NYnijC0EE7rJGh1qBuo:ba9a090)* status: 1 unresolved discussion, LGTM missing from assignee cohnt, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/iris.cc` line 588 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-NYnhhJz-UM__gNX8IZi:-NYnhhJz-UM__gNX8IZj:b-hv4hsu) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e3a6d54849b0689ccaca5aa71615b7258207d64/geometry/optimization/iris.cc#L588)):*
> ```cpp
> 
>   while (true) {
>     log()->info("IrisInConfigurationSpace iteration {}", iteration);
> ```

I'm not very familiar with how logging works behind the scenes in Drake (and how those changes percolate upwards into the Python bindings), but is there a way to turn off this logging? In scenarios where I'm generating large numbers of IRIS regions (possibly with multithreading), this sort of per-region logging could become inconvenient. Perhaps it would be nice to include a field in IrisOptions to toggle logging?

RussTedrake (commented):
Thanks!  If you're happy, please mark this as :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-:-NYo5yr2BegG45F7otKg:b4yfdhh)* status: 1 unresolved discussion, LGTM missing from assignees cohnt,ggould-tri(platform)

___
*[`geometry/optimization/iris.cc` line 588 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-NYnhhJz-UM__gNX8IZi:-NYo5Z41C6EVhFubiZKB:b-203wnm) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e3a6d54849b0689ccaca5aa71615b7258207d64/geometry/optimization/iris.cc#L588)):*

I'm not very familiar with how logging works behind the scenes in Drake (and how those changes percolate upwards into the Python bindings), but is there a way to turn off this logging? In scenarios where I'm generating large numbers of IRIS regions (possibly with multithreading), this sort of per-region logging could become inconvenient. Perhaps it would be nice to include a field in IrisOptions to toggle logging?
</blockquote></details>

The beauty of using this logging feature is that users can set the verbosity globally (in both python and c++).  The syntax you see here says this is at the "info" level.  Any reporting level less verbose than "info" will turn it off.

https://github.com/gabime/spdlog

cohnt (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-:-NYoWKEM6N__GRz0Unf5:bg7igkv)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform)

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673)* status: 1 unresolved discussion, LGTM missing from assignee ggould-tri(platform)

___
*[`geometry/optimization/iris.cc` line 588 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-NYnhhJz-UM__gNX8IZi:-NYr_-0p3vR7D5HWhH6Y:b64zwgz) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e3a6d54849b0689ccaca5aa71615b7258207d64/geometry/optimization/iris.cc#L588)):*

The beauty of using this logging feature is that users can set the verbosity globally (in both python and c++).  The syntax you see here says this is at the "info" level.  Any reporting level less verbose than "info" will turn it off.

https://github.com/gabime/spdlog
</blockquote></details>

@cohnt -- i'll need you to clear this comment before I can merge.

cohnt (commented):
Sorry -- thought putting :lgtm: was sufficient. All sounds good to me!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-:-NYraomp7Kt7FHdAYqCw:bof4ig3)* status: LGTM missing from assignee ggould-tri(platform)

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19673)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),cohnt

___
*[`geometry/optimization/iris.cc` line 662 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19673#-NYsi9BM0FYIW8BYKs4R:-NYy0jDp4YediZ4F7UYr:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/6e3a6d54849b0689ccaca5aa71615b7258207d64/geometry/optimization/iris.cc#L662)):*

nit:  A naive reader could think that this is a stop condition.
</blockquote></details>

Done.

PR #19676, opened 2023-06-26
[solvers] prefer SCS to CSDP.

hongkai-dai:
In ChooseBestSolver, prefers SCS to CSDP. SCS might give a low accuracy solution, but CSDP fail to solve some trivial problems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676)

### DIFF
--- a/solvers/choose_best_solver.cc
+++ b/solvers/choose_best_solver.cc
@@ -163,10 +163,10 @@ void GetAvailableSolversHelper(
           // Dispreferred (generic nonlinear solvers).
           // I generally find SNOPT faster than IPOPT. Nlopt is less reliable.
           SnoptSolver, IpoptSolver, NloptSolver,
-          // Dispreferred (cannot handle free variables).
-          CsdpSolver,
           // Dispreferred (ADMM, low accuracy).
-          ScsSolver>(result);
+          ScsSolver,
+          // Dispreferred (cannot handle free variables).
+          CsdpSolver>(result);
       return;
     }
     case ProgramType::kQP: {
@@ -202,10 +202,10 @@ void GetAvailableSolversHelper(
           // According to http://plato.asu.edu/ftp/socp.html, Mosek is slightly
           // faster than Gurobi for SOCP, but the difference is small.
           MosekSolver, GurobiSolver,
-          // Dispreferred (cannot handle free variables).
-          CsdpSolver,
           // Dispreferred (ADMM, low accuracy).
           ScsSolver,
+          // Dispreferred (cannot handle free variables).
+          CsdpSolver,
           // Dispreferred (generic nonlinear solvers).
           // I strongly suggest NOT to use these solvers for SOCP. These
           // gradient-based solvers ignore the convexity of the problem, and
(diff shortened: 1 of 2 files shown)


hongkai-dai (commented):
+@RussTedrake for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NYrzzF18XfdPqVH_zZJ:b-7vq6ss)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers

RussTedrake (commented):
:lgtm: Thanks!  I'll wait til you look at the CI failures before passing to platform.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NYxeeUi8Hg3tZ4HOy_x:bvyen7c)* status: needs at least two assigned reviewers

RussTedrake:
We should probably discuss whether those CI failures are real enough to suggest that SCS is brittle, too.
For GCS, at least, it might be reasonable to prefer CSDP (e.g. setting our own preferred solvers instead of using ChooseBestSolver) since we expect all of the decision variables to be bounded.

hongkai-dai (commented):
> We should probably discuss whether those CI failures are real enough to suggest that SCS is brittle, too.  
For GCS, at least, it might be reasonable to prefer CSDP (e.g. setting our own preferred solvers instead of using ChooseBestSolver) since we expect all of the decision variables to be bounded.

I think these variables don't have the lower or upper bound https://github.com/RobotLocomotion/drake/blob/31eae0b156a77763822a5aeb4fc1633471984755/geometry/optimization/graph_of_convex_sets.cc#L701-L703?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NYyMX4B27dDqqtc6uc8:b-w4sfvl)* status: needs at least two assigned reviewers

hongkai-dai (commented):
@jwnimmer-tri How can I enable a specific `GTEST_TEST` when gurobi/mosek are enabled? After we switch to SCS, many problems cannot be solved accurately, so I only want to have the CI to test these problems with gurobi/mosek.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NZ3v-I6BurJ0Hladkb8:bah6h6r)* status: needs at least two assigned reviewers

RussTedrake (commented):
We have 
```
if (!GurobiOrMosekSolverAvailable()) {
return;
}
```
in many of the unit tests in `gcs_trajectory_optimization_test.cc`

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NZ5gkwx3U1E6eQejBpU:b1g8g2g)* status: needs at least two assigned reviewers

jwnimmer-tri:
> After we switch to SCS, many problems cannot be solved accurately, ...

Isn't this a canary in the coal mine?  If a bunch of tests are unhappy now, won't all of our users' programs that are similar to those tests also start failing?

hongkai-dai (commented):
It would affect many of our users who rely on solvers::Solve() to use CSDP solver.

I can spend a day to understand why CSDP solver fails for the simple problem and try to fix CSDP.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19676#-:-NZ6OeKE04Hbwg0CrCih:b-b612wo)* status: needs at least two assigned reviewers

Issue #19680, opened 2023-06-26
[pydrake] Disable spdlog redirection when using multithreaded C++ code?


EricCousineau-TRI:
**Is your feature request related to a problem? Please describe.**
I am writing Python bindings in a downstream project (Anzu) that does work in a multithreaded fashion, and calls `drake::log()->info()`, `->debug()`, `->trace()`, etc., in those separate threads.

Even though this code doesn't explicitly call Python code in those other threads (e.g. not via `std::function<>` or C++ <-> Python polymorphism), it still seemed to deadlock.

If I commented out this line, my multithreaded code seemed to work fine:
https://github.com/RobotLocomotion/drake/blob/b9813a712454580cd6bc0724f8d2506d85412e58/bindings/pydrake/common/module_py.cc#L103

**Describe the solution you'd like**
Perhaps an environment variable, like `PYDRAKE_NO_SPDLOG_REDIRECT`, or a function to call from pydrake that undoes the effects of `MaybeRedirectPythonLogging()` and related methods.

**Describe alternatives you've considered**
(1) Writing a C++ pybind11 module as follows, then importing it before `pydrake`, fixes the "seems to deadlock" issue:
https://gist.github.com/EricCousineau-TRI/ff52a8e3be0ca5fd085307c16ff09dbc

(2) Set spdlog's debugging level such that no messages are printed

(3) Add an option to only redirect to Python if it's the main thread. If it isn't the main thread, it should ideally emit directly to `stderr` as the nominal sink would.

**Additional context**
https://drakedevelopers.slack.com/archives/C3YB3EV5W/p1687807563590549

\cc @jwnimmer-tri @dmcconachie-tri @sammy-tri

jwnimmer-tri:
If we end up with an environment variable, my vote to best align with [the current names](https://drake.mit.edu/doxygen_cxx/group__environment__variables.html) would be to name it starting with `DRAKE_`.

Perhaps `DRAKE_PYTHON_LOGGING`, where unset or set-to-empty means the status quo and `DRAKE_PYTHON_LOGGING=0` switches `MaybeRedirectPythonLogging()` to be a no-op.

jwnimmer-tri:
> Perhaps an environment variable ... or a function to call from pydrake that undoes the effects of MaybeRedirectPythonLogging ...

Either (or both) of those features sound fine to me.  The environment variable is probably the easier one to implement, but the undo function would probably cover more use cases.

[...]

PR #19683, opened 2023-06-26
[bindings] Only log to Python sink for main thread

EricCousineau-TRI:
Implements workaround (3) mentioned in #19680

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19683)

### DIFF
--- a/bindings/pydrake/common/text_logging_pybind.cc
+++ b/bindings/pydrake/common/text_logging_pybind.cc
@@ -24,6 +24,16 @@ namespace internal {
 
 #ifdef HAVE_SPDLOG
 namespace {
+
+bool& mutable_is_main_thread() {
+  thread_local static bool is_main{false};
+  return is_main;
+}
+
+bool is_main_thread() {
+  return mutable_is_main_thread();
+}
+
 class pylogging_sink final
     // We use null_mutex below because we'll use the GIL as our *only* mutex.
     // This is critically important to avoid deadlocks by lock order inversion.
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI (commented):
@jwnimmer-tri What do you think about this approach?
It would probably keep "90%" of pydrake code looking the same, and only deviate when there are threads involved.
I could also make this optional.

Alternatively, I can implement the environment variable approach if you think that's simplest.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19683#-:-NYtupmU9EMZMgvrXfEk:b-vfi3rs)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

jwnimmer-tri:
I'm not sure why this is better than just turning off the Python redirection as a whole.  Could you clarify the pro/con as you see it?

On the 'con' side: Log messages will no longer be uniformly formatted within the process, and changes to the Python logging level will affect more than just the Python sink.  I also have the general impression that `thread_local` is a pretty expensive language construct, so I would want to benchmark this (myself) if you think this a worthwhile direction, as a sanity check.

EricCousineau-TRI:
For this approach:
The pro would be that logging from the main thread (probably most use cases) would still function normally, such as in Jupyter notebooks, and something we could potentially turn "always on" in certain places like Anzu.
The cons would be as you mentioned - inconsistency and possibly performance.

For disabling everything:
The pro is simplicity.
The con is if you're say launching controllers from a Jupyter notebook (which IMO not sure if it's the best idea), then all of the logging will go to terminal stderr vs. the redirected stderr.

Given the above, though, I'm fine with going the route of disabling everything.

jwnimmer-tri:
In the case where using Python `logging` is important (because stderr text gets discarded, or because it's hooked up to a network logger, or etc.), I think the better answer is to use our current default and eat the GIL slowdown.  (Possibly the user would set the log threshold to WARN in this case, no INFO.)  Log messages from worker threads are not any less important than log message from the main thread, so if there is an important use case for having the main thread stay on Python `logging`, it applies equally well for worker threads.  That argues against any per-thread policy variation.

EricCousineau-TRI:
Aye, sgtm. Closing!

Issue #19532, opened 2023-06-05
Improving RenderEngineGltfClient fidelity by consuming glTF files


SeanCurtis-TRI:
### The problem
Drake currently supports an incredibly impoverished material specification for the OBJ meshes it consumes. Essentially, it only uses a phong illumination model and, even then, only allows specification of diffuse color (or diffuse texture).

The `RenderEngineGltfClient` implementation is intended to be able to farm Drake state to an external rendering process to create a higher-fidelity rendering of the simulation state. Frequently, model specifications begin with high-fidelity materials (multiple textured parameters or even different rendering models like PBR). Passing them through Drake (and the `RenderEngine` and `SceneGraph` APIs) inevitably filters those specifications down to a fisher-price version of the model. The resulting external rendering suffers for this fidelity reduction.

### The solution
Given `RenderEngineGltfClient` broadcasts the Drake scene as a glTF file, it would be nice to support input models as glTF files. If a model is specified as a glTF file, we'd like the full fidelity of that model to be available to the rendering server. The high level proposal (and impact on APIs) would consist of the following:

1. Allow users to specify "mesh" for perception as a gltf file (a change from our current state where it is obj only).
2. Promise that where such meshes are specified, they are included in `RenderEngineGltfClient`'s output in a lossless manner.

#### Specification

There are a lot of subtleties inherent in this. How do we simultaneously allow URDF/SDF specification that can nevertheless be meaningfully read by other parsers? The answer to this question is still TBD. For now we'll do the following:

Simply specify the .gltf file in the `<visual><mesh>...` tag.

This will cause all of our visualizers and most of our `RenderEngine` implementations to yell ("What is a glTF? I'm ignoring it!") However, the `RenderEngineGltfClient` will happily use it.

In the short-term, it allows for the specification of a high-fidelity representation for external rendering, but it means there will be no geometry with a usable illustration role for that geometry. For now, we recommend the following:

1. Add a *second* `<visual>` tag referencing an OBJ version of the gltf file.
2. Add the `<drake:accepting_renderer>` tag to that visual and provide a nonsense name -- the name of a render engine that definitely does not exist.

The .gltf file will implicitly serve as a perception-only geometry (as only one `RenderEngine` implementation will use it). The obj file will implicitly serve as an illustration only (as no `RenderEngine` implementation will have that name). In the future, the second `<visual>` will become unnecessary as we offer support for .gltf across all visualizers.

#### Passing input glTF to render server

[...]

PR #19684, opened 2023-06-27
Allowing RenderEngineGltfClient to accept gltf

SeanCurtis-TRI:
The file is treated like a single obj file. If there are multiple root nodes, their transforms are defined relative to a root frame belonging to the whole file. This frame, F, is what we pose in Drake.

relates #19532 (PR 3 of 4)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)


SeanCurtis-TRI (commented):
+(release notes: none) 

+@zachfang for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NYuNirQ0oRnZKdKbSsI:bnugo93)* status: LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

SeanCurtis-TRI:
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NYvipbSEbBDgI0g32l8:-NYvipbSEbBDgI0g32l9:b-ic20j4) ([raw file](https://github.com/RobotLocomotion/drake/blob/eaa68fd8ab8550a2fac0af086291d17067fbb9a3/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L21)):*
> ```objc
> namespace internal {
> 
> #pragma GCC visibility push(hidden)
> ```

@jwnimmer-tri I've got a *private* nested struct below (`GltfRecord`). One of the members of that struct is an `nlohmann::json`. The compiler is unhappy that I'm including a "hidden" type in an "unhidden" type.

These pragmas were my quick and dirty trick to get CI a chance to at least pass.

My next plan was to change my map from `map<GeometryId, GltfRecord>` to `map<GeometryId, GltfRecord*>` and defer the definition of what the `GltfRecord` is until the .cc file. Given the json itself is a mess of heap-allocated memory, starting with one more allocation and pointer is probably harmless.

Any other ideas for resolving this particular conundrum?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NYvipbSEbBDgI0g32l8:-NYyOL6V6tIsgLsJiCNU:brun82l) ([raw file](https://github.com/RobotLocomotion/drake/blob/eaa68fd8ab8550a2fac0af086291d17067fbb9a3/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L21)):*

@jwnimmer-tri I've got a *private* nested struct below (`GltfRecord`). One of the members of that struct is an `nlohmann::json`. The compiler is unhappy that I'm including a "hidden" type in an "unhidden" type.

These pragmas were my quick and dirty trick to get CI a chance to at least pass.

My next plan was to change my map from `map<GeometryId, GltfRecord>` to `map<GeometryId, GltfRecord*>` and defer the definition of what the `GltfRecord` is until the .cc file. Given the json itself is a mess of heap-allocated memory, starting with one more allocation and pointer is probably harmless.

Any other ideas for resolving this particular conundrum?
</blockquote></details>

For sure, the solution is to set these Drake class(es) to hidden visibility.

In fact, formally we are supposed to be marking every class that's declared in a non-installed header file as hidden visibility -- we haven't done that yet only because there's no linter that can assist us with getting it precisely right.

In any case, here's the best spelling for this file:

```patch
--- a/geometry/render_gltf_client/internal_render_engine_gltf_client.h
+++ b/geometry/render_gltf_client/internal_render_engine_gltf_client.h
@@ -7,6 +7,7 @@
 
 #include <drake_vendor/nlohmann/json.hpp>
 
+#include "drake/common/drake_export.h"
 #include "drake/geometry/render/render_camera.h"
 #include "drake/geometry/render_gltf_client/internal_merge_gltf.h"
 #include "drake/geometry/render_gltf_client/internal_render_client.h"
@@ -18,12 +19,11 @@ namespace geometry {
 namespace render_gltf_client {
 namespace internal {
 
-#pragma GCC visibility push(hidden)
-
 /* A RenderEngine that exports
  <a href="https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html">glTF
  </a> scenes, uploads to a render server, and retrieves the renderings back. */
-class RenderEngineGltfClient : public render_vtk::internal::RenderEngineVtk {
+class DRAKE_NO_EXPORT RenderEngineGltfClient
+    : public render_vtk::internal::RenderEngineVtk {
  public:
   /* @name Does not allow copy, move, or assignment  */
   //@{
@@ -158,8 +158,6 @@ class RenderEngineGltfClient : public render_vtk::internal::RenderEngineVtk {
   std::map<GeometryId, GltfRecord> gltfs_;
 };
 
-#pragma GCC visibility pop
-
 }  // namespace internal
 }  // namespace render_gltf_client
 }  // namespace geometry
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NYvipbSEbBDgI0g32l8:-NYzHNIVBQ3odNQd9weC:b22zyxh) ([raw file](https://github.com/RobotLocomotion/drake/blob/eaa68fd8ab8550a2fac0af086291d17067fbb9a3/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L21)):*

For sure, the solution is to set these Drake class(es) to hidden visibility.

In fact, formally we are supposed to be marking every class that's declared in a non-installed header file as hidden visibility -- we haven't done that yet only because there's no linter that can assist us with getting it precisely right.

In any case, here's the best spelling for this file:

```patch
--- a/geometry/render_gltf_client/internal_render_engine_gltf_client.h
+++ b/geometry/render_gltf_client/internal_render_engine_gltf_client.h
@@ -7,6 +7,7 @@
 
 #include <drake_vendor/nlohmann/json.hpp>
 
+#include "drake/common/drake_export.h"
 #include "drake/geometry/render/render_camera.h"
 #include "drake/geometry/render_gltf_client/internal_merge_gltf.h"
 #include "drake/geometry/render_gltf_client/internal_render_client.h"
@@ -18,12 +19,11 @@ namespace geometry {
 namespace render_gltf_client {
 namespace internal {
 
-#pragma GCC visibility push(hidden)
-
 /* A RenderEngine that exports
  <a href="https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html">glTF
  </a> scenes, uploads to a render server, and retrieves the renderings back. */
-class RenderEngineGltfClient : public render_vtk::internal::RenderEngineVtk {
+class DRAKE_NO_EXPORT RenderEngineGltfClient
+    : public render_vtk::internal::RenderEngineVtk {
  public:
   /* @name Does not allow copy, move, or assignment  */
   //@{
@@ -158,8 +158,6 @@ class RenderEngineGltfClient : public render_vtk::internal::RenderEngineVtk {
   std::map<GeometryId, GltfRecord> gltfs_;
 };
 
-#pragma GCC visibility pop
-
 }  // namespace internal
 }  // namespace render_gltf_client
 }  // namespace geometry
```
</blockquote></details>

Thanks. I'd missed that going in.

zachfang (commented):
Checkpoint for reviewing everything except the test file and the new anonymous functions in the .cc file.

> Reviewed 7 of 11 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ6ZtAW4MFAVLKvwpvv:b-a38rdn)* status: 9 unresolved discussions, LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6QIli-iOlUW14YeyF:-NZ6QIli-iOlUW14YeyG:b-mi95ok) (no related file):*
BTW. I wonder if adding a `from_drake: true/false` tag would make the server/client relationship too "intimate" or the glTF too "customized"?

___
___
*[`geometry/render_gltf_client/BUILD.bazel` line 96 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6UDNC0Kai5GfegTiT:-NZ6UDNC0Kai5GfegTiU:b89iik2) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/BUILD.bazel#L96)):*
> ```Starlark
>         ":internal_http_service",
>         ":internal_http_service_curl",
>         ":internal_merge_gltf",
> ```

nit. `internal_render_client.{cc, h}` are not touched in this PR, I assume the added dependencies (and `:nlohmann`) should be moved down to `internal_render_engine_gltf_client`?



_Code quote:_
```Starlark
":internal_merge_gltf",
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 94 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6H0-HBt0OphOWwc0b:-NZ6H0-HBt0OphOWwc0c:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L94)):*
> ```objc
>                    render_vtk::internal::ImageType image_type) const;
> 
>   /* Override RenderEngineVtk's default handling of the mesh types. We want to
> ```

nit.



_Suggestion:_
```objc
Overrides
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 105 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6Yr1k1-8DBN9ObGMf:-NZ6Yr1k1-8DBN9ObGMg:b-g5dpj3) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L105)):*
> ```objc
>                      void* user_data);
>   void ImplementGltf(const std::filesystem::path& gltf_path, double scale,
>                      void* user_data);
> ```

BTW. Maybe we can use a new line to separate these two GltfClient-only functions from the override functions above for clarity. (Even though they might move to the base class eventually.)



_Code quote:_
```objc
  void ImplementMesh(const std::filesystem::path& mesh_path, double scale,
                     void* user_data);
  void ImplementGltf(const std::filesystem::path& gltf_path, double scale,
                     void* user_data);
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 123 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6Hfww810flJFzPzJz:-NZ6Hfww810flJFzPzK-:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L123)):*
> ```objc
>   };
> 
>   // TODO(SeanCurtis-TRI) Based on file path, I should load a gltf *one time*.
> ```

nit.



_Suggestion:_
```objc
Based on a/the file path
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 526 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6JCd5B9HzNv5TUI6Y:-NZ6JCd5B9HzNv5TUI6Z:b-5das0w) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L526)):*
> ```cpp
>                                          ImageType image_type) const {
>   // TODO(SeanCurtis-TRI): Given the ability to edit the gltf in place, we
>   //  should use VTK to create the vtk and merge the gltfs *once* and use that
> ```

nit. The `vtk` appears to be a typo?



_Code quote:_
```cpp
the vtk
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 529 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ42ezR0NmUkM-udHcb:-NZ42ezR0NmUkM-udHcc:b-8zsg8s) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L529)):*
> ```cpp
>   //  as the reference gltf. Then, with a mapping from geometry id to node
>   //  index, we can simply mutate the poses before saving the result. This
>   //  would decrease the cost of transmitting the gltf file *significantly*.
> ```

BTW. Transmitting to me is the RPC part, which should remain mostly the same (regarding the glTF file size). I would assume you meant constructing or converting to a glTF file.



_Code quote:_
```cpp
transmitting
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 569 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6X8hD892yK9K2MgdR:-NZ6X8hD892yK9K2MgdS:b-bhpnct) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L569)):*
> ```cpp
> }
> 
> void RenderEngineGltfClient::DoUpdateVisualPose(
> ```

BTW/question. To confirm my understanding, we talked about having multiple visuals for a model. In `RenderEngineGltfClient`'s perspective, a mesh or convex type will be either a .gltf or .obj, so `DoUpdateVisualPose() or DoRemoveGeometry()` will only need to handle one of them. 



_Code quote:_
```cpp
void RenderEngineGltfClient::DoUpdateVisualPose(
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 602 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6MeBo-3V4HFzzTW01:-NZ6MeBo-3V4HFzzTW02:b-3sj0kk) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L602)):*
> ```cpp
>     const std::filesystem::path& mesh_path, double scale, void* user_data) {
>   std::string extension = mesh_path.extension();
>   std::transform(extension.begin(), extension.end(), extension.begin(),
> ```

BTW. Wow I thought we can simply do `new_std_string = old_std_string.tolower();` and be done with it. :sweat_smile: 



_Code quote:_
```cpp
std::transform(extension.begin(), extension.end(), extension.begin(),
```

zachfang (commented):
Another checkpoint due to some Github disruption. (Almost done.)

> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ7LQR1EAbgZrSB2A5L:b-nf8xeo)* status: 16 unresolved discussions, LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 195 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ73n4E8XB50RqB7unC:-NZ73n4E8XB50RqB7unD:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L195)):*
> ```cpp
>           if (node.contains("translation")) {
>             const auto& t = node["translation"];
>             Vector3<double> p_GN(t[0].get<double>(), t[1].get<double>(),
> ```

nit.



_Suggestion:_
```cpp
const Vector3<double> p_GN
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ73r-j3JMFJsU6Qbif:-NZ73r-j3JMFJsU6Qbig:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L202)):*
> ```cpp
>             const auto& q = node["rotation"];
>             // glTF rotation is (x, y, z, w) quaternion. Eigen is (w, x, y, z).
>             Quaternion<double> quat_GN(q[3].get<double>(), q[0].get<double>(),
> ```

nit.



_Suggestion:_
```cpp
const Quaternion<double>
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 284 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7C0XlB_E9ONNmNeQx:-NZ7C0Xm51sEWn7wEB4n:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L284)):*
> ```cpp
> // renders shouldn't transmit any of this otherwise useless data.
> 
> /* Change all material definitions to be an emissive flat color. This removes
> ```

nit.



_Suggestion:_
```cpp
Changes
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 95 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7GT1O4WNyh4LBkmV-:-NZ7GT1O4WNyh4LBkmV0:b-hqzd75) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L95)):*
> ```cpp
> 
> TEST_F(RenderEngineGltfClientTest, Constructor) {
>   // Make sure that default values are passed to the underlying client.
> ```

BTW. Maybe we can move this comment down to be adjacent to:
```
  const RenderEngineGltfClient default_engine{default_params};
  EXPECT_EQ(default_engine.get_params().GetUrl(), default_params.GetUrl());
  EXPECT_EQ(default_engine.get_params().verbose, default_params.verbose);
  EXPECT_EQ(default_engine.get_params().cleanup, default_params.cleanup);
```



_Code quote:_
```cpp
// Make sure that default values are passed to the underlying client.
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 99 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7Fh2379WkaRq6sKHB:-NZ7Fh2379WkaRq6sKHC:b-jvh3js) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L99)):*
> ```cpp
>   // We want some test params completely different from default, but
>   // params_.cleanup matches default, but we don't want cleanup disabled
>   // throughout the text.
> ```

nit. the test(s)?



_Code quote:_
```cpp
the text
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 359 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7IQ8I1K8PGmJdNOjA:-NZ7IQ8I1K8PGmJdNOjB:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L359)):*
> ```cpp
>   /* Exports the glTF file from the engine, and reads it back as json. */
>   json ExportAndReadJson(std::string_view file_name, ImageType image_type) {
>     std::filesystem::path file_path = temp_dir_ / file_name;
> ```

nit.



_Suggestion:_
```cpp
const std::filesystem::path
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 393 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7JLF92OTDa1KO0paE:-NZ7JLF92OTDa1KO0paF:b-tpdff) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L393)):*
> ```cpp
>                                       Convex("some.stl"), properties_, X_WG_));
> 
>   // Objs and glTfs are both accepted.
> ```

nit. `glTFs` or `gltfs`.



_Code quote:_
```cpp
glTfs
```

zachfang (commented):
:lgtm: Mostly small stuff.

> Reviewed 1 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ7Vj881TO3YjbyWloX:birvnis)* status: 18 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 5 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7VQWrCYvtSM9snDb6:-NZ7VQWrCYvtSM9snDb7:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L5)):*
> ```cpp
> #include <filesystem>
> #include <fstream>
> #include <set>
> ```

nit.



_Suggestion:_
```cpp
#include <map>
#include <set>
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 442 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7NL5W1j4DR08LWZlT:-NZ7NL5W1j4DR08LWZlU:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L442)):*
> ```cpp
>   const GeometryId gltf_id = AddGltf();
> 
>   /* Given the pose of a node */
> ```

nit.



_Suggestion:_
```cpp
a node.
```

jwnimmer-tri (commented):
> Reviewed 2 of 11 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 23 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.h` line 118 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7pA304WSA7ssQsrde:-NZ7pA318s2z8yyEQYb3:b-35lbgx) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.h#L118)):*
> ```objc
>     std::map<int, Matrix4<double>> root_nodes;
>     // The isotropic scale of the mesh.
>     double scale;
> ```

nit GSG



_Suggestion:_
```objc
double scale{};
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZ7qKzLAAZ08L5_DlwI:b-h4ak4t) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*
> ```cpp
>   // https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-extras)
>   for (nlohmann::json& node : gltf["nodes"]) {
>     node["extras"]["from_drake"] = true;
> ```

How about we don't do this until we have a specific use case that demands it?

It's more work to implement and test, chips away at our GPL firewall, and does not appear to me to be useful in any way.

(Admittedly, I haven't read the unit testing yet.  Possibly the unit testing relies on this somehow?)
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 566 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qZ_r1rjMEf3W2QjT:-NZ7qZ_s4FJ_rX7IeLli:bn8dd7z) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L566)):*
> ```cpp
> 
>   std::ofstream f(export_path);
>   f << gltf;
> ```

If I recall correctly, by default `ofstream` does not throw upon write errors.  If the disk is full (for example), I don't think this will warn the user.  We should throw in case the stream is bad.
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 186 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7oMSa5KkHXF2m6Sdl:-NZ7oMSa5KkHXF2m6Sdm:b7ipwiz) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_vtk/internal_render_engine_vtk.h#L186)):*
> ```objc
>   RenderingPipeline& get_mutable_pipeline(ImageType image_type) const;
> 
>   /* A package of data required to register a visual geometry. */
> ```

nit It seems worth a mention that this is the type of `user_data` in the overrides below?
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 225 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7o1IL216EzIlojLXq:-NZ7o1IL216EzIlojLXr:b-efsrsk) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_vtk/internal_render_engine_vtk.h#L225)):*
> ```objc
> 
>   // Initializes the VTK pipelines.
>   void InitializePipelines();
> ```

nit I am unconvinced that `InitializePipelines` should be exposed to subclasses as `protected`.

SeanCurtis-TRI (commented):
Thanks.

@sherm1 for platform review (on next TRI workday.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ8elWO1_grqE95WOB3:b-7giy6v)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6QIli-iOlUW14YeyF:-NZ8UXIa4UpsdR7Bw79x:b-pheieo) (no related file):*

BTW. I wonder if adding a `from_drake: true/false` tag would make the server/client relationship too "intimate" or the glTF too "customized"?
</blockquote></details>

Those are some interesting thoughts. I don't have ready answers.

1. I considered swapping `from_drake` to `from_gltf_client`. But, really, `gltf_client` would just be an alias to drake and doesn't really provide much more distance beyond `from_drake`.
2. As glTF was *designed* to support extensible data (the "extras"), I don't think exercise of those attributes (which don't make the glTF any less compliant) increases the "intimacy".
3. Up to this PR, our logic for determining what came via the gltf client is based on VTK behavior. That seemed overly rigid and brittle. However, if we feel that's the safest solution, I can massage this code to conform to that rubric.

@jwnimmer-tri You have any thoughts?

--------------------------------

Between typing the above, responding to reviewer comments, and publishing changes, @jwnimmer-tri had already weighed in below. I've also provided my justification below.

___
___
*[`geometry/render_gltf_client/BUILD.bazel` line 96 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6UDNC0Kai5GfegTiT:-NZ8VhdW7GJHXtHTPPV7:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/BUILD.bazel#L96)):*

nit. `internal_render_client.{cc, h}` are not touched in this PR, I assume the added dependencies (and `:nlohmann`) should be moved down to `internal_render_engine_gltf_client`?
</blockquote></details>

Done
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZ8cQ-51gdAF_OTA6iW:b-1jbd49) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*

How about we don't do this until we have a specific use case that demands it?

It's more work to implement and test, chips away at our GPL firewall, and does not appear to me to be useful in any way.

(Admittedly, I haven't read the unit testing yet.  Possibly the unit testing relies on this somehow?)
</blockquote></details>

There is a specific use case.

It *vastly* simplifies the merging code here and the server code in `drake-blender`.

Current state of affairs

1. VTK creates a "Render Node" node that it stashes all of its geometry in.
    - `MultibodyPlant` doesn't register hierarchies of nodes; everything is in the world. So, literally everything that gets added is a child of the world which maps to a child of "Render Node" in the resultant glTF.
2. `server.py` handles label images by tweaking those nodes that are immediate children of "Render Node".

Realities of glTFs

1. glTFs can be built of arbitrary hierarchies.
2. Those hierarchies are preserved in the merging process.
2. Merging gltfs puts the root nodes of a glTF file as root nodes in the scene (that would be a sibling of "Render Node" and not a child).

State of affairs *with* this tag
1. Merging is straightforward and simple.
2. `server.py` changes its test from `bpy_object.parent.name == "Renderer Node"` to (essentially) `bpy_object.get("from_drake") is not None`.

Without this tag (or some alternately named tag with the same semantics)
1. Merging becomes more arduous and indirect, depending on VTK logic and *not* Drake logic.
2. `server.py` has to look for nodes whose *ancestor* is "Render Node" instead of just its parent.

So, for all of those reasons, I added a tag. If intimacy is a concern, I'd most strongly advocate changing its name. But it is the case that the mere presence of an "extras" tag in a glTF file suggests the existence of an entity that created it. (Is this the blind watchmaker?)
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 569 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6X8hD892yK9K2MgdR:-NZ8Wj-WBqCzNs6YLaaM:bhb7x4w) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L569)):*

BTW/question. To confirm my understanding, we talked about having multiple visuals for a model. In `RenderEngineGltfClient`'s perspective, a mesh or convex type will be either a .gltf or .obj, so `DoUpdateVisualPose() or DoRemoveGeometry()` will only need to handle one of them. 
</blockquote></details>

Correct. If it's a `glTF` it is managed by this `RenderEngine`, otherwise it defers to the parent class.
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 602 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ6MeBo-3V4HFzzTW01:-NZ8WNsJASyinlLCMDoQ:b-z67qbk) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L602)):*

BTW. Wow I thought we can simply do `new_std_string = old_std_string.tolower();` and be done with it. :sweat_smile: 
</blockquote></details>

I know, right?

In the follow-up PR, `geometry::Mesh` gets a new method called `extension()`. So, all of this extension extraction will disappear into there. But for now, the ugliness is copy-pasta'd here.
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 95 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7GT1O4WNyh4LBkmV-:-NZ8eA_w6PjDPkE5SqWG:byvxtnh) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L95)):*

BTW. Maybe we can move this comment down to be adjacent to:
```
  const RenderEngineGltfClient default_engine{default_params};
  EXPECT_EQ(default_engine.get_params().GetUrl(), default_params.GetUrl());
  EXPECT_EQ(default_engine.get_params().verbose, default_params.verbose);
  EXPECT_EQ(default_engine.get_params().cleanup, default_params.cleanup);
```
</blockquote></details>

I've changed the documentation.
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 225 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7o1IL216EzIlojLXq:-NZ8eS2n3OcWSU3D43zm:b-ht801f) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_vtk/internal_render_engine_vtk.h#L225)):*

nit I am unconvinced that `InitializePipelines` should be exposed to subclasses as `protected`.
</blockquote></details>

Agreed. I just wanted to avoid re-ordering.

SeanCurtis-TRI (commented):
(BTW @jwnimmer-tri, as you've been looking already, would you be willing to make an out-of-order platform review?)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ8jt2u0VH-qy8wYjjB:b-nc2y7q)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

zachfang (commented):
> Reviewed 3 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 602 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ8nGu47RHc1WrFxfmL:-NZ8nGu47RHc1WrFxfmM:b-rm7ae1) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac21be68ae022951d98b6af91b4ed7f8f0e12b/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L602)):*
> ```cpp
> }
> 
>   /* As with registering geometry, we should be able to remove both kinds of
> ```

BTW. These indents are probably unintentional.

jwnimmer-tri (commented):
> Reviewed 4 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZ8kj2QBn7HunkciVbR:bclycx5) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*

There is a specific use case.

It *vastly* simplifies the merging code here and the server code in `drake-blender`.

Current state of affairs

1. VTK creates a "Render Node" node that it stashes all of its geometry in.
    - `MultibodyPlant` doesn't register hierarchies of nodes; everything is in the world. So, literally everything that gets added is a child of the world which maps to a child of "Render Node" in the resultant glTF.
2. `server.py` handles label images by tweaking those nodes that are immediate children of "Render Node".

Realities of glTFs

1. glTFs can be built of arbitrary hierarchies.
2. Those hierarchies are preserved in the merging process.
2. Merging gltfs puts the root nodes of a glTF file as root nodes in the scene (that would be a sibling of "Render Node" and not a child).

State of affairs *with* this tag
1. Merging is straightforward and simple.
2. `server.py` changes its test from `bpy_object.parent.name == "Renderer Node"` to (essentially) `bpy_object.get("from_drake") is not None`.

Without this tag (or some alternately named tag with the same semantics)
1. Merging becomes more arduous and indirect, depending on VTK logic and *not* Drake logic.
2. `server.py` has to look for nodes whose *ancestor* is "Render Node" instead of just its parent.

So, for all of those reasons, I added a tag. If intimacy is a concern, I'd most strongly advocate changing its name. But it is the case that the mere presence of an "extras" tag in a glTF file suggests the existence of an entity that created it. (Is this the blind watchmaker?)
</blockquote></details>

Okay, I see.  Let me try to restate to confirm I got it right.

When rendering a label image, a server needs to know what geometry came in from the command line versus over the wire, in order to paint the command-line geometry with the "background" label.

To decide which geometry was which, the drake-blender code was relying on an implementation detail of the client, beyond what was promised by the protocol.

When we land this PR (either with or without the `from_drake` change) and a user adds glTF meshes to their scene graph, the current drake-blender server code will render incorrect label images.  No matter what, we'll need to adjust the server code to make it work correctly again.

That leads to two questions:

(1) Do we need to fix the drake-blender code and then tag a new drake-blender release, prior to landing this PR?

If we don't take care with release sequencing, then users who try to use this feature (glTF meshes) for label images might experience the drake-blender bug.

My first choice would be to sequence the releases such that users who consistently use the latest released version numbers have no risk of hitting the bug.

My second choice would be to fail-fast (e.g., throw an exception) in case the label image might be buggy, and work to fix bug in the next release(s).  For example, the client could refuse to make label rendering requests iff there was any merged gltf geometry.

(2) How should we fix the drake-blender server code?

I don't think having the client spam a magic tag in the gltf-on-the-wire is a good answer, even setting aside the legal firewall questions.  At minimum, if we decide to do that, then we need to update the protocol specification doxygen to promise it, and mark it as "protocol version 2" on the wire.

If the server is side-loading additional geometry, I think its the job of the server to keep track of it. Is there any way it the server could load the blend file, then tag all of _that_ geometry (recursively) as "will be background label", and _then_ load the wire-gltf data secondarily?  I have to think that one way or another, the client should be able to send vanilla gltf and the server should accept vanilla gltf, and everyone is happy.  Just because someone made a mistake in the early days of drake-blender for how geometry is disambiguated does not seem like a good reason to torque the protocol.

jwnimmer-tri (commented):
(If I end up with time, I'll do platform review, but I'm not sure yet about my holiday schedule.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZ93KIy77-5SZulYUZe:bycc4y1)* status: 3 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZ981CY2pury9uAfy5z:b-pbl65s) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*

Okay, I see.  Let me try to restate to confirm I got it right.

When rendering a label image, a server needs to know what geometry came in from the command line versus over the wire, in order to paint the command-line geometry with the "background" label.

To decide which geometry was which, the drake-blender code was relying on an implementation detail of the client, beyond what was promised by the protocol.

When we land this PR (either with or without the `from_drake` change) and a user adds glTF meshes to their scene graph, the current drake-blender server code will render incorrect label images.  No matter what, we'll need to adjust the server code to make it work correctly again.

That leads to two questions:

(1) Do we need to fix the drake-blender code and then tag a new drake-blender release, prior to landing this PR?

If we don't take care with release sequencing, then users who try to use this feature (glTF meshes) for label images might experience the drake-blender bug.

My first choice would be to sequence the releases such that users who consistently use the latest released version numbers have no risk of hitting the bug.

My second choice would be to fail-fast (e.g., throw an exception) in case the label image might be buggy, and work to fix bug in the next release(s).  For example, the client could refuse to make label rendering requests iff there was any merged gltf geometry.

(2) How should we fix the drake-blender server code?

I don't think having the client spam a magic tag in the gltf-on-the-wire is a good answer, even setting aside the legal firewall questions.  At minimum, if we decide to do that, then we need to update the protocol specification doxygen to promise it, and mark it as "protocol version 2" on the wire.

If the server is side-loading additional geometry, I think its the job of the server to keep track of it. Is there any way it the server could load the blend file, then tag all of _that_ geometry (recursively) as "will be background label", and _then_ load the wire-gltf data secondarily?  I have to think that one way or another, the client should be able to send vanilla gltf and the server should accept vanilla gltf, and everyone is happy.  Just because someone made a mistake in the early days of drake-blender for how geometry is disambiguated does not seem like a good reason to torque the protocol.
</blockquote></details>

Actually, I think I just had an epiphany on what we can do on the server side to trivialize everything. Here's the sketch while I work out the details:

1. Server creates a custom collection.
2. gltf gets imported *into* this collection.
3. When doing label stuff, uses membership of the collection as being indicative of whether it came from the gltf client or not.

That divorces the server from any knowledge of anything encoded in the glTF. I'll get back to you on this.
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 602 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ8nGu47RHc1WrFxfmL:-NZ98hQlCW0pL_ccLkf-:b-kr3kwt) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac21be68ae022951d98b6af91b4ed7f8f0e12b/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L602)):*

BTW. These indents are probably unintentional. 
</blockquote></details>

Nope. Intentional and, as far as I can tell, consistent with the rest of the file. (At least where `/*` is used in place of `//`. Although the file is currently a mix of `//` and `/* */`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZ9Djz247mtdunf0-gE:bjt4ibb) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*

Actually, I think I just had an epiphany on what we can do on the server side to trivialize everything. Here's the sketch while I work out the details:

1. Server creates a custom collection.
2. gltf gets imported *into* this collection.
3. When doing label stuff, uses membership of the collection as being indicative of whether it came from the gltf client or not.

That divorces the server from any knowledge of anything encoded in the glTF. I'll get back to you on this.
</blockquote></details>

That sounds great. Anything along the lines of "the server keeps tabs on where things came from" would work for me.

zachfang (commented):
> Reviewed 1 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 602 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ8nGu47RHc1WrFxfmL:-NZ9JbOeBJl4nxtuimhY:bbdriz4) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac21be68ae022951d98b6af91b4ed7f8f0e12b/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L602)):*

Nope. Intentional and, as far as I can tell, consistent with the rest of the file. (At least where `/*` is used in place of `//`. Although the file is currently a mix of `//` and `/* */`.
</blockquote></details>

I should be more explicit. I meant the two-space indents for this `TEST_F` added in r3.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 556 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ7qKzLAAZ08L5_DlwH:-NZA8hRXCHUxvW88Qelz:bi9uuqp) ([raw file](https://github.com/RobotLocomotion/drake/blob/02692a289ee3e7bb634b591e98842e42a85862e4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L556)):*

That sounds great. Anything along the lines of "the server keeps tabs on where things came from" would work for me.
</blockquote></details>

Ok, the plan worked. I've removed the tag.
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 602 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZ8nGu47RHc1WrFxfmL:-NZA8qcM7zCflMSkZYrF:b-8agds0) ([raw file](https://github.com/RobotLocomotion/drake/blob/efac21be68ae022951d98b6af91b4ed7f8f0e12b/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L602)):*

I should be more explicit. I meant the two-space indents for this `TEST_F` added in r3.
</blockquote></details>

D'oh!

SeanCurtis-TRI (commented):
+(status: do not merge) I'm going to submit the PR to `drake-blender`.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZA9WZw977OAAtBVrNa:blrywoz)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge" (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI:
Waiting on https://github.com/RobotLocomotion/drake-blender/pull/53 to merge.

jwnimmer-tri:
+@jwnimmer-tri for platform review.

jwnimmer-tri (commented):
> Reviewed 1 of 11 files at r1, 3 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: LGTM missing from assignee jwnimmer-tri(platform), labeled "do not merge" (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Reviewed 1 of 11 files at r1, 1 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 8 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 161 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCD0mc57EISw7Ij0pI:-NZCD0mc57EISw7Ij0pJ:bbcxexy) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L161)):*
> ```cpp
> 
> /* Searches the gltf and identifies all *mesh* root nodes: nodes that are not
>   referenced in a node's "children" list but do have the "mesh" attribute. */
> ```

nit It would help to document the semantics of the return value.  I can guess that the key is a glTF index, but I can't really guess what the value is (other than some kind of transform or transform&scale; the question is _which_ transform (frames)).
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 168 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCDTwe2UyZCeG6xX9k:-NZCDTwe2UyZCeG6xX9l:b5fv3q5) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L168)):*
> ```cpp
>     // Cull children.
>     const nlohmann::json& nodes = gltf["nodes"];
>     const int N = ssize(nodes);
> ```

nit I'm not sure that `N` capital letter meets GSG?  (It's not a matrix.)
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 169 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCEOc6AmnNiFZdKkcf:-NZCEOc76U3QSAcoOF_2:b-e6u39f) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L169)):*
> ```cpp
>     const nlohmann::json& nodes = gltf["nodes"];
>     const int N = ssize(nodes);
>     for (int i = 0; i < N; ++i) indices.insert(i);
> ```

BTW Passing the `indices.end()` hint here might help with performance?
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 215 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCGkJNDjkzDJXcNoam:-NZCGkJNDjkzDJXcNoan:b-7iknvd) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L215)):*
> ```cpp
>           }
>         }
>         roots[n] = X_FN;
> ```

BTW Passing the `roots.end()` hint here might help with performance?
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 234 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCGz_p3ECVdFYur-zw:-NZCGz_p3ECVdFYur-zx:b-bbqfw0) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L234)):*
> ```cpp
>                      properties from the targeted nodes. */
> void SetRootPoses(nlohmann::json* gltf,
>                   const std::map<int, Matrix4<double>> target_nodes,
> ```

nit Accidental copy?



_Suggestion:_
```cpp
const std::map<int, Matrix4<double>>& target_nodes,
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 236 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCHDIJ6oKmLfUZDWPr:-NZCHDIJ6oKmLfUZDWPs:b3qkx92) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L236)):*
> ```cpp
>                   const std::map<int, Matrix4<double>> target_nodes,
>                   const math::RigidTransformd& X_WG, double scale,
>                   bool strip = true) {
> ```

nit The default value is never used.  (Also for file-local functions, IMO typically defaults do more harm than good.)



_Suggestion:_
```cpp
bool strip
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 268 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCIIMp8N2MTeXGTpDL:-NZCIIMp8N2MTeXGTpDM:buxqd9f) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L268)):*
> ```cpp
>   if (gltf->contains("nodes")) {
>     nlohmann::json& nodes = (*gltf)["nodes"];
>     const int N = ssize(nodes);
> ```

nit I'm not sure that `N` capital letter meets GSG.
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 570 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZCLv_f0E3a9Fak9pY0:-NZCLv_f0E3a9Fak9pY1:b-t22yy4) ([raw file](https://github.com/RobotLocomotion/drake/blob/3281a0fea697b4a5e6310903f06fff78072be9ec/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L570)):*
> ```cpp
>     GeometryId id, const math::RigidTransformd& X_WG) {
>   if (gltfs_.count(id) == 1) {
>     GltfRecord& gltf = gltfs_[id];
> ```

BTW This is doing a second map lookup.  It would probably be a little bit faster to use `find` instead of `count`.

SeanCurtis-TRI (commented):
Thanks.  I've addressed your comments in the .cc file. Enjoy your holiday weekend.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZP288ZC8dCDuNiYv2j:b-1tptr8)* status: LGTM missing from assignee jwnimmer-tri(platform), labeled "do not merge" (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI:
@drake-jenkins-bot retest mac-arm-ventura-clang-bazel-experimental-release please.

jwnimmer-tri (commented):
:lgtm: platform.

It seems to me like this PR should be noted in the notes: -(release notes: none) +(release notes: fix).

> Reviewed 1 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZaTDBn9vg4oFRt0r8E:b3c3d4u)* status: 5 unresolved discussions, labeled "do not merge" (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaW4M_COlJdkgjH9lg:-NZaW4MaBEcSjcHEqOQb:b-d8kz83) (no related file):*
Working

If I have time, I'm going to try some manual end-to-end testing of some rendered images.

___
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 376 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZT8YYfE2eEPjO5vDOv:-NZT8YYfE2eEPjO5vDOw:b-hzq4gb) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L376)):*
> ```cpp
>   const render::RenderLabel label_;
>   const double scale_ = 2.0;
>   std::filesystem::path temp_dir_;
> ```

nit It seems like other unvarying test-fixture members are `const`?



_Suggestion:_
```cpp
const std::filesystem::path temp_dir_;
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 427 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaU-Iv3soeUMAeZHHj:-NZaU-Iv3soeUMAeZHHk:b1jfx4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L427)):*
> ```cpp
> /* Currently, our server expects us to output a z-up gltf (because a) Drake is
>  z-up and b) VTK does not correct it). The input gltf files are y-up. Therefore,
>  we have to do some special math to apply a z-up pose to a y-up object.
> ```

nit The parenthetical groupings here is too easy to misread.



_Suggestion:_
```cpp
/* Currently, our server expects us to output a z-up gltf (because (a) Drake is
 z-up and (b) VTK does not correct it). The input gltf files are y-up. Therefore,
 we have to do some special math to apply a z-up pose to a y-up object.
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 449 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaUv5e1fIRT9MTJK-r:-NZaUv5e1fIRT9MTJK-s:b-2lpjoi) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L449)):*
> ```cpp
>   const GeometryId gltf_id = AddGltf();
> 
>   /* Given the non-rigid transform of a node (X_WN), the rigid pose of the
> ```

BTW Would it be better to avoid `X` for transforms that are non-rigid, like we did in the implementation?



_Suggestion:_
```cpp
T_WN
```
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 475 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaVbXADYKEimVvvP7G:-NZaVbXADYKEimVvvP7H:b-qet5j0) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L475)):*
> ```cpp
>     EXPECT_TRUE(node.contains("matrix"));
>     const json& matrix = node["matrix"];
>     // For glTF, transform matrix is a *column-major* matrix.
> ```

BTW Possibly something like `internal_merge_gltf.h` should offer a json-to-mat4x4 helper function?  This is the third time we have this pattern copy-pasted.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 4 unresolved discussions, labeled "do not merge" (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaW4M_COlJdkgjH9lg:-NZaoRWt0pVaHQCOeLnM:b-v0ssx4) (no related file):*

Working

If I have time, I'm going to try some manual end-to-end testing of some rendered images.
</blockquote></details>

Hmm, not enough time.  We can continue to improve if we find integration issues later on.

___

sherm1 (commented):
BTW this PR has two lgtms but is marked "do not merge"

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZax6bsAXsXCNWlsouo:b-akp65e)* status: 4 unresolved discussions, labeled "do not merge" (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
We need the other PR to land before we can land this one:

> Waiting on https://github.com/RobotLocomotion/drake-blender/pull/53 to merge.

SeanCurtis-TRI (commented):
@jwnimmer-tri Here's my reasoning for why I put release notes none on this:

1. At best, this feels like a feature, not a fix. We're adding capabilities, not correcting existing ones.
2. Just because this `RenderEngine` implementation has this added capability under the hood, trying to exercise it through any normal mechanism (parsing, scene graph, etc.) will undoubtedly lead to landmines. True, it can be exercised, but someone would have to be using the engine outside of the normal context.
3. That's why my intention was to tag the *final* PR with `feature` and call it "Drake initial support for glTF files" (or some such thing). That PR is what makes the system sufficiently "defined" across the board such that we can predict what will happen if you specify a .gltf in an sdf/urdf file. It would also highlight the new capabilities. 

Thoughts?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZc-2G69Q_nbUZIS2ep:b-r18p24)* status: 1 unresolved discussion, labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 449 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaUv5e1fIRT9MTJK-r:-NZcB-C7AOPqa2el2Wjf:b-a2uliw) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L449)):*

BTW Would it be better to avoid `X` for transforms that are non-rigid, like we did in the implementation?
</blockquote></details>

I had done a better job in the actual implementation of cleaning up this nomenclature.

I took a pass at this code to (slight) unify it. At the very least, I'm no longer using `X` to mean a non-`RigidTransform`.
___
*[`geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc` line 475 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZaVbXADYKEimVvvP7G:-NZcBLhk9R4zhtHDxoWP:b-pbdqz3) ([raw file](https://github.com/RobotLocomotion/drake/blob/7a1c8c90878858fd41665babca739292f24836f7/geometry/render_gltf_client/test/internal_render_engine_gltf_client_test.cc#L475)):*

BTW Possibly something like `internal_merge_gltf.h` should offer a json-to-mat4x4 helper function?  This is the third time we have this pattern copy-pasted.
</blockquote></details>

Agreed.

SeanCurtis-TRI (commented):
-(status: do not merge) 

The `drake-blender` PR has now merged.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZccjFa0j2UheIcjzba:bpjcfom)* status: 1 unresolved discussion, labeled "do not merge" (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
With the changes in R6 and R7, I'm deferring merging until reviewers have a chance to weigh in.

+(status: do not merge) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZcu5FQEAEw7JotIaO_:bnqfde1)* status: labeled "do not merge" (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Reviewed 6 of 6 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: labeled "do not merge" (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
I'll defer to you on what label you want here.

> At best, this feels like a feature, not a fix. We're adding capabilities, not correcting existing ones.

Using "notes: feature" is OK by me, and probably better.  My main thesis was that it's notable, I don't really mind which one it is. I choose the nearest neighbor to "notes: none".

> Just because this RenderEngine implementation has this added capability under the hood, trying to exercise it through any normal mechanism (parsing, scene graph, etc.) will undoubtedly lead to landmines.

This doesn't register with me -- I don't see anything here that is either dangerous or incomplete.  Sure, it won't show up in MeshcatVisualizer (modulo other PRs) but that's not relevant for rendering.  (If there _is_ something dangerous here, we should repair that before merging to master anyway.)

My opinion is that significant changes to how shipped code operates should always be noted.  If we reach the release day and the feature is only halfway complete, we can adjust the notes document at that point with appropriate caveats or cautions.

Or put another way -- once we do "notes: none", the opportunity to reflect on what's notable in the release is essentially lost.  On the other hand, if we use a fix/feature tag then we have the opportunity to revisit the question again once we have all of the information about what is or is not part of a release.

zachfang (commented):
Added some feel-free-to-ignore FYIs. Everything looks good to me. 

> Reviewed 3 of 5 files at r4, 2 of 6 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZgAtqa6qOENP5TzhA7:bgd5ro9)* status: labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 221 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZg7tKk9Q35KDqJdd6C:-NZg7tKlDte10ReMzPKr:b-30fk1g) ([raw file](https://github.com/RobotLocomotion/drake/blob/79b4e9d668ca4d7ea6bb3bb10c2dc8e647e66555/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L221)):*
> ```cpp
> /* Effectively sets the world pose of the single "geometry" represented by one
>  or more nodes defined in a glTF file. It achieves this by setting the
>  transforms on the `root_nodes` of the glTF. Applying the pose to non-root
> ```

FYI. Adding this sentence about applying the pose to `non-root` nodes is a bit distracting (at least for me).

My understanding is we change the `T_FN` to `T_WN` for the root nodes and keep the transforms of their child nodes intact because everything is relative to their root node. Maybe adding that would connect better with the first part, i,e., `sets the world pose of the single"geometry" ...`? (Feel free to ignore.)  



_Code quote:_
```cpp
Applying the pose to non-root
```
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 262 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZg7-q22Y6J6nIKuMvt:-NZg7-q22Y6J6nIKuMvu:bb6vfrx) ([raw file](https://github.com/RobotLocomotion/drake/blob/79b4e9d668ca4d7ea6bb3bb10c2dc8e647e66555/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L262)):*
> ```cpp
>   Matrix4<double> T_WF = X_WG.GetAsMatrix4();
>   T_WF.block<3, 3>(0, 0) *= scale;  // S_WG.
>   math::RigidTransformd X_GF(math::RotationMatrixd::MakeXRotation(M_PI / 2));
> ```

FYI.



_Suggestion:_
```cpp
const math::RigidTransformd X_GF
```
___
*[`geometry/render_gltf_client/test/internal_merge_gltf_test.cc` line 31 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZgAGql7mBjgIPlroZA:-NZgAGql7mBjgIPlroZB:bb6vfrx) ([raw file](https://github.com/RobotLocomotion/drake/blob/79b4e9d668ca4d7ea6bb3bb10c2dc8e647e66555/geometry/render_gltf_client/test/internal_merge_gltf_test.cc#L31)):*
> ```cpp
>   Matrix4<double> M;
>   M << 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16;
>   json M_json = GltfMatrixFromEigenMatrix(M);
> ```

FYI.



_Suggestion:_
```cpp
const json M_json
```

SeanCurtis-TRI (commented):
> I don't see anything here that is either dangerous or incomplete.

I believe the primary problem is that if you've got either of the other two `RenderEngine` types and a .gltf file in a `<visual>` tag things will blow up in inexplicable ways. Although, given that it's no worse than passing in a .dae file in the same tag, it's not the end of the world. What's *unique* in this case is declaring the feature on this PR is tantamount to announcing we support gltTF. Nowhere do we imply support of Collada.

> significant changes to how shipped code operates should always be noted.

I agree. I was just proposing they be noted in conjunction with the terminal PR in the sequence and not an intermediate PR. I view the whole sequence as one big feature PR that has to be broken up to be reviewable.

That said, I don't care enough to change the tag back.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-:-NZh5rIO2pKdkakSAsIL:b-m24751)* status: labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 221 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZg7tKk9Q35KDqJdd6C:-NZh8z7m0eG_r1o-ya7n:b-eq0h5l) ([raw file](https://github.com/RobotLocomotion/drake/blob/79b4e9d668ca4d7ea6bb3bb10c2dc8e647e66555/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L221)):*

FYI. Adding this sentence about applying the pose to `non-root` nodes is a bit distracting (at least for me).

My understanding is we change the `T_FN` to `T_WN` for the root nodes and keep the transforms of their child nodes intact because everything is relative to their root node. Maybe adding that would connect better with the first part, i,e., `sets the world pose of the single"geometry" ...`? (Feel free to ignore.)  
</blockquote></details>

I tried strengthening the statement. PTAL.

zachfang (commented):
> Reviewed 1 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: 1 unresolved discussion, labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 221 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZg7tKk9Q35KDqJdd6C:-NZhB4HW3RPFqEmBswXQ:b9j9n43) ([raw file](https://github.com/RobotLocomotion/drake/blob/79b4e9d668ca4d7ea6bb3bb10c2dc8e647e66555/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L221)):*

I tried strengthening the statement. PTAL.
</blockquote></details>

Looks good to me.
___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 222 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZhArv02njyvs4JGKew:-NZhArv02njyvs4JGKex:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/c64c67977288c9b6f96874f5ce12d3c0061dc5b4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L222)):*
> ```cpp
>  or more nodes defined in a glTF file. It achieves this by setting the
>  transforms on the `root_nodes` of the glTF and *only* the root nodes. All other
>  nodes should maintain their fixed poses *relative* ro their root nodes. It's
> ```

nit.



_Suggestion:_
```cpp
to
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19684)* status: labeled "do not merge" (waiting on @SeanCurtis-TRI)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 222 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19684#-NZhArv02njyvs4JGKew:-NZhB_kk9G1j2wUV3zPU:b-o2plka) ([raw file](https://github.com/RobotLocomotion/drake/blob/c64c67977288c9b6f96874f5ce12d3c0061dc5b4/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L222)):*

nit.
</blockquote></details>

Ruh roh!

SeanCurtis-TRI:
-(status: do not merge)

SeanCurtis-TRI:
BTW It used to be really easy to tell if reviewers had stamped `lgtm`. Now they're lost in the conversation. I think I'll have to post an issue with Piotr.

PR #19685, opened 2023-06-27
[bindings] Add env var DRAKE_PYTHON_LOGGING to disable redirect

### LINKED ISSUE #19680: [pydrake] Disable spdlog redirection when using multithreaded C++ code?

EricCousineau-TRI:
Resolves #19680

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)

### DIFF
--- a/bindings/pydrake/common/BUILD.bazel
+++ b/bindings/pydrake/common/BUILD.bazel
@@ -493,13 +493,16 @@ drake_py_unittest(
     ],
 )
 
+# We make this importable outside of pydrake so we can configure spdlog
+# before pydrake's import logic.
 drake_pybind_library(
-    name = "text_logging_test_py",
+    name = "text_logging_test_helpers_py",
     testonly = 1,
     add_install = False,
-    cc_so_name = "test/text_logging_test",
-    cc_srcs = ["test/text_logging_test_py.cc"],
+    cc_so_name = "test/text_logging_test_helpers",
+    cc_srcs = ["test/text_logging_test_helpers_py.cc"],
     package_info = PACKAGE_INFO,
+    py_imports = ["test"],
     visibility = ["//visibility:private"],
 )
 
@@ -509,7 +512,7 @@ drake_py_binary(
     srcs = ["test/text_logging_example.py"],
     deps = [
         ":module_py",
-        ":text_logging_test_py",
+        ":text_logging_test_helpers_py",
     ],
 )
 
--- a/bindings/pydrake/common/BUILD.bazel
+++ b/bindings/pydrake/common/BUILD.bazel
@@ -493,8 +493,6 @@ drake_py_unittest(
     ],
 )
 
-# We make this importable outside of pydrake so we can configure spdlog
-# before pydrake's import logic.
 drake_pybind_library(
     name = "text_logging_test_helpers_py",
     testonly = 1,
@@ -502,7 +500,6 @@ drake_pybind_library(
     cc_so_name = "test/text_logging_test_helpers",
     cc_srcs = ["test/text_logging_test_helpers_py.cc"],
     package_info = PACKAGE_INFO,
-    py_imports = ["test"],
     visibility = ["//visibility:private"],
 )
 
@@ -534,7 +531,7 @@ drake_py_unittest(
     ],
     deps = [
         ":module_py",
-        ":text_logging_test_py",
+        ":text_logging_test_helpers_py",
     ],
 )
 
--- a/bindings/pydrake/common/test/text_logging_example.py
+++ b/bindings/pydrake/common/test/text_logging_example.py
@@ -1,14 +1,8 @@
 import argparse
 import logging
-import os
-
-from text_logging_test_helpers import do_log_test, set_log_level
-if "_TEST_SPDLOG_LEVEL" in os.environ:
-    # Configure logging before we import pydrake so that we can intercept debug
-    # messages while handling the redirect.
-    set_log_level(os.environ["_TEST_SPDLOG_LEVEL"])
 
 from pydrake.common import configure_logging
+from pydrake.common.test.text_logging_test_helpers import do_log_test
 
 
 def main():
--- a/bindings/pydrake/common/test/text_logging_example.py
+++ b/bindings/pydrake/common/test/text_logging_example.py
@@ -1,8 +1,14 @@
 import argparse
 import logging
+import os
+
+from text_logging_test_helpers import do_log_test, set_log_level
+if "_TEST_SPDLOG_LEVEL" in os.environ:
+    # Configure logging before we import pydrake so that we can intercept debug
+    # messages while handling the redirect.
+    set_log_level(os.environ["_TEST_SPDLOG_LEVEL"])
 
 from pydrake.common import configure_logging
-from pydrake.common.test.text_logging_test import do_log_test
 
 
 def main():
--- a/bindings/pydrake/common/test/text_logging_gil_test.py
+++ b/bindings/pydrake/common/test/text_logging_gil_test.py
@@ -3,7 +3,7 @@ import time
 import unittest
 
 from pydrake.common import configure_logging
-from pydrake.common.test.text_logging_test import (
+from pydrake.common.test.text_logging_test_helpers import (
     Worker,
     do_log_test,
 )
--- a/bindings/pydrake/common/test/text_logging_test.py
+++ b/bindings/pydrake/common/test/text_logging_test.py
@@ -147,19 +147,26 @@ class TestTextLoggingExample(unittest.TestCase,
         self.assertEqual(effective_level, expected_level)
 
     def test_disabled_via_env(self):
+        """When the magic environment variable is set, C++ logging is not
+        redirected to Python; it continues to write to stderr directly, using
+        its own independent formatter and log level threshold.
+        """
+        # Disable the code that injects the pylogging_sink.
         env = dict(os.environ)
-        env["_TEST_SPDLOG_LEVEL"] = "debug"
         env["DRAKE_PYTHON_LOGGING"] = "0"
+        # Configure the Python logging to not print anything.
+        python_level = CRITICAL + 1
         try:
             output = subprocess.check_output(
                 ["bindings/pydrake/common/text_logging_example",
-                    "--use_nice_format=0",
-                    "--root_level=-1",
-                    "--drake_level=-1"],
+                    "--use_nice_format=1",
+                    f"--root_level={python_level}",
+                    f"--drake_level={python_level}"],
                 stderr=subprocess.STDOUT,
                 encoding="utf8",
                 env=env)
         except subprocess.CalledProcessError as e:
             print(e.output, file=sys.stderr, flush=True)
             raise
-        self.assertIn("Will not redirect", output)
+        # The C++ logger should still have printed (INFO and higher)
+        self.assertIn("Test Info message", output)
--- a/bindings/pydrake/common/test/text_logging_test.py
+++ b/bindings/pydrake/common/test/text_logging_test.py
@@ -156,10 +156,10 @@ class TestTextLoggingExample(unittest.TestCase,
         (2) We do not specifically change anything about C++ logging level.
         (3) We set the Python logging level to "no logging at all".
         (4) The do_log_test function invokes a bunch of C++ log statements.
-        The test predicate is that the C++ log statements made it out to stderr.
-        If step (1) didn't happen, then step (3) would mean that no messages
-        would ever appear, so the test predicate would fail (the output would
-        be empty).
+        The test predicate is that the C++ log statements made it out to
+        stderr. If step (1) didn't happen, then step (3) would mean that no
+        messages would ever appear, so the test predicate would fail (the
+        output would be empty).
         """
         # Disable the code that injects the pylogging_sink.
         env = dict(os.environ)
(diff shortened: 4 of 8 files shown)


EricCousineau-TRI (commented):
+@jwnimmer-tri Would you be up for feature review on this?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-:-NYxNtXv42kobw0I4vby:bzdudz4)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

jwnimmer-tri (commented):
Sure.

The implementation looks fine; I'll need to circle back a little later to grok the tests.

> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-:-NYxPaUeBFX13I4-uicH:bqzikfz)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYxOsXR765C2XnFhEir:-NYxOsXSD_hooqVSXAgn:b8dwj3m) (no related file):*
Users will need documentation of this new variable to make use of it.

The https://drake.mit.edu/doxygen_cxx/group__environment__variables.html page should list this variable in the inventory, and then either explain what it does inline or else link to somewhere that explains what it does (e.g., the pydrake docs).

___

jwnimmer-tri (commented):
https://github.com/EricCousineau-TRI/drake/pull/17 with suggestions.

> Reviewed 4 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-:-NYxuY3a1cuaAvdwqSx3:bbcznqs)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYxOsXR765C2XnFhEir:-NYyu1neALdAqlBSFf2c:b-41t7rf) (no related file):*

Users will need documentation of this new variable to make use of it.

The https://drake.mit.edu/doxygen_cxx/group__environment__variables.html page should list this variable in the inventory, and then either explain what it does inline or else link to somewhere that explains what it does (e.g., the pydrake docs).
</blockquote></details>

Attempted in latest revision, but I'm not immediately able to figure out why `doxygen.h` can't resolve to groups in `pydrake_doxygen.h`.

Is there any chance you see something obvious that I missed?

___

jwnimmer-tri (commented):
> Reviewed 5 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYxOsXR765C2XnFhEir:-NYz-1fz1Eqti3BaOF8z:b3pnexc) (no related file):*

Attempted in latest revision, but I'm not immediately able to figure out why `doxygen.h` can't resolve to groups in `pydrake_doxygen.h`.

Is there any chance you see something obvious that I missed?
</blockquote></details>

For reasons I don't understand, that particular file has always been _very_ fussy about triple backticks.  A newline and removing the `sh` hint seems to fix it:

```patch
--- a/bindings/pydrake/pydrake_doxygen.h
+++ b/bindings/pydrake/pydrake_doxygen.h
@@ -646,7 +646,8 @@ then logging will not be redirected.
 
 For example, to disable logging redirect, you can set the following in
 your terminal before you run your process that uses pydrake:
-```sh
+
+```
 export PYDRAKE_PYTHON_LOGGING=0
 ```
 
```
```


___
___
*[`bindings/pydrake/pydrake_doxygen.h` line 641 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYz-HuxB4enLrgX3A7-:-NYz-HuxB4enLrgX3A70:b1ukuz7) ([raw file](https://github.com/RobotLocomotion/drake/blob/a8aa4031ecbfffd99125a1f3f406a92350fee31d/bindings/pydrake/pydrake_doxygen.h#L641)):*
> ```objc
> @addtogroup environment_variables
> @{
> @defgroup pydrake_python_logging PYDRAKE_PYTHON_LOGGING
> ```

nit The name is `DRAKE_PYTHON_LOGGING`.

Ditto below.

**Review by EricCousineau-TRI — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYxOsXR765C2XnFhEir:-NZ1t7zR6ovVR_dZUFdZ:b-qcv5w1) (no related file):*

For reasons I don't understand, that particular file has always been _very_ fussy about triple backticks.  A newline and removing the `sh` hint seems to fix it:

```patch
--- a/bindings/pydrake/pydrake_doxygen.h
+++ b/bindings/pydrake/pydrake_doxygen.h
@@ -646,7 +646,8 @@ then logging will not be redirected.
 
 For example, to disable logging redirect, you can set the following in
 your terminal before you run your process that uses pydrake:
-```sh
+
+```
 export PYDRAKE_PYTHON_LOGGING=0
 ```
 
```
```

</blockquote></details>

Done - thanks! (I was able to get it work by only removing `sh`)

___
___
*[`bindings/pydrake/pydrake_doxygen.h` line 641 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NYz-HuxB4enLrgX3A7-:-NZ1t6Mc-9y8XAW5VtYD:b-opq3hi) ([raw file](https://github.com/RobotLocomotion/drake/blob/a8aa4031ecbfffd99125a1f3f406a92350fee31d/bindings/pydrake/pydrake_doxygen.h#L641)):*

nit The name is `DRAKE_PYTHON_LOGGING`.

Ditto below.
</blockquote></details>

Done, derp

**Review by jwnimmer-tri — COMMENTED** [review]
Oops, I'd mean to :lgtm:.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-:-NZ26kbJ7nMJeIbJl-_L:b-tc74ce)* status: needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @EricCousineau-TRI)

**Comment by jwnimmer-tri** [comment]
I'll propose +@xuchenhan-tri for platform review, unless you had some other plan.

**Comment by EricCousineau-TRI** [comment]
Yup, works for me - thanks!

**Review by xuchenhan-tri — COMMENTED** [review]
Platform :lgtm: with one question.

> Reviewed 1 of 5 files at r1, 5 of 5 files at r2, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-:-NZ75NFF6uGx7z6EwMRj:b-j6g7ls)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/test/text_logging_test.py` line 158 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NZ74EGODSVaoiTUTFih:-NZ74EGODSVaoiTUTFii:b-901sgb) ([raw file](https://github.com/RobotLocomotion/drake/blob/ae7a482aaf0f57e7ae5ec0595004aead9f95fcc3/bindings/pydrake/common/test/text_logging_test.py#L158)):*
> ```Python
>         env["DRAKE_PYTHON_LOGGING"] = "0"
>         # Configure the Python logging to not print anything.
>         python_level = CRITICAL + 1
> ```

BTW, I'm confused by the contradiction between this comment and my perceived testing strategy: My understanding is that we want to set the python level so that things are logged in python if the environment variable had not been set. Therefore, when the logging doesn't go through to python but instead shows up in the c++ logger, we know that the environment variable has an effect.

If my understanding is correct, shouldn't we configure python logging to print whatever is redirected by c++ logger?



_Code quote:_
```Python
        # Configure the Python logging to not print anything.
        python_level = CRITICAL + 1
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/test/text_logging_test.py` line 158 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NZ74EGODSVaoiTUTFih:-NZ7Si7_6mvMUWtH10HL:bvw7sr1) ([raw file](https://github.com/RobotLocomotion/drake/blob/ae7a482aaf0f57e7ae5ec0595004aead9f95fcc3/bindings/pydrake/common/test/text_logging_test.py#L158)):*

BTW, I'm confused by the contradiction between this comment and my perceived testing strategy: My understanding is that we want to set the python level so that things are logged in python if the environment variable had not been set. Therefore, when the logging doesn't go through to python but instead shows up in the c++ logger, we know that the environment variable has an effect.

If my understanding is correct, shouldn't we configure python logging to print whatever is redirected by c++ logger?
</blockquote></details>

Here's how I see it.  Maybe you can help us figure out how to better explain it?

(1) We set the env var so that Python logging and C++ logging are independent.
(2) We do not specifically change anything about C++ logging level.
(3) We set the Python logging level to "no logging at all".
(4) The do_log_test function invokes a bunch of C++ log statements.

The test predicate is that the C++ log statements made it out to stderr.

If step (1) didn't happen, then step (3) would mean that no messages would ever appear, so the test predicate would fail (the output would be empty).

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/test/text_logging_test.py` line 158 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19685#-NZ74EGODSVaoiTUTFih:-NZ7XuJjDV9oOYyXdR2p:bggs7i) ([raw file](https://github.com/RobotLocomotion/drake/blob/ae7a482aaf0f57e7ae5ec0595004aead9f95fcc3/bindings/pydrake/common/test/text_logging_test.py#L158)):*
I see, that makes sense! Can we just use this as-is for the doc-string? It's abundantly clear.

> (1) We set the env var so that Python logging and C++ logging are independent.  
(2) We do not specifically change anything about C++ logging level.  
(3) We set the Python logging level to "no logging at all".  
(4) The do\_log\_test function invokes a bunch of C++ log statements.
The test predicate is that the C++ log statements made it out to stderr.
If step (1) didn't happen, then step (3) would mean that no messages would ever appear, so the test predicate would fail (the output would be empty).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19685)* status: commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @EricCousineau-TRI)

PR #19635, opened 2023-06-20
Meshcat can handle Mesh(foo.gltf)

SeanCurtis-TRI:
Meshcat passes embedded .gltf files along to the meshcat session. This includes an incidental recentering of the objects created in the meshcat_manual_test.cc.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19635)

PR #19624, opened 2023-06-19
Allow Mesh specification as gltf (overview)

SeanCurtis-TRI:
This draft PR serves as the overview for the PR train that will allow Mesh and Convex shapes to use (in a limited way) glTF file. The final feature will be implemented with four PRs (each with their own collection of TODOs):

1. Introduce an internal dependency to [nlohmann/json](https://github.com/nlohmann/json) using the self-contained header version.
2. Add functionality for merging one json-representation glTF file into another.
3. Modify `RenderEngineGltfClient` to consume and propagate `Mesh(foo.gltf)`.
4. Modify the remaining geometry ecosystem to have appropriate responses (sometimes throw, sometimes ignore). (Note: as of initial publication of this PR, this final PR is not quite complete.)

The four stages are represented in the first four revisions of this PR.

Relates #19532.

PR #19691, opened 2023-06-28
Confirm meldis support for .gltf

SeanCurtis-TRI:
Confirm that meldis supports .gltf (assuming that Drake and meshcat do).

After #19624 and #19635 have merged into master, this can be rebased and merged. 

This is a proof of concept. The solar system example for scene graph uses `DrakeVisualizer` to visualize an orrery. We've replaced the geometry for the sun (providing a .gltf textured sphere).

Licensing:
 1. The sun texture is a google reduced summary of an image from https://www.solarsystemscope.com/textures/.
 2. The license is [Attribution 4.0 International license](https://creativecommons.org/licenses/by/4.0/).
 3. Prior to merging, we should make sure we've resolved issues regarding the texture's license. We can swap it out if we need to. For now, it's sufficient to have a place holder to evaluate meldis and drake_visualizer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19691)


SeanCurtis-TRI:
If you run meldis in one window and `bazel run //examples/scene_graph:solar_system_run_dynamics` in another, you should see something akin to:

image

jwnimmer-tri:
FYI In terms of a longer-lived manual acceptance test, the idea approach would be running `model_visualizer` on an SDFormat file that uses a glTF mesh. That's the workflow that best matches what users would see on their own. If there's value in improving `examples/scene_graph` to demo glTF as well then so be it, but I don't think the solar_system is the best acceptance test for Meldis.

SeanCurtis-TRI (commented):
Agreed. I went with the solar system as the fastest way to get visual geometry into meldis. I suppose until we have more broad .glTF support, I'm going to have to create a new model (I don't want to take an existing model and suddenly make it broken sometimes.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19691#-:-NZ3LxQY8jrTyPsguAMw:b-10xozw)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
Maybe even `model_visualizer drake/geometry/render_gltf_client/test/textured_green_box.gltf` is sufficient?  The parser could (should) load gltf meshes just like we do for obj meshes.  In the meantime, we could add `textured_green_box.sdf` into git as a stop-gap.

SeanCurtis-TRI (commented):
I'll look at the `textured_green_box.sdf`. That was made to be small. However, I changed it for `meshcat_manual_test.cc` because I wanted to have a clear visual artifact that could not have been loaded from .obj (at least as per current support). I'll probably end up using that dented box. But the principle is the same.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19691#-:-NZ82MD02DJ6QGdeMGdr:b-wv05vk)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI:
FYI `model_visualizer something.gltf` won't work for a while.

If we want it to have similar semantics as obj, we have to have `Parser` accept .gltf which leads us to `detail_mesh_parser.cc` which will need to do the same thing for .gltf as it does for .obj. The two key ideas there are: 1) we have to go from .gltf to mass properties and 2) be able to assign gltf as collision geometry.  The good news is that being able to do the former implicitly empowers us to do the latter (well, 95%).

SeanCurtis-TRI (commented):
On a related note: how do you envision implementing a "longer-lived manual acceptance test"?

Providing an .sdf which references a .gltf file is easy. Making it available as `package://drake/something/something.gltf` is also straightforward. But that's merely the ingredients. How does it pulled over to becoming an acceptance test?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19691#-:-NZYtuCi3h67pNXUr_aw:b98uh8)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
> How does it pulled over to becoming an acceptance test?

I didn't have anything specific in mind.  I imagine that a couple of sentences comment somewhere would suffice.

My goal would be that someone who wanted to re-test glTF support could figure out how to do it, without digging through PR overviews.  I don't think we need a formal procedure like `meshcat_manual_test` -- that would be overkill.

SeanCurtis-TRI:
Well, right now, I've got it documented in a commit message.  :)  Not quite enough.

My basic plan was to add an .sdf with the sun. So, the sun appears in that particular toy example. But, by having it as part of examples, then `package://drake/examples/scene_graph/sun.sdf` is now visible to `bazel run //tools:model_visualizer`. 

In many ways, `model_visualizer` is not good proof that `meldis` is gltf friendly. The solar system is a perfect candidate. The only visualization mechanism it instantiates is a `DrakeVisualizer` -- so, the *only* way to visualize it is through `meldis`.

I'm writing a doxygen module on file support as a follow up. I could add it there, but that seems overkill. Generally, putting it any place that gets rendered into documentation feels overkill -- as time passes, the fact that .gltf is supported in meldis will become so common place that calling it out specially would become absurd.

jwnimmer-tri:
> In many ways, model_visualizer is not good proof that meldis is gltf friendly...

I don't follow this reasoning.  If we run `meldis` first and then `model_visualizer` second, we see the model in _both_ meldis and meshcat, so we can check all visualizers from a single model file, or even visually cross-check the two browser windows against each other.

> My basic plan was to add an .sdf with the sun.

I don't object to this on principle, but in practice dealing with the license stuff is strictly more annoying than using something we already have (e.g., the meshcat_test's cube, if we could wrap it in a sdf).

In any case, I agree with the outline -- we should have at least one sdf file that references a gltf mesh in Drake git somewhere, so that we can spot-check the gltf visualization.  It can be anywhere (examples, or some test dir, etc).

In terms of where to document the existence of such a file -- I think the PR overview and commit message would be good enough.  It will be easy enough to rediscover later.  (Though uncommon in Drake, in the wider world its not uncommon to have a sentence or two in the commit message that says "Test strategy: ..." explaining how the code was tested.)

SeanCurtis-TRI:
re: sun.sdf

I replaced the texture with something I painted in gimp in two minutes. So, there are no licensing issues.

The big thing about using the .gltf from `meshcat_manual_test.cc` is finding a home for it. It's currently only a test-only resource. So, either those tests would have to be included in the set of models made available to `model_visualizer` or it would have to be moved into one of the environments that already support it. That's kind of why the sun seemed the lowest effort path forward.

I'm closing this PR now and we can pick up the conversation, as necessary, in #19739.

jwnimmer-tri:
Model visualizer can visualize any model we want.  It doesn't need to be part of bazel runfiles to visualize it.  The "test only" file group is of no consequence.

PR #19693, opened 2023-06-28
Add lights to RenderEngineGlParams

meggar:
Adds a lights parameter to RenderEngineGlParams to enable configuration of multiple lights.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)


xuchenhan-tri (commented):
Per platform checklist, +@SeanCurtis-TRI for feature review?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NZ83I7t32uGBETfaK5S:b4arsy3)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @meggar)

xuchenhan-tri (commented):
Good day, @meggar. I’m assigning @SeanCurtis-TRI  as the most relevant team member to assist with this pull request. If you already had a different team member in mind, please let us know. To @SeanCurtis-TRI, feel free to delegate in case there is someone else more suitable.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NZ843o9AC_MCN-SEkEu:b62hj1l)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @meggar)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZ83ZiQ3lc8MpOS6F_z:-NZ83ZiQ3lc8MpOS6Fa-:b-nzn6ma) (no related file):*
@drake-jenkins-bot ok to test

___

jwnimmer-tri (commented):
Note that since there was no issue filed to discuss the design ahead of time, @SeanCurtis-TRI we should start with a design/requirements review first, and only after that reaches consensus should we move on to reviewing the implementation, docs, tests, etc.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NZ84yAqDn_NTwaFpPc6:bm4ij0n)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @meggar)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZ84DBV2kyoTHYfwIOA:-NZ84DBV2kyoTHYfwIOB:b-rgp3pi) (no related file):*
@SeanCurtis-TRI for your consideration, here's my proposal for forward-looking semantics:

(1) SceneGraph might eventually denote lights (e.g., moving lights).  For the moment, it does not.

(2) File formats like SDFormat can already specify lights.  Eventually, the Parser should parse those into the SceneGraph.

(3) Beyond that, the render params (for a particular engine) can specify additional lights (like we have here).  The rendered images would have both the SceneGraph lights as well as engine-specific lights.

(4) The default render params should not have any lights.

(5) In the event that the engine is asked to render a color image and has no lights of any kind (no scene graph, no params), then a default light is added -- the same default light we already have.

(6) As some point if the scene graph lighting support becomes quite mature (e.g., has "accepting renderer" semantics), we could imagine deprecating the engine-specific lighting parameters for removal.

___

SeanCurtis-TRI (commented):
Sorry for the delay. Thanks for submitting this, I'll dig into this (and @jwnimmer-tri's comments today.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NZOQ_CM2DM8djqgmksZ:b-7zi01n)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @meggar)

meggar (commented):
Do I need to provide anything else for this review?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-N_4s7rKD1dseJANGDhs:b-t1ug8q)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes

jwnimmer-tri:
Apologies for the delay.  We have a lot of team members on vacation.  Hopefully @SeanCurtis-TRI can reply soon.

SeanCurtis-TRI:
@meggar As you are a new contributor to Drake, there can be a lot of details to resolve in your first review. Rather than inundating you with minutiae, it might be simpler if I push modifications to your PR with notes in the review explaining the why and wherefore of the changes. Would that be alright with you? Do you have any particular preferences? The goal here is to make the PR go as smoothly as possible.

------ addendum

I've been working on this intermittently but am currently on the other side of the earth. I'm not done with my initial pass, but I promise you things are making progress.

SeanCurtis-TRI:

@SeanCurtis-TRI for your consideration, here's my proposal for forward-looking semantics:

(1) SceneGraph might eventually denote lights (e.g., moving lights).  For the moment, it does not.

(2) File formats like SDFormat can already specify lights.  Eventually, the Parser should parse those into the SceneGraph.

(3) Beyond that, the render params (for a particular engine) can specify additional lights (like we have here).  The rendered images would have both the SceneGraph lights as well as engine-specific lights.

(4) The default render params should not have any lights.

(5) In the event that the engine is asked to render a color image and has no lights of any kind (no scene graph, no params), then a default light is added -- the same default light we already have.

(6) As some point if the scene graph lighting support becomes quite mature (e.g., has "accepting renderer" semantics), we could imagine deprecating the engine-specific lighting parameters for removal.
</blockquote></details>

Generally, I agree with what your proposal. It slots this PR nicely into an initial foray which, nevertheless, takes us in a coherent direction.

I've given sporadic thoughts/comments on some of your bullets. Absence of a number merely indicates that I had nothing to add.

(1) Not just moving lights. *any* lights. The lack of light specification has gnawed at me a long time. In fact, has been lurking in my OKRs since fall of 2019, but it never got enough prioritization to bubble up to the top. Ideally, `SceneGraph` would be the path through which any of the `RenderEngine` implementations would receive light specification. The only reason I would prefer per-RenderEngine specifications is if one RenderEngine supported a light type (or light parameters) that others don't support. But this is mostly the maturity to which you allude in (6).

(2) Definite thumbs-up on the SDFormat. It also seems as we look more towards using glTF, we might consider the kind of data we load from it. That might also be a source of lighting.

(3/5) If we implemented (1) and (2) first, then it would make sense to have no lights by default. And that case (described in (5)) would naturally arise. As we go from the other direction (starting at (3)), we have the question: does the presence of engine-specific lights *supplant* the default light? I.e, if you specify *any* lights, you have to specify *all* lights. It certainly seems that it would be consistent with the end goal and is functionally equivalent to the end semantics (`SceneGraph` has specified no lights, `params` has specified no lights, so we use the current "headlamp" style of light).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @meggar)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZ84DBV2kyoTHYfwIOA:-N_5vPD_3Ozs5XRcxRia:bt43w6n) (no related file):*

Generally, I agree with what your proposal. It slots this PR nicely into an initial foray which, nevertheless, takes us in a coherent direction.

I've given sporadic thoughts/comments on some of your bullets. Absence of a number merely indicates that I had nothing to add.

(1) Not just moving lights. *any* lights. The lack of light specification has gnawed at me a long time. In fact, has been lurking in my OKRs since fall of 2019, but it never got enough prioritization to bubble up to the top. Ideally, `SceneGraph` would be the path through which any of the `RenderEngine` implementations would receive light specification. The only reason I would prefer per-RenderEngine specifications is if one RenderEngine supported a light type (or light parameters) that others don't support. But this is mostly the maturity to which you allude in (6).

(2) Definite thumbs-up on the SDFormat. It also seems as we look more towards using glTF, we might consider the kind of data we load from it. That might also be a source of lighting.

(3/5) If we implemented (1) and (2) first, then it would make sense to have no lights by default. And that case (described in (5)) would naturally arise. As we go from the other direction (starting at (3)), we have the question: does the presence of engine-specific lights *supplant* the default light? I.e, if you specify *any* lights, you have to specify *all* lights. It certainly seems that it would be consistent with the end goal and is functionally equivalent to the end semantics (`SceneGraph` has specified no lights, `params` has specified no lights, so we use the current "headlamp" style of light).

</blockquote></details>

(1) Agreed.  (That's why I used `e.g.` not `i.e.`.)

(3) I don't see a need to implement "supplant" anytime soon.  My intuition is that lights could always be additive, and nobody would object.  I will note, however, that SDFormat lights have a "name".  If we decided to name the engine-specific lights, it would make sense for the engine light of a given name to trump the scene graph light with the same name.

___

SeanCurtis-TRI:
> My intuition is that lights could always be additive, and nobody would object.

My intuition says the opposite. The default headlamp's only true virtue is that it guarantees that what you're looking at is illuminated on the camera's side. In my imagination, an effort to introduce specific lights would stem from the effort to recreate something specific, and it seems highly improbably that that *thing* includes a headlamp.

jwnimmer-tri:
You misunderstand me.  Any lights the user writes down would be additive.  The headlamp would only be added when nothing else is there.  It's not a _default_, it's a _fallback_.

meggar:
> @meggar As you are a new contributor to Drake, there can be a lot of details to resolve in your first review. Rather than inundating you with minutiae, it might be simpler if I push modifications to your PR with notes in the review explaining the why and wherefore of the changes. Would that be alright with you? Do you have any particular preferences? The goal here is to make the PR go as smoothly as possible.
> 
> ------ addendum
> 
> I've been working on this intermittently but am currently on the other side of the earth. I'm not done with my initial pass, but I promise you things are making progress.

Sounds good, thank you!

SeanCurtis-TRI:
>  Any lights the user writes down would be additive. The headlamp would only be added when nothing else is there. It's not a default, it's a fallback.

I think this is merely a nomenclature debate. Changing the light to become a "fallback" is synonymous with what I meant by user-specified lights supplanting the headlamp. So, we're in agreement on behavior.

SeanCurtis-TRI:
I've merged master into the branch preparatory to my own changes.

SeanCurtis-TRI (commented):
Alright, I've finally had the chance to finish up the pass.

First, thanks a million for the initial PR. The core of the PR was great. It did serve as the catalyst we needed to start pushing in the lighting we've wanted all along. As you've seen below, it inspired some discussion that had been deferred up to this point. So, thanks for that.

 I've made a number of changes (which I'll flag out down below). Some of the changes are big picture stuff (like the fallback light that arose out of the conversation inspired by your PR), some of it is about fitting your PR into Drake a bit more natively (i.e., being willing to make *bigger* changes to fit the lights in). And then, there are a number of smaller things about style and organization.

I've sprinkled comments around which I've addressed in my additional commits. PTAL.

> Reviewed 5 of 9 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-N_bwM44CXeS9GOgH_44:bdc25p6)* status: 33 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZ84DBV2kyoTHYfwIOA:-N_fTQiy-njJ0hZXMgRR:b-pcmoz2) (no related file):*

> My intuition is that lights could always be additive, and nobody would object.

My intuition says the opposite. The default headlamp's only true virtue is that it guarantees that what you're looking at is illuminated on the camera's side. In my imagination, an effort to introduce specific lights would stem from the effort to recreate something specific, and it seems highly improbably that that *thing* includes a headlamp.
</blockquote></details>

I've updated the PR to include the fallback light. Specify nothing, and we get the legacy behavior. Specify anything, and we use those lights.

___
___
*[`geometry/render_gl/BUILD.bazel` line 132 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_By-wo6ENliY1JcTWK:-N_By-wo6ENliY1JcTWL:bueugez) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/BUILD.bazel#L132)):*
> ```Starlark
>     name = "render_engine_gl_params",
>     hdrs = [
>         "light_parameter.h",
> ```

In anticipation of using this more broadly, I'm moving this to `geometry/render` so it won't be *strictly* `RenderEngineGl` affiliated, but will be able to be used by other `RenderEngine` implementations.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 51 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fTfx9-UHVzCmtJLyV:-N_fTfx9-UHVzCmtJLyW:b-xadldd) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L51)):*
> ```cpp
> constexpr char kInternalGroup[] = "render_engine_gl_internal";
> constexpr char kHasTexCoordProperty[] = "has_tex_coord";
> constexpr int MAX_LIGHT_NUM = 5;
> ```

nit:

- The [styleguide](https://drake.mit.edu/styleguide/cppguide.html#Constant_Names) requires us to name constants like this as `kMaxLightNum`.
- Also, it makes more sense to make this a static member of `LightingShader`.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 60 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fVBWh6QCS-R4y0ZAA:-N_fVBWi-IlQm9Tx6AzT:b-dddbhk) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L60)):*
> ```cpp
> };
> 
> class LightingShader : public ShaderProgram {
> ```

I made a big push to pull everything lighting-related into the `LightingShader`. That includes the GLSL code, the uniform management, etc. The goal was for the diffuse and texture shaders to have a single job: specify the diffuse color that should be illuminated. It eliminated GLSL duplication and allowed each shader to focus on its unique contribution.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 65 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fUBjO0VVQKCzJLion:-N_fUBjO0VVQKCzJLioo:b-mgkj9y) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L65)):*
> ```cpp
>   LightingShader() : ShaderProgram() {}
> 
>   GLint GetLightFieldLocation(int index, std::string field_name) const {
> ```

nit: Moved to be private.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 66 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_Bz-BMAUqrhQd1yQIW:-N_Bz-BMAUqrhQd1yQIX:blzfqwn) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L66)):*
> ```cpp
> 
>   GLint GetLightFieldLocation(int index, std::string field_name) const {
>     return GetUniformLocation("lights[" + std::to_string(index) + "]." +
> ```

nit Generally prefer `fmt::format()` over string concatenation.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 71 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fULSEC3KL-X0macdX:-N_fULSEC3KL-X0macdY:bk4j88h) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L71)):*
> ```cpp
> 
>   void SetAllLights(
>     const std::vector<render::LightParameter>& lights) const final {
> ```

This gave me the chance to *remove* lighting artifacts from `ShaderProgram` and keep them *only* in `LightingShader`. So, no virtual method here.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fUVtmCDRgcu8QLp2x:-N_fUVtmCDRgcu8QLp2y:bjjgfhb) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L102)):*
> ```cpp
>   }
> 
>   // Children of the Lighting Shader cannot use this function.
> ```

nit: As part of the lighting functionality, this function is now gone.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 158 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fUovW9AgUCAUfTOlc:-N_fUovW9AgUCAUfTOld:b-rbg2un) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L158)):*
> ```cpp
>   }
> 
>   void DoGeometryToWorldMatrix(const Eigen::Matrix4f& X_WG) const override {
> ```

This function went away and became part of setting the model view matrix.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 158 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fUtjD54VJ4dBdk16H:-N_fUtjD54VJ4dBdk16I:bx56dpb) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L158)):*
> ```cpp
>   }
> 
>   void DoGeometryToWorldMatrix(const Eigen::Matrix4f& X_WG) const override {
> ```

nit: A note on naming. `X_WG` is best reserved for *rigid* transformations between frames. However, this may include scale factors, so it should be `T_WG`. And, actually, I ended up distinguishing between the geometry's G frame and the "models" M frame -- the relationship is the scale matrix.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 188 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_C31kMEVI-dq8L3bjo:-N_C31kMEVI-dq8L3bjp:b-pa63f4) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L188)):*
> ```cpp
> uniform mat4 T_CM;  // The "model view matrix" (in OpenGl terms).
> uniform mat4 T_DC;  // The "projection matrix" (in OpenGl terms).
> uniform mat4 T_WG;  // The Geometry to World Matrix.
> ```

nit: in this context, this should be called `T_WM` because the "geometry" is the model, so G = M and that makes the rest of the math notation clean up.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 191 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_C2jpL9ksgyobXHT_h:-N_C2jpL9ksgyobXHT_i:b7ws1yr) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L191)):*
> ```cpp
> uniform mat3 normal_mat;
> varying vec3 n_C;
> out vec3 frag_position; // Vertex position in world space
> ```

nit: We should use monogram notation: `p_WV` or `p_WF`. The choice of name can be ambiguous:

1. Here, we're definitely outputting the world-position of the input vertex.
2. In the fragment shader, it gets interpolated across the primitive, so its the world position of the fragment. I opted for `p_WV` with a comment.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 212 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fWZMX2dzmQE9AxWXT:-N_fWZMX2dzmQE9AxWXU:b6be52z) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L212)):*
> ```cpp
> const int MAX_LIGHT_NUM = 5;
> 
> struct Light {
> ```

nit: See `light_parameter.h` for the changes in the light specification.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 217 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fVykm130e3pK8i3Rk:-N_fVykm130e3pK8i3Rl:b-5gh5mm) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L217)):*
> ```cpp
>     // 2 for Spot Light
>     // 3 for Directional Light
>     int light_type;
> ```

nit For a struct called `Light` you don't really need fields with the prefix `light_`. 99.9% of the time it'll become apparent because I'll have `light.type` or `light.direction`.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 232 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fWnXTDnVlNbQCvR5e:-N_fWnXTDnVlNbQCvR5f:b-shd3in) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L232)):*
> ```cpp
> uniform Light lights[MAX_LIGHT_NUM];
> 
> vec3 GetPointLightRGB(Light light, vec3 nhat_C, vec3 rgb) {
> ```

Some of the new light features (e.g., which frame the light is defined in), required some additional code.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 260 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fWhQP0PZtSyNjDAPH:-N_fWhQP0PZtSyNjDAPI:b-vlh7c3) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L260)):*
> ```cpp
>     } else if (light.light_type == 3) {
>         // Directional light
>         return rgb * max(dot(nhat_C, normalize(-light.light_dir)), 0.0) * light.intensity;
> ```

nit: I refactored the code so that we didn't have the full illumination value computed in several places.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 260 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fWxRb5JF9SmiASium:-N_fWxRb5JF9SmiASiun:b-e7drt0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L260)):*
> ```cpp
>     } else if (light.light_type == 3) {
>         // Directional light
>         return rgb * max(dot(nhat_C, normalize(-light.light_dir)), 0.0) * light.intensity;
> ```

This was a math error. The fragment normal was expressed in the *camera* frame. But the light direction was expressed in the world frame. As such, the dot product between them is generally meaningless. I've updated so that the calculations are always done in the world frame.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 926 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_C1bDE368kX1331eIa:-N_C1bDE368kX1331eIb:bhhqif2) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_render_engine_gl.cc#L926)):*
> ```cpp
> 
>     Eigen::Matrix4f X_WG = instance.X_WG.GetAsMatrix4().cast<float>();
>     shader_program.SetGeometryToWorldMatrix(X_WG, instance.scale.cast<float>());
> ```

nit: If we change the how `SetModelViewMatrix()` is parameterized, we can roll this `SetGeometryToWorldMatrix()` into part of that call.  Something like:

```c++
void SetModelViewMatrix(const Eigen::Matrix4f& X_CW,
                        const RigidTransformd& X_WG,
                        const Vector3d& scale);
```
___
*[`geometry/render_gl/internal_shader_program.h` line 259 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPfLj_1hwtLLWcuTS6:-NZPfLj_1hwtLLWcuTS7:b-dorxem) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/internal_shader_program.h#L259)):*
> ```objc
>   /* Sets the pose of the model in the world frame with the given matrix.
>    @param X_WG   The pose of the model in the World frame. */
>   virtual void DoGeometryToWorldMatrix(const Eigen::Matrix4f&) const {}
> ```

nit: For NVI functions, we keep the names aligned, So, `Foo()` is paired with `DoFoo()`. In this case, we've lost the "Set". It's moot here as I deleted the function.
___
*[`geometry/render_gl/internal_shader_program.h` line 140 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fXU7cBtf3Isrcc3on:-N_fXU7cBtf3Isrcc3oo:b-dyzawy) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_shader_program.h#L140)):*
> ```objc
> 
>   /* Gives a hook to the inheriting class to set up the lighting values. */
>   virtual void SetAllLights(
> ```

As mentioned, I pushed all of the lighting-related APIs into the `LightingShader`.
___
*[`geometry/render_gl/internal_shader_program.h` line 162 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fXZ9U8NdEE-p27hvN:-N_fXZ9U8NdEE-p27hvO:bae2rbh) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/internal_shader_program.h#L162)):*
> ```objc
>    @param X_CM   The pose of the *model* frame relative to the camera frame.
>    @param scale  The per-axis scale of the geometry.  */
>   void SetModelViewMatrix(const Eigen::Matrix4f& X_CM,
> ```

By decomposing `X_CM` into `X_CW`, `X_WG` and `scale`, we can do more work internal to this function and change the `DoModelViewMatrix` to more generally support lighting.
___
*[`geometry/render_gl/render_engine_gl_params.h` line 22 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPdu63-dHArcOqA1o1:-NZPdu63-dHArcOqA1o2:b-4izb70) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/render_engine_gl_params.h#L22)):*
> ```objc
>     a->Visit(DRAKE_NVP(default_label));
>     a->Visit(DRAKE_NVP(default_diffuse));
>     a->Visit(DRAKE_NVP(default_clear_color));
> ```

nit: As alluded to elsewhere, we need to make sure the lights get serialized as well.
___
*[`geometry/render_gl/render_engine_gl_params.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPgeAMBgbf_lq9-2ac:-NZPgeAMBgbf_lq9-2ad:b-tlc1n5) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/render_engine_gl_params.h#L36)):*
> ```objc
>   Rgba default_clear_color{204 / 255., 229 / 255., 255 / 255., 1.0};
> 
>   /** Lights in the scene  */
> ```

nit: Missing final period.
___
*[`geometry/render_gl/render_engine_gl_params.h` line 36 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPghbH0kSgR-N5DK_A:-NZPghbH0kSgR-N5DK_B:bm7oeq6) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/render_engine_gl_params.h#L36)):*
> ```objc
>   Rgba default_clear_color{204 / 255., 229 / 255., 255 / 255., 1.0};
> 
>   /** Lights in the scene  */
> ```

We should be clear that we have a limit to how many lights. 
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 205 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fZ96zESxKv3yTWJbp:-N_fZ96zESxKv3yTWJbq:b7wg3ju) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/test/internal_render_engine_gl_test.cc#L205)):*
> ```cpp
> }
> 
> // Helper to construct a downward facing light for RenderEngineGlParams.
> ```

This was a decent effort at making the tests pass. It works because the camera is placed above the objects, looking in the -z direction. But this doesn't actually reproduce the fallback camera.

So, I've added specific tests:

1. Make sure the fallback camera really has expected behavior.
2. Make sure the three light types work as expected (with all of the parameters contributing as expected).
3. Make sure we can do multiple lights.
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 355 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_f_M1l6h2uc4fUba6T:-N_f_M1l6h2uc4fUba6U:b-nf2n7o) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/bindings/pydrake/geometry/geometry_py_render.cc#L355)):*
> ```cpp
>     py::class_<Class> cls(m, "LightParameter", cls_doc.doc);
>     cls  // BR
>         .def_readwrite("type", &LightParameter::type)
> ```

We needed a bit more in the bindings of this type. And unit tests as well.
___
*[`geometry/render_gl/light_parameter.h` line 10 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPdRdU7weBxrWMWJ0Q:-NZPdRdU7weBxrWMWJ0R:b-a43mvk) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/light_parameter.h#L10)):*
> ```objc
> 
> enum LightType {
>   POINT = 1,
> ```

nit: According to our styleguide, [enum values should be named like constants](https://drake.mit.edu/styleguide/cppguide.html#Enumerator_Names). E.g., `kPoint` instead of `POINT`. 
___
*[`geometry/render_gl/light_parameter.h` line 16 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPd_2j1hQDZW8Dgqux:-NZPd_2j1hQDZW8Dgquy:bc170td) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/light_parameter.h#L16)):*
> ```objc
> 
> /** Light parameter for RenderEngineGlParams.  */
> struct LightParameter {
> ```

nit: We need to make sure this can be serialized so that the `RenderEngineGlParams` can be fully serialized.
___
*[`geometry/render_gl/light_parameter.h` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZPdnJX42d4P2UZ1oa-:-NZPdnJX42d4P2UZ1oa0:b3qqytq) ([raw file](https://github.com/RobotLocomotion/drake/blob/fdc8c4f5895f60c470084f04aaaf6a179ef01c28/geometry/render_gl/light_parameter.h#L19)):*
> ```objc
>   /** The light type is either Point, Spotlight, or Directional.
>     Default is Directional light, because that is what it was in VTK.  */
>   LightType type{LightType::DIRECTIONAL};
> ```

nit: A little whitespace between fields aids reading.
___
*[`geometry/render_gl/light_parameter.h` line 15 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fXzGXE2_yqAcqVZy5:-N_fXzGXE2_yqAcqVZy6:b77nsom) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/light_parameter.h#L15)):*
> ```objc
> };
> 
> /** Light parameter for RenderEngineGlParams.  */
> ```

I put a great deal of effort in adding documentation.

- Vectors and positions should always be clear about what frame they are measured and expressed in.
- Attenuation coefficients and intensity should be sufficiently documented such that a user can make intelligent decisions in how to tune it. Including URLs (in order to not reinvent the wheel) is a good thing.
- `cone_angle` needed to be clear about whether it was the entire span of the spotlight, or the half angle (from central axis to boundary).
___
*[`geometry/render_gl/light_parameter.h` line 15 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fYOUz7D5Rp7VAiuFn:-N_fYOUz7D5Rp7VAiuFo:bszmd2y) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/light_parameter.h#L15)):*
> ```objc
> };
> 
> /** Light parameter for RenderEngineGlParams.  */
> ```

I've added some properties:

1. Light color -- this seemed like a no-brainer. It's nice to be able to specify color.
2. Frame. The legacy behavior had the light fixed to the camera frame. So, we needed to be able to specify this here as well.
___
*[`geometry/render_gl/light_parameter.h` line 32 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fYhRQDXpfnYp1a7dg:-N_fYhRQDXpfnYp1a7dh:b-scs2p5) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/light_parameter.h#L32)):*
> ```objc
>     lights  */
>   Eigen::Vector3d light_direction = {0, 0, 1};
> };
> ```

I've added tests to the struct:

1. Putting a regression test on the default values.
2. Confirm successful serialization.
___
*[`geometry/render_gl/test/multithread_safety_test.cc` line 202 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_f_7dnCP_6ezcNQnD5:-N_f_7dnCP_6ezcNQnD6:b-e9ub4q) ([raw file](https://github.com/RobotLocomotion/drake/blob/f578409b85d81fd7a7b5a8515ff3fa24bc8d9781/geometry/render_gl/test/multithread_safety_test.cc#L202)):*
> ```cpp
> 
> /* Helper to construct a downward facing light for RenderEngineGlParams. */
> LightParameter ConstructDownwardDirectionalLight() {
> ```

This all got reverted for the reasons mentioned in `internal_render_engine_gl_test.cc`.

SeanCurtis-TRI:
*[`geometry/render/light_parameter.h` line 1 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-N_fj-wVBhPx2jbZmx4W:-N_fj-wVBhPx2jbZmx4X:b-cz6z3s) ([raw file](https://github.com/RobotLocomotion/drake/blob/865e85a63ba5e18aed47856aec1d8d6808089a55/geometry/render/light_parameter.h#L1)):*
> ```objc
> #pragma once
> ```

Sadly, the move and the merge sufficiently confused reviewable that you can't distinguish the two acts.

SeanCurtis-TRI (commented):
@jwnimmer-tri I've done enough changes to the code, that we'll probably need another feature reviewer. Thoughts?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-N_fiwwk5SWOXyR1aPAo:b-l3b65d)* status: 34 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

jwnimmer-tri:
@drake-jenkins-bot ok to test

@SeanCurtis-TRI right, I don't think you can serve as feature reviewer.  You'll need to find someone else.  Maybe @joemasterjohn would be interested?

joemasterjohn (commented):
Sure thing, +@joemasterjohn.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-N_inG94CG2kAwo3UhGl:blspx3m)* status: 34 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),joemasterjohn, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

SeanCurtis-TRI:
@joemasterjohn Where is this on your schedule?

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 34 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),joemasterjohn, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NZ84DBV2kyoTHYfwIOA:-NaHhxoUBlwdZUrt7TBt:b4blms0) (no related file):*

I've updated the PR to include the fallback light. Specify nothing, and we get the legacy behavior. Specify anything, and we use those lights.
</blockquote></details>

Taking a look today, I'll have some comments up by this afternoon.

___

joemasterjohn (commented):
:lgtm: Looks great, nits are cleanup and optimization suggestions. I have one blocking comment to consider doing lighting calculations in camera space. PTAL and consider, but I'd settle for a TODO and follow up PR. But we can get platform rolling on this: +@EricCousineau-TRI for platform, please.

> Reviewed 16 of 22 files at r3, 3 of 5 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NaLHebu6JVuBUrEiAg4:balljpr)* status: 46 unresolved discussions, LGTM missing from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`geometry/render/light_parameter.h` line 23 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaKwuieBB-RLK3h0PoO:-NaKwuieBB-RLK3h0PoP:bh1s6si) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render/light_parameter.h#L23)):*
> ```objc
>            set of directions.
>   - kDirectional: a light, infinitely far away, casting light parallel rays of
>                   light (like a sun). */
> ```

nit: doc cleanup



_Suggestion:_
```objc
  - kDirectional: a light source, infinitely far away, casting parallel rays of
                  light (like a sun). */
```
___
*[`geometry/render/light_parameter.h` line 120 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaKytEj3IlTZYgOx82h:-NaKytEj3IlTZYgOx82i:bpv20al) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render/light_parameter.h#L120)):*
> ```objc
>    example, in the world frame, a spotlight with direction [0, 0, -1]ᵀ points
>    straight down. In the camera frame, the direction [0, 0, 1]ᵀ points forward
>    in the camera's view direction. This field only applies to spotlight and
> ```

BTW I'm used to (0, 0, -1) being forward in the camera frame. Is this a typo or convention in our renderers?



_Code quote:_
```objc
   straight down. In the camera frame, the direction [0, 0, 1]ᵀ points forward
   in the camera's view direction. This field only applies to spotlight and
```
___
*[`geometry/render/test/light_parameter_test.cc` line 22 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL000g3cwjCLfNfsIM:-NaL000g3cwjCLfNfsIN:bijru88) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render/test/light_parameter_test.cc#L22)):*
> ```cpp
>   const LightParameter light;
>   EXPECT_EQ(light.type, LightType::kDirectional);
>   EXPECT_EQ(light.color, Rgba(1, 1, 1));
> ```

nit: missing check for default attenuation
___
*[`geometry/render/test/light_parameter_test.cc` line 44 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL02uMADlJiKNQVYD9:-NaL02uMADlJiKNQVYDA:b50bkmf) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render/test/light_parameter_test.cc#L44)):*
> ```cpp
> 
>   EXPECT_EQ(light.type, dut.type);
>   EXPECT_EQ(light.color, dut.color);
> ```

nit: missing check for attenuation
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 82 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLCeuhBIzkk9df9JUx:-NaLCeuhBIzkk9df9JUy:bhglsu8) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L82)):*
> ```cpp
>   }
> 
>   static constexpr int kMaxNumLights{5};
> ```

nit: Subclasses do an `fmt::format` on the shader string anyway, so why not make the constant `MAX_LIGHT_NUM` in the shader a format argument and format it with `lights.size()`?



_Code quote:_
```cpp
  static constexpr int kMaxNumLights{5};
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 105 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL2Y9E4p92d2cMXfDh:-NaL2Y9E4p92d2cMXfDi:b-rdyedc) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L105)):*
> ```cpp
> uniform mat3 T_WM_normals;  // Rotation * inverse_scale to transform normals.
> out vec3 n_W;
> out vec3 p_WV; // Vertex position in world space.
> ```

I'd advocate for doing lighting calculations in camera frame. Lights may have to be transformed into camera frame, but there are undoubtedly fewer lights than vertices. You would instead store (the already intermediately calculated) `p_CV = T_CM * vec4(p_MV, 1);` and `n_C = normalize(T_CM_normals * n_M)` for lighting calculations, and avoid computing the vertex positions in world frame.



_Code quote:_
```cpp
out vec3 n_W;
out vec3 p_WV; // Vertex position in world space.
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 161 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL55Iy4egwuxu0sYrv:-NaL55Iy4egwuxu0sYrw:bj91jvh) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L161)):*
> ```cpp
> uniform Light lights[MAX_LIGHT_NUM];
> 
> vec3 GetLightPositionInWorld(Light light, vec3 v) {
> ```

BTW: curious as to why the argument `v` is used here. Is it not always the swizzled `light.position.xyz`?



_Code quote:_
```cpp
vec3 GetLightPositionInWorld(Light light, vec3 v) {
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 227 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL6iMFARLe4TGR4PuR:-NaL6iMFARLe4TGR4PuS:bbw7moy) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L227)):*
> ```cpp
>   float attenuation = 1 / (light.atten_coeff[0] +
>                            light.atten_coeff[1] * dist_FL +
>                            light.atten_coeff[2] * (dist_FL * dist_FL));
> ```

nit: Horner's can save you a multiplication. You can also compute `inverse_attenuation` and save the divide for the end.



_Suggestion:_
```cpp
  float attenuation = 1 / (light.atten_coeff[0] +
                          (light.atten_coeff[1] +
                           light.atten_coeff[2] * dist_FL) * dist_FL);
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 246 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL8Z5ZBodWaZGMpu5P:-NaL8Z5ZBodWaZGMpu5Q:b4jaa2) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L246)):*
> ```cpp
>   color.rgb = illum * diffuse.rgb;
>   color.a = diffuse.a;
>   return color;
> ```

nit: code simplification.



_Suggestion:_
```cpp
  return vec4(illum * diffuse.rgb, diffuse.a);
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 296 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLAkKLA8BHGtUBPtyF:-NaLAkKLA8BHGtUBPtyG:b-tqttuh) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L296)):*
> ```cpp
>     // TODO(SeanCurtis-TRI): Consider packing "type" in with "color" and
>     // cos_half_angle in with dir (as vec4). This would reduce the number of
>     // uniforms and OpenGL context changes.
> ```

BTW: Love these little shader optimization hacks. `intensity` could also be packed with `attenuation`.



_Code quote:_
```cpp
    // TODO(SeanCurtis-TRI): Consider packing "type" in with "color" and
    // cos_half_angle in with dir (as vec4). This would reduce the number of
    // uniforms and OpenGL context changes.
```
___
*[`geometry/render_gl/internal_shader_program.h` line 246 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLEfp_EFZl8Xir-dUt:-NaLEfp_EFZl8Xir-dUu:b5ftk4o) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_shader_program.h#L246)):*
> ```objc
>    @param X_WG     The pose of the geometry in the world frame.
>    @param scale    The per-axis scale of the geometry.  */
>   virtual void DoModelViewMatrix(const Eigen::Matrix4f& /* X_CW */,
> ```

nit: name alignment



_Suggestion:_
```objc
 virtual void DoSetModelViewMatrix
```
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 1576 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLGYi7A3nzGj20XYR9:-NaLGYi7A3nzGj20XYRA:b-lyepdg) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/test/internal_render_engine_gl_test.cc#L1576)):*
> ```cpp
> }
> 
> // Confirm the properties of the fallback camera using the following
> ```

BTW: Great unit test!

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 46 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-NaN37HBD1WtBU_4S7Np:bteoj2m) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*

BTW Thanks for the clarification!

While the serialization above for string <-> enum is nice, things like this means you're actually doing an archive-specialized serialization (e.g. YAML does string, Python should maintain original type).

I don't have a better solution at the moment, but perhaps the above `Serialize` functions could look like this:
```
if constexpr (is_yaml_archive<Archive>) {
  // Do string stuff.
} else {
  // Do normal stuff
}
```

Maybe doesn't need to be solved on this go around.
\cc @jwnimmer-tri @rpoyner-tri if y'all have additional thoughts / comments.
</blockquote></details>

Broadly, until we invent some project-wide architectural sugar for enums, I think that params structs should always use `std::string` in lieu of enums, and let the consuming implementation convert and sanity-check the strings.  See for example `MultibodyPlantConfig::contact_model` or `SimulatorConfig::integration_scheme`.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 47 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-NaNYaEH62G6XgePLf4a:blz3co9) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*

Broadly, until we invent some project-wide architectural sugar for enums, I think that params structs should always use `std::string` in lieu of enums, and let the consuming implementation convert and sanity-check the strings.  See for example `MultibodyPlantConfig::contact_model` or `SimulatorConfig::integration_scheme`.
</blockquote></details>

Gotcha, raising disposition to "discussion" (nit): For consistency, can we change the param struct to use a string?

SeanCurtis-TRI (commented):
I'm tackling them now -- I'm currently in Tokyo, so my work day is more or less opposite yours. :)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NaOvlfvBp9f53NG2men:bu3ydhr)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`geometry/render/light_parameter.h` line 120 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaKytEj3IlTZYgOx82h:-NaOw2bvByk3UGoe82uK:b-dc5c8c) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render/light_parameter.h#L120)):*

BTW I'm used to (0, 0, -1) being forward in the camera frame. Is this a typo or convention in our renderers?
</blockquote></details>

If you look in `camera_info.h` to see Drake's [definition of the camera and its frame](https://github.com/RobotLocomotion/drake/blob/master/systems/sensors/camera_info.h#L50-L55).  tl;dr Yep. Looks in the +Cz direciton.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 82 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLCeuhBIzkk9df9JUx:-NaOx5I9-_Oe21vJ3v4N:b-lf5xn5) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L82)):*

nit: Subclasses do an `fmt::format` on the shader string anyway, so why not make the constant `MAX_LIGHT_NUM` in the shader a format argument and format it with `lights.size()`?
</blockquote></details>

A good idea. I've added a TODO in the shader.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 105 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL2Y9E4p92d2cMXfDh:-NaOyklFCaoH8HMOUoDw:b-v52btr) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L105)):*

I'd advocate for doing lighting calculations in camera frame. Lights may have to be transformed into camera frame, but there are undoubtedly fewer lights than vertices. You would instead store (the already intermediately calculated) `p_CV = T_CM * vec4(p_MV, 1);` and `n_C = normalize(T_CM_normals * n_M)` for lighting calculations, and avoid computing the vertex positions in world frame.
</blockquote></details>

Done  Added TODO.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 161 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL55Iy4egwuxu0sYrv:-NaOzBNq3T2n6hWvReg5:b-pqup4) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L161)):*

BTW: curious as to why the argument `v` is used here. Is it not always the swizzled `light.position.xyz`?
</blockquote></details>

Done You're right. That's a relic of a previous version. Same for light direction.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 227 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL6iMFARLe4TGR4PuR:-NaOzsNlDjVIcpaK8F9d:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L227)):*

nit: Horner's can save you a multiplication. You can also compute `inverse_attenuation` and save the divide for the end.
</blockquote></details>

Done
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 246 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaL8Z5ZBodWaZGMpu5P:-NaP-L_r00u9Tc4wsARl:b-o8k7qr) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L246)):*

nit: code simplification.
</blockquote></details>

Definitely a "compaction".  Simplification? Maybe. I'd generally assume that the shader compiler would take care of it, but I'm too lazy to go looking.  :)
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 296 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLAkKLA8BHGtUBPtyF:-NaP-W7f3qUw-YXd3ahw:b1h05m3) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/internal_render_engine_gl.cc#L296)):*

BTW: Love these little shader optimization hacks. `intensity` could also be packed with `attenuation`.
</blockquote></details>

I did a better job documenting that up in the shader. I just realized I'd documented it twice, so I'm going to delete this one in favor of that one.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 1576 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaLGYi7A3nzGj20XYR9:-NaP0assEK7qL_Wfk19h:b-hay274) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/geometry/render_gl/test/internal_render_engine_gl_test.cc#L1576)):*

BTW: Great unit test!
</blockquote></details>

Thanks. 
___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-NaP1UDk-T5lid9aFqNU:b-5zlsaa) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*

Gotcha, raising disposition to "discussion" (nit): For consistency, can we change the param struct to use a string?
</blockquote></details>

Then, to be perfectly frank, I'd rather remove the serialization.

1. `CameraConfig` doesn't admit for the specification of `RenderEngine` parameters. So, the most Drake-canonical form for yaml-specifying rendering does not benefit from this being serializable.
2. Admittedly, by making it serializable, I enable authors to roll their *own* `CameraConfig` where they *can* specify light parameters in a yaml.

I think it more likely that Drake will solve enums before we have widespread adoption of people writing their own `CameraConfig` (despite our best efforts with `Scenario` examples).

(Alternatively, we change the `renderer_type` string in `CameraConfig` to a variant on `RenderEngineParameters`. Then it would immediately benefit from this being serializable.)

I haven't taken any specific action yet, because I wanted to collect more thoughts.

Thoughts?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-NaSn35Q1bQNDP9lu1V3:b-73s8o6) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*
> So, the most Drake-canonical form for yaml-specifying rendering does not benefit from this being serializable.

That's not accurate.  The CameraConfig can refer to a pre-existing renderer by name.  The user could easily have a Scenario struct that had a dict of renderer names and their params, and then cite those by name in their dict of CameraConfigs.

The benefit of keeping all configs uniformly and plainly serializable is high.

The benefit of compile-time checking of hard-coded params values (i.e., by using enums instead of strings) is low.  Admittedly it does feel weird to use strings for something that has a fixed set of valid choices, but people do it _all the time_ and it works fine in practice.  If we step back with a cold eye, the compiler checking isn't actually adding that much value.

joemasterjohn (commented):
> Reviewed 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-NaczBCG9atd17zsFjmP:b-b9x8qt) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*

> So, the most Drake-canonical form for yaml-specifying rendering does not benefit from this being serializable.

That's not accurate.  The CameraConfig can refer to a pre-existing renderer by name.  The user could easily have a Scenario struct that had a dict of renderer names and their params, and then cite those by name in their dict of CameraConfigs.

The benefit of keeping all configs uniformly and plainly serializable is high.

The benefit of compile-time checking of hard-coded params values (i.e., by using enums instead of strings) is low.  Admittedly it does feel weird to use strings for something that has a fixed set of valid choices, but people do it _all the time_ and it works fine in practice.  If we step back with a cold eye, the compiler checking isn't actually adding that much value.
</blockquote></details>

Sorry, "roll their own `CameraConfig`" was horrible shorthand for wrapping `CameraConfig` with extra stuff to include parameters. Essentially, the same as your `Scenario` suggestion (although, I wouldn't put it at the `Scenario` level, hence "rolling my own".

I guess I have a strong bias against strings. Although, the fact that the strings are immediately validated limits the ways it can go wrong; data can't silently disappear. So, in this case, I can buy the argument to make them strings.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693)* status: LGTM missing from assignee SeanCurtis-TRI(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @meggar)

___
*[`bindings/pydrake/geometry/geometry_py_render.cc` line 366 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-NaN0kd25jDLXudiFBkU:-Nado5Eo6DIr7dbDLC2E:bhcm4l2) ([raw file](https://github.com/RobotLocomotion/drake/blob/9114aa2c3fe5299aed41f0929b82eb67afdc2b17/bindings/pydrake/geometry/geometry_py_render.cc#L366)):*

Sorry, "roll their own `CameraConfig`" was horrible shorthand for wrapping `CameraConfig` with extra stuff to include parameters. Essentially, the same as your `Scenario` suggestion (although, I wouldn't put it at the `Scenario` level, hence "rolling my own".

I guess I have a strong bias against strings. Although, the fact that the strings are immediately validated limits the ways it can go wrong; data can't silently disappear. So, in this case, I can buy the argument to make them strings.


</blockquote></details>

Done Swapped to strings.

(Rather than pay the per-rendering cost of "converting" string to int, I've gone all in on "specify lights at construction". If we want to edit lighting after creation, we can revisit it later.)

SeanCurtis-TRI:
-@SeanCurtis-TRI (stopped being feature reviewer, and slid to become co-author).

SeanCurtis-TRI (commented):
+(status: squashing now) 

@meggar Thank you so much for getting this rolling. It's nice to be able to configure lights.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19693#-:-NaiGDaP3vjZsQcQJZNB:brdzqkr)* status: :shipit: complete! all discussions resolved, LGTM from assignees EricCousineau-TRI(platform),joemasterjohn

PR #19695, opened 2023-06-28
[bindings] Add functions to explicitly enable/disable text logging

EricCousineau-TRI:
Towards #19680
Relates #19685

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695)

### DIFF
--- a/bindings/pydrake/common/BUILD.bazel
+++ b/bindings/pydrake/common/BUILD.bazel
@@ -513,10 +513,21 @@ drake_py_binary(
     ],
 )
 
+drake_py_binary(
+    name = "text_logging_example_toggle",
+    testonly = 1,
+    srcs = ["test/text_logging_example_toggle.py"],
+    deps = [
+        ":module_py",
+        ":text_logging_test_py",
+    ],
+)
+
 drake_py_unittest(
     name = "text_logging_test",
     data = [
         ":text_logging_example",
+        ":text_logging_example_toggle",
     ],
     shard_count = 8,
     deps = [
--- a/bindings/pydrake/common/module_py.cc
+++ b/bindings/pydrake/common/module_py.cc
@@ -78,6 +78,7 @@ void def_testing(py::module m) {
     return std::unique_ptr<RegisteredType>(new UnregisteredDerivedType());
   });
 }
+
 }  // namespace testing
 
 PYBIND11_MODULE(_module_py, m) {
@@ -101,6 +102,9 @@ PYBIND11_MODULE(_module_py, m) {
       doc.logging.set_log_level.doc);
 
   internal::MaybeRedirectPythonLogging();
+
+  m.def("MaybeUndoRedirectPythonLogging",
+      &internal::MaybeUndoRedirectPythonLogging);
 
   py::enum_<drake::ToleranceType>(m, "ToleranceType", doc.ToleranceType.doc)
       .value("kAbsolute", drake::ToleranceType::kAbsolute,
@@ -134,10 +138,10 @@ deterministic given the C++ random seed (see drake issue #12632 for the
 discussion), use e.g.
 
 .. code-block:: python
-    
+
    generator = pydrake.common.RandomGenerator()
    random_state = numpy.random.RandomState(generator())
-   
+
    my_random_value = random_state.uniform()
    ...
 
--- a/bindings/pydrake/common/test/text_logging_example_toggle.py
+++ b/bindings/pydrake/common/test/text_logging_example_toggle.py
@@ -1,0 +1,24 @@
+import logging
+
+from pydrake.common import MaybeUndoRedirectPythonLogging
+from pydrake.common.test.text_logging_test import do_log_info
+
+
+def main():
+    do_log_info(f"Redirected")
+
+    # Takes effect.
+    assert MaybeUndoRedirectPythonLogging()
+    # Does not take effect twice in a row.
+    assert not MaybeUndoRedirectPythonLogging()
+
+    do_log_info(f"Not redirected")
+
+    drake_logger = logging.getLogger("drake")
+    assert not drake_logger._tied_to_spdlog
+
+    # TODO: Verify monkey patches are reverted / no longer have an effect?
+
+
+if __name__ == "__main__":
+    main()
--- a/bindings/pydrake/common/test/text_logging_test.py
+++ b/bindings/pydrake/common/test/text_logging_test.py
@@ -144,3 +144,25 @@ class TestTextLoggingExample(unittest.TestCase,
 
         # Confirm that the correct levels came out.
         self.assertEqual(effective_level, expected_level)
+
+    def test_toggle(self):
+        try:
+            output = subprocess.check_output(
+                ["bindings/pydrake/common/text_logging_example_toggle"],
+                stderr=subprocess.STDOUT,
+                encoding="utf8")
+        except subprocess.CalledProcessError as e:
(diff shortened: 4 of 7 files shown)


EricCousineau-TRI (commented):
One more option per issue, can close that after they both land.

+@jwnimmer-tri Might you be up for review once more?
\cc @dmcconachie-tri 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-:-NZ2XPuh02B8Jvpqplhs:b-wj4g7m)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

jwnimmer-tri (commented):
Sure.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-:-NZ2ZaMzEZ7zPeyNoDwc:b-80tlqk)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2ZIxk3YoYpg4DXlpJ:-NZ2ZIxk3YoYpg4DXlpK:bcqb996) (no related file):*
I understand the use case for providing `MaybeUndoRedirectPythonLogging` to users.

Could you explain the use case for providing `MaybeRedirectPythonLogging`?  (An undo-redo cycle is just one more thing to write tests for, so it'd be best to avoid that effort if there's no use case.)

___
___
*[`bindings/pydrake/common/text_logging_pybind.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2Zw4ADaH8hE4BXNHE:-NZ2Zw4B2OYo8YvbQm6o:bwckcpw) ([raw file](https://github.com/RobotLocomotion/drake/blob/7e9169b7d1128167e35b11511d164d8b2c326a98/bindings/pydrake/common/text_logging_pybind.cc#L38)):*
> ```cpp
> 
>     // Annotate that the logger is alive (fed by spdlog).
>     py::setattr(logger, "_tied_to_spdlog", py::cast(true));
> ```

When logging redirection is undone, we somehow need to undo `_monkey_patch_logger_level_events()` as well.  Post-undo log level changes in Python should not affect the C++ levels anymore (and if I'm reading correctly, it looks like they still do).  We'll also need a regression test that that would have caught this.

The whole thing is pretty subtle.  If you want to kick it over to me to work on a bulletproof implementation with comprehensive test suit coverage, let me know.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2ZIxk3YoYpg4DXlpJ:-NZ30p7kEO7xQkuDzUBo:bp0werc) (no related file):*

I understand the use case for providing `MaybeUndoRedirectPythonLogging` to users.

Could you explain the use case for providing `MaybeRedirectPythonLogging`?  (An undo-redo cycle is just one more thing to write tests for, so it'd be best to avoid that effort if there's no use case.)
</blockquote></details>

To be honest, I don't have a concrete use case for it, but I wasn't sure if there'd be a point where we'd want a context manager for confined scopes.

I can reduce this to just `MaybeUndoRedirectPythonLogging()`.

___
___
*[`bindings/pydrake/common/text_logging_pybind.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2Zw4ADaH8hE4BXNHE:-NZ317W1D-e9AJhzOLjq:bsfolks) ([raw file](https://github.com/RobotLocomotion/drake/blob/7e9169b7d1128167e35b11511d164d8b2c326a98/bindings/pydrake/common/text_logging_pybind.cc#L38)):*

When logging redirection is undone, we somehow need to undo `_monkey_patch_logger_level_events()` as well.  Post-undo log level changes in Python should not affect the C++ levels anymore (and if I'm reading correctly, it looks like they still do).  We'll also need a regression test that that would have caught this.

The whole thing is pretty subtle.  If you want to kick it over to me to work on a bulletproof implementation with comprehensive test suit coverage, let me know.
</blockquote></details>

I think it'd be great to have the full gamut, but I don't know if we need it per your above question.
I'll just shrink this to "undo-only" for now!

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`bindings/pydrake/common/text_logging_pybind.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2Zw4ADaH8hE4BXNHE:-NZ31k3-0sbxtGG0xest:bpxnlye) ([raw file](https://github.com/RobotLocomotion/drake/blob/7e9169b7d1128167e35b11511d164d8b2c326a98/bindings/pydrake/common/text_logging_pybind.cc#L38)):*

I think it'd be great to have the full gamut, but I don't know if we need it per your above question.
I'll just shrink this to "undo-only" for now!
</blockquote></details>

Ah, sorry, misread this comment - I see that this is invariant of undo+redo.

Yes, could I ask if your help on that front?

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2ZIxk3YoYpg4DXlpJ:-NZ33Mns7zTzZkYPflES:b-896fix) (no related file):*

To be honest, I don't have a concrete use case for it, but I wasn't sure if there'd be a point where we'd want a context manager for confined scopes.

I can reduce this to just `MaybeUndoRedirectPythonLogging()`.
</blockquote></details>

Done.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19695)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

___
*[`bindings/pydrake/common/text_logging_pybind.cc` line 38 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19695#-NZ2Zw4ADaH8hE4BXNHE:-NZ7moTeCNugZXd2TmMq:b51rtfs) ([raw file](https://github.com/RobotLocomotion/drake/blob/7e9169b7d1128167e35b11511d164d8b2c326a98/bindings/pydrake/common/text_logging_pybind.cc#L38)):*

Ah, sorry, misread this comment - I see that this is invariant of undo+redo.

Yes, could I ask if your help on that front?
</blockquote></details>

=> https://github.com/RobotLocomotion/drake/pull/19703

For me, starting over from scratch turned out easier than trying to amend this PR.

EricCousineau-TRI:
works for me, thanks!

PR #19667, opened 2023-06-23
Add Reset functions for realtime rate calculation

Xining-Du:
Previously when resuming simulator from a paused state, the realtime rate (RTR) calculator will use a wrong prev_sim_time_, resulting in a sudden drop of RTR in meshcat. This adds a new API ResetRealtimeRateCalculator() to meshcat visualizer which should be called upon simulation resuming to fix the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19667)

PR #19700, opened 2023-06-29
[geometry] Add Meshcat::GetRealtimeRate

jwnimmer-tri:
Tidy up recent Meshcat changes (amends #19667):
- Add missing pydrake bindings.
- Strongly prefer testing the public API (eschew test-friendship hacks). We want to guard against regressions in the end-user experience; using private API goes against that goal.
- Fix indentation, typos, and eschew abbreviation.

\CC @Xining-Du @EricCousineau-TRI FYI

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_visualizers.cc
+++ b/bindings/pydrake/geometry/geometry_py_visualizers.cc
@@ -109,6 +109,8 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py::arg("meshcat"), py::arg("params") = MeshcatVisualizerParams{},
             // `meshcat` is a shared_ptr, so does not need a keep_alive.
             cls_doc.ctor.doc)
+        .def("ResetRealtimeRateCalculator", &Class::ResetRealtimeRateCalculator,
+            cls_doc.ResetRealtimeRateCalculator.doc)
         .def("Delete", &Class::Delete, cls_doc.Delete.doc)
         .def("StartRecording", &Class::StartRecording,
             py::arg("set_transforms_while_recording") = true,
@@ -281,6 +283,8 @@ void DoScalarIndependentDefinitions(py::module m) {
         .def("Delete", &Class::Delete, py::arg("path") = "", cls_doc.Delete.doc)
         .def("SetRealtimeRate", &Class::SetRealtimeRate, py::arg("rate"),
             cls_doc.SetRealtimeRate.doc)
+        .def("GetRealtimeRate", &Class::GetRealtimeRate,
+            cls_doc.GetRealtimeRate.doc)
         .def("SetProperty",
             py::overload_cast<std::string_view, std::string, bool,
                 const std::optional<double>&>(&Class::SetProperty),
--- a/bindings/pydrake/geometry/test/visualizers_test.py
+++ b/bindings/pydrake/geometry/test/visualizers_test.py
@@ -193,6 +193,7 @@ class TestGeometryVisualizers(unittest.TestCase):
         self.assertEqual(len(gamepad.button_values), 0)
         self.assertEqual(len(gamepad.axes), 0)
         meshcat.SetRealtimeRate(1.0)
+        meshcat.GetRealtimeRate()
         meshcat.Flush()
 
         meshcat.StartRecording(frames_per_second=64.0,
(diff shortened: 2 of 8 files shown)


joemasterjohn (commented):
:lgtm: one small nit. +@xuchenhan-tri for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-:-NZ6f2A68BT94vHDy5x6:b-u1f008)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 622 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-NZ6e6jH35cO9CBc3vBB:-NZ6e6jH35cO9CBc3vBC:bdq2tjq) ([raw file](https://github.com/RobotLocomotion/drake/blob/61a3e9c092422cd7c8e571faf047b4343a55acc2/geometry/test/meshcat_visualizer_test.cc#L622)):*
> ```cpp
>   // One more step causes an update.
>   simulator.AdvanceTo(0.005);
>   EXPECT_NE(meshcat->GetRealtimeRate(), slow_rate);
> ```

nit: I think it's safe enough to say that we expect the rate to be greater than the extremely slow rate. You could even measure the rate prior to the call to sleep and assert that the rate after update is relatively close to that.



_Suggestion:_
```cpp
  EXPECT_GT(meshcat->GetRealtimeRate(), slow_rate);
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700)* status: 1 unresolved discussion, LGTM missing from assignee xuchenhan-tri(platform)

___
*[`geometry/test/meshcat_visualizer_test.cc` line 622 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-NZ6e6jH35cO9CBc3vBB:-NZ6gWY4-T8QZ5DX4PQo:b-3aofcg) ([raw file](https://github.com/RobotLocomotion/drake/blob/61a3e9c092422cd7c8e571faf047b4343a55acc2/geometry/test/meshcat_visualizer_test.cc#L622)):*

nit: I think it's safe enough to say that we expect the rate to be greater than the extremely slow rate. You could even measure the rate prior to the call to sleep and assert that the rate after update is relatively close to that.
</blockquote></details>

Clarified with a comment.  It's possible (and would very likely to show up in CI) that the kernel pauses our process during this `AdvanceTo`.  There's no easy way to upper-bound the amount of wall clock spent during `AdvanceTo` here.

xuchenhan-tri (commented):
:lgtm: with one pedantic comment. Feel free to ignore.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-:-NZ79Z3P7phJc8oikuwy:bcrl0wp)* status: 2 unresolved discussions

___
*[`geometry/test/meshcat_visualizer_test.cc` line 593 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-NZ78r3N95oStCPha8vN:-NZ78r3N95oStCPha8vO:b9legtr) ([raw file](https://github.com/RobotLocomotion/drake/blob/b6b4764a8b154f2c9ecdff6beb4a7e4f2c478d59/geometry/test/meshcat_visualizer_test.cc#L593)):*
> ```cpp
>   // around publish event timing, we'll configure the visualizer to publish at
>   // 1024 Hz but we'll manually step time at 1000 Hz. This guarantees that
>   // exactly one publish event has been triggered after each one of our steps.
> ```

BTW, this is not accurate and is only true up to a certain time (which the test hasn't hit).



_Code quote:_
```cpp
  // 1024 Hz but we'll manually step time at 1000 Hz. This guarantees that
  // exactly one publish event has been triggered after each one of our steps.
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700)* status: 1 unresolved discussion

___
*[`geometry/test/meshcat_visualizer_test.cc` line 622 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19700#-NZ6e6jH35cO9CBc3vBB:-NZ7VSTI8FQQK1ninIt3:b-y38xnb) ([raw file](https://github.com/RobotLocomotion/drake/blob/61a3e9c092422cd7c8e571faf047b4343a55acc2/geometry/test/meshcat_visualizer_test.cc#L622)):*
> ... very likely to show up in CI ...

To clarify -- it certainly won't happen every time.  (My guess would be that it would happen at least one a month.)

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19700)* status: :shipit: complete! all discussions resolved, LGTM from assignees joemasterjohn,xuchenhan-tri(platform)

PR #19703, opened 2023-06-29
[py common] Add use_native_cpp_logging feature

### LINKED ISSUE #19680: [pydrake] Disable spdlog redirection when using multithreaded C++ code?

jwnimmer-tri:
Closes #19695.  Closes #19680. Relates #19685.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)

### DIFF
--- a/bindings/pydrake/common/BUILD.bazel
+++ b/bindings/pydrake/common/BUILD.bazel
@@ -526,7 +526,7 @@ drake_py_unittest(
     data = [
         ":text_logging_example",
     ],
-    shard_count = 8,
+    shard_count = 16,
     deps = [
         "//bindings/pydrake/common/test_utilities:meta_py",
     ],
@@ -537,6 +537,22 @@ drake_py_unittest(
     tags = [
         "cpu:2",
     ],
+    deps = [
+        ":module_py",
+        ":text_logging_test_helpers_py",
+    ],
+)
+
+drake_py_unittest(
+    name = "text_logging_threading_direct_to_stderr_test",
+    deps = [
+        ":module_py",
+        ":text_logging_test_helpers_py",
+    ],
+)
+
+drake_py_unittest(
+    name = "text_logging_threading_with_gil_release_test",
     deps = [
         ":module_py",
         ":text_logging_test_helpers_py",
--- a/bindings/pydrake/common/__init__.py
+++ b/bindings/pydrake/common/__init__.py
@@ -7,7 +7,7 @@ import typing
 import numpy as np
 
 from ._module_py import *
-from ._module_py import _set_log_level
+from ._module_py import _set_log_level, _use_native_cpp_logging
 
 _root_logger = _logging.getLogger()
 _drake_logger = _logging.getLogger("drake")
@@ -39,6 +39,8 @@ def _sync_spdlog_level():
     "enabled for" level would also incorporate the global "logging.disable"
     setting, but we do not reflect that back into C++ spdlog at the moment.
     """
+    if not _drake_logger._tied_to_spdlog:
+        return
     level = _drake_logger.level or _root_logger.level
     _set_log_level(_python_level_to_spdlog(level))
 
@@ -99,6 +101,33 @@ def configure_logging():
     _logging.basicConfig(level=_logging.INFO, format=format)
     _drake_logger.setLevel(_logging.NOTSET)
     _logging.addLevelName(5, "TRACE")
+
+
+def use_native_cpp_logging():
+    """By default, pydrake's C++ code routes all of its log messages to
+    Python's ``logging`` module; this function opts-out of that feature.
+
+    After this function has been called, pydrake's C++ code will log directly
+    to stderr, with no interaction with Python. (The Python settings for log
+    level threshold and message formatting will no longer affect the C++ log
+    messages.)
+
+    This can be useful to avoid C++ logging touching the GIL, e.g., when using
+    Python's ``threading`` module.
+
+    This function is not thread-safe. No other threads should be running when
+    it is called.
+
+    See also the `environment variable
+    <https://drake.mit.edu/doxygen_cxx/group__environment__variables.html>`_
+    ``DRAKE_PYTHON_LOGGING``, which can also be used to opt-out.
+
+    Raises:
+        RuntimeError: If the reconfiguration is not possible, e.g., if you have
+            already manually adjusted the C++ logging configuration.
+    """
+    _use_native_cpp_logging()
+    _drake_logger._tied_to_spdlog = False
 
 
 def _wrap_to_match_input_shape(f):
--- a/bindings/pydrake/common/module_py.cc
+++ b/bindings/pydrake/common/module_py.cc
@@ -101,6 +101,7 @@ PYBIND11_MODULE(_module_py, m) {
       doc.logging.set_log_level.doc);
 
   internal::MaybeRedirectPythonLogging();
+  m.def("_use_native_cpp_logging", &internal::UseNativeCppLogging);
 
   py::enum_<drake::ToleranceType>(m, "ToleranceType", doc.ToleranceType.doc)
       .value("kAbsolute", drake::ToleranceType::kAbsolute,
--- a/bindings/pydrake/common/test/text_logging_example.py
+++ b/bindings/pydrake/common/test/text_logging_example.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 
-from pydrake.common import configure_logging
+from pydrake.common import configure_logging, use_native_cpp_logging
 from pydrake.common.test.text_logging_test_helpers import do_log_test
 
 
@@ -11,6 +11,9 @@ def main():
     parser.add_argument(
         "--use_nice_format", metavar="1|0", type=int, required=True,
         help="Switch on (or off) nicely formatted Python output.")
+    parser.add_argument(
+        "--use_native_cpp_logging", metavar="1|0", type=int, required=True,
+        help="Whether to unhook the C++ => Python log redirection.")
     parser.add_argument(
         "--root_level", metavar="INT", type=int, required=True,
         help="Set Python root logger to this level, or use -1 for a no-op.")
@@ -22,6 +25,8 @@ def main():
     # Configure logging as instructed.
     if args.use_nice_format:
         configure_logging()
+    if args.use_native_cpp_logging:
+        use_native_cpp_logging()
     if args.root_level >= 0:
         logging.getLogger().setLevel(args.root_level)
     if args.drake_level >= 0:
--- a/bindings/pydrake/common/test/text_logging_test.py
+++ b/bindings/pydrake/common/test/text_logging_test.py
@@ -46,21 +46,25 @@ class TestTextLoggingExample(unittest.TestCase,
             CRITICAL,
         ]
         for use_nice_format in [False, True]:
-            for root_level in all_levels:
-                for drake_level in all_levels:
-                    yield dict(
-                        use_nice_format=use_nice_format,
-                        root_level=root_level,
-                        drake_level=drake_level)
+            for use_native_cpp_logging in [False, True]:
+                for root_level in all_levels:
+                    for drake_level in all_levels:
+                        yield dict(
+                            use_nice_format=use_nice_format,
+                            use_native_cpp_logging=use_native_cpp_logging,
+                            root_level=root_level,
+                            drake_level=drake_level)
 
     @run_with_multiple_values(_all_permutations())
-    def test_example(self, *, use_nice_format, root_level, drake_level):
+    def test_example(self, *, use_nice_format, use_native_cpp_logging,
+                     root_level, drake_level):
         """Runs the text_logging_example and checks its output."""
         # Invoke the test program.
         try:
             output = subprocess.check_output(
                 ["bindings/pydrake/common/text_logging_example",
                  f"--use_nice_format={int(use_nice_format)}",
+                 f"--use_native_cpp_logging={int(use_native_cpp_logging)}",
                  f"--root_level={root_level}",
                  f"--drake_level={drake_level}"],
                 stderr=subprocess.STDOUT, encoding="utf8")
@@ -69,19 +73,27 @@ class TestTextLoggingExample(unittest.TestCase,
             raise
 
         # Parse all of its output lines, e.g., the nice format:
-        #  [2022-02-11 14:49:17,952] [drake] [error] Test Error message
+        #  [2022-02-11 14:49:17,952] [drake] [ERROR] Test Error message
         # or the basic format:
-        #  ERROR:drake:Test Error message
+        #  ERROR:drake:Test Error message:
+        # or the C++ native format:
+        #  [2023-07-19 16:29:50.914] [console] [error] Test Error message
         found = []
-        if use_nice_format:
-            pattern = r"\[.{23}\] \[drake\] \[(.*?)\] (.*)"
+        if use_native_cpp_logging:
+            pattern = r"\[.{23}\] \[console\] \[([a-z]*?)\] (.*)"
+        elif use_nice_format:
+            pattern = r"\[.{23}\] \[drake\] \[([LevelA-Z0-9]*?)\] (.*)"
         else:
             pattern = r"(.*?):drake:(.*)"
         matcher = re.compile(pattern)
         for line in output.splitlines():
             match = matcher.match(line)
             self.assertTrue(match, f"{line!r} does not match {pattern}")
-            found.append(match.groups())
+            level, message = match.groups()
+            if use_native_cpp_logging:
+                self.assertTrue(level.islower(), msg=level)
+                level = level.upper()
+            found.append((level, message))
 
         # Confirm the message bodies vs the reported level.
         trace_name = "TRACE" if use_nice_format else f"Level {DRAKE_TRACE}"
@@ -142,6 +154,8 @@ class TestTextLoggingExample(unittest.TestCase,
         assert drake_level >= 0
         assert root_level >= 0
         expected_level = drake_level or root_level or DRAKE_TRACE
+        if use_native_cpp_logging:
+            expected_level = INFO
 
         # Confirm that the correct levels came out.
         self.assertEqual(effective_level, expected_level)
@@ -170,6 +184,7 @@ class TestTextLoggingExample(unittest.TestCase,
             output = subprocess.check_output(
                 ["bindings/pydrake/common/text_logging_example",
                     "--use_nice_format=1",
+                    "--use_native_cpp_logging=0",
                     f"--root_level={python_level}",
                     f"--drake_level={python_level}"],
                 stderr=subprocess.STDOUT,
--- a/bindings/pydrake/common/test/text_logging_test_helpers_py.cc
+++ b/bindings/pydrake/common/test/text_logging_test_helpers_py.cc
@@ -1,5 +1,6 @@
 #include <atomic>
 #include <chrono>
+#include <future>
 #include <thread>
 
 #include "pybind11/pybind11.h"
@@ -21,6 +22,12 @@ void do_log_test() {
   drake::log()->warn("Test Warn message");
   drake::log()->error("Test Error message");
   drake::log()->critical("Test Critical message");
+}
+
+// Logs one message at every available level, from an async worker.
+void do_log_test_async() {
+  auto handle = std::async(std::launch::async, &do_log_test);
+  handle.get();
 }
 
 // Launches a C++ thread that logs periodically.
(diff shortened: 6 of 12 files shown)


jwnimmer-tri (commented):
> Reviewed 5 of 5 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/common/test/text_logging_threading_test.py` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-N_j3ZKLBhlCX9OiZT7m:-N_j3ZKN74Ncs2XllwqY:b-9d4aak) ([raw file](https://github.com/RobotLocomotion/drake/blob/f43bc5e5d9879d8b239dd1c51575bec3e7faad1d/bindings/pydrake/common/test/text_logging_threading_test.py#L8)):*
> ```Python
> from pydrake.common.test.text_logging_test_helpers import do_log_test
> 
> # With this commented out, the test should fail (block).
> ```

@EricCousineau-TRI I tried to create a repro test case for  your bug report, but I was unable to make it deadlock.

Could you help point me in the right direction to reproduce the deadlock?

My best guess is shown below.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/common/test/text_logging_threading_test.py` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-N_j3ZKLBhlCX9OiZT7m:-NaIwTMHBpZqpJOUyntE:b-1dt2js) ([raw file](https://github.com/RobotLocomotion/drake/blob/f43bc5e5d9879d8b239dd1c51575bec3e7faad1d/bindings/pydrake/common/test/text_logging_threading_test.py#L8)):*

@EricCousineau-TRI I tried to create a repro test case for  your bug report, but I was unable to make it deadlock.

Could you help point me in the right direction to reproduce the deadlock?

My best guess is shown below.
</blockquote></details>

Working: Trying it out now.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/common/test/text_logging_threading_test.py` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-N_j3ZKLBhlCX9OiZT7m:-NaIxfvg4KnEvlsi9963:b-hhjmg4) ([raw file](https://github.com/RobotLocomotion/drake/blob/f43bc5e5d9879d8b239dd1c51575bec3e7faad1d/bindings/pydrake/common/test/text_logging_threading_test.py#L8)):*

Working: Trying it out now.
</blockquote></details>

Ah, sorry, you said it failed to deadlock, missed that.

I've commented in Anzu PR 10614, but restating my best guess of the sequence of events:
- We call something like `MoveRobotAsync(...)` causes GIL to still be actively held (default binding behavior, I believe)
- In a separate thread, the executing method calls `drake::log()->info(...)`
- Due to the redirect, this logging call tries to acquire the GIL that is already held, thus is in a deadlock

So the solutions in this case are:
(1) Not do redirect (as we are doing now)
(2) Tell our Python bindings to release the GIL in those bindings of interest

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 1 unresolved discussion, LGTM missing from assignee EricCousineau-TRI(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @jwnimmer-tri)

___
*[`bindings/pydrake/common/test/text_logging_threading_test.py` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-N_j3ZKLBhlCX9OiZT7m:-NaN1G2_9F1kJfcOC2C1:b-qo9r8f) ([raw file](https://github.com/RobotLocomotion/drake/blob/f43bc5e5d9879d8b239dd1c51575bec3e7faad1d/bindings/pydrake/common/test/text_logging_threading_test.py#L8)):*

Ah, sorry, you said it failed to deadlock, missed that.

I've commented in Anzu PR 10614, but restating my best guess of the sequence of events:
- We call something like `MoveRobotAsync(...)` causes GIL to still be actively held (default binding behavior, I believe)
- In a separate thread, the executing method calls `drake::log()->info(...)`
- Due to the redirect, this logging call tries to acquire the GIL that is already held, thus is in a deadlock

So the solutions in this case are:
(1) Not do redirect (as we are doing now)
(2) Tell our Python bindings to release the GIL in those bindings of interest
</blockquote></details>

Ah.  It sounds like the key difference that I didn't capture is that the C++ code itself launches multiple worker threads (while holding the GIL).  I think I should be able to adjust this test to try that.

EricCousineau-TRI (commented):
:lgtm_strong: with questions about `set_log_level` and `py::call_guard<py::gil_scoped_release>()`.

> Reviewed 5 of 5 files at r1, 3 of 5 files at r2, 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-:-NaNczUG2jFLsu-2Zycx:bqpjxly)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/common/__init__.py` line 10 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNc85K74JiiaX912AP:-NaNc85L9xGAoMt3qcay:b-7utg76) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/__init__.py#L10)):*
> ```Python
> 
> from ._module_py import *
> from ._module_py import _set_log_level, _use_native_cpp_logging
> ```

nit As a user debugging things, when using `use_native_cpp_logging()`, I find that I may still want to adjust logging levels.

In an Anzu dev branch, I ended up calling `pydrake.common._set_log_level` directly, which is admittedly fragile.
Is it possible to expose this function publicly, say `set_native_cpp_log_level`, and have it fail fast if _drake_loggger._tied_to_spdlog` is true?
___
*[`bindings/pydrake/common/test/text_logging_test_helpers_py.cc` line 76 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNcewu9CmAApZZn4G3:-NaNcewu9CmAApZZn4G4:b33n5pl) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/test/text_logging_test_helpers_py.cc#L76)):*
> ```cpp
> 
>   m.def("do_log_test", &do_log_test);
>   m.def("do_log_test_async", &do_log_test_async);
> ```

nit Per above, the alternative solution is to decorate each and every C++ entry point with possible multi-threaded logging with `py::call_guard<py::gil_scoped_release>()` (or something equivalent).

It may be useful to illustrate / test here. I've made an (somewhat messy) attempt here:
https://github.com/jwnimmer-tri/drake/pull/12

FTR, I prefer `use_native_cpp_logging()` as it is less to reason about, but it may be good to be aware of / test the workflow with a GIL release.
___
*[`bindings/pydrake/common/test/text_logging_threading_test.py` line 8 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-N_j3ZKLBhlCX9OiZT7m:-NaNcc8U8wgEoDL6HbDl:b556ppi) ([raw file](https://github.com/RobotLocomotion/drake/blob/f43bc5e5d9879d8b239dd1c51575bec3e7faad1d/bindings/pydrake/common/test/text_logging_threading_test.py#L8)):*

Done.

The new test case isn't as complicated as that example, but I think it sufficiently covers the use case.
</blockquote></details>

Awesome, thanks!

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/common/__init__.py` line 10 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNc85K74JiiaX912AP:-NaNeW7vBR2wtDgvDFB_:b202887) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/__init__.py#L10)):*

nit As a user debugging things, when using `use_native_cpp_logging()`, I find that I may still want to adjust logging levels.

In an Anzu dev branch, I ended up calling `pydrake.common._set_log_level` directly, which is admittedly fragile.
Is it possible to expose this function publicly, say `set_native_cpp_log_level`, and have it fail fast if _drake_loggger._tied_to_spdlog` is true?
</blockquote></details>

Alternatively, perhaps the suggested route is to configure the logging level before calling `use_native_cpp_logging()`.
Then users can use existing public interfaces? e.g.
```
logging.getLogger("drake").setLevel(logging.TRACE)
use_native_cpp_logging()
# Restore Python logging.
logging.getLogger("drake").setLevel(logging.INFO)
```

jwnimmer-tri (commented):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: needs at least two assigned reviewers

___
*[`bindings/pydrake/common/__init__.py` line 10 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNc85K74JiiaX912AP:-NaNhJAN0ASrahjF0_lX:b-e8n9vo) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/__init__.py#L10)):*

Alternatively, perhaps the suggested route is to configure the logging level before calling `use_native_cpp_logging()`.
Then users can use existing public interfaces? e.g.
```
logging.getLogger("drake").setLevel(logging.TRACE)
use_native_cpp_logging()
# Restore Python logging.
logging.getLogger("drake").setLevel(logging.INFO)
```
</blockquote></details>

I don't think we should offer public pydrake API to configure spdlog.

(1) It's not terribly valuable.  Users in the `use_native_cpp_logging()` camp are already doing their own custom witchcraft, so it shouldn't be too much of a lift for them to write and bind whatever C++ log configuration code they want.

(2) It's somewhat costly.  Coupling the spdlog API choices through into a public pydrake API will be an extra compatibility burden as spdlog evolves and we need to figure out how to keep the sugar intact and well-tested.
___
*[`bindings/pydrake/common/test/text_logging_test_helpers_py.cc` line 76 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNcewu9CmAApZZn4G3:-NaO2W1H0y95sY66NUk2:b-gnqcmv) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/test/text_logging_test_helpers_py.cc#L76)):*

nit Per above, the alternative solution is to decorate each and every C++ entry point with possible multi-threaded logging with `py::call_guard<py::gil_scoped_release>()` (or something equivalent).

It may be useful to illustrate / test here. I've made an (somewhat messy) attempt here:
https://github.com/jwnimmer-tri/drake/pull/12

FTR, I prefer `use_native_cpp_logging()` as it is less to reason about, but it may be good to be aware of / test the workflow with a GIL release.
</blockquote></details>

Done.

I added the demo, but as a separate program for utmost clarity.  (The old test has been renamed to `threading1`, with some comment improvements, but Reviewable didn't notice that.)

EricCousineau-TRI (commented):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/common/__init__.py` line 10 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNc85K74JiiaX912AP:-NaOBbjREuaAR41Xok4V:b-ba7jcz) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/__init__.py#L10)):*

I don't think we should offer public pydrake API to configure spdlog.

(1) It's not terribly valuable.  Users in the `use_native_cpp_logging()` camp are already doing their own custom witchcraft, so it shouldn't be too much of a lift for them to write and bind whatever C++ log configuration code they want.

(2) It's somewhat costly.  Coupling the spdlog API choices through into a public pydrake API will be an extra compatibility burden as spdlog evolves and we need to figure out how to keep the sugar intact and well-tested.
</blockquote></details>

Sounds good to me.
___
*[`bindings/pydrake/common/test/text_logging_test_helpers_py.cc` line 76 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaNcewu9CmAApZZn4G3:-NaOBxAuCn-XzIyritvn:bru5xav) ([raw file](https://github.com/RobotLocomotion/drake/blob/09982315a1a7b612c028055c532407a31b3bb59e/bindings/pydrake/common/test/text_logging_test_helpers_py.cc#L76)):*

Done.

I added the demo, but as a separate program for utmost clarity.  (The old test has been renamed to `threading1`, with some comment improvements, but Reviewable didn't notice that.)
</blockquote></details>

Sweet, thanks. Just some small follow-up nits on the usage of `1, 2` vs. something more descriptive.
___
*[`bindings/pydrake/common/test/text_logging_threading1_test.py` line 1 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaOBmpfCl24MtK9Hmxc:-NaOBmpfCl24MtK9Hmxd:b-k58gfz) ([raw file](https://github.com/RobotLocomotion/drake/blob/74ba41031d950ae7c80165fc128585b983756562/bindings/pydrake/common/test/text_logging_threading1_test.py#L1)):*
> ```Python
> import logging
> ```

nit Can the name of this test reflect its goal? e.g.
```
text_loging_threading_use_native_cpp_threading.py
```
___
*[`bindings/pydrake/common/test/text_logging_threading2_test.py` line 1 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NaOBs4w4Mqz957pWNyK:-NaOBs4w4Mqz957pWNyL:bputypn) ([raw file](https://github.com/RobotLocomotion/drake/blob/74ba41031d950ae7c80165fc128585b983756562/bindings/pydrake/common/test/text_logging_threading2_test.py#L1)):*
> ```Python
> import logging
> ```

nit Can the name of this test reflect its goal? e.g.
```
text_loging_threading_gil_release.py
```

EricCousineau-TRI (commented):
+@ggould-tri for platform review, please!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-:-NaOC5reChlg1IHQyMHE:bcav8qn)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform)

EricCousineau-TRI:
\cc @dmcconachie-tri for better insight + alternatives for the GIL deadlock problem

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: LGTM missing from assignee ggould-tri(platform)

ggould-tri (commented):
:lgtm_strong: ; minor notes.

> Reviewed 5 of 5 files at r1, 3 of 5 files at r2, 2 of 5 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-:-Nah-jqVBncPcRtFL9eQ:bru7ff4)* status: 2 unresolved discussions

___
*[`bindings/pydrake/common/__init__.py` line 111 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NagzRlF49nUWyVEeNvl:-NagzRlF49nUWyVEeNvm:begolmg) ([raw file](https://github.com/RobotLocomotion/drake/blob/86fce938723fce12a626c988db78b271c163b0b1/bindings/pydrake/common/__init__.py#L111)):*
> ```Python
> 
>     After this function has been called, pydrake's C++ code will log directly
>     to stderr, with no interaction with Python. (The Python settings for log
> ```

nit, super-picky:  "Directly" is misleading, as the logging still goes via spdlog and respects any C++ actions subsequently done to spdlog (e.g., level reconfiguration, or if we ever bring the `text_log_lcm_bridge` over).



_Suggestion:_
```Python
    After this function has been called, pydrake's C++ code will log
    to stderr, with no interaction with Python. (The Python settings for log
```
___
*[`bindings/pydrake/common/text_logging_pybind.h` line 14 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-Nah0_Zg8rfS7ZFirbiS:-Nah0_Zg8rfS7ZFirbiT:b-f3lvc9) ([raw file](https://github.com/RobotLocomotion/drake/blob/86fce938723fce12a626c988db78b271c163b0b1/bindings/pydrake/common/text_logging_pybind.h#L14)):*
> ```objc
> /* Implements pydrake.common.use_native_cpp_logging(). This function is not
> thread-safe; no other threads should be running when it is called. */
> void UseNativeCppLogging();
> ```

minor:  It's odd that one of these functions is named with "maybe" and the other isn't, given that they have nearly exactly identical precondition checks and warn-and-nop failure behaviour.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19703)* status: 1 unresolved discussion

___
*[`bindings/pydrake/common/__init__.py` line 111 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-NagzRlF49nUWyVEeNvl:-NahI6Y4CePfp5oHGC9V:b-dtdau6) ([raw file](https://github.com/RobotLocomotion/drake/blob/86fce938723fce12a626c988db78b271c163b0b1/bindings/pydrake/common/__init__.py#L111)):*

nit, super-picky:  "Directly" is misleading, as the logging still goes via spdlog and respects any C++ actions subsequently done to spdlog (e.g., level reconfiguration, or if we ever bring the `text_log_lcm_bridge` over).
</blockquote></details>

The text is correct insofar as "after" means "immediately after".  If someone comes by later and further changes things, that's their own fault -- for all we know, they could disable all logging entirely or send it to a file instead of stderr, so that "to stderr" on its own is no longer true.  I don't think we need to anticipate those kind of changes in the text here.
___
*[`bindings/pydrake/common/text_logging_pybind.h` line 14 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19703#-Nah0_Zg8rfS7ZFirbiS:-NahIXhQ4e9LmWlDurdZ:bo00lnu) ([raw file](https://github.com/RobotLocomotion/drake/blob/86fce938723fce12a626c988db78b271c163b0b1/bindings/pydrake/common/text_logging_pybind.h#L14)):*

minor:  It's odd that one of these functions is named with "maybe" and the other isn't, given that they have nearly exactly identical precondition checks and warn-and-nop failure behaviour.
</blockquote></details>

Their behavior is not the same.  In case of any error, the "Maybe" function will give up (do nothing and return early), but the non-"Maybe" function will throw an exception.  To my ear, that's exactly what "Maybe" always means in a function -- it's allowed to fail without throwing.

PR #19711, opened 2023-06-30
Optimize diagram loop check by using direct feedthrough maps

Xining-Du:
Trying to save intermediate system direct-feedthrough results in a map when a diagram calls `ThrowIfAlgebraicLoopsExist()` to check algebraic loops during the diagram build process. This helps to avoid repeatedly checking system direct-feedthroughs in the recursive calls. We were able to reduce the build time from ~5 mins to ~5 secs for a large diagram.

Some thoughts on the PR:
1. I could make it optional to use direct feedthrough map in diagram builder, but chose not to do so since I will need to add optional parameters in several APIs like `DiagramBuilder::Compile()` and `DiagramBuilder::ThrowIfAlgebraicLoopsExist()` which I'm not sure if is a good idea. Instead, the direct-feedthrough map is *always* enabled in this PR.
2. No additional unit tests are added here, partially because I couldn't find a good way to test public APIs with the choice made above. Instead I'm relying on existing diagram loop unit checks to confirm the correctness of the algorithm.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19711)

### DIFF
--- a/systems/framework/diagram.cc
+++ b/systems/framework/diagram.cc
@@ -72,11 +72,12 @@ Diagram<T>::get_output_port_locator(OutputPortIndex port_index) const {
 }
 
 template <typename T>
-std::multimap<int, int> Diagram<T>::GetDirectFeedthroughs() const {
+std::multimap<int, int> Diagram<T>::GetDirectFeedthroughsImpl(
+        DirectFeedThroughMap* directfeedthrough_map) const {
   std::multimap<int, int> pairs;
   for (InputPortIndex u(0); u < this->num_input_ports(); ++u) {
     for (OutputPortIndex v(0); v < this->num_output_ports(); ++v) {
-      if (DiagramHasDirectFeedthrough(u, v)) {
+      if (DiagramHasDirectFeedthrough(u, v, directfeedthrough_map)) {
         pairs.emplace(u, v);
       }
     }
@@ -1090,7 +1091,9 @@ const SystemBase& Diagram<T>::GetRootSystemBase() const {
 }
 
 template <typename T>
-bool Diagram<T>::DiagramHasDirectFeedthrough(int input_port, int output_port)
+bool Diagram<T>::DiagramHasDirectFeedthrough(int input_port, int output_port,
+  DirectFeedThroughMap* directfeedthrough_map
+)
     const {
   DRAKE_ASSERT(input_port >= 0);
   DRAKE_ASSERT(input_port < this->num_input_ports());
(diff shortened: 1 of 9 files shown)


jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19711)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @Xining-Du)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19711#-NZarx6TEeUn9zcR474w:-NZarx6TEeUn9zcR474x:b-7sinod) (no related file):*
The first step for me to understand / evaluate this change is to look at the some benchmarks, both to quantify the performance improvement and also to acutely understand how the gain is accomplished.

Do you have any benchmark program you could share (whether or not we would merge it into Drake)?

If not, could you try to explain a relevant use case, so I could create my own benchmark?  Things like how many systems, whether they are recursively nested (sub-diagrams within diagrams), a rough degree of each system (how many in/out-bound port connections), etc.

I ask because my first impression is that the equivalent speedup seems like it could be accomplished with a narrower change that's fully isolated within `diagram.cc`, without exposing any memoization argument to users in the SystemBase class.  If so, it would be a lot simpler to document, implement, and test.  Having a benchmark I can examine and experiment with would allow me to understand.

___
___
*[`systems/framework/diagram.cc` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19711#-NZau4j8DXjwvbdLfmG0:-NZau4j95zbGA3ozO4k4:b97wvpu) ([raw file](https://github.com/RobotLocomotion/drake/blob/2d5b1cb067dfa386619967a25399d53e5617928e/systems/framework/diagram.cc#L80)):*
> ```cpp
>   for (InputPortIndex u(0); u < this->num_input_ports(); ++u) {
>     for (OutputPortIndex v(0); v < this->num_output_ports(); ++v) {
>       if (DiagramHasDirectFeedthrough(u, v, directfeedthrough_map)) {
> ```

BTW My first impression is that _this_ change is the _only_ one we need for better performance -- that only the `GetDirectFeedthroughs()` needs the memoization dictionary, and nowhere else.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19711)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @Xining-Du)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19711#-NZarx6TEeUn9zcR474w:-NZvMmsdCr8SKSNH07AB:b-obv6oj) (no related file):*

The first step for me to understand / evaluate this change is to look at the some benchmarks, both to quantify the performance improvement and also to acutely understand how the gain is accomplished.

Do you have any benchmark program you could share (whether or not we would merge it into Drake)?

If not, could you try to explain a relevant use case, so I could create my own benchmark?  Things like how many systems, whether they are recursively nested (sub-diagrams within diagrams), a rough degree of each system (how many in/out-bound port connections), etc.

I ask because my first impression is that the equivalent speedup seems like it could be accomplished with a narrower change that's fully isolated within `diagram.cc`, without exposing any memoization argument to users in the SystemBase class.  If so, it would be a lot simpler to document, implement, and test.  Having a benchmark I can examine and experiment with would allow me to understand.
</blockquote></details>

See https://github.com/RobotLocomotion/drake/pull/19746.

___

PR #19716, opened 2023-07-03
Implement IsEmpty for ConvexSet and derived classes

cohnt:
This function already existed for HPolyedron, but this adds it to the base ConvexSet class, and implements it for all derived classes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -45,6 +45,7 @@ void DefineGeometryOptimization(py::module m) {
         .def("IntersectsWith", &ConvexSet::IntersectsWith, py::arg("other"),
             cls_doc.IntersectsWith.doc)
         .def("IsBounded", &ConvexSet::IsBounded, cls_doc.IsBounded.doc)
+        .def("IsEmpty", &ConvexSet::IsEmpty, cls_doc.IsEmpty.doc)
         .def("MaybeGetPoint", &ConvexSet::MaybeGetPoint,
             cls_doc.MaybeGetPoint.doc)
         .def("PointInSet", &ConvexSet::PointInSet, py::arg("x"),
@@ -133,7 +134,6 @@ void DefineGeometryOptimization(py::module m) {
             py::arg("tol") = 1E-9, cls_doc.ReduceInequalities.doc)
         .def("FindRedundant", &HPolyhedron::FindRedundant,
             py::arg("tol") = 1E-9, cls_doc.FindRedundant.doc)
-        .def("IsEmpty", &HPolyhedron::IsEmpty, cls_doc.IsEmpty.doc)
         .def("MaximumVolumeInscribedEllipsoid",
             &HPolyhedron::MaximumVolumeInscribedEllipsoid,
             cls_doc.MaximumVolumeInscribedEllipsoid.doc)
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -43,6 +43,7 @@ class TestGeometryOptimization(unittest.TestCase):
         mut.Point()
         p = np.array([11.1, 12.2, 13.3])
         point = mut.Point(p)
+        self.assertFalse(point.IsEmpty())
         self.assertEqual(point.ambient_dimension(), 3)
         np.testing.assert_array_equal(point.x(), p)
         np.testing.assert_array_equal(point.MaybeGetPoint(), p)
@@ -61,6 +62,7 @@ class TestGeometryOptimization(unittest.TestCase):
         np.testing.assert_array_equal(hpoly.A(), self.A)
         np.testing.assert_array_equal(hpoly.b(), self.b)
         self.assertTrue(hpoly.PointInSet(x=[0, 0, 0], tol=0.0))
+        self.assertFalse(hpoly.IsEmpty())
         self.assertFalse(hpoly.IsBounded())
         new_vars, new_constraints = hpoly.AddPointInSetConstraints(
             self.prog, self.x)
@@ -166,6 +168,7 @@ class TestGeometryOptimization(unittest.TestCase):
         mut.Hyperellipsoid()
         ellipsoid = mut.Hyperellipsoid(A=self.A, center=self.b)
         self.assertEqual(ellipsoid.ambient_dimension(), 3)
+        self.assertFalse(ellipsoid.IsEmpty())
         np.testing.assert_array_equal(ellipsoid.A(), self.A)
         np.testing.assert_array_equal(ellipsoid.center(), self.b)
         self.assertTrue(ellipsoid.PointInSet(x=self.b, tol=0.0))
@@ -215,6 +218,7 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertEqual(sum2.ambient_dimension(), 3)
         self.assertEqual(sum2.num_terms(), 2)
         self.assertIsInstance(sum2.term(0), mut.Point)
+        self.assertFalse(sum.IsEmpty())
 
     def test_spectrahedron(self):
         s = mut.Spectrahedron()
@@ -229,6 +233,7 @@ class TestGeometryOptimization(unittest.TestCase):
         mut.VPolytope()
         vertices = np.array([[0.0, 1.0, 2.0], [3.0, 7.0, 5.0]])
         vpoly = mut.VPolytope(vertices=vertices)
+        self.assertFalse(vpoly.IsEmpty())
         self.assertEqual(vpoly.ambient_dimension(), 2)
         np.testing.assert_array_equal(vpoly.vertices(), vertices)
         self.assertTrue(vpoly.PointInSet(x=[1.0, 5.0], tol=1e-8))
@@ -307,6 +312,7 @@ class TestGeometryOptimization(unittest.TestCase):
         h_box = mut.HPolyhedron.MakeBox(
             lb=[-1, -1, -1], ub=[1, 1, 1])
         sum = mut.CartesianProduct(setA=point, setB=h_box)
+        self.assertFalse(sum.IsEmpty())
         self.assertEqual(sum.ambient_dimension(), 6)
         self.assertEqual(sum.num_factors(), 2)
         sum2 = mut.CartesianProduct(sets=[point, h_box])
@@ -325,6 +331,7 @@ class TestGeometryOptimization(unittest.TestCase):
         h_box = mut.HPolyhedron.MakeBox(
             lb=[-1, -1, -1], ub=[1, 1, 1])
         intersect = mut.Intersection(setA=point, setB=h_box)
+        self.assertFalse(intersect.IsEmpty())
         self.assertEqual(intersect.ambient_dimension(), 3)
         self.assertEqual(intersect.num_elements(), 2)
         intersect2 = mut.Intersection(sets=[point, h_box])
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -114,6 +114,18 @@ bool CartesianProduct::DoIsBounded() const {
   return true;
 }
 
+bool CartesianProduct::DoIsEmpty() const {
+  if (sets_.size() == 0) {
+    return true;
+  }
+  for (const auto& s : sets_) {
+    if (s->IsEmpty()) {
+      return true;
+    }
+  }
+  return false;
+}
+
 std::optional<VectorXd> CartesianProduct::DoMaybeGetPoint() const {
   // Check if all sets are points.
   std::vector<VectorXd> points;
--- a/geometry/optimization/cartesian_product.h
+++ b/geometry/optimization/cartesian_product.h
@@ -77,6 +77,8 @@ class CartesianProduct final : public ConvexSet {
 
   bool DoIsBounded() const final;
 
+  bool DoIsEmpty() const final;
+
   std::optional<Eigen::VectorXd> DoMaybeGetPoint() const final;
 
   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
--- a/geometry/optimization/convex_set.cc
+++ b/geometry/optimization/convex_set.cc
@@ -31,6 +31,27 @@ bool ConvexSet::IntersectsWith(const ConvexSet& other) const {
   return result.is_success();
 }
 
+bool ConvexSet::DoIsEmpty() const {
+  solvers::MathematicalProgram prog;
+  auto point = prog.NewContinuousVariables(ambient_dimension());
+  AddPointInSetConstraints(&prog, point);
+  auto result = solvers::Solve(prog);
+  auto status = result.get_solution_result();
+  switch (status) {
+    case solvers::SolutionResult::kSolutionFound:
+      return false;
+    case solvers::SolutionResult::kInfeasibleConstraints:
+    case solvers::SolutionResult::kInfeasibleOrUnbounded:
+      return true;
+    default:
+      throw std::runtime_error(
+          fmt::format("ConvexSet::IsEmpty() has solution result {}. "
+                      "We are uncertain if the set if empty or not.",
+                      status));
+      return true;
+  }
+}
+
 std::optional<Eigen::VectorXd> ConvexSet::DoMaybeGetPoint() const {
   return std::nullopt;
 }
--- a/geometry/optimization/convex_set.h
+++ b/geometry/optimization/convex_set.h
@@ -90,6 +90,15 @@ class ConvexSet : public ShapeReifier {
     return DoIsBounded();
   }
 
+  /** Returns true iff the set is empty. Note: for some derived classes, this
+  check is trivial, but for others, it can require solving a (typically small)
+  optimization problem. Check the derived class documentation for any notes.
+  @throws std::exception if ambient_dimension() == 0 */
+  bool IsEmpty() const {
+    DRAKE_THROW_UNLESS(ambient_dimension() > 0);
+    return DoIsEmpty();
+  }
+
   /** If this set trivially contains exactly one point, returns the value of
   that point. Otherwise, returns nullopt. When ambient_dimension is zero,
   returns nullopt. By "trivially", we mean that representation of the set
@@ -212,6 +221,11 @@ class ConvexSet : public ShapeReifier {
   /** Non-virtual interface implementation for IsBounded().
   @pre ambient_dimension() > 0 */
   virtual bool DoIsBounded() const = 0;
+
+  /** Non-virtual interface implementation for IsEmpty(). The default
+  implementation solves a feasibility optimization problem, but derived
+  classes can override with a custom (more efficient) implementation. */
+  virtual bool DoIsEmpty() const;
 
   /** Non-virtual interface implementation for MaybeGetPoint(). The default
   implementation returns nullopt. Sets that can model a single point should
--- a/geometry/optimization/hpolyhedron.cc
+++ b/geometry/optimization/hpolyhedron.cc
@@ -54,16 +54,6 @@ std::tuple<bool, solvers::MathematicalProgramResult> IsInfeasible(
           result};
 }
 
-// Checks if Ax ≤ b defines an empty set.
-bool DoIsEmpty(const Eigen::Ref<const MatrixXd>& A,
-               const Eigen::Ref<const VectorXd>& b) {
-  solvers::MathematicalProgram prog;
-  solvers::VectorXDecisionVariable x =
-      prog.NewContinuousVariables(A.cols(), "x");
-  prog.AddLinearConstraint(A, VectorXd::Constant(b.rows(), -kInf), b, x);
-  return std::get<0>(IsInfeasible(prog));
-}
-
 /* Checks whether the constraint cᵀ x ≤ d is already implied by the linear
  constraints in prog. This is done by solving a small linear program
  and modifying the coefficients of  `new_constraint` binding. This method may
@@ -363,11 +353,19 @@ bool HPolyhedron::DoIsBounded() const {
   return result.is_success();
 }
 
+bool HPolyhedron::DoIsEmpty() const {
+  solvers::MathematicalProgram prog;
+  solvers::VectorXDecisionVariable x =
+      prog.NewContinuousVariables(A_.cols(), "x");
+  prog.AddLinearConstraint(A_, VectorXd::Constant(b_.rows(), -kInf), b_, x);
+  return std::get<0>(IsInfeasible(prog));
+}
+
 bool HPolyhedron::ContainedIn(const HPolyhedron& other, double tol) const {
   DRAKE_THROW_UNLESS(other.A().cols() == A_.cols());
   // `this` defines an empty set and therefore is contained in any `other`
   // HPolyhedron.
-  if (DoIsEmpty(A_, b_)) {
+  if (DoIsEmpty()) {
     return true;
   }
 
@@ -439,8 +437,9 @@ HPolyhedron HPolyhedron::DoIntersectionWithChecks(const HPolyhedron& other,
       A.row(num_kept) = other.A().row(i);
       b.row(num_kept) = other.b().row(i);
       ++num_kept;
-      if (DoIsEmpty(A.topRows(num_kept), b.topRows(num_kept))) {
-        return {A.topRows(num_kept), b.topRows(num_kept)};
+      HPolyhedron maybe_empty(A.topRows(num_kept), b.topRows(num_kept));
+      if (maybe_empty.IsEmpty()) {
+        return maybe_empty;
       }
     }
   }
@@ -520,10 +519,6 @@ std::set<int> HPolyhedron::FindRedundant(double tol) const {
     }
   }
   return redundant_indices;
-}
-
-bool HPolyhedron::IsEmpty() const {
-  return DoIsEmpty(A_, b_);
 }
 
 std::unique_ptr<ConvexSet> HPolyhedron::DoClone() const {
--- a/geometry/optimization/hpolyhedron.h
+++ b/geometry/optimization/hpolyhedron.h
@@ -103,9 +103,6 @@ class HPolyhedron final : public ConvexSet {
   negative tol means it is less likely to remote a constraint.  */
   [[nodiscard]] HPolyhedron ReduceInequalities(double tol = 1E-9) const;
 
-  /** Checks if this HPolyhedron defines an empty set.  */
-  [[nodiscard]] bool IsEmpty() const;
-
   /** Solves a semi-definite program to compute the inscribed ellipsoid.
   From Section 8.4.2 in Boyd and Vandenberghe, 2004, we solve
   @verbatim
@@ -211,6 +208,8 @@ class HPolyhedron final : public ConvexSet {
   std::unique_ptr<ConvexSet> DoClone() const final;
 
   bool DoIsBounded() const final;
+
+  bool DoIsEmpty() const final;
 
   // N.B. No need to override DoMaybeGetPoint here.
 
--- a/geometry/optimization/hyperellipsoid.cc
+++ b/geometry/optimization/hyperellipsoid.cc
@@ -180,6 +180,10 @@ bool Hyperellipsoid::DoIsBounded() const {
   return qr.dimensionOfKernel() == 0;
 }
 
+bool Hyperellipsoid::DoIsEmpty() const {
+  return false;
+}
+
 bool Hyperellipsoid::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
                                   double tol) const {
   DRAKE_DEMAND(A_.cols() == x.size());
--- a/geometry/optimization/hyperellipsoid.h
+++ b/geometry/optimization/hyperellipsoid.h
@@ -109,6 +109,8 @@ class Hyperellipsoid final : public ConvexSet {
 
   bool DoIsBounded() const final;
 
+  bool DoIsEmpty() const final;
+
   // N.B. No need to override DoMaybeGetPoint here.
 
   bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
--- a/geometry/optimization/minkowski_sum.cc
+++ b/geometry/optimization/minkowski_sum.cc
@@ -103,6 +103,19 @@ bool MinkowskiSum::DoIsBounded() const {
   return true;
 }
 
+bool MinkowskiSum::DoIsEmpty() const {
+  if (sets_.size() == 0) {
+    return true;
+  }
+  // The empty set is annihilatory in Minkowski addition.
+  for (const auto& s : sets_) {
+    if (s->IsEmpty()) {
+      return true;
+    }
+  }
+  return false;
+}
+
 std::optional<VectorXd> MinkowskiSum::DoMaybeGetPoint() const {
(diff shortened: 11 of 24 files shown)


hongkai-dai (commented):
BTW, I think you could provide an implementation in the base class
```cc
bool ConvexSet::DoIsEmpty() const {
  solvers::MathematicalProgram prog;
  auto pt = prog.NewContinuousVariable(...);
  AddPointInSetConstraints(&prog, pt);
  auto result = solvers::Solve(prog);
  return !result.IsSuccess();
}
```

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-:-NZS-DHUBXjRSwD-XX1a:b-skmdf8)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

cohnt:
Done.

Also, +hongkai-dai for review

svenevs:

I deleted your runner by accident, CI is either off or on with ventura and I deleted yours rather than mine... sorry xD (we're updating ventura image right now)

svenevs:
status update: [check slack](https://drakedevelopers.slack.com/archives/C270MN28G/p1688137097344999), there is another delay sorry :/

cohnt:
No problem :)

hongkai-dai (commented):
> Reviewed 17 of 24 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.h` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZWz2bq81bVV3H2eA2H:-NZWz2bq81bVV3H2eA2I:b-9av1kj) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/convex_set.h#L224)):*
> ```objc
>   implementation solves a feasibility optimization problem, but derived
>   classes can override with a custom (more efficient) implementation. */
>   virtual bool DoIsEmpty() const = 0;
> ```

Now that you have a base class implementation, I don't think `=0` is needed here?
___
*[`geometry/optimization/convex_set.cc` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZWz7ZV2gzOPI-lf3kT:-NZWz7ZV2gzOPI-lf3kU:b9mphuh) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/convex_set.cc#L39)):*
> ```cpp
>   AddPointInSetConstraints(&prog, pt);
>   auto result = solvers::Solve(prog);
>   return !result.is_success();
> ```

Thinking more about it, I think here we should check if the problem is feasible. So return something like this
```cc
switch (result.get_solution_result()) {
  case solvers::SolutionResult::kSolutionFound:
    return true;
  case solvers::SolutionResult::kInfeasibleConstraints:
  case solvers::SolutionResult::kInfeasibleOrUnbounded:
    return false;
  default:
    throw std::runtime_error(fmt::format("ConvexSet::IsEmpty() has solution result {}. We are uncertain if the set if empty or not.", result.get_solution_result));
}
```
When `result.is_success()= false`, it can be many reasons, for example running out of iterations limit. So we don't know for sure if the set is empty or not.
___
*[`geometry/optimization/intersection.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZX0EdcDLmmFOaK7zEY:-NZX0EdcDLmmFOaK7zEZ:b4s5l02) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/intersection.cc#L78)):*
> ```cpp
>     return false;
>   }
>   solvers::MathematicalProgram prog;
> ```

This part is `DoIsEmpty()` in the base class. I think you can call
```cc
return ConvexSet::IsEmpty();
```
___
*[`geometry/optimization/spectrahedron.cc` line 84 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZX0n2s2Du-6M8P9bB3:-NZX0n2tCxeaZhWkQ5Tr:bnv5007) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/spectrahedron.cc#L84)):*
> ```cpp
> }
> 
> bool Spectrahedron::DoIsEmpty() const {
> ```

Why not to call the base class `ConvexSet::IsEmpty()`?

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 4 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.h` line 224 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZWz2bq81bVV3H2eA2H:-NZXIjg14yxNmOZASqy2:bpvljfs) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/convex_set.h#L224)):*

Now that you have a base class implementation, I don't think `=0` is needed here?
</blockquote></details>

Ah, yes. My C++ is pretty rusty. Fixed!
___
*[`geometry/optimization/convex_set.cc` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZWz7ZV2gzOPI-lf3kT:-NZXIo5y1FG94Z5i4BSW:bugpcva) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/convex_set.cc#L39)):*

Thinking more about it, I think here we should check if the problem is feasible. So return something like this
```cc
switch (result.get_solution_result()) {
  case solvers::SolutionResult::kSolutionFound:
    return true;
  case solvers::SolutionResult::kInfeasibleConstraints:
  case solvers::SolutionResult::kInfeasibleOrUnbounded:
    return false;
  default:
    throw std::runtime_error(fmt::format("ConvexSet::IsEmpty() has solution result {}. We are uncertain if the set if empty or not.", result.get_solution_result));
}
```
When `result.is_success()= false`, it can be many reasons, for example running out of iterations limit. So we don't know for sure if the set is empty or not.
</blockquote></details>

Good thinking! I just had to change the logic a bit. (If a solution is found, it's nonempty, so should return false.)
___
*[`geometry/optimization/intersection.cc` line 78 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZX0EdcDLmmFOaK7zEY:-NZXIx0-ED3-0YZl0Jcr:b-wit3gt) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/intersection.cc#L78)):*

This part is `DoIsEmpty()` in the base class. I think you can call
```cc
return ConvexSet::IsEmpty();
```
</blockquote></details>

Originally, I needed to handle the empty case separately for Intersection. I've moved the ambient_dimension check into the base class, so it should work for everything now.
___
*[`geometry/optimization/spectrahedron.cc` line 84 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZX0n2s2Du-6M8P9bB3:-NZXJAGL7eZHiAnXpDZE:b-2bwaqi) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/spectrahedron.cc#L84)):*

Why not to call the base class `ConvexSet::IsEmpty()`?
</blockquote></details>

Done! I wasn't sure about the Spectrahedron logic, but the test cases suggest that this works.

hongkai-dai (commented):
:lgtm: +@sherm1 for Wednesday's platform review please, thanks!

> Reviewed 1 of 24 files at r1, 5 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-:-NZXKbkb5DAV_lke57Ou:b-7g5p1o)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.cc` line 39 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZWz7ZV2gzOPI-lf3kT:-NZXKMaw2ekSi33iOHq8:bda0zit) ([raw file](https://github.com/RobotLocomotion/drake/blob/99a572d2d30bc0fefb1635f31d41753bc94df201/geometry/optimization/convex_set.cc#L39)):*

Good thinking! I just had to change the logic a bit. (If a solution is found, it's nonempty, so should return false.)
</blockquote></details>

Ah right, I forgot to flip the logic here.

jwnimmer-tri:
+@jwnimmer-tri taking platform review for -@sherm1.

jwnimmer-tri (commented):
> Reviewed 18 of 24 files at r1, 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZaYBAg5df0_3BqhATj:b-rhanp8) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*
> ```objc
>   /** Returns true iff the set is empty. Note: for some derived classes, this
>   check is trivial, but for others, it can require solving a (typically small)
>   optimization problem. Check the derived class documentation for any notes. */
> ```

It's unclear from these docs whether a set with ambient dimension of zero would be considered "empty" or not.

When the ambient dimension is zero, we have that `PointInSet(x)` always returns false for all `x`.  That seems like `empty == true` to me?   For sure, the docs here should explain what the return value is in this case.  It might also be worth elaborating somewhere on _why_ we chose true (or false, or throwing) for the degenerate case.
___
*[`geometry/optimization/convex_set.h` line 96 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYlwE0wQLF6mZe6Of:-NZaYlwE0wQLF6mZe6Og:bd6mdzg) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L96)):*
> ```objc
>   check is trivial, but for others, it can require solving a (typically small)
>   optimization problem. Check the derived class documentation for any notes. */
>   bool IsEmpty() const { return DoIsEmpty(); }
> ```

For most of the other NVI pairs (Foo + DoFoo), the "ambient dimension is zero" case is dealt with in the base class, and then the DoFoo function has a precondition that the dimension is strictly greater than zero.

This makes it very easy to reason about whether the degenerate case is handled correctly, by checking degenerate-case code in just one place, instead of every single subclass. 

Here's the code change I have in mind.  (Further docs changes would also be required if we go this route.)

If we _don't_ need a guard in the base class for some reason, there should be an implementation comment to clarify why not.



_Suggestion:_
```objc
bool IsEmpty() const {
  if (ambient_dimension() == 0) {
    return TBD /* ... whatever boolean constant we pick */;
  }
  return DoIsEmpty();
}
```
___
*[`geometry/optimization/convex_set.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaZnIf0Qg4xe92GPHw:-NZaZnIf0Qg4xe92GPHx:bi7o64a) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.cc#L41)):*
nit

> Do not abbreviate by deleting letters within a word.

-- https://drake.mit.edu/styleguide/cppguide.html#General_Naming_Rules



_Suggestion:_
```cpp
point
```
___
*[`geometry/optimization/hpolyhedron.cc` line 444 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZabOhj9jlXg_CZWV8G:-NZabOhj9jlXg_CZWV8H:bnctc2o) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/hpolyhedron.cc#L444)):*
> ```cpp
>       if (HPolyhedron(A_keep, b_keep).IsEmpty()) {
>         return {A_keep, b_keep};
>       }
> ```

BTW It seems to me like we're unnecessarily creating the result object twice?



_Suggestion:_
```cpp
      HPolyhedron maybe_empty(A.topRows(num_kept), b.topRows(num_kept));
      if (maybe_empty.IsEmpty()) {
        return maybe_empty;
      }
```
___
*[`geometry/optimization/hyperellipsoid.h` line 112 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaaB6DCQwj9W2sX0db:-NZaaB6DCQwj9W2sX0dc:b-q6h9c8) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/hyperellipsoid.h#L112)):*
nit Overrides of virtual functions should generally never be declared inline.

The other classes that also return a constant should likewise be changed.

See https://drake.mit.edu/styleguide/cppguide.html#Inline_Functions:

> ... function definitions should customarily be placed into `*.cc` files; only if there is an acute reason to put it in a header may it appear there.

This function does not meet condition 1 (it's not snake_case getter) nor 5a (because its virtual).



_Suggestion:_
```objc
bool DoIsEmpty() const final;
```
___
*[`geometry/optimization/intersection.cc` line 7 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaagdk4WX5URIXpiSe:-NZaagdmCflt5Oe3SFzc:b-ra1vwc) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/intersection.cc#L7)):*
> ```cpp
> #include <fmt/format.h>
> 
> #include "drake/solvers/solve.h"
> ```

nit Unused include.
___
*[`geometry/optimization/minkowski_sum.cc` line 110 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZabsn64iGqV7jf03iZ:-NZabsn64iGqV7jf03i_:b72rxv2) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/minkowski_sum.cc#L110)):*
> ```cpp
>     return true;
>   }
>   // The empty set is annihilatory in Minkowski addition
> ```

nit punctuation



_Suggestion:_
```cpp
  // The empty set is annihilatory in Minkowski addition.
```
___
*[`geometry/optimization/vpolytope.h` line 92 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZa_Q4Q2Ba30OF7npqZ:-NZa_Q4Q2Ba30OF7npq_:b-qtusyx) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/vpolytope.h#L92)):*
> ```objc
>   bool DoIsBounded() const final { return true; }
> 
>   bool DoIsEmpty() const final { return false; }
> ```

What if we have `vertices_.size() == 0`?  That seems like it should be empty?
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 26 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZadhtLE5cgHD2UFdxG:-NZadhtLE5cgHD2UFdxH:baz2qnj) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/test/spectrahedron_test.cc#L26)):*
> ```cpp
>   Spectrahedron spect;
>   EXPECT_EQ(spect.ambient_dimension(), 0);
>   EXPECT_FALSE(spect.IsEmpty());
> ```

BTW It's a little weird for the test case named "Empty" to not be empty.  Might merit an explanatory comment?
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 246 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZadppt3_YalSJwiBxY:-NZadppt3_YalSJwiBxZ:b-xa3it6) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/test/spectrahedron_test.cc#L246)):*
> ```cpp
> }
> 
> GTEST_TEST(SpectrahedronTest, Empty2) {
> ```

nit We can be more descriptive here.



_Suggestion:_
```cpp
NontriviallyEmpty
```

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZbFc391PDMRGTRxqTg:b-318j7t) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*

It's unclear from these docs whether a set with ambient dimension of zero would be considered "empty" or not.

When the ambient dimension is zero, we have that `PointInSet(x)` always returns false for all `x`.  That seems like `empty == true` to me?   For sure, the docs here should explain what the return value is in this case.  It might also be worth elaborating somewhere on _why_ we chose true (or false, or throwing) for the degenerate case.
</blockquote></details>

TL;DR If it's alright with you, I'll leave existing interfaces unchanged, and create a separate issue to discuss this.

This is a really good point to raise, and actually gets into some subtle issues regarding the difference between an empty set and a zero-dimensional set.

The set {0}=R^0 is the unique vector space of dimension zero. The empty set is not a vector space at all (since a vector space must always include the zero vector), and is often assigned the dimension -1 (although this is just a convention). It's important to treat these objects distinctly -- for example, an empty HPolyhedron still keeps track of its ambient dimension.

I was wary of messing with the behavior of such things as PointInSet, but if I were rewriting it, I would actually prefer that it returns true. This is because the only input that could be accepted without throwing an error due to a dimension mismatch is a zero-length vector, and I think it makes sense to treat that as the unique zero point in R^0. Then the following python script would exhibit consistent behavior for any acceptable dimension:
```
import numpy as np
from pydrake.all import HPolyhedron

for dim in range(3, -1, -1):
	A = np.zeros((0,dim))
	b = np.zeros(0)
	h = HPolyhedron(A, b)
	print(h.ambient_dimension(), h.PointInSet(np.zeros(dim)))
```

Likewise, I would also consider changing the behavior of IntersectsWith to return true if both sets are {0}. And sets initialized with empty lists (MinkowskiSum, CartesianProduct, Intersection) should be considered to be empty, and have ambient_dimension -1. But I think changing this interface is probably beyond the scope of this PR, and I'd want to hear what others think.
___
*[`geometry/optimization/convex_set.h` line 96 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYlwE0wQLF6mZe6Of:-NZbGOwW59MGI8yp9RUt:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L96)):*

For most of the other NVI pairs (Foo + DoFoo), the "ambient dimension is zero" case is dealt with in the base class, and then the DoFoo function has a precondition that the dimension is strictly greater than zero.

This makes it very easy to reason about whether the degenerate case is handled correctly, by checking degenerate-case code in just one place, instead of every single subclass. 

Here's the code change I have in mind.  (Further docs changes would also be required if we go this route.)

If we _don't_ need a guard in the base class for some reason, there should be an implementation comment to clarify why not.
</blockquote></details>

Done.
___
*[`geometry/optimization/convex_set.cc` line 41 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaZnIf0Qg4xe92GPHw:-NZbHGUr04tw_-DWHJI1:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.cc#L41)):*

nit

> Do not abbreviate by deleting letters within a word.

-- https://drake.mit.edu/styleguide/cppguide.html#General_Naming_Rules
</blockquote></details>

Done.
___
*[`geometry/optimization/hpolyhedron.cc` line 444 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZabOhj9jlXg_CZWV8G:-NZbGUxp2z9oI3PNG4LT:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/hpolyhedron.cc#L444)):*

BTW It seems to me like we're unnecessarily creating the result object twice?
</blockquote></details>

Done.
___
*[`geometry/optimization/hyperellipsoid.h` line 112 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaaB6DCQwj9W2sX0db:-NZbGZFg8klfHV-N3-GN:b9yefvp) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/hyperellipsoid.h#L112)):*

nit Overrides of virtual functions should generally never be declared inline.

The other classes that also return a constant should likewise be changed.

See https://drake.mit.edu/styleguide/cppguide.html#Inline_Functions:

> ... function definitions should customarily be placed into `*.cc` files; only if there is an acute reason to put it in a header may it appear there.

This function does not meet condition 1 (it's not snake_case getter) nor 5a (because its virtual).
</blockquote></details>

Fixed. I also fixed several cases where DoIsBounded was declared inline.
___
*[`geometry/optimization/minkowski_sum.cc` line 110 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZabsn64iGqV7jf03iZ:-NZbHIsPErSqjVMyupWt:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/minkowski_sum.cc#L110)):*

nit punctuation
</blockquote></details>

Done.
___
*[`geometry/optimization/vpolytope.h` line 92 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZa_Q4Q2Ba30OF7npqZ:-NZbGm4K3Rx_ifFpQOqE:b-u7eovp) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/vpolytope.h#L92)):*

What if we have `vertices_.size() == 0`?  That seems like it should be empty?
</blockquote></details>

Good catch -- thanks! I've also added a test case to check this behavior.
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 26 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZadhtLE5cgHD2UFdxG:-NZbGoLcBXF1G3mD2bKE:b-qfdn7c) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/test/spectrahedron_test.cc#L26)):*

BTW It's a little weird for the test case named "Empty" to not be empty.  Might merit an explanatory comment?
</blockquote></details>

It was this way on Drake master. I've updated the name to be more in-line with the other ConvexSet test cases.
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 246 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZadppt3_YalSJwiBxY:-NZbHAwe6xOHBdV7-ic3:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/test/spectrahedron_test.cc#L246)):*

nit We can be more descriptive here.
</blockquote></details>

Done.
___
*[`geometry/optimization/intersection.cc` line 7 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaagdk4WX5URIXpiSe:-NZbH9Z53-bzo00dpv35:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/intersection.cc#L7)):*

nit Unused include.
</blockquote></details>

Done.

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 15 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-:-NZbJ50z5Vc-9sIbJ9sa:b-vtj4vo)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZbJ7NLC7y-53AAju0c:b-cohii5) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*

TL;DR If it's alright with you, I'll leave existing interfaces unchanged, and create a separate issue to discuss this.

This is a really good point to raise, and actually gets into some subtle issues regarding the difference between an empty set and a zero-dimensional set.

The set {0}=R^0 is the unique vector space of dimension zero. The empty set is not a vector space at all (since a vector space must always include the zero vector), and is often assigned the dimension -1 (although this is just a convention). It's important to treat these objects distinctly -- for example, an empty HPolyhedron still keeps track of its ambient dimension.

I was wary of messing with the behavior of such things as PointInSet, but if I were rewriting it, I would actually prefer that it returns true. This is because the only input that could be accepted without throwing an error due to a dimension mismatch is a zero-length vector, and I think it makes sense to treat that as the unique zero point in R^0. Then the following python script would exhibit consistent behavior for any acceptable dimension:
```
import numpy as np
from pydrake.all import HPolyhedron

for dim in range(3, -1, -1):
	A = np.zeros((0,dim))
	b = np.zeros(0)
	h = HPolyhedron(A, b)
	print(h.ambient_dimension(), h.PointInSet(np.zeros(dim)))
```

Likewise, I would also consider changing the behavior of IntersectsWith to return true if both sets are {0}. And sets initialized with empty lists (MinkowskiSum, CartesianProduct, Intersection) should be considered to be empty, and have ambient_dimension -1. But I think changing this interface is probably beyond the scope of this PR, and I'd want to hear what others think.
</blockquote></details>

I would defer to @hongkai-dai (or Russ) to decide what to do here.

If we are not sure yet what to do in the 0-dim case, we could forbid it instead of choose one or the other in this first PR.  (Forbidding it would be having it as `@pre ambient... > 0` in the docs, and throwing here in the base class if we see zero.)

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZbK1T9BKZ-o_U_75Mo:b-64p7hr) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*

I would defer to @hongkai-dai (or Russ) to decide what to do here.

If we are not sure yet what to do in the 0-dim case, we could forbid it instead of choose one or the other in this first PR.  (Forbidding it would be having it as `@pre ambient... > 0` in the docs, and throwing here in the base class if we see zero.)
</blockquote></details>

I think mathematically we could define a 0-dimensional space. A set in this space could be empty or not. But that leads to a lot of potential confusion for a coder to think about it (a set in a zero dimensional space, is that an empty set or not? I will always need to pause and think about it) I like Jeremy's proposal to just forbid having a 0-dim case at the constructor of `ConvexSet`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZbMrp51a7rxjDq5JnJ:b-sv0a7c) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*
> ... forbid ... at the constructor of `ConvexSet`.

I did not mean to suggest that.  My suggestion was that `IsEmpty` could throw, in particular.

We cannot disallow 0-dim sets existing in the first place.  They occur all over the place.  Our only choice is to specify what each operation (function) should do about them when it sees one.

See also https://github.com/RobotLocomotion/drake/pull/19279#pullrequestreview-1406745625 for related discussion.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.h` line 95 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-NZaYBAg5df0_3BqhATi:-NZbQIHG8KpAxe_Ss4QF:b-64gdee) ([raw file](https://github.com/RobotLocomotion/drake/blob/121ddb14709f74e2696284ca5116efa953b01b65/geometry/optimization/convex_set.h#L95)):*

> ... forbid ... at the constructor of `ConvexSet`.

I did not mean to suggest that.  My suggestion was that `IsEmpty` could throw, in particular.

We cannot disallow 0-dim sets existing in the first place.  They occur all over the place.  Our only choice is to specify what each operation (function) should do about them when it sees one.

See also https://github.com/RobotLocomotion/drake/pull/19279#pullrequestreview-1406745625 for related discussion.
</blockquote></details>

If it's alright with both of you, I'd like to centralize the discussion here: #19723

sherm1 (commented):
Thanks for taking this one, Jeremy!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716#-:-NZblMJ3BknpM98qK4G2:b-rtyqn8)* status: 1 unresolved discussion

cohnt:
For now, IsEmpty will throw an error if the ambient dimension is zero. Later on, we can discuss the best way to implement that behavior in #19723.

jwnimmer-tri (commented):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

hongkai-dai (commented):
> Reviewed 8 of 15 files at r4, 2 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19716)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

PR #19719, opened 2023-07-04
GraphOfConvexSets returns kInfeasibleConstraints when no path exists.

RussTedrake:
This fixes some small bugs in our handling of the case where no path exists.

(I found this when toying with the default parameter changes, which I will PR once this lands).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719)

### DIFF
--- a/geometry/optimization/graph_of_convex_sets.cc
+++ b/geometry/optimization/graph_of_convex_sets.cc
@@ -668,6 +668,10 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
   std::map<EdgeId, Variable> relaxed_phi;
   std::vector<Variable> excluded_phi;
 
+  // The flow constraints below assume that we have some edge out of the source
+  // and into the target, so we handle that case explicitly.
+  bool has_edges_out_of_source = false;
+  bool has_edges_into_target = false;
   for (const auto& [edge_id, e] : edges_) {
     // If an edge is turned off (ϕ = 0) or excluded by preprocessing, don't
     // include it in the optimization.
@@ -680,6 +684,12 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
         excluded_phi.push_back(phi);
       }
       continue;
+    }
+    if (e->u().id() == source_id) {
+      has_edges_out_of_source = true;
+    }
+    if (e->v().id() == target_id) {
+      has_edges_into_target = true;
     }
     outgoing_edges[e->u().id()].emplace_back(e.get());
     incoming_edges[e->v().id()].emplace_back(e.get());
@@ -743,6 +753,18 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
       // (on the vertices).
       AddPerspectiveConstraint(&prog, b, vars);
     }
+  }
+  if (!has_edges_out_of_source) {
+    MathematicalProgramResult result;
+    log()->info("Source vertex {} has no outgoing edges.", source_id);
+    result.set_solution_result(SolutionResult::kInfeasibleConstraints);
+    return result;
+  }
+  if (!has_edges_into_target) {
+    MathematicalProgramResult result;
+    log()->info("Target vertex {} has no incoming edges.", target_id);
+    result.set_solution_result(SolutionResult::kInfeasibleConstraints);
+    return result;
   }
 
   for (const std::pair<const VertexId, std::unique_ptr<Vertex>>& vpair :
@@ -929,7 +951,7 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
       ++num_trials;
 
       // Find candidate path by traversing the graph with a depth first search
-      // where edges are taken with prbability proportional to their flow.
+      // where edges are taken with probability proportional to their flow.
       std::vector<VertexId> visited_vertex_ids{source_id};
       std::vector<VertexId> path_vertex_ids{source_id};
       std::vector<const Edge*> new_path;
@@ -948,6 +970,9 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
         if (candidate_edges.size() == 0) {
           path_vertex_ids.pop_back();
           new_path.pop_back();
+          // Since this code requires result.is_success() to be true, we should
+          // always have a path. We assert that..
+          DRAKE_ASSERT(path_vertex_ids.size() > 0);
           continue;
         }
         Eigen::VectorXd candidate_flows(candidate_edges.size());
--- a/geometry/optimization/test/graph_of_convex_sets_test.cc
+++ b/geometry/optimization/test/graph_of_convex_sets_test.cc
@@ -1496,6 +1496,115 @@ GTEST_TEST(ShortestPathTest, RoundingBacktrack) {
   ASSERT_TRUE(result.is_success());
 }
 
+// Cover the case where there is no path from source to target.
+GTEST_TEST(ShortestPathTest, NoPath) {
+  GraphOfConvexSets spp;
+  auto source = spp.AddVertex(Point(Vector2d(0, 0)));
+  auto v1 = spp.AddVertex(Point(Vector2d(0, 1)));
+  auto v2 = spp.AddVertex(Point(Vector2d(1, 0)));
+  auto target = spp.AddVertex(Point(Vector2d(1, 1)));
+  spp.AddEdge(*source, *v1);
+  spp.AddEdge(*v2, *target);
+
+  GraphOfConvexSetsOptions options;
+  options.convex_relaxation = true;
+  options.preprocessing = false;
+  options.max_rounded_paths = 10;
+  auto result = spp.SolveShortestPath(*source, *target, options);
+  EXPECT_FALSE(result.is_success());
+  EXPECT_EQ(result.get_solution_result(),
+            SolutionResult::kInfeasibleConstraints);
+
+  options.preprocessing = true;
+  options.max_rounded_paths = 10;
+  result = spp.SolveShortestPath(*source, *target, options);
+  EXPECT_FALSE(result.is_success());
+  EXPECT_EQ(result.get_solution_result(),
+            SolutionResult::kInfeasibleConstraints);
+
+  if (!MixedIntegerSolverAvailable()) {
+    return;
+  }
+
+  options.convex_relaxation = false;
+  result = spp.SolveShortestPath(*source, *target, options);
+  EXPECT_FALSE(result.is_success());
+  EXPECT_EQ(result.get_solution_result(),
+            SolutionResult::kInfeasibleConstraints);
+}
+
+// Cover the special case of the source not having any outgoing edges.
+GTEST_TEST(ShortestPathTest, NoPathDetachedSource) {
+  GraphOfConvexSets spp;
+  auto source = spp.AddVertex(Point(Vector2d(0, 0)));
+  spp.AddVertex(Point(Vector2d(0, 1)));
+  auto v2 = spp.AddVertex(Point(Vector2d(1, 0)));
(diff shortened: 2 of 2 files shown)


geometry/optimization/graph_of_convex_sets.cc
```diff
@@ -908,6 +930,7 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
   // Implements the rounding scheme put forth in Section 4.2 of
   // "Motion Planning around Obstacles with Convex Optimization":
   // https://arxiv.org/abs/2205.04422
+  bool has_paths_from_source_to_target = true;
```
TobiaMarcucci, line 933: I'm a bit confused by this: are we asking if the discrete graph has a path? Because if the convex relaxation of the SPP in GCS is feasible then, I believe, that the graph must have a path (since there is a unit of flow that somehow goes from s to t). In other words, does it actually happen in practice that the depth-first search backtracks to the source? I would imagine that that cannot happen if the convex relaxation is feasible, while if the convex relaxation is infeasible we should not do any rounding?

Or is this because we want a rounding function that does not make any assumptions on where the flows come from?

TobiaMarcucci (commented):
Just a question on why we need a flag to determine whether the graph has a path at the rounding stage.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719)* status: 1 unresolved discussion, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 933 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19719#gh-1252294429:-NZXPloHES3_9Wfq68N1:bz9eo3q) ([raw file](https://github.com/RobotLocomotion/drake/blob/de9dd98b70aa1129a2cd6adde5f44e9031e35d2e/geometry/optimization/graph_of_convex_sets.cc#L933)):*

I'm a bit confused by this: are we asking if the discrete graph has a path? Because if the convex relaxation of the SPP in GCS is feasible then, I believe, that the graph must have a path (since there is a unit of flow that somehow goes from s to t). In other words, does it actually happen in practice that the depth-first search backtracks to the source? I would imagine that that cannot happen if the convex relaxation is feasible, while if the convex relaxation is infeasible we should not do any rounding?

Or is this because we want a rounding function that does not make any assumptions on where the flows come from?
</blockquote></details>

Done.  You're absolutely right.

The initial bug was inside this loop (we did backtrack to zero), but that was actually a result of the bug which is now fixed above -- we got a success from the solver in a case where there was no path.

Now that the upstream bug is fixed, this case never happens. I've removed the checks.

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-:-NZXXug432LF5-4Er0Gg:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers

wrangelvid (commented):
Looks good to me. Only a minor nitpick. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-:-NZX_QkV9oeyw1oYr1tj:b-yhu707)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 973 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZX_8VA0eND7BS4gmFh:-NZX_8VA0eND7BS4gmFi:bq45o61) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/graph_of_convex_sets.cc#L973)):*
> ```cpp
>           path_vertex_ids.pop_back();
>           new_path.pop_back();
>           DRAKE_ASSERT(path_vertex_ids.size() > 0);
> ```

It doesn't look like this is getting caught in the tests.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719)* status: needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 973 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZX_8VA0eND7BS4gmFh:-NZXyBoI0_DwHtLTIeev:b-vp46ee) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/graph_of_convex_sets.cc#L973)):*

It doesn't look like this is getting caught in the tests.
</blockquote></details>

It should never fail.  I include it here only to give a more legible error message if we ever have another bug.  (the DRAKE_ASSERTs do not need explicit test coverage; in fact they are active only in debug mode).

SeanCurtis-TRI:

SeanCurtis-TRI (commented):
:LGTM: With some small points.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-:-NZY6H2p5PbL7RiDTEx1:b-g59ffz)* status: 3 unresolved discussions

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 973 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY4FaF0SW5qTU3-vEJ:-NZY4FaF0SW5qTU3-vEK:bvsu9tz) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/graph_of_convex_sets.cc#L973)):*
> ```cpp
>           path_vertex_ids.pop_back();
>           new_path.pop_back();
>           DRAKE_ASSERT(path_vertex_ids.size() > 0);
> ```

BTW The context in which this makes sense as a `DRAKE_ASSERT` instead of a `throw` is the fact that we've already validated the `source` node has outbound edges. Given that action takes place 180 lines above this assertion, it might be nice to put a note on why this validating implementation and not input data. 
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1499 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY5N04-MiaeFUjddsc:-NZY5N04-MiaeFUjddsd:brctnja) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/test/graph_of_convex_sets_test.cc#L1499)):*
> ```cpp
> }
> 
> GTEST_TEST(ShortestPathTest, NoPath) {
> ```

For these tests you've got three cases for not having a path:

1. Source not attached to the graph.
2. Target not attached to the graph.
3. Source and target lie in disjoint subgraphs.

Two thoughts:

1. `NoPath` and `NoPath2` don't do a good job of highlighting what is significant in the tests.
2. `NoPath` merges two cases in one. Such that correctness in one case may mask an error in the other case. As those two cases use independent machinery (their own bool, and their own tests), it would be good to test them independently.

I'd propose three tests:

1. `NoPathDetachedSource`
2. `NoPathDetachedTarget`
3. `NoPathDisjointGraphs`.
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1541 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY5xoF9HiS5QcVQGt7:-NZY5xoF9HiS5QcVQGt8:b-2dnjiz) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/test/graph_of_convex_sets_test.cc#L1541)):*
> ```cpp
>   auto target = spp.AddVertex(Point(Vector2d(1, 1)));
>   spp.AddEdge(*source, *v1);
>   spp.AddEdge(*v2, *target);
> ```

nit: As with the previous test, it would be worth highlighting that you've intentionally omitted edges connecting the two subgraphs. As the test name doesn't indicate that as being the feature under test, it should be mentioned here.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719)* status: 1 unresolved discussion

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 973 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY4FaF0SW5qTU3-vEJ:-NZYdUgZDUX4ASwHhkeh:btrwmrz) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/graph_of_convex_sets.cc#L973)):*

BTW The context in which this makes sense as a `DRAKE_ASSERT` instead of a `throw` is the fact that we've already validated the `source` node has outbound edges. Given that action takes place 180 lines above this assertion, it might be nice to put a note on why this validating implementation and not input data. 
</blockquote></details>

Done.  There is a simpler justification nearby.
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1499 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY5N04-MiaeFUjddsc:-NZYeUvC3i3VNZ699EBQ:bqpknx5) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/test/graph_of_convex_sets_test.cc#L1499)):*

For these tests you've got three cases for not having a path:

1. Source not attached to the graph.
2. Target not attached to the graph.
3. Source and target lie in disjoint subgraphs.

Two thoughts:

1. `NoPath` and `NoPath2` don't do a good job of highlighting what is significant in the tests.
2. `NoPath` merges two cases in one. Such that correctness in one case may mask an error in the other case. As those two cases use independent machinery (their own bool, and their own tests), it would be good to test them independently.

I'd propose three tests:

1. `NoPathDetachedSource`
2. `NoPathDetachedTarget`
3. `NoPathDisjointGraphs`.
</blockquote></details>

Done.  But if you call the rule of three on me, I'll be grumpy! 
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1541 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY5xoF9HiS5QcVQGt7:-NZYeiktC7F7H1RIIQVF:bing70y) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/test/graph_of_convex_sets_test.cc#L1541)):*

nit: As with the previous test, it would be worth highlighting that you've intentionally omitted edges connecting the two subgraphs. As the test name doesn't indicate that as being the feature under test, it should be mentioned here.
</blockquote></details>

Done. I've added comments above each test.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19719)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,SeanCurtis-TRI(platform)

___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1499 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19719#-NZY5N04-MiaeFUjddsc:-NZYmfqlBby3WEp51rMk:bwvox08) ([raw file](https://github.com/RobotLocomotion/drake/blob/60eb317ea8d6201a107feb510089a006b07a769c/geometry/optimization/test/graph_of_convex_sets_test.cc#L1499)):*

Done.  But if you call the rule of three on me, I'll be grumpy! 
</blockquote></details>

I'll simply appeal to your conscience. :) I'd probably press harder if this weren't test code. But I can tell it's gnawing at you already.  ;)

PR #19720, opened 2023-07-05
Optional default options for GraphOfConvexSets.

RussTedrake:
NOTE: Also changes GraphOfConvexSets to use `convex_relaxation = false` by default.

GcsTrajectoryOptimization (and perhaps other future downstream consumers of GCS) should use different default GCS options than the main GCS class. To support this, we make the relevant options std::optional, so that each method can distinguish between whether they have been set by the user or can be set to new appropriate default values before passing those options to the main GCS class.

For legibility, also adds an info-level log message telling user which of those most important options they've ended up with.

+@tobiamarcucci for feature review.
+@wrangelvid for additional feature review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)

### DIFF
--- a/geometry/optimization/graph_of_convex_sets.cc
+++ b/geometry/optimization/graph_of_convex_sets.cc
@@ -13,8 +13,8 @@
 #include <fmt/format.h>
 
 #include "drake/math/quadratic_form.h"
+#include "drake/solvers/choose_best_solver.h"
 #include "drake/solvers/create_cost.h"
-#include "drake/solvers/solve.h"
 
 namespace drake {
 namespace geometry {
@@ -62,7 +62,33 @@ MathematicalProgramResult Solve(const MathematicalProgram& prog,
   if (options.solver) {
     options.solver->Solve(prog, {}, solver_options, &result);
   } else {
-    result = solvers::Solve(prog, {}, solver_options);
+    std::unique_ptr<solvers::SolverInterface> solver{};
+    try {
+      solvers::SolverId solver_id = solvers::ChooseBestSolver(prog);
+      solver = solvers::MakeSolver(solver_id);
+    } catch (const std::exception&) {
+      // We should only get here if the user is trying to solve the MIP.
+      DRAKE_DEMAND(options.convex_relaxation == false && rounding == false);
+
+      // TODO(russt): Consider calling MixedIntegerBranchAndBound automatically
+      // here. The small trick is that we need to pass the SolverId into that
+      // constructor manually, and ChooseBestSolver doesn't make it easy to
+      // figure out what the best solver would be if we removed the integer
+      // variables.
+
+      throw std::runtime_error(
+          "GraphOfConvexSets: There is no solver available that can solve the "
+          "mixed-integer version of this problem. Please check "
+          "https://drake.mit.edu/doxygen_cxx/group__solvers.html for more "
+          "details about supported solvers and how to enable them.\n\n "
+          "Alternatively, you can try to solve the problem without integer "
+          "variables by setting options.convex_relaxation=true (and likely "
+          "setting options.max_rounded_paths > 0 if you want an "
+          "integer-feasible solution). See the documentation for "
+          "GraphOfConvexSetsOptions for more details.");
+    }
+    DRAKE_DEMAND(solver != nullptr);
+    solver->Solve(prog, {}, solver_options, &result);
   }
   return result;
 }
@@ -332,8 +358,16 @@ std::string GraphOfConvexSets::GetGraphvizString(
 std::set<EdgeId> GraphOfConvexSets::PreprocessShortestPath(
     VertexId source_id, VertexId target_id,
     const GraphOfConvexSetsOptions& options) const {
-  DRAKE_DEMAND(vertices_.find(source_id) != vertices_.end());
-  DRAKE_DEMAND(vertices_.find(target_id) != vertices_.end());
+  if (vertices_.find(source_id) == vertices_.end()) {
+    throw std::runtime_error(fmt::format(
+        "Source vertex {} is not a vertex in this GraphOfConvexSets.",
+        source_id));
+  }
+  if (vertices_.find(target_id) == vertices_.end()) {
+    throw std::runtime_error(fmt::format(
+        "Target vertex {} is not a vertex in this GraphOfConvexSets.",
+        target_id));
+  }
 
   std::map<VertexId, std::vector<int>> incoming_edges;
   std::map<VertexId, std::vector<int>> outgoing_edges;
@@ -649,12 +683,33 @@ void GraphOfConvexSets::AddPerspectiveConstraint(
 
 MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
     VertexId source_id, VertexId target_id,
-    const GraphOfConvexSetsOptions& options) const {
-  DRAKE_DEMAND(vertices_.find(source_id) != vertices_.end());
-  DRAKE_DEMAND(vertices_.find(target_id) != vertices_.end());
+    const GraphOfConvexSetsOptions& specified_options) const {
+  if (vertices_.find(source_id) == vertices_.end()) {
+    throw std::runtime_error(fmt::format(
+        "Source vertex {} is not a vertex in this GraphOfConvexSets.",
+        source_id));
+  }
+  if (vertices_.find(target_id) == vertices_.end()) {
+    throw std::runtime_error(fmt::format(
+        "Target vertex {} is not a vertex in this GraphOfConvexSets.",
+        target_id));
+  }
+
+  // Fill in default options. Note: if these options change, they must also be
+  // updated in the method documentation.
+  GraphOfConvexSetsOptions options = specified_options;
+  if (!options.convex_relaxation) {
+    options.convex_relaxation = false;
+  }
+  if (!options.preprocessing) {
+    options.preprocessing = false;
+  }
+  if (!options.max_rounded_paths) {
+    options.max_rounded_paths = 0;
+  }
 
   std::set<EdgeId> unusable_edges;
-  if (options.preprocessing) {
+  if (*options.preprocessing) {
     unusable_edges = PreprocessShortestPath(source_id, target_id, options);
   }
 
@@ -679,7 +734,7 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
       // Track excluded edges (ϕ = 0 and preprocessed) so that their variables
       // can be set in the optimization result.
       excluded_edges.emplace_back(e.get());
-      if (options.convex_relaxation) {
+      if (*options.convex_relaxation) {
         Variable phi("phi_excluded");
         excluded_phi.push_back(phi);
       }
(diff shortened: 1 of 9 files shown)


planning/trajectory_optimization/gcs_trajectory_optimization.cc
```diff
@@ -647,9 +647,21 @@ void GcsTrajectoryOptimization::AddVelocityBounds(
...
+GcsTrajectoryOptimization::SolvePath(
+    const Subgraph& source, const Subgraph& target,
+    const GraphOfConvexSetsOptions& specified_options) {
+  // Fill in default options.
+  GraphOfConvexSetsOptions options = specified_options;
+  if (!options.convex_relaxation) {
+    options.convex_relaxation = true;
+  }
+  if (!options.preprocessing) {
+    options.preprocessing = true;
```
TobiaMarcucci, line 659: Maybe `preprocessing` could be `false` by default? It can be expensive if the graph has many edges (should be an LP per edge). And it seems the kind of thing that the expert user wants to activate to have a performance boost for some particular problems?

geometry/optimization/graph_of_convex_sets.cc
```diff
@@ -62,7 +62,33 @@ MathematicalProgramResult Solve(const MathematicalProgram& prog,
...
+      // We should only get here if the user is trying to solve the MIP.
+      DRAKE_DEMAND(options.convex_relaxation == false && rounding == false);
+
+      // TODO(russt): Consider calling MixedIntegerBranchAndBound automatically
+      // here. The small trick is that we need to pass the SolverId into that
+      // constructor manually, and ChooseBestSolver doesn't make it easy to
+      // figure out what the best solver would be if we removed the integer
+      // variables.
+
+      throw std::runtime_error(
```
TobiaMarcucci, line 79: I like this!

TobiaMarcucci (commented):
I like the changes and I think the log is very useful to have!

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-:-NZb7ao--tBI-CdudoM8:bnfp4nl)* status: 2 unresolved discussions, LGTM missing from assignee wrangelvid, needs platform reviewer assigned

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: 2 unresolved discussions, LGTM missing from assignee wrangelvid, needs platform reviewer assigned

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#gh-1253429748:-NZbFhc2Bs5rZRw3ldKz:busvok4) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/geometry/optimization/graph_of_convex_sets.cc#L79)):*

I like this!
</blockquote></details>

Great!  (please mark yourself satisfied, so that reviewable stops thinking of this comment as blocking?).
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 659 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#gh-1253426694:-NZbFPGV0YcdYMve81PU:b-ucg164) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L659)):*

Maybe `preprocessing` could be `false` by default? It can be expensive if the graph has many edges (should be an LP per edge). And it seems the kind of thing that the expert user wants to activate to have a performance boost for some particular problems?
</blockquote></details>

In my recent experience, it always seemed to help.  My thinking was that in the trajopt formulation, we are loading more and more complexity into each set.  So the merits of spending a little time on the discrete graph is almost always helpful.

@wrangelvid -- wdyt?

TobiaMarcucci (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: 1 unresolved discussion, LGTM missing from assignee wrangelvid, needs platform reviewer assigned

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 659 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#gh-1253426694:-NZbHKyq-YWjWxn_zQLd:biznrf2) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L659)):*

In my recent experience, it always seemed to help.  My thinking was that in the trajopt formulation, we are loading more and more complexity into each set.  So the merits of spending a little time on the discrete graph is almost always helpful.

@wrangelvid -- wdyt?
</blockquote></details>

I see. For problems like the maze (maybe not the kind of thing we want to solve in practice) the preprocessing can take a very long time (~5k LPs) while the convex relaxation is an SOCP of the same size. However, I understand, and probably the new log should be enough for the user to understand what's going on if the solution of the convex relaxation does not start for a while. Eventually, I think it should not be impossible to automatically decide if to trigger the preprocessing depending on the sparsity of the graph or something like that?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: LGTM missing from assignee wrangelvid, needs platform reviewer assigned

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 659 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#gh-1253426694:-NZcIVVH7gEQvCBZE3p2:b-9qxjkh) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L659)):*

I see. For problems like the maze (maybe not the kind of thing we want to solve in practice) the preprocessing can take a very long time (~5k LPs) while the convex relaxation is an SOCP of the same size. However, I understand, and probably the new log should be enough for the user to understand what's going on if the solution of the convex relaxation does not start for a while. Eventually, I think it should not be impossible to automatically decide if to trigger the preprocessing depending on the sparsity of the graph or something like that?
</blockquote></details>

I see.  The maze example is a good counter-example.  I'm ok either way.

RussTedrake (commented):
-@wrangelvid 
+@sammy-tri for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-:-NZeZ5lzErBoFIJz2MSU:b1te8qj)* status: LGTM missing from assignee sammy-tri(platform)

wrangelvid (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: 1 unresolved discussion, LGTM missing from assignee sammy-tri(platform)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 659 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#gh-1253426694:-NZfStcr4MA4w6Ga4NmQ:b7rfado) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L659)):*

I see.  The maze example is a good counter-example.  I'm ok either way.
</blockquote></details>

The test cases are dealing with smaller graphs where the preprocessing is reducing the solve time by a lot. Even with the Kuka arms, we tend to have only a few regions so a majority of the users would end-up setting the preprocessing to `true`. Since we are targeting those applications initially, it makes sense to set the default value to `true`. 
___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 662 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-NZfU4-zCFI2gw0o_fBQ:-NZfU4-zCFI2gw0o_fBR:b42qyof) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L662)):*
> ```cpp
>   }
>   if (!options.max_rounded_paths) {
>     options.max_rounded_paths = 5;
> ```

Do you think it makes sense to add the default values to the documentation in the header file?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: LGTM missing from assignee sammy-tri(platform)

___
*[`planning/trajectory_optimization/gcs_trajectory_optimization.cc` line 662 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-NZfU4-zCFI2gw0o_fBQ:-NZfdAW7EOojSRjUqa6C:bc4sjck) ([raw file](https://github.com/RobotLocomotion/drake/blob/47b617b6d983742beb44ff1641968bf0f164bf91/planning/trajectory_optimization/gcs_trajectory_optimization.cc#L662)):*

Do you think it makes sense to add the default values to the documentation in the header file?
</blockquote></details>

Done.  Good call.

sammy-tri (commented):
:lgtm:

> Reviewed 5 of 8 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-:-NZgWE-a53GtANjzKopd:bnfp4nl)* status: 1 unresolved discussion

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 984 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-NZgUm3BCAQwxqpJ3-_i:-NZgUm3BCAQwxqpJ3-_j:bcbmhxj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9c61f5e9022fce3177435136b5075694e25993aa/geometry/optimization/graph_of_convex_sets.cc#L984)):*
> ```cpp
> 
>   MathematicalProgramResult result = Solve(prog, options, false);
>   log()->info(
> ```

nit: Should this (and the finished message) be "info" level logs (vs. debug or trace)?  I noticed in #19673 that this directory has been adding some text logging for long running tasks / debugging improvements, so if you want to leave it that's fine.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19720)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,sammy-tri(platform)

___
*[`geometry/optimization/graph_of_convex_sets.cc` line 984 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19720#-NZgUm3BCAQwxqpJ3-_i:-NZgtA8k7P9jBbx2RD1w:bsgyznt) ([raw file](https://github.com/RobotLocomotion/drake/blob/9c61f5e9022fce3177435136b5075694e25993aa/geometry/optimization/graph_of_convex_sets.cc#L984)):*

nit: Should this (and the finished message) be "info" level logs (vs. debug or trace)?  I noticed in #19673 that this directory has been adding some text logging for long running tasks / debugging improvements, so if you want to leave it that's fine.
</blockquote></details>

info feels like the right level to me.

PR #11344, opened 2019-04-26
mbp: Ensure that all API entry points for adding joints registers topology

EricCousineau-TRI:
Hotfix for #11257

@amcastro-tri I have to jet; if you can fix the compilation error and push it here, that'd be totes awesome. Otherwise, I can fix it later.

Closes #11343.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake/11344)

PR #19724, opened 2023-07-05
[multibody] Reject duplicated frame names

jwnimmer-tri:
Unwind some code duplication between the plant and the tree, so that we only need to fix the logic one place. (Amends #11344 to remove its copy-pasta.)

To avoid name conflicts for a joint's parent body's offset frame, register the offset frame using the child body's model instance instead of the parent's model instance.

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724)

### DIFF
--- a/multibody/plant/multibody_plant.h
+++ b/multibody/plant/multibody_plant.h
@@ -965,13 +965,15 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
 
   /// This method adds a Joint of type `JointType` between two bodies.
   /// For more information, see the below overload of `AddJoint<>`.
-  template <template<typename Scalar> class JointType>
+  template <template <typename Scalar> class JointType>
   const JointType<T>& AddJoint(std::unique_ptr<JointType<T>> joint) {
-    DRAKE_MBP_THROW_IF_FINALIZED();
     static_assert(std::is_convertible_v<JointType<T>*, Joint<T>*>,
                   "JointType must be a sub-class of Joint<T>.");
-    RegisterJointInGraph(*joint);
-    return this->mutable_tree().AddJoint(std::move(joint));
+    DRAKE_MBP_THROW_IF_FINALIZED();
+    const JointType<T>& result =
+        this->mutable_tree().AddJoint(std::move(joint));
+    RegisterJointInGraph(result);
+    return result;
   }
 
   /// This method adds a Joint of type `JointType` between two bodies.
@@ -1059,31 +1061,12 @@ class MultibodyPlant : public internal::MultibodyTreeSystem<T> {
     //  consider changing frame F to frame Jp and changing frame M to frame Jc.
     static_assert(std::is_base_of_v<Joint<T>, JointType<T>>,
                   "JointType<T> must be a sub-class of Joint<T>.");
-
-    const Frame<T>* frame_on_parent{nullptr};
-    if (X_PF) {
-      frame_on_parent = &this->AddFrame(
-          std::make_unique<FixedOffsetFrame<T>>(
-              name + "_parent", parent, *X_PF));
-    } else {
-      frame_on_parent = &parent.body_frame();
-    }
-
-    const Frame<T>* frame_on_child{nullptr};
-    if (X_BM) {
-      frame_on_child = &this->AddFrame(
-          std::make_unique<FixedOffsetFrame<T>>(
-              name + "_child", child, *X_BM));
-    } else {
-      frame_on_child = &child.body_frame();
-    }
-
-    const JointType<T>& joint = AddJoint(
-        std::make_unique<JointType<T>>(
-            name,
-            *frame_on_parent, *frame_on_child,
-            std::forward<Args>(args)...));
-    return joint;
+    DRAKE_MBP_THROW_IF_FINALIZED();
+    const JointType<T>& result =
+        this->mutable_tree().template AddJoint<JointType>(
+            name, parent, X_PF, child, X_BM, std::forward<Args>(args)...);
+    RegisterJointInGraph(result);
+    return result;
   }
 
   /// Welds `frame_on_parent_F` and `frame_on_child_M` with relative pose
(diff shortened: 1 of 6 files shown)


jwnimmer-tri:
+@sherm1 for feature review or delegation, please.

sherm1 (commented):
Feature pass done. CI needs repair. One design discussion, PTAL.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-:-NZgNjxoDIDJB7J03zFr:baupk3)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_test.cc` line 396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-NZgHjCd2DWXBulDJ9DX:-NZgHjCd2DWXBulDJ9DY:b-uhh9gx) ([raw file](https://github.com/RobotLocomotion/drake/blob/483d4494e32772c2f8ee3e0bd6cb1d82c2056f12/multibody/plant/test/multibody_plant_test.cc#L396)):*
> ```cpp
>   EXPECT_EQ(pendulum_frame_indices[3], pin_joint.frame_on_child().index());
>   EXPECT_EQ(pendulum_frame_indices[4], weld_joint.frame_on_child().index());
>   EXPECT_EQ(pendulum_frame_indices[5], weld_joint.frame_on_parent().index());
> ```

BTW swapping these apparently caused some test failures in CI
___
*[`multibody/tree/multibody_tree-inl.h` line 333 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-NZgKWaq5aB0TFpDlC9A:-NZgKWaq5aB0TFpDlC9B:bhamjyr) ([raw file](https://github.com/RobotLocomotion/drake/blob/483d4494e32772c2f8ee3e0bd6cb1d82c2056f12/multibody/tree/multibody_tree-inl.h#L333)):*
> ```objc
>   static_assert(std::is_base_of_v<Joint<T>, JointType<T>>,
>                 "JointType<T> must be a sub-class of Joint<T>.");
>   const ModelInstanceIndex joint_instance = child.model_instance();
> ```

BTW consider calling this `frame_instance` instead. Currently it is true that the joint model instance will ultimately be inherited from the child frame's model instance. But that policy is independent of the policy here for assigning model instances to frames. 

A nicer alternative IMO would be to have the Joint model instance passed in to tree.AddJoint (as we do for AddBody), and then have any created frames and the joint use that. Then the joint model instance policy could be pushed up to MbP where it belongs so it would be known by the time we're creating frames and joints.
___
*[`multibody/tree/test/multibody_tree_test.cc` line 250 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-NZgLvaZ-agiN908be0m:-NZgLvaZ-agiN908be0n:b-lnue5t) ([raw file](https://github.com/RobotLocomotion/drake/blob/483d4494e32772c2f8ee3e0bd6cb1d82c2056f12/multibody/tree/test/multibody_tree_test.cc#L250)):*
> ```cpp
>                                         model->GetBodyByName("iiwa_link_1"),
>                                         RigidTransform<double>()),
>       /* Verify this method is throwing for the right reasons. */
> ```

BTW this comment seems superfluous

jwnimmer-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-:-NZgWHfs7YJ9KlH0iLGW:b-1krawd)* status: 1 unresolved discussion, LGTM missing from assignees sherm1(platform),sammy-tri(platform)

___
*[`multibody/tree/multibody_tree-inl.h` line 333 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-NZgKWaq5aB0TFpDlC9A:-NZgTZ-Z6o2tMvy49ws-:b5ak3gn) ([raw file](https://github.com/RobotLocomotion/drake/blob/483d4494e32772c2f8ee3e0bd6cb1d82c2056f12/multibody/tree/multibody_tree-inl.h#L333)):*
I've improved the implementation to cross-check this.

> A nicer alternative IMO would be to have the Joint model instance passed in to tree.AddJoint (as we do for AddBody), and then have any created frames and the joint use that.

That would withdraw the promise from the Joint documentation that its model instance is _always_ the same as its child Body.  I don't feel comfortable undoing that invariant, and certainly not as part of this PR.
___
*[`multibody/plant/test/multibody_plant_test.cc` line 396 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-NZgHjCd2DWXBulDJ9DX:-NZgVa3Q4rYjEOrZlVIM:b-94h0zg) ([raw file](https://github.com/RobotLocomotion/drake/blob/483d4494e32772c2f8ee3e0bd6cb1d82c2056f12/multibody/plant/test/multibody_plant_test.cc#L396)):*

BTW swapping these apparently caused some test failures in CI
</blockquote></details>

The perils of calling mutation unctions ("add frame") from inside an argument list.  The order of evaluation is unspecified.

sammy-tri (commented):
:lgtm:

> Reviewed 4 of 7 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-:-NZga7VdCQ4EEvW6w-Bw:bnfp4nl)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724)* status: LGTM missing from assignee sherm1(platform)

jwnimmer-tri:
@sherm1 BTW we still need your LGTM emoji, once you're finished.

sherm1 (commented):
Sorry, forgot the stamp! :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19724#-:-NZh9BjL4iQ0HP5Iii8A:bfn3bv2)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),sherm1(platform)

PR #19731, opened 2023-07-06
[pydrake/multibody] Add fix_inertia tool

rpoyner-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)

### DIFF
--- a/bindings/pydrake/multibody/BUILD.bazel
+++ b/bindings/pydrake/multibody/BUILD.bazel
@@ -264,6 +264,32 @@ drake_py_binary(
     ],
 )
 
+drake_py_library(
+    name = "inertia_fixer_py",
+    srcs = [
+        "_inertia_fixer.py",
+        "fix_inertia.py",
+    ],
+    imports = PACKAGE_INFO.py_imports,
+    deps = [
+        ":parsing_py",
+        ":plant_py",
+        "//bindings/pydrake/common",
+        "//bindings/pydrake/geometry",
+    ],
+)
+
+drake_py_binary(
+    name = "fix_inertia",
+    srcs = [
+        "fix_inertia.py",
+    ],
+    visibility = ["//tools:__pkg__"],
+    deps = [
+        ":inertia_fixer_py",
+    ],
+)
+
 drake_jupyter_py_binary(
     name = "examples/door_hinge_inspector",
     add_test_rule = 1,
@@ -282,14 +308,25 @@ cmake_configure_file(
     ],
 )
 
+cmake_configure_file(
+    name = "generate_run_installed_fix_inertia",
+    src = "run_installed_fix_inertia.py.in",
+    out = "run_installed_fix_inertia.py",
+    defines = [
+        "PYTHON_SITE_PACKAGES_RELPATH=" + PYTHON_SITE_PACKAGES_RELPATH,
+    ],
+)
+
 install_files(
     name = "install_wrapper_scripts",
     dest = "bin",
     files = [
+        "run_installed_fix_inertia.py",
         "run_installed_mesh_to_model.py",
     ],
     rename = {
         "bin/run_installed_mesh_to_model.py": "mesh_to_model",
+        "bin/run_installed_fix_inertia.py": "fix_inertia",
     },
 )
 
@@ -310,6 +347,7 @@ PY_LIBRARIES = [
     ":jupyter_widgets_py",
     ":module_py",
     ":mesh_model_maker_py",
+    ":inertia_fixer_py",
 ]
 
 # Symbol roll-up (for user ease).
@@ -338,6 +376,7 @@ filegroup(
     name = "models",
     srcs = glob([
         "**/*.sdf",
+        "**/*.urdf",
     ]),
     visibility = ["//:__pkg__"],
 )
@@ -507,8 +546,31 @@ drake_py_unittest(
     ],
 )
 
+drake_py_unittest(
+    name = "fix_inertia_test",
+    data = [
+        ":fix_inertia",
+        ":models",
+        "//examples/acrobot:models",
+        "//examples/multibody/four_bar:models",
+        "//manipulation/util:test_models",
+        "//multibody:models",
+        "//multibody/benchmarks/acrobot:models",
+        "//multibody/parsing:test_models",
+    ],
+    deps = [
+        ":inertia_fixer_py",
+        "//bindings/pydrake/common",
+        "//bindings/pydrake/common/test_utilities:numpy_compare_py",
+        "//bindings/pydrake/geometry",
+        "//bindings/pydrake/multibody:parsing_py",
+        "//bindings/pydrake/multibody:plant_py",
+    ],
+)
+
 add_lint_tests_pydrake(
     python_lint_extra_srcs = [
         "run_installed_mesh_to_model.py.in",
+        "run_installed_fix_inertia.py.in",
     ],
 )
--- a/bindings/pydrake/multibody/_inertia_fixer.py
+++ b/bindings/pydrake/multibody/_inertia_fixer.py
@@ -1,0 +1,703 @@
+"""Fix broken (non-physical) inertias in SDFormat/URDF, by writing a revised
+file.
+
+The common data structure being repaired in both formats is the content of the
+`<inertial>` tag, which contains the mass, the inertia tensor, and also defines
+a central inertia frame in which the tensor is expressed.
+
+By definitions used in both URDF and SDFormat documentation, the relevant
+frames are the link origin frame (L) and the central inertia frame (C). Drake
+software often uses the terminology "body frame" (B), which is equivalent to L.
+Therefore, in Drake monogram notation, the inertia tensor is expressed in B,
+and the `<inertial>` tag includes the pose of C relative to B, X_BC.
+
+"""
+from __future__ import annotations
+
+import abc
+import argparse
+from dataclasses import dataclass, field
+from typing import Callable
+import xml.parsers.expat as expat
+
+import numpy as np
+
+from pydrake.geometry import (Role, SceneGraph)
+from pydrake.multibody.plant import MultibodyPlant
+from pydrake.multibody.parsing import Parser
+from pydrake.multibody.tree import (BodyIndex, CalcSpatialInertia, RigidBody,
+                                    SpatialInertia, UnitInertia)
+
+
+@dataclass
+class SourceLocation:
+    """The location of an XML syntax item as reported by expat parsing."""
+
+    index: int
+    "0-based byte index from start of input stream."
+
+    line: int
+    "1-based line number from start of input stream."
+
+
+@dataclass
+class ElementFacts:
+    """Facts about an XML element as collecting by expat parsing."""
+
+    prior_data: str
+    "Character data immediately preceding the element text."
+
+    name: str
+    "The name of the XML element."
+
+    attributes: dict
+    "The attributes of the XML element."
+
+    depth: int
+    "The nesting depth of the XML element; 0 for the root element."
+
+    start: SourceLocation
+    "The location of the start of the XML element text."
+
+    end: SourceLocation = None
+    """The location of the end of the XML element text; may require
+    adjustments, depending on the input syntax used (optional).
+    """
+
+    parent: ElementFacts = None
+    "Facts for the parent of this element (optional)."
+
+    children: list[ElementFacts] = field(default_factory=list)
+    "Facts about relevant children of this element (optional)."
+
+    parent_model: ElementFacts = None
+    """Facts for the enclosing model of this model element (optional).
+    The definition of a model element is determined by each file format;
+    see FormatDriver.is_model_element().
+    """
+
+
+# This function has an aggressively short name for embedding into f-strings.
+def gfmt(x: float) -> str:
+    """Converts a floating point number to a string, using some rounding to
+    avoid platform differences in results.
+    """
+    return f"{x:.5g}"
+
+
+class FormatDriver(object, metaclass=abc.ABCMeta):
+    """This interface encapsulates the differences between URDF and SDFormat
+    processing, for purposes of the implementation of XmlInertiaMapper.
+    """
+
+    @abc.abstractmethod
+    def is_model_element(self, name: str) -> bool:
+        """Returns True if the element directly encloses link elements.
+        Args:
+            name: the name of the XML element
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def associate_plant_models(self, models: list, links: list,
+                               plant: MultibodyPlant) -> dict:
+        """Returns a mapping of body_index to inertial ElementFacts.
+
+        Args:
+            models: a list of ElementFacts for model elements
+            links: a list of ElementFacts for link elements
+            plant: a MultibodyPlant containing the parsed models
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def format_inertia(self, input_bytes: bytes, inertial_facts: ElementFacts,
+                       pose: list, mass: float, moments: list,
+                       products: list) -> bytes:
+        """Returns formatted XML bytes for the repaired inertial tag.
+
+        See above for definitions of the frames involved.
+
+        Args:
+            input_bytes: full contents of the input file
+            inertial_facts: facts about the element to reformat
+            pose: Pose of the central inertia frame (C) relative to the body
+                  frame (B), as a 6-element list: [x, y, z, r, p, y].
+            mass: the mass of the body (kg).
+            moments: the principle moments of inertia, expressed in C --
+                     [ixx, iyy, izz].
+            products: the products of inertia, expressed in C --
+                      [ixy, ixz, iyz].
+        """
+        raise NotImplementedError
+
+
+def is_synth_element(el: ElementFacts) -> bool:
+    """Returns True if the ElementFacts is synthetic.
+
+    Synthetic elements are placeholders for later expansion into output XML.
+    """
+    return el.end.index == el.start.index
+
+
+def make_synth_element(parent: ElementFacts, name: str) -> ElementFacts:
+    """Makes a new synthetic child ElementFacts for a given parent
+    ElementFacts. The parent must already have at least one child.
+
+    The returned synthetic element will be used for later expansion into the
+    output XML.
+
+    Args:
+        parent: the ElementFacts that will get a new synthetic child
+        name: the XML element name of the new child
+
+    """
+    assert parent.children, parent
+    kid0 = parent.children[0]
+    synth_el = ElementFacts(kid0.prior_data, name, {},
+                            kid0.depth, kid0.start)
+    synth_el.end = synth_el.start
+    return synth_el
+
+
+def find_inertial_facts_for_link(link: ElementFacts) -> ElementFacts:
+    """Fetches or synthesizes a child inertial ElementFacts for a link
+    ElementFacts.
+    """
+    inertial_kids = [x for x in link.children if x.name == "inertial"]
+    if inertial_kids:
+        return inertial_kids[0]
+    return make_synth_element(link, "inertial")
+
+
+def maybe_synthesize_children(parent: ElementFacts, names: list[str]):
+    """Given a parent ElementFacts and a list of required child elements,
+    synthesizes ElementFacts for any children that are missing. The parent must
+    already have at least one child.
+
+    The new synthetic elements will be used for later expansion into the output
+    XML. Note that synthetic elements are inserted at the beginning of the list
+    of child elements (for formatting convenience), so they will appear in the
+    formatted output in the reverse order of that given by the input list of
+    names.
+
+    Args:
+        parent: the ElementFacts that will get new synthetic children
+        names: a list of XML element names to insert if missing
+    """
+    assert parent.children, parent
+    kid_names = [x.name for x in parent.children]
+    for needed_name in names:
+        if needed_name in kid_names:
+            continue
+        parent.children.insert(
+            0, make_synth_element(parent, needed_name))
+
+
+def adjusted_element_end_index(input_bytes: bytes, facts: ElementFacts) -> int:
+    """Returns the index of the end of all of the text (including the element
+    closure), when given an ElementFacts produced by the expat parse.
+
+    Args:
+        input_bytes: full contents of the input file
+        facts: an ElementFacts structure
+    """
+    if is_synth_element(facts):
+        # Empty, synthetic, "pseudo" element.
+        return facts.end.index
+
+    if input_bytes[facts.end.index:].startswith(
+            f"</{facts.name}".encode('utf-8')):
+        # Typical case:
+        # `<inertial>...</inertial>`.
+        #               ^                # Raw index points here.
+        #                          ^     # Adjusted index points here.
+        return facts.end.index + input_bytes[facts.end.index:].find(b'>') + 1
+
+    # `<inertial/>` corner case.
+    #             ^     # Raw index points here; good to go
+    return facts.end.index
+
+
+def make_format_helpers(facts: ElementFacts) -> (
+        Callable[[int], str], Callable[[], str]):
+    """Returns two callable functions that help format white space for output
+    XML elements.
+
+    The returned callables are intended to be convenient to invoke in Python
+    f-strings; callers typically assign them as `d` and `coda`.
+
+    d(int) -> str
+        indents n levels in addition to the current depth of the element
+        described by `facts`.
+
+    coda() -> str
+        provides appropriate trailing white space to maintain typical
+        formatting:
+            * newline plus d(0) if `facts` is synthetic,
+            * otherwise, empty string.
+
+    Args:
+        facts: an ElementFacts structure
+
+    Returns:
+        d: the indent function described above
+        coda: the trailing space function described above
+    """
+    spaces = len(facts.prior_data)
+    indentation = spaces // facts.depth
+
+    # Build an element-specific indenter.
+    depth = facts.depth
+    indent = " " * indentation
+
+    def d(more_depth):
+        return indent * (depth + more_depth)
+
+    synth = facts.end.index == facts.start.index
+
+    def coda():
+        return '\n' + d(0) if synth else ""
+
+    return (d, coda)
+
+
+class UrdfDriver(FormatDriver):
+    """Format driver for URDF files."""
+
+    def is_model_element(self, name: str) -> bool:
+        return name == "robot"
+
+    def _is_element_ignored(self, facts: ElementFacts) -> bool:
+        # TODO(rpoyner_tri): The 'drake_ignore' attribute is regrettable legacy
+        # cruft that should be removed when the associated URDF parser cruft is
+        # removed.
+        return ((facts.attributes.get("drake_ignore") == "true")
+                or (facts.attributes.get("name") == "world"))
+
+    def associate_plant_models(self, models: list, links: list,
+                               plant: MultibodyPlant) -> dict:
+        # This assertion is weakened for files that use 'drake_ignore'.
+        assert len(links) >= plant.num_bodies() - 1, (
+            links, plant.num_bodies())
+        mapping = {}
+        serial_number = 0
+        for link in links:
+            if self._is_element_ignored(link):
+                continue
+            serial_number += 1
+            if not link.children:
+                # Without inertia (mass) or geometry, there is nothing to do.
+                continue
+            link_name = link.attributes.get("name")
+            assert serial_number < plant.num_bodies()
+            bix = BodyIndex(serial_number)
+            mapping[bix] = find_inertial_facts_for_link(link)
+            assert plant.get_body(bix).name() == link.attributes.get("name"), (
+                plant.get_body(bix).name(), bix, link.attributes.get("name"))
+        return mapping
+
+    def format_inertia(self, input_bytes: bytes, inertial_facts: ElementFacts,
+                       pose: list, mass: float, moments: list,
+                       products: list) -> bytes:
+        end = adjusted_element_end_index(input_bytes, inertial_facts)
+        inertia_output = f"""\
+<inertia ixx="{gfmt(moments[0])}" ixy="{gfmt(products[0])}"\
+ ixz="{gfmt(products[1])}" iyy="{gfmt(moments[1])}"\
+ iyz="{gfmt(products[2])}" izz="{gfmt(moments[2])}"/>"""
+        xyz_output = ' '.join([gfmt(x) for x in pose[:3]])
+        rpy_output = ' '.join([gfmt(x) for x in pose[3:]])
+        # * self-closing or no inertial tag: pave it all.
+        if end == inertial_facts.end.index:
+            d, coda = make_format_helpers(inertial_facts)
+            return f"""\
+<inertial>
+{d(1)}{inertia_output}
+{d(1)}<mass value="{gfmt(mass)}"/>
+{d(1)}<origin rpy="{rpy_output}" xyz="{xyz_output}"/>
+{d(0)}</inertial>{coda()}""".encode('utf-8')
+
+        maybe_synthesize_children(
+            inertial_facts, ["origin", "mass", "inertia"])
+        # Now we get to slice up the inertial element, and only rewrite the
+        # parts that we must.
+        output = bytes()
+        index = inertial_facts.start.index
+        for facts in inertial_facts.children:
+            output += input_bytes[index:facts.start.index]
+            d, coda = make_format_helpers(facts)
+            if facts.name == "inertia":
+                output += f"{inertia_output}{coda()}".encode('utf-8')
+            if facts.name == "mass":
+                output += f'<mass value="{gfmt(mass)}"/>{coda()}'.encode(
+                    'utf-8')
+            if facts.name == "origin":
+                output += f"""\
+<origin rpy="{rpy_output}" xyz="{xyz_output}"/>{coda()}""".encode('utf-8')
+            index = adjusted_element_end_index(input_bytes, facts)
+        output += input_bytes[index:end]
+        return output
+
+
+class SdformatDriver(FormatDriver):
+    """Format driver for SDFormat files."""
+
+    def is_model_element(self, name: str) -> bool:
+        return name == "model"
+
+    def associate_plant_models(self, models: list, links: list,
+                               plant: MultibodyPlant) -> dict:
+        # Because SDFormat has both nested models and inclusions, we will have
+        # to rummage around in the plant, finding body indices by using name
+        # strings.
+        mapping = {}
+        for link in links:
+            if not link.children:
+                continue
+            model = link.parent
+            enclosing_models = [model]
+            while model.parent_model:
+                model = model.parent_model
+                enclosing_models.append(model)
+            enclosing_models.reverse()
+            model_name = '::'.join(
+                [m.attributes["name"] for m in enclosing_models])
+            model_instance = plant.GetModelInstanceByName(model_name)
+            body = plant.GetBodyByName(link.attributes["name"], model_instance)
+            bix = body.index()
+            mapping[bix] = find_inertial_facts_for_link(link)
+            assert plant.get_body(bix).name() == link.attributes.get("name"), (
+                plant.get_body(bix).name(), bix, link.attributes.get("name"))
+        return mapping
+
+    def format_inertia(self, input_bytes: bytes, inertial_facts: ElementFacts,
+                       pose: list, mass: float, moments: list,
+                       products: list) -> bytes:
+        end = adjusted_element_end_index(input_bytes, inertial_facts)
+        pose_output = ' '.join([gfmt(x) for x in pose])
+        # * self-closing or no inertial tag: pave it all.
+        if end == inertial_facts.end.index:
+            d, coda = make_format_helpers(inertial_facts)
+            return f"""\
+<inertial>
+{d(1)}<inertia>
+{d(2)}<ixx>{gfmt(moments[0])}</ixx>
+{d(2)}<ixy>{gfmt(products[0])}</ixy>
+{d(2)}<ixz>{gfmt(products[1])}</ixz>
+{d(2)}<iyy>{gfmt(moments[1])}</iyy>
+{d(2)}<iyz>{gfmt(products[2])}</iyz>
+{d(2)}<izz>{gfmt(moments[2])}</izz>
+{d(1)}</inertia>
+{d(1)}<mass>{gfmt(mass)}</mass>
+{d(1)}<pose>{pose_output}</pose>
+{d(0)}</inertial>{coda()}""".encode('utf-8')
+
+        maybe_synthesize_children(
+            inertial_facts, ["pose", "mass", "inertia"])
+        # Now we get to slice up the inertial element, and only rewrite the
+        # parts that we must.
+        output = bytes()
+        index = inertial_facts.start.index
+        for facts in inertial_facts.children:
+            output += input_bytes[index:facts.start.index]
+            d, coda = make_format_helpers(facts)
+            if facts.name == "inertia":
+                output += f"""\
+<inertia>
+{d(1)}<ixx>{gfmt(moments[0])}</ixx>
+{d(1)}<ixy>{gfmt(products[0])}</ixy>
+{d(1)}<ixz>{gfmt(products[1])}</ixz>
+{d(1)}<iyy>{gfmt(moments[1])}</iyy>
+{d(1)}<iyz>{gfmt(products[2])}</iyz>
+{d(1)}<izz>{gfmt(moments[2])}</izz>
+{d(0)}</inertia>{coda()}""".encode('utf-8')
+            if facts.name == "mass":
+                output += f"<mass>{gfmt(mass)}</mass>{coda()}".encode('utf-8')
+            if facts.name == "pose":
+                output += f"<pose>{pose_output}</pose>{coda()}".encode('utf-8')
+            index = adjusted_element_end_index(input_bytes, facts)
+        end = adjusted_element_end_index(input_bytes, inertial_facts)
+        output += input_bytes[index:end]
+        return output
+
+
+class XmlInertiaMapper:
+    """Handles the parsing, indexing, and output generation details for editing
+    of inertias in some formats of XML robot model files.
+    """
+
+    def __init__(self, input_text: str):
+        """Initialize an XmlInertiaMapper.
+
+        Args:
+            input_text: the full contents of the input file
+        """
+        self._input_text = input_text
+        self._input_bytes = input_text.encode('utf-8')
+
+        # Configure the parser.
+        self._parser = expat.ParserCreate()
+        self._parser.StartElementHandler = self._handle_start_element
+        self._parser.EndElementHandler = self._handle_end_element
+        self._parser.CharacterDataHandler = self._handle_char_data
+
+        # Declare some state to remember during parsing.
+        self._depth = 0
+        self._element_stack = []
+        self._models = []
+        self._model_stack = []
+        self._links = []
+        self._format_driver = None
+        self._last_char_data = ""
+
+        # Eventually build a mapping from body_index to inertial_facts.
+        self._mapping = {}
+
+    def _make_location(self) -> SourceLocation:
+        return SourceLocation(self._parser.CurrentByteIndex,
+                              self._parser.CurrentLineNumber)
+
+    def _make_element_facts(self, name: str, attributes: dict) -> ElementFacts:
+        return ElementFacts(self._last_char_data, name, attributes,
+                            self._depth, self._make_location())
+
+    def _handle_char_data(self, data: str):
+        self._last_char_data = data
+
+    def _handle_start_element(self, name: str, attributes: dict):
+        if not self._format_driver and self._depth == 0:
+            if name == "robot":
+                self._format_driver = UrdfDriver()
+            elif name == "sdf":
+                self._format_driver = SdformatDriver()
+            else:
+                raise RuntimeError("unknown file format!")
+
+        element = self._make_element_facts(name, attributes)
+        if self._element_stack:
+            element.parent = self._element_stack[-1]
+            element.parent.children.append(element)
+
+        if self._format_driver.is_model_element(name):
+            if self._model_stack:
+                element.parent_model = self._model_stack[-1]
+            self._models.append(element)
+            self._model_stack.append(element)
+
+        if name == "link":
+            self._links.append(element)
+
+        self._depth += 1
+        self._element_stack.append(element)
+
+    def _handle_end_element(self, name: str):
+        self._element_stack[-1].end = self._make_location()
+        self._element_stack.pop()
+        self._depth -= 1
+
+        if self._format_driver.is_model_element(name):
+            self._model_stack.pop()
+
+    def parse(self):
+        """Execute the parsing of the XML text."""
+        self._parser.Parse(self._input_text)
+
+    def build_plant_models_association(self, plant: MultibodyPlant):
+        """Matches body indices to inertial elements in the input text.
+
+        Args:
+            plant: a MultibodyPlant containing the parsed models
+        """
+        assert self._format_driver is not None
+        self._mapping = self._format_driver.associate_plant_models(
+            self._models, self._links, plant)
+
+    def mapping(self) -> dict:
+        """Returns a mapping from body indices to inertial element facts."""
+        return self._mapping
+
+    def build_output(self, new_inertias_mapping: dict) -> str:
+        """Builds and returns the complete output text.
+
+        Args:
+            new_inertias_mapping: a dictionary whose keys are BodyIndex values,
+               and whose values are a tuple of the body's spatial inertia and
+               6D inertial pose as [x, y, z, r, p, y].
+        """
+        output = bytes()
+        input_bytes = self._input_bytes
+        input_bytes_index = 0
+        for body_index, new_inertia in sorted(new_inertias_mapping.items()):
+            inertial_facts = self._mapping[body_index]
+            output += input_bytes[input_bytes_index:inertial_facts.start.index]
+            output += self._format_inertia(inertial_facts, *new_inertia)
+            input_bytes_index = adjusted_element_end_index(
+                input_bytes, inertial_facts)
+        output += input_bytes[input_bytes_index:]
+        return output.decode('utf-8')
+
+    def _format_inertia(self, inertial_facts: ElementFacts,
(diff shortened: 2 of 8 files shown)


rpoyner-tri (commented):
+@trowell-tri for eventual feature review -- not yet.
+(status: do not merge) +(status: do not review) 

This is a reorg of code started as #19463, to accommodate testing, and to be compatible with distribution formats.

Motivating issue is #18773 .

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NZgiFJW1U2idDA1EUZe:bntwkvp)* status: LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
I believe `r2` is substantially correct and has decent test coverage. Documentation is still sorely lacking and/or stale.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-N_uR0do2ecYvg9OBpEQ:b5uoxka)* status: LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 1 unresolved discussion, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 194 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-N_uS5sB63TxEEV9fGe7:-N_uS5sB63TxEEV9fGe8:b-ehazby) ([raw file](https://github.com/RobotLocomotion/drake/blob/c0123643b61e7569363ab2bee2b3f972d7b412b7/multibody/parsing/parser.h#L194)):*
> ```objc
>   bool GetAutoRenaming() const { return enable_auto_rename_; }
> 
>   /// Cause all subsequent Add*Model*() operations to replace physically
> ```

These entry points should be marked "advanced" or "internal use only" or some such.

rpoyner-tri (commented):
As always, writing bold assertions about correctness is the best way to find bugs. `r3` fixes at least one bug, adds some but not all docs, and adds some previously missing tests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NaEE2Px6sVvNJwwYTx2:b-uywwpf)* status: 1 unresolved discussion, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 1 unresolved discussion, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 194 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-N_uS5sB63TxEEV9fGe7:-NaMs4sO0D93Z-Veszg9:b-pyas8g) ([raw file](https://github.com/RobotLocomotion/drake/blob/c0123643b61e7569363ab2bee2b3f972d7b412b7/multibody/parsing/parser.h#L194)):*

These entry points should be marked "advanced" or "internal use only" or some such.
</blockquote></details>

I did some skimming of the PR, but it's not yet clear to me why this feature is related to the tool.  I assume there is some reason why the rewriting tool wants to enable this, but I couldn't tease out what it is?

trowell-tri (commented):
> Reviewed 2 of 5 files at r1, 12 of 14 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 19 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_common.h` line 288 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaIth4QAu1FQhMe3erM:-NaIth4QAu1FQhMe3erN:b-si4ddy) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.h#L288)):*
> ```objc
> // @param spoil_invalid_inertia  If true, return NaN-filled inertia rather than
> //                                some valid approximation when the input values
> //                                are physically invalid.
> ```

btw elsewhere you note that the returned SpatialInertia is default-constructed and I wonder if the most clear phrasing is something like "default-constructed (NaN-filled) SpatialInertia". Or are the differing levels of technical detail provided on purpose?
___
*[`multibody/parsing/detail_common.cc` line 270 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNAL6M7YE27sdRKvqV:-NaNAL6M7YE27sdRKvqW:b-u2lsls) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.cc#L270)):*
> ```cpp
>     bool spoil_invalid_inertia) {
>   SpatialInertia<double> invalid_inertia_replacement;
>   const Vector3d& p_BoBcm_B = X_BBi.translation();
> ```

btw this got yanked way up here but no code other than the original stanza seems to use it? Detritus from a prior refactoring?
___
*[`multibody/parsing/detail_common.cc` line 277 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN8Jwl3Cn34k2BK881:-NaN8Jwl3Cn34k2BK882:ba8fxf7) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.cc#L277)):*
> ```cpp
>       (std::isfinite(mass) && mass > 0.0) ? mass : 1.0;
>   if (spoil_invalid_inertia) {
>     // The goal here is to transmit mass, to the extent we have a valid one,
> ```

btw maybe "a valid value" instead of "a valid one" since it's not, like, an object.
___
*[`multibody/parsing/detail_common.cc` line 300 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN9jj67SQVKhWazDSN:-NaN9jj67SQVKhWazDSO:bizjlqm) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.cc#L300)):*
> ```cpp
>     // Bcm to Bo.  Note: It is unwise to directly create a solid sphere about
>     // Bo as this does not guarantee that the spatial inertia about Bcm is
>     // valid. Bcm (B's center of mass) is the ground-truth point for validity
> ```

nit you seem to care about two spaces after a period but missed this one. (I don't care, although I would try to be consistent, and other comments hereabouts are consistently using a single space.)
___
*[`multibody/parsing/detail_sdf_parser.cc` line 1374 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNFKAsCJNDn5eBFte8:-NaNFKAsCJNDn5eBFte9:b-x2xhjh) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_sdf_parser.cc#L1374)):*
> ```cpp
>   drake::log()->trace("sdf_parser: Add nested models");
>   // auto resolver = workspace.collision_resolver;
>   // auto package_map = workspace.package_map;
> ```

nit looks like refactoring debris?
___
*[`multibody/parsing/parser.h` line 184 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaIun6kEmreek-8wBwe:-NaIun6kEmreek-8wBwf:b-w0u5fi) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L184)):*
> ```objc
>   /// Cause all subsequent Add*Model*() operations to use strict parsing;
>   /// warnings will be treated as errors.
>   void SetStrictParsing() { is_strict_ = true; }
> ```

btw the strict parsing option is now the only one (of three) options with no corresponding getter; is that purposeful or should we add one?
___
*[`multibody/parsing/parser.h` line 196 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaIuft0D9X6etUME3vX:-NaIuft0D9X6etUME3vY:b2chl5a) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L196)):*
> ```objc
>   /// Cause all subsequent Add*Model*() operations to replace physically
>   /// invalid inertia configurations with default-constructed SpatialInertia.
>   void SpoilInvalidInertia() { spoil_invalid_inertia_ = true; }
> ```

btw I don't mind these new method names but I note that the other options are Set/Get pairs (well, at least "SetFoo" with one "GetFoo"); should we keep that convention? If that convention is more Drake-wide then we probably should…
___
*[`multibody/parsing/parser.h` line 199 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaItmSzB6mXQy0ZMK2b:-NaItmSzB6mXQy0ZMK2c:bx876f8) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L199)):*
> ```objc
> 
>   /// Cause all subsequent Add*Model*() operations to replace physically
>   /// invalid inertia configurations with default-constructed SpatialInertia.
> ```

btw elsewhere you note that the returned SpatialInertia is NaN-filled and I wonder if the most clear phrasing is something like "default-constructed (NaN-filled) SpatialInertia". Or are the differing levels of technical detail provided on purpose?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 163 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNZWx45wspB_Ut3OQ2:-NaNZWx45wspB_Ut3OQ3:b-xavok4) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L163)):*
> ```Python
>     Args:
>         parent: the ElementFacts that will get new synthetic children
>         names: a list of XML element names to maybe insert
> ```

btw maybe "to insert if missing"?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 167 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNZha86pJkZroIY7GZ:-NaNZha86pJkZroIY7G_:byouyig) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L167)):*
> ```Python
>     assert parent.children, parent
>     kid_names = [x.name for x in parent.children]
>     kid0 = parent.children[0]
> ```

nit refactoring debris?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 185 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN_onM0LEDLjoqYuT6:-NaN_onM0LEDLjoqYuT7:b-uo6zl) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L185)):*
> ```Python
>     if is_synth_element(facts):
>         # Empty, synthetic, "pseudo" element.
>         return facts.end.index
> ```

This doesn't seem like it can be right, both by itself and in comparison to the top-level `is_synth_element` (which makes sense) and it looks like it isn't tested, so … an error?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-NaO0YPjDCizz84lrph1:bqm6380) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*
> ```Python
>         return facts.end.index
> 
>     if input_text[facts.end.index:].startswith(f"</{facts.name}"):
> ```

btw I assume that the expat parser leaves the index pointing at the start of the next tag when it consumes an element (because this works for files with inertia tags on separate lines) but could this fail for inertia blocks where there is an XML comment in the wrong place?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 482 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNTcob-_Q9bcMYqoB5:-NaNTcob-_Q9bcMYqoB6:b-vj8pmg) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L482)):*
> ```Python
> 
>     def associate_plant_models(self, plant: MultibodyPlant):
>         """Match body indices to inertial elements in the input text.
> ```

nit should we echo the return information here as well?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 506 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNyQkPBPWBANYl7mVv:-NaNyQkQ1ra7jMPZ6TaW:bdp155k) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L506)):*
> ```Python
>         input_text = self._input_text
>         input_text_index = 0
>         for body_index, new_inertia in sorted(new_inertias_mapping.items()):
> ```

btw how do body indices work for included files? It doesn't seem like we can be supporting them in this utility, but I wanted to confirm.

Originally I was trying to make sure that BodyIndex was monotonically increasing with file position (looks that way to me) and wondering why we didn't sort our mapping by element start, but from what I've read it seems like this should be fine.
___
*[`bindings/pydrake/multibody/fix_inertia.py` line 2 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNBuwg-lDWITvZxycT:-NaNBuwhE-PxvcjOX3xY:bihhhr3) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/fix_inertia.py#L2)):*
> ```Python
> r"""
> XXX docs!!
> ```

Flagging this so so no one forgets this later.
___
*[`bindings/pydrake/multibody/fix_inertia.py` line 34 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNCR7z38AeP8Livw02:-NaNCR7z38AeP8Livw03:bwqsfem) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/fix_inertia.py#L34)):*
> ```Python
>     parser.add_argument(
>         "--invalid_only", action="store_true",
>         help="only fix physically invalid inertias.")
> ```

nit Capitalize first letter
___
*[`bindings/pydrake/multibody/fix_inertia.py` line 37 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNCX-HDlrqJANbJtma:-NaNCX-I-KIIL9BqaHoK:bwqsfem) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/fix_inertia.py#L37)):*
> ```Python
>     parser.add_argument(
>         "--in_place", action="store_true",
>         help="modify the input file in-place. Any output_file argument"
> ```

nit Capitalize first letter
___
*[`bindings/pydrake/multibody/test/fix_inertia_test.py` line 321 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO92XK7PVH7axsIn4n:-NaO92XK7PVH7axsIn4o:b7dpb17) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/test/fix_inertia_test.py#L321)):*
> ```Python
>     def test_massless_sphere_urdf(self):
>         # See above comments about sdformat inertia processing; here we
>         # get a magic mass of 1 from nowhere.
> ```

Copypasta, or does this really apply to URDF as well? (The zero mass below makes me think maybe not?)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 12 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`multibody/parsing/detail_common.cc` line 270 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNAL6M7YE27sdRKvqV:-Nah0S65-gElwLXGpI8x:b-2zc3al) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.cc#L270)):*

btw this got yanked way up here but no code other than the original stanza seems to use it? Detritus from a prior refactoring?
</blockquote></details>

It needs to have broader scope than just the else block now -- it gets used again at the very end.
___
*[`multibody/parsing/detail_common.cc` line 300 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN9jj67SQVKhWazDSN:-Nah0wtq163vpsPOSeIZ:b-kxfu6k) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/detail_common.cc#L300)):*

nit you seem to care about two spaces after a period but missed this one. (I don't care, although I would try to be consistent, and other comments hereabouts are consistently using a single space.)
</blockquote></details>

I prefer the one-space -- the two-space text got pasted from another author. Will resolve to one-space.
___
*[`multibody/parsing/parser.h` line 194 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-N_uS5sB63TxEEV9fGe7:-Nah22Mo5B3q2zFrljLa:bnibrk1) ([raw file](https://github.com/RobotLocomotion/drake/blob/c0123643b61e7569363ab2bee2b3f972d7b412b7/multibody/parsing/parser.h#L194)):*

I did some skimming of the PR, but it's not yet clear to me why this feature is related to the tool.  I assume there is some reason why the rewriting tool wants to enable this, but I couldn't tease out what it is?
</blockquote></details>

The difficulty is in implementing `--invalid_only` -- the tool never attempts to hand-build inertias from XML , so forwarding the "invalid in source file" bit is necessary. There a other ways we could do this, but I was unpersuaded by the idea of writing yet-another xml-to-math stack in python.
___
*[`multibody/parsing/parser.h` line 184 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaIun6kEmreek-8wBwe:-Nah1jRo6mxpjuE2eZvp:b-brtoht) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L184)):*

btw the strict parsing option is now the only one (of three) options with no corresponding getter; is that purposeful or should we add one?
</blockquote></details>

Meh? There are tests that detect the effect of strict parsing, so the getter didn't seem so important.
___
*[`multibody/parsing/parser.h` line 196 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaIuft0D9X6etUME3vX:-Nah2oKKCASWGzg3IjbM:b-r3s42i) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L196)):*

btw I don't mind these new method names but I note that the other options are Set/Get pairs (well, at least "SetFoo" with one "GetFoo"); should we keep that convention? If that convention is more Drake-wide then we probably should…
</blockquote></details>

I don't really want this to be part of supported API, so meh?
___
*[`multibody/parsing/parser.h` line 199 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaItmSzB6mXQy0ZMK2b:-Nah2bd591hlHxLjoWha:b3mvsmj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/multibody/parsing/parser.h#L199)):*

btw elsewhere you note that the returned SpatialInertia is NaN-filled and I wonder if the most clear phrasing is something like "default-constructed (NaN-filled) SpatialInertia". Or are the differing levels of technical detail provided on purpose?
</blockquote></details>

The nan proposal didn't work out. These should get resolved to document the later implementation.
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 482 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNTcob-_Q9bcMYqoB5:-Nah5-Pj0jef6g7R9nI1:b-4jsglg) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L482)):*

nit should we echo the return information here as well?
</blockquote></details>

I think the annotations on the interface are wrong -- working.
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 506 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNyQkPBPWBANYl7mVv:-Nah4Gkw8f2-ITXZAkOo:b-du6wf0) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L506)):*

btw how do body indices work for included files? It doesn't seem like we can be supporting them in this utility, but I wanted to confirm.

Originally I was trying to make sure that BodyIndex was monotonically increasing with file position (looks that way to me) and wondering why we didn't sort our mapping by element start, but from what I've read it seems like this should be fine.
</blockquote></details>

We don't support bodies from included files. It could be that there is a test missing.
___
*[`bindings/pydrake/multibody/test/fix_inertia_test.py` line 321 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO92XK7PVH7axsIn4n:-Nah4yoM9zm8RUMcO-D_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/test/fix_inertia_test.py#L321)):*

Copypasta, or does this really apply to URDF as well? (The zero mass below makes me think maybe not?)
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 12 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`multibody/parsing/parser.h` line 194 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-N_uS5sB63TxEEV9fGe7:-NahMTgu5qcwJmqUFO05:b8gqh13) ([raw file](https://github.com/RobotLocomotion/drake/blob/c0123643b61e7569363ab2bee2b3f972d7b412b7/multibody/parsing/parser.h#L194)):*

The difficulty is in implementing `--invalid_only` -- the tool never attempts to hand-build inertias from XML , so forwarding the "invalid in source file" bit is necessary. There a other ways we could do this, but I was unpersuaded by the idea of writing yet-another xml-to-math stack in python.
</blockquote></details>

Ah, okay.

I'm not convinced that `--invalid_only` is useful in the first place.  Can you explain the use case?  It seems like scope creep.

My nominal expectation is that:

(1) Users know how to use source control, and are capable of taking whatever parts of the diffs they prefer and leaving the rest behind.

(2) Files with an invalid inertia are typically _all_ wrong, not just 10% wrong.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 12 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-NahMRpC0azL2ZvnDqHb:b-5o5kjm) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

btw I assume that the expat parser leaves the index pointing at the start of the next tag when it consumes an element (because this works for files with inertia tags on separate lines) but could this fail for inertia blocks where there is an XML comment in the wrong place?
</blockquote></details>

Perhaps there is a test missing? I'm not sure I understand the misplaced comment scenario.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 12 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

___
*[`multibody/parsing/parser.h` line 194 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-N_uS5sB63TxEEV9fGe7:-NahNVQJ2vTdkZaoTy7U:bkknxnh) ([raw file](https://github.com/RobotLocomotion/drake/blob/c0123643b61e7569363ab2bee2b3f972d7b412b7/multibody/parsing/parser.h#L194)):*

Ah, okay.

I'm not convinced that `--invalid_only` is useful in the first place.  Can you explain the use case?  It seems like scope creep.

My nominal expectation is that:

(1) Users know how to use source control, and are capable of taking whatever parts of the diffs they prefer and leaving the rest behind.

(2) Files with an invalid inertia are typically _all_ wrong, not just 10% wrong.
</blockquote></details>

If we drop invalid-only, then the whole Parser subdir set of mods can go. I like that.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 8 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 185 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN_onM0LEDLjoqYuT6:-NahP-Pe8Rqjs7HZzR6F:b9fjizj) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L185)):*

This doesn't seem like it can be right, both by itself and in comparison to the top-level `is_synth_element` (which makes sense) and it looks like it isn't tested, so … an error?
</blockquote></details>

I'm not sure what about it looks wrong? It could be that I could delete this path and fall through to the (implementation-identical) final return statement. If I add `assert False` here, it definitely gets hit by the test suite.

FWIW, somewhere (perhaps the other PR) Nimmer said he would be fine with file/text level tests, perhaps at the expense of unit tests on every class and function.
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 482 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNTcob-_Q9bcMYqoB5:-NahQ8MxCyaGte7TlIm3:b-lqtiq5) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L482)):*

I think the annotations on the interface are wrong -- working.
</blockquote></details>

This returns None. I've changed the method name to clarify that is not implementing the FormatDriver interface.

trowell-tri (commented):
> Reviewed 1 of 8 files at r4, 14 of 14 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 2 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 185 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaN_onM0LEDLjoqYuT6:-NamM96YAIaJHy4Ji8zt:b-izxvis) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L185)):*

I'm not sure what about it looks wrong? It could be that I could delete this path and fall through to the (implementation-identical) final return statement. If I add `assert False` here, it definitely gets hit by the test suite.

FWIW, somewhere (perhaps the other PR) Nimmer said he would be fine with file/text level tests, perhaps at the expense of unit tests on every class and function.
</blockquote></details>

Whoops, never mind; I completely misread this stanza to mean something entirely different and I have no idea why. Retracted.
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-NamMi3d6h3Izp2MYZP2:bv25qa8) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

Perhaps there is a test missing? I'm not sure I understand the misplaced comment scenario.
</blockquote></details>

The (very contrived) error corner case would be:

```
<inertial>
   <some useful stuff>
   <!--A comment-->
</inertial>
```

Maybe expat just skips the comment in this case?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 506 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNyQkPBPWBANYl7mVv:-NamYpn0Dh2N1V8L7MX4:b-3joygi) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L506)):*

We don't support bodies from included files. It could be that there is a test missing.
</blockquote></details>

Ah, I see now. I had missed how the association is built just from links we've read from our immediate input file, not from the entire plant. Given that, I don't see how there's a useful test case here.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 2 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-NaqjCyjE8yFii6beSFR:bxm6qs6) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

The (very contrived) error corner case would be:

```
<inertial>
   <some useful stuff>
   <!--A comment-->
</inertial>
```

Maybe expat just skips the comment in this case?
</blockquote></details>

In that case, the expat end index would still end up at the start of `</inertial>`. I think that this is in fact a very common case, given that nesting (of anything) is more-or-less the point of XML.

Perhaps the comment is misleading, because it looks like it only applies to one-line elements?

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 3 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-NaqwflREjAnLSasc2H3:b-kn7a1m) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

In that case, the expat end index would still end up at the start of `</inertial>`. I think that this is in fact a very common case, given that nesting (of anything) is more-or-less the point of XML.

Perhaps the comment is misleading, because it looks like it only applies to one-line elements?
</blockquote></details>

I don't think the comment was misleading, I just haven't read the expat docs. If any comment is needed then maybe a quick precis of what the expat is behavior could be added somewhere higher up, especially if there are non-obvious things you rely on. But on the other hand, perhaps it's silly for a future maintainer to try to work on this without reading those docs. Although if there are non-obvious behaviors you've discovered in expat (I don't mean this unless it isn't documented) then I would note them down somewhere.

I'm still not sure how I feel about it skipping comments. It's handy, but what if you want to process them separately? Anyway that's super off-topic.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 3 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-Nar5DL-D3IY9HMwRuQn:b-484qna) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

I don't think the comment was misleading, I just haven't read the expat docs. If any comment is needed then maybe a quick precis of what the expat is behavior could be added somewhere higher up, especially if there are non-obvious things you rely on. But on the other hand, perhaps it's silly for a future maintainer to try to work on this without reading those docs. Although if there are non-obvious behaviors you've discovered in expat (I don't mean this unless it isn't documented) then I would note them down somewhere.

I'm still not sure how I feel about it skipping comments. It's handy, but what if you want to process them separately? Anyway that's super off-topic.
</blockquote></details>

The comments are handled by the underlying parser by triggering different comment-specific callbacks. I don't currently use those, but I suppose we could if there were sufficient reason. My overall goal is to just preserve what comments I can (with some caveats) -- that is done by limiting how much text is replaced, and just copying the rest through from the input text.

trowell-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 2 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 187 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaO0YPjDCizz84lrph0:-Nar6fUz8DUhka-HCAaK:b-ugvacs) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/_inertia_fixer.py#L187)):*

The comments are handled by the underlying parser by triggering different comment-specific callbacks. I don't currently use those, but I suppose we could if there were sufficient reason. My overall goal is to just preserve what comments I can (with some caveats) -- that is done by limiting how much text is replaced, and just copying the rest through from the input text.
</blockquote></details>

Ah, that makes sense.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 2 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-Naqrb8YE8pKURRwCpEh:-Narkz0EEAzfMGpK3vr7:bchniex) (no related file):*

The offending file contains non-ascii UTF8 characters, which throw off the string-slicing math.

```
rico@torg:~/checkout/drake$ grep -P '[^\x00-\x7f]' examples/multibody/four_bar/four_bar.sdf 
   Wx  <- - - ·
        <!-- Rotation around Ax by 90° = π/2 radians ≈ 1.57079632679 -->
        <!-- Rotation around Bx by 90° = π/2 radians ≈ 1.57079632679 -->
        <!-- Rotation around Cx by 90° = π/2 radians ≈ 1.57079632679 -->
      <!-- This frame is rotated relative from link B, around Bx, by -90° = -π/2 radians ≈ -1.57079632679 -->
      <!-- This frame is rotated relative from link C, around Cx, by -90° = -π/2 radians ≈ -1.57079632679 -->
```

I'll have to study `expat`'s behavior and see what I can do to fix things.
</blockquote></details>

BTW Maybe expat is using `bytes` indexing instead of `str` indexing?  Possibly doing the text-replacement in `bytes` land would resolve this.

___

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 2 unresolved discussions, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-Naqrb8YE8pKURRwCpEh:-NarlSCG93wb-gAR_8nr:b-6xo8od) (no related file):*

BTW Maybe expat is using `bytes` indexing instead of `str` indexing?  Possibly doing the text-replacement in `bytes` land would resolve this.
</blockquote></details>

Yes, expat math is in bytes. I'll change the input-text handling to use bytes.

___

rpoyner-tri (commented):

I've got more doc and hopefully one more look from Todd, and then I'll nominate a platform reviewer.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-Nas7I786AWQLwMSQRo1:b-njubfn)* status: 1 unresolved discussion, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-Naqrb8YE8pKURRwCpEh:-Nas74NBD7VcVDz6Ju08:b-i2t808) (no related file):*

Yes, expat math is in bytes. I'll change the input-text handling to use bytes.
</blockquote></details>

`r6` clears up unicode difficulties and adds tests.

___
___
*[`bindings/pydrake/multibody/fix_inertia.py` line 2 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NaNBuwg-lDWITvZxycT:-Nas7D8bBWrRxD9CXapG:b-kx65jq) ([raw file](https://github.com/RobotLocomotion/drake/blob/9bb157763bbc96ee29091d289194146e1bf41bf4/bindings/pydrake/multibody/fix_inertia.py#L2)):*

Flagging this so so no one forgets this later.
</blockquote></details>

I need to write more here; keeping it `Working` for now.

rpoyner-tri:
I worry the unit tests may be floating-point brittle; let's see:

rpoyner-tri:

Fears confirmed. I'll have to figure out a reasonable python float rounding strategy, or rework the tests to avoid text comparison of floats.
</blockquote></details>

rpoyner-tri (commented):
`r7` is looking pretty good. Mac CI issues addressed, docs largely done.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NawiUef1XUWTNjkf7iz:b-ho026l)* status: LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @rpoyner-tri)

trowell-tri (commented):
> Reviewed 5 of 7 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: 1 unresolved discussion, LGTM missing from assignee trowell-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @rpoyner-tri)

___
*[`bindings/pydrake/multibody/fix_inertia.py` line 2 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-Nax9lVKEmI1CRUTt-xP:-Nax9lVKEmI1CRUTt-xQ:b-kt3hfj) ([raw file](https://github.com/RobotLocomotion/drake/blob/ea525fd6b899a11f05051d858ea112e47dff7061/bindings/pydrake/multibody/fix_inertia.py#L2)):*
> ```Python
> r"""Utility for fixing invalid inertia data in URDF or SDFormat files, by
> writing a new, complete, modified file.
> ```

BTW consider choosing a word other than "modified" because it has a commonly-used meaning wrt files that conflicts with "new".

trowell-tri (commented):
:lgtm:

> Reviewed 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NaxGASpBwk6C0eOxMm0:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+@sherm1 for Monday platform review

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NaxGtEy-fvoFE-eDF4u:b-ghhs9g)* status: LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-Nb-qAuLBpGpHMF6JTrj:b-l8xbmn)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Woo hoo! That's quite an undertaking. Platform :lgtm_strong: with a few minor comments

> Reviewed 1 of 5 files at r1, 11 of 14 files at r5, 3 of 7 files at r6, 2 of 2 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-:-NbHEBc-9LNbskam4Im7:b-567qea)* status: 8 unresolved discussions

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 72 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH2xl5CX0CyZd6UmBJ:-NbH2xl5CX0CyZd6UmBK:b6gp4p2) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L72)):*
> ```Python
> # This function has an aggressively short name for embedding into f-strings.
> def gfmt(x: float) -> str:
>     """Convert a floating point number to a string, using some rounding to
> ```

BTW in the C++ styleguide we would require descriptive "Converts" rather than imperative "Convert" in the function descriptions. Does Python have a different convention?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 112 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH3ZuX0K11ndtvq1Wg:-NbH3ZuY6uGbJs76HjeF:bwdsu13) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L112)):*
> ```Python
>             input_bytes: full contents of the input file
>             inertial_facts: facts about the element to reformat
>             pose: Pose of the center-of-mass frame relative to the body frame,
> ```

BTW consider calling that the "central inertia frame" rather than the "center of mass" frame since it has both the COM location (as the frame origin) and the basis vectors in which the inertia is expressed. Also if you give the frame a symbol here (say "C") then you can say that moments and products are "expressed in C".
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 217 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH5Jiv2MbSPh1E4ny4:-NbH5Jiv2MbSPh1E4ny5:bik1zyh) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L217)):*
> ```Python
>     d(int) -> str
>         indents n levels in addition to the current depth of the element
>          described by `facts`.
> ```

BTW did you want this 1-space indent?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 512 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH76fyBWi36AG10qT5:-NbH76fyBWi36AG10qT6:bl2jxpe) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L512)):*
> ```Python
>             new_inertias_mapping: a dictionary whose keys are BodyIndex values,
>                and whose values are a tuple of the body's spatial inertia and
>                6D inertial pose as [x, y, z, r, p, y].
> ```

BTW the meaning of "inertial pose" is unclear to me. Does it mean the pose relative to the link frame L of the central inertia frame C (i.e. X_LC?) in which the inertia was expressed? And then is the returned spatial inertia expressed in C? Can you make this more precise?
___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 584 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH8SUC2i_gS0ENubbM:-NbH8SUC2i_gS0ENubbN:bleevb6) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L584)):*
> ```Python
>                                      M_BBo_B_one.get_unit_inertia())
> 
>         # TODO: rotate to zero out products of inertia, and return both inertia
> ```

BTW consider attaching a name or issue number to this TODO
___
*[`bindings/pydrake/multibody/fix_inertia.py` line 44 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbHA3HU3gHaEDqe4eHN:-NbHA3HU3gHaEDqe4eHO:btpm7ji) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/fix_inertia.py#L44)):*
> ```Python
>     may not be exactly suited to the purpose at hand. Be sure to examine the
>     model file differences, and use `model_visualizer` to examine the resulting
>     inertial ellipses.
> ```

nit: ellipses -> ellipsoids
___
*[`bindings/pydrake/multibody/test/fix_inertia_test.py` line 245 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbHBlTz5NB8WHq5J21q:-NbHBlTz5NB8WHq5J21r:b-3t3mr4) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/test/fix_inertia_test.py#L245)):*
> ```Python
>       <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0"/>
>       <mass value="0"/>
>       <origin rpy="0 0 0" xyz="0 0 0"/>
> ```

BTW why does a massless urdf sphere stay massless while the sdf one above gets inertias? Consider adding an explanation of this oddity for the naive reader.
___
*[`bindings/pydrake/multibody/test/fix_inertia_test.py` line 442 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbHDFIC9cAq0O04HHWm:-NbHDFIC9cAq0O04HHWn:bw8csq7) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/test/fix_inertia_test.py#L442)):*
> ```Python
>         # Repeated processing of the output gives the same result. This step
>         # also ensures XML well-formed-ness and Drake parsing compatibility.
>         self.assert_files_equal(direct_result, repeat_result)
> ```

BTW nice! I was going to ask about idempotence. Can you guarantee idempotence for any already-processed file? If so consider whether to document that.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19731)* status: :shipit: complete! all discussions resolved, LGTM from assignees trowell-tri,sherm1(platform)

___
*[`bindings/pydrake/multibody/_inertia_fixer.py` line 512 at r9](https://reviewable.io/reviews/RobotLocomotion/drake/19731#-NbH76fyBWi36AG10qT5:-NbLB9HF6VKcqBBqp2qc:b-x2tpgz) ([raw file](https://github.com/RobotLocomotion/drake/blob/45a00f4934985834e5225c0dcca216b6c7020691/bindings/pydrake/multibody/_inertia_fixer.py#L512)):*

I'm giving this a go, but we may need a few iterations to get things precise.
</blockquote></details>

Did my best -- PTAL.

sherm1:
Nice!

RussTedrake:
FWIW -- I was just trying to point a student to this tool, but it took me 10 minutes (finally looking here) to find it.  I think we need to do something to improve it's discoverability?  Probably mention it in the authoring_multibody tutorial?

BTW -- since it uses AddModels, can it load an OBJ directly, and spit out an SDF?  That seems like potentially a very common/useful version of the workflow?

jwnimmer-tri:
> BTW -- since it uses AddModels, can it load an OBJ directly, and spit out an SDF? That seems like potentially a very common/useful version of the workflow?

For that use case, users should try [mesh_to_model](https://drake.mit.edu/pydrake/pydrake.multibody.mesh_to_model.html) instead.

Issue #19717, opened 2023-07-03
Transform Between HPolyhedron and VPolytope when Sets Aren't Full Dimensional


cohnt:
When transforming back and forth between representing a set as an HPolyhedron and a VPolytope, Drake makes a call to QHull. However, when this set is not full-dimensional, QHull throws an error (in both directions). QHull provides the following suggestion:

If the input is lower dimensional:
  - use 'QJ' to joggle the input and make it full dimensional
  - use 'Qbk:0Bk:0' to delete coordinate k from the input.  You should
    pick the coordinate with the least range.  The hull will have the
    correct topology.
  - **determine the flat containing the points, rotate the points
    into a coordinate plane, and delete the other coordinates.**
  - add one or more points to make the input full dimensional.

I'm planning to implement the third option, and am creating this issue to track progress across multiple PRs. My plan is the following:
 - [x] Check if a ConvexSet is empty. (Sometimes, this can be done in O(1) time. Others, it may require solving an optimization problem.) #19716
 - [x] Clarify the distinction between empty and nonempty zero-dimensional convex sets. Discussed further in #19723.  #19732
 - [x] Compute a feasible point within an arbitrary ConvexSet. (Sometimes, this can be done in O(1) time. Others, it may require solving an optimization problem.) #19804
     - [x] Add improved feasible point methods for certain ConvexSet subclasses. #19824
 - [x] Create a new subclass of ConvexSet, "AffineSubspace", described as a basis plus a translation. #19828
     - [x] Implement checks for containment and equivalence. #19850
     - [x] Implement methods for projection, projection to local coordinates, and lifting from local coordinates to ambient space. #19851
 - [x] Add constructor to AffineSubspace to compute the affine hull of another convex set. #19864
 - [x] When converting between HPolyhedron and VPolytope, if QHull throws an error due to the dimension, handle it. Compute the affine hull, change coordinates onto the flat so that it is full dimensional, perform the QHull computations, and then transform back to the original coordinate system.
     - [x] VPolytope -> HPolyhedron. #19979
     - [x] HPolyhedron -> VPolytope. #20001

VPolytope to HPolyhedron Error (not enough points):
```
import numpy as np
from pydrake.all import HPolyhedron, VPolytope
v = VPolytope(np.array([[1, 0], [2, 0]]).T)
h = HPolyhedron(v)
```
Output: `RuntimeError: QH6214 qhull input error: not enough points(2) to construct initial simplex (need 3)`

[...]

PR #19732, opened 2023-07-06
Update ConvexSet logic to properly handle zero-dimensional sets

### LINKED ISSUE #19723: For ConvexSet, Clarify the Difference Between Empty and Zero-Dimensional
This was spawned from a discussion in #19716, and is related to #19717.

Within ConvexSet and its derived classes, the difference in treatment between empty sets and zero-dimensional sets is somewhat unclear. The set {0}=R^0 is the unique vector space of dimension zero. The empty set is not a vector space at all (since a vector space must always include the zero vector), and is often assigned the dimension -1 (although this is just a convention).

The empty set can be considered as a subset of any dimensional vector space, including the zero-dimensional one. (In fact, there are only two zero-dimensional sets: {} and {0}. It's important to treat these objects distinctly -- for example, an empty HPolyhedron still keeps track of its ambient dimension. Further, towards #19717, it will be important to properly handle zero-dimensional nonempty affine subspaces.

Overall, I'm suggesting that we begin treating these two objects as distinct, and modify/update all parts of ConvexSet and its derived classes to account for this. Specifically, I propose the following changes:

- Treat the empty vector [] as being the unique element of {0}, the zero-dimensional vector space.
- Update several default constructors to have more intuitive mathematical behavior
    - **CartesianProduct**: The empty Cartesian product should certainly be zero-dimensional. By convention, it's also nonempty, so it should be the zero-dimensional set {0}. See [wikipedia](https://en.wikipedia.org/wiki/Empty_product#Nullary_Cartesian_product), or just note the fact that we should be able to add an empty product to any chain of products without changing it.
    - **Intersection**: The empty intersection should always be the whole space. We currently construct it as being zero-dimensional, and I'm fine maintaining this behavior, so it should be treated as the zero-dimensional set {0}.
    - **MinkowskiSum**: The singleton {0} (in any dimension) acts as the identity element for Minkowski addition, whereas {} is annihilatory. We currently construct it as being zero-dimensional, and I'm fine maintaining this behavior, so it should be treated as the zero-dimensional set {0}.
    - **VPolytope**: I think the default constructor in this case should be zero vertices in the zero dimensional space, so {}.
    - **Spectrahedron**: This falls back to the default constructor for ConvexSet (which isn't used anywhere else), creating a zero dimensional set, but not defining whether or not it's empty. I suggest making it nonempty by default, to match the behavior of all the other ConvexSet's default constructors.
- The other constructors can all remain unchanged.
- Update IntersectsWith to return true if both sets are {0}.
- Update PointInSet to return true when checking if 0 (represented as the empty vector []) is contained in {0}. The tolerance argument will be ignored.
- Continue throwing errors in AddPointInSetConstraints and AddPointInNonnegativeScalingConstraints when the ambient dimension is zero, since the only way this would make sense is if an empty list of variables is passed in, which seems to be clearly bad behavior.

Tagging @RussTedrake, @hongkai-dai, and @AlexandreAmice for further discussion before I start working on a PR.

cohnt:
Fixes #19723, towards the long term goal of #19717.

Aside from documenting and clarifying the behavior of the IsEmpty method when the set has an ambient dimension of zero (previously, it would throw an error), this also adds an error if you try to construct a CartesianProduct, Intersection, or MinkowskiSum with a set whose ambient dimension is zero. Not sure if that qualifies as a breaking change -- you technically could do so before, but I imagine it would have led to errors of some sort.

+@hongkai-dai for feature review

### DIFF
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -41,10 +41,16 @@ int SumAmbientDimensions(const ConvexSets& sets) {
 CartesianProduct::CartesianProduct() : CartesianProduct(ConvexSets{}) {}
 
 CartesianProduct::CartesianProduct(const ConvexSets& sets)
-    : ConvexSet(SumAmbientDimensions(sets)), sets_(sets) {}
+    : ConvexSet(SumAmbientDimensions(sets)), sets_(sets) {
+  for (const auto& s : sets_) {
+    DRAKE_THROW_UNLESS(s->ambient_dimension() > 0);
+  }
+}
 
 CartesianProduct::CartesianProduct(const ConvexSet& setA, const ConvexSet& setB)
     : ConvexSet(setA.ambient_dimension() + setB.ambient_dimension()) {
+  DRAKE_THROW_UNLESS(setA.ambient_dimension() > 0);
+  DRAKE_THROW_UNLESS(setB.ambient_dimension() > 0);
   sets_.emplace_back(setA.Clone());
   sets_.emplace_back(setB.Clone());
 }
@@ -116,7 +122,9 @@ bool CartesianProduct::DoIsBounded() const {
 
 bool CartesianProduct::DoIsEmpty() const {
   if (sets_.size() == 0) {
-    return true;
+    // By convention, the empty cartesian product is considered nonempty.
+    // See https://en.wikipedia.org/wiki/Empty_product#Nullary_Cartesian_product
+    return false;
   }
   for (const auto& s : sets_) {
     if (s->IsEmpty()) {
--- a/geometry/optimization/cartesian_product.h
+++ b/geometry/optimization/cartesian_product.h
@@ -27,13 +27,15 @@ class CartesianProduct final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(CartesianProduct)
 
-  /** Constructs a default (zero-dimensional) set. */
+  /** Constructs a default (zero-dimensional, nonempty) set. */
   CartesianProduct();
 
-  /** Constructs the product from a vector of convex sets. */
+  /** Constructs the product from a vector of convex sets. Each set must have
+  ambient_dimension greater than 0 and be nonempty.*/
   explicit CartesianProduct(const ConvexSets& sets);
 
-  /** Constructs the product from a pair of convex sets. */
+  /** Constructs the product from a pair of convex sets. Each set must have
+  ambient_dimension greater than 0 and be nonempty.*/
   CartesianProduct(const ConvexSet& setA, const ConvexSet& setB);
 
   /** Constructs the product of convex sets in the transformed coordinates:
--- a/geometry/optimization/convex_set.cc
+++ b/geometry/optimization/convex_set.cc
@@ -21,7 +21,7 @@ ConvexSet::~ConvexSet() = default;
 bool ConvexSet::IntersectsWith(const ConvexSet& other) const {
   DRAKE_THROW_UNLESS(other.ambient_dimension() == this->ambient_dimension());
   if (ambient_dimension() == 0) {
-    return false;
+    return !other.IsEmpty() && !this->IsEmpty();
   }
   solvers::MathematicalProgram prog{};
   const auto& x = prog.NewContinuousVariables(this->ambient_dimension(), "x");
@@ -32,6 +32,9 @@ bool ConvexSet::IntersectsWith(const ConvexSet& other) const {
 }
 
 bool ConvexSet::DoIsEmpty() const {
+  if (ambient_dimension() == 0) {
+    return false;
+  }
   solvers::MathematicalProgram prog;
   auto point = prog.NewContinuousVariables(ambient_dimension());
   AddPointInSetConstraints(&prog, point);
--- a/geometry/optimization/convex_set.h
+++ b/geometry/optimization/convex_set.h
@@ -93,33 +93,32 @@ class ConvexSet : public ShapeReifier {
   /** Returns true iff the set is empty. Note: for some derived classes, this
   check is trivial, but for others, it can require solving a (typically small)
   optimization problem. Check the derived class documentation for any notes.
-  @throws std::exception if ambient_dimension() == 0 */
-  bool IsEmpty() const {
-    DRAKE_THROW_UNLESS(ambient_dimension() > 0);
-    return DoIsEmpty();
-  }
+  Zero-dimensional sets are considered to be nonempty by default. Sets which
+  can be zero-dimensional and empty must handle this behavior in their
+  derived implementation of DoIsEmpty. An example of such a subclass is
+  VPolytope. */
+  bool IsEmpty() const { return DoIsEmpty(); }
 
   /** If this set trivially contains exactly one point, returns the value of
-  that point. Otherwise, returns nullopt. When ambient_dimension is zero,
-  returns nullopt. By "trivially", we mean that representation of the set
-  structurally maps to a single point; if checking for point-ness would require
-  solving an optimization program, returns nullopt. In other words, this is a
-  relatively cheap function to call. */
+  that point. Otherwise, returns nullopt. By "trivially", we mean that
+  representation of the set structurally maps to a single point; if checking
+  for point-ness would require solving an optimization program, returns nullopt.
+  In other words, this is a relatively cheap function to call. */
   std::optional<Eigen::VectorXd> MaybeGetPoint() const {
     if (ambient_dimension() == 0) {
-      return std::nullopt;
+      if (IsEmpty()) {
+        return std::nullopt;
+      } else {
+        return Eigen::VectorXd::Zero(0);
+      }
     }
     return DoMaybeGetPoint();
   }
 
-  /** Returns true iff the point x is contained in the set.  When
-  ambient_dimension is zero, returns false. */
+  /** Returns true iff the point x is contained in the set. */
   bool PointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
                   double tol = 0) const {
     DRAKE_THROW_UNLESS(x.size() == ambient_dimension());
-    if (ambient_dimension() == 0) {
-      return false;
-    }
     return DoPointInSet(x, tol);
   }
 
@@ -219,23 +218,26 @@ class ConvexSet : public ShapeReifier {
   virtual std::unique_ptr<ConvexSet> DoClone() const = 0;
 
   /** Non-virtual interface implementation for IsBounded().
-  @pre ambient_dimension() > 0 */
+  @pre ambient_dimension() >= 0 */
   virtual bool DoIsBounded() const = 0;
 
   /** Non-virtual interface implementation for IsEmpty(). The default
   implementation solves a feasibility optimization problem, but derived
-  classes can override with a custom (more efficient) implementation. */
+  classes can override with a custom (more efficient) implementation.
+  Zero-dimensional sets are considered to be nonempty by default. Sets which
+  can be zero-dimensional and empty must handle this behavior in their
+  derived implementation of DoIsEmpty. */
   virtual bool DoIsEmpty() const;
 
   /** Non-virtual interface implementation for MaybeGetPoint(). The default
   implementation returns nullopt. Sets that can model a single point should
   override with a custom implementation.
-  @pre ambient_dimension() > 0 */
+  @pre ambient_dimension() >= 0 */
   virtual std::optional<Eigen::VectorXd> DoMaybeGetPoint() const;
 
   /** Non-virtual interface implementation for PointInSet().
   @pre x.size() == ambient_dimension()
-  @pre ambient_dimension() > 0 */
+  @pre ambient_dimension() >= 0 */
   virtual bool DoPointInSet(const Eigen::Ref<const Eigen::VectorXd>& x,
                             double tol) const = 0;
 
--- a/geometry/optimization/hpolyhedron.cc
+++ b/geometry/optimization/hpolyhedron.cc
@@ -354,6 +354,10 @@ bool HPolyhedron::DoIsBounded() const {
 }
 
 bool HPolyhedron::DoIsEmpty() const {
+  if (ambient_dimension() == 0) {
+    // By convention, a zero-dimensional HPolyhedron is considered nonempty.
+    return false;
+  }
   solvers::MathematicalProgram prog;
   solvers::VectorXDecisionVariable x =
       prog.NewContinuousVariables(A_.cols(), "x");
--- a/geometry/optimization/hpolyhedron.h
+++ b/geometry/optimization/hpolyhedron.h
@@ -23,7 +23,7 @@ class HPolyhedron final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(HPolyhedron)
 
-  /** Constructs a default (zero-dimensional) polyhedron. */
+  /** Constructs a default (zero-dimensional, nonempty) polyhedron. */
   HPolyhedron();
 
   /** Constructs the polyhedron.
--- a/geometry/optimization/hyperellipsoid.h
+++ b/geometry/optimization/hyperellipsoid.h
@@ -29,7 +29,7 @@ class Hyperellipsoid final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Hyperellipsoid)
 
-  /** Constructs a default (zero-dimensional) set. */
+  /** Constructs a default (zero-dimensional, nonempty) set. */
   Hyperellipsoid();
 
   /** Constructs the ellipsoid.
--- a/geometry/optimization/intersection.cc
+++ b/geometry/optimization/intersection.cc
@@ -4,6 +4,8 @@
 
 #include <fmt/format.h>
 
+#include "drake/solvers/solve.h"
+
 namespace drake {
 namespace geometry {
 namespace optimization {
@@ -35,11 +37,16 @@ int GetAmbientDimension(const ConvexSets& sets) {
 Intersection::Intersection() : Intersection(ConvexSets{}) {}
 
 Intersection::Intersection(const ConvexSets& sets)
-    : ConvexSet(GetAmbientDimension(sets)), sets_(sets) {}
+    : ConvexSet(GetAmbientDimension(sets)), sets_(sets) {
+  for (const auto& s : sets_) {
+    DRAKE_THROW_UNLESS(s->ambient_dimension() > 0);
+  }
+}
 
 Intersection::Intersection(const ConvexSet& setA, const ConvexSet& setB)
     : ConvexSet(setA.ambient_dimension()) {
   DRAKE_THROW_UNLESS(setB.ambient_dimension() == setA.ambient_dimension());
+  DRAKE_THROW_UNLESS(setB.ambient_dimension() > 0);
   sets_.emplace_back(setA.Clone());
   sets_.emplace_back(setB.Clone());
 }
@@ -66,6 +73,22 @@ bool Intersection::DoIsBounded() const {
   throw std::runtime_error(
       "Determining the boundedness of an Intersection made up of unbounded "
       "elements is not currently supported.");
+}
+
+bool Intersection::DoIsEmpty() const {
+  if (sets_.size() == 0) {
+    // When we have sets_.size() == 0, we treat the intersection as being
+    // {0}, the unique zero-dimensional vector space, which is nonempty.
+    return false;
+  }
+  // The empty set is annihilatory in intersection.
+  for (const auto& s : sets_) {
+    if (s->IsEmpty()) {
+      return true;
+    }
+  }
+  // Now actually see if the intersection is nonempty.
+  return ConvexSet::DoIsEmpty();
 }
 
 std::optional<VectorXd> Intersection::DoMaybeGetPoint() const {
--- a/geometry/optimization/intersection.h
+++ b/geometry/optimization/intersection.h
@@ -20,13 +20,15 @@ class Intersection final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Intersection)
 
-  /** Constructs a default (zero-dimensional) set. */
+  /** Constructs a default (zero-dimensional, nonempty) set. */
   Intersection();
 
-  /** Constructs the intersection from a vector of convex sets. */
+  /** Constructs the intersection from a vector of convex sets. Each set must
+  have ambient_dimension greater than 0 and be nonempty. */
   explicit Intersection(const ConvexSets& sets);
 
-  /** Constructs the intersection from a pair of convex sets. */
+  /** Constructs the intersection from a pair of convex sets. Each set must have
+  ambient_dimension greater than 0 and be nonempty. */
   Intersection(const ConvexSet& setA, const ConvexSet& setB);
 
   ~Intersection() final;
@@ -42,6 +44,8 @@ class Intersection final : public ConvexSet {
   std::unique_ptr<ConvexSet> DoClone() const final;
 
   bool DoIsBounded() const final;
+
+  bool DoIsEmpty() const final;
 
   std::optional<Eigen::VectorXd> DoMaybeGetPoint() const final;
 
--- a/geometry/optimization/minkowski_sum.cc
+++ b/geometry/optimization/minkowski_sum.cc
@@ -45,11 +45,16 @@ int GetAmbientDimension(const ConvexSets& sets) {
 MinkowskiSum::MinkowskiSum() : MinkowskiSum(ConvexSets{}) {}
 
 MinkowskiSum::MinkowskiSum(const ConvexSets& sets)
-    : ConvexSet(GetAmbientDimension(sets)), sets_(sets) {}
+    : ConvexSet(GetAmbientDimension(sets)), sets_(sets) {
+  for (const auto& s : sets_) {
+    DRAKE_THROW_UNLESS(s->ambient_dimension() > 0);
+  }
+}
 
 MinkowskiSum::MinkowskiSum(const ConvexSet& setA, const ConvexSet& setB)
     : ConvexSet(setA.ambient_dimension()) {
   DRAKE_THROW_UNLESS(setB.ambient_dimension() == setA.ambient_dimension());
+  DRAKE_THROW_UNLESS(setB.ambient_dimension() > 0);
   sets_.emplace_back(setA.Clone());
   sets_.emplace_back(setB.Clone());
 }
@@ -105,7 +110,9 @@ bool MinkowskiSum::DoIsBounded() const {
 
 bool MinkowskiSum::DoIsEmpty() const {
   if (sets_.size() == 0) {
-    return true;
+    // When we have sets_.size() == 0, we treat the Minkowski sum as being
+    // {0}, the unique zero-dimensional vector space, which is nonempty.
+    return false;
   }
   // The empty set is annihilatory in Minkowski addition.
   for (const auto& s : sets_) {
--- a/geometry/optimization/minkowski_sum.h
+++ b/geometry/optimization/minkowski_sum.h
@@ -21,13 +21,15 @@ class MinkowskiSum final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(MinkowskiSum)
 
-  /** Constructs a default (zero-dimensional) set. */
+  /** Constructs a default (zero-dimensional, nonempty) set. */
   MinkowskiSum();
 
-  /** Constructs the sum from a vector of convex sets. */
+  /** Constructs the sum from a vector of convex sets. Each set must have
+  ambient_dimension greater than 0 and be nonempty.*/
   explicit MinkowskiSum(const ConvexSets& sets);
 
-  /** Constructs the sum from a pair of convex sets. */
+  /** Constructs the sum from a pair of convex sets. Each set must have
+  ambient_dimension greater than 0 and be nonempty. */
   MinkowskiSum(const ConvexSet& setA, const ConvexSet& setB);
 
   /** Constructs a MinkowskiSum from a SceneGraph geometry and pose in the
--- a/geometry/optimization/point.h
+++ b/geometry/optimization/point.h
@@ -19,7 +19,7 @@ class Point final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Point)
 
-  /** Constructs a default (zero-dimensional) set. */
+  /** Constructs a default (zero-dimensional, nonempty) set. */
   Point();
 
   /** Constructs a Point. */
--- a/geometry/optimization/spectrahedron.h
+++ b/geometry/optimization/spectrahedron.h
@@ -21,7 +21,7 @@ class Spectrahedron final : public ConvexSet {
  public:
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(Spectrahedron)
 
-  /** Default constructor (which constructs the empty set). */
+  /** Default constructor (yields the zero-dimensional nonempty set). */
   Spectrahedron();
 
   /** Constructs the spectrahedron from a MathematicalProgram.
--- a/geometry/optimization/test/cartesian_product_test.cc
+++ b/geometry/optimization/test/cartesian_product_test.cc
@@ -77,11 +77,11 @@ GTEST_TEST(CartesianProductTest, DefaultCtor) {
   EXPECT_EQ(dut.num_factors(), 0);
   EXPECT_NO_THROW(dut.Clone());
   EXPECT_EQ(dut.ambient_dimension(), 0);
-  EXPECT_FALSE(dut.IntersectsWith(dut));
+  EXPECT_TRUE(dut.IntersectsWith(dut));
   EXPECT_TRUE(dut.IsBounded());
-  EXPECT_THROW(dut.IsEmpty(), std::exception);
-  EXPECT_FALSE(dut.MaybeGetPoint().has_value());
-  EXPECT_FALSE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
+  EXPECT_FALSE(dut.IsEmpty());
+  EXPECT_TRUE(dut.MaybeGetPoint().has_value());
+  EXPECT_TRUE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
 }
 
 GTEST_TEST(CartesianProductTest, Move) {
@@ -417,23 +417,24 @@ GTEST_TEST(CartesianProductTest, Rotated) {
   EXPECT_FALSE(Solve(prog).is_success());
 }
 
-GTEST_TEST(CartesianProductTest, EmptyCartesianProductTest) {
-  Eigen::MatrixXd A{5, 3};
-  Eigen::VectorXd b{5};
-  // Rows 1-3 define an infeasible set of inequalities.
-  // clang-format off
-  A << 1, 0, 0,
-       1, -1, 0,
-       -1, 0, 1,
-       0, 1, -1,
-       0, 0, -1;
-  b << 1, -1, -1, -1, 0;
-  // clang-format off
+GTEST_TEST(CartesianProductTest, ZeroDimensionalInput) {
+  // If any of the inputs are zero dimensional, it should throw an error.
+  const Point P1(Vector2d{1.2, 3.4});
+  const Point P2(Eigen::VectorXd::Zero(0));
+  EXPECT_THROW(CartesianProduct(P1, P2), std::exception);
 
-  const HPolyhedron H1{A, b};
-  const Point P1(Vector3d{0.1, 1.2, 0.3});
-  const CartesianProduct S(P1, H1);
+  ConvexSets sets;
+  sets.emplace_back(P1);
+  sets.emplace_back(P2);
+  EXPECT_THROW((CartesianProduct(sets)), std::exception);
+}
 
+GTEST_TEST(CartesianProductTest, EmptyInput) {
+  const Point P(Vector2d{1.2, 3.4});
+  // A VPolytope constructed from an empty list of vertices is empty.
+  Eigen::Matrix<double, 2, 0> vertices;
+  const VPolytope V = VPolytope(vertices);
+  CartesianProduct S(P, V);
   EXPECT_TRUE(S.IsEmpty());
 }
 
--- a/geometry/optimization/test/hpolyhedron_test.cc
+++ b/geometry/optimization/test/hpolyhedron_test.cc
@@ -45,10 +45,10 @@ GTEST_TEST(HPolyhedronTest, DefaultConstructor) {
   EXPECT_EQ(H.A().size(), 0);
   EXPECT_EQ(H.b().size(), 0);
   EXPECT_NO_THROW(H.Clone());
-  EXPECT_FALSE(H.IntersectsWith(H));
+  EXPECT_TRUE(H.IntersectsWith(H));
   EXPECT_TRUE(H.IsBounded());
-  EXPECT_THROW(H.IsEmpty(), std::exception);
-  EXPECT_FALSE(H.PointInSet(Eigen::VectorXd::Zero(0)));
+  EXPECT_FALSE(H.IsEmpty());
+  EXPECT_TRUE(H.PointInSet(Eigen::VectorXd::Zero(0)));
 }
 
 GTEST_TEST(HPolyhedronTest, UnitBoxTest) {
--- a/geometry/optimization/test/hyperellipsoid_test.cc
+++ b/geometry/optimization/test/hyperellipsoid_test.cc
@@ -102,10 +102,10 @@ GTEST_TEST(HyperellipsoidTest, DefaultCtor) {
   EXPECT_THROW(dut.MinimumUniformScalingToTouch(dut), std::exception);
   EXPECT_NO_THROW(dut.Clone());
   EXPECT_EQ(dut.ambient_dimension(), 0);
-  EXPECT_FALSE(dut.IntersectsWith(dut));
+  EXPECT_TRUE(dut.IntersectsWith(dut));
   EXPECT_TRUE(dut.IsBounded());
-  EXPECT_THROW(dut.IsEmpty(), std::exception);
-  EXPECT_FALSE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
+  EXPECT_FALSE(dut.IsEmpty());
+  EXPECT_TRUE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
 }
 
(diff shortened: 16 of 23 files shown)


cohnt:
Let's see if I can do this right...

hongkai-dai (commented):
Sorry for the delay. PTAL.

> Reviewed 5 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-NZoAC596viuIuu6ovdd:b-osmlt4)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/cartesian_product.h` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZhUEgX14t91h-HQlEa:-NZhUEgX14t91h-HQlEb:btqcwvz) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/cartesian_product.h#L34)):*
> ```objc
> 
>   /** Constructs the product from a vector of convex sets. Each set must have
>   ambient_dimension greater than 0 and be nonempty.*/
> ```

Why we want to require that each set is non-empty? It is not checked in the code.
___
*[`geometry/optimization/convex_set.h` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo9a51DU011NQN4-xb:-NZo9a51DU011NQN4-xc:b-xk4usj) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/convex_set.h#L99)):*
> ```objc
>   can be zero-dimensional and empty must handle this behavior in their
>   derived implementation of DoIsEmpty. An example of such a subclass is
>   VPolytope. */
> ```

I would suggest to explain more on what is a zero-dimensional set. I like what you wrote in the github issue, that 0-dimensional set is a vector space with only one element, namely 0 ∈ R⁰
___
*[`geometry/optimization/convex_set.h` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo9xHC7PV1AfoDeYHg:-NZo9xHC7PV1AfoDeYHh:b-2v3urd) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/convex_set.h#L118)):*
> ```objc
>   }
> 
>   /** Returns true iff the point x is contained in the set. */
> ```

I would suggest to mention what happens when ambient_dimension is zero. So far as I see, when ambient_dimension is zero, the only element in the zero-dimension set is 0 ∈ R⁰, hence we return true if x.size() = 0, and false otherwise.
___
*[`geometry/optimization/vpolytope.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo84VM0KKbakN9GjNW:-NZo84VM0KKbakN9GjNX:b-xht1eh) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.h#L29)):*
> ```objc
>   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(VPolytope)
> 
>   /** Constructs a default (zero-dimensional, empty) set. */
> ```

I would suggest some explanation here, why the zero-dimensional v polytope is empty?
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-NZo7WwR9mngyxrV2sAc:b-xi2ao3) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*
> ```cpp
>   if (vertices_.size() == 0) {
>     return false;
>   } else if (ambient_dimension() == 0) {
> ```

I am not sure I understand this logic here.

Why we allow `vertices_.size() != 0`, and `ambient_dimension() == 0`? I would think that we should have the invariant vertices_.rows() == ambient_dimension(). Hence ambient_dimension() == 0 should imply vertices_.size() == 0?

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/cartesian_product.h` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZhUEgX14t91h-HQlEa:-N_ZroOuEURmQGddAEjX:b8r9793) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/cartesian_product.h#L34)):*

Why we want to require that each set is non-empty? It is not checked in the code.
</blockquote></details>

I had added this because it would require some strange logic within some of the functions (e.g. DoAddPointInSetConstraints). I went ahead and implemented that logic, as well as sending a warning to the user, since it's probably unintended behavior.

I've also added many test cases to check this logic.
___
*[`geometry/optimization/convex_set.h` line 99 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo9a51DU011NQN4-xb:-N_ZsDjt9nR3VOrJBnB6:b-902fn2) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/convex_set.h#L99)):*

I would suggest to explain more on what is a zero-dimensional set. I like what you wrote in the github issue, that 0-dimensional set is a vector space with only one element, namely 0 ∈ R⁰
</blockquote></details>

Note that there are two different zero dimensional sets -- the zero vector space (which is nonempty), and the empty set with ambient dimension zero.
___
*[`geometry/optimization/convex_set.h` line 118 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo9xHC7PV1AfoDeYHg:-N_ZsLL85F-tpCw-laca:booa8al) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/convex_set.h#L118)):*

I would suggest to mention what happens when ambient_dimension is zero. So far as I see, when ambient_dimension is zero, the only element in the zero-dimension set is 0 ∈ R⁰, hence we return true if x.size() = 0, and false otherwise.
</blockquote></details>

Good call on checking that here. We do need to split apart the cases for whether or not the set is empty when it is zero-dimensional.
___
*[`geometry/optimization/vpolytope.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo84VM0KKbakN9GjNW:-N_ZsbmM9ZoMbaRA-fFQ:b-7shqg0) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.h#L29)):*

I would suggest some explanation here, why the zero-dimensional v polytope is empty?
</blockquote></details>

The default constructor creates a VPolytope whose vertex list is 0-by-0, implying that it's a list of zero vertices in a zero-dimensional space. Thus, it should be empty.

Likewise, one could also construct a VPolytope with zero vertices in an n-dimensional space. This set will still be empty, but its ambient dimension will be n.
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_ZtVs31cQ7hbPQ9Lqf:b9qca76) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*

I am not sure I understand this logic here.

Why we allow `vertices_.size() != 0`, and `ambient_dimension() == 0`? I would think that we should have the invariant vertices_.rows() == ambient_dimension(). Hence ambient_dimension() == 0 should imply vertices_.size() == 0?
</blockquote></details>

Done.

Good catch!

hongkai-dai (commented):
> Reviewed 3 of 23 files at r1, 2 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/cartesian_product.h` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZhUEgX14t91h-HQlEa:-N_aRkI15yg1Bivck83o:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/cartesian_product.h#L34)):*

I had added this because it would require some strange logic within some of the functions (e.g. DoAddPointInSetConstraints). I went ahead and implemented that logic, as well as sending a warning to the user, since it's probably unintended behavior.

I've also added many test cases to check this logic.
</blockquote></details>

Thanks!
___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_bgRNF8dKpd7kcUnuN:b10svqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*
> ```cpp
> 
> bool ConvexSet::DoIsEmpty() const {
>   if (ambient_dimension() == 0) {
> ```

I am not sure about this one. When the dimension is 0, the set could be either an empty set or a singleton set {0}. In the latter case it is not empty.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_cBy1w8S_mbIhiM-_n:b-v90j8x) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

I am not sure about this one. When the dimension is 0, the set could be either an empty set or a singleton set {0}. In the latter case it is not empty.
</blockquote></details>

This check is needed here, as otherwise, it's possible for AddPointInSetConstraints to be called -- that method will always throw an error if the ambient dimension is zero. We treat this as the "default" behavior (since most subclasses of ConvexSet follow this convention), but give the opportunity for that behavior to be overridden (as is done in VPolytope and Intersection, for example).

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_icuT12M2fLY4yXuKu:bwue2gf) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

This check is needed here, as otherwise, it's possible for AddPointInSetConstraints to be called -- that method will always throw an error if the ambient dimension is zero. We treat this as the "default" behavior (since most subclasses of ConvexSet follow this convention), but give the opportunity for that behavior to be overridden (as is done in VPolytope and Intersection, for example).
</blockquote></details>

Possibly that is subtle enough that it would be worth adding a comment here highlighting the situation to other Drake developers.  If Hongkai was momentarily confused about it, others will be too.

The header file comment for `DoIsEmpty` is pretty clear on this point.  Perhaps just citing that here would be sufficient, or repeat a quick summary of it inline here.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_ig3Qn2ijmwT0bL7xv:b-x9ehj4) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

Possibly that is subtle enough that it would be worth adding a comment here highlighting the situation to other Drake developers.  If Hongkai was momentarily confused about it, others will be too.

The header file comment for `DoIsEmpty` is pretty clear on this point.  Perhaps just citing that here would be sufficient, or repeat a quick summary of it inline here.
</blockquote></details>

Good call. How's this sound?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_iozPy2cSfyazzMdgP:b-9wa6ak) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

Good call. How's this sound?
</blockquote></details>

LGTM.  I'll let Hongkai respond, too.

hongkai-dai (commented):
> Reviewed 3 of 23 files at r1, 4 of 10 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 8 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/cartesian_product.cc` line 204 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_j0whj0B5QHaN8rVJn:-N_j0whj0B5QHaN8rVJo:be2rkjq) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/cartesian_product.cc#L204)):*
> ```cpp
>   int index = 0;
>   for (const auto& s : sets_) {
>     if (s->ambient_dimension() == 0) {
> ```

BTW, we have a "rule of 3" that if a code block is repeated for 3 or more times, we should factor it out into a function.
___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_iq8TD9EgrS8Md0gsP:bxkyxlt) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

LGTM.  I'll let Hongkai respond, too.
</blockquote></details>

Thanks! This is very clear. Sorry I forgot the documentation in the header class.
___
*[`geometry/optimization/intersection.h` line 27 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ivJBn1eAzXt08PHz2:-N_ivJBn1eAzXt08PHz3:bp6m4k6) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.h#L27)):*
> ```objc
> 
>   /** Constructs the intersection from a vector of convex sets. Each set must
>   have ambient_dimension greater than 0 and be nonempty. */
> ```

BTW, the constructor doesn't throw an error if a set has ambient_dimension = 0 or is empty. Would you like to throw an error? Ditto for MinkowskiSum
___
*[`geometry/optimization/intersection.cc` line 7 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_iqkZ28zMwbJUn4V8_:-N_iqkZ28zMwbJUn4V8a:b1wk7z8) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L7)):*
> ```cpp
> #include <fmt/format.h>
> 
> #include "drake/solvers/solve.h"
> ```

This is not needed?
___
*[`geometry/optimization/intersection.cc` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ita-n403fL9sHBzzW:-N_ita-n403fL9sHBzzX:b1tzppa) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L75)):*
> ```cpp
> bool Intersection::DoIsEmpty() const {
>   if (sets_.size() == 0) {
>     // When we have sets_.size() == 0, we treat the intersection as being
> ```

I would suggest to put this documentation in the header file for the entire class. The users probably won't read this documentation in the cc file.

Generally I would suggest that in each ConvexSet subclass documentation, we mention what happens when ambient_dimension =  0 for this class? Is it empty?

I often see that the default class constructor mentions that "construct a 0-dimensional nonempty set", which suggests that ambient_dimension = 0 is NOT an empty set for this class. Could you mention this in the class documention also?
___
*[`geometry/optimization/minkowski_sum.h` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_iuxbABIm4L6wMpB2n:-N_iuxbABIm4L6wMpB2o:b-89a4l7) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/minkowski_sum.h#L32)):*
> ```objc
> 
>   /** Constructs the sum from a pair of convex sets. Each set must have
>   ambient_dimension greater than 0 and be nonempty. */
> ```

Ditto
___
*[`geometry/optimization/minkowski_sum.cc` line 176 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_izVRtCEKGwtboHDvu:-N_izVRtCEKGwtboHDvv:bcjfxdq) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/minkowski_sum.cc#L176)):*
> ```cpp
>   }
>   for (int j = 0; j < num_terms(); ++j) {
>     if (sets_[j]->ambient_dimension() == 0) {
> ```

BTW, we have a "rule of 3", that if a code block is repeated for 3 or more times, then we should factor it out as a separate function.
___
*[`geometry/optimization/vpolytope.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo84VM0KKbakN9GjNW:-N_izrhdBLiUJg5iPjM2:b-kufkxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.h#L29)):*

The default constructor creates a VPolytope whose vertex list is 0-by-0, implying that it's a list of zero vertices in a zero-dimensional space. Thus, it should be empty.

Likewise, one could also construct a VPolytope with zero vertices in an n-dimensional space. This set will still be empty, but its ambient dimension will be n.
</blockquote></details>

I see, thanks!
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_j-xm91Io7DkZpZYFY:be3zmsx) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*

Done.

Good catch!
</blockquote></details>

I would like to make sure I get the logic correct here.

If vertices_ is a 0 x n matrix where n != 0, namely we have several vertices, and each vertex is a 0-dimensional vector, then `x` is in the v-polytope if and only if `x` is zero dimensional.

Otherwise if `vertices_` is an `n x 0` matrix where `n` can also 0, then `x` cannot be in the set.

Is this right? If so, then I think the code should be
```cc
if (vertices_.cols() == 0) {
  return false;
} else if (vertices.rows() == 0) {
  // ambient_dimension = 0, but the v-polytope is not empty.
  return x.size() == 0;
}
```

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 7 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/cartesian_product.cc` line 204 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_j0whj0B5QHaN8rVJn:-N_jZjAE5zLB6EVk1k0u:b9lgvlp) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/cartesian_product.cc#L204)):*

BTW, we have a "rule of 3" that if a code block is repeated for 3 or more times, we should factor it out into a function.
</blockquote></details>

I moved this block to a function in ConvexSet, so CartesianProduct and MinkowskiSum can both use it. No more repeated code block :)
___
*[`geometry/optimization/convex_set.cc` line 35 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_bgRNF8dKpd7kcUnuM:-N_jZwzW1UXmrMPUY9yg:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cbdec9ce571a1e6ed62c335e4f5887217fc6f232/geometry/optimization/convex_set.cc#L35)):*

Thanks! This is very clear. Sorry I forgot the documentation in the header class.
</blockquote></details>

Done.
___
*[`geometry/optimization/intersection.h` line 27 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ivJBn1eAzXt08PHz2:-N_jZyeI5VHWm8n4LEMO:b-6s2i41) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.h#L27)):*

BTW, the constructor doesn't throw an error if a set has ambient_dimension = 0 or is empty. Would you like to throw an error? Ditto for MinkowskiSum
</blockquote></details>

No error necessary, since we added logic to handle this case. Removed both comments.
___
*[`geometry/optimization/intersection.cc` line 7 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_iqkZ28zMwbJUn4V8_:-N_j_6Ci751cLsy-RLhm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L7)):*

This is not needed?
</blockquote></details>

Done.
___
*[`geometry/optimization/intersection.cc` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ita-n403fL9sHBzzW:-N_j_6orDbi7yR9UaqSO:b-n2lc2l) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L75)):*

I would suggest to put this documentation in the header file for the entire class. The users probably won't read this documentation in the cc file.

Generally I would suggest that in each ConvexSet subclass documentation, we mention what happens when ambient_dimension =  0 for this class? Is it empty?

I often see that the default class constructor mentions that "construct a 0-dimensional nonempty set", which suggests that ambient_dimension = 0 is NOT an empty set for this class. Could you mention this in the class documention also?
</blockquote></details>

I've added documentation throughout the various header files explaining any quirks in the behavior of DoIsEmpty for the various subclasses. I'm not sure what does and doesn't appear in the Doxygen -- please let me know if there are any particular spots in the code you want me to put the documentation.
___
*[`geometry/optimization/minkowski_sum.h` line 32 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_iuxbABIm4L6wMpB2n:-N_j_W3QEq3mo1qq1LBg:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/minkowski_sum.h#L32)):*

Ditto
</blockquote></details>

Done.
___
*[`geometry/optimization/vpolytope.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo84VM0KKbakN9GjNW:-N_j_a_XEv4dXXzL1SN4:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.h#L29)):*

I see, thanks!
</blockquote></details>

Done.
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_j_c4F2q7Acres-S_W:b-rth355) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*

I would like to make sure I get the logic correct here.

If vertices_ is a 0 x n matrix where n != 0, namely we have several vertices, and each vertex is a 0-dimensional vector, then `x` is in the v-polytope if and only if `x` is zero dimensional.

Otherwise if `vertices_` is an `n x 0` matrix where `n` can also 0, then `x` cannot be in the set.

Is this right? If so, then I think the code should be
```cc
if (vertices_.cols() == 0) {
  return false;
} else if (vertices.rows() == 0) {
  // ambient_dimension = 0, but the v-polytope is not empty.
  return x.size() == 0;
}
```
</blockquote></details>

I think we just need to check the number of columns of vertices_. If it has zero columns, then it has zero points, so it's empty. If it has n columns, then it has n points, so it's nonempty. A 0xn array denotes n copies of 0, contained in {0}.

hongkai-dai (commented):
> Reviewed 11 of 15 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 1 unresolved discussion, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/intersection.cc` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ita-n403fL9sHBzzW:-N_jsdg33IYPjAQ9m1II:bpj08di) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L75)):*

I've added documentation throughout the various header files explaining any quirks in the behavior of DoIsEmpty for the various subclasses. I'm not sure what does and doesn't appear in the Doxygen -- please let me know if there are any particular spots in the code you want me to put the documentation.
</blockquote></details>

Any `/** */` or `///` in the public or protected methods will appear in doxygen. The private method documentation don't. 

Many documentations (such as mentioning if a zero-dimensional set is empty or not) is added on the private function `DoIsEmpty()`, hence they will not show up in doxygen. That's why I suggested to explain it in the class documentation.

BTW, if you want to check the documentation, you can build and test it locally by following the instructions on this page https://drake.mit.edu/documentation_instructions.html
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_jtiPiC4ciY5C1bEzA:bylbqho) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*
> I think we just need to check the number of columns of vertices\_. If it has zero columns, then it has zero points, so it's empty. If it has n columns, then it has n points, so it's nonempty. A 0xn array denotes n copies of 0, contained in {0}.

I agree.

But currently the code checks `vertices_.size() == 0`, not `vertices_.cols() == 0`? Moreover, if `vertices_.cols() != 0`, then we will solve an optimization program to check if the point is within the set. That seems too expensive to me for `vertices_` being an `0 x n` matrix. For this special case (`vertices_` being 0 x n matrix), we know it should always return true.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732)* status: 1 unresolved discussion, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/intersection.cc` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ita-n403fL9sHBzzW:-N_jyCZf5lFLneiT1nlt:b-g1iczc) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L75)):*

Any `/** */` or `///` in the public or protected methods will appear in doxygen. The private method documentation don't. 

Many documentations (such as mentioning if a zero-dimensional set is empty or not) is added on the private function `DoIsEmpty()`, hence they will not show up in doxygen. That's why I suggested to explain it in the class documentation.

BTW, if you want to check the documentation, you can build and test it locally by following the instructions on this page https://drake.mit.edu/documentation_instructions.html
</blockquote></details>

Done!

I've added an explanation of any quirks in IsEmpty (including the zero-dimensional case) to the documentation of all subclasses.
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_jy8Sq75VoQ1skiRW-:b-uzb0ii) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*

> I think we just need to check the number of columns of vertices\_. If it has zero columns, then it has zero points, so it's empty. If it has n columns, then it has n points, so it's nonempty. A 0xn array denotes n copies of 0, contained in {0}.

I agree.

But currently the code checks `vertices_.size() == 0`, not `vertices_.cols() == 0`? Moreover, if `vertices_.cols() != 0`, then we will solve an optimization program to check if the point is within the set. That seems too expensive to me for `vertices_` being an `0 x n` matrix. For this special case (`vertices_` being 0 x n matrix), we know it should always return true.
</blockquote></details>

Ah, I had fixed `IsEmpty`, but not `DoPointInSet`. It should be good now.

The special case where `vertices_` is a `0 x n` matrix can never occur, because `PointInSet` already handles the case where the ambient dimension is zero. (See: `convex_set.h:126`)

hongkai-dai (commented):
:lgtm: +@sherm1 for platform review please, thanks!

> Reviewed 1 of 10 files at r2, 1 of 15 files at r4, 8 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_k0HCD4WewP_4RYOup:b-z000um)* status: LGTM missing from assignee sherm1(platform), missing label for release notes (waiting on @cohnt)

___
*[`geometry/optimization/intersection.cc` line 75 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_ita-n403fL9sHBzzW:-N_k-f81E_C-hNEeA6g8:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/cfb3ba980c01a0887d20c075bf81ec87ab95e7fd/geometry/optimization/intersection.cc#L75)):*

Done!

I've added an explanation of any quirks in IsEmpty (including the zero-dimensional case) to the documentation of all subclasses.
</blockquote></details>

Thanks!
___
*[`geometry/optimization/vpolytope.cc` line 350 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-NZo7WwR9mngyxrV2sAb:-N_k01GS7fh6nhLxIk-m:bdddyif) ([raw file](https://github.com/RobotLocomotion/drake/blob/b2b78a5a4fac3a1a291eb435a9f86f348b7d9955/geometry/optimization/vpolytope.cc#L350)):*

Ah, I had fixed `IsEmpty`, but not `DoPointInSet`. It should be good now.

The special case where `vertices_` is a `0 x n` matrix can never occur, because `PointInSet` already handles the case where the ambient dimension is zero. (See: `convex_set.h:126`)
</blockquote></details>

Ah I see, I forgot that the base class already handles the 0 dimensional case. Sorry.

sherm1 (commented):
I'm assuming this is a breaking change since the definition of IsEmpty() has changed: +(release notes: breaking change) 
Platform :lgtm: pending some minor comments, PTAL

> Reviewed 5 of 23 files at r1, 3 of 10 files at r2, 6 of 15 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_kH27bB0y_wGTYlFcT:bdiq44j)* status: 4 unresolved discussions

___
*[`geometry/optimization/convex_set.h` line 99 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kBN4Q6HpiBKG1P86a:-N_kBN4Q6HpiBKG1P86b:b-xhyy25) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/convex_set.h#L99)):*
> ```objc
>   in a zero-dimensional space -- the empty set, and the whole set (which is
>   simply the "zero vector space", {0}.) Zero-dimensional sets are considered
>   to be nonempty by default. Sets which can be zero-dimensional and empty must
> ```

BTW if you have a convenient reference for that convention, consider linking it here. I found your Cartesian product link helpful.
___
*[`geometry/optimization/minkowski_sum.h` line 71 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kDffCD2AFjNZK37oa:-N_kDffCD2AFjNZK37ob:b-fp7mc) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/minkowski_sum.h#L71)):*
> ```objc
>   bool DoIsBounded() const final;
> 
>   /** When we have sets_.size() == 0, we treat the Minkowski sum as being
> ```

nit: don't use Doxygen comments in the private section -- it is misleading since we don't apply Doxygen to private members. You can use `/*` or `//` here. (Also for similar changes elsewhere in this PR)
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 422 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kFTqOBg5E9n07xxd-:-N_kEsx-DRBnsRtqeZxP:b-yvy18y) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/test/cartesian_product_test.cc#L422)):*
> ```cpp
> GTEST_TEST(CartesianProductTest, ZeroDimensionalInputNonempty) {
>   // If any of the inputs are zero dimensional, special logic is used in several
>   // of the methods. This checks them.
> ```

nit: I found this comment confusing. I didn't notice the "Non" in the test name at first so the comment made me think this was the zero-dim test. Consider adding "This test verifies the non-zero dimensional case."
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 472 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kG6RcAew3bVkEaBu-:-N_kG6RcAew3bVkEaBu0:b1t5jn0) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/test/cartesian_product_test.cc#L472)):*
> ```cpp
>   EXPECT_EQ(new_constraints.size(), 2);
>   auto result = solvers::Solve(prog);
>   EXPECT_FALSE(result.is_success());
> ```

minor: not obvious from the code that there is something zero-dimensional for this test and the next one. Please add a comment explaining to the ignorant reader how these two are testing zero-dim behavior. Can you add an `ambient_dimension()==0` check?

sherm1 (commented):
BTW there are conflicts which need manual merging

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_kHuPD6ee6NGJZFEC2:bd5fn4r)* status: 4 unresolved discussions

jwnimmer-tri:
I don't think its a breaking change.  Previously, the ambient-zero case was disallowed with an exception.  Now, we have specific behavior specified instead.

Also, `IsEmpty()` is very new (not yet subject to Stable API governance).

sherm1 (commented):
Thanks, -(release notes: breaking change)  +(release notes: feature)  then.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_kN7Dz3zu3zKqpDr7H:b-rnofeu)* status: 4 unresolved discussions

cohnt (commented):
I went ahead and merged master into this branch to fix the conflicts. When this gets merged in, I assume the commits can just be squashed together?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_klX8W2x1YXDgXwkcc:b-2lkoyw)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @cohnt)

___
*[`geometry/optimization/convex_set.h` line 99 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kBN4Q6HpiBKG1P86a:-N_klmcb3FsMnRXg4ULV:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/convex_set.h#L99)):*

BTW if you have a convenient reference for that convention, consider linking it here. I found your Cartesian product link helpful.
</blockquote></details>

Done.
___
*[`geometry/optimization/convex_set.h` line 124 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_km4TU7VOhAglC1Dy7:-N_km4TU7VOhAglC1Dy8:b-qkz7i9) ([raw file](https://github.com/RobotLocomotion/drake/blob/ecf1b83f78b02ceb09af3329fb61dd9eb724849d/geometry/optimization/convex_set.h#L124)):*
> ```objc
>   /** Returns a feasible point within this convex set if it is nonempty,
>   and nullopt otherwise. */
>   std::optional<Eigen::VectorXd> MaybeGetFeasiblePoint() const {
> ```

@hongkai-dai  note the change in logic here, so that `MaybeGetFeasiblePoint` matches the convention of `MaybeGetPoint`.

@sherm1, this wasn't covered by Hongkai's feature review, so please wait to merge into master until he gives the okay.
___
*[`geometry/optimization/minkowski_sum.h` line 71 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kDffCD2AFjNZK37oa:-N_klpX81f1M7OfMcJl2:b-3q2s9h) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/minkowski_sum.h#L71)):*

nit: don't use Doxygen comments in the private section -- it is misleading since we don't apply Doxygen to private members. You can use `/*` or `//` here. (Also for similar changes elsewhere in this PR)
</blockquote></details>

Done. It's included in the subclass documentation as well, so we're not losing anything.
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 422 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kFTqOBg5E9n07xxd-:-N_klwlN7lP04FrgINpH:b-cxzqcc) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/test/cartesian_product_test.cc#L422)):*

nit: I found this comment confusing. I didn't notice the "Non" in the test name at first so the comment made me think this was the zero-dim test. Consider adding "This test verifies the non-zero dimensional case."
</blockquote></details>

Added some additional description.
___
*[`geometry/optimization/test/cartesian_product_test.cc` line 472 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kG6RcAew3bVkEaBu-:-N_klzAD9izQZdB_QCaK:b-2j1xds) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/test/cartesian_product_test.cc#L472)):*

minor: not obvious from the code that there is something zero-dimensional for this test and the next one. Please add a comment explaining to the ignorant reader how these two are testing zero-dim behavior. Can you add an `ambient_dimension()==0` check?
</blockquote></details>

Added some additional checks on the ambient dimensions and emptiness, that should help clarify what we're checking here.

hongkai-dai (commented):
:lgtm: 

> Reviewed 8 of 16 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_lk38d23IKsqDn5WZe:bg7igkv)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @cohnt)

sherm1 (commented):
Thanks for the clarifications -- very helpful. Yes, we'll squash when merging.

> Reviewed 16 of 16 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-:-N_nuLlAC72kVH1Zzhwn:b-kzgqa3)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @cohnt)

___
*[`geometry/optimization/test/cartesian_product_test.cc` line 472 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19732#-N_kG6RcAew3bVkEaBu-:-N_nu4Ou1dPbbsZMh0hP:b-ayz2sz) ([raw file](https://github.com/RobotLocomotion/drake/blob/180cc44b1726ae5b311df9734e8a0bf6214aaab9/geometry/optimization/test/cartesian_product_test.cc#L472)):*

Added some additional checks on the ambient dimensions and emptiness, that should help clarify what we're checking here.
</blockquote></details>

Very helpful, thanks!

PR #19736, opened 2023-07-06
CollisionChecker validates distance/interpolation funcs with default config

SeanCurtis-TRI:
Previously, when setting a distance (or interpolation) function, `CollisionChecker` would validate it by testing the distance between the zero configuration and itself (or by interpolating between the zero configuration and itself). It is perfectly reasonable that a distance or interpolation function (e.g., one with quaternions) would reject such an invalid configuration as being incalculable. By using the default configuration, we assume we're starting with a valid configuration and the distance/interpolation functions can use their most natural semantics.

The valid test now validates using the plant's default configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)

### DIFF
--- a/planning/collision_checker.cc
+++ b/planning/collision_checker.cc
@@ -517,7 +517,7 @@ void CollisionChecker::SetConfigurationDistanceFunction(
     const ConfigurationDistanceFunction& distance_function) {
   DRAKE_THROW_UNLESS(distance_function != nullptr);
   const double test_distance =
-      distance_function(GetZeroConfiguration(), GetZeroConfiguration());
+      distance_function(GetDefaultConfiguration(), GetDefaultConfiguration());
   DRAKE_THROW_UNLESS(test_distance == 0.0);
   configuration_distance_function_ = distance_function;
 }
@@ -538,12 +538,12 @@ void CollisionChecker::SetConfigurationInterpolationFunction(
     return;
   }
   const Eigen::VectorXd test_interpolated_q = interpolation_function(
-      GetZeroConfiguration(), GetZeroConfiguration(), 0.0);
+      GetDefaultConfiguration(), GetDefaultConfiguration(), 0.0);
   DRAKE_THROW_UNLESS(test_interpolated_q.size() ==
-                     GetZeroConfiguration().size());
+                     GetDefaultConfiguration().size());
   for (int index = 0; index < test_interpolated_q.size(); ++index) {
     DRAKE_THROW_UNLESS(test_interpolated_q(index) ==
-                       GetZeroConfiguration()(index));
+                       GetDefaultConfiguration()(index));
   }
   configuration_interpolation_function_ = interpolation_function;
 }
@@ -803,6 +803,11 @@ CollisionChecker::CollisionChecker(CollisionCheckerParams params,
       supports_parallel_checking_(supports_parallel_checking) {
   // Initialize the zero configuration.
   zero_configuration_ = Eigen::VectorXd::Zero(plant().num_positions());
+  // Initialize the default configuration.
+  auto plant_context = plant().CreateDefaultContext();
+  default_configuration_ =
+      plant().GetPositions(*plant().CreateDefaultContext());
+
   // Initialize the collision padding matrix.
   collision_padding_ =
       Eigen::MatrixXd::Zero(plant().num_bodies(), plant().num_bodies());
--- a/planning/collision_checker.cc
+++ b/planning/collision_checker.cc
@@ -804,7 +804,6 @@ CollisionChecker::CollisionChecker(CollisionCheckerParams params,
   // Initialize the zero configuration.
   zero_configuration_ = Eigen::VectorXd::Zero(plant().num_positions());
   // Initialize the default configuration.
-  auto plant_context = plant().CreateDefaultContext();
   default_configuration_ =
       plant().GetPositions(*plant().CreateDefaultContext());
 
--- a/planning/collision_checker.h
+++ b/planning/collision_checker.h
@@ -1216,6 +1216,12 @@ class CollisionChecker {
   std::string CriticizePaddingMatrix(const Eigen::MatrixXd& padding,
                                      const char* func) const;
 
+  /* @returns a generalized position vector, sized according to the full model,
+   whose values come from the plant's default context. */
+  const Eigen::VectorXd& GetDefaultConfiguration() const {
+    return default_configuration_;
+  }
+
   /* This class allocates and maintains the contexts associated with OpenMP
    threads. When the CollisionChecker is evaluated in its implicit mode, the
    contexts used are drawn from this collection and each context is associated
@@ -1394,6 +1400,10 @@ class CollisionChecker {
 
   /* We maintain a "zero configuration" of the model. */
   Eigen::VectorXd zero_configuration_;
+
+  /* We maintain a "default configuration" of the model (based on the plant's
+   default allocated context). */
+  Eigen::VectorXd default_configuration_;
 
   /* Determines whether the checker reports support for parallel evaluation.
    This is defined upon construction by implementations. */
--- a/planning/collision_checker_params.h
+++ b/planning/collision_checker_params.h
@@ -19,7 +19,8 @@ as a double. The returned distance will be strictly non-negative.
 To be valid, the function must satisfy the following condition:
 
  - dist(q, q) ≡ 0
-*/
+
+for values of q that are valid for the CollisionChecker's plant. */
 using ConfigurationDistanceFunction =
     std::function<double(const Eigen::VectorXd&, const Eigen::VectorXd&)>;
 
@@ -34,7 +35,8 @@ To be valid, the function must satisfy the following conditions:
  - interpolate(q1, q2, 0) ≡ q1
  - interpolate(q1, q2, 1) ≡ q2
  - interpolate(q, q, r) ≡ q, for all r in [0, 1]
-*/
+
+for values of q, q1, and q2 that are valid for the CollisionChecker's plant. */
 using ConfigurationInterpolationFunction = std::function<Eigen::VectorXd(
     const Eigen::VectorXd&, const Eigen::VectorXd&, double)>;
 
--- a/planning/test/collision_checker_test.cc
+++ b/planning/test/collision_checker_test.cc
@@ -610,6 +610,57 @@ TEST_F(TrivialCollisionCheckerTest, ClonesAndContexts) {
             dut_->num_allocated_contexts() + 1 /* prototype context */);
 }
 
+// We want to confirm that when setting distance/interpolation functions, they
+// are validated using *default* configuration and *not* zero configuration.
+// So, we'll create functions that have two properties:
+//
+//   1. They report they've been called.
+//   2. They get angry if they're not provided the default configuration as
+//      parameter values.
+//
+TEST_F(TrivialCollisionCheckerTest, NonZeroDefaultConfiguration) {
+  // We need a plant that has a non-zero default configuration, so we can't
+  // use dut_.
+  auto [robot, robot_index] =
+      MakeModel({.weld_robot = false, .on_env_base = true});
+
+  // Retrieve the default configuration, and cross-check that it's non-zero.
+  const VectorXd default_q =
+      robot->plant().GetPositions(*robot->plant().CreateDefaultContext());
+  ASSERT_TRUE((default_q.array() != 0).any());
+
+  std::unique_ptr<CollisionChecker> dut =
+      MakeUnallocatedChecker(std::move(robot), {robot_index});
+
+  // Create distance function that rejects all configs except the default. This
+  // will prove that the dut only probes the default config when validating the
+  // distance function during SetConfigurationDistanceFunction.
+  bool distance_called = false;
+  auto distance = [&distance_called, &default_q](const VectorXd& q0,
+                                                 const VectorXd& q1) {
+    distance_called = true;
+    EXPECT_EQ(q0, default_q);
+    EXPECT_EQ(q1, default_q);
+    return 0.0;
+  };
+  EXPECT_NO_THROW(dut->SetConfigurationDistanceFunction(distance));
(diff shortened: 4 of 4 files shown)


SeanCurtis-TRI (commented):
+(release notes: fix) 

+@hongkai-dai for feature review (this should resolve your problem)

cc @calderpg-tri 

I'll be running this against anzu CI...stay tuned.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-NZhRu0p9WEiAtSN2HZE:b-axk01k)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

SeanCurtis-TRI (commented):
The build can be found in `Anzu-master-Drake/214`.  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-NZhT1_92zOGgyz5dnO0:b7hcrkh)* status: LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

hongkai-dai (commented):
:lgtm: 

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-NZhVrr2AgrWabYZ1oFy:bg7igkv)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`planning/collision_checker.cc` line 806 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZhVg4v2nuQoXVCFQsK:-NZhVg4v2nuQoXVCFQsL:bggbuvb) ([raw file](https://github.com/RobotLocomotion/drake/blob/61f1d4658b71ffdf02b2629dd985bf343b2a92bf/planning/collision_checker.cc#L806)):*
> ```cpp
>   // Initialize the zero configuration.
>   zero_configuration_ = Eigen::VectorXd::Zero(plant().num_positions());
>   auto plant_context = plant().CreateDefaultContext();
> ```

BTW, document this as "Initialize the default configuration"? (As the line above was documented as "Initialize the zero configuration.").

SeanCurtis-TRI (commented):
As it's essentially end of day Thursday for TRI, +@jwnimmer-tri for Monday platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-NZhkeY-2XZ4nYoS-1xI:b-x50a9d)* status: LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri (commented):
Checkpoint.  I still need to take a more careful look at the unit test.

> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-NZiNZFD3AHfGh0fQUpg:b-832d0z)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`planning/collision_checker.cc` line 810 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZiMsr7EIidSs3y8PWT:-NZiMsr8411UIQiEbARq:bv2kr3f) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/collision_checker.cc#L810)):*
> ```cpp
>   default_configuration_ =
>       plant_context->get_continuous_state_vector().CopyToVector().head(
>           plant().num_positions());
> ```

I'm not sure why `xc` is the right answer here.  What if its a discrete-time plant?



_Suggestion:_
```cpp
  default_configuration_ = plant().GetPositions(*plant().CreateDefaultContext());
```
___
*[`planning/test/collision_checker_test.cc` line 622 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZiNML91vMQURP3CN5c:-NZiNMLA63U19o1FdsNb:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/test/collision_checker_test.cc#L622)):*
> ```cpp
>       MakeModel({.weld_robot = false, .on_env_base = true});
> 
>   // Well get the default and zero configurations.
> ```

typo



_Suggestion:_
```cpp
We'll
```

jwnimmer-tri (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`planning/test/collision_checker_test.cc` line 613 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZqRxFxC-TiM-JA8zB1:-NZqRxFy16qXlu37TaQq:b-6ba1aw) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/test/collision_checker_test.cc#L613)):*
> ```cpp
> }
> 
> // We want to confirm that a distance/interpolation function that can't operate
> ```

I don't buy having such a low-level test for this.  It's too finely tuned for the exact conditions where the bug occurred, thus providing little to no lasting value given the number of lines this test costs and the likelihood of that exact bug being reintroduced.

The test I would buy is using an MbP that actually has a quaternion joint, and the two callback functions use MbP kinematics (or inverse kinematics) to calculate the return values, such that having q==0 would have thrown from inside MbP.

That would detect not only this particular bug, but other kinds bugs that might turn up when handling quaternions inside a CollisionChecker.

If you like, I could write up and push that test case here.

SeanCurtis-TRI (commented):
Updated.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-N_1eqP23gq8kv9fQfJb:b-bw8e14)* status: 2 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`planning/collision_checker.cc` line 810 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZiMsr7EIidSs3y8PWT:-N_1WM6V570jD8IDHY8O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/collision_checker.cc#L810)):*

I'm not sure why `xc` is the right answer here.  What if its a discrete-time plant?
</blockquote></details>

Done.
___
*[`planning/test/collision_checker_test.cc` line 613 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZqRxFxC-TiM-JA8zB1:-N_1WTtR3Kaf7TUGBDha:bhenxal) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/test/collision_checker_test.cc#L613)):*
While I didn't completely buy your assertion/solution, it did cause me to reconsider what I'd presented (and massage the test slightly).

> That would detect not only this particular bug, but other kinds bugs that might turn up when handling quaternions inside a CollisionChecker.

This feels like an integration test. Without testing large swathes of the CollisionChecker API, it's not clear to me that we actually get that benefit. If the only API we exercise is the `Set...Function` APIs, then it has no such power. I think a test along the lines of "Does CollisionChecker handle quaternions correctly?", is an important question to answer. I just don't see this PR as the motivation for it.

I've massaged the test to be more concrete in its scope. We validate candidate functions using the default configuration. Period.

This led me to explore a bit further and I'd offer the following observation:

In `collision_checker_params.h` the definition of a valid distance function is that `dist(q, q) = 0` but it makes no mention of whether `q` is valid or not. A similarly phrased definition is provided for interpolation.

So, based on the documentation, we could simply argue that @hongkai-dai's functions that don't handle arbitrary `q` is *de jure* incorrect. Taken to that level of granularity, it doesn't seem a very helpful definition.

I'd argue the following:

1. The definition of valid functions should pick up a `for valid q` caveat.
2. Then the use of the default configuration is simply an easy way to *sample* the domain of valid `q` to give us a hint that things are good.

jwnimmer-tri (commented):
> Reviewed 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`planning/test/collision_checker_test.cc` line 613 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZqRxFxC-TiM-JA8zB1:-N_4Ki3zACHnNTKzPFdl:b-kf3cez) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/test/collision_checker_test.cc#L613)):*

While I didn't completely buy your assertion/solution, it did cause me to reconsider what I'd presented (and massage the test slightly).

> That would detect not only this particular bug, but other kinds bugs that might turn up when handling quaternions inside a CollisionChecker.

This feels like an integration test. Without testing large swathes of the CollisionChecker API, it's not clear to me that we actually get that benefit. If the only API we exercise is the `Set...Function` APIs, then it has no such power. I think a test along the lines of "Does CollisionChecker handle quaternions correctly?", is an important question to answer. I just don't see this PR as the motivation for it.

I've massaged the test to be more concrete in its scope. We validate candidate functions using the default configuration. Period.

This led me to explore a bit further and I'd offer the following observation:

In `collision_checker_params.h` the definition of a valid distance function is that `dist(q, q) = 0` but it makes no mention of whether `q` is valid or not. A similarly phrased definition is provided for interpolation.

So, based on the documentation, we could simply argue that @hongkai-dai's functions that don't handle arbitrary `q` is *de jure* incorrect. Taken to that level of granularity, it doesn't seem a very helpful definition.

I'd argue the following:

1. The definition of valid functions should pick up a `for valid q` caveat.
2. Then the use of the default configuration is simply an easy way to *sample* the domain of valid `q` to give us a hint that things are good.
</blockquote></details>

I agree, this new version is better.

I still don't buy testing the test code.  It's too much bloat to maintain for too little value.  I've opened a PR that trims down the test: https://github.com/SeanCurtis-TRI/drake/pull/10.  I think the simplified lambdas are correct by inspection, with no need for runtime probing of them.

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-:-N_5wa7M2Zea1BspmEum:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),hongkai-dai

___
*[`planning/test/collision_checker_test.cc` line 613 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-NZqRxFxC-TiM-JA8zB1:-N_5wO1GB80IFqvWIORJ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/1c10aff6d9d6de7ef32457893d50e65da615fd04/planning/test/collision_checker_test.cc#L613)):*

I agree, this new version is better.

I still don't buy testing the test code.  It's too much bloat to maintain for too little value.  I've opened a PR that trims down the test: https://github.com/SeanCurtis-TRI/drake/pull/10.  I think the simplified lambdas are correct by inspection, with no need for runtime probing of them.
</blockquote></details>

Done.

hongkai-dai (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: 1 unresolved discussion

___
*[`planning/collision_checker.cc` line 809 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-N_6G2mH3uPMd6EUOwJM:-N_6G2mH3uPMd6EUOwJN:b-1ad8ao) ([raw file](https://github.com/RobotLocomotion/drake/blob/61e0d4002b54f8acc42d496d1475524b79ec0274/planning/collision_checker.cc#L809)):*
> ```cpp
>   auto plant_context = plant().CreateDefaultContext();
>   default_configuration_ =
>       plant().GetPositions(*plant().CreateDefaultContext());
> ```

Curious why we call `plant().CreateDefaultContext()` in this line again as we have already called it in the line above.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: 1 unresolved discussion

___
*[`planning/collision_checker.cc` line 809 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-N_6G2mH3uPMd6EUOwJM:-N_6GwrzCjBMqMrwG5VB:boqlh1u) ([raw file](https://github.com/RobotLocomotion/drake/blob/61e0d4002b54f8acc42d496d1475524b79ec0274/planning/collision_checker.cc#L809)):*

Curious why we call `plant().CreateDefaultContext()` in this line again as we have already called it in the line above.
</blockquote></details>

Oops.  That part of my patch suggestion got lost.  We should delete the prior line.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19736)* status: 1 unresolved discussion, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`planning/collision_checker.cc` line 809 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19736#-N_6G2mH3uPMd6EUOwJM:-N_6HP5AB33ltCUs-dua:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/61e0d4002b54f8acc42d496d1475524b79ec0274/planning/collision_checker.cc#L809)):*

Oops.  That part of my patch suggestion got lost.  We should delete the prior line.
</blockquote></details>

Done.

PR #19739, opened 2023-07-07
[render] Handle geometry::Mesh("foo.gltf", 1) throughout Drake

### LINKED ISSUE #19532: Improving RenderEngineGltfClient fidelity by consuming glTF files

SeanCurtis-TRI:
New feature summary:

1. RenderEngineGltfClient can render Mesh/Convex which uses a .glTF file (from a previous PR).
2. Meshcat can visualize a Mesh/Convex which uses a .glTF file.
3. Meldis (via DrakeVisualizer) can likewise visualize .gltf files.
4. All other consumers of such meshes either dispatch exceptions or warning messages that the geometry will be ignored.

Implementation details:

If we allow a user to specify a Mesh using a .gltf (instead of an .obj) then every system that can't reify a .gltf file needs to respond.

1. geometry::Mesh and geometry::Convex conveniently report their extension.
2. Updating the following shape reifiers:
   - DrakeVisualizer accepts gltf and warns and ignores unsupported types.
   - Meshcat warns and ignores unsupported geometry types.
     - The two primary unsupported geometry types (STL and GLTF) are in flux.
   - Hydroelastic internal - MakeRigidRepresentation
       - Mesh already throws for unsupported mesh type. - Convex extended to likewise throw for unsupported mesh type.
     - MakeSoftRepresentation - Mesh calls into MakeVolumeMeshFromVtk; we defer to that to know if it has the expected vtk. - Convex calls into MakeConvexVolumeMesh; we defer to that to know if it has the expected obj.
   - RenderEngineGl warns and ignores non-objs.
   - RenderEngineVtk
3. Skipping the following shape reifiers:
   - MeshBuilderForDeformable. It currently only supports vtk files with tetrahedral meshes. Presumably, it already handles the case of a non-vtk file provided (e.g., obj). We'll rely on that logic to handle a mesh with the .gltf extension as well.
   - ProximityEngine. It already has the protocol to throw on any non-obj.
   - ShapeToString. It doesn't care about extensions.
   - deformable_contact_internal (Geometries). This actually defers to hydroleastic_internal (see above).
   - ShapeConfigurations and MakeFclShape (in characterization_utilities.h)
     - They are hard coded to use a box mesh and ignore the filename.
4. Introduce .gltf file in the solar_system example
   - In addition to the underlying .gtlf file referred to in solar_system.cc, this also adds sun.gltf. This enables the following:

          bazel run //tools:model_visualizer -- \
              package://drake/examples/scene_graph/sun.sdf

      as an alternate mechanism for observing meldis support. The solar
      system demo only attaches a `DrakeVisualizer` instance and,
      therefore, requires meldis for visualization.

closes #19532

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739)


SeanCurtis-TRI (commented):
+@jwnimmer-tri for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-NZhu-viDfkL59RLyUaH:b-qf68jg)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
:lgtm: feature.

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-NZqHw0K6_bH9M1myGnd:b-n37teh)* status: 18 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqHTrvDe_iKdcdpt2a):*
What about `geometry::optimization::VPolytope`?  Does `ProximityEngine` already notice and throw before it reaches there?  It wasn't obvious to me that it would.
___
*[`-- commits` line 33 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqESf124QZPkYok4Qv):*
nit The description of what RenderEngineVtk does seems incomplete.
___
*[`examples/scene_graph/BUILD.bazel` line 79 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqFhYhBHayMxDuUANY:-NZqFhYhBHayMxDuUANZ:bsvw5g3) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/examples/scene_graph/BUILD.bazel#L79)):*
> ```Starlark
>         "**/*.urdf",
>         "**/*.xml",
>         "**/*.gltf",
> ```

nit Keep this list alpha-sorted.
___
*[`examples/scene_graph/solar_system.cc` line 162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqG0A95obmMOBpAn1c:-NZqG0A95obmMOBpAn1d:b386jnd) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/examples/scene_graph/solar_system.cc#L162)):*
> ```cpp
>   scene_graph->RegisterAnchoredGeometry(
>       source_id_,
>       MakeShape<Mesh>(RigidTransformd::Identity(), "Sun", Vector4d(1, 1, 0, 1),
> ```

BTW Is the yellow diffuse color here doing anything anymore?  If not, perhaps we should make that more obvious to the reader, by setting it to `nullopt` or at least white?



_Code quote:_
```cpp
Vector4d(1, 1, 0, 1)
```
___
*[`geometry/drake_visualizer.cc` line 469 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqCOuR1CSsY3G7pWn9:-NZqCOuR1CSsY3G7pWnA:b36vzfu) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L469)):*
> ```cpp
>   // Tests the extension to see if the mesh type is unsupported; if so, it
>   // sets the bool* (referenced as data) to false and dispatches a one-time
>   // warning.
> ```

I don't yet understand the reasoning for this design choice (i.e., I disagree with it, pending learning your thought process).

The set of supported mesh types are a function of the receiving program, not this sending program.  Who's to say what program is on the other end of this message?  Checking here will give false positives, and is anyway extra code complexity for no gain.

If we want to have `Meldis` warn when it gets an unsupported Mesh type, I'd buy that.  (I think it probably already does?)
___
*[`geometry/drake_visualizer.cc` line 474 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqAhETAExWGcMWlk_X:-NZqAhETAExWGcMWlk_Y:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L474)):*
> ```cpp
>       static const logging::Warn one_time(
>           "DrakeVisualizer only supports Mesh/Convex specifications which use "
>           ".obj files. Mesh specifications using other mesh types (e.g., "
> ```

typo



_Suggestion:_
```cpp
.obj files or .gltf files
```
___
*[`geometry/drake_visualizer.cc` line 705 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqBrxeCmPoK7W0tNnZ:-NZqBrxeCmPoK7W0tNn_:b-eise56) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L705)):*
> ```cpp
>     // num_geom reflects the *upper bound* -- if geometries get skipped, we'll
>     // decrement it.
>     message.link[0].num_geom = anchored_count;
> ```

nit This line (and its comment) should disappear.  See the discussion immediately below.
___
*[`geometry/drake_visualizer.cc` line 716 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqBL5zCK6IXeUHJ_nd:-NZqBL5zCK6IXeUHJ_ne:bpnm5rn) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L716)):*
> ```cpp
>           message.link[0].geom.push_back(*geo);
>         } else {
>           --message.link[0].num_geom;
> ```

This is too roundabout.  There is a simpler, canonical pattern when filling up an LCM message.  First populate the vectors, then at the end just do this:

```
for (auto& link : message.link) {
  link.num_geom = link.geom.size();
}
```

That covers both the 0'th link and the other links.
___
*[`geometry/drake_visualizer.cc` line 735 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqC-46BUqpgMS31m_Q:-NZqC-46BUqpgMS31m_R:bfpa06y) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L735)):*
> ```cpp
>       std::optional<lcmt_viewer_geometry_data> geo = make_geometry(g_id);
>       if (geo.has_value()) {
>         message.link[link_index].geom.push_back(*geo);
> ```

nit We might as well move it into place?

`std::move(*geo)`

Ditto above.
___
*[`geometry/meshcat.cc` line 336 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqGfJ012NQ5ekVVn11:-NZqGfJ012NQ5ekVVn12:b-nfrq8b) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/meshcat.cc#L336)):*
> ```cpp
>       // Use filename path as the base directory for textures.
>       std::filesystem::path file_path(mesh.filename());
>       const std::filesystem::path basedir = file_path.parent_path();
> ```

BTW The `file_path` isn't actually used afterwards, and thus it doesn't really need to be a mutable named variable?



_Suggestion:_
```cpp
      const fs::path basedir = fs::path(mesh.filename()).parent_path();
```
___
*[`geometry/meshcat.cc` line 426 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqH80Z6us-ePryVpf9:-NZqH80Z6us-ePryVpfA:b-ia8qkj) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/meshcat.cc#L426)):*
> ```cpp
>         static const logging::Warn one_time(
>             "Drake's Meshcat only supports Mesh/Convex specifications which "
>             "use .obj or dae files. Mesh specifications using other mesh types "
> ```

nit The comment is inaccurate; we can handle glTF meshes also.
___
*[`geometry/shape_specification.h` line 208 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq7EGs5dRr_NmHgVby:-NZq7EGuAR46j8SUTmdS:b-srywbu) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/shape_specification.h#L208)):*
> ```objc
>    the dot. In other words /foo/bar/mesh.obj and /foo/bar/mesh.OBJ would both
>    report the ".obj" extension. */
>   std::string extension() const;
> ```

BTW What's the plan for binding these two new functions in pydrake?
___
*[`geometry/shape_specification.cc` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq7NZUAxAm9hOuoIsQ:-NZq7NZUAxAm9hOuoIsR:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/shape_specification.cc#L18)):*
> ```cpp
> namespace {
> 
> std::string GetExtensionLower(const std::string filename) {
> ```

typo



_Suggestion:_
```cpp
const std::string& filename
```
___
*[`geometry/shape_specification.cc` line 97 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq7unbBUWZvfl7bDu6:-NZq7unbBUWZvfl7bDu7:b5wy5xw) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/shape_specification.cc#L97)):*
> ```cpp
> 
> std::string Convex::extension() const {
>   return GetExtensionLower(filename_);
> ```

This is pushing the boundary of a `snake_case` getter, I think past the line. Constructing a `fs::path` immediately splits the whole string on `/` into multiple components, so this function is `O(len(filename_))`.

Possible solutions:

* The extension could be stored as a member field, computed in the constructor.
* 
* If `GetExtensionLower` was just an `rfind` on the raw string (i.e., not using `fs::path`), then I think it would be OK as snake_case.  It's still not strictly constant time, but O(len(result)) is pretty darn close for the typical case of 3-4 letter extensions.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 964 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqD_PF9ArSvUXiOojC:-NZqD_PG12r2m3hflXIw:b732jxx) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/render_gl/internal_render_engine_gl.cc#L964)):*
> ```cpp
>   const fs::path path_in(filename_in);
>   const fs::path file_path =
>       fs::is_symlink(path_in) ? fs::read_symlink(path_in) : path_in;
> ```

BTW I don't grok why we'd check for `is_symlink` only, versus calling `fs::canonical` to also resolve `/path/to/foo/../bar/quux.obj` to `/path/to/bar/quux.obj` (for better caching).  I don't mind leaving this oddity for another day, in case it doesn't seem relevant to this PR.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 967 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqDmVc-P3qEh0kJIE4:-NZqDmVc-P3qEh0kJIE5:b-mrjh7n) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/render_gl/internal_render_engine_gl.cc#L967)):*
> ```cpp
>   const std::string filename = file_path.string();
> 
>   if (Mesh(filename).extension() != ".obj") {
> ```

Why are we checking the extension of the resolved symlink, instead of the original filename?  If the symlink is to a file canonically named `/var/cache/01234567.content-addressible` but the passed-in `filename_in` is `/path/to/foo.obj`, I seems to me like we should still allow it an an OBJ file.
___
*[`geometry/test/drake_visualizer_test.cc` line 631 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqHrVa0Y50-qqzGV7c:-NZqHrVa0Y50-qqzGV7d:b-cuhkyu) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/test/drake_visualizer_test.cc#L631)):*
> ```cpp
> }
> 
> /* Positive confirmation that files with supported extensions are included. */
> ```

Working

If the code change stays intact, I'll need to circle back and read this test code.
___
*[`geometry/test/shape_specification_test.cc` line 700 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq9yYWB0nHBEJauGjJ:-NZq9yYXDjecaGITo4wm:bcrur75) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/test/shape_specification_test.cc#L700)):*
> ```cpp
>   EXPECT_EQ(Mesh("a/b.ObJ").extension(), ".obj");
>   // Arbitrary extensions.
>   EXPECT_EQ(Mesh("a/b.extension").extension(), ".extension");
> ```

BTW I think it's probably worth a test case clarifying that `foobar.quux.obj` has an extension of `.obj` not `.quux.obj`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739)* status: 18 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
I'll wait to assign platform reviewer after I've confirmed I haven't introduced lint.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-N_1UuHS7bFXuepwYywv:b-cxea90)* status: 5 unresolved discussions, needs at least two assigned reviewers

___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqHTrvDe_iKdcdpt2a):*

What about `geometry::optimization::VPolytope`?  Does `ProximityEngine` already notice and throw before it reaches there?  It wasn't obvious to me that it would.
</blockquote></details>

Thanks for catching it. I'd searched for `ImplementGeometry(Mesh)` thinking that would catch `Convex` in the net. But `VPolytope` historically only had `Convex`. I'll do another search to make sure nothing else has slipped through.
___
*[`geometry/shape_specification.h` line 208 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq7EGs5dRr_NmHgVby:-N_1MTWXDQ6ZzJEFVszH:bulga8j) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/shape_specification.h#L208)):*

BTW What's the plan for binding these two new functions in pydrake?
</blockquote></details>

Adding it now.
___
*[`geometry/shape_specification.cc` line 97 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZq7unbBUWZvfl7bDu6:-N_1N9U28Ama-5sFabfm:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/shape_specification.cc#L97)):*

This is pushing the boundary of a `snake_case` getter, I think past the line. Constructing a `fs::path` immediately splits the whole string on `/` into multiple components, so this function is `O(len(filename_))`.

Possible solutions:

* The extension could be stored as a member field, computed in the constructor.
* 
* If `GetExtensionLower` was just an `rfind` on the raw string (i.e., not using `fs::path`), then I think it would be OK as snake_case.  It's still not strictly constant time, but O(len(result)) is pretty darn close for the typical case of 3-4 letter extensions.
</blockquote></details>

Done.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 967 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqDmVc-P3qEh0kJIE4:-N_1TZM9Cu6PwzaoOLyO:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/render_gl/internal_render_engine_gl.cc#L967)):*

Why are we checking the extension of the resolved symlink, instead of the original filename?  If the symlink is to a file canonically named `/var/cache/01234567.content-addressible` but the passed-in `filename_in` is `/path/to/foo.obj`, I seems to me like we should still allow it an an OBJ file.
</blockquote></details>

Done.
___
*[`geometry/drake_visualizer.cc` line 469 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqCOuR1CSsY3G7pWn9:-N_1Jsnb0MZI56sre0vW:bqp6nwx) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L469)):*

I don't yet understand the reasoning for this design choice (i.e., I disagree with it, pending learning your thought process).

The set of supported mesh types are a function of the receiving program, not this sending program.  Who's to say what program is on the other end of this message?  Checking here will give false positives, and is anyway extra code complexity for no gain.

If we want to have `Meldis` warn when it gets an unsupported Mesh type, I'd buy that.  (I think it probably already does?)
</blockquote></details>

The reasoning was the ability to make a definitive statements about `DrakeVisualizer` and the messages it produces rather than about the behavior of the receiver.

However, I can buy the argument that as all `DrakeVisualizer` does is pass along file names, that kind of gate keeping is overkill. I'll remove it.
___
*[`geometry/drake_visualizer.cc` line 716 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqBL5zCK6IXeUHJ_nd:-N_1J__G8r1BxphHFbsd:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/drake_visualizer.cc#L716)):*

This is too roundabout.  There is a simpler, canonical pattern when filling up an LCM message.  First populate the vectors, then at the end just do this:

```
for (auto& link : message.link) {
  link.num_geom = link.geom.size();
}
```

That covers both the 0'th link and the other links.
</blockquote></details>

Done.
___
*[`geometry/test/drake_visualizer_test.cc` line 631 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-NZqHrVa0Y50-qqzGV7c:-N_1Tzpc1BfmiSsZTttM:b-l1kvyg) ([raw file](https://github.com/RobotLocomotion/drake/blob/16ec8d6087d6821b8949e2781d4785b73ac8b671/geometry/test/drake_visualizer_test.cc#L631)):*

Working

If the code change stays intact, I'll need to circle back and read this test code.
</blockquote></details>

Code and test have been reverted.

jwnimmer-tri (commented):
> Reviewed 19 of 19 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739)* status: 4 unresolved discussions, needs at least two assigned reviewers

___
*[`geometry/shape_specification.h` line 208 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1YxBu0l0Ee7XF_wbU:-N_1YxBu0l0Ee7XF_wbV:bxg4jx0) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/shape_specification.h#L208)):*
> ```objc
>    the dot. In other words /foo/bar/mesh.obj and /foo/bar/mesh.OBJ would both
>    report the ".obj" extension. The "extension" portion of the filename is
>    defined as in std::filesystem::path::extension()*/
> ```

nit Period at end of sentence.

Ditto below.
___
*[`geometry/shape_specification.h` line 340 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1Yupv7VpBV782-8uW:-N_1YupwCv9Tlz8YmpsO:b-lmiqg5) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/shape_specification.h#L340)):*
> ```objc
>   // NOTE: Cannot be const to support default copy/move semantics.
>   std::string filename_;
>   std::string extension_;
> ```

Working

Hmm, do we need `reset_after_move` here?  If the filename is >15 characters but extension is very short, will SBO mean they get out of sync, and if even so do we actually care?  Is out-of-sync still a "valid but indeterminate" state?
___
*[`geometry/optimization/vpolytope.h` line 55 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1_mhD0Tt8GeV2V9sa:-N_1_mhD0Tt8GeV2V9sb:bt19yiq) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/optimization/vpolytope.h#L55)):*
> ```objc
>   VPolytope(const Shape& shape,
>             const math::RigidTransformd& shape_pose_in_world,
>             const math::RigidTransformd& reference_frame_pose_in_world);
> ```

Working

Giving myself a moment to ponder what I think about this as a public API, and with these argument names.

Unit testing `GetVertices` on its own should suffice for covering the new code, with no API churn.
___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 301 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1_fhd-OUWFXmzcAwl:-N_1_fhd-OUWFXmzcAwm:b2ni8l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/bindings/pydrake/geometry/test/optimization_test.py#L301)):*
> ```Python
>         # result is implicitly tested in the constructor that consumes a
>         # QueryObject (see below); that constructor delegates to this one.
>         mut.VPolytope(shape=Box(1., 1., 1.),
> ```

nit



_Suggestion:_
```Python
1.0, 1.0, 1.0
```

SeanCurtis-TRI:
*[`geometry/optimization/vpolytope.h` line 55 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1_mhD0Tt8GeV2V9sa:-N_1fFmC2blOHn5g4d5t:b-krtwn3) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/optimization/vpolytope.h#L55)):*

Working

Giving myself a moment to ponder what I think about this as a public API, and with these argument names.

Unit testing `GetVertices` on its own should suffice for covering the new code, with no API churn.

</blockquote></details>

Actually, I just realized I should've gone back to my previous version:

1. This constructor is private.
2. It simply uses `X_WS` and `X_WE` directly as parameters (which is fine for a private API in C++).
3. I remove the extra tests knowing that every public API that relies on this is already tested.

SeanCurtis-TRI:
*[`geometry/shape_specification.h` line 340 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1Yupv7VpBV782-8uW:-N_1fdo51HTFSJbKgcdp:bupi48y) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/shape_specification.h#L340)):*

Working

Hmm, do we need `reset_after_move` here?  If the filename is >15 characters but extension is very short, will SBO mean they get out of sync, and if even so do we actually care?  Is out-of-sync still a "valid but indeterminate" state?
</blockquote></details>

The eventuality you describe is very likely. The path is an absolute path and the extensions will be short (or rejected by consumers).

The consequences of an empty filename having a non-empty extension is that a moved mesh/convex might jump over extension hurdles, but then crash hard when the file is accessed. An empty extension would fall at the first hurdle and print a message indicating that the extension "" isn't acceptable. (Although, that depends on the error site enclosing the extension in quotes).

My gut says enforcing in-sync and handling it probably isn't worth it.

(But thanks for shaking the tree and reminding me to keep this in my thoughts.)

jwnimmer-tri (commented):
+(priority: high) given that other PRs are being blocked on this.

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-N_45K9OAEgTjz6l7A3f:b-wwetsw)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`geometry/shape_specification.h` line 340 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1Yupv7VpBV782-8uW:-N_4JUnW6-gmPHLFcQ8h:bd3ne6w) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/shape_specification.h#L340)):*

The eventuality you describe is very likely. The path is an absolute path and the extensions will be short (or rejected by consumers).

The consequences of an empty filename having a non-empty extension is that a moved mesh/convex might jump over extension hurdles, but then crash hard when the file is accessed. An empty extension would fall at the first hurdle and print a message indicating that the extension "" isn't acceptable. (Although, that depends on the error site enclosing the extension in quotes).

My gut says enforcing in-sync and handling it probably isn't worth it.

(But thanks for shaking the tree and reminding me to keep this in my thoughts.)
</blockquote></details>

Fixed in https://github.com/SeanCurtis-TRI/drake/pull/9.

The libstdc++ implements "moved-from strings are empty".
___
*[`geometry/optimization/vpolytope.h` line 55 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1_mhD0Tt8GeV2V9sa:-N_46R1f-ZsPUAz5UCsX:b-nh6j3m) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/optimization/vpolytope.h#L55)):*

I went with private and monogram names, but did *not* delete the tests.

1. Currently, all of the tests work by constructing a `SceneGraph` and applying `ProximityProperties` to the assigned geometries.
2. Currently, `ProximityEngine` barks loudly for non-obj meshes.
3. At some point in the future, `ProximityEngine` will support non-obj meshes. However, `ProximityEngine`'s support does not necessarily imply that `VPolytope` will know what to do with some other file type.
4. In whatever window exists between `ProximityEngine` no longer barking and `VPolytope` supporting a new file type, `VPolytope` must defend itself.

Hence, the apparent need of a test that shows even if nothing else trips over non-Objs, `VPolytope` will complain appropriately.


</blockquote></details>

Fixed in https://github.com/SeanCurtis-TRI/drake/pull/9.

Whenever practical, we should test using the public API, to match what users would see.  We can pretty easily avoid ProximityEngine interference.

SeanCurtis-TRI (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-N_5yM1a2X4-Mad1TZip:b223nro)* status: 2 unresolved discussions, LGTM missing from assignee xuchenhan-tri(platform)

___
*[`geometry/shape_specification.h` line 340 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1Yupv7VpBV782-8uW:-N_5y7XE8V4zCqLi58Sk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/shape_specification.h#L340)):*

Fixed in https://github.com/SeanCurtis-TRI/drake/pull/9.

The libstdc++ implements "moved-from strings are empty".
</blockquote></details>

Done.
___
*[`geometry/optimization/vpolytope.h` line 55 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_1_mhD0Tt8GeV2V9sa:-N_5y8nf6dcKr7AD1wNw:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/df388fe230d21a911b0b2cce46833dfb478d9145/geometry/optimization/vpolytope.h#L55)):*

Fixed in https://github.com/SeanCurtis-TRI/drake/pull/9.

Whenever practical, we should test using the public API, to match what users would see.  We can pretty easily avoid ProximityEngine interference.
</blockquote></details>

Done.

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-N_5ytjl11V6CYh21cyJ:bnfp4nl)* status: LGTM missing from assignee xuchenhan-tri(platform)

xuchenhan-tri (commented):
:lgtm: platform

> Reviewed 10 of 22 files at r1, 11 of 19 files at r2, 6 of 9 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-:-N_6PRTGAj3PCadOBuJL:bcpz082)* status: 3 unresolved discussions

___
*[`geometry/meshcat.cc` line 407 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_6H5frAn_M-JJEVbam:-N_6H5frAn_M-JJEVban:beoz70p) ([raw file](https://github.com/RobotLocomotion/drake/blob/14e5a14156d239f41d7a119916e933740c74fb6d/geometry/meshcat.cc#L407)):*
> ```cpp
>     } else {
>       // Neither obj with mtllib nor gltf. E.g., a Collada .dae file, .stl, or
>       // geometry-only obj.
> ```

nit, this sentence is a mouthful and I had a hard time understanding what it's trying to say until I read the predicates above.



_Code quote:_
```cpp
      // Neither obj with mtllib nor gltf. E.g., a Collada .dae file, .stl, or
      // geometry-only obj.
```
___
*[`geometry/meshcat.cc` line 411 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_6GOzm4IROMfgaEi20:-N_6GOzn4AInIQHqsgld:bq8gwdb) ([raw file](https://github.com/RobotLocomotion/drake/blob/14e5a14156d239f41d7a119916e933740c74fb6d/geometry/meshcat.cc#L411)):*
> ```cpp
>       // TODO(SeanCurtis-TRI): This doesn't work for STL even though meshcat
>       // supports STL. Meshcat treats STL differently from obj or dae.
>       // https://github.com/rdeits/meshcat/blob/master/src/index.js#L130-L146.
> ```

minor, consider using a permalink so that it ages better.
___
*[`geometry/proximity/hydroelastic_internal.cc` line 276 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19739#-N_6IvB02sqeHMfLStC5:-N_6IvB02sqeHMfLStC6:byp64c7) ([raw file](https://github.com/RobotLocomotion/drake/blob/14e5a14156d239f41d7a119916e933740c74fb6d/geometry/proximity/hydroelastic_internal.cc#L276)):*
> ```cpp
>   if (convex_spec.extension() != ".obj") {
>         "hydroelastic::MakeRigidRepresentation(): unsupported mesh file: {}",
> ```

Btw, the message may be more useful if it states what the (only) supported type is.



_Code quote:_
```cpp
"hydroelastic::MakeRigidRepresentation(): unsupported mesh file: {}",
```

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19739)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),xuchenhan-tri(platform)

PR #19741, opened 2023-07-07
Add BlockSparseCholeskySolver

xuchenhan-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -15,6 +15,7 @@ drake_cc_package_library(
     visibility = ["//visibility:public"],
     deps = [
         ":block_3x3_sparse_matrix",
+        ":block_sparse_cholesky_solver",
         ":block_sparse_lower_triangular_or_symmetric_matrix",
         ":block_sparse_matrix",
         ":conex_supernodal_solver",
@@ -53,6 +54,20 @@ drake_cc_library(
         "//common:default_scalars",
         "//common:essential",
         "//common:unused",
+    ],
+)
+
+drake_cc_library(
+    name = "block_sparse_cholesky_solver",
+    srcs = ["block_sparse_cholesky_solver.cc"],
+    hdrs = ["block_sparse_cholesky_solver.h"],
+    deps = [
+        ":block_sparse_lower_triangular_or_symmetric_matrix",
+        ":minimum_degree_ordering",
+        "//common:copyable_unique_ptr",
+        "//common:essential",
+        "//common:reset_after_move",
+        "//multibody/contact_solvers/sap:partial_permutation",
     ],
 )
 
@@ -239,6 +254,14 @@ drake_cc_googletest(
     name = "block_3x3_sparse_matrix_test",
     deps = [
         ":block_3x3_sparse_matrix",
+        "//common/test_utilities:eigen_matrix_compare",
+    ],
+)
+
+drake_cc_googletest(
+    name = "block_sparse_cholesky_solver_test",
+    deps = [
+        ":block_sparse_cholesky_solver",
         "//common/test_utilities:eigen_matrix_compare",
     ],
 )
--- a/multibody/contact_solvers/block_sparse_cholesky_solver.cc
+++ b/multibody/contact_solvers/block_sparse_cholesky_solver.cc
@@ -1,0 +1,280 @@
+#include "drake/multibody/contact_solvers/block_sparse_cholesky_solver.h"
+
+#include <algorithm>
+#include <utility>
+#include <vector>
+
+#include "drake/multibody/contact_solvers/minimum_degree_ordering.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+template <typename BlockType>
+BlockSparseCholeskySolver<BlockType>::~BlockSparseCholeskySolver() = default;
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::SetMatrix(const SymmetricMatrix& A) {
+  const BlockSparsityPattern& A_block_pattern = A.sparsity_pattern();
+  /* Compute the elimination ordering using Minimum Degree algorithm. */
+  const std::vector<int> elimination_ordering =
+      ComputeMinimumDegreeOrdering(A_block_pattern);
+  BlockSparsityPattern L_block_pattern =
+      SymbolicFactor(A, elimination_ordering);
+  SetMatrixImpl(A, elimination_ordering, std::move(L_block_pattern));
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::UpdateMatrix(
+    const SymmetricMatrix& A) {
+  PermuteAndCopyToL(A);
+  solver_mode_ = SolverMode::kAnalyzed;
+}
+
+template <typename BlockType>
+bool BlockSparseCholeskySolver<BlockType>::Factor() {
+  DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kAnalyzed);
+  for (int j = 0; j < L_->block_cols(); ++j) {
+    /* Update diagonal. */
+    const BlockType& Ajj = L_->diagonal_block(j);
+    L_diag_[j].compute(Ajj);
+    if (L_diag_[j].info() != Eigen::Success) {
+      solver_mode_ = SolverMode::kEmpty;
+      return false;
+    }
+    L_->SetBlockFlat(0, j, L_diag_[j].matrixL());
+    /* Update L₂₁ column.
+     | a₁₁  *  | = | λ₁₁  0 | * | λ₁₁ᵀ L₂₁ᵀ |
+     | a₂₁ a₂₂ |   | L₂₁ L₂₂|   |  0   L₂₂ᵀ |
+     So we have
+      L₂₁λ₁₁ᵀ = a₂₁, and thus
+      λ₁₁L₂₁ᵀ = a₂₁ᵀ */
+    const std::vector<int>& row_blocks = L_->block_row_indices(j);
+    const auto Ljj = L_diag_[j].matrixL();
+    /* We start from flat = 1 here to skip the j,j diagonal entry. */
+    for (int flat = 1; flat < ssize(row_blocks); ++flat) {
+      const BlockType& Aij = L_->block_flat(flat, j);
+      BlockType Lij = Ljj.solve(Aij.transpose()).transpose();
+      L_->SetBlockFlat(flat, j, std::move(Lij));
+    }
+    /* Update L₂₂ according to L₂₂ = a₂₂ - L₂₁⋅L₂₁ᵀ. */
+    RightLookingSymmetricRank1Update(j);
+  }
+  solver_mode_ = SolverMode::kFactored;
+  return true;
+}
+
+template <typename BlockType>
+VectorX<double> BlockSparseCholeskySolver<BlockType>::Solve(
+    const Eigen::Ref<const VectorX<double>>& b) const {
+  VectorX<double> x(b);
+  SolveInPlace(&x);
+  return x;
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::SolveInPlace(
+    VectorX<double>* b) const {
+  DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kFactored);
+  DRAKE_THROW_UNLESS(b != nullptr);
+  DRAKE_THROW_UNLESS(b->size() == L_->cols());
+  VectorX<double> permuted_b(*b);
+  scalar_permutation_.Apply(*b, &permuted_b);
+
+  const BlockSparsityPattern& block_sparsity_pattern = L_->sparsity_pattern();
+  const std::vector<int>& block_sizes = block_sparsity_pattern.block_sizes();
+  const std::vector<int>& starting_cols = L_->starting_cols();
+
+  /* Solve Lz = b in place. */
+  for (int j = 0; j < L_->block_cols(); ++j) {
+    const int block_size = block_sizes[j];
+    const int offset = starting_cols[j];
+    /* Solve for the j-th block entry. */
+    const VectorX<double> bj =
+        L_diag_[j].matrixL().solve(permuted_b.segment(offset, block_size));
+    permuted_b.segment(offset, block_size) = bj;
+    /* Eliminate for the j-th block entry from the system. */
+    const auto& blocks_in_col_j = L_->block_row_indices(j);
+    for (int flat = 1; flat < ssize(blocks_in_col_j); ++flat) {
+      const int i = blocks_in_col_j[flat];
+      permuted_b.segment(starting_cols[i], block_sizes[i]).noalias() -=
+          L_->block_flat(flat, j) * bj;
+    }
+  }
+
+  VectorX<double>& permuted_z = permuted_b;
+  /* Solve Lᵀx = z in place. */
+  for (int j = L_->block_cols() - 1; j >= 0; --j) {
+    /* Eliminate all solved variables. */
+    const auto& blocks_in_col_j = L_->block_row_indices(j);
+    for (int flat = 1; flat < ssize(blocks_in_col_j); ++flat) {
+      const int i = blocks_in_col_j[flat];
+      permuted_z.segment(starting_cols[j], block_sizes[j]).noalias() -=
+          L_->block_flat(flat, j).transpose() *
+          permuted_z.segment(starting_cols[i], block_sizes[i]);
+    }
+    /* Solve for the j-th block entry. */
+    const VectorX<double> zj = L_diag_[j].matrixU().solve(
+        permuted_z.segment(starting_cols[j], block_sizes[j]));
+    permuted_z.segment(starting_cols[j], block_sizes[j]) = zj;
+  }
+  scalar_permutation_.ApplyInverse(permuted_z, b);
+}
+
+template <typename BlockType>
+typename BlockSparseCholeskySolver<BlockType>::LowerTriangularMatrix
+BlockSparseCholeskySolver<BlockType>::L() const {
+  DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kFactored);
+  return *L_;
+}
+
+template <typename BlockType>
+Eigen::PermutationMatrix<Eigen::Dynamic>
+BlockSparseCholeskySolver<BlockType>::CalcPermutationMatrix() const {
+  DRAKE_THROW_UNLESS(solver_mode() != SolverMode::kEmpty);
+  const std::vector<int>& p = scalar_permutation_.permutation();
+  return Eigen::PermutationMatrix<Eigen::Dynamic>(
+      Eigen::Map<const VectorX<int>>(p.data(), p.size()));
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::SetMatrixImpl(
+    const SymmetricMatrix& A, const std::vector<int>& elimination_ordering,
+    BlockSparsityPattern&& L_pattern) {
+  /* First documented responsibility: set `block_permutation_`. */
+  /* Construct the inverse of the elimination ordering, which permutes the
+   original indices to new indices. */
+  std::vector<int> permutation(elimination_ordering.size());
+  for (int i = 0; i < ssize(permutation); ++i) {
+    permutation[elimination_ordering[i]] = i;
+  }
+  block_permutation_ = PartialPermutation(std::move(permutation));
+  /* Second documented responsibility: set `scalar_permutation_`. */
+  SetScalarPermutation(A, elimination_ordering);
+  /* Third documented responsibility: allocate for `L_` and `L_diag_`. */
+  L_ = std::make_unique<LowerTriangularMatrix>(std::move(L_pattern));
+  L_diag_.resize(A.block_cols());
+  /* Fourth documented responsibility: UpdateMatrix. */
+  UpdateMatrix(A);
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::SetScalarPermutation(
+    const SymmetricMatrix& A, const std::vector<int>& elimination_ordering) {
+  /* It's easier to build the scalar elimination ordering first from block
+   elimination ordering and then convert it to the scalar permutation (the
+   inverse of the scalar elimination ordering) that induces the permutation P
+   such that L⋅Lᵀ = P⋅A⋅Pᵀ.
+   More specificially, Pᵢⱼ = 1 for j = scalar_elimination_ordering[i] (or
+   equivalently i = scalar_permutation_[j]) and Pᵢⱼ = 0 otherwise. See
+   CalcPermutationMatrix(). */
+  std::vector<int> scalar_elimination_ordering(A.cols());
+  {
+    const BlockSparsityPattern& A_block_pattern = A.sparsity_pattern();
+    const std::vector<int>& A_block_sizes = A_block_pattern.block_sizes();
+    const std::vector<int>& starting_indices = A.starting_cols();
+    int i_permuted = 0;
+    for (int block_permuted = 0; block_permuted < ssize(elimination_ordering);
+         ++block_permuted) {
+      const int block = elimination_ordering[block_permuted];
+      const int start = starting_indices[block];
+      const int size = A_block_sizes[block];
+      for (int i = start; i < start + size; ++i) {
+        scalar_elimination_ordering[i_permuted++] = i;
+      }
+    }
+  }
+  /* Invert the elimination ordering to get the permutation. */
+  std::vector<int> scalar_permutation(scalar_elimination_ordering.size());
+  for (int i_permuted = 0; i_permuted < ssize(scalar_permutation);
+       ++i_permuted) {
+    scalar_permutation[scalar_elimination_ordering[i_permuted]] = i_permuted;
+  }
+  scalar_permutation_ = PartialPermutation(std::move(scalar_permutation));
+}
+
+template <typename BlockType>
+BlockSparsityPattern BlockSparseCholeskySolver<BlockType>::SymbolicFactor(
+    const SymmetricMatrix& A, const std::vector<int>& elimination_ordering) {
+  /* 1. Compute the block permutation as well as the scalar permutation. */
+  const int n = elimination_ordering.size();
+  /* Construct the inverse of the elimination ordering, which permutes the
+   original indices to new indices. */
+  std::vector<int> permutation(n);
+  for (int i = 0; i < n; ++i) {
+    permutation[elimination_ordering[i]] = i;
+  }
+  const PartialPermutation block_permutation(std::move(permutation));
+
+  /* Find the sparsity pattern of the permuted A (under the permutation induced
+   by the elimination ordering). */
+  const BlockSparsityPattern& A_block_pattern = A.sparsity_pattern();
+  const std::vector<int>& A_block_sizes = A_block_pattern.block_sizes();
+  const std::vector<std::vector<int>>& sparsity_pattern =
+      A_block_pattern.neighbors();
+  std::vector<std::vector<int>> permuted_sparsity_pattern(
+      sparsity_pattern.size());
+  for (int i = 0; i < ssize(sparsity_pattern); ++i) {
+    const int pi = block_permutation.permuted_index(i);
+    for (int j : sparsity_pattern[i]) {
+      const int pj = block_permutation.permuted_index(j);
+      permuted_sparsity_pattern[std::min(pi, pj)].emplace_back(
+          std::max(pi, pj));
+    }
+  }
+  std::vector<int> permuted_block_sizes(A.block_cols());
+  block_permutation.Apply(A_block_sizes, &permuted_block_sizes);
+
+  /* Compute the sparsity pattern of L given the sparsity pattern of A in the
+   new ordering. */
+  return contact_solvers::internal::SymbolicCholeskyFactor(
+      BlockSparsityPattern(permuted_block_sizes, permuted_sparsity_pattern));
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::RightLookingSymmetricRank1Update(
+    int j) {
+  const std::vector<int>& blocks_in_col_j = L_->block_row_indices(j);
+  const int n = blocks_in_col_j.size();
+  /* We start from k = 1 here to skip the j,j diagonal entry. */
+  for (int k = 1; k < n; ++k) {
+    const int col = blocks_in_col_j[k];
+    const BlockType& B = L_->block_flat(k, j);
+    for (int l = k; l < n; ++l) {
+      const int row = blocks_in_col_j[l];
+      const BlockType& A = L_->block_flat(l, j);
+      L_->AddToBlock(row, col, -A * B.transpose());
+    }
+  }
+}
+
+template <typename BlockType>
+void BlockSparseCholeskySolver<BlockType>::PermuteAndCopyToL(
+    const SymmetricMatrix& A) {
+  const int n = A.block_cols();
+  DRAKE_DEMAND(n == block_permutation_.domain_size());
+  DRAKE_DEMAND(n == block_permutation_.permuted_domain_size());
+  L_->SetZero();
+  for (int j = 0; j < n; ++j) {
+    const std::vector<int>& row_indices = A.block_row_indices(j);
+    for (int i : row_indices) {
+      const BlockType& block = A.block(i, j);
+      const int pi = block_permutation_.permuted_index(i);
+      const int pj = block_permutation_.permuted_index(j);
+      if (pi >= pj) {
+        L_->SetBlock(pi, pj, block);
+      } else {
+        L_->SetBlock(pj, pi, block.transpose());
+      }
+    }
+  }
+}
+
+template class BlockSparseCholeskySolver<MatrixX<double>>;
+template class BlockSparseCholeskySolver<Matrix3<double>>;
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
--- a/multibody/contact_solvers/block_sparse_cholesky_solver.h
+++ b/multibody/contact_solvers/block_sparse_cholesky_solver.h
@@ -1,0 +1,220 @@
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include "drake/common/copyable_unique_ptr.h"
+#include "drake/common/drake_copyable.h"
+#include "drake/common/eigen_types.h"
+#include "drake/common/reset_after_move.h"
+#include "drake/multibody/contact_solvers/block_sparse_lower_triangular_or_symmetric_matrix.h"
+#include "drake/multibody/contact_solvers/sap/partial_permutation.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+/* A Cholesky solver for solving the symmetric positive definite
+ system
+   A⋅x = b
+ where A is block sparse.
+ Example use case:
+ ```
+  BlockSparseCholeskySolver<MatrixX<double>> solver;
+  // Sets the matrix A.
+  solver.SetMatrix(A);
+  // Factorize the matrix.
+  solver.Factor();
+  // Solve A⋅x1 = b1.
+  x1 = solver.Solve(b1);
+  // Solve A⋅x2 = b2. This reuses the factorization (important for speed!).
+  x2 = solver.Solve(b2);
+  // Update the numerical values in the matrix but the sparsity pattern doesn't
+  // change.
+  solver.UpdateMatrix(A2);
+  // Need to refactor after the matrix has changed.
+  solver.Factor();
+  // Solve A2⋅x = b using updated factorization.
+  x = solver.Solve(b);
+ ```
+
+ @tparam BlockType The matrix type for individual block matrices;
+ MatrixX<double> or Matrix3<double>. The fixed-size matrix version is preferred
+ if you know the sizes of blocks are uniform and fixed. */
+template <typename BlockType>
+class BlockSparseCholeskySolver {
+ public:
+  DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(BlockSparseCholeskySolver);
+
+  /* The state of the solver. */
+  enum class SolverMode {
+    kEmpty,     // Matrix A is yet to be set.
+    kAnalyzed,  // Matrix A is symbolically analyzed and ready to be factored,
+                // but not yet numerically factored.
+    kFactored,  // Matrix A is factored in place into L, its lower triangular
+                // Cholesky factorization.
+  };
+
+  using SymmetricMatrix =
+      BlockSparseLowerTriangularOrSymmetricMatrix<BlockType, true>;
+  using LowerTriangularMatrix =
+      BlockSparseLowerTriangularOrSymmetricMatrix<BlockType, false>;
+
+  /* Constructs a BlockSparseCholeskySolver. */
+  BlockSparseCholeskySolver() = default;
+
+  ~BlockSparseCholeskySolver();
+
+  /* Sets the matrix to be factored and analyzes its sparsity pattern to find an
+   efficient elimination ordering. Factor() may be called after call to
+   SetMatrix(). SetMatrix() can be called repeatedly in a row, but prefer the
+   following pattern when the sparsity patterns don't change.
+
+     solver.SetMatrix(A0);
+     ...
+     solver.UpdateMatrix(A1);
+     ...
+     solver.UpdateMatrix(A2);
+     ...
+     solver.UpdateMatrix(A3);
+     ...
+   See UpdateMatrix().
+
+   @pre A is positive definite.
+   @post solver_mode() == SolverMode::kAnalyzed. */
+  void SetMatrix(const SymmetricMatrix& A);
+
+  /* Updates the matrix to be factored. This is useful for solving a series of
+   matrices with the same sparsity pattern using the same elimination ordering.
+   For example, with matrices A and B with the same sparisty pattern. It's more
+   efficient to call
+     solver.SetMatrix(A);
+     solver.Factor();
+     solver.Solve(...)
+     solver.UpdateMatrix(B);
+     solver.Factor();
+     solver.Solve(...)
+   than to call
+     solver.SetMatrix(A);
+     solver.Factor();
+     solver.Solve(...)
+     solver.SetMatrix(B);
+     solver.Factor();
+     solver.Solve(...)
+   Factor() may be called after call to UpdateMatrix().
+   @pre SetMatrix() has been invoked and the argument to the last call of
+   SetMatrix() has the same sparsity pattern of A.
+   @pre A is positive definite.
+   @post solver_mode() == SolverMode::kAnalyzed. */
+  void UpdateMatrix(const SymmetricMatrix& A);
+
+  /* Computes the block sparse Cholesty factorization. Returns true if
+   factorization succeeds, otherwise returns false. Failure is triggered by an
+   internal failure of Eigen::LLT.  This can fail if, for instance, the input
+   matrix set in SetMatrix() or UpdateMatrix() is not positive definite. If
+   failure is encountered, the user should verify that the specified matrix is
+   positive definite and not poorly conditioned.
+   @throws std::exception if solver_mode() is not SolverMode::kAnalyzed.
+   @post solver_mode() is SolverMode::kFactored if factorization is successful
+   and is SolverMode::kEmpty otherwise. */
+  [[nodiscard]] bool Factor();
+
+  /* Solves the system A⋅x = b and returns x.
+   @throws std::exception if b.size() is incompatible with the size of the
+   matrix set by SetMatrix().
+   @throws std::exception unless solver_mode() == SolverMode::kFactored. */
+  VectorX<double> Solve(const Eigen::Ref<const VectorX<double>>& b) const;
+
+  /* Solves the system A⋅x = b and writes the result in b.
+   @throws std::exception if b == nullptr or b->size() is incompatible with the
+   size of the matrix set by SetMatrix().
+   @throws std::exception unless solver_mode() == SolverMode::kFactored. */
+  void SolveInPlace(VectorX<double>* b) const;
+
+  /* Returns the current mode of the solver. See SolverMode. */
+  SolverMode solver_mode() const { return solver_mode_; }
+
+  /* Returns (the lower triangular) Cholesky factorization matrix L as a
+   dense matrix. L is defined by L⋅Lᵀ = P⋅A⋅Pᵀ, where A is the matrix set via
+   SetMatrix() or UpdateMatrix() and P is the permutation matrix induced by the
+   elimination ordering (see CalcPermutationMatrix()).
+   @throws std::exception unless solver_mode() == SolverMode::kFactored. */
+  LowerTriangularMatrix L() const;
+
+  /* Returns the permutation matrix induced by the elimination ordering.
+   The permutation matrix is defined through L⋅Lᵀ = P⋅A⋅Pᵀ, where A is the
+   matrix set via SetMatrix() or UpdateMatrix() and L is the lower triangular
+   Cholesky factorization of the permuted A.
+   @throws std::exception if solver_mode() == SolverMode::kEmpty. */
+  Eigen::PermutationMatrix<Eigen::Dynamic> CalcPermutationMatrix() const;
+
+ private:
+  /* Helper for SetMatrix() to set the matrix given its elimination ordering and
+   the sparsity pattern of its Cholesky factorization. It performs the
+   following:
+    1. sets `block_permutation_`;
+    2. sets `scalar_permutation_`;
+    3. allocates for `L_` and `L_diag_`;
+    4. calls UpdateMatrix(A) to copy the numeric values of A to L_.
+   @param[in] A                     The matrix to be factored.
+   @param[in] elimination_ordering  Elimination ordering of the blocks of A.
+                                    Must be a permutation of {0, 1, ...,
+                                    A.block_cols() - 1}.
+   @param[in] L_pattern             The block sparsity pattern of the lower
+                                    triangular matrix L if A is to be factored
+                                    with the given elimination ordering.  */
+  void SetMatrixImpl(const SymmetricMatrix& A,
+                     const std::vector<int>& elimination_ordering,
+                     BlockSparsityPattern&& L_pattern);
+
+  /* Sets `scalar_permutation_` given the matrix A and the prescribed
+   elimination ordering.
+   @param[in] A                     The matrix to be factored.
+   @param[in] elimination_ordering  Elimination ordering of the blocks of A.
+                                    Must be a permutation of {0, 1, ...,
+                                    A.block_cols() - 1}. */
+  void SetScalarPermutation(const SymmetricMatrix& A,
+                            const std::vector<int>& elimination_ordering);
+
+  /* Returns the block sparsity pattern of the L matrix from a block sparse
+   Cholesky factorization following the prescribed elimination ordering.
+   @param[in] A                     The matrix to be factored.
+   @param[in] elimination_ordering  Elimination ordering of the blocks of A.
+                                    Must be a permutation of {0, 1, ...,
+                                    A.block_cols() - 1}. */
+  BlockSparsityPattern SymbolicFactor(
+      const SymmetricMatrix& A, const std::vector<int>& elimination_ordering);
+
+  /* Performs L(j+1:, j+1:) -= L(j+1:, j) * L(j+1:, j).transpose().
+   @pre 0 <= j < L.block_cols(). */
+  void RightLookingSymmetricRank1Update(int j);
+
+  /* Permutes the given matrix A with `block_permutation_` p and set L such that
+   the lower triangular part of L satisfies L(p(i), p(j)) = A(i, j).
+   @pre SetMarix() has been called. */
+  void PermuteAndCopyToL(const SymmetricMatrix& A);
+
+  /* The cholesky factorization of the permuted matrix, i.e. L⋅Lᵀ = P⋅A⋅Pᵀ,
+   where P is the permutation matrix induced by the `scalar_permutation_`. */
+  copyable_unique_ptr<LowerTriangularMatrix> L_;
+  std::vector<Eigen::LLT<BlockType>> L_diag_;
+
+  /* Block and scalar representations of the permutation matrix P (see
+   CalcPermutationMatrix()). */
+  /* Permutation for block indices, same size as A.block_cols(). For a given
+   block index i into A, `block_permutation_[i]` gives the permuted block index
+   into L_. */
+  PartialPermutation block_permutation_;
+  /* Permutation for scarlar indices, same size as A.cols(). For a given
+   scalar index i into A, `scalar_permutation_[i]` gives the permuted scalar
+   index into L_. */
+  PartialPermutation scalar_permutation_;
+
+  reset_after_move<SolverMode> solver_mode_{SolverMode::kEmpty};
+};
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
--- a/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc
+++ b/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc
@@ -1,0 +1,212 @@
+#include "drake/multibody/contact_solvers/block_sparse_cholesky_solver.h"
+
+#include <memory>
+#include <utility>
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/unused.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+namespace {
+
+using Eigen::Matrix4d;
+using Eigen::MatrixXd;
+using Eigen::MatrixXi;
+using Eigen::VectorXd;
+
+/* Makes an arbitrary SPD matrix with the following nonzero pattern.
+
+  X X | O O O | O O O O | O O O
+  X X | O O O | O O O O | O O O
+  --- | ----- |---------| -----
+  O O | X X X | X X X X | X X X
+  O O | X X X | X X X X | X X X
+  O O | X X X | X X X X | X X X
+  --- | ----- |---------| -----
+  O O | X X X | X X X X | O O O
+  O O | X X X | X X X X | O O O
+  O O | X X X | X X X X | O O O
+  O O | X X X | X X X X | O O O
+  --- | ----- |---------| -----
+  O O | X X X | O O O O | X X X
+  O O | X X X | O O O O | X X X
+  O O | X X X | O O O O | X X X
+
+ The scaling factor can be used to control the values of each nonzero entry. */
+Eigen::Matrix<double, 12, 12> MakeSpdMatrix(double scale) {
+  Eigen::Matrix<double, 12, 12> A;
+  for (int i = 0; i < 12; ++i) {
+    for (int j = 0; j < 12; ++j) {
+      A(i, j) = 0.03 * i + 0.02 * j;
+    }
+  }
+  Eigen::Matrix<double, 12, 12> I = Eigen::Matrix<double, 12, 12>::Identity();
+  /* Add a large diagonal entry to ensure diagonal dominance. */
+  Eigen::Matrix<double, 12, 12> result = scale * (A * A.transpose() + 12 * I);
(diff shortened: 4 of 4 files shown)


xuchenhan-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-NZilbSmBDUJiaLHNCQW:b-57wv7n)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

amcastro-tri (commented):
Excellent! mostly asking for comments in the docs.
:lgtm_strong: 

+@jwnimmer-tri for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_-Pa8g-YpMwfnFQO1z:b-xa3785)* status: 23 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzm6ab2pX-0Z86hXQB:-NZzm6ab2pX-0Z86hXQC:bpjwont) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L53)):*
> ```objc
>       BlockSparseLowerTriangularOrSymmetricMatrix<BlockType, true>;
> 
>   /* Constructs a BlockSparseCholeskySovler. */
> ```

minor, typo "Sovler"
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzmZ6R-hseuquckJHq:-NZzmZ6R-hseuquckJHr:b-1h7kvs) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L59)):*
> ```objc
>    efficient elimination ordering. Factor() may be called after call to
>    SetMatrix().
>    @pre A is positive definite.
> ```

nit, can I call this twice in a row? 
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzmrWtCjAlaRiJ9sZu:-NZzmrWtCjAlaRiJ9sZv:b-pyvewu) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L65)):*
> ```objc
>   /* Updates the matrix to be factored. This is useful for solving a series of
>    matrices with the same sparsity pattern using the same elimination ordering.
>    For example, with matrices A, B, and C with the same sparisty pattern. It's
> ```

minor, ".. with matrices A and B .." (matrix C was not used)
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzn6ff3qInJwyF-ZpB:-NZzn6ff3qInJwyF-ZpC:b-2z29y7) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L83)):*
> ```objc
>    @pre A has the same elimination ordering as the input of the last invocation
>    of SetMatrix(). UpdateMatrix() must not be called before any invocation of
>    SetMatrix().
> ```

nit, maybe a separate `@pre matrix_set() is true` for this one?
You might even consider throwing since it's easy to check.



_Suggestion:_
```objc
   UpdateMatrix() must not be called before any invocation of
   SetMatrix().
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZznxnFCs3stZsQ6Xku:-NZznxnFCs3stZsQ6Xkv:b-1v9ynv) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L95)):*
> ```objc
>    positive definite and not poorly conditioned.
>    @throws std::exception if matrix_set() returns false.
>    @post matrix_set() == false and is_factored() == true. */
> ```

maybe just semantics, but I find it odd that a call to this method sets `matrix_set() = false`? 
since `is_factored() = true` you already know the solver now is in "factored" state. 
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzniAS7CWqTngEk_Eb:-NZzniAS7CWqTngEk_Ec:b-9b7wgg) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*
> ```objc
>    @throws std::exception if matrix_set() returns false.
>    @post matrix_set() == false and is_factored() == true. */
>   bool Factor();
> ```

can `is_factored()` be `true` when I call this method? that is, can I call `Factor()` twice in a row?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzoKeP0SagdScofZ6O:-NZzoKeP0SagdScofZ6P:b-lzvi4p) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*
> ```objc
>    @throws std::exception if matrix_set() returns false.
>    @post matrix_set() == false and is_factored() == true. */
>   bool Factor();
> ```

I'd imagine that if the return is `false`, then the post condition is `is_factored() = false`. 
You might want to document that.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 105 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzo_M24hGZibc9p-SP:-NZzo_M24hGZibc9p-SQ:b-4b3kcc) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L105)):*
> ```objc
>   /* Solves the system A⋅x = b and writes the result in b.
>    @pre y != nullptr and y->size() is compatible with the size of the matrix set
>    by SetMatrix().
> ```

the mix of `@pre` and `@throws` surprises me a bit. I usually use `@pre` for difficult to verify numeric things that require to perform an expensive computation. 
For easy things (nullptr, size), why not to just throw? Otherwise, what criteria do you use to choose one over the other?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzp2r02b2SRfELNkd9:-NZzp2r02b2SRfELNkdA:b-scgtf9) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L115)):*
> ```objc
> 
>   /* Returns true iff the matrix has been factorized (via Factor() or
>    CalcSchurComplementAndFactor()) and is ready for back solve. */
> ```

fyi, this might go out of date if we decide we don't like the method's name in the next PR.
I think you can leave the note in parentheses out just fine here, and above in `matrix_set()`. It's clear without the note already.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 167 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzr6XX2RDnYEFLTUa3:-NZzr6XX2RDnYEFLTUa4:b-xpdg4x) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L167)):*
> ```objc
>   void PermuteAndCopyToL(const SymmetricMatrix& A);
> 
>   copyable_unique_ptr<LowerTriangularMatrix> L_;
> ```

I wonder how L relates to the permutation below? would it be correct to say this is doing `L⋅Lᵀ = P⋅A⋅Pᵀ`? with P the permutation below.
Maybe even worth returning L, and P for additional unit testing of this identity, or whichever variant you are using.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 168 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-LSGk-IocdGhLSGss:-N_-LSGl4O0YSSqhpHUh:bouagu1) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L168)):*
> ```objc
> 
>   copyable_unique_ptr<LowerTriangularMatrix> L_;
>   std::vector<Eigen::LLT<BlockType>> L_diag_;
> ```

question, wondering if you considered implementing `L⋅D⋅Lᵀ` instead
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 171 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzsuC822WH3hM870s4:-NZzsuC822WH3hM870s5:b-dwo45a) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L171)):*
> ```objc
>   /* The mapping from the internal indices (i.e, the indices for L_) to
>    the indices of the original matrix supplied in SetMatrix(). */
>   PartialPermutation block_permutation_;  // permutation for block indices, same
> ```

for the `i-th` block of A, is `block_permutation_[i]` the permuted index? or is it the other way around, ditto below.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 173 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzq2s-EMPqNPdN6obo:-NZzq2s-EMPqNPdN6obp:b4bmho1) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L173)):*
> ```objc
>   PartialPermutation block_permutation_;  // permutation for block indices, same
>                                           // size as A.block_cols().
>   PartialPermutation scalar_permutation_;  // permutation for scalar indices,
> ```

are these two `foo_permutation_` a different representation of the same permutation matrix `P`? maybe add a comment on this.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-NP5A5wxIQ1U2bmKP:-N_-NP5A5wxIQ1U2bmKQ:b77gsv5) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L51)):*
> ```cpp
>     diagonal block of L_. Instead, we directly invoke L_diag_. */
> 
>     /* Update column.
> ```

nit, consider



_Suggestion:_
```cpp
 Update L₂₁ column.
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-NVXV0UE0gTk9oWJB:-N_-NVXV0UE0gTk9oWJC:b-nl7161) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L65)):*
> ```cpp
>       L_->SetBlockFlat(flat, j, std::move(Lij));
>     }
>     RightLookingSymmetricRank1Update(j);
> ```

nit, consider adding comment, update L₂₂ according to `L₂₂ = a₂₂ - L₂₁⋅L₂₁ᵀ`
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-Np_m85UnxzHZwTKM:-N_-Np_m85UnxzHZwTKN:b8l4h3d) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L84)):*
> ```cpp
>   DRAKE_THROW_UNLESS(is_factored_);
>   DRAKE_DEMAND(y != nullptr);
>   DRAKE_DEMAND(y->size() == L_->cols());
> ```

nit, consider THROW here as well.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 155 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZztiTsA760qYryhq5a:-NZztiTsA760qYryhq5b:b-nhn7ok) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L155)):*
> ```cpp
>    elimination ordering and then convert it to the scalar permutation (the
>    inverse of the scalar elimination ordering). The scalar permutation maps the
>    right hand side of A * x = b to the right handside of
> ```

minor,



_Suggestion:_
```cpp
right-hand side 
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 155 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzumRSBr1MY_5aansa:-NZztdCX1g6hK_EE8ory:b-gs663b) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L155)):*
> ```cpp
>    elimination ordering and then convert it to the scalar permutation (the
>    inverse of the scalar elimination ordering). The scalar permutation maps the
>    right hand side of A * x = b to the right handside of
> ```

I wonder if easier to write in terms of a permutation matrix.
Lets say `p` is the scalar permutation and we define matrix P such that `Pᵢⱼ = 1` if `i = pⱼ` and zero otherwise.
Then could we write `L⋅Lᵀ=P⋅A⋅Pᵀ`? or equivalently, L⋅Lᵀ ⋅(P⋅x) = P⋅b?

I'm only proposing the notation, I might have gotten the transpose incorrectly.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-8siHDbz64wzaXM-h:-N_-8siHDbz64wzaXM-i:b-wl92n6) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L162)):*
> ```cpp
>   const std::vector<int>& starting_indices = A.starting_cols();
>   int scalar_order = 0;
>   for (int i = 0; i < ssize(elimination_ordering); ++i) {
> ```

it'd be nice to use a consistent notation all throughout to know when we are working with an original index or its permutation.
Maybe `i`, vs `i_permuted`? similarly `block` and `block_permuted`?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 244 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-I0Xi-6x-XHAfzsRi:-N_-I0Xi-6x-XHAfzsRj:be8768y) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L244)):*
> ```cpp
>     for (int i : row_indices) {
>       const BlockType& block = A.block(i, j);
>       const int pi = block_permutation_.push(i);
> ```

I believe your intention here is not to update `block_permutation_`. 
Therefore call `permuted_index()` instead?
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZziOr763KDZyAw9ssD:-NZziOr763KDZyAw9ssE:b-z5mbrn) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L48)):*
> ```cpp
>   Eigen::Matrix<double, 12, 12> I = Eigen::Matrix<double, 12, 12>::Identity();
>   /* Now result is guaranteed to be SPD. */
>   Eigen::Matrix<double, 12, 12> result = scale * (A * A.transpose() + 12 * I);
> ```

nit,I haven't done the math, but below you require diagonal dominance, so maybe (if true) update comment to:  "Now the result is diagonally dominant"

Diagonal dominant implies SPD, but not the other way around.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzi2yoAXu7pmn6-ULQ:-NZzi2yoAXu7pmn6-ULR:ba49nyn) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L50)):*
> ```cpp
>   Eigen::Matrix<double, 12, 12> result = scale * (A * A.transpose() + 12 * I);
>   /* Zeroing off-diagonal entries of a diagnally dominant matrix doesn't affect
>    the SPDness of the matrix. */
> ```

nit, 
Zeroing off-diagonal entries of a diagonally dominant matrix does not affect diagonal dominance, which in turns implies SPDness.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzkSkl88tLjRa5DvPJ:-NZzkSkm4prhKn3lSo1E:b-wz0j5j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L90)):*
> ```cpp
>   const VectorXd b1 = VectorXd::LinSpaced(A.cols(), 0.0, 1.0);
>   const VectorXd x1 = solver.Solve(b1);
>   const VectorXd expected_x1 = Eigen::LLT<MatrixXd>(dense_A).solve(b1);
> ```

minor, fyi, you can also do `dense_A.llt().solve(b1)` for a shorter spelling.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 110 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzl-2HD0sFqdu2VBg0:-NZzl-2HD0sFqdu2VBg1:bg9aisx) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L110)):*
> ```cpp
>   BlockSparseSymmetricMatrix A2 = MakeSparseSpdMatrix(10);
>   MatrixX<double> dense_A2 = A2.MakeDenseMatrix();
>   solver.UpdateMatrix(A2);
> ```

what happens if A2 does not have the same sparsity pattern?

jwnimmer-tri (commented):
(There are enough questions on the header file that I'm going to pause my review until after the next push.)

> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_-fewz3lMY6Eqb34tT:bhsvbf5)* status: 24 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 4 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-engs8wWM0MXa0VGU:-N_-engs8wWM0MXa0VGV:b-qrpv79) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L4)):*
> ```objc
> 
> #include <memory>
> #include <unordered_set>
> ```

nit Unused `unordered_set`

xuchenhan-tri (commented):
Thanks for the quick turnaround! Comments addressed, PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_0VsPi8BvIcGOwYtX5:b-3aeokw)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 4 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-engs8wWM0MXa0VGU:-N_-gigP7VrQI96mmqe2:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L4)):*

nit Unused `unordered_set`
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 59 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzmZ6R-hseuquckJHq:-N_-hP9U8NdsjafYG6UO:b-g4rral) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L59)):*

nit, can I call this twice in a row? 
</blockquote></details>

Done.

Yes, but it's more efficient to call UpdateMatrix if the sparsity pattern stays unchanged.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzmrWtCjAlaRiJ9sZu:-N_-hguB9CobsUVZJm1E:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L65)):*

minor, ".. with matrices A and B .." (matrix C was not used)
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzn6ff3qInJwyF-ZpB:-N_-iF8D7sthUyxY2I7V:bbsabm0) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L83)):*
> maybe a separate `@pre matrix_set() is true` for this one?

matrix_set() == true is not a prerequisite for this function. In fact, in the example workflow

```
     solver.SetMatrix(A);
     solver.Factor();
     solver.Solve(...)
     solver.UpdateMatrix(B);
     solver.Factor();
     solver.Solve(...)
```
`matrix_set() == false` right before `solver.UpdateMatrix(B)`.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZznxnFCs3stZsQ6Xku:-N_-iHpD-CkoripD5c1T:b-m7p5pl) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L95)):*

maybe just semantics, but I find it odd that a call to this method sets `matrix_set() = false`? 
since `is_factored() = true` you already know the solver now is in "factored" state. 
</blockquote></details>

The matrix is factored in place; the original matrix is gone, so the matrix is "not set".
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzniAS7CWqTngEk_Eb:-N_-iTvZBOgLjkX7sD5P:b-t4ddic) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*
No, Factored cannot be called twice in a row because of the combination of
>@throws std::exception if matrix_set() returns false.
 @post matrix_set() == false ....
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzoKeP0SagdScofZ6O:-N_-z9Of7u6qE-imHYYi:b-qu6n94) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*

I'd imagine that if the return is `false`, then the post condition is `is_factored() = false`. 
You might want to document that.
</blockquote></details>

Good point. Done.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 105 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzo_M24hGZibc9p-SP:-N_-zP8H7xxJYoqZCvwN:b-t3q4hk) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L105)):*

the mix of `@pre` and `@throws` surprises me a bit. I usually use `@pre` for difficult to verify numeric things that require to perform an expensive computation. 
For easy things (nullptr, size), why not to just throw? Otherwise, what criteria do you use to choose one over the other?
</blockquote></details>

Done. Changed to throw.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 115 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzp2r02b2SRfELNkd9:-N_-zl-QBX41e4Ti8rOk:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L115)):*

fyi, this might go out of date if we decide we don't like the method's name in the next PR.
I think you can leave the note in parentheses out just fine here, and above in `matrix_set()`. It's clear without the note already.

</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 167 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzr6XX2RDnYEFLTUa3:-N_07iUU2nY_hFUlmP7o:b-da19uo) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L167)):*
> would it be correct to say this is doing `L⋅Lᵀ = P⋅A⋅Pᵀ`? with P the permutation below.

Yes, this is correct. There is one caveat: I was cutting corner in r1 by not storing the diagonal blocks of L_ because we never touch it. I have since added it in r2 to perform the unit test you proposed.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 168 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-LSGk-IocdGhLSGss:-N_08CgP7cNl9U3cdFha:btog8az) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L168)):*

question, wondering if you considered implementing `L⋅D⋅Lᵀ` instead
</blockquote></details>

Eigen::LLT is faster than Eigen::LDLT so LLT is chosen over LDLT.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 171 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzsuC822WH3hM870s4:-N_0UCyB58gUeFWyVtEc:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L171)):*

for the `i-th` block of A, is `block_permutation_[i]` the permuted index? or is it the other way around, ditto below.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 173 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzq2s-EMPqNPdN6obo:-N_0UGReAZRl9s0iO7Fi:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L173)):*

are these two `foo_permutation_` a different representation of the same permutation matrix `P`? maybe add a comment on this.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-NP5A5wxIQ1U2bmKP:-N_08eoi5M-s7R-o9kEB:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L51)):*

nit, consider
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-NVXV0UE0gTk9oWJB:-N_08muq4yfn6A_N5_iS:b-dtkdrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L65)):*

nit, consider adding comment, update L₂₂ according to `L₂₂ = a₂₂ - L₂₁⋅L₂₁ᵀ`
</blockquote></details>

Done, thanks.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 84 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-Np_m85UnxzHZwTKM:-N_08oEkDZ5_PCCcUls_:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L84)):*

nit, consider THROW here as well.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 155 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZztiTsA760qYryhq5a:-N_08skm6K0CobjCmPZ3:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L155)):*

minor,
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 155 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzumRSBr1MY_5aansa:-N_09Mt_3MuNuBXP86vj:b-gk5vq9) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L155)):*

I wonder if easier to write in terms of a permutation matrix.
Lets say `p` is the scalar permutation and we define matrix P such that `Pᵢⱼ = 1` if `i = pⱼ` and zero otherwise.
Then could we write `L⋅Lᵀ=P⋅A⋅Pᵀ`? or equivalently, L⋅Lᵀ ⋅(P⋅x) = P⋅b?

I'm only proposing the notation, I might have gotten the transpose incorrectly.
</blockquote></details>

Done, I think?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-8siHDbz64wzaXM-h:-N_0BM266yzoLpfz5D0B:b-gc9xfi) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L162)):*

it'd be nice to use a consistent notation all throughout to know when we are working with an original index or its permutation.
Maybe `i`, vs `i_permuted`? similarly `block` and `block_permuted`?
</blockquote></details>

The proposed notation doesn't help me in particular, but I'm neutral about the change. I tried it out in the new revision. See if you like it.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 244 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-I0Xi-6x-XHAfzsRi:-N_0Bk7p4h3J28OLNIJy:b-iry7x8) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L244)):*

I believe your intention here is not to update `block_permutation_`. 
Therefore call `permuted_index()` instead?
</blockquote></details>

Done, thanks
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZziOr763KDZyAw9ssD:-N_0Bxu11nrTpPyJ30qI:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L48)):*

nit,I haven't done the math, but below you require diagonal dominance, so maybe (if true) update comment to:  "Now the result is diagonally dominant"

Diagonal dominant implies SPD, but not the other way around.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 50 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzi2yoAXu7pmn6-ULQ:-N_0C4SI2cRkwbOwrdy2:ba4di8c) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L50)):*

nit, 
Zeroing off-diagonal entries of a diagonally dominant matrix does not affect diagonal dominance, which in turns implies SPDness.

</blockquote></details>

Done thanks.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzkSkl88tLjRa5DvPJ:-N_0CNFjALIxUiVTB75J:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L90)):*

minor, fyi, you can also do `dense_A.llt().solve(b1)` for a shorter spelling.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 110 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzl-2HD0sFqdu2VBg0:-N_0SuN0BcpECQFcuN-J:bw2722e) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L110)):*

what happens if A2 does not have the same sparsity pattern?
</blockquote></details>

A precondition of calling UpdateMatrix is that A2 has the same sparsity pattern as A. I clarified in the function documentation.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZznxnFCs3stZsQ6Xku:-N_0W_iX8-pX1r_Bb2OR:b-p7k91o) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L95)):*

The matrix is factored in place; the original matrix is gone, so the matrix is "not set".
</blockquote></details>

I think the pair of bools probably the cause of this trouble (it's challenging for me, too).

Isn't what we have an `enum SolverMode` that has exactly three values `Empty, Unfactored, Factored`.  Then the pre/post-conditions update the class's mode, and can be very concisely stated?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 95 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZznxnFCs3stZsQ6Xku:-N_0dGP97D-a1g2KO0q1:b-dtkdrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L95)):*

I think the pair of bools probably the cause of this trouble (it's challenging for me, too).

Isn't what we have an `enum SolverMode` that has exactly three values `Empty, Unfactored, Factored`.  Then the pre/post-conditions update the class's mode, and can be very concisely stated?
</blockquote></details>

Done, thanks.

xuchenhan-tri (commented):
+(priority: high) in hope to make the next Drake release.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_0dUw98MAtqfz16he0:b-d932hv)* status: 13 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

jwnimmer-tri (commented):
:lgtm: platform.

> Reviewed 1 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_0dNln8m7SgCOBiIbb:b-vtj4vo)* status: 20 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0_tBZ6DhRpTsY9d30:-N_0_tBZ6DhRpTsY9d31:b9gyuxh) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.h#L43)):*
> ```objc
>  @tparam BlockType The matrix type for individual block matrices, usually
>                    MatrixX<double>, but fixed size matrices are preferred if you
>                    know the sizes of blocks are uniform and fixed. */
> ```

This is true but incomplete, isn't it?  With most code in the cc file, it seems like `Matrix3d` and `MatrixXd` are the _only_ two valid types?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 45 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0a8du3Pf-uDTMHlWt:-N_0a8du3Pf-uDTMHlWu:b56zmi3) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.h#L45)):*
> ```objc
>                    know the sizes of blocks are uniform and fixed. */
> template <typename BlockType>
> class BlockSparseCholeskySolver {
> ```

https://drake.mit.edu/styleguide/cppguide.html#Inline_Functions

Other than the "mode" getter, none of the other functions seem like they should be inline?  (I guess the constructor is probably fine inline.)

In particular, the destructor is pretty fat so should be in cc.

Likewise the copying getters for L and P, etc.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 108 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0Z8Et-KBHtWyv9CSU:-N_0Z8Et-KBHtWyv9CSV:b-c1obsl) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.h#L108)):*
> ```objc
>    @post is_factored() == true iff the factorization is successful and `true` is
>    returned. */
>   bool Factor();
> ```

A call site that fails to check the error return is probably a buggy call site?



_Suggestion:_
```objc
[[nodiscard]] bool Factor();
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 116 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0b7nZ4bPOT4LmIif8:-N_0b7nZ4bPOT4LmIif9:b3mef65) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.h#L116)):*
> ```objc
>   VectorX<double> Solve(const Eigen::Ref<const VectorX<double>>& b) const;
> 
>   /* Solves the system A⋅x = b and writes the result in b.
> ```

nit The doc text "the result in **b**" does not match the C++ argument name `y`.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 4 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0_kmv2uRqRvpx1poy:-N_0_kmwA-uOpTC-m0_y:b-hszb06) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L4)):*
> ```cpp
> 
> #include <algorithm>
> #include <numeric>
> ```

nit Unused `numeric`?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 180 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0bnc031DvSNZFHc4t:-N_0bnc10h_ntsWUOtjF:bybmd41) ([raw file](https://github.com/RobotLocomotion/drake/blob/21dfb40fa131580b05eb8ce4628f2d344bd6d5d9/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L180)):*
> ```cpp
>     const SymmetricMatrix& A, const std::vector<int>& elimination_ordering) {
>   /* 1. Compute the block permutation as well as the scalar permutation. */
>   const int N = elimination_ordering.size();
> ```

nit Non-GSG (non-matrix) variable name.

Ditto throughout.



_Suggestion:_
```cpp
n
```
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 4 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0cM2851ZGM24XV7gY:-N_0cM2851ZGM24XV7gZ:b-hszb06) ([raw file](https://github.com/RobotLocomotion/drake/blob/21dfb40fa131580b05eb8ce4628f2d344bd6d5d9/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L4)):*
> ```cpp
> 
> #include <memory>
> #include <numeric>
> ```

nit Unused `numeric`?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 17 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 13 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 43 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_0_tBZ6DhRpTsY9d30:-N_0mPY_7-eEwBZEhVsp:b-w6it0w) ([raw file](https://github.com/RobotLocomotion/drake/blob/635fdb2bfcc463290be3e14bf256d5006e604a75/multibody/contact_solvers/block_sparse_cholesky_solver.h#L43)):*

I'm not sure what the exact requested change is. I modified the language to emphasize the limited possibility. PTAL.
</blockquote></details>

You got it right -- to convey that anything besides those two would fail to compile.

amcastro-tri (commented):
Excellent! I love the new mode introduced.
There is only one last lingering comment in `PermuteAndCopyToL()`, all other comments are minor and non-blocking.

> Reviewed 2 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_46kGF0JTeC-WA8VHD:bz813k1)* status: 7 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzoKeP0SagdScofZ6O:-N_40Hb2Cy7UrEvPuqBQ:boazfsu) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*

Good point. Done.
</blockquote></details>

minor, I don't think the return is documented yet.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 167 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzr6XX2RDnYEFLTUa3:-N_41BhxE5tPjCuR8jrd:b-6tcdih) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L167)):*

> would it be correct to say this is doing `L⋅Lᵀ = P⋅A⋅Pᵀ`? with P the permutation below.

Yes, this is correct. There is one caveat: I was cutting corner in r1 by not storing the diagonal blocks of L_ because we never touch it. I have since added it in r2 to perform the unit test you proposed.
</blockquote></details>

Thanks! the new docs do help a lot! (at least me)
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 168 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-LSGk-IocdGhLSGss:-N_41Kz4CyAEgPDe2tGx:bnd8hc9) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L168)):*

Eigen::LLT is faster than Eigen::LDLT so LLT is chosen over LDLT.
</blockquote></details>

I thought that had to do with pivoting though, and probably a bad implementation. In principle LDLT does not require square roots. Anyway, I see your point. Thanks.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 52 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_3zaEv3UeXGHV-wJjK:-N_3zaEv3UeXGHV-wJjL:b-ebkb5o) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.h#L52)):*
> ```objc
>   enum class SolverMode {
>     kEmpty,       // Matrix A is yet to be set.
>     kUnfactored,  // Matrix A is set but not yet factored.
> ```

minor, I think for those of us who understand the implications of using a sparse solver (the typical user), `kUnfactored` sounds just too vague, stating nothing (not even in these docs) about the work that was already done (symbolic assembly + permutation). What about `kSymbolicallyAnalyzed` or smilar?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 146 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_40u_e17QApThv77VN:-N_40u_e17QApThv77VO:b-l3i55r) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.h#L146)):*
> ```objc
>    Cholesky factorization of the permuted A.
>    @throws std::exception if solver_mode() == SolverMode::kEmpty. */
>   Eigen::PermutationMatrix<Eigen::Dynamic> CalcPermutationMatrix() const;
> ```

Since this is mostly for debugging I have no complains. However interesting you chose a sparse format for P but only the dense format for L. Why not to return a block sparse lower triangular matrix?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 206 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_41byuBSvzwRSzQNav:-N_41byv-Wkc12zK3GmE:b8lufbl) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.h#L206)):*
> ```objc
>   /* Permutation for block indices, same size as A.block_cols(). For a given
>    block index i into A, `block_permutation_[i]` gives the permuted block index
>    into L_. */
> ```

btw, very nice, thanks!
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 155 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzumRSBr1MY_5aansa:-N_4554o1N-7EAHJH7IN:bx9tzsn) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L155)):*

Done, I think?
</blockquote></details>

wel.. you tell me? I think it'd help for completeness to state something like "`Pᵢⱼ=1` for `j=scalar_elimination_ordering[i]` and zero otherwise"
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 162 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-8siHDbz64wzaXM-h:-N_45MhjE4NbwrZy_Sna:bhwk435) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L162)):*

The proposed notation doesn't help me in particular, but I'm neutral about the change. I tried it out in the new revision. See if you like it.
</blockquote></details>

thanks! being completely selfish, it does help me at least
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 244 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_-I0Xi-6x-XHAfzsRi:-N_45i_GB2NE1FfF7nR0:b-99dvue) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L244)):*

Done, thanks
</blockquote></details>

you updated above... but I believe here you also need `permuted_index()`? otherwise successive invocations of this method would keep changing the permutations.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 30 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_42W9Y5w-viUASYrvB:-N_42W9Y5w-viUASYrvC:b-3af4z) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L30)):*
> ```cpp
> template <typename BlockType>
> void BlockSparseCholeskySolver<BlockType>::UpdateMatrix(
>     const SymmetricMatrix& A) {
> ```

nit, 
consider `DRAKE_THROW_UNLESS(mode() != kEmpty)`.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 110 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzl-2HD0sFqdu2VBg0:-N_46Jfr5_2sOrmyJKM6:b-x2yv3c) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L110)):*

A precondition of calling UpdateMatrix is that A2 has the same sparsity pattern as A. I clarified in the function documentation.
</blockquote></details>

ah, thanks!
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 174 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_46WJZ5lbsORbku9uc:-N_46WJZ5lbsORbku9ud:b-10gtzk) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L174)):*
> ```cpp
>   const MatrixXd lhs = L * L.transpose();
>   const MatrixXd rhs = P * A_dense * P.transpose();
>   EXPECT_TRUE(CompareMatrices(lhs, rhs, 1e-14));
> ```

btw, nice!
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 182 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_46dey5LRSJ-u-24Nn:-N_46dey5LRSJ-u-24No:b-nhn7ok) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L182)):*
> ```cpp
>   /* CalcPermutatoinMatrix() before setting the matrix throws. */
>   EXPECT_THROW(solver.CalcPermutationMatrix(), std::exception);
>   /* After setting the matrix, CalcPermutatoinMatrix() returns the same result
> ```

minor,



_Suggestion:_
```cpp
CalcPermutationMatrix() 
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 7 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzoKeP0SagdScofZ6O:-N_47ROIDICuvSupwH15:bfkitye) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*

minor, I don't think the return is documented yet.
</blockquote></details>

The second sentence documents exactly that.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 52 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_3zaEv3UeXGHV-wJjK:-N_47Icf8vlBjo6Llzdj:bscflaq) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.h#L52)):*

minor, I think for those of us who understand the implications of using a sparse solver (the typical user), `kUnfactored` sounds just too vague, stating nothing (not even in these docs) about the work that was already done (symbolic assembly + permutation). What about `kSymbolicallyAnalyzed` or smilar?
</blockquote></details>

My vote would be to keep the constant name brief, but it would be fine to expand upon the comments here to explain in more detail.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 7 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-NZzoKeP0SagdScofZ6O:-N_4MUDD4NdEdy9kNzYJ:b-c1pgso) ([raw file](https://github.com/RobotLocomotion/drake/blob/412a21f795565d9a995a80fa6887c00b918b4107/multibody/contact_solvers/block_sparse_cholesky_solver.h#L96)):*

The second sentence documents exactly that.
</blockquote></details>

nvm, you are right. Thanks.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 7 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @xuchenhan-tri)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 52 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_3zaEv3UeXGHV-wJjK:-N_4MpHDCPw5yH2kxJJr:b-i2pikd) ([raw file](https://github.com/RobotLocomotion/drake/blob/58e5b8847f3cf9e53d13535b2b9f21ea98d2f4d1/multibody/contact_solvers/block_sparse_cholesky_solver.h#L52)):*

My vote would be to keep the constant name brief, but it would be fine to expand upon the comments here to explain in more detail.
</blockquote></details>

`kAnalyzed`? I feel like `kUnfactored` sounds a lot like "no work done".
I'll let Xuchen decide though.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: 2 unresolved discussions

amcastro-tri (commented):
Very good! there is one last typo. Ready to merge! 
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-:-N_55OOuEdjYB_xKbcwd:brjyb98)* status: 1 unresolved discussion

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 170 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_55FT_14EDB3_ovnQ3:-N_55FT_14EDB3_ovnQ4:b-xuxnw5) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d640674eb10de749114f5bdfc21f1eedd6c74c2/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L170)):*
> ```cpp
>    such that L⋅Lᵀ = P⋅A⋅Pᵀ.
>    More specificially, Pᵢⱼ = 1 for j = scalar_elimination_ordering[i] (or
>    equivalently i = scalar_permutation_[j]) and Pᵢⱼ = 1 otherwise. See
> ```

minor, woops!



_Suggestion:_
```cpp
Pᵢⱼ = 0 otherwise
```

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19741)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),amcastro-tri

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 170 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19741#-N_55FT_14EDB3_ovnQ3:-N_56jVmBCtIYt0Gr9Pb:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/9d640674eb10de749114f5bdfc21f1eedd6c74c2/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L170)):*

minor, woops!
</blockquote></details>

Done

PR #19742, opened 2023-07-09
CspaceFreeBox find separation certificates for a given box.

hongkai-dai:
Refactor the code in CspaceFreePolytope::FindSeparationCertificateGivenPolytope and CspaceFreeBox::FindSeparationCertificateGivenBox. Factor out the common part to CspaceFreePolytopeBase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)

### DIFF
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -280,6 +280,7 @@ drake_cc_googletest(
         "//common/test_utilities:symbolic_test_util",
         "//geometry/optimization:c_iris_test_utilities",
         "//geometry/optimization:cspace_free_box",
+        "//multibody/inverse_kinematics",
         "//solvers:mosek_solver",
     ],
 )
--- a/geometry/optimization/cspace_free_box.cc
+++ b/geometry/optimization/cspace_free_box.cc
@@ -1,9 +1,11 @@
 #include "drake/geometry/optimization/cspace_free_box.h"
 
 #include <array>
+#include <limits>
 #include <map>
 #include <memory>
 #include <unordered_set>
+#include <utility>
 
 #include "drake/common/fmt_eigen.h"
 #include "drake/geometry/optimization/cspace_free_internal.h"
@@ -11,6 +13,9 @@
 namespace drake {
 namespace geometry {
 namespace optimization {
+namespace {
+const double kInf = std::numeric_limits<double>::infinity();
+}  // namespace
 
 CspaceFreeBox::SeparationCertificateResult
 CspaceFreeBox::SeparationCertificate::GetSolution(
@@ -51,6 +56,33 @@ CspaceFreeBox::CspaceFreeBox(const multibody::MultibodyPlant<double>* plant,
     : CspaceFreePolytopeBase(plant, scene_graph, plane_order,
                              CspaceFreePolytopeBase::SForPlane::kOnChain,
                              options) {}
+
+bool CspaceFreeBox::FindSeparationCertificateGivenBox(
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+    const IgnoredCollisionPairs& ignored_collision_pairs,
+    const FindSeparationCertificateOptions& options,
+    std::unordered_map<SortedPair<geometry::GeometryId>,
+                       SeparationCertificateResult>* certificates) const {
+  std::vector<std::optional<SeparationCertificateResult>> certificates_vec =
+      this->FindSeparationCertificateGivenBox(
+          ignored_collision_pairs, q_box_lower, q_box_upper, options);
+
+  certificates->clear();
+  bool is_success = true;
+  for (const auto& certificate : certificates_vec) {
+    if (certificate.has_value()) {
+      const auto& plane = separating_planes()[certificate->plane_index];
+      certificates->emplace(
+          SortedPair<geometry::GeometryId>(plane.positive_side_geometry->id(),
+                                           plane.negative_side_geometry->id()),
+          std::move(certificate.value()));
+    } else {
+      is_success = false;
+    }
+  }
+  return is_success;
+}
 
 CspaceFreeBox::~CspaceFreeBox() {}
 
@@ -244,6 +276,77 @@ CspaceFreeBox::ConstructPlaneSearchProgram(
 
   DRAKE_DEMAND(gram_var_count == gram_vars.rows());
   return ret;
+}
+
+std::vector<std::optional<CspaceFreeBox::SeparationCertificateResult>>
+CspaceFreeBox::FindSeparationCertificateGivenBox(
+    const IgnoredCollisionPairs& ignored_collision_pairs,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+    const FindSeparationCertificateOptions& options) const {
+  Eigen::VectorXd s_box_lower;
+  Eigen::VectorXd s_box_upper;
+  Eigen::VectorXd q_star;
+  this->ComputeSBox(q_box_lower, q_box_upper, &s_box_lower, &s_box_upper,
+                    &q_star);
+  PolynomialsToCertify polynomials_to_certify;
+  this->GeneratePolynomialsToCertify(s_box_lower, s_box_upper, q_star,
+                                     ignored_collision_pairs,
+                                     &polynomials_to_certify);
+  std::vector<int> active_plane_indices;
+  active_plane_indices.reserve(polynomials_to_certify.plane_geometries.size());
+  for (const auto& plane_geometries : polynomials_to_certify.plane_geometries) {
+    active_plane_indices.push_back(plane_geometries.plane_index);
+  }
+  std::vector<std::optional<SeparationCertificateResult>> ret(
+      active_plane_indices.size(), std::nullopt);
+
+  // This lambda function formulates and solves a small SOS program for each
+  // pair of geometries.
+  auto solve_small_sos = [this, &polynomials_to_certify, &active_plane_indices,
+                          &options, &ret](int plane_count) -> bool {
+    const int plane_index = active_plane_indices[plane_count];
+    auto certificate_program = this->ConstructPlaneSearchProgram(
+        polynomials_to_certify.plane_geometries[plane_count],
+        polynomials_to_certify.s_minus_s_box_lower,
+        polynomials_to_certify.s_box_upper_minus_s);
+
+    solvers::MathematicalProgramResult result;
+    solvers::MakeSolver(options.solver_id)
+        ->Solve(*certificate_program.prog, std::nullopt, options.solver_options,
+                &result);
+    if (result.is_success()) {
+      ret[plane_count].emplace(certificate_program.certificate.GetSolution(
+          plane_index, separating_planes()[plane_index].a,
+          separating_planes()[plane_index].b,
+          separating_planes()[plane_index].decision_variables, result));
+      return true;
+    } else {
+      ret[plane_count].reset();
+      return false;
+    }
+  };
+  this->SolveCertificationForEachPlaneInParallel(
+      active_plane_indices, solve_small_sos, options.num_threads,
+      options.verbose, options.terminate_at_failure);
+  return ret;
+}
+
+void CspaceFreeBox::AddCspaceBoxContainment(
+    solvers::MathematicalProgram* prog,
+    const VectorX<symbolic::Variable>& s_box_lower,
+    const VectorX<symbolic::Variable>& s_box_upper,
+    const Eigen::MatrixXd& s_inner_pts) const {
+  const int s_size = this->rational_forward_kin().s().rows();
+  DRAKE_THROW_UNLESS(s_inner_pts.rows() == s_size);
+  DRAKE_THROW_UNLESS(s_box_lower.rows() == s_size);
+  DRAKE_THROW_UNLESS(s_box_upper.rows() == s_size);
+  // We have the constraint s_box_lower <= s <= s_box_upper.
+  prog->AddBoundingBoxConstraint(Eigen::VectorXd::Constant(s_size, -kInf),
+                                 s_inner_pts.rowwise().minCoeff(), s_box_lower);
+  prog->AddBoundingBoxConstraint(s_inner_pts.rowwise().maxCoeff(),
+                                 Eigen::VectorXd::Constant(s_size, kInf),
+                                 s_box_upper);
 }
 }  // namespace optimization
 }  // namespace geometry
--- a/geometry/optimization/cspace_free_box.h
+++ b/geometry/optimization/cspace_free_box.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <optional>
+#include <unordered_map>
 #include <vector>
 
 #include "drake/geometry/optimization/cspace_free_polytope_base.h"
@@ -154,6 +156,31 @@ class CspaceFreeBox : public CspaceFreePolytopeBase {
                 SeparatingPlaneOrder plane_order,
                 const Options& options = Options{});
 
+  /** Finds the certificates that the C-space box {q | q_box_lower <= q <=
+   * q_box_upper} is collision free.
+   *
+   * @param q_box_lower The lower bound of the C-space box.
+   * @param q_box_upper The upper bound of the C-space box.
+   * @param ignored_collision_pairs We ignore the pair of geometries in
+   * `ignored_collision_pairs`.
+   * @param[out] certificates Contains the certificate we successfully found for
+   * each pair of geometries. Notice that depending on `options`, the program
+   * could search for the certificate for each geometry pair in parallel, and
+   * will terminate the search once it fails to find the certificate for any
+   * pair. At termination, the pair of geometries whose optimization hasn't been
+   * finished will not show up in @p certificates.
+   * @retval success If true, then we have certified that the C-space box
+   * {q | q_box_lower<=q<=q_box_upper} is collision free. Otherwise
+   * success=false.
+   */
+  bool FindSeparationCertificateGivenBox(
+      const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+      const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+      const IgnoredCollisionPairs& ignored_collision_pairs,
+      const FindSeparationCertificateOptions& options,
+      std::unordered_map<SortedPair<geometry::GeometryId>,
+                         SeparationCertificateResult>* certificates) const;
+
  private:
   // Forward declare the tester class that will test the private members.
   friend class CspaceFreeBoxTester;
@@ -211,6 +238,37 @@ class CspaceFreeBox : public CspaceFreePolytopeBase {
       const PlaneSeparatesGeometries& plane_geometries,
       const VectorX<symbolic::Polynomial>& s_minus_s_lower,
       const VectorX<symbolic::Polynomial>& s_upper_minus_s) const;
+
+  /*
+   Finds the certificates that the C-space box {q | q_box_lower <= q <=
+   q_box_upper} is collision free.
+   @retval certificates certificates[i] is the separation certificate for a pair
+   of geometries. If we cannot certify or haven't certified the separation for
+   this pair, then certificates[i] contains std::nullopt. Note that when we run
+   this function in parallel and options.terminate_at_failure=true, we will
+   terminate all the remaining certification programs that have been launched,
+   so certificates[i] = std::nullopt could be either because that we have
+   attempted to find the certificate for this pair of geometry but failed, or it
+   could be that we fail to find the certificate for another pair and haven't
+   attempted to find the certificate for this pair.
+   The geometry pair which certificates[i] certifies is given by
+   separating_planes()[certificates[i].plane_index].geometry_pair().
+   */
+  [[nodiscard]] std::vector<std::optional<SeparationCertificateResult>>
+  FindSeparationCertificateGivenBox(
+      const IgnoredCollisionPairs& ignored_collision_pairs,
+      const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+      const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+      const FindSeparationCertificateOptions& options) const;
+
+  /*
+   Adds the constraint that each column of s_inner_pts is in the box s_box_lower
+   <= s <= s_box_upper.
+   */
+  void AddCspaceBoxContainment(solvers::MathematicalProgram* prog,
+                               const VectorX<symbolic::Variable>& s_box_lower,
+                               const VectorX<symbolic::Variable>& s_box_upper,
+                               const Eigen::MatrixXd& s_inner_pts) const;
 };
 }  // namespace optimization
 }  // namespace geometry
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -18,18 +18,6 @@ namespace optimization {
 namespace {
 const double kInf = std::numeric_limits<double>::infinity();
 
-// Checks if a future has completed execution.
-// This function is taken from monte_carlo.cc. It will be used in the "thread
-// pool" implementation (which doesn't use the openMP).
-template <typename T>
-bool IsFutureReady(const std::future<T>& future) {
-  // future.wait_for() is the only method to check the status of a future
-  // without waiting for it to complete.
-  const std::future_status status =
-      future.wait_for(std::chrono::milliseconds(1));
-  return (status == std::future_status::ready);
-}
-
 // Solves an optimization problem. If the optimization problem has a cost, then
 // after we find the optimal solution for that cost (where the optimal solution
 // would be on the boundary of the feasible set), we back-off a little bit and
@@ -301,8 +289,6 @@ CspaceFreePolytope::FindSeparationCertificateGivenPolytope(
     }
   }
 
-  std::vector<std::optional<bool>> is_success(active_plane_indices.size(),
-                                              std::nullopt);
   std::vector<std::optional<SeparationCertificateResult>> ret(
       active_plane_indices.size(), std::nullopt);
 
@@ -311,7 +297,7 @@ CspaceFreePolytope::FindSeparationCertificateGivenPolytope(
   auto solve_small_sos = [this, &d_minus_Cs, &C_redundant_indices,
                           &s_lower_redundant_indices,
                           &s_upper_redundant_indices, &active_plane_indices,
-                          &options, &is_success, &ret](int plane_count) {
+                          &options, &ret](int plane_count) -> bool {
     const int plane_index = active_plane_indices[plane_count];
     auto certificate_program = this->ConstructPlaneSearchProgram(
         this->plane_geometries_[plane_index], d_minus_Cs, C_redundant_indices,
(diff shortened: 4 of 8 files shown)


hongkai-dai (commented):
+@AlexandreAmice for feature review please, thanks!

@jwnimmer-tri may I ask you a question? I have two classes `CspaceFreePolytope` and `CspaceFreeBox`, both derived from `CspaceFreePolytopeBase`. Both these derived classes implement a function `FindSeparationCertificateGivenXXX` (where `XXX` is either `Polytope` or `Box`). These two functions do very similar things, except their input argument types are different. I would like to factor out these two functions into a common function, which I did in this PR.

I created a templated function in the base class `CspaceFreePolytopeBase::FindSeparationCertificateGivenPolytope`. This template function can be instantiated with different types in the child classes. Is this a good way to factor out the code with different input types? Thanks for your attention!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-:-N_26WRq6ZIoIFWROUzH:bjsjq4m)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_5hlEaBmKacSPlqPXI:b-xsr0dt) (no related file):*
> I have two classes `CspaceFreePolytope` and `CspaceFreeBox`, both derived from `CspaceFreePolytopeBase`. Both these derived classes implement a function `FindSeparationCertificateGivenXXX` (where `XXX` is either `Polytope` or `Box`). These two functions do very similar things, except their input argument types are different. I would like to factor out these two functions into a common function, which I did in this PR.
>
> I created a templated function in the base class `CspaceFreePolytopeBase::FindSeparationCertificateGivenPolytope`. This template function can be instantiated with different types in the child classes. Is this a good way to factor out the code with different input types?

If the only difference were in one argument type, I would say use `std::variant<>` for that one argument and no need for templates.

However, it seems like multiple argument types are different.  In that case, a function template is a fair choice.  However, don't put the code in the header.  Put the code in the cc file, and use explicit template instantiation for the two variations you need.

You might also want to document the helper function as "Internal use only" so that it doesn't become stable API.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_5mvusEt2MyxnzE5hM:bgbh8l0) (no related file):*
Thanks Jeremy!

> However, don't put the code in the header. Put the code in the cc file, and use explicit template instantiation for the two variations you need.

Sorry I am not sure how to do the explicit template instantiation here. The templated code is in the header file `cspace_free_polytope_base.h`. It is later instantiated with the template type `CspaceFreePolytope::SeparationCertificateProgram` and `CspaceFreeBox::SeparationCertificateProgram`. Notice that `cspace_free_polytope_base.h/cc` doesn't know about these two types because they are defined in the child classes. So where should I explicitly instantiate the template function?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_69-yh6mdYcb9FeiD3:b-p91nhv) (no related file):*

Thanks Jeremy!

> However, don't put the code in the header. Put the code in the cc file, and use explicit template instantiation for the two variations you need.

Sorry I am not sure how to do the explicit template instantiation here. The templated code is in the header file `cspace_free_polytope_base.h`. It is later instantiated with the template type `CspaceFreePolytope::SeparationCertificateProgram` and `CspaceFreeBox::SeparationCertificateProgram`. Notice that `cspace_free_polytope_base.h/cc` doesn't know about these two types because they are defined in the child classes. So where should I explicitly instantiate the template function?
</blockquote></details>

Maybe it should be a free function instead of a member function?  It could be in its own header file `cspace_free_internal.h` so that users would not need to see it.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_6EBcEEyuQ7Cb6I06c:ba49vsk) (no related file):*

Maybe it should be a free function instead of a member function?  It could be in its own header file `cspace_free_internal.h` so that users would not need to see it.
</blockquote></details>

Thanks Jeremy, do you mean that I put this template function `FindSeparationCertificateGivenPolytope` as a free function in the header file `cspace_free_internal.h`, but not doing explicit instantiation? I think that would work. I don't know how to do explicit instantiating of this template function because to instantiate the template it has to know about `CspaceFreeBox::SeparationCertificateProgram` and `CspaceFreePolytope::SeparationCertificateProgram`, which live in different library targets.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_6FEOR5jyz5-EwLROg:b-m0pd3u) (no related file):*

Thanks Jeremy, do you mean that I put this template function `FindSeparationCertificateGivenPolytope` as a free function in the header file `cspace_free_internal.h`, but not doing explicit instantiation? I think that would work. I don't know how to do explicit instantiating of this template function because to instantiate the template it has to know about `CspaceFreeBox::SeparationCertificateProgram` and `CspaceFreePolytope::SeparationCertificateProgram`, which live in different library targets.
</blockquote></details>

If the header file with the free function is only `#included` in those two cc files, and no header files, then we don't need explicit instantiation.  It will only ever be implicitly instantiated in the two places that need it, and will not pollute extra code into users' projects.

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_BgufA1U_-w6G5DaL4:bhr8pwx) (no related file):*

If the header file with the free function is only `#included` in those two cc files, and no header files, then we don't need explicit instantiation.  It will only ever be implicitly instantiated in the two places that need it, and will not pollute extra code into users' projects.
</blockquote></details>

I implemented the version that move the templated function to a header file, and only include the header file in the two cc files.

Honestly I like the previous version (with the templated function as a protected member function of the parent class) better. The templated function is already protected, so only the derived class can access the template function. But now with the template function in the header file, anybody who can access that header file can see this templated function and it becomes a public API. Does that make sense?

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_KtsRl27pyZcaEwkc1:b-n65syx) (no related file):*

I implemented the version that move the templated function to a header file, and only include the header file in the two cc files.

Honestly I like the previous version (with the templated function as a protected member function of the parent class) better. The templated function is already protected, so only the derived class can access the template function. But now with the template function in the header file, anybody who can access that header file can see this templated function and it becomes a public API. Does that make sense?
</blockquote></details>

@jwnimmer-tri would love to hear your thought on this one. Thanks!

___

AlexandreAmice (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_Ytlwt2pqaVNTfVvmF:binudt0) (no related file):*

@jwnimmer-tri would love to hear your thought on this one. Thanks!
</blockquote></details>

Happy to review once @jwnimmer-tri pitches in with his thoughts!

___

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_dtPtU-gUmqiY16Y83:b-x11nje) (no related file):*

Happy to review once @jwnimmer-tri pitches in with his thoughts!
</blockquote></details>

Jeremy says that he doesn't have time until July 31. So I reverted the last commit. Alex, could you review the PR? Thanks!

___

AlexandreAmice (commented):
> Reviewed 1 of 7 files at r1, 2 of 4 files at r3, 5 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 9 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_lELWCCWvVicjVfat-:b-b4zci3) (no related file):*

Jeremy says that he doesn't have time until July 31. So I reverted the last commit. Alex, could you review the PR? Thanks!
</blockquote></details>

Checkpoint.

I believe I understand the new flow and will double check myself soon. Will review tests shortly.

___
___
*[`geometry/optimization/cspace_free_box.h` line 167 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l1oR8BMGLQEFno2Ko:-N_l1oR8BMGLQEFno2Kp:bi6a9bp) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L167)):*
> ```objc
>    * `ignored_collision_pairs`.
>    * @param[out] certificates Contains the certificate we successfully found for
>    * each pair of geometries. Notice that depending on `options`, the program
> ```

This doesn't quite tell me what the `certificates` contain at the end if I terminate early. If I terminate early, will I certificates only contain the geometry pairs I successfully certified? Or will certificates be a `nullptr`.

I think my preference would be that `certificates` be of type `std::unordered_map<SortedPair<geometry::GeometryId>,std::optional<SeparationCertificateResult>>*` but would be happy with also just a clarification of the comment.
___
*[`geometry/optimization/cspace_free_box.h` line 241 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l2mVR7QOXKGUBAkeb:-N_l2mVR7QOXKGUBAkec:b-53us84) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L241)):*
> ```objc
>       const VectorX<symbolic::Polynomial>& s_upper_minus_s) const;
> 
>   /* Finds the certificates that the C-space box {q | q_box_lower <= q <=
> ```

nit: this comment should start on a new line.
___
*[`geometry/optimization/cspace_free_box.h` line 243 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l3KBm5O7C92PkmfX9:-N_l3KBm5O7C92PkmfXA:b-3ahq1m) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L243)):*
> ```objc
>   /* Finds the certificates that the C-space box {q | q_box_lower <= q <=
>    q_box_upper} is collision free.
>    @retval certificates certificates[i] is the separation certificate for a pair
> ```

Is there an invariant here. Is `certificates[i]` the certificate for the pair corresponding to `separating_planes()[i]`?
___
*[`geometry/optimization/cspace_free_box.h` line 261 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l3mJB0_4EFdPLdw7-:-N_l3mJC7wIXqYd5c4JZ:bb7lc46) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L261)):*
> ```objc
> 
>   /**
>    Adds the constraint that each column of s_inner-pts is in the box s_box_lower
> ```

nit



_Suggestion:_
```objc
s_inner_pts
```
___
*[`geometry/optimization/cspace_free_box.cc` line 307 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l5RyT3_jN4Yn4phF9:-N_l5RyT3_jN4Yn4phFA:b-1hzelo) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.cc#L307)):*
> ```cpp
>       polynomials_to_certify.plane_geometries, ignored_collision_pairs,
>       polynomials_to_certify, options,
>       [this](const PlaneSeparatesGeometries& plane_geometries,
> ```

Marking a note so I return once I understand this function.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 200 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l84MP2MXX9AxGF9pt:-N_l84MP2MXX9AxGF9pu:bfgafl7) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.h#L200)):*
> ```objc
>       SForPlane s_for_plane_enum) const;
> 
>   template <typename CspacePolytopeType, typename CertificateProgramType,
> ```

I see the problem with this being in the header. I unfortunately do not know how to resolve your problem either and will think about it on a second pass.

This needs documentation though
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 272 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lD1YYEzfJN_vl0G0Z:-N_lD1YYEzfJN_vl0G0_:bjxdmnl) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L272)):*
> ```cpp
>     const std::function<bool(int)>& solve_plane_sos, int num_threads,
>     bool verbose, bool terminate_at_failure) const {
>   num_threads = num_threads > 0
> ```

FYI: I was warned by Jeremy that if `num_threads = std::thread::hardware_concurrency` and then Mosek also attempts to multi-thread, we risk losing a good bit of performance. For toy examples, I would guess that having the number of threads be the hardware concurrency is a good idea, but for say the 14-dof example it may be worth cutting this down to let `Mosek` run faster.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 335 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lE6mb6btL27iJtScj:-N_lE6mb6btL27iJtSck:b-ifkb1q) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L335)):*
> ```cpp
>                   })) {
>     if (verbose) {
>       drake::log()->debug("Found Lagrangian multipliers and separating planes");
> ```

Since verbose is an option, I think it makes more sense to make this an `info` call?
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 356 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lEBQW2JE2fgRK6AR0:-N_lDs6c1iyWrvHeL1gP:b5h3860) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L356)):*
> ```cpp
>       }
> 
>       drake::log()->warn(
> ```

perhaps this should be a `debug` also to avoid flooding the user?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 7 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_box.h` line 167 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l1oR8BMGLQEFno2Ko:-N_oMrD1089eTNW7vQCv:b-vq1ccc) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L167)):*

This doesn't quite tell me what the `certificates` contain at the end if I terminate early. If I terminate early, will I certificates only contain the geometry pairs I successfully certified? Or will certificates be a `nullptr`.

I think my preference would be that `certificates` be of type `std::unordered_map<SortedPair<geometry::GeometryId>,std::optional<SeparationCertificateResult>>*` but would be happy with also just a clarification of the comment.

</blockquote></details>

Done. I added some documentation.

If the geometry pair hasn't finished certification at termination, then that geometry pair won't show up in the returned certificates.
___
*[`geometry/optimization/cspace_free_box.h` line 243 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l3KBm5O7C92PkmfX9:-N_oTPcO92NKyGGXtvHF:bxt0k4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_box.h#L243)):*

Is there an invariant here. Is `certificates[i]` the certificate for the pair corresponding to `separating_planes()[i]`?
</blockquote></details>

No, we don't have that invariance here.

If ignored_collision_pairs is empty, then certificates[i] is for the pair in separating_planes()[i]. Otherwise, the order in certificates is the same order in separating_planes(), but removing the pairs in ignored_collision_pairs.

I don't want to promise any specific ordering in `certificates`. Each certificates[i].plane_index indicates the plane and the geometry pair for that certificate.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 200 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_l84MP2MXX9AxGF9pt:-N_pjW241bn0CNRSp-AI:bgdbe6t) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.h#L200)):*

I see the problem with this being in the header. I unfortunately do not know how to resolve your problem either and will think about it on a second pass.

This needs documentation though
</blockquote></details>

I changed back to the version without the templates. There are actually not so much code duplication now. I only copied 11 lines from `cspace_free_polytope.cc` to `cspace_free_box.cc`.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 272 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lD1YYEzfJN_vl0G0Z:-N_pjkXTAQ0VQekH8pb4:b-4p07sb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L272)):*

FYI: I was warned by Jeremy that if `num_threads = std::thread::hardware_concurrency` and then Mosek also attempts to multi-thread, we risk losing a good bit of performance. For toy examples, I would guess that having the number of threads be the hardware concurrency is a good idea, but for say the 14-dof example it may be worth cutting this down to let `Mosek` run faster.
</blockquote></details>

When Mosek uses multi-thread, actually only a small portion of the MSK_optimize() function runs with multi-thread. Most of time I think it runs with only one thread. So I think although this might hurt the performance, it doesn't matter so much.

BTW, on the 14-dof example, the limitation is actually on the RAM size. I had to run it with only one thread, otherwise multiple threads would eat up all my 128GB memory.
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 335 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lE6mb6btL27iJtScj:-N_pklONC6Oa9IPjO1do:bb6a712) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L335)):*

Since verbose is an option, I think it makes more sense to make this an `info` call?
</blockquote></details>

I had it originally using `info`, but Jeremy suggested changing to `debug` in the PR https://reviewable.io/reviews/RobotLocomotion/drake/19492
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 356 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lEBQW2JE2fgRK6AR0:-N_plgpi-r1mz0I4rnKH:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L356)):*

perhaps this should be a `debug` also to avoid flooding the user?
</blockquote></details>

Done.

AlexandreAmice (commented):
> Reviewed 2 of 7 files at r4, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 2 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-N_uSBhQ6iXqAgmDq2tU:bg7igkv) (no related file):*

Checkpoint.

I believe I understand the new flow and will double check myself soon. Will review tests shortly.
</blockquote></details>

:lgtm: 

___
___
*[`geometry/optimization/cspace_free_box.h` line 254 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_tlsdY-_irW5uzsUFe:-N_tlsdY-_irW5uzsUFf:bp4qxd6) ([raw file](https://github.com/RobotLocomotion/drake/blob/7381f7f91178f32398aec0c4fb0f46d11891ead8/geometry/optimization/cspace_free_box.h#L254)):*
> ```objc
>    could be that we fail to find the certificate for another pair and haven't
>    attempted to find the certificate for this pair.
>    Note that the returned certificates[i] does NOT necessarily correspond to the
> ```

I would be more explicit and say `The geometry pair which certificates[i]  certifies is given by ...`
___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 272 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_lD1YYEzfJN_vl0G0Z:-N_uRgafEYZj2j4Ez4_q:b-67naa1) ([raw file](https://github.com/RobotLocomotion/drake/blob/3165cc35b37dd007264ff58531e97046241b701a/geometry/optimization/cspace_free_polytope_base.cc#L272)):*

When Mosek uses multi-thread, actually only a small portion of the MSK_optimize() function runs with multi-thread. Most of time I think it runs with only one thread. So I think although this might hurt the performance, it doesn't matter so much.

BTW, on the 14-dof example, the limitation is actually on the RAM size. I had to run it with only one thread, otherwise multiple threads would eat up all my 128GB memory.
</blockquote></details>

That is good to know. Thanks!

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform), missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_5hlEaBmKacSPlqPXH:-Na2zvvb8RcvvipJB2mh:b-o9wybp) (no related file):*

:lgtm: 
</blockquote></details>

+@rpoyner-tri for Monday's platform review please, thanks!

___
___
*[`geometry/optimization/cspace_free_box.h` line 254 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-N_tlsdY-_irW5uzsUFe:-Na2zlLt6pSiN7j34-rl:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/7381f7f91178f32398aec0c4fb0f46d11891ead8/geometry/optimization/cspace_free_box.h#L254)):*

I would be more explicit and say `The geometry pair which certificates[i]  certifies is given by ...`
</blockquote></details>

Done.

rpoyner-tri (commented):
:lgtm: api doc questions

> Reviewed 1 of 7 files at r1, 2 of 7 files at r4, 4 of 5 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-:-Na7hPJy1vG0KkXjVgBq:b7qn7tc)* status: 4 unresolved discussions, missing label for release notes (waiting on @hongkai-dai)

___
*[`geometry/optimization/cspace_free_box.h` line 264 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-Na7c_956BvEh1v-2Qfo:-Na7c_956BvEh1v-2Qfp:b-t1yapi) ([raw file](https://github.com/RobotLocomotion/drake/blob/081249e0c6987b2140277519553becd0d7c7ae27/geometry/optimization/cspace_free_box.h#L264)):*
> ```objc
>       const FindSeparationCertificateOptions& options) const;
> 
>   /**
> ```

This function is private, so it should not use the doxygen-aware comment marking.
___
*[`geometry/optimization/cspace_free_polytope_base.h` line 200 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19742#-Na7dR4l8rNufkFI8yKm:-Na7dR4l8rNufkFI8yKn:b-rl4ai3) ([raw file](https://github.com/RobotLocomotion/drake/blob/081249e0c6987b2140277519553becd0d7c7ae27/geometry/optimization/cspace_free_polytope_base.h#L200)):*
> ```objc
>       SForPlane s_for_plane_enum) const;
> 
>   /* For each pair of geometries, solve the certification problem to find their
> ```

As currently written, this function is still part of public API; there is no real way to prevent non-Drake clients from subclassing this public class.

So, it should either stay in public API and be "marked internal use only", or we should work out a different layout to remove it from public API. Given the PR history, it seems like you tried a few things already, so maybe that's not going to work out.

If it is to stay public API, it needs to have the doxygen-aware comment markings `/** */`. Depending on how you feel about *supporting* this as public API (possible deprecations etc.), it should maybe be marked "internal use only".

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 2 unresolved discussions, missing label for release notes (waiting on @hongkai-dai)

svenevs:
The mac-arm-ventura queue has been broken today, I cleared out all jobs and am relaunching the one needed for your PR.

AlexandreAmice (commented):
> Reviewed 1 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19742)* status: 1 unresolved discussion

PR #19746, opened 2023-07-09
[framework] Improve DiagramBuilder::Build performance

jwnimmer-tri:
Closes #19711.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746)

### DIFF
--- a/systems/framework/diagram.cc
+++ b/systems/framework/diagram.cc
@@ -73,10 +73,11 @@ Diagram<T>::get_output_port_locator(OutputPortIndex port_index) const {
 
 template <typename T>
 std::multimap<int, int> Diagram<T>::GetDirectFeedthroughs() const {
+  std::unordered_map<const System<T>*, std::multimap<int, int>> memoize;
   std::multimap<int, int> pairs;
   for (InputPortIndex u(0); u < this->num_input_ports(); ++u) {
     for (OutputPortIndex v(0); v < this->num_output_ports(); ++v) {
-      if (DiagramHasDirectFeedthrough(u, v)) {
+      if (DiagramHasDirectFeedthrough(u, v, &memoize)) {
         pairs.emplace(u, v);
       }
     }
@@ -1090,12 +1091,15 @@ const SystemBase& Diagram<T>::GetRootSystemBase() const {
 }
 
 template <typename T>
-bool Diagram<T>::DiagramHasDirectFeedthrough(int input_port, int output_port)
+bool Diagram<T>::DiagramHasDirectFeedthrough(
+    int input_port, int output_port,
+    std::unordered_map<const System<T>*, std::multimap<int, int>>* memoize)
     const {
   DRAKE_ASSERT(input_port >= 0);
   DRAKE_ASSERT(input_port < this->num_input_ports());
   DRAKE_ASSERT(output_port >= 0);
   DRAKE_ASSERT(output_port < this->num_output_ports());
+  DRAKE_ASSERT(memoize != nullptr);
 
   const auto input_ids = GetInputPortLocators(InputPortIndex(input_port));
   std::set<InputPortLocator> target_input_ids(input_ids.begin(),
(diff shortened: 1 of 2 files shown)


jwnimmer-tri:
+@Xining-Du please take a look at this draft, related to #19711.  Is it moving in the right direction?

(1) Does the benchmark roughly capture the kinds of things you care about?  If so, are there particular values for `num_subsystems` and `depth` we should add to the suite?

(2) Does the algorithmic change to the implementation work as fast as your code in #19711?  This is a simpler approach, so if its equally fast, we should use this version.

Xining-Du (commented):
(1) Yes the benchmark does capture our problem. I think our diagram is roughly (20,2) here, with much less number of systems but much more number of input/output ports. In particular, the benchmark shows much better scaling behavior with `depth`, which is the most important thing.
(2) I can confirm that this PR is as fast as PR`19711`, so this approach should be the way to go. Thanks for the simplification!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-:-N_GpziuAKdOyy2TBBmc:b-v3r6zm)* status: 2 unresolved discussions, LGTM missing from assignee Xining-Du, needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @jwnimmer-tri)

___
*[`systems/benchmarking/framework_benchmarks.cc` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-N_GtejJDtylQEOPEoJM:-N_GtejJDtylQEOPEoJN:b-hp3qiv) ([raw file](https://github.com/RobotLocomotion/drake/blob/b18ae6a6d4fe3b28c56ced5059ca1ba01d3946b5/systems/benchmarking/framework_benchmarks.cc#L120)):*
> ```cpp
>     ->Unit(benchmark::kMillisecond)
>     ->Args({3, 0})
>     ->Args({30, 0})
> ```

Just make sure I understand correctly: With memoize, the diagram loop check time should scale as O(N) where N is the total number of systems (assuming that GetDirectFeedthrough() is the most time consuming part)?
However for (30,1) I see the time is about 1000X the time for (30, 0), while (20,1) is only 20X (20,0). Is that caused by some bad performance with unordered_map::find() with (30, 1)? 
___
*[`systems/framework/diagram.cc` line 1126 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-N_GtMfI5ONbtaxEy8Kk:-N_GtMfI5ONbtaxEy8Kl:b-4qz7mv) ([raw file](https://github.com/RobotLocomotion/drake/blob/b18ae6a6d4fe3b28c56ced5059ca1ba01d3946b5/systems/framework/diagram.cc#L1126)):*
> ```cpp
>       DRAKE_ASSERT(inserted);
>     }
> #else  // This is the original (slow) version.
> ```

Is the idea to keep the old version for correctness verification and benchmark comparison, or we want to delete it eventually?

jwnimmer-tri (commented):
-@Xining-Du thanks!  I'll work on merging this now.

I moved the benchmark to #19810 so that we can merge benchmark vs the fix separately.

> Dismissed @Xining-Du from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-:-N_jC6fu13l8h0Uj-ERq:b-j0zje6)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`systems/framework/diagram.cc` line 1126 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-N_GtMfI5ONbtaxEy8Kk:-N_ZLqt-0DQYBoiT4iFR:bn3xrip) ([raw file](https://github.com/RobotLocomotion/drake/blob/b18ae6a6d4fe3b28c56ced5059ca1ba01d3946b5/systems/framework/diagram.cc#L1126)):*
<details><summary><i>Previously, Xining-Du wrote…</i></summary><blockquote>

Is the idea to keep the old version for correctness verification and benchmark comparison, or we want to delete it eventually?
</blockquote></details>

It was only temporary; during prototyping, I had both versions of the code here, to make it easier for me to run A/B tests without needing `git` hijinks -- I can just change the `#if 1` to `#if 0` as a quick test.
___
*[`systems/benchmarking/framework_benchmarks.cc` line 120 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-N_GtejJDtylQEOPEoJM:-N_jDiJj5LZ8CsxR8FYc:b-54h736) ([raw file](https://github.com/RobotLocomotion/drake/blob/b18ae6a6d4fe3b28c56ced5059ca1ba01d3946b5/systems/benchmarking/framework_benchmarks.cc#L120)):*
<details><summary><i>Previously, Xining-Du wrote…</i></summary><blockquote>

Just make sure I understand correctly: With memoize, the diagram loop check time should scale as O(N) where N is the total number of systems (assuming that GetDirectFeedthrough() is the most time consuming part)?
However for (30,1) I see the time is about 1000X for (30, 0), while (20,1) is only 20X (20,0). Is that caused by some bad performance with unordered_map::map() with (30, 1)? 
</blockquote></details>

This change only removes redundant calls to the same function more than once.

The overall graph readability search code is still pretty inefficient, by doing all pairs one by one.  Some kind of dynamic programming would be a lot better.

sherm1 (commented):
That's lovely! Feature :lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-:-N_kMU3D-J-PeE2pD1nN:b-m6b2la)* status: needs at least two assigned reviewers

sherm1 (commented):
+@sammy-tri for platform review Thu per schedule, please

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-:-N_kMiB-9V006Li7_fql:b-zcaiee)* status: LGTM missing from assignee sammy-tri(platform)

sammy-tri (commented):
Played with the benchmark a bit too, that's quite the speedup. :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19746#-:-N_onCIG3rsQNWa_oVIB:bk2ngu)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),sherm1(platform)

PR #19747, opened 2023-07-10
RenderEngineGl loads .mtl files

### LINKED ISSUE #19188: RenderEngineGl supports MTL files
Currently, `RenderEngineGl` **can** load MTL files via `tiny_obj_loader`, but the code ignores it to be consistent with other RenderEngine implementations.

As the first pass to support MTL files in RenderEngineGl, we will only accept an MTL file with one texture, i.e., a phong texture or `map_Kd`, and ignore if an MTL file has multiple textures with a warning.

zachfang:
Toward #18844 
Close #19188


zachfang (commented):
+@SeanCurtis-TRI for feature review. Thanks!

From a high-level visual test (in my local branch), I can correctly render textures using RenderEngineGl in different scenarios.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-NZxcpEw-PesCD0rsAbJ:bd3ba6z)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-NZxdHoP19k05Qk4KgRd:-NZxdHoP19k05Qk4KgRe:b-ripqks) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/test/internal_render_engine_gl_test.cc#L376)):*
> ```cpp
>     PerceptionProperties material;
>     material.AddProperty("label", "id", expected_label_);
>     if (use_texture) {
> ```

FYI. Most of the changes in the test file are combining the original Gl tests and the updated VTK tests. 
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 987 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-NZxdmyO2KEOFXJIZ4m9:-NZxdmyO2KEOFXJIZ4mA:bvxb0jh) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/test/internal_render_engine_gl_test.cc#L987)):*
> ```cpp
> 
> // Mostly identical as `MeshTest` except for the geometry type being Convex.
> TEST_F(RenderEngineGlTest, ConvexTest) {
> ```

FYI. I quickly attempted to merge `MeshTest` and `ConvexTest` by enumerating different test cases across the Shape type, the obj path, etc. However, it became a bit messy. Thus, I remain the current structure.



_Code quote:_
```cpp
TEST_F(RenderEngineGlTest, ConvexTest) {
```

SeanCurtis-TRI (commented):
:LGTM: With some small stuff.

One request, could I persuade you to not merge this until after #19739 has merged? I really want to get that one merged for the next release of Drake (to facilitate using the gltf stuff here in Tokyo). I suspect there *might* be some merge conflicts and would like to defer handling those until after the core .gltf stuff is merged.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-N_1onua4DCNvP7gmhT7:b-33a9tr)* status: 9 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`geometry/render/render_material.h` line 23 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1ghrvBmOyRcnv7iYx:-N_1ghrvBmOyRcnv7iYy:bxzhtvx) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render/render_material.h#L23)):*
> ```objc
>   std::filesystem::path diffuse_map;
> 
>   /* Whether the material definition is loaded from an .mtl file. */
> ```

nit: I anticipate that `RenderMesh` and `RenderMaterial` will be used more broadly than just .obj. As such, I'd rather not have specific references to .obj concepts. Can we rename this to `from_file` or some such thing?
___
*[`geometry/render/render_material.h` line 24 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1h15466iaKLuZNCT0:-N_1h15466iaKLuZNCT1:b-g49xje) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render/render_material.h#L24)):*
> ```objc
> 
>   /* Whether the material definition is loaded from an .mtl file. */
>   bool from_mtl;
> ```

nit: This needs to be default initialized. I'd default it to `false` so that you don't have to set it in `render_material.cc` as you have done now.
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 351 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1hpca9qxNa55fx3SP:-N_1hpca9qxNa55fx3SQ:b6v1sps) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.h#L351)):*
> ```objc
>   // Mapping from obj filename to the index into geometries_ containing the
>   // OpenGlGeometry representation of the mesh.
>   std::unordered_map<std::string, int> meshes_;
> ```

Had you considered making a single map that goes from filename to a struct containing the OpenGlGeometry index and an *optional* render material?
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 354 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1hf6gDCE096usAACp:-N_1hf6gDCE096usAACq:bn4rdet) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.h#L354)):*
> ```objc
> 
>   // Mapping from obj filename to its material iff the material definition is
>   // from the mesh itself, i.e., an .mtl file.
> ```

nit A linguistic subtlety to distance the code from being all about .obj files.



_Suggestion:_
```objc
// from the mesh itself, e.g., an .mtl file.
```
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 487 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1j80M4Cs0BobXD-lS:-N_1j80M4Cs0BobXD-lT:b-a2yiky) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L487)):*
> ```cpp
> 
> /* Returns the file name if it is a symlink. */
> std::string MaybeReadSymlinkFileName(const std::string& filename_in) {
> ```

nit: This would be better as:

```c++
// Given a filename (e.g., of a mesh), this produces a string that we use in
// our maps to guarantee we only load the file once.
std::string GetPathKey(const std::string& filename) {
  std::error_code path_error;
  const fs::path path = fs::canonical(filename, path_error);
  if (path_error) {
    throw std::runtime_error(
        fmt::format("RenderEngineGl: unable to access the file {}; {}",
                    filename, path_error.message()));
  }
  return path.string();
}
```

Links can link to other links. `fs::read_symlink` only follows the chain one step. `fs::canonical` follows it all the way. (Something I learned today.)

As for the new name, see my comment below.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 638 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1ldGz1r5B-iH4vQT9:-N_1ldGz1r5B-iH4vQTA:b-egewnq) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L638)):*
> ```cpp
>       kInternalGroup, kHasTexCoordProperty, geometry.has_tex_coord);
> 
>   const std::string filename = MaybeReadSymlinkFileName(filename_in);
> ```

nit: See the note in `GetMesh()` about this variable.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 973 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1meV-5pXe_5EWyX2n:-N_1meV02HXO7PeVgcFW:b-i580k4) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L973)):*
> ```cpp
> int RenderEngineGl::GetMesh(const string& filename_in) {
>   int mesh = -1;
>   const std::string filename = MaybeReadSymlinkFileName(filename_in);
> ```

nit: Calling this `filename` could lead to confusion about why we have two filenames, but we're passing `filename_in` to `LoadRenderMeshFromObj` (which is the right thing to do, of course).

Above, I proposed swapping `MaybeReadSymlinkFileName` with `GetPathKey()`. I'd recommend swapping `filename` with `path_key` and then it makes sense that you're only using it as a key in the map. (Same for `ImplementMesh()`.

(I recognize you're inheriting my mess, but there's nothing like stepping away from a problem to gain fresh eyes in how you messed it up in the first place.)
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 984 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1mc7Q6BNgGeBo617d:-N_1mc7Q6BNgGeBo617e:b-73iqii) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L984)):*
> ```cpp
>     const RenderMaterial material = mesh_data.material;
>     // If the material is defined by an .mtl file, we will associate it with
>     // the mesh every time the mesh is rendered.
> ```

nit: Semantic error.



_Suggestion:_
```cpp
// the mesh every time the mesh is instanced.
```
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 345 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1nsyd-f6cSnG-wRsb:-N_1nsye59-Dl5U_CG9x:b-kimxt4) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/test/internal_render_engine_gl_test.cc#L345)):*
> ```cpp
>   // Tests that don't instantiate their own renderers should invoke this.
>   void Init(const RigidTransformd& X_WR, bool add_terrain = false) {
>     const RenderEngineGlParams params{.default_clear_color = kBgColor};
> ```

BTW Thanks!

zachfang:
> One request, could I persuade you to not merge this until after https://github.com/RobotLocomotion/drake/pull/19739 has merged?

Yes, that is my plan as well when I was following up on the PR last night. I will continue fixing and rebasing until that PR is merged.

zachfang (commented):
All comments addressed. The CI would fail pending the update on 19739. Locally, `//geometry/render` and `//geometry/render_gl` passed the tests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-N_50L6eD6o7aNuX61CC:bxzpy9z)* status: 1 unresolved discussion, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`geometry/render_gl/internal_render_engine_gl.h` line 351 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1hpca9qxNa55fx3SP:-N_5-mSn9-BHVmABt6ub:blnzyz6) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.h#L351)):*

Had you considered making a single map that goes from filename to a struct containing the OpenGlGeometry index and an *optional* render material?
</blockquote></details>

Done. PTAL if this makes sense.
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 354 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1hf6gDCE096usAACp:-N_5-iCFBfHFZ6kKBxn4:b-nfj2bz) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.h#L354)):*

nit A linguistic subtlety to distance the code from being all about .obj files.
</blockquote></details>

Done. This part is removed due to merging two maps.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 487 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1j80M4Cs0BobXD-lS:-N_501xm1PfbPgTvdcbS:b69gotv) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L487)):*

nit: This would be better as:

```c++
// Given a filename (e.g., of a mesh), this produces a string that we use in
// our maps to guarantee we only load the file once.
std::string GetPathKey(const std::string& filename) {
  std::error_code path_error;
  const fs::path path = fs::canonical(filename, path_error);
  if (path_error) {
    throw std::runtime_error(
        fmt::format("RenderEngineGl: unable to access the file {}; {}",
                    filename, path_error.message()));
  }
  return path.string();
}
```

Links can link to other links. `fs::read_symlink` only follows the chain one step. `fs::canonical` follows it all the way. (Something I learned today.)

As for the new name, see my comment below.


</blockquote></details>

Done. Updated the code as suggested.
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 973 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_1meV-5pXe_5EWyX2n:-N_50A5q1Zj9jeTjtpid:b-hdlb4f) ([raw file](https://github.com/RobotLocomotion/drake/blob/f435b096d699c4971be8b2c8d731bd9657067ebf/geometry/render_gl/internal_render_engine_gl.cc#L973)):*

nit: Calling this `filename` could lead to confusion about why we have two filenames, but we're passing `filename_in` to `LoadRenderMeshFromObj` (which is the right thing to do, of course).

Above, I proposed swapping `MaybeReadSymlinkFileName` with `GetPathKey()`. I'd recommend swapping `filename` with `path_key` and then it makes sense that you're only using it as a key in the map. (Same for `ImplementMesh()`.

(I recognize you're inheriting my mess, but there's nothing like stepping away from a problem to gain fresh eyes in how you messed it up in the first place.)
</blockquote></details>

Done. I think it's a good idea!

SeanCurtis-TRI (commented):
A handful of comments requesting you clean up the mess I introduced in VTK.  I'll withhold the stamp until after we get all of the cruft out of the way.  Thanks.

> Reviewed 5 of 9 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-N_6QmpH0nzEjAjSKJU3:b-kj2k5w)* status: 4 unresolved discussions, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @zachfang)

___
*[`geometry/render_gl/internal_render_engine_gl.h` line 360 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6F6ku2TBZSPOR-WCz:-N_6F6ku2TBZSPOR-WD-:b-lk8qfp) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/internal_render_engine_gl.h#L360)):*
> ```objc
> 
>   struct RenderGlMesh {
>     int mesh_index;
> ```

nit: this also needs a moustache (`{}`).
___
*[`geometry/render_gl/internal_render_engine_gl.h` line 361 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6FdhP-vVG8yIBn_vO:-N_6FdhP-vVG8yIBn_vP:bpatm2v) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/internal_render_engine_gl.h#L361)):*
> ```objc
>   struct RenderGlMesh {
>     int mesh_index;
>     std::optional<geometry::internal::RenderMaterial> mesh_material{
> ```

nit: Probably worth explicitly documenting the semantics of what it means when this field is `nullopt` or not.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 394 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6HbRu8Bp7pnqkvbuw:-N_6HbRu8Bp7pnqkvbux:b-919e6q) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/test/internal_render_engine_gl_test.cc#L394)):*
> ```cpp
>           FindResourceOrThrow("drake/geometry/render/test/meshes/box.png"));
>     } else {
>       const Rgba color_n(default_color_.r / 255.0, default_color_.g / 255.0,
> ```

BTW What's the significance of the `_n` suffix? Apparently copypasta from what I'd done in VTK. Now, I'm a bit hard-pressed to remember why I added that.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 656 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6I4-9D0CiJoOHbEu8:-N_6I4-A7Tqh7ibWyzOW:b-riz3da) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/test/internal_render_engine_gl_test.cc#L656)):*
> ```cpp
>       // size of the box. We confirm that doesn't actually happen. We test both
>       // the un-tiled default behavior and the ability to scale the texture.
>       PerceptionProperties props = simple_material(use_texture);
> ```

nit: There's no point in calling `use_texture` and `UpdateProperty`. Calling `simple_material()` (without any parameters) and `props.AddProperty()` would suffice. (Again, this one is on me.)

zachfang:
Check if I rebase correctly first, I will address the latest comments now and remove the `don't review tag` once I get to the review-able state again.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747)* status: 2 unresolved discussions, needs at least two assigned reviewers, missing label for release notes (waiting on @zachfang)

___
*[`geometry/render_gl/internal_render_engine_gl.h` line 361 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6FdhP-vVG8yIBn_vO:-N_6qxmI0buGvkPVFTvv:b-q4ve1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/internal_render_engine_gl.h#L361)):*

nit: Probably worth explicitly documenting the semantics of what it means when this field is `nullopt` or not.
</blockquote></details>

I updated the comments. PTAL.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 394 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6HbRu8Bp7pnqkvbuw:-N_6qmyA-6ryOyW7Sfvg:b6a8ztz) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/test/internal_render_engine_gl_test.cc#L394)):*

BTW What's the significance of the `_n` suffix? Apparently copypasta from what I'd done in VTK. Now, I'm a bit hard-pressed to remember why I added that.
</blockquote></details>

Changed to `default_color`.

zachfang:
RenderEngineVtk doesn't like the changes with an error:
```
[ RUN      ] RenderEngineVtkTest.BoxTest
geometry/render_vtk/test/internal_render_engine_vtk_test.cc:472: Failure
Value of: CompareColor(expected_color_, color, inlier)
  Actual: false (Expected: (4, 241, 33, 255) at (320, 240), tested: (4, 216, 30, 255) with tolerance: 1.0009999999999999)
Expected: true
Color at: (320, 240) for test: Box test - unscaled texture
geometry/render_vtk/test/internal_render_engine_vtk_test.cc:472: Failure
Value of: CompareColor(expected_color_, color, inlier)
  Actual: false (Expected: (130, 119, 16, 255) at (320, 240), tested: (118, 107, 14, 255) with tolerance: 1.0009999999999999)
Expected: true
Color at: (320, 240) for test: Box test - scaled texture
```

Looks like it's still comparing to the correct image just something is slightly off. I will investigate more tomorrow.

SeanCurtis-TRI (commented):
:LGTM:

+@EricCousineau-TRI for platform review, please.

> Reviewed 1 of 9 files at r3, 23 of 24 files at r5, 2 of 4 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-N_BXyRS5wUknOsh98AV:b-a3qv2x)* status: 3 unresolved discussions, LGTM missing from assignee EricCousineau-TRI(platform)

___
*[`geometry/render_gl/internal_render_engine_gl.h` line 370 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_BWVEY2T37_dR9QvVu:-N_BWVEY2T37_dR9QvVv:bczuw0u) ([raw file](https://github.com/RobotLocomotion/drake/blob/8740f0038d88a5da65eeff91253dd6afd25fffbe/geometry/render_gl/internal_render_engine_gl.h#L370)):*
> ```objc
>   };
> 
>   // Mapping from the obj filename to RenderGlMesh.
> ```

BTW In retrospect, it might be worth noting it's a mapping from the obj's __*canonical*__ filename to `RenderGlTMesh`.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 656 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6I4-9D0CiJoOHbEu8:-N_BWfB547WJPFM4DUvD:byhq7fe) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/test/internal_render_engine_gl_test.cc#L656)):*

OK. I think the original intent is when `use_texture=true`, we add the RGBA=(1, 1, 1, 1) diffuse color to show faithful texture color. As a diffuse_map is added in `simple_material()`, we need `UpdateProperty()` instead of `AddProperty()`.

By not supplying `use_texture` as an arg, the texture will then be modulated by the default color and thus introduces the color difference to cause the test to fail. 
</blockquote></details>

Oh, yeah. There is a point. :) So, we have two options:

1. *Also* set the `(phong, diffuse)` value to white, or
2. Put a note on why we're doing it this way to prevent future confusion.

I'm not sure either one is the clear winner. But I guess I have a slight bias for number 1.

1. The documentation explaining why we're doing this probably takes as many lines as simply setting the diffuse color. And it still leaves this weird indirection in place.
2. Putting both together in the test will make it clear *at the test* what matters.
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 647 at r7](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_BXgvt-seBTlwPQMcD:-N_BXgvt-seBTlwPQMcE:b-vo01bn) ([raw file](https://github.com/RobotLocomotion/drake/blob/8740f0038d88a5da65eeff91253dd6afd25fffbe/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L647)):*
> ```cpp
>       if (use_texture) {
>         props.UpdateProperty("phong", "diffuse_map",
>                           FindResourceOrThrow(
> ```

nit: Ended up with some weird horizontal whitespace here.

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747)* status: LGTM missing from assignee EricCousineau-TRI(platform)

___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 656 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_6I4-9D0CiJoOHbEu8:-N_BhVEJ6ZDW6SejbMXk:begri3g) ([raw file](https://github.com/RobotLocomotion/drake/blob/e5e416f5b472c7013cc8a8b6a2db65219a81b936/geometry/render_gl/test/internal_render_engine_gl_test.cc#L656)):*

Oh, yeah. There is a point. :) So, we have two options:

1. *Also* set the `(phong, diffuse)` value to white, or
2. Put a note on why we're doing it this way to prevent future confusion.

I'm not sure either one is the clear winner. But I guess I have a slight bias for number 1.

1. The documentation explaining why we're doing this probably takes as many lines as simply setting the diffuse color. And it still leaves this weird indirection in place.
2. Putting both together in the test will make it clear *at the test* what matters.
</blockquote></details>

Done. I went with the first option.

EricCousineau-TRI (commented):
:lgtm: platform w/ some small nits, thanks!

> Reviewed 4 of 9 files at r3, 1 of 3 files at r4, 1 of 4 files at r6, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-:-N_Ead7H3Ktys6T9Smnr:bdne49b)* status: 3 unresolved discussions

___
*[`geometry/render_gl/internal_render_engine_gl.h` line 359 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EYQG06u1myjWtXp48:-N_EYQG1DNfmRatfjK9C:b-hpe9ja) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/internal_render_engine_gl.h#L359)):*
> ```objc
>   // and allowed deviation within a small tolerance, then we could reuse them.
> 
>   // A data struct includes the index into geometries_ containing the
> ```

BTW grammar "A data struct that includes" ?
___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 651 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EaP4E-Lvu8JkrYmej:-N_EaP4E-Lvu8JkrYmek:bts8pdw) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/internal_render_engine_gl.cc#L651)):*
> ```cpp
>     material = MakeMeshFallbackMaterial(data.properties, filename_in,
>                                         parameters_.default_diffuse,
>                                         drake::internal::DiagnosticPolicy());
> ```

nit (minor) Unclear why a temporary `DiagnosticPolicy` is constructed vs. using something shared / centralized.

Consider centralizing in class as member, e.g. `diagnostic_policy_`.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 952 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_E_DNo8Gc1kTzU2Ee9:-N_E_DNpBgZQogL5LqkU:b-ilav4q) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/test/internal_render_engine_gl_test.cc#L952)):*
> ```cpp
> // Performs the shape centered in the image with a mesh (which happens to be a
> // box). The textured box will be one that is textured via its mtl library. We
> // use it to confirm that VTK is properly _invoking_ the obj material handling;
> ```

nit Why is VTK involved here? Do you mean OpenGl?
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 1001 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EZfTv4-D9PbhDIaKj:-N_EZfTw-aDbE8qfGLQt:b59wxct) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/test/internal_render_engine_gl_test.cc#L1001)):*
> ```cpp
>     Init(X_WR_, true);
> 
>     // N.B. box_no_mtl.obj doesn't exist in the repo. It is generated from
> ```

nit "repo" is ambiguous. Do you mean "This file is generated by Bazel to [...]" , or "This file does not directly live in the source tree; it is generated by the build system [...]" ?

zachfang (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform)

___
*[`geometry/render_gl/internal_render_engine_gl.cc` line 651 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EaP4E-Lvu8JkrYmej:-N_EwMuV5bSXiFoV-bLU:b-47wcxl) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/internal_render_engine_gl.cc#L651)):*

nit (minor) Unclear why a temporary `DiagnosticPolicy` is constructed vs. using something shared / centralized.

Consider centralizing in class as member, e.g. `diagnostic_policy_`.
</blockquote></details>

Done. I also added a private member to `RenderEngineVtk`.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 952 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_E_DNo8Gc1kTzU2Ee9:-N_EwXUw0xGKPtaoQd-p:b-i5w8vj) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/test/internal_render_engine_gl_test.cc#L952)):*

nit Why is VTK involved here? Do you mean OpenGl?
</blockquote></details>

Done. I changed to `render engine` as it's technically `tinyobjloader` doing all the work now for both render engines.
___
*[`geometry/render_gl/test/internal_render_engine_gl_test.cc` line 1001 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EZfTv4-D9PbhDIaKj:-N_Ewnlc3a2XnqF7SGrp:b-gqke88) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/test/internal_render_engine_gl_test.cc#L1001)):*

nit "repo" is ambiguous. Do you mean "This file is generated by Bazel to [...]" , or "This file does not directly live in the source tree; it is generated by the build system [...]" ?
</blockquote></details>

Done. I have updated other places using (copying) that.

EricCousineau-TRI (commented):
> Reviewed 6 of 6 files at r9.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19747)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform)

jwnimmer-tri:
*[`geometry/render_gl/internal_render_engine_gl.cc` line 651 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19747#-N_EaP4E-Lvu8JkrYmej:-N_F1fCu0Fhnq8cc_25D:b-hn40ac) ([raw file](https://github.com/RobotLocomotion/drake/blob/8212cf50bc2fa1d21de847a71509346ed1598ff7/geometry/render_gl/internal_render_engine_gl.cc#L651)):*

Done. I also added a private member to `RenderEngineVtk`.
</blockquote></details>

The temporary object made it 100% clear that the policy is always the default one.  Using a member field implies that the policy might be configurable, when in fact it is not.

zachfang:
> The temporary object made it 100% clear that the policy is always the default one.

I can submit a quick PR to invert that.

PR #19762, opened 2023-07-12
Compute MD ordering with priority nodes

xuchenhan-tri:
Compute the minimum degree ordering of a sparsity pattern under the constraint that certain nodes are eliminated before others. This is useful for computing Schur complement as a side effect of a right-looking Cholesky factorization as the variables w.r.t which the Schur complement is computed must be eliminated first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19762)

PR #19764, opened 2023-07-13
Compute in-group Minimum Degree ordering

xuchenhan-tri:
Given a block sparsity pattern G on vertices v = {0, 1, ..., n-1} and a partition on v = v₁ ∪ v₂ (such that v₁ ∩ v₂ = ∅), computes an elimination for v₁ and v₂ respectively using their induced graphs and concatenates them.

This is useful for computing Schur complement as a side effect of a right-looking Cholesky factorization as the variables w.r.t which the Schur complement is computed must be eliminated first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -157,6 +157,7 @@ drake_cc_library(
     deps = [
         ":block_sparse_lower_triangular_or_symmetric_matrix",
         "//common:essential",
+        "//multibody/contact_solvers/sap:partial_permutation",
     ],
 )
 
--- a/multibody/contact_solvers/minimum_degree_ordering.cc
+++ b/multibody/contact_solvers/minimum_degree_ordering.cc
@@ -5,6 +5,8 @@
 #include <set>
 #include <utility>
 
+#include "drake/multibody/contact_solvers/sap/partial_permutation.h"
+
 namespace drake {
 namespace multibody {
 namespace contact_solvers {
@@ -209,6 +211,80 @@ BlockSparsityPattern SymbolicCholeskyFactor(
     }
   }
   return BlockSparsityPattern(block_sizes, L_sparsity);
+}
+
+std::vector<int> CalcAndConcatenateMdOrderingWithinGroup(
+    const BlockSparsityPattern& global_pattern,
+    const std::unordered_set<int>& v1) {
+  /* Sizes of v, v1, and v2. */
+  const int n = global_pattern.block_sizes().size();
+  const int n1 = v1.size();
+  const int n2 = n - n1;
+
+  /* Mapping from v to v1 and v2 and the inverse mappings. */
+  PartialPermutation v1_permutation(n);
+  PartialPermutation v2_permutation(n);
+  std::vector<int> global_to_local(n);
+
+  auto in_v1 = [&](int b) {
+    return v1.count(b) > 0;
+  };
+
+  for (int i = 0; i < n; ++i) {
+    if (in_v1(i)) {
+      v1_permutation.push(i);
+      global_to_local[i] = v1_permutation.permuted_index(i);
+    } else {
+      v2_permutation.push(i);
+      global_to_local[i] = v2_permutation.permuted_index(i);
+    }
+  }
+
+  /* The number of scalar variables in each block (as needed for the block
+   sparsity pattern for G1 and G2). */
+  std::vector<int> v1_block_sizes(n1);
+  std::vector<int> v2_block_sizes(n2);
+  const std::vector<int>& global_block_sizes = global_pattern.block_sizes();
+  v1_permutation.Apply(global_block_sizes, &v1_block_sizes);
+  v2_permutation.Apply(global_block_sizes, &v2_block_sizes);
+
+  /* Build the induced graphs G1 and G2 from the global graph G. */
+  const std::vector<std::vector<int>>& G = global_pattern.neighbors();
+  std::vector<std::vector<int>> G1(n1);
+  std::vector<std::vector<int>> G2(n2);
+  for (int a = 0; a < n; ++a) {
+    for (int b : G[a]) {
+      if (in_v1(a) != in_v1(b)) {
+        /* One of a and b is in v1 and the other is in v2, so the edge ab is not
+         in either of the induced graph. */
+        continue;
+      }
+      const int j = std::min(global_to_local[a], global_to_local[b]);
+      const int i = std::max(global_to_local[a], global_to_local[b]);
+      if (in_v1(b)) {
+        G1[j].emplace_back(i);
+      } else {
+        G2[j].emplace_back(i);
+      }
+    }
+  }
+
+  const std::vector<int> v1_ordering = ComputeMinimumDegreeOrdering(
+      BlockSparsityPattern(std::move(v1_block_sizes), std::move(G1)));
+  const std::vector<int> v2_ordering = ComputeMinimumDegreeOrdering(
+      BlockSparsityPattern(std::move(v2_block_sizes), std::move(G2)));
+
+  std::vector<int> result;
+  result.reserve(n);
+  /* The v1 vertices come first. */
+  for (int v : v1_ordering) {
+    result.emplace_back(v1_permutation.domain_index(v));
+  }
+  /* The v2 vertices follow. */
+  for (int v : v2_ordering) {
+    result.emplace_back(v2_permutation.domain_index(v));
+  }
+  return result;
 }
 
 }  // namespace internal
--- a/multibody/contact_solvers/minimum_degree_ordering.h
+++ b/multibody/contact_solvers/minimum_degree_ordering.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <unordered_set>
 #include <vector>
 
 #include "drake/multibody/contact_solvers/block_sparse_lower_triangular_or_symmetric_matrix.h"
@@ -109,6 +110,21 @@ std::vector<int> ComputeMinimumDegreeOrdering(
  matrix. */
 BlockSparsityPattern SymbolicCholeskyFactor(
     const BlockSparsityPattern& block_sparsity);
+
+/* Given a block sparsity pattern G on vertices v = {0, 1, ..., n-1} and a
+ partition on v = v₁ ∪ v₂ (such that v₁ ∩ v₂ = ∅), computes an elimination
+ ordering on v in the following way:
+  1. Generate the v₁-induced subgraph G₁ and the v₂-induced subgraph G₂.
+  2. Compute the Minimum Degree ordering on G₁ and G₂ respectively.
+  3. Concatenate the orderings so that all vertices in v₁ come before vertices
+     in v₂.
+ @param[in] global_pattern  The block sparsity pattern G.
+ @param[in] v1              The vertices in the set v₁.
+ @returns  The elimination ordering obtained by following the algorithm
+ described above. */
+std::vector<int> CalcAndConcatenateMdOrderingWithinGroup(
+    const BlockSparsityPattern& global_pattern,
+    const std::unordered_set<int>& v1);
 
 }  // namespace internal
 }  // namespace contact_solvers
--- a/multibody/contact_solvers/test/minimum_degree_ordering_test.cc
+++ b/multibody/contact_solvers/test/minimum_degree_ordering_test.cc
@@ -217,6 +217,53 @@ GTEST_TEST(MinimumDegreeOrderingTest, SymbolicCholeskyFactor) {
   EXPECT_EQ(L_sparsity[5], std::vector<int>({5}));
 }
 
+/* In this test, we make a graph with 8 vertices, {0, 1, ..., 7}, such that
+ odd indexed vertices belong to v₁ and even indexed vertices belong to v₂.
+ Within v₁ and v₂, the block sparsity pattern looks like
+    X X | O O O | O O O O | O O O
+    X X | O O O | O O O O | O O O
+    --- | ----- |---------| -----
+    O O | X X X | X X X X | O O O
+    O O | X X X | X X X X | O O O
+    O O | X X X | X X X X | O O O
+    --- | ----- |---------| -----
+    O O | X X X | X X X X | O O O
+    O O | X X X | X X X X | O O O
+    O O | X X X | X X X X | O O O
+    O O | X X X | X X X X | O O O
+    --- | ----- |---------| -----
+    O O | O O O | O O O O | X X X
+    O O | O O O | O O O O | X X X
+    O O | O O O | O O O O | X X X
+ The expected elimination ordering for this block sparsity pattern is
+ [0, 3, 2, 1] from pen and paper calculation. 2 is eliminated before 1 because
+ when 0 and 3 are eliminated, the degree of 2 is 3 and the degree of 1 is 4.
+
+ The global to local index mapping looks like
+  0->0, 2->1, 4->2, 6->3
+  1->0, 3->1, 5->2, 7->3.
+ Because vertices in v₁ appear first in the resulting ordering, the final result
+ should be [1, 7, 5, 3, 0, 6, 4, 2].
+ We arbitrarily add edges across v₁ and v₂ (4-5, 4-7, 0-7, 0-5, 2-1) but they do
+ not affect the result. */
+GTEST_TEST(BlockSparseCholeskySolverTest, ConcatenateMdOrderingWithinGroup) {
+  std::vector<std::vector<int>> sparsity;
+  sparsity.emplace_back(std::vector<int>{0, 5, 7});
+  sparsity.emplace_back(std::vector<int>{1, 2});
+  sparsity.emplace_back(std::vector<int>{2, 4});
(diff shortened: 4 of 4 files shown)


xuchenhan-tri (commented):
+@joemasterjohn for feature review please.

This is the heuristic that you proposed in softsim. For the examples I tested this on, it's comparable with the alternative heuristic in #19762. I plan to run both and perform the factorization with the winner.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-:-N_BihCKA_967OZrvqxt:by4353a)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

joemasterjohn (commented):
Feature pass done. PTAL.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-:-N_bhaY48NVHuhgqG2Uc:blhsnxn)* status: 7 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bY7mZ7htkdJ8d6ZaF:-N_bY7mZ7htkdJ8d6ZaG:botuxie) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L246)):*
> ```cpp
>   }
> 
>   auto in_v1 = [&](int b) -> bool {
> ```

nit: explicit reutrn type not needed.



_Code quote:_
```cpp
-> bool
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 248 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bgzqU-5tpQzOYPSvr:-N_bgzqU-5tpQzOYPSvs:b-10bnxw) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L248)):*
> ```cpp
>   auto in_v1 = [&](int b) -> bool {
>     return v1.count(b) > 0;
>   };
> ```

nit: Here you could use 2 `contact_solvers::internal::PartialPermutation`s instead of plain old vectors to give some semantics to these mappings and shorten the code below.



_Suggestion:_
```cpp
  /* Mapping from v to v1 and v2 and the inverse mappings. */
  PartialPermutation v1_permutation(n);
  PartialPermutation v2_permutation(n);
  /* The number of scalar variables in each block (as needed for the block
   sparsity pattern for G1 and G2). */
  std::vector<int> v1_block_sizes(n1);
  std::vector<int> v2_block_sizes(n2);

  auto in_v1 = [&](int b) {
    return v1.count(b) > 0;
  };

  for (int i = 0; i < n; ++i) {
    if (in_v1(i)) {
      v1_permutation.push(i);
    } else {
      v2_permutation.push(i);
    }
  }
  
  const std::vector<int>& global_block_sizes = global_pattern.block_sizes();
  v1_permutation.Apply(global_block_sizes, &v1_block_sizes);
  v2_permutation.Apply(global_block_sizes, &v2_block_sizes);
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 253 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bcgyG4zngmq1FvdFH:-N_bcgyG4zngmq1FvdFI:b-a6ka61) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L253)):*
> ```cpp
>   std::vector<std::vector<int>> G1(n1);
>   std::vector<std::vector<int>> G2(n2);
>   for (int a = 0; a < n; ++a) {
> ```

nit: some STL tricks you can use here:



_Suggestion:_
```cpp
  for (int a = 0; a < n; ++a) {
    const bool in_v1_a = in_v1(a);
    const int a_local = in_v1_a ? v1_permutation.permuted_index(a)
                                : v2_permutation.permuted_index(a);
    std::vector<int>& G_a = in_v1_a ? G1[a_local] : G2[a_local];
    std::copy_if(G[a].begin(), G[a].end(), G_a.begin(), [&](int v) {
      return in_v1(v) == in_v1_a;
    });
    std::transform(G_a.begin(), G_a.end(), G_a.begin(), [&](int v) {
      return in_v1_a ? v1_permutation.permuted_index(v)
                     : v2_permutation.permuted_index(v);
    });
  }
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 260 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZmIDA3JUgbs6liNS:-N_bZmIDA3JUgbs6liNT:b776upm) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L260)):*
> ```cpp
>         continue;
>       }
>       const int j = std::min(global_to_local[a], global_to_local[b]);
> ```

nit: this local index access can be moved out of the inner loop. See my comment below about why `a` is always the min index.



_Suggestion:_
```cpp
const int j = global_to_local[a];
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_b_DKpEGCfnPh8I_i8:-N_b_DKpEGCfnPh8I_i9:b3ttqwu) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L261)):*
> ```cpp
>       }
>       const int j = std::min(global_to_local[a], global_to_local[b]);
>       const int i = std::max(global_to_local[a], global_to_local[b]);
> ```

BTW the global to local mapping is stable so global_to_local[a] < global_to_local[b] for all b in G[a] because a < b for all b in G[a] (when `in_v1(a) == in_v1(b)`, so no need to do this min/max. `j = global_to_local`



_Suggestion:_
```cpp
      const int i = global_to_local[b];
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 262 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_baAsB4IcGLkgCOl_S:-N_baAsB4IcGLkgCOl_T:brga9vu) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L262)):*
> ```cpp
>       const int j = std::min(global_to_local[a], global_to_local[b]);
>       const int i = std::max(global_to_local[a], global_to_local[b]);
>       if (in_v1(b)) {
> ```

I think this is incorrect, shouldn't it be whether `j` is in v1 or not? If you take out the min/max for i/j above you can just create a reference to `G1[global_to_local[a]]` or `G2[global_to_local[a]]` depending on `is_v1(a)` outside of the inner loop and use it here without a conditional.



_Suggestion:_
```cpp
      if (in_v1(j))
```
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 284 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZJ-Z0J_8KGD0wpwG:-N_bZJ-Z0J_8KGD0wpwH:b-1ndhgd) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L284)):*
> ```cpp
>   for (int v : v2_ordering) {
>     result.emplace_back(v2_to_global[v]);
>   }
> ```

nit: If you use PartialPermutation this can be written concisely as:



_Suggestion:_
```cpp
  std::vector<int> result;
  result.reserve(n);
  v1_permutation.ApplyInverse(v1_ordering, &result);
  v2_permutation.ApplyInverse(v2_ordering, &result);
```

xuchenhan-tri (commented):
@joemasterjohn I addressed the comments. PTAL, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-:-N_eOqd-0aQd2Et12EQX:b-8c3q42)* status: 5 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bY7mZ7htkdJ8d6ZaF:-N_eIcBw1ipLiQwlZUtr:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L246)):*

nit: explicit reutrn type not needed.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 248 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bgzqU-5tpQzOYPSvr:-N_eKz6S-2uF43bFxwTo:b-dtkds3) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L248)):*

nit: Here you could use 2 `contact_solvers::internal::PartialPermutation`s instead of plain old vectors to give some semantics to these mappings and shorten the code below.
</blockquote></details>

Done, thanks!
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 253 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bcgyG4zngmq1FvdFH:-N_eL-QO6ZUdhSr8v7nl:bghc58q) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L253)):*

nit: some STL tricks you can use here:
</blockquote></details>

I'm going to pass on this one. The stl trick is very neat but I feel like the dumbed down version is more readable.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 260 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZmIDA3JUgbs6liNS:-N_eNAEw6c1RBA0CGb1g:b-becr91) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L260)):*

nit: this local index access can be moved out of the inner loop. See my comment below about why `a` is always the min index.
</blockquote></details>

I agree that this is true, but I think it's a lot more clear what the intention is with the add `std::max/min`.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_b_DKpEGCfnPh8I_i8:-N_eNJZo34rs7H66OG0X:b-becr91) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L261)):*

BTW the global to local mapping is stable so global_to_local[a] < global_to_local[b] for all b in G[a] because a < b for all b in G[a] (when `in_v1(a) == in_v1(b)`, so no need to do this min/max. `j = global_to_local`
</blockquote></details>

I agree that this is true, but I think it's a lot more clear what the intention is with the add `std::max/min`.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 262 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_baAsB4IcGLkgCOl_S:-N_eNKGqDnbON-dn9TNi:b-61eduo) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L262)):*

I think this is incorrect, shouldn't it be whether `j` is in v1 or not? If you take out the min/max for i/j above you can just create a reference to `G1[global_to_local[a]]` or `G2[global_to_local[a]]` depending on `is_v1(a)` outside of the inner loop and use it here without a conditional.
</blockquote></details>

I think `in_v1(b)` is correct. 

`in_v1()` operates on unpermuted indices (a and b, instead of i and j).
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 284 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZJ-Z0J_8KGD0wpwG:-N_eOINAAf5gV7j9w55o:b-jl3u6o) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L284)):*

nit: If you use PartialPermutation this can be written concisely as:
</blockquote></details>

I think this is not quite right. PTAL and the change and see if you agree.

joemasterjohn (commented):
:lgtm: thanks! +@rpoyner-tri for platform review tomorrow, please.

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-:-N_etquU2XG8_ia9izjR:bch0nem)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform)

___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 253 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bcgyG4zngmq1FvdFH:-N_etmoQ-5Vin0vTCG-o:brxp9sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L253)):*

I'm going to pass on this one. The stl trick is very neat but I feel like the dumbed down version is more readable.
</blockquote></details>

Sounds good.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 260 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZmIDA3JUgbs6liNS:-N_eti3X6Lc0nOFod1Kc:brxp9sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L260)):*

I agree that this is true, but I think it's a lot more clear what the intention is with the add `std::max/min`.
</blockquote></details>

Sounds good.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_b_DKpEGCfnPh8I_i8:-N_etak5Em09GeIY9xnE:brxp9sq) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L261)):*

I agree that this is true, but I think it's a lot more clear what the intention is with the add `std::max/min`.
</blockquote></details>

Sounds good.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 262 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_baAsB4IcGLkgCOl_S:-N_eVic711IGLA3fAvvL:blfx2uu) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L262)):*

I think `in_v1(b)` is correct. 

`in_v1()` operates on unpermuted indices (a and b, instead of i and j).
</blockquote></details>

Sorry my bad, you're absolutely right.
___
*[`multibody/contact_solvers/minimum_degree_ordering.cc` line 284 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-N_bZJ-Z0J_8KGD0wpwG:-N_eqwPwDcT031M4T_KC:b-b7qw9j) ([raw file](https://github.com/RobotLocomotion/drake/blob/75e6c2048e0485a94b80d6e2acd275bbb69ef46b/multibody/contact_solvers/minimum_degree_ordering.cc#L284)):*

I think this is not quite right. PTAL and the change and see if you agree.
</blockquote></details>

Ok you're right, ApplyInverse is the wrong thing. The ordering being the important thing here :man_facepalming:

rpoyner-tri (commented):
:lgtm:

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19764#-:-N_irbAaDA-UPx3Xl-8q:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),joemasterjohn

PR #19774, opened 2023-07-13
[mkdoc] Test comments between docstring and symbol

EricCousineau-TRI:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19774)

### DIFF
--- a/tools/workspace/pybind11/test/sample_header.h
+++ b/tools/workspace/pybind11/test/sample_header.h
@@ -24,15 +24,18 @@
 /// @def PREPROCESSOR_DEFINITION
 /// Preprocessor definitions are ignored. In nibh mauris cursus mattis
 /// molestie a. Non arcu risus quis varius quam quisque id.
+// (Test) Ensure this comment does not ruin docstring association.
 #define PREPROCESSOR_DEFINITION "Nisl purus in mollis nunc sed id."
 
 /// Root-level symbol. Magna fermentum iaculis eu non diam phasellus
 /// vestibulum.
+// (Test) Ensure this comment does not ruin docstring association.
 struct RootLevelSymbol {};
 
 /// @namespace drake
 /// Namespaces are ignored. Enim blandit volutpat maecenas volutpat blandit. Eu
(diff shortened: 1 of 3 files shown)


sherm1 (commented):
Thanks for the confirmation, Eric.
I don't understand mkdoc well enough to get what you're doing here though. Is the "expected" stuff what mkdoc _should_ have generated but doesn't with the `//` comments in the way?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19774#-:-N_FW36R75hHmWEW5bs9:b-ahg130)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

EricCousineau-TRI (commented):
Sorry for not clarifying, but yup, exactly.

This PR is showing that injecting `//` comments between the `///`documentation comment does not cause `mkdoc` (`clang.cindex`) to lose track of the association between the documentation comment and the symbol.

The fact that the regression file (`sample\_header\_documentation.expected.h`) has been updated with only line number changes and the regression test (`sample_header_documentation_py_test`) passes across all platforms (including Mac) shows that we should be OK for the injected `//` comments.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19774#-:-N_K7uPR0_eewb2WX2vK:b-aif4dt)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @EricCousineau-TRI)

jwnimmer-tri:
> ... should be OK for the injected // comments.

It shows that _these exact comments_ do not cause a problem.  It does not show that multi-like TODOs are safe, or that some particular Doxygen comment-opener formatting might also go awry.  Even with this regression test, it will still be incumbent upon developers to generate the documentation and make sure it looks good.  At the end of the day, there is no way around that.  That's why we buy them all real computers so that can generate the docs in just a few quick minutes.

EricCousineau-TRI:
sounds good to me, will go ahead and close this

PR #19785, opened 2023-07-14
Add GraphOfConvexSets::SolveConvexRestriction()

RussTedrake:
+@TobiaMarcucci for feature review, please
cc @wrangelvid

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -670,7 +670,11 @@ void DefineGeometryOptimization(py::module m) {
                 &GraphOfConvexSets::SolveShortestPath),
             py::arg("source"), py::arg("target"),
             py::arg("options") = GraphOfConvexSetsOptions(),
-            cls_doc.SolveShortestPath.doc_by_reference);
+            cls_doc.SolveShortestPath.doc_by_reference)
+        .def("SolveConvexRestriction",
+            &GraphOfConvexSets::SolveConvexRestriction, py::arg("active_edges"),
+            py::arg("options") = GraphOfConvexSetsOptions(),
+            cls_doc.SolveConvexRestriction.doc);
   }
 
   // NOLINTNEXTLINE(readability/fn_size)
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -541,6 +541,11 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertIsInstance(spp.SolveShortestPath(
             source=source, target=target, options=options),
             MathematicalProgramResult)
+        self.assertIsInstance(
+            spp.SolveConvexRestriction(active_edges={edge0.id(),
+                                                     edge1.id()},
+                                       options=options),
+            MathematicalProgramResult)
 
         self.assertIn("source", spp.GetGraphvizString(
             result=result, show_slacks=True, precision=2, scientific=False))
--- a/geometry/optimization/graph_of_convex_sets.cc
+++ b/geometry/optimization/graph_of_convex_sets.cc
@@ -1197,6 +1197,80 @@ MathematicalProgramResult GraphOfConvexSets::SolveShortestPath(
   return SolveShortestPath(source.id(), target.id(), options);
 }
 
+MathematicalProgramResult GraphOfConvexSets::SolveConvexRestriction(
+    const std::set<EdgeId>& active_edge_ids,
+    const GraphOfConvexSetsOptions& options) const {
+  MathematicalProgram prog;
+
+  std::set<VertexId> vertices;
+  for (const auto& edge_id : active_edge_ids) {
+    if (edges_.count(edge_id) == 0) {
+      throw std::runtime_error(
+          fmt::format("EdgeID {} is not in the graph.", edge_id));
+    }
+    const Edge* e = edges_.at(edge_id).get();
+    vertices.emplace(e->u().id());
+    vertices.emplace(e->v().id());
+  }
+
+  for (const auto& vertex_id : vertices) {
+    const Vertex* v = vertices_.at(vertex_id).get();
+    prog.AddDecisionVariables(v->x());
+    v->set().AddPointInSetConstraints(&prog, v->x());
+
+    // Vertex costs.
+    for (const Binding<Cost>& b : v->costs_) {
+      prog.AddCost(b);
+    }
+    // Vertex constraints.
+    for (const Binding<Constraint>& b : v->constraints_) {
+      prog.AddConstraint(b);
+    }
+  }
+
+  for (const auto& edge_id : active_edge_ids) {
+    const Edge* e = edges_.at(edge_id).get();
+
+    // Edge costs.
+    for (const Binding<Cost>& b : e->costs_) {
+      prog.AddCost(b);
+    }
+    // Edge constraints.
+    for (const Binding<Constraint>& b : e->constraints_) {
+      prog.AddConstraint(b);
+    }
+  }
+
+  MathematicalProgramResult result = Solve(prog, options, false);
+
+  // Add any excluded vertices to the result.
+  int num_excluded_vars = 0;
+  std::vector<const Vertex*> excluded_vertices;
+  for (const auto& pair : vertices_) {
+    const VertexId v_id = pair.first;
+    if (vertices.count(v_id) == 0) {
+      const Vertex* v = pair.second.get();
+      num_excluded_vars += v->x().size();
+      excluded_vertices.emplace_back(v);
+    }
+  }
+  int count = result.get_x_val().size();
+  Eigen::VectorXd x_val(count + num_excluded_vars);
+  x_val.head(count) = result.get_x_val();
+  std::unordered_map<symbolic::Variable::Id, int> decision_variable_index =
+      prog.decision_variable_index();
+  for (const Vertex* v : excluded_vertices) {
+    for (int i = 0; i < v->x().size(); ++i) {
+      decision_variable_index.emplace(v->x()[i].get_id(), count);
+      x_val[count++] = std::numeric_limits<double>::quiet_NaN();
+    }
+  }
+  result.set_decision_variable_index(decision_variable_index);
+  result.set_x_val(x_val);
+
+  return result;
+}
+
 }  // namespace optimization
 }  // namespace geometry
 }  // namespace drake
--- a/geometry/optimization/graph_of_convex_sets.h
+++ b/geometry/optimization/graph_of_convex_sets.h
@@ -75,8 +75,8 @@ struct GraphOfConvexSetsOptions {
 };
 
 /**
-GraphOfConvexSets implements the design pattern and optimization problems first
-introduced in the paper "Shortest Paths in Graphs of Convex Sets".
+GraphOfConvexSets (GCS) implements the design pattern and optimization problems
+first introduced in the paper "Shortest Paths in Graphs of Convex Sets".
 
 "Shortest Paths in Graphs of Convex Sets" by Tobia Marcucci, Jack Umenberger,
 Pablo A. Parrilo, Russ Tedrake. https://arxiv.org/abs/2101.11565
@@ -87,9 +87,9 @@ Each vertex in the graph is associated with a convex set over continuous
 variables, edges in the graph contain convex costs and constraints on these
 continuous variables.  We can then formulate optimization problems over this
 graph, such as the shortest path problem where each visit to a vertex also
-corresponds to selecting an element from the convex set subject to the costs and
-constraints.  Behind the scenes, we construct efficient mixed-integer convex
-transcriptions of the graph problem using MathematicalProgram.
+corresponds to selecting an element from the convex set subject to the costs
+and constraints.  Behind the scenes, we construct efficient mixed-integer
+convex transcriptions of the graph problem using MathematicalProgram.
 
 Design note: This class avoids providing any direct access to the
 MathematicalProgram that it constructs nor to the decision variables /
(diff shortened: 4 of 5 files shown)


RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)* status: 2 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 515 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JK3Y0EMDRYPwmJ0NU:-N_JK3Y13VK5iCwoOYQ_:b-onigp4) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.h#L515)):*
> ```objc
>   vertices associated with at least one of the active edges. Decision variables
>   for all other convex sets will be set to NaN. */
>   solvers::MathematicalProgramResult SolveWithActiveEdges(
> ```

btw -- do we like the name?  i'm open to suggestions.
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 1246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JJjFu-9_OFyHH4fOM:-N_JJjFu-9_OFyHH4fON:b-huew1) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.cc#L1246)):*
> ```cpp
>   MathematicalProgramResult result = Solve(prog, options, false);
> 
>   // Add any excluded vertices to the result.
> ```

btw -- i debated whether to add these or simply leave them out.  (or potentially to have an optional argument `bool set_excluded_vertex_results=true`.  We'll pay a very small price for this if we're calling the method repeatedly during rounding.

TobiaMarcucci (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)* status: 3 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 515 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JK3Y0EMDRYPwmJ0NU:-N_KMDWN0tsAf2iERNX3:bpzn7jd) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.h#L515)):*

btw -- do we like the name?  i'm open to suggestions.
</blockquote></details>

In my current work I'm trying to think about these as subgraphs (the SPP is looking for a path subgraph of the original graph, the MST is looking for a spanning tree, etc). In this sense, an alternative name might be SolveOnSubgraph or something like that. Note that the subgraph is uniquely defined by the edges -- no need to also specify the vertices, except for weird subgraphs where one has isolated vertices. This function should work fine if one passes, e.g., the edges of a spanning tree or a tour, correct?
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 1246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JJjFu-9_OFyHH4fOM:-N_KOLeF-IKHNhDdh2EI:bmaa2tp) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.cc#L1246)):*

btw -- i debated whether to add these or simply leave them out.  (or potentially to have an optional argument `bool set_excluded_vertex_results=true`.  We'll pay a very small price for this if we're calling the method repeatedly during rounding.
</blockquote></details>

This is setting the position of the other vertices to NaN? I'm happy either way, I don't see any particular disadvantage of setting also the the values for the other vertices.
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1775 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_KPIgS35popC8vnPID:-N_KPIgS35popC8vnPIE:b-bkospv) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/test/graph_of_convex_sets_test.cc#L1775)):*
> ```cpp
>   // Test that solving with the known shortest path returns consistent results.
>   // The solution to this problem is not actually unique, so we cannot demand
>   // that they return exactly the same results.
> ```

This is the case with the cost Euclidean distance *not* squared, right? If yes, I agree that the solution is not unique, since if you look at the figure in the paper you see that the trajectory is stepping in the interior of a set, and moving that point along the line connecting the previous and the next vertex does not change the cost. Maybe a more robust test is with the Euclidean distance squared? In that case the minimizer should always be unique.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)* status: 3 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 515 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JK3Y0EMDRYPwmJ0NU:-N_L7OKi2slqe-koteE3:b-nrvhcx) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.h#L515)):*

In my current work I'm trying to think about these as subgraphs (the SPP is looking for a path subgraph of the original graph, the MST is looking for a spanning tree, etc). In this sense, an alternative name might be SolveOnSubgraph or something like that. Note that the subgraph is uniquely defined by the edges -- no need to also specify the vertices, except for weird subgraphs where one has isolated vertices. This function should work fine if one passes, e.g., the edges of a spanning tree or a tour, correct?
</blockquote></details>

Yes, I believe it should work correctly if one passes the edges of a spanning tree or tour.  But I don't love SolveOnSubgraph, because that doesn't make it clear that all of the binaries are fixed and the problem becomes convex.
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 1246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JJjFu-9_OFyHH4fOM:-N_L7e0t2pLnDZRpD1Zx:bhm0xui) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.cc#L1246)):*

This is setting the position of the other vertices to NaN? I'm happy either way, I don't see any particular disadvantage of setting also the the values for the other vertices.
</blockquote></details>

My only concern is that it adds an extremely modest amount of computational cost to what I envision might be a frequently called subroutine.  My thought was that it's better to have it and remove it if/when it shows up in a profiler.  

If we don't think it adds much value, (and the absence of those values in the result would not be unclear to the user) then i would prefer to simply remove it now.
___
*[`geometry/optimization/test/graph_of_convex_sets_test.cc` line 1775 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_KPIgS35popC8vnPID:-N_LDADv6U7TNlpcoLRS:b-hlzxed) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/test/graph_of_convex_sets_test.cc#L1775)):*

This is the case with the cost Euclidean distance *not* squared, right? If yes, I agree that the solution is not unique, since if you look at the figure in the paper you see that the trajectory is stepping in the interior of a set, and moving that point along the line connecting the previous and the next vertex does not change the cost. Maybe a more robust test is with the Euclidean distance squared? In that case the minimizer should always be unique.
</blockquote></details>

Done.  I went ahead and switched the the squared norm cost.

TobiaMarcucci (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)* status: 2 unresolved discussions, LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 515 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JK3Y0EMDRYPwmJ0NU:-N_LPrbJ05QByy1m-rNC:b-xmyrk2) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.h#L515)):*

Yes, I believe it should work correctly if one passes the edges of a spanning tree or tour.  But I don't love SolveOnSubgraph, because that doesn't make it clear that all of the binaries are fixed and the problem becomes convex.
</blockquote></details>

As by our discussion, one other term could be `SolveConvexRestriction`
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 1246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JJjFu-9_OFyHH4fOM:-N_LQ4p1C-b-N2E463yG:b4kcyo4) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.cc#L1246)):*

My only concern is that it adds an extremely modest amount of computational cost to what I envision might be a frequently called subroutine.  My thought was that it's better to have it and remove it if/when it shows up in a profiler.  

If we don't think it adds much value, (and the absence of those values in the result would not be unclear to the user) then i would prefer to simply remove it now.
</blockquote></details>

I think the first make sense. I don't think it'll ever show up in a profiler, but it might save some minor headaches/bugs.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785)* status: LGTM missing from assignee TobiaMarcucci, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

___
*[`geometry/optimization/graph_of_convex_sets.h` line 515 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JK3Y0EMDRYPwmJ0NU:-N_LZHmZ3j5yc591NJ5O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.h#L515)):*

As by our discussion, one other term could be `SolveConvexRestriction`
</blockquote></details>

Done.
___
*[`geometry/optimization/graph_of_convex_sets.cc` line 1246 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_JJjFu-9_OFyHH4fOM:-N_LZJbx0KKBNoWqLgCo:b-g4p7ox) ([raw file](https://github.com/RobotLocomotion/drake/blob/83a3f239ae33de7165e3344352a74003fa4aa2bc/geometry/optimization/graph_of_convex_sets.cc#L1246)):*

I think the first make sense. I don't think it'll ever show up in a profiler, but it might save some minor headaches/bugs.
</blockquote></details>

Done.  Agreed.

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-:-N_LZc9vAnnxaj6MGN8p:bnfp4nl)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

xuchenhan-tri (commented):
+@xuchenhan-tri 
platform :lgtm: 
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-:-N_M0lbi9BOQSYAFQ1zT:ba06zzb)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,xuchenhan-tri(platform)

jwnimmer-tri:
*[`geometry/optimization/graph_of_convex_sets.cc` line 1269 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19785#-N_k-BBe-4XHmRZZkKBQ:-N_k-BBf5I3D_hYqz_pH:b-gx0k69) ([raw file](https://github.com/RobotLocomotion/drake/blob/1ff75b28bba4ab4278e377df5ca1bef59c684ff8/geometry/optimization/graph_of_convex_sets.cc#L1269)):*
> ```cpp
>   }
>   result.set_decision_variable_index(decision_variable_index);
>   result.set_x_val(x_val);
> ```

BTW If the `result` contained suboptimal solutions, I think they would be the wrong size now?  Is that relevant / do we care?

RussTedrake:
Can you say more?  The results returned are different, I agree.  But I don't understand anything to be the wrong size?  (are you saying it would error)?

jwnimmer-tri:
I have in mind the invariant that `result.x_val_.size() == result.suboptimal_x_val_[*].size()` for all suboptimal results.

The code in this PR nan-padded the `x_val` for the unused vertices, but did not nan-pad the `suboptimal_x_val_`.

jwnimmer-tri:
Probably the best fix would be to add a function on `MathematicalProgramResult` that takes in a `VectorXDecisionVariable` and appends it to the index, filling in the new `x_val` positions with NaNs across the board.  Trying to get those details right at GCS call site violates encapsulation.

RussTedrake:
i see.  I believe that suboptimal_x_val should be empty here .... since this codepath is always solving the convex optimization (the suboptimal pools are only for mixed-integer problems)

PR #19791, opened 2023-07-15
[pybind] split solvers_py_mathematicalprogram.

hongkai-dai:
Put the Evaluator/Cost/Constraint/Binding into solvers_py_evaluator.cc

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)


hongkai-dai (commented):
@jwnimmer-tri and @EricCousineau-TRI I remember we had the discussion before to split the giant `solvers_py_mathematicalprogram.cc`, and here is my first attempt.

May I ask you for help on this one? Currently I am not sure how to handle `PyFunctionConstraint` and `PyFunctionCost`. I moved them to `solvers_py_evaluator.cc` (I think it makes more sense as that's where all other cost/constraint live). But in `solvers_py_mathematicalprogram.cc` it uses these two classes in binding `prog.AddCost(PyFunctionCost) and `prog.AddConstraint(PyFunctionConstraint)`. Is there any way I can make `solvers_py_mathematicalprogram.cc` to be aware of these two classes?

One approach is to put these two classes in a header file and include that header file in `solvers_py_mathematicalprogram.cc`. Another approach is to move `PyFunctionConstraint` and `PyFunctionCost` back to `solvers_py_mathematicalprogram.cc`, but I wonder if there is a better approach.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-:-N_PdvFPBTpdqDpqZOTp:b6l9cd)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

jwnimmer-tri:
The they are used by two different cc files, then yes -- you would need to put them in a header file.

However, from a quick skim I don't see any reason to place them in two different files.  If we keep them in the same file as MathematicalProgram, then everything is happy.

I can understand the aspiration to move them into `evaluators`, but that move seems like more worth than it's worth.

hongkai-dai:
Got it, I will close this PR then.

jwnimmer-tri:
I do think we should split the C++ evaluators into their own file.

> However, from a quick skim I don't see any reason to place them in two different files.

Here I was only saying that the `PyFunctionCost` and `PyFunctionConstraint` should not be split out.

hongkai-dai:
> Here I was only saying that the PyFunctionCost and PyFunctionConstraint should not be split out.

But `PyFunctionConstraint` is also used in `BindEvaluatorsAndBindings` function in https://github.com/RobotLocomotion/drake/blob/4aa538dbfae9e53753dabf3b121c0a0c2c405701/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc#L1676-L1686, which defines the binding for all other cost/constraints.

jwnimmer-tri:
Other than clarity / consistency, there's no need for those 10 lines to live inside that function though, right?  We can keep that binding definition in the original file.  We'll just want some comment(s) in the `solvers_py_evaluator.cc` to clarify that it only holds the bound C++ costs and constraints; the python shims still live in the other file.

hongkai-dai (commented):
Done. Thanks Jeremy

+@EricCousineau-TRI for feature review please, thanks! 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-:-N_a-ekA7i6SC6bjIhoV:b90dlpd)* status: LGTM missing from assignee EricCousineau-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

EricCousineau-TRI (commented):
:lgtm: feature, nice!
Briefly confirmed move-only with:
```
git diff FETCH_HEAD~ --color-moved=zebra --ignore-space-change
git fetch upstream pull/19791/head
```

+@rpoyner-tri for platform, please!

> Reviewed 2 of 5 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-:-N_ai_ax2wl9V2EVkyU4:b2fzojb)* status: LGTM missing from assignee rpoyner-tri(platform)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform)

___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 1 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_dKXW52FxCAywoQCPT:-N_dKXW52FxCAywoQCPU:b-s6dco1) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L1)):*
> ```cpp
> #include <memory>
> ```

nit The filename is "evaluator" singular but the function name is "DefineSolversEvaluators" plural.  The two should be consistent. (I don't think it matter which direction you resolve it.)
___
*[`bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc` line 1545 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_dL1HHCsRFJwNjbqZu:-N_dL1HHCsRFJwNjbqZv:b-h7zg9a) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_mathematicalprogram.cc#L1545)):*
> ```cpp
>   BindMathematicalProgram(m);
>   BindFreeFunctions(m);
>   BindPyFunctionConstraint(m);
> ```

nit I think with the bindings we have so far it won't bite us, but we should probably fix it anyway: the `PyFunctionConstraint` is a lower dependency layer than `MathematicalProgram`, so it should be bound first.  This call should be moved to be the first line of the function (where it used to be).

rpoyner-tri (commented):
:lgtm: minor stuff

> Reviewed 2 of 5 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-:-N_dgafSEEwGWmcuiP3g:bo23mka)* status: 4 unresolved discussions

___
*[`bindings/pydrake/solvers/BUILD.bazel` line 241 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_deifB8hJeuej3KL_X:-N_deifB8hJeuej3KL_Y:b-5h73gc) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/BUILD.bazel#L241)):*
> ```Starlark
>     deps = [
>         ":solvers",
>         "//bindings/pydrake:math_py",
> ```

The three "//bindings/blah" dependencies can be removed here.
___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_d_X2kAndbjbfB_2FJ:-N_d_X2m9MjRBg6OVMhW:b-qv50y) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L47)):*
> ```cpp
> namespace {
> 
> /*
> ```

This doc comment was stale before the code move and is stale now. Maybe we would rather fix it in a separate follow-up PR?

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 1 unresolved discussion

___
*[`bindings/pydrake/solvers/BUILD.bazel` line 241 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_deifB8hJeuej3KL_X:-N_doBbP39MRqE0lfClM:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/BUILD.bazel#L241)):*

The three "//bindings/blah" dependencies can be removed here.
</blockquote></details>

Done.
___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 1 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_dKXW52FxCAywoQCPT:-N_dp-zLByBy3v2s6TBe:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L1)):*

nit The filename is "evaluator" singular but the function name is "DefineSolversEvaluators" plural.  The two should be consistent. (I don't think it matter which direction you resolve it.)
</blockquote></details>

Done.
___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_d_X2kAndbjbfB_2FJ:-N_dpILfDqEjREsuU0_y:b-ogrims) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L47)):*

This doc comment was stale before the code move and is stale now. Maybe we would rather fix it in a separate follow-up PR?
</blockquote></details>

Done. I modified the doc.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),EricCousineau-TRI(platform)

___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_d_X2kAndbjbfB_2FJ:-N_dvwzTD0J0A_r6KvYv:buh7fkj) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L47)):*

Done. I modified the doc.
</blockquote></details>

Any chance I can get you to document the template param (`@tparam`) `C`? It seems like it is the interesting part.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),EricCousineau-TRI(platform)

___
*[`bindings/pydrake/solvers/solvers_py_evaluator.cc` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_d_X2kAndbjbfB_2FJ:-N_e7N6Q5q9_lxuD0B-p:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/396e5b290e0f72525bce4455713391dad3a55dc2/bindings/pydrake/solvers/solvers_py_evaluator.cc#L47)):*

Any chance I can get you to document the template param (`@tparam`) `C`? It seems like it is the interesting part.
</blockquote></details>

Done.

rpoyner-tri (commented):
> Reviewed 3 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),EricCousineau-TRI(platform)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 1 unresolved discussion

___
*[`bindings/pydrake/solvers/solvers_py_evaluators.cc` line 49 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_eAZE83_QreBt5BZhV:-N_eAZE83_QreBt5BZhW:bgs24p0) ([raw file](https://github.com/RobotLocomotion/drake/blob/cb44ac94b5fe12302adcb36d7c690e6e7ac184fe/bindings/pydrake/solvers/solvers_py_evaluators.cc#L49)):*
> ```cpp
> /*
>  * Register a Binding template.
>  * @tparam C A EvaluatorBase (or its child) class.
> ```

BTW typo



_Suggestion:_
```cpp
An EvaluatorBase
```

rpoyner-tri (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 2 unresolved discussions

___
*[`bindings/pydrake/solvers/test/evaluators_test.py` line 4 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_ebqeY9qelIrYtE9dp:-N_ebqeY9qelIrYtE9dq:b-18zqnz) ([raw file](https://github.com/RobotLocomotion/drake/blob/8f865283bd0b3e2b3aecbfd3a3b484fc6ceaf32f/bindings/pydrake/solvers/test/evaluators_test.py#L4)):*
> ```Python
> 
> import numpy as np
> import scipy
> ```

This line definitely makes the problem easier to debug, but I don't think it is ultimately necessary. As far as I can tell, Drake doesn't even require scipy any more, but Nimmer has written an elaborate set of "fake" scipy code to allow pybind11 to "call scipy" when dealing with the py-bound versions of Eigen::Sparse objects.

What we need (I think) is to add back the `scipy_stub_py` dependency in the bazel file, something like this:
```
diff --git a/bindings/pydrake/solvers/BUILD.bazel b/bindings/pydrake/solvers/BUILD.bazel
index b0bbe9b87b..5e94d1406e 100644
--- a/bindings/pydrake/solvers/BUILD.bazel
+++ b/bindings/pydrake/solvers/BUILD.bazel
@@ -238,6 +238,7 @@ drake_py_unittest(
     name = "evaluators_test",
     deps = [
         ":solvers",
+        "//bindings/pydrake/common/test_utilities:scipy_stub_py",
     ],
 )
```

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 2 unresolved discussions

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: 1 unresolved discussion

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19791)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),EricCousineau-TRI(platform)

___
*[`bindings/pydrake/solvers/test/evaluators_test.py` line 4 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19791#-N_ebqeY9qelIrYtE9dp:-N_et1p305XvKEngcPPC:b7bgsuk) ([raw file](https://github.com/RobotLocomotion/drake/blob/8f865283bd0b3e2b3aecbfd3a3b484fc6ceaf32f/bindings/pydrake/solvers/test/evaluators_test.py#L4)):*

This line definitely makes the problem easier to debug, but I don't think it is ultimately necessary. As far as I can tell, Drake doesn't even require scipy any more, but Nimmer has written an elaborate set of "fake" scipy code to allow pybind11 to "call scipy" when dealing with the py-bound versions of Eigen::Sparse objects.

What we need (I think) is to add back the `scipy_stub_py` dependency in the bazel file, something like this:
```
diff --git a/bindings/pydrake/solvers/BUILD.bazel b/bindings/pydrake/solvers/BUILD.bazel
index b0bbe9b87b..5e94d1406e 100644
--- a/bindings/pydrake/solvers/BUILD.bazel
+++ b/bindings/pydrake/solvers/BUILD.bazel
@@ -238,6 +238,7 @@ drake_py_unittest(
     name = "evaluators_test",
     deps = [
         ":solvers",
+        "//bindings/pydrake/common/test_utilities:scipy_stub_py",
     ],
 )
```
</blockquote></details>

Thanks Rico!

Issue #18726, opened 2023-02-01
Jammy CI tests failing with: "X Error of failed request:  BadShmSeg (invalid shared segment parameter)"


BetsyMcPhail:
### What happened?

Some tests begain failing in Jammy unprovisioned jobs on January 30th. For example, https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/linux-jammy-unprovisioned-gcc-bazel-nightly-release/201/

The failing tests are:

- `//bindings/pydrake/geometry:py/scene_graph_test`
- `//bindings/pydrake/visualization:py/model_visualizer_test`
- `//bindings/pydrake/visualization:py/video_test`
- `//examples/hardware_sim:py/hardware_sim_cc_test`
- `//examples/hardware_sim:py/hardware_sim_py_test`
- `//geometry/render_gltf_client:py/acceptance_test`
- `//geometry/render_gltf_client:py/integration_test`
- `//tutorials:py/rendering_multibody_plant_test`

They all fail with the message:

```
....X Error of failed request:  BadShmSeg (invalid shared segment parameter)
  Major opcode of failed request:  130 (MIT-SHM)
  Minor opcode of failed request:  3 (X_ShmPutImage)
  Segment id in failed request:  0x200005
  Serial number of failed request:  46
  Current serial number in output stream:  47
```

### What operating system are you using?

Ubuntu 22.04

BetsyMcPhail:
See slack discussion: https://drakedevelopers.slack.com/archives/C270MN28G/p1675172012911669

BetsyMcPhail:
After comparing package versions on a working job and a failing job, we found that updating `libglapi-mesa`, which updates `libegl-mesa0`, `libgbm1`,  `libgl1-mesa-dri`, and `libglx-mesa0` and installs `libllvm15`, causes the failure.

BetsyMcPhail:
We don't directly install any of these gl/mesa packages, instead they are installed as dependencies of `xvfb` as part of the [CI setup](https://github.com/RobotLocomotion/drake-ci/blob/1ba3cf4b53222f602034968651418c0e5b3949c7/setup/ubuntu/install_prereqs#L71). Therefore, none of the gl/mesa packgaes or `xvfb` are installed on the unprovisioned images. If we simply try to `apt-mark hold` the gl/mesa packages, `xvfb` can't be installed. 

I ran the following commands to update the **unprovisioned** Jammy image:

```
[...]

PR #19803, opened 2023-07-18
Add SpatialInertia::SolidCylinderWithMassAboutEnd()

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803)

### DIFF
--- a/bindings/pydrake/multibody/BUILD.bazel
+++ b/bindings/pydrake/multibody/BUILD.bazel
@@ -327,8 +327,8 @@ drake_py_library(
 
 install(
     name = "install",
-    targets = PY_LIBRARIES + [":all_py"],
     py_dest = PACKAGE_INFO.py_dest,
+    targets = PY_LIBRARIES + [":all_py"],
     deps = get_drake_py_installs(PY_LIBRARIES_WITH_INSTALL) + [
         ":install_wrapper_scripts",
     ],
@@ -406,6 +406,7 @@ drake_py_unittest(
         ":parsing_py",
         ":plant_py",
         ":tree_py",
+        "//bindings/pydrake/common/test_utilities:deprecation_py",
         "//bindings/pydrake/common/test_utilities:numpy_compare_py",
         "//bindings/pydrake/common/test_utilities:pickle_compare_py",
         "//bindings/pydrake/common/test_utilities:scipy_stub_py",
--- a/bindings/pydrake/multibody/BUILD.bazel
+++ b/bindings/pydrake/multibody/BUILD.bazel
@@ -327,8 +327,8 @@ drake_py_library(
 
 install(
     name = "install",
-    py_dest = PACKAGE_INFO.py_dest,
     targets = PY_LIBRARIES + [":all_py"],
+    py_dest = PACKAGE_INFO.py_dest,
     deps = get_drake_py_installs(PY_LIBRARIES_WITH_INSTALL) + [
         ":install_wrapper_scripts",
     ],
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -97,6 +97,7 @@ from pydrake.common import FindResourceOrThrow
 from pydrake.common.deprecation import install_numpy_warning_filters
 from pydrake.common.eigen_geometry import Quaternion_
 from pydrake.common.test_utilities import numpy_compare
+from pydrake.common.test_utilities.deprecation import catch_drake_warnings
 from pydrake.common.test_utilities.pickle_compare import assert_pickle
 from pydrake.common.value import AbstractValue, Value
 from pydrake.geometry import (
@@ -737,6 +738,9 @@ class TestPlant(unittest.TestCase):
         self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(
             radius=0.1, length=0.4, unit_vector=[0, 0, 1]),
                               UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(r=1, L=4),
+                                  UnitInertia)
         self.assertIsInstance(
             UnitInertia.AxiallySymmetric(J=1, K=2, b_E=[1, 2, 3]), UnitInertia)
         self.assertIsInstance(UnitInertia.StraightLine(K=1.5, b_E=[1, 2, 3]),
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -734,7 +734,8 @@ class TestPlant(unittest.TestCase):
         self.assertIsInstance(UnitInertia.SolidCapsule(r=1, L=2,
                                                        unit_vector=[0, 0, 1]),
                               UnitInertia)
-        self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(r=1, L=2),
+        self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(
+            radius=0.1, length=0.4, unit_vector=[0, 0, 1]),
                               UnitInertia)
         self.assertIsInstance(
             UnitInertia.AxiallySymmetric(J=1, K=2, b_E=[1, 2, 3]), UnitInertia)
@@ -773,6 +774,8 @@ class TestPlant(unittest.TestCase):
             mass=0.123, radius=0.1, length=0.4, unit_vector=[0, 0, 1])
         SpatialInertia.SolidCylinderWithDensityAboutEnd(
             density=1000, radius=0.1, length=0.4, unit_vector=[0, 0, 1])
+        SpatialInertia.SolidCylinderWithMassAboutEnd(
+            mass=0.123, radius=0.1, length=0.4, unit_vector=[0, 0, 1])
         SpatialInertia.ThinRodWithMass(
             mass=2, length=0.3, unit_vector=[0, 0, 1])
         SpatialInertia.ThinRodWithMassAboutEnd(
--- a/bindings/pydrake/multibody/tree_py.cc
+++ b/bindings/pydrake/multibody/tree_py.cc
@@ -1158,7 +1158,7 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def_static("SolidCapsule", &Class::SolidCapsule, py::arg("r"),
             py::arg("L"), py::arg("unit_vector") = Vector3<T>::UnitZ().eval(),
             cls_doc.SolidCapsule.doc)
-        // TODO(2023-10-01) Remove overload wrapping when deprecation complete.
+        // TODO(2023-11-01) Remove overload wrapping when deprecation complete.
         .def_static("SolidCylinderAboutEnd",
             overload_cast_explicit<Class, const T&, const T&,
                 const Vector3<T>&>(&Class::SolidCylinderAboutEnd),
--- a/bindings/pydrake/multibody/tree_py.cc
+++ b/bindings/pydrake/multibody/tree_py.cc
@@ -1159,7 +1159,8 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py::arg("L"), py::arg("unit_vector") = Vector3<T>::UnitZ().eval(),
             cls_doc.SolidCapsule.doc)
         .def_static("SolidCylinderAboutEnd", &Class::SolidCylinderAboutEnd,
-            py::arg("r"), py::arg("L"), cls_doc.SolidCylinderAboutEnd.doc)
+            py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
+            cls_doc.SolidCylinderAboutEnd.doc)
         .def_static("AxiallySymmetric", &Class::AxiallySymmetric, py::arg("J"),
             py::arg("K"), py::arg("b_E"), cls_doc.AxiallySymmetric.doc)
         .def_static("StraightLine", &Class::StraightLine, py::arg("K"),
@@ -1215,6 +1216,10 @@ void DoScalarDependentDefinitions(py::module m, T) {
             &Class::SolidCylinderWithDensityAboutEnd, py::arg("density"),
             py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
             cls_doc.SolidCylinderWithDensityAboutEnd.doc)
+        .def_static("SolidCylinderWithMassAboutEnd",
+            &Class::SolidCylinderWithMassAboutEnd, py::arg("mass"),
+            py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
+            cls_doc.SolidCylinderWithMassAboutEnd.doc)
         .def_static("ThinRodWithMass", &Class::ThinRodWithMass, py::arg("mass"),
             py::arg("length"), py::arg("unit_vector"),
             cls_doc.ThinRodWithMass.doc)
--- a/bindings/pydrake/multibody/tree_py.cc
+++ b/bindings/pydrake/multibody/tree_py.cc
@@ -1158,9 +1158,22 @@ void DoScalarDependentDefinitions(py::module m, T) {
         .def_static("SolidCapsule", &Class::SolidCapsule, py::arg("r"),
             py::arg("L"), py::arg("unit_vector") = Vector3<T>::UnitZ().eval(),
             cls_doc.SolidCapsule.doc)
-        .def_static("SolidCylinderAboutEnd", &Class::SolidCylinderAboutEnd,
+        // TODO(2023-10-01) Remove overload wrapping when deprecation complete.
+        .def_static("SolidCylinderAboutEnd",
+            overload_cast_explicit<Class, const T&, const T&,
+                const Vector3<T>&>(&Class::SolidCylinderAboutEnd),
             py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
             cls_doc.SolidCylinderAboutEnd.doc)
+        .def_static("SolidCylinderAboutEnd",
+            WrapDeprecated(cls_doc.SolidCylinderAboutEnd.doc_deprecated,
+                [](const T& r, const T& L) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+                  return Class::SolidCylinderAboutEnd(r, L);
+#pragma GCC diagnostic pop
+                }),
+            py::arg("r"), py::arg("L"),
+            cls_doc.SolidCylinderAboutEnd.doc_deprecated)
         .def_static("AxiallySymmetric", &Class::AxiallySymmetric, py::arg("J"),
             py::arg("K"), py::arg("b_E"), cls_doc.AxiallySymmetric.doc)
         .def_static("StraightLine", &Class::StraightLine, py::arg("K"),
--- a/multibody/plant/test/joint_limits_test.cc
+++ b/multibody/plant/test/joint_limits_test.cc
@@ -138,13 +138,10 @@ GTEST_TEST(JointLimitsTest, RevoluteJoint) {
 
   for (double time_step : {2.5e-4, 5.0e-4, 1.0e-3}) {
     MultibodyPlant<double> plant(time_step);
-    // The COM of the rod is right at its center, though we place the body frame
-    // B on the left end of the rod to connect it to the world with a revolute
-    // joint.
-    const auto M_B = SpatialInertia<double>::MakeFromCentralInertia(
-        mass, Vector3<double>(rod_length / 2.0, 0.0, 0.0),
-        mass * UnitInertia<double>::SolidCylinder(
-            rod_radius, rod_length, Vector3<double>::UnitX()));
+    // The rod's cm is at its centroid. The rod's body frame B is at the rod's
+    // left end and connects to the world with a revolute joint.
+    const auto M_B = SpatialInertia<double>::SolidCylinderWithMassAboutEnd(
+        mass, rod_radius, rod_length, Vector3<double>::UnitX());
     const RigidBody<double>& body = plant.AddRigidBody("Body", M_B);
     const RevoluteJoint<double>& pin = plant.AddJoint<RevoluteJoint>(
         "Pin", plant.world_body(), std::nullopt, body, std::nullopt,
--- a/multibody/plant/test/joint_limits_test.cc
+++ b/multibody/plant/test/joint_limits_test.cc
@@ -138,8 +138,8 @@ GTEST_TEST(JointLimitsTest, RevoluteJoint) {
 
   for (double time_step : {2.5e-4, 5.0e-4, 1.0e-3}) {
     MultibodyPlant<double> plant(time_step);
-    // The rod's cm is at its centroid. The rod's body frame B is at the rod's
-    // left end and connects to the world with a revolute joint.
+    // The rod's center of mass is at its centroid. The rod's body frame B is at
+    // the rod's left end and connects to the world with a revolute joint.
     const auto M_B = SpatialInertia<double>::SolidCylinderWithMassAboutEnd(
         mass, rod_radius, rod_length, Vector3<double>::UnitX());
     const RigidBody<double>& body = plant.AddRigidBody("Body", M_B);
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -181,11 +181,6 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithMass(
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
   ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
   const Vector3<T> p_BoBcm_B = Vector3<T>::Zero();
-
-  // Note: Although a check is made that ‖unit_vector‖ ≈ 1, even if imperfect,
-  // UnitInertia::SolidCylinder() allows for a near zero-vector due to code in
-  // UnitInertia::AxiallySymmetric() which normalizes unit_vector before use.
-  // TODO(Mitiguy) remove normalization in UnitInertia::AxiallySymmetric().
   const UnitInertia<T> G_BBo_B =
       UnitInertia<T>::SolidCylinder(radius, length, unit_vector);
   return SpatialInertia<T>(mass, p_BoBcm_B, G_BBo_B);
@@ -199,12 +194,23 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensityAboutEnd(
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
   ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
-  SpatialInertia<T> M_BBcm_B =
-      SpatialInertia<T>::SolidCylinderWithDensity(
-          density, radius, length, unit_vector);
-  const Vector3<T> p_BcmBp_B = -0.5 * length * unit_vector;
-  M_BBcm_B.ShiftFromCenterOfMassInPlace(p_BcmBp_B);
-  return M_BBcm_B;  // Due to shift, this actually returns M_BBp_B.
+  const T volume = M_PI * radius * radius * length;  // π r² l
+  const T mass = density * volume;
+  return SolidCylinderWithMassAboutEnd(mass, radius, length, unit_vector);
+}
+
+template <typename T>
+SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithMassAboutEnd(
+    const T& mass, const T& radius, const T& length,
+    const Vector3<T>& unit_vector) {
+  ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
+  ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
+  ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
+  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  const Vector3<T> p_BpBcm_B = 0.5 * length * unit_vector;
+  const UnitInertia<T> G_BBp_B =
+      UnitInertia<T>::SolidCylinderAboutEnd(radius, length, unit_vector);
+  return SpatialInertia(mass, p_BpBcm_B, G_BBp_B);  // Returns M_BBp_B.
 }
 
 template <typename T>
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -23,7 +23,7 @@ const boolean<T> is_nonnegative_finite(const T& value) {
 
 template<typename T>
 void ThrowUnlessValueIsPositiveFinite(const T& value,
-    const std::string& value_name, const std::string& function_name) {
+    std::string_view value_name, std::string_view function_name) {
   if (!is_positive_finite(value)) {
     DRAKE_DEMAND(!value_name.empty());
     DRAKE_DEMAND(!function_name.empty());
@@ -37,16 +37,18 @@ void ThrowUnlessValueIsPositiveFinite(const T& value,
 // Throw unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
 // Note: 1E-14 ≈ 2^5.5 * std::numeric_limits<double>::epsilon();
 template<typename T>
-void ThrowUnlessUnitVectorIsMagnitudeOne(const T& unit_vector,
-    const std::string& function_name) {
-  using std::abs;
-  constexpr double kTolerance = 1E-14;
-  if (abs(unit_vector.norm() - 1) > kTolerance) {
-    DRAKE_DEMAND(!function_name.empty());
-    const std::string error_message = fmt::format(
-        "{}(): The unit_vector argument {} is not a unit vector.",
-        function_name, fmt_eigen(unit_vector.transpose()));
-    throw std::logic_error(error_message);
+void ThrowUnlessVectorIsMagnitudeOne(const Vector3<T>& unit_vector,
+    std::string_view function_name) {
+  if constexpr (scalar_predicate<T>::is_bool) {
+    using std::abs;
+    constexpr double kTolerance = 1E-14;
+    if (abs(unit_vector.norm() - 1) > kTolerance) {
+      DRAKE_DEMAND(!function_name.empty());
+      const std::string error_message =
+          fmt::format("{}(): The unit_vector argument {} is not a unit vector.",
+                      function_name, fmt_eigen(unit_vector.transpose()));
+      throw std::logic_error(error_message);
+    }
   }
 }
 }  // namespace
@@ -136,7 +138,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCapsuleWithDensity(
   ThrowUnlessValueIsPositiveFinite(density, "density", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
 
   // Volume = π r² L + 4/3 π r³
   const T pi_r_squared = M_PI * radius * radius;
@@ -152,7 +154,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCapsuleWithMass(
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   const Vector3<T> p_BoBcm_B = Vector3<T>::Zero();
   const UnitInertia<T> G_BBo_B =
       UnitInertia<T>::SolidCapsule(radius, length, unit_vector);
@@ -166,7 +168,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensity(
   ThrowUnlessValueIsPositiveFinite(density, "density", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   const T volume = M_PI * radius * radius * length;  // π r² l
   const T mass = density * volume;
   return SolidCylinderWithMass(mass, radius, length, unit_vector);
@@ -179,7 +181,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithMass(
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   const Vector3<T> p_BoBcm_B = Vector3<T>::Zero();
   const UnitInertia<T> G_BBo_B =
       UnitInertia<T>::SolidCylinder(radius, length, unit_vector);
@@ -193,7 +195,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithDensityAboutEnd(
   ThrowUnlessValueIsPositiveFinite(density, "density", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   const T volume = M_PI * radius * radius * length;  // π r² l
   const T mass = density * volume;
   return SolidCylinderWithMassAboutEnd(mass, radius, length, unit_vector);
@@ -206,7 +208,7 @@ SpatialInertia<T> SpatialInertia<T>::SolidCylinderWithMassAboutEnd(
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(radius, "radius", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   const Vector3<T> p_BpBcm_B = 0.5 * length * unit_vector;
   const UnitInertia<T> G_BBp_B =
       UnitInertia<T>::SolidCylinderAboutEnd(radius, length, unit_vector);
@@ -218,7 +220,7 @@ SpatialInertia<T> SpatialInertia<T>::ThinRodWithMass(
     const T& mass, const T& length, const Vector3<T>& unit_vector) {
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
 
   // Although a check is made that ‖unit_vector‖ ≈ 1, we normalize regardless.
   // Note: UnitInertia::ThinRod() calls UnitInertia::AxiallySymmetric() which
@@ -235,7 +237,7 @@ SpatialInertia<T> SpatialInertia<T>::ThinRodWithMassAboutEnd(
     const T& mass, const T& length, const Vector3<T>& unit_vector) {
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
-  ThrowUnlessUnitVectorIsMagnitudeOne(unit_vector, __func__);
+  ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
   SpatialInertia<T> M_BBcm_B =
       SpatialInertia<T>::ThinRodWithMass(mass, length, unit_vector);
   const Vector3<T> p_BcmBp_B = -0.5 * length * unit_vector;
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -34,8 +34,9 @@ void ThrowUnlessValueIsPositiveFinite(const T& value,
   }
 }
 
-// Throw unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
+// Throws unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
 // Note: 1E-14 ≈ 2^5.5 * std::numeric_limits<double>::epsilon();
+// Note: This function is a no-op when type T is symbolic::Expression.
 template<typename T>
 void ThrowUnlessVectorIsMagnitudeOne(const Vector3<T>& unit_vector,
     std::string_view function_name) {
--- a/multibody/tree/spatial_inertia.h
+++ b/multibody/tree/spatial_inertia.h
@@ -266,7 +266,7 @@ class SpatialInertia {
       const Vector3<T>& unit_vector);
 
   /// Creates a spatial inertia for a uniform-density solid cylinder B about an
-  /// end-point Bp of the cylinder's central longitudinal axis (see below).
+  /// end-point Bp of the cylinder's axis (see below for more about Bp).
   /// @param[in] density mass per volume (kg/m³).
   /// @param[in] radius radius of cylinder (meters).
   /// @param[in] length length of cylinder in unit_vector direction (meters).
@@ -286,7 +286,7 @@ class SpatialInertia {
       const Vector3<T>& unit_vector);
 
   /// Creates a spatial inertia for a uniform-density solid cylinder B about an
-  /// end-point Bp of the cylinder's central longitudinal axis (see below).
+  /// end-point Bp of the cylinder's axis (see below for more about Bp).
   /// @param[in] mass mass of the solid cylinder (kg).
   /// @param[in] radius radius of cylinder (meters).
   /// @param[in] length length of cylinder in unit_vector direction (meters).
--- a/multibody/tree/spatial_inertia.h
+++ b/multibody/tree/spatial_inertia.h
@@ -265,15 +265,15 @@ class SpatialInertia {
       const T& mass, const T& radius, const T& length,
       const Vector3<T>& unit_vector);
 
-  /// Creates a spatial inertia for a uniform-density solid cylinder B about an
-  /// end-point Bp (Bp is at the center of one of the cylinder's circular ends).
+  /// Creates a spatial inertia for a uniform-density solid cylinder B about the
+  /// point Bp (Bp is at a center of the cylinder's circular ends; see below).
   /// @param[in] density mass per volume (kg/m³).
   /// @param[in] radius radius of cylinder (meters).
   /// @param[in] length length of cylinder in unit_vector direction (meters).
-  /// @param[in] unit_vector unit vector defining the axial direction of the
-  /// cylinder, expressed in B.
+  /// @param[in] unit_vector unit vector parallel to the axis of the cylinder
+  /// and directed from Bp to Bcm (B's center of mass), expressed in B.
   /// @retval M_BBp_B B's spatial inertia about Bp, expressed in B.
-  /// @note The position from Bp to Bcm is length / 2 * unit_vector.
+  /// @note The position from Bp to Bcm is p_BpBcm = length / 2 * unit_vector.
   /// @note B's rotational inertia about Bp is axially symmetric, meaning B has
   /// an equal moment of inertia about any line that both passes through Bp
   /// and is perpendicular to unit_vector.
@@ -285,15 +285,15 @@ class SpatialInertia {
       const T& density, const T& radius, const T& length,
       const Vector3<T>& unit_vector);
 
-  /// Creates a spatial inertia for a uniform-density solid cylinder B about an
-  /// end-point Bp (Bp is at the center of one of the cylinder's circular ends).
+  /// Creates a spatial inertia for a uniform-density solid cylinder B about the
+  /// point Bp (Bp is at a center of the cylinder's circular ends; see below).
   /// @param[in] mass mass of the solid cylinder (kg).
   /// @param[in] radius radius of cylinder (meters).
   /// @param[in] length length of cylinder in unit_vector direction (meters).
-  /// @param[in] unit_vector unit vector defining the axial direction of the
-  /// cylinder, expressed in B.
+  /// @param[in] unit_vector unit vector parallel to the axis of the cylinder
+  /// and directed from Bp to Bcm (B's center of mass), expressed in B.
   /// @retval M_BBp_B B's spatial inertia about Bp, expressed in B.
-  /// @note The position from Bp to Bcm is length / 2 * unit_vector.
+  /// @note The position from Bp to Bcm is p_BpBcm = length / 2 * unit_vector.
   /// @note B's rotational inertia about Bp is axially symmetric, meaning B has
   /// an equal moment of inertia about any line that both passes through Bp
   /// and is perpendicular to unit_vector.
--- a/multibody/tree/spatial_inertia.h
+++ b/multibody/tree/spatial_inertia.h
@@ -259,7 +259,8 @@ class SpatialInertia {
   /// and is perpendicular to unit_vector.
   /// @throws std::exception if mass, radius, or length is not positive and
   /// finite or if ‖unit_vector‖ is not within 1.0E-14 of 1.0.
-  //  TODO(Mitiguy) Create SolidCylinderWithMassAboutEnd().
+  /// @see SolidCylinderWithMassAboutEnd() to calculate M_BBp_B, B's spatial
+  /// inertia about Bp (at the center of one of the cylinder's circular ends).
   static SpatialInertia<T> SolidCylinderWithMass(
       const T& mass, const T& radius, const T& length,
       const Vector3<T>& unit_vector);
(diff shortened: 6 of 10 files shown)


mitiguy (commented):
@drake-jenkins-bot linux-jammy-gcc-bazel-experimental-debug   please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-N_oguZH7EtXFMHdK1xn:b-xtjm9a)* status: needs platform reviewer assigned, needs at least two assigned reviewers

mitiguy (commented):
Feature review +@sherm1 

Note: For reasons that perplex me, this does not pass CI. It may (or may not) have to do with the error message that I see for linux-jammy-gcc-bazel-experimental-debug, which is:

Matplotlib created a temporary config/cache directory at /tmp/matplotlib-sco6bphl because the default path (/home/ubuntu/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.

Running tests...
----------------------------------------------------------------------
INFO:drake:Meshcat listening for connections at http://localhost:7010
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  130 (MIT-SHM)
  Minor opcode of failed request:  3 (X\_ShmPutImage)
  Value in failed request:  0x5c0
  Serial number of failed request:  46
  Current serial number in output stream:  47
Matplotlib created a temporary config/cache directory at /tmp/matplotlib-iwm76bst because the default path (/home/ubuntu/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-N_oxK-_AwjR36lC8m1V:b-923q6e)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers

___
*[`multibody/tree/unit_inertia.cc` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-N_oxDP-34TbdgsW1QnY:-N_oxDP-34TbdgsW1QnZ:bh3sfr7) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L62)):*
> ```cpp
>   DRAKE_THROW_UNLESS(r >= 0);
>   DRAKE_THROW_UNLESS(L >= 0);
>   // UnitInertia::SolidCylinder() allows for a near zero-vector due to code in
> ```

BTW  Paul - self blocking.  Remove previous line as it should not be there.

sherm1 (commented):
I'm out all week so to avoid delay -@sherm1 +@SeanCurtis-TRI for feature review, please (or reassign)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-N_zGZS36LOrlrurMEkE:b-xy1x34)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers

SeanCurtis-TRI:
I suspect your test error is related to #18726.

We'll try retriggering it and see how it goes.

SeanCurtis-TRI:
@drake-jenkins-bot retest linux-jammy-gcc-bazel-experimental-debug please

SeanCurtis-TRI (commented):
As I read this PR, I don't see an obvious argument for why "breaking" is appropriate here. There are two reasons to make this a breaking change:

1. This code isn't covered by the stability promise, or
2. Deprecating would be "impractical, too expensive, or logically unsound" (according to the [stable API definition](https://drake.mit.edu/stable.html#stable-api).

I can't see that either of those apply here. Now, I can easily imagine that you're not sure how to deprecate. Because I didn't believe breaking here was appropriate, I prepared a commit that changes breaking to deprecation and I can simply apply it to your PR if you like. Happy to help. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-Na41mfw00xOtGv_0_BU:b-3og98d)* status: 13 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3ty364xD_ebaOGAft):*
nit: The fact that UnitInertia::SolidCyilnderAboutEnd() changed is an easily observable fact from looking at the contents of the commit. Spending characters on the summary line to report that isn't helpful.

Summarizing the nature and reason of the change *is* helpful. That kind of text is excellent fodder for commit messages. If I were to re-write this commit message, it would be something like this (assuming we agree on deprecation over breaking):

```
Add SpatialInertia::SolidCylinderWithMassAboutEnd()

In support of the SpatialInertia function, the corresponding UnitInertia
API has changed to support cylinders with arbitrary axis orientation.
This required deprecating the old API in favor of the new.
```
___
*[`multibody/tree/spatial_inertia.h` line 289 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3deRp9szPwZ3DWJhW:-Na3deRp9szPwZ3DWJhX:b-ez3fcx) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/spatial_inertia.h#L289)):*
> ```objc
> 
>   /// Creates a spatial inertia for a uniform-density solid cylinder B about an
>   /// end-point Bp (Bp is at the center of one of the cylinder's circular ends).
> ```

BTW: I feel that "one of" is needlessly vague. This unique contribution of this function (compared to `SolidCylinderWithMass`) is 100% dependent upon Bp.  Surely the reader shouldn't have to wonder or dig (and there is *a lot* of text to dig through) to gain even an intuition for Bp, right?

Possible suggestion:

```
...about the point Bp (Bp is at the center of the cylinder's "bottom" circular end; see below).
...
@note p_BpBcm = length / 2 * unit_vector
```

(replacing '=' with your favorite "is defined as" symbol).

(FTR, "end point" works well for a line segment, but less well for a cylinder.)

I'd note that the same issue exists for the density-variant of this function.



_Code quote:_
```objc
one of the cylinder's circular ends
```
___
*[`multibody/tree/spatial_inertia.h` line 878 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3gbQm3T2geLi6euWi:-Na3gbQm3T2geLi6euWj:bu6wajj) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5e5569cf572130ec308ad274f5abd440f7a7b2b/multibody/tree/spatial_inertia.h#L878)):*
> ```objc
>   // Gzz = 1/5 (a² + b²)                     | Gzz = 1/3 (a² + b²)
>   // shape_factor = 1/5                      | shape_factor = 1/3
>   //-----------------------------------------|----------------------------------
> ```

BTW Out of curiosity? This deletion doesn't seem to be captured by the documentation of this PR. Why this as well?
___
*[`multibody/tree/unit_inertia.h` line 300 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3i8ii4k0z66PwKsCm:-Na3i8ii4k0z66PwKsCn:bjnjjh8) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.h#L300)):*
> ```objc
> 
>   /// Creates a unit inertia for a uniform-density solid cylinder B about an
>   /// end-point Bp (Bp is at the center of one of the cylinder's circular ends).
> ```

nit Same note here about "one of" (re: Bp). And *this* function doesn't even define Bp.

Given the correctness of `SpatialInertia::Compute.....End()` depends on this function's definition of `Bp`, this function should probably document the definition and support it with a test.
___
*[`multibody/tree/unit_inertia.h` line 305 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3ifgzCOjDJvv4fv7u:-Na3ifgzCOjDJvv4fv7v:bkozmv) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.h#L305)):*
> ```objc
>   /// @param[in] unit_vector unit vector defining the axial direction of the
>   /// cylinder, expressed in B.
>   /// @throws std::exception if radius or length is negative or if
> ```

BTW I just realized that the *unit inertia* functions don't care about finite measures? They throw for negative, but pass infinite or nan values through happily? That seems inconsistent. Either they should really validate the values or they should just declare prerequisites. But it seems they are currently halfway between.

Taking any action based on this observation is *not* intended to be part of *this* PR. I mention it to make sure it's on your radar. If it already is, feel free to dismiss this comment as redundant.
___
*[`multibody/tree/unit_inertia.cc` line 13 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3jiw_3O2nNHpu8yXg:-Na3jiw_3O2nNHpu8yXh:b-lups7f) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L13)):*
> ```cpp
> // Note: 1E-14 ≈ 2^5.5 * std::numeric_limits<double>::epsilon();
> template<typename T>
> void ThrowUnlessUnitVectorIsMagnitudeOne(const T& unit_vector,
> ```

BTW This applies to `SpatialInertia` as well. This function will choke if someone creates a `SpatialInertia<symbolic::Expression>` whose direction includes unbounded variables. You'll get the error:

```
The following environment does not have an entry for the variable <variable name>
```

Is this what you want?

One possible solution is to *not* evaluate for `T = Expression`.
Another possible solution is to only evaluate if the expression has no variables.
There are probably other solutions -- it's too bad that Drake hasn't documented a guidebook for being "Expression compatible" as I'm sure there are other "standard" practices.

Furthermore, based on the subtlety of handling all of the T correctly, you probably don't want this defined twice, but once and re-used twice.
___
*[`multibody/tree/unit_inertia.cc` line 13 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3maKBBStYe2I8xgwd:-Na3maKBBStYe2I8xgwe:bfxp4vz) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L13)):*
> ```cpp
> // Note: 1E-14 ≈ 2^5.5 * std::numeric_limits<double>::epsilon();
> template<typename T>
> void ThrowUnlessUnitVectorIsMagnitudeOne(const T& unit_vector,
> ```

BTW The name of this function is verbose to the point of absurdity, literally. The name currently expresses a condition which can never throw: if a "unit vector" is not "unit length". That's a tautology and, as such, is never false.

This would be more concisely (and correctly) named, `ThrowUnlessVectorIsMagnitudeOne` or `ThrowUnlessVectorIsUnitLength` -- after all, the point of the function is to prove that the given vector *is* a unit vector.
___
*[`multibody/tree/unit_inertia.cc` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3lNpvDBIvJY-VNdZz:-Na3lNpvDBIvJY-VNd_-:b-gtye1e) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L14)):*
> ```cpp
> template<typename T>
> void ThrowUnlessUnitVectorIsMagnitudeOne(const T& unit_vector,
>     const std::string& function_name) {
> ```

nit: this shouldn't be `const std::string&` it should simply be `std::string_view`. What you currently have constructs a string, which is a shame when you're passing compile-time `const char*`. `string_view` just keeps a pointer and doesn't do heap allocation. As you're doing these tests in release mode, might as well forego the unnecessary allocation.
___
*[`multibody/tree/unit_inertia.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3na7nC4jNAFFyGLjh:-Na3na7nC4jNAFFyGLji:b-kzqc7e) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L78)):*
> ```cpp
>   const T r2 = radius * radius;
>   const T l2 = length * length;
>   const T J = T(0.5) * r2;  // Axial moment of inertia J = ½ r².
> ```

BTW My best understanding is that `T(0.5)` isn't necessary. Simply 0.5 would be enough. Do you remember your reason for explicitly doing the `T(##)` thing?
___
*[`multibody/tree/test/spatial_inertia_test.cc` line 274 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3oAtHDpUkcEBCc2UG:-Na3oAtHDpUkcEBCc2UH:b-td0int) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/spatial_inertia_test.cc#L274)):*
> ```cpp
> 
>   // Test a solid cylinder B about Bcm whose unit_vector is in the z-direction.
>   Vector3<double> unit_vec(0, 0, 1);
> ```

BTW When I first read this code, I wondered why you were doing a test with all of those identities in the values -- I generally dislike identities as they can mask all sorts of errors in the tested code. But then I noted below, that the reason `unit_vec` isn't `const` is because you change it to a more "interesting" (non-axis-aligned) direction vector. So, now my concern has evolved.

You've essentially duplicated the tests: once with <0, 0, 1> and once with ... the other one  :)  It seems to me that the tests on the non-axis-aligned unit vector give us full test coverage. What persistent value does the additional toy version add to these unit tests?

I can understand when writing the unit tests, that it makes sense to start with really simple values to help human reasoning. But what conditions will the <0, 0, 1>-valued tests catch that the interesting vector wouldn't catch?

Certainly, the actual code of the unit test reads *exactly* the same whichever value of the unit vector you use -- so there's no human readability benefit.

If the extra tests don't provide value, they should be excised.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3r-Km3HOoZdjnMBCX:-Na3r-Km3HOoZdjnMBCY:b-sicrn6) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/unit_inertia_test.cc#L261)):*
> ```cpp
>   const UnitInertia<double> G_expected(I_perp, I_perp, I_axial);
>   UnitInertia<double> G_BBp_B = UnitInertia<double>::SolidCylinderAboutEnd(
>       r, L, Vector3<double>::UnitZ());
> ```

 BTW: Regrettable that you're testing this with the vector <0, 0, 1>. As this is the fundamental, underlying function that truly depends on the unit vector, this is where the testing should take place.

Right now you're in a state where you can argue that the unit tests on `SpatialInertia` provide the evidence that `UnitInertia` is correct, but that's not optimal. Presumably, the argument is that `AxiallySymmetric` has been sufficiently tested that we don't need an interesting axis, just and indication that a `UnitInertia` is produced. Yes?
___
*[`multibody/tree/test/unit_inertia_test.cc` line 605 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3rMKSC9ND7BQ6bOpW:-Na3rMKSC9ND7BQ6bOpX:bcgejje) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/unit_inertia_test.cc#L605)):*
> ```cpp
>   const UnitInertia<double> G_expected =
>       UnitInertia<double>::SolidCylinderAboutEnd(
>           r, L, Vector3<double>::UnitZ());
> ```

btw: Same note re: <0 , 0, 1>.

SeanCurtis-TRI:
[@drake-jenkins-bot](https://github.com/drake-jenkins-bot) retest linux-jammy-gcc-bazel-experimental-debug please

SeanCurtis-TRI:

mitiguy (commented):
In view of Sean's suggestion, planned commit message is:

```
Add SpatialInertia::SolidCylinderWithMassAboutEnd()

In support of the SpatialInertia function, the corresponding UnitInertia
API has changed to support cylinders with arbitrary axis orientation.
This required deprecating the old API in favor of the new.
```

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-Na8r88f8zVH4CAn8mUu:b-ke7kdo)* status: 8 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3ty364xD_ebaOGAft):*

nit: The fact that UnitInertia::SolidCyilnderAboutEnd() changed is an easily observable fact from looking at the contents of the commit. Spending characters on the summary line to report that isn't helpful.

Summarizing the nature and reason of the change *is* helpful. That kind of text is excellent fodder for commit messages. If I were to re-write this commit message, it would be something like this (assuming we agree on deprecation over breaking):

```
Add SpatialInertia::SolidCylinderWithMassAboutEnd()

In support of the SpatialInertia function, the corresponding UnitInertia
API has changed to support cylinders with arbitrary axis orientation.
This required deprecating the old API in favor of the new.
```
</blockquote></details>

I like and accept your suggestions. I do not fully understand deprecation vs. breaking, but have changed the labels to align with your suggestion.
___
*[`multibody/tree/spatial_inertia.h` line 289 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3deRp9szPwZ3DWJhW:-Na90xU-7KUFXj5rSq67:b-1g0ps4) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/spatial_inertia.h#L289)):*

BTW: I feel that "one of" is needlessly vague. This unique contribution of this function (compared to `SolidCylinderWithMass`) is 100% dependent upon Bp.  Surely the reader shouldn't have to wonder or dig (and there is *a lot* of text to dig through) to gain even an intuition for Bp, right?

Possible suggestion:

```
...about the point Bp (Bp is at the center of the cylinder's "bottom" circular end; see below).
...
@note p_BpBcm = length / 2 * unit_vector
```

(replacing '=' with your favorite "is defined as" symbol).

(FTR, "end point" works well for a line segment, but less well for a cylinder.)

I'd note that the same issue exists for the density-variant of this function.
</blockquote></details>

Improved here and elsewhere, but without "bottom".
The idea of "bottom" is biased towards the default Vector3::UnitZ() direction (perhaps due to defaults for Drake cylinders).  My mental image was a cylinder rolling on a table, so bottom was not immediately obvious to me.
___
*[`multibody/tree/spatial_inertia.h` line 878 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3gbQm3T2geLi6euWi:-Na91Nat5Y7X8nE__oJe:b-prqa9v) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5e5569cf572130ec308ad274f5abd440f7a7b2b/multibody/tree/spatial_inertia.h#L878)):*

BTW Out of curiosity? This deletion doesn't seem to be captured by the documentation of this PR. Why this as well?
</blockquote></details>

Opportunistic removal of erroneous comment.
Hollow ellipsoid was cut/paste -- but not tested/verified.
TBD for Hollow box no longer warrants attention.
This comment (and related code for implemented function) was added in the last few months.
I want to remove this now as I may otherwise forget to get rid of it.
___
*[`multibody/tree/unit_inertia.h` line 300 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3i8ii4k0z66PwKsCm:-Na9bYH2EKRRYIcUrTMY:b-k3qipv) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.h#L300)):*

nit Same note here about "one of" (re: Bp). And *this* function doesn't even define Bp.

Given the correctness of `SpatialInertia::Compute.....End()` depends on this function's definition of `Bp`, this function should probably document the definition and support it with a test.
</blockquote></details>

The function uses DRAKE_DEMAND(radius >=0) and DRAKE_DEMAND(length>=0).
Alternatively, I can implement code to test for radius >= 0 and throw useful message.
Similarly for length >= 0.  Was planning on doing this in subsequent PR.  See below.
___
*[`multibody/tree/unit_inertia.h` line 305 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3ifgzCOjDJvv4fv7u:-Na99f1w3ANvKD6Tjz2g:b-ds66d3) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.h#L305)):*

BTW I just realized that the *unit inertia* functions don't care about finite measures? They throw for negative, but pass infinite or nan values through happily? That seems inconsistent. Either they should really validate the values or they should just declare prerequisites. But it seems they are currently halfway between.

Taking any action based on this observation is *not* intended to be part of *this* PR. I mention it to make sure it's on your radar. If it already is, feel free to dismiss this comment as redundant.
</blockquote></details>

I agree.  Will do in subsequent PR. 
___
*[`multibody/tree/unit_inertia.cc` line 13 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3jiw_3O2nNHpu8yXg:-Na9TTxy28T5bDgU-xUl:b-ua7d1f) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L13)):*

BTW This applies to `SpatialInertia` as well. This function will choke if someone creates a `SpatialInertia<symbolic::Expression>` whose direction includes unbounded variables. You'll get the error:

```
The following environment does not have an entry for the variable <variable name>
```

Is this what you want?

One possible solution is to *not* evaluate for `T = Expression`.
Another possible solution is to only evaluate if the expression has no variables.
There are probably other solutions -- it's too bad that Drake hasn't documented a guidebook for being "Expression compatible" as I'm sure there are other "standard" practices.

Furthermore, based on the subtlety of handling all of the T correctly, you probably don't want this defined twice, but once and re-used twice.
</blockquote></details>

With help from Hongkai and Xuchen, I put in some guards.  It would be better if this was only defined once.  Feel free to direct me as to where the one occurence should go -- but in a subsequent PR.

Possibly more to also consider with is_positive_finite() and is_nonnegative_finite().
___
*[`multibody/tree/unit_inertia.cc` line 13 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3maKBBStYe2I8xgwd:-Na9GG26E4yeFrhloRr2:bh620qs) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L13)):*

BTW The name of this function is verbose to the point of absurdity, literally. The name currently expresses a condition which can never throw: if a "unit vector" is not "unit length". That's a tautology and, as such, is never false.

This would be more concisely (and correctly) named, `ThrowUnlessVectorIsMagnitudeOne` or `ThrowUnlessVectorIsUnitLength` -- after all, the point of the function is to prove that the given vector *is* a unit vector.
</blockquote></details>

Done.  Changed here and in spatial_inertia.cc
___
*[`multibody/tree/unit_inertia.cc` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3lNpvDBIvJY-VNdZz:-Na9XcZxBOQOEo73-pGC:bgpfdg2) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L14)):*

nit: this shouldn't be `const std::string&` it should simply be `std::string_view`. What you currently have constructs a string, which is a shame when you're passing compile-time `const char*`. `string_view` just keeps a pointer and doesn't do heap allocation. As you're doing these tests in release mode, might as well forego the unnecessary allocation.
</blockquote></details>

Done  Here and elsewhere.
___
*[`multibody/tree/unit_inertia.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3na7nC4jNAFFyGLjh:-Na9Y125ATHHaJIv8ZFv:bsf08lt) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L78)):*

BTW My best understanding is that `T(0.5)` isn't necessary. Simply 0.5 would be enough. Do you remember your reason for explicitly doing the `T(##)` thing?
</blockquote></details>

Done.  No, I do not remember.  I removed the T().
___
*[`multibody/tree/test/spatial_inertia_test.cc` line 274 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3oAtHDpUkcEBCc2UG:-Na9gron6B8LhNfrYkY4:b8retda) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/spatial_inertia_test.cc#L274)):*

BTW When I first read this code, I wondered why you were doing a test with all of those identities in the values -- I generally dislike identities as they can mask all sorts of errors in the tested code. But then I noted below, that the reason `unit_vec` isn't `const` is because you change it to a more "interesting" (non-axis-aligned) direction vector. So, now my concern has evolved.

You've essentially duplicated the tests: once with <0, 0, 1> and once with ... the other one  :)  It seems to me that the tests on the non-axis-aligned unit vector give us full test coverage. What persistent value does the additional toy version add to these unit tests?

I can understand when writing the unit tests, that it makes sense to start with really simple values to help human reasoning. But what conditions will the <0, 0, 1>-valued tests catch that the interesting vector wouldn't catch?

Certainly, the actual code of the unit test reads *exactly* the same whichever value of the unit vector you use -- so there's no human readability benefit.

If the extra tests don't provide value, they should be excised.

</blockquote></details>

Done. Excised.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3r-Km3HOoZdjnMBCX:-Na9kkxg0OaUuyuCtmVe:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/unit_inertia_test.cc#L261)):*

 BTW: Regrettable that you're testing this with the vector <0, 0, 1>. As this is the fundamental, underlying function that truly depends on the unit vector, this is where the testing should take place.

Right now you're in a state where you can argue that the unit tests on `SpatialInertia` provide the evidence that `UnitInertia` is correct, but that's not optimal. Presumably, the argument is that `AxiallySymmetric` has been sufficiently tested that we don't need an interesting axis, just and indication that a `UnitInertia` is produced. Yes?
</blockquote></details>

Done.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 605 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3rMKSC9ND7BQ6bOpW:-Na9l5B69SkX8riN5mfz:b-plmrd8) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/test/unit_inertia_test.cc#L605)):*

btw: Same note re: <0 , 0, 1>.
</blockquote></details>

Objection noted (and I sort of agree with) -- but I am not planning on changing this. 
Reason: The only change to this test is to make it compatible with the changes to arguments for SolidCylinderAboutEnd().  I agree with the requested test changes for<0, 0, 1> for SolidCylinderAboutEnd().  However, the Vector3<double>::UnitZ() was the default argument used in this old SolidCapsule test -- which is really more of a rewrite of a test that already passed review.

SeanCurtis-TRI (commented):
BTW Another byproduct of roaming around Tokyo yesterday, was a thought on the subject of break vs deprecate.

One of the justifications of "breaking" is "too expensive". I'd originally dismissed it because the deprecation code wasn't particularly tricky to write. So, I was judging the price on an absolute scale. However, if the `UnitInertia` API is functionally unused, then the value of deprecation is so small that even the small absolute price becomes "too expensive".

So, would you make that argument? If so, I'd recommend that once you have CI passing on this PR, you should run it against Anzu to make sure that Anzu won't break (I'll message you the link you need on slack.) Although, given the mysterious nature of your CI failure, I don't think you have to wait to run the Anzu build.

Alternatively, you weren't explicit about accepting my deprecate branch (however, you seemed to accept it implicitly). Let me know, and I'll push my deprecation code to your branch.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaEBGhmB8Bx5ECNizO7:b-qdrmtu)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 289 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3deRp9szPwZ3DWJhW:-NaE4dRa5FoRzyygDl3-:bas5b4a) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/spatial_inertia.h#L289)):*

Improved here and elsewhere, but without "bottom".
The idea of "bottom" is biased towards the default Vector3::UnitZ() direction (perhaps due to defaults for Drake cylinders).  My mental image was a cylinder rolling on a table, so bottom was not immediately obvious to me.
</blockquote></details>

All you really did there is double down on the word "endpoint". What you missed is what I highlighted; you've eliminated "one of" and left "an". To boil this down to its simplest terms: there are obviously two caps on the cylinder. The summary text actively obscures which cap `P` lies on. My proposed text was directed toward giving an *intuition* (whose more formal definition could be found via digging).

As I was tootling around Tokyo yesterday, I had the same thought about my term "bottom" that you had. But your argument assumes a frame: the world frame. My suggestion suggests the cylinder is the reference frame. I obviously have a bias towards the geometry frame.  :) But beyond that, it's not obvious that one is clearly right.

However, you do define the cylinder's axis, *not* with a line, but with a *vector*, you have a direction, ready to hand, for defining the "top" of the cylinder. I did put the word "bottom" in quotes exactly for that reason. But I realize that only works if the `unit_vector` is also associated with the word "top".

Ultimately, my core concern went untouched. Whether or not you like "top"/"bottom" is up to you. But it's an objective fact that given that your text implies two possible definitions for the location of P, the *only* way a reader can infer which it is, is to *carefully* read all the text.

(And the lawyer in my soul doubts that an "axis" has endpoints at all.)
___
*[`multibody/tree/spatial_inertia.h` line 878 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3gbQm3T2geLi6euWi:-NaE6Hy5CaDuyZnCOYVu:bi3m4os) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5e5569cf572130ec308ad274f5abd440f7a7b2b/multibody/tree/spatial_inertia.h#L878)):*

Opportunistic removal of erroneous comment.
Hollow ellipsoid was cut/paste -- but not tested/verified.
TBD for Hollow box no longer warrants attention.
This comment (and related code for implemented function) was added in the last few months.
I want to remove this now as I may otherwise forget to get rid of it.
</blockquote></details>

It would serve you well not to surprise your reviewers, otherwise they may think it's a mistake.
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaE7H7sASct9KT20NZN:-NaE7H7sASct9KT20NZO:b-14fk8a) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b2b542affcae4e5dee2c80e38d5e1d02ae57109/multibody/tree/spatial_inertia.cc#L42)):*
> ```cpp
> void ThrowUnlessVectorIsMagnitudeOne(const Vector3<T>& unit_vector,
>     std::string_view function_name) {
>   if constexpr (scalar_predicate<T>::is_bool) {
> ```

With this change, your documentation should be clear that expressions don't get validated. For every function that relies on this function, its documentation should be updated.
___
*[`multibody/tree/unit_inertia.h` line 300 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3i8ii4k0z66PwKsCm:-NaE7eja0O9AnW_zcXe3:b-46hb8n) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.h#L300)):*

The function uses DRAKE_DEMAND(radius >=0) and DRAKE_DEMAND(length>=0).
Alternatively, I can implement code to test for radius >= 0 and throw useful message.
Similarly for length >= 0.  Was planning on doing this in subsequent PR.  See below.
</blockquote></details>

Depending on how my treatise above about "one of" vs "an" goes above, it applies here as well. I'm marking my resolved here so that the only conversation happens above.

I'm a bit confused about your response mentioning `radius` and `length` as my comment wasn't about validating parameters, but something this function implicitly defines. However, it doesn't seem I need to worry about that confusion.

Your change certainly addresses the the concern about it not being documented (`P` is now defined) and your tests test it. So, I'm content.
___
*[`multibody/tree/unit_inertia.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3na7nC4jNAFFyGLjh:-NaE9gEYB52ewCi4qbW1:bnv2wpc) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L78)):*

Done.  No, I do not remember.  I removed the T().
</blockquote></details>

You might consider a subsequent, superficial PR that goes through these files and removes them all so the whole document is stylistically consistent. I only mentioned it in *this* PR because it was highlighted and jumped out at me.  :)
___
*[`multibody/tree/unit_inertia.cc` line 15 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaE9Y_m04ORwTGz3248:-NaE9Y_m04ORwTGz3249:b-iwy4l9) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b2b542affcae4e5dee2c80e38d5e1d02ae57109/multibody/tree/unit_inertia.cc#L15)):*
> ```cpp
> void ThrowUnlessVectorIsMagnitudeOne(const Vector3<T>& unit_vector,
>                                      std::string_view function_name) {
>   if constexpr (scalar_predicate<T>::is_bool) {
> ```

nit: Same note about the functions that rely on this validation documenting that unit length isn't tested for `Expression`.

mitiguy (commented):
@SeanCurtis-TRI 
1. Small documentation enhancements and copious argument in their support.
2. As you mentioned, it seems issue #18726 is causing problems with CI failures in this PR #19803 
3. Feel free to push your deprecation code to the branch (thank you).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaEsSMB4fQfmumnS9DQ:b3t23en)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 289 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3deRp9szPwZ3DWJhW:-NaEYNm89puSaAMkkbCC:b-z6dmm8) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/spatial_inertia.h#L289)):*

All you really did there is double down on the word "endpoint". What you missed is what I highlighted; you've eliminated "one of" and left "an". To boil this down to its simplest terms: there are obviously two caps on the cylinder. The summary text actively obscures which cap `P` lies on. My proposed text was directed toward giving an *intuition* (whose more formal definition could be found via digging).

As I was tootling around Tokyo yesterday, I had the same thought about my term "bottom" that you had. But your argument assumes a frame: the world frame. My suggestion suggests the cylinder is the reference frame. I obviously have a bias towards the geometry frame.  :) But beyond that, it's not obvious that one is clearly right.

However, you do define the cylinder's axis, *not* with a line, but with a *vector*, you have a direction, ready to hand, for defining the "top" of the cylinder. I did put the word "bottom" in quotes exactly for that reason. But I realize that only works if the `unit_vector` is also associated with the word "top".

Ultimately, my core concern went untouched. Whether or not you like "top"/"bottom" is up to you. But it's an objective fact that given that your text implies two possible definitions for the location of P, the *only* way a reader can infer which it is, is to *carefully* read all the text.

(And the lawyer in my soul doubts that an "axis" has endpoints at all.)
</blockquote></details>


I hope you and family are enjoying your time in Tokyo.  It is wonderful that you are taking time to explore.

For sake of linguistic argument (hopefully, this makes you smile):

1. You write: "My proposed text was directed toward giving an _intuition_ (whose more formal definition could be found via digging)."   I agree that "digging" helps clarify what is meant by Bp.

2. I'll argue that a world frame is inherent with "Bottom".  For Earth dwelling folks, "bottom" has to do with vertical and gravity (or if on a "flat" part of Earth, outward normal from Earth's mostly spherical surface).  Earth, Earth, Earth (World).

I mentioned that my mental image was of a cylinder rolling on a table.  However, the documentation makes no such reference.  Instead it only discusses the geometry of a cylinder and the unit vector -- both essential to defining Bp.  We both have a strong bias towards focusing on the cylinder's geometry.

3. Your linguistic lawyer seems so right.  An "axis" does not have endpoints. Perhaps a more explicitly clear set of words would be "an end-point of the cylinder's central longitudinal line segment."  

However, in view of the definition of the axis of cylinder [here](https://www.varsitytutors.com/hotmath/hotmath_help/topics/cylinder#:~:text=The%20axis%20of%20a%20cylinder,it%20is%20an%20oblique%20cylinder), the **axis** of a cylinder **is** the segment containing the centers of the two bases.

It seems reasonable to emphasize that understanding Bp requires some digging.
I will emphasize "digging" by enhancing the documentation to: 
  /// Creates a spatial inertia for a uniform-density solid cylinder B about an
  /// end-point Bp of the cylinder's axis (see below for more about Bp).
___
*[`multibody/tree/spatial_inertia.h` line 878 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3gbQm3T2geLi6euWi:-NaEfyes7JMSzbHO4Dac:b-qxq1u0) ([raw file](https://github.com/RobotLocomotion/drake/blob/b5e5569cf572130ec308ad274f5abd440f7a7b2b/multibody/tree/spatial_inertia.h#L878)):*

It would serve you well not to surprise your reviewers, otherwise they may think it's a mistake.
</blockquote></details>

Agreed.  I should have put a well-placed comment in Reviewable when I pushed my first commit.
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaE7H7sASct9KT20NZN:-NaEhIhD8a9n5GIpRjLu:bma8oze) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b2b542affcae4e5dee2c80e38d5e1d02ae57109/multibody/tree/spatial_inertia.cc#L42)):*

With this change, your documentation should be clear that expressions don't get validated. For every function that relies on this function, its documentation should be updated.
</blockquote></details>

Perhaps I am misinterpreting what you are saying.  Currently, ThrowUnlessVectorIsMagnitudeOne() is used 9 times.  Are you suggesting that public documentation for functions which use ThrowUnlessVectorIsMagnitudeOne() should clarify that expressions are not checked for vector's magnitude?   

If that is true, then why should the same not apply to DRAKE_DEMAND(), DRAKE_THROW_UNLESS(), etc.  For example, in rotation_matrix.cc, we use checks such as 

  void set(const Matrix3<T>& R) {
    DRAKE_ASSERT_VOID(ThrowIfNotValid(R));
    SetUnchecked(R);
  }

Here is the start of the code in rotation_matrix.cc
void RotationMatrix<T>::ThrowIfNotValid(const Matrix3<T>& R) {
  if constexpr (scalar_predicate<T>::is_bool) { ...

So, this seems to be the defacto standard for related functions.  
As they say in court "Your witness".
___
*[`multibody/tree/unit_inertia.cc` line 78 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3na7nC4jNAFFyGLjh:-NaEqwvE68MqTeYYafH4:bk1oher) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/unit_inertia.cc#L78)):*

You might consider a subsequent, superficial PR that goes through these files and removes them all so the whole document is stylistically consistent. I only mentioned it in *this* PR because it was highlighted and jumped out at me.  :)
</blockquote></details>

Yes -- that seems reasonable.  The younger Paul put T() in other places and I could do this all in one PR.
___
*[`multibody/tree/unit_inertia.cc` line 15 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaE9Y_m04ORwTGz3248:-NaEqYwp7sqwr-TbgGDF:bhhhe4u) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b2b542affcae4e5dee2c80e38d5e1d02ae57109/multibody/tree/unit_inertia.cc#L15)):*

nit: Same note about the functions that rely on this validation documenting that unit length isn't tested for `Expression`.
</blockquote></details>

See my response above.

SeanCurtis-TRI (commented):
:LGTM:

I know when to throw in the towel.

> Reviewed 2 of 2 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaFFW44-j4Id9uX4pub:bcq1a66)* status: LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/spatial_inertia.h` line 289 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-Na3deRp9szPwZ3DWJhW:-NaFBwMa8uiLVDNJMuJc:b9g0quz) ([raw file](https://github.com/RobotLocomotion/drake/blob/e95bfc0a7e0fd90cc202950d248657fbe3890caa/multibody/tree/spatial_inertia.h#L289)):*


I hope you and family are enjoying your time in Tokyo.  It is wonderful that you are taking time to explore.

For sake of linguistic argument (hopefully, this makes you smile):

1. You write: "My proposed text was directed toward giving an _intuition_ (whose more formal definition could be found via digging)."   I agree that "digging" helps clarify what is meant by Bp.

2. I'll argue that a world frame is inherent with "Bottom".  For Earth dwelling folks, "bottom" has to do with vertical and gravity (or if on a "flat" part of Earth, outward normal from Earth's mostly spherical surface).  Earth, Earth, Earth (World).

I mentioned that my mental image was of a cylinder rolling on a table.  However, the documentation makes no such reference.  Instead it only discusses the geometry of a cylinder and the unit vector -- both essential to defining Bp.  We both have a strong bias towards focusing on the cylinder's geometry.

3. Your linguistic lawyer seems so right.  An "axis" does not have endpoints. Perhaps a more explicitly clear set of words would be "an end-point of the cylinder's central longitudinal line segment."  

However, in view of the definition of the axis of cylinder [here](https://www.varsitytutors.com/hotmath/hotmath_help/topics/cylinder#:~:text=The%20axis%20of%20a%20cylinder,it%20is%20an%20oblique%20cylinder), the **axis** of a cylinder **is** the segment containing the centers of the two bases.

It seems reasonable to emphasize that understanding Bp requires some digging.
I will emphasize "digging" by enhancing the documentation to: 
  /// Creates a spatial inertia for a uniform-density solid cylinder B about an
  /// end-point Bp of the cylinder's axis (see below for more about Bp).

</blockquote></details>

If I gave you a cylindrical water bottle, you'd never argue that the bottom of the bottle is the side closest to the ground. You'd acknowledge there is an unambiguous top, regardless of how it was oriented. And *identifying* that top was central to the water bottle working right.

However, for all of the arguments about ancillary details, I've now resigned myself to the fact that the documentation will not inform a user's intuition. The *only* way to find out what P is is to carefully read through the whole thing. 
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaE7H7sASct9KT20NZN:-NaFClzSEpLG3HhoOTRt:b8yxfb6) ([raw file](https://github.com/RobotLocomotion/drake/blob/1b2b542affcae4e5dee2c80e38d5e1d02ae57109/multibody/tree/spatial_inertia.cc#L42)):*

Perhaps I am misinterpreting what you are saying.  Currently, ThrowUnlessVectorIsMagnitudeOne() is used 9 times.  Are you suggesting that public documentation for functions which use ThrowUnlessVectorIsMagnitudeOne() should clarify that expressions are not checked for vector's magnitude?   

If that is true, then why should the same not apply to DRAKE_DEMAND(), DRAKE_THROW_UNLESS(), etc.  For example, in rotation_matrix.cc, we use checks such as 

  void set(const Matrix3<T>& R) {
    DRAKE_ASSERT_VOID(ThrowIfNotValid(R));
    SetUnchecked(R);
  }

Here is the start of the code in rotation_matrix.cc
void RotationMatrix<T>::ThrowIfNotValid(const Matrix3<T>& R) {
  if constexpr (scalar_predicate<T>::is_bool) { ...

So, this seems to be the defacto standard for related functions.  
As they say in court "Your witness".
</blockquote></details>

Your understanding of my request is correct.

First of all, you know that "I found code in Drake that does this" has never been *de jure* proof that what you're doing is right. (And when it's code that you yourself have written, it's hardly an independent witness.)

What I'm saying is that the header currently says variants of the following multiple times:

```
/// @throws std::exception if mass, radius, or length is not positive and
/// finite or if ‖unit_vector‖ is not within 1.0E-14 of 1.0.
```

This is a documented contract for the function. And, when `T = Expression`, this statement doesn't hold. With the change made up above, this is now a fact.

The only question is, what do you do with this fact?

1. You can argue that it's generally understood that we don't validate mathematical properties against `Expression`. I'm open to the argument, but a cherry-picked example from your own code is not proof for that argument. That said, I'm always open to evidence-based persuasion.
2. Back out the change entirely and allow it to throw for `Expression` based purely on the presence of unbound variables with the argument that *that's* the generally understood practice vis a vis `Expression`.
3. You can change the documentation to reflect the truth.
4. Some third creative option that escapes me right now.

To be perfectly frank, I don't know which is more prevalent between 1 & 2. Or which is preferred. I don't spend enough time coding against `Expression`. (The joys of geometry.)

Perhaps the simplest thing for you to do is to defer it until you can ask a better-informed developer and reduce this PR to its simplest spelling.

My aesthetic preference was obviously to not throw when you have an `Expression` you can't validate, and be clear that you don't validate when you can't. But I can't argue that is *the* correct thing to do.

So, I'm going to mark myself satisfied and let you choose what appeals to you.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803)* status: LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

mitiguy (commented):
@SeanCurtis-TRI   -- thank you for the quick turn-around, much appreciated.
Signing out for the night.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaFHhNP4WCNM3-ZHrMj:b-uzz993)* status: LGTM missing from assignee xuchenhan-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

xuchenhan-tri (commented):
Platform :lgtm: 

> Reviewed 1 of 9 files at r1, 3 of 6 files at r2, 1 of 2 files at r3, 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaI9UAz2dBml_sQhDYW:b-cxwtpw)* status: 6 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/plant/test/joint_limits_test.cc` line 141 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaHzUat-27sG1KaFgem:-NaHzUat-27sG1KaFgen:b2ni8l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/plant/test/joint_limits_test.cc#L141)):*
> ```cpp
>   for (double time_step : {2.5e-4, 5.0e-4, 1.0e-3}) {
>     MultibodyPlant<double> plant(time_step);
>     // The rod's cm is at its centroid. The rod's body frame B is at the rod's
> ```

nit



_Suggestion:_
```cpp
COM?
```
___
*[`multibody/tree/spatial_inertia.cc` line 37 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI0GCR2U8TgeSyrBxB:-NaI0GCR2U8TgeSyrBxC:b2nhbfk) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/spatial_inertia.cc#L37)):*
> ```cpp
> }
> 
> // Throw unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
> ```

BTW



_Suggestion:_
```cpp
// Throws unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
```
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI16Wg5xVAKPrnitfp:-NaI16Wg5xVAKPrnitfq:b-4ieyie) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/spatial_inertia.cc#L42)):*
> ```cpp
> void ThrowUnlessVectorIsMagnitudeOne(const Vector3<T>& unit_vector,
>     std::string_view function_name) {
>   if constexpr (scalar_predicate<T>::is_bool) {
> ```

minor, it'd be nice to mention (in the documentation of this function) that this function is a no-op for T=symbolic::Expression.
___
*[`multibody/tree/unit_inertia.h` line 306 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI6-9FD2JHL7_Cs17y:-NaI6-9FD2JHL7_Cs17z:b-3s2scq) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/unit_inertia.h#L306)):*
> ```objc
>   /// @param[in] L The length of the cylinder.
>   /// @throws std::exception if r or L is negative.
>   DRAKE_DEPRECATED("2023-10-01",
> ```

nit three months from now (plus change) should be 11/1.



_Suggestion:_
```objc
2023-11-01
```
___
*[`multibody/tree/unit_inertia.cc` line 76 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI6lN88elQqlsqUSCZ:-NaI6lN88elQqlsqUSC_:b-qfcvcp) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/unit_inertia.cc#L76)):*
> ```cpp
>   DRAKE_THROW_UNLESS(radius >= 0);
>   DRAKE_THROW_UNLESS(length >= 0);
>   ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
> ```

minor Personally, I find DRAKE_THROW_UNLESS is to be better here e.g. something like
```
DRAKE_THROW_UNLESS(unit_vector.norm() - 1.0 < 1e-14 && 1.0 - unit_vector.norm() < 1e-14);

```
especially considering your hand crafted throw helper needs to deal with symbolic as a special case.

I'm fine with the current spelling if you prefer it, but we should be consistent in using this helper. (I'm looking at the implementation of SolidCapsule).



_Code quote:_
```cpp
ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
```
___
*[`multibody/tree/test/unit_inertia_test.cc` line 271 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI8txB5Om4i0a_li_B:-NaI8txB5Om4i0a_li_C:b2ni8l0) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/test/unit_inertia_test.cc#L271)):*
> ```cpp
> 
>   // Create G_BBp_B more directly via SolidCylinderAboutEnd().
>   const Vector3<double>unit_vec_A = Vector3<double>::UnitZ();
> ```

nit



_Suggestion:_
```cpp
const Vector3<double> unit_vec_A
```

SeanCurtis-TRI (commented):
> Reviewed 7 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803)* status: 5 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/unit_inertia.h` line 306 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI6-9FD2JHL7_Cs17y:-NaJb9QPEIskbuw1r4Uz:bwjdgrk) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/unit_inertia.h#L306)):*

nit three months from now (plus change) should be 11/1.
</blockquote></details>

Apparently, I can't count.  Thanks for catching it.
___
*[`multibody/tree/unit_inertia.cc` line 76 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI6lN88elQqlsqUSCZ:-NaJaoso6LsXDP8AwdQj:br6i60k) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/unit_inertia.cc#L76)):*

minor Personally, I find DRAKE_THROW_UNLESS is to be better here e.g. something like
```
DRAKE_THROW_UNLESS(unit_vector.norm() - 1.0 < 1e-14 && 1.0 - unit_vector.norm() < 1e-14);

```
especially considering your hand crafted throw helper needs to deal with symbolic as a special case.

I'm fine with the current spelling if you prefer it, but we should be consistent in using this helper. (I'm looking at the implementation of SolidCapsule).
</blockquote></details>

I believe, this is a stopgap until I can get around to finish my variant of `DRAKE_THROW_UNLESS` that allows you to print arbitrary values as part of the exception message. The value of printing out the normal provided is high enough to justify the extra infrastructure (for now).

Although, nerfing `DRAKE_THROW_UNLESS` for `T = Expression` is going to be an interesting problem generally.
___
*[`bindings/pydrake/multibody/BUILD.bazel` line 331 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaJacga4uQqtBtYVPo4:-NaJacga4uQqtBtYVPo5:b-xgbzad) ([raw file](https://github.com/RobotLocomotion/drake/blob/efc6b63b5a84259ebcaad987591c76a320db75f0/bindings/pydrake/multibody/BUILD.bazel#L331)):*
> ```Starlark
>     name = "install",
>     py_dest = PACKAGE_INFO.py_dest,
>     targets = PY_LIBRARIES + [":all_py"],
> ```

BTW I believe this change caused the CI failure.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

___
*[`multibody/plant/test/joint_limits_test.cc` line 141 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaHzUat-27sG1KaFgem:-NaJJCPx64CaCV-5h6sD:b-k3zeba) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/plant/test/joint_limits_test.cc#L141)):*

nit
</blockquote></details>

Good catch.  I'll be explicit and change cm to center of mass.
Note: I prefer CM to COM (or CG).
In any event, I should not have used a lower case cm.
___
*[`multibody/tree/spatial_inertia.cc` line 37 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI0GCR2U8TgeSyrBxB:-NaJL60_CnuYzeZNjlOY:bih6m9d) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/spatial_inertia.cc#L37)):*

BTW
</blockquote></details>

Done here and one other place.
___
*[`multibody/tree/spatial_inertia.cc` line 42 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI16Wg5xVAKPrnitfp:-NaJLkRQ9KAhjmE4IHoD:b-c59e2c) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/spatial_inertia.cc#L42)):*

minor, it'd be nice to mention (in the documentation of this function) that this function is a no-op for T=symbolic::Expression.
</blockquote></details>

Done here and in one other place.
___
*[`multibody/tree/unit_inertia.cc` line 76 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI6lN88elQqlsqUSCZ:-NaJOX2sE2RSsmAW7fM2:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/unit_inertia.cc#L76)):*

minor Personally, I find DRAKE_THROW_UNLESS is to be better here e.g. something like
```
DRAKE_THROW_UNLESS(unit_vector.norm() - 1.0 < 1e-14 && 1.0 - unit_vector.norm() < 1e-14);

```
especially considering your hand crafted throw helper needs to deal with symbolic as a special case.

I'm fine with the current spelling if you prefer it, but we should be consistent in using this helper. (I'm looking at the implementation of SolidCapsule).
</blockquote></details>

Done.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 271 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaI8txB5Om4i0a_li_B:-NaJOj1c7Y0UlSALacg7:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f2527a160516d4c27f52b9344c3533fbdd61bd84/multibody/tree/test/unit_inertia_test.cc#L271)):*

nit
</blockquote></details>

Done.
___
*[`bindings/pydrake/multibody/BUILD.bazel` line 331 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-NaJacga4uQqtBtYVPo4:-NaJjMhE7iMuh-rXDygO:b-vwzm5x) ([raw file](https://github.com/RobotLocomotion/drake/blob/efc6b63b5a84259ebcaad987591c76a320db75f0/bindings/pydrake/multibody/BUILD.bazel#L331)):*

BTW I believe this change caused the CI failure.
</blockquote></details>

Yes -- For some inexplicable reason, CLion reordered these lines when I was viewing it and made an innocuous change.  Xuchen did a sanity check with me as this was very odd - and used a different editor to revert the order.  I need to check on what my CLion settings are for Bazel files (strange) ...

mitiguy (commented):
+(status: squashing now) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803#-:-NaJkQnf03v2FsrozuuI:bm9o57a)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform)

xuchenhan-tri (commented):
> Reviewed 6 of 7 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19803)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),xuchenhan-tri(platform)

mitiguy:
PR #19803 helps resolve issue #18416.

PR #19807, opened 2023-07-19
BlockSparseCholeskySolver computes Schur complement

xuchenhan-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)

### DIFF
--- a/multibody/contact_solvers/block_sparse_cholesky_solver.cc
+++ b/multibody/contact_solvers/block_sparse_cholesky_solver.cc
@@ -34,35 +34,137 @@ void BlockSparseCholeskySolver<BlockType>::UpdateMatrix(
 
 template <typename BlockType>
 bool BlockSparseCholeskySolver<BlockType>::Factor() {
-  DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kAnalyzed);
-  for (int j = 0; j < L_->block_cols(); ++j) {
-    /* Update diagonal. */
-    const BlockType& Ajj = L_->diagonal_block(j);
-    L_diag_[j].compute(Ajj);
-    if (L_diag_[j].info() != Eigen::Success) {
-      solver_mode_ = SolverMode::kEmpty;
-      return false;
-    }
-    L_->SetBlockFlat(0, j, L_diag_[j].matrixL());
-    /* Update L₂₁ column.
-     | a₁₁  *  | = | λ₁₁  0 | * | λ₁₁ᵀ L₂₁ᵀ |
-     | a₂₁ a₂₂ |   | L₂₁ L₂₂|   |  0   L₂₂ᵀ |
-     So we have
-      L₂₁λ₁₁ᵀ = a₂₁, and thus
-      λ₁₁L₂₁ᵀ = a₂₁ᵀ */
-    const std::vector<int>& row_blocks = L_->block_row_indices(j);
-    const auto Ljj = L_diag_[j].matrixL();
-    /* We start from flat = 1 here to skip the j,j diagonal entry. */
-    for (int flat = 1; flat < ssize(row_blocks); ++flat) {
-      const BlockType& Aij = L_->block_flat(flat, j);
-      BlockType Lij = Ljj.solve(Aij.transpose()).transpose();
-      L_->SetBlockFlat(flat, j, std::move(Lij));
-    }
-    /* Update L₂₂ according to L₂₂ = a₂₂ - L₂₁⋅L₂₁ᵀ. */
-    RightLookingSymmetricRank1Update(j);
+  DRAKE_THROW_UNLESS(solver_mode_ == SolverMode::kAnalyzed);
+  const bool success = CalcPartialFactorization(0, L_->block_cols());
+  solver_mode_ = success ? SolverMode::kFactored : SolverMode::kEmpty;
+  return success;
+}
+
+template <typename BlockType>
+std::optional<MatrixX<double>>
+BlockSparseCholeskySolver<BlockType>::FactorAndCalcSchurComplement(
+    const SymmetricMatrix& A,
+    const std::unordered_set<int>& eliminated_blocks) {
+  const int num_eliminated_blocks = ssize(eliminated_blocks);
+  const int num_total_blocks = A.block_cols();
+  const int num_remaining_blocks = num_total_blocks - num_eliminated_blocks;
+  DRAKE_DEMAND(num_remaining_blocks >= 0);
+  if (num_remaining_blocks == 0) {
+    SetMatrix(A);
+    const bool success = Factor();
+    return success ? std::optional<MatrixX<double>>(MatrixX<double>::Zero(0, 0))
+                   : std::nullopt;
   }
-  solver_mode_ = SolverMode::kFactored;
-  return true;
+
+  /* The Schur complement of the `eliminated_blocks` appear on the bottom
+   right corner of the matrix under Cholesky factorization when the
+   `eliminated_blocks` have been factorized first with a right-looking
+   Cholesky. So we need an elimination ordering that ensures that the blocks
+   associated with `eliminated_blocks` are eliminated first. There are
+   many ordering that satisfy this requirement, and we look for one that reduces
+   fill-in. */
+  const std::vector<int> elimination_ordering =
+      ComputeMinimumDegreeOrdering(A.sparsity_pattern(), eliminated_blocks);
+  SetMatrixImpl(A, elimination_ordering,
+                SymbolicFactor(A, elimination_ordering));
+
+  /* Reset solver mode and exit if factorization of the eliminated blocks fails.
+   */
+  if (!CalcPartialFactorization(0, num_eliminated_blocks)) {
+    solver_mode_ = SolverMode::kEmpty;
+    return std::nullopt;
+  }
+
+  const MatrixX<double> permuted_S =
+      L_->MakeDenseBottomRightCorner(num_remaining_blocks)
+          .template selfadjointView<Eigen::Lower>();
+
+  /* Finish the factorization. If the factorization fails, set the solver mode
+   and exit. */
+  bool success =
+      CalcPartialFactorization(num_eliminated_blocks, L_->block_cols());
+  solver_mode_ = success ? SolverMode::kFactored : SolverMode::kEmpty;
+  if (!success) {
+    return std::nullopt;
+  }
+
+  /* The Schur complement computed here is permuted according to the elimination
+   ordering. We need to invert this permutation to obtain the original Schur
+   complement. To do that, we need to find the _scalar_ permutation for the
+   remaining blocks, and for that we need the _block_ permutation of the
+   remaining blocks as well as the sizes of those blocks. */
+
+  /* The block elimination ordering implies a permutation of the remaining
+   blocks. For example, suppose the elimination ordering is
+   [0, 5, 2, 4, 6, 3, 1], eliminated blocks are {0, 2, 5}, and the remaining
+   blocks are {1, 3, 4, 6} (notice how the eliminated blocks come before
+   remaining blocks). Then the induced elimination ordering on the remaining
+   blocks is [2, 3, 1, 0]. To see that, note that 1, 3, 4, 6 are the 0th, 1st,
+   2nd, 3rd elements in the remaining group; the order that they are eliminated
+   is
+
+     4 (2nd element in remaining blocks).
+     6 (3rd element in remaining blocks).
+     3 (1st element in remaining blocks).
+     1 (0th element in remaining blocks).
+
+   We name this elimination ordering on the remaining blocks
+   `remaining_ordering`. We also store the mapping from global indices to
+   remaining indices in `global_to_remaining` where a value of -1 indicates that
+   the element is eliminated.
+
+   In particular, in the above example, `remaining ordering` is [2, 3, 1, 0] and
+   `global_to_remaining` is [-1, 0, -1, 1, 2, -1, 3]. Note that
+   `global_to_remaining` assigns the set {1, 3, 4, 6} indices in ascending
+   order. */
+  std::vector<int> global_to_remaining(num_total_blocks, -1);
+  const std::vector<int>& block_sizes = A.sparsity_pattern().block_sizes();
+  std::vector<int> remaining_block_sizes;
+  remaining_block_sizes.reserve(num_remaining_blocks);
+  int remaining_index = 0;
+  for (int i = 0; i < num_total_blocks; ++i) {
+    if (eliminated_blocks.count(i) == 0) {
+      global_to_remaining[i] = remaining_index++;
+      remaining_block_sizes.emplace_back(block_sizes[i]);
+    }
+  }
+  /* The scalar index of the first scalar in each remaining block. */
+  std::vector<int> remaining_starting_indices(num_remaining_blocks);
+  remaining_starting_indices[0] = 0;
+  for (int i = 1; i < num_remaining_blocks; ++i) {
+    remaining_starting_indices[i] =
+        remaining_starting_indices[i - 1] + remaining_block_sizes[i - 1];
+  }
+  /* The block elimination ordering within remaining blocks. */
+  std::vector<int> remaining_ordering;
+  remaining_ordering.reserve(num_remaining_blocks);
+  /* We start with permuted_i = ssize(eliminated_blocks) to skip all
+   eliminated blocks. */
+  for (int permuted_i = ssize(eliminated_blocks);
+       permuted_i < ssize(elimination_ordering); ++permuted_i) {
+    const int i = elimination_ordering[permuted_i];
+    DRAKE_DEMAND(global_to_remaining[i] >= 0);
+    remaining_ordering.emplace_back(global_to_remaining[i]);
+  }
+  const int num_remaining_scalars =
+      remaining_starting_indices.back() + remaining_block_sizes.back();
+  /* Build the scalar elimination ordering (for scalars in the remaining blocks)
+   from the block elimination ordering. */
+  VectorX<int> scalar_remaining_ordering(num_remaining_scalars);
+  int new_scalar_index = 0;
+  for (int i = 0; i < ssize(remaining_ordering); ++i) {
+    const int block = remaining_ordering[i];
+    const int start = remaining_starting_indices[block];
+    const int size = remaining_block_sizes[block];
+    for (int s = start; s < start + size; ++s) {
+      scalar_remaining_ordering(new_scalar_index++) = s;
+    }
+  }
+  /* The elimination ordering is the inverse of the permutation. */
+  const auto& inverse_scalar_permutation = scalar_remaining_ordering;
+  Eigen::PermutationMatrix<Eigen::Dynamic> P_inv(inverse_scalar_permutation);
+  const auto P = P_inv.transpose();
+  return P_inv * permuted_S * P;
 }
 
 template <typename BlockType>
@@ -230,6 +332,41 @@ BlockSparsityPattern BlockSparseCholeskySolver<BlockType>::SymbolicFactor(
    new ordering. */
   return contact_solvers::internal::SymbolicCholeskyFactor(
       BlockSparsityPattern(permuted_block_sizes, permuted_sparsity_pattern));
+}
+
+template <typename BlockType>
+bool BlockSparseCholeskySolver<BlockType>::CalcPartialFactorization(
+    int starting_col_block, int ending_col_block) {
+  DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kAnalyzed);
+  DRAKE_DEMAND(starting_col_block >= 0 &&
+               starting_col_block <= L_->block_cols());
+  DRAKE_DEMAND(ending_col_block >= 0 && ending_col_block <= L_->block_cols());
+  for (int j = starting_col_block; j < ending_col_block; ++j) {
+    /* Update diagonal. */
+    const BlockType& Ajj = L_->diagonal_block(j);
+    L_diag_[j].compute(Ajj);
+    if (L_diag_[j].info() != Eigen::Success) {
+      return false;
+    }
+    L_->SetBlockFlat(0, j, L_diag_[j].matrixL());
+    /* Update L₂₁ column.
+     | a₁₁  *  | = | λ₁₁  0 | * | λ₁₁ᵀ L₂₁ᵀ |
+     | a₂₁ a₂₂ |   | L₂₁ L₂₂|   |  0   L₂₂ᵀ |
+     So we have
+      L₂₁λ₁₁ᵀ = a₂₁, and thus
+      λ₁₁L₂₁ᵀ = a₂₁ᵀ */
+    const std::vector<int>& row_blocks = L_->block_row_indices(j);
+    const auto Ljj = L_diag_[j].matrixL();
+    /* We start from flat = 1 here to skip the j,j diagonal entry. */
+    for (int flat = 1; flat < ssize(row_blocks); ++flat) {
+      const BlockType& Aij = L_->block_flat(flat, j);
+      BlockType Lij = Ljj.solve(Aij.transpose()).transpose();
+      L_->SetBlockFlat(flat, j, std::move(Lij));
+    }
+    /* Update L₂₂ according to L₂₂ = a₂₂ - L₂₁⋅L₂₁ᵀ. */
+    RightLookingSymmetricRank1Update(j);
+  }
+  return true;
 }
 
 template <typename BlockType>
--- a/multibody/contact_solvers/block_sparse_cholesky_solver.h
+++ b/multibody/contact_solvers/block_sparse_cholesky_solver.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <memory>
+#include <optional>
+#include <unordered_set>
 #include <vector>
 
 #include "drake/common/copyable_unique_ptr.h"
@@ -120,6 +122,47 @@ class BlockSparseCholeskySolver {
    and is SolverMode::kEmpty otherwise. */
   [[nodiscard]] bool Factor();
 
+  /* Computes the block sparse Cholesky factorization of the given matrix `A`.
+   Returns std::nullopt if the factorization fails. Failure is triggered by an
+   internal failure of Eigen::LLT. This can fail if, for instance, the input
+   matrix `A` is not positive definite. If failure is encountered, the user
+   should verify that the specified matrix is positive definite and not poorly
+   conditioned.
+
+   In addition, this function computes the Schur complement matrix of the input
+   matrix A in the following sense:
+
+   Let E be the set of `eliminated_blocks`. We define permutation p on the block
+   indices of `A` such that p(i) < p(j) iff
+
+    (1) i ∈ E and j ∉ E or
+    (2) i ∈ E and j ∈ E and i < j or
+    (3) i ∉ E and j ∉ E and i < j.
+
+   For example, if A has 7 block rows/columns and E = {2, 5, 6}, then
+
+    [p(0), p(1), p(2), p(3), p(4), p(5), p(6)] =
+    [3,    4,    0,    5,    6,    1,    2]
+
+   We then define Â = P⋅A⋅Pᵀ, where P is the permutation matrix representation
+   of p (i.e. Pᵢⱼ = 1 if i = p(j) and Pᵢⱼ = 0 otherwise). In other words, Â is
+   permuted from A so that blocks with indices in `eliminated_blocks` appear on
+   the top left corner of the matrix Â and all other blocks appear on the bottom
+   right corner of the matrix. The matrix Â can be written in block form as Â =
+   [D, B; Bᵀ, C] where D corresponds to the blocks with indices in E, C
+   corresponds to the blocks with indices outside of E, and B is the resulting
+   off-diagonal block from the permutation.
+
+   If the fatorization of A is successful, returns the Schur complement
+   S = C - BᵀD⁻¹B.
+
+   @pre `eliminated_blocks` has all its entries in [0, A.block_cols()).
+   @post solver_mode() is SolverMode::kFactored if factorization is successful
+   and is SolverMode::kEmpty otherwise. */
+  [[nodiscard]] std::optional<MatrixX<double>> FactorAndCalcSchurComplement(
+      const SymmetricMatrix& A,
+      const std::unordered_set<int>& eliminated_blocks);
+
   /* Solves the system A⋅x = b and returns x.
    @throws std::exception if b.size() is incompatible with the size of the
    matrix set by SetMatrix().
@@ -185,6 +228,16 @@ class BlockSparseCholeskySolver {
                                     A.block_cols() - 1}. */
   BlockSparsityPattern SymbolicFactor(
       const SymmetricMatrix& A, const std::vector<int>& elimination_ordering);
+
+  /* Factorizes matrix A but in particular only processes a range of columns.
+   If the range of columns is [0, L_.block_cols()), then this method performs a
+   full factorization of A. Otherwise, this leaves the underlying factorization
+   in an intermediate state and therefore successive calls to this method
+   must be performed with care.
+   @note this function does not modify solver mode.
+   @pre solver_mode() == kAnalyzed.
+   @pre 0 <= starting_col_block <= ending_col_block <= L.block_cols(). */
+  bool CalcPartialFactorization(int starting_col_block, int ending_col_block);
 
   /* Performs L(j+1:, j+1:) -= L(j+1:, j) * L(j+1:, j).transpose().
    @pre 0 <= j < L.block_cols(). */
--- a/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc
+++ b/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc
@@ -1,6 +1,7 @@
 #include "drake/multibody/contact_solvers/block_sparse_cholesky_solver.h"
 
 #include <memory>
+#include <numeric>
 #include <utility>
 
 #include <gtest/gtest.h>
@@ -14,6 +15,7 @@ namespace contact_solvers {
 namespace internal {
 namespace {
 
+using Eigen::Matrix3d;
 using Eigen::Matrix4d;
 using Eigen::MatrixXd;
 using Eigen::MatrixXi;
@@ -125,22 +127,40 @@ GTEST_TEST(BlockSparseCholeskySolverTest, Solve) {
   EXPECT_TRUE(CompareMatrices(x4, expected_x4, 1e-13));
 }
 
-GTEST_TEST(BlockSparseCholeskySolverTest, SolveFailureDueToNonSpdness) {
+GTEST_TEST(BlockSparseCholeskySolverTest, FailureDueToNonSpdness) {
   std::vector<std::vector<int>> sparsity;
-  sparsity.emplace_back(std::vector<int>{0});
-  std::vector<int> block_sizes = {4};
+  sparsity.emplace_back(std::vector<int>{0, 1});
+  sparsity.emplace_back(std::vector<int>{1});
+  std::vector<int> block_sizes = {4, 3};
   BlockSparsityPattern block_pattern(block_sizes, sparsity);
   BlockSparseSymmetricMatrix A(std::move(block_pattern));
   A.AddToBlock(0, 0, -Matrix4d::Identity());
+  A.AddToBlock(1, 1, -Matrix3d::Identity());
 
   BlockSparseCholeskySolver<MatrixXd> solver;
   solver.SetMatrix(A);
   EXPECT_EQ(solver.solver_mode(),
             BlockSparseCholeskySolver<MatrixXd>::SolverMode::kAnalyzed);
-  const bool success = solver.Factor();
-  EXPECT_FALSE(success);
-  EXPECT_EQ(solver.solver_mode(),
-            BlockSparseCholeskySolver<MatrixXd>::SolverMode::kEmpty);
+  /* Calling Factor() fails. */
+  {
+    const bool success = solver.Factor();
+    EXPECT_FALSE(success);
+    EXPECT_EQ(solver.solver_mode(),
+              BlockSparseCholeskySolver<MatrixXd>::SolverMode::kEmpty);
+  }
+  /* Calling FactorAndCalcSchurComplement() also fails. */
+  {
+    for (const std::unordered_set<int>& eliminated_blocks :
+         std::vector<std::unordered_set<int>>{std::unordered_set<int>{},
+                                              std::unordered_set<int>{0},
+                                              std::unordered_set<int>{0, 1}}) {
+      const std::optional<MatrixXd> schur_complement =
+          solver.FactorAndCalcSchurComplement(A, eliminated_blocks);
+      EXPECT_FALSE(schur_complement.has_value());
+      EXPECT_EQ(solver.solver_mode(),
+                BlockSparseCholeskySolver<MatrixXd>::SolverMode::kEmpty);
+    }
+  }
 }
 
 GTEST_TEST(BlockSparseCholeskySolverTest, FactorBeforeSetMatrixThrows) {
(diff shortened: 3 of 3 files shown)


xuchenhan-tri (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-N_fbaDoBSdFewOag24v:b-57wv7n)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

amcastro-tri (commented):
working... Almost done.

> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaCdiG43KhuG4hzoOFK:b-tsb7u0)* status: 7 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 132 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCZntH7kRjecs46hdj:-NaCZntH7kRjecs46hdk:b-7ihjxn) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L132)):*
minor, consider `FactorAndCalcSchurComplement()`

> **In addition**, this function computes the Schur complement

I like how you documented this because that's how I think about it. The Schur complement is an intermediate result obtained during the main computation, the factorization.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCc1Fu2wAGLbBiJeSf:-NaCc1Fu2wAGLbBiJeSg:b8irox0) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*
> ```objc
>     (1) i ∈ E and j ∉ E or
>     (2) i ∈ E and j ∈ E and i < j or
>     (3) i ∉ E and j ∉ E and i < j.
> ```

I obviously understand that `p` corresponds to a permutation so that all eliminated blocks would appear first on `M̂`.
Now, I have trouble parsing these three condition. Not sure because I am slow or because there is something not quite right. Is there another way to write this? 
In the meantime I'll reverse engineer the source to see if I understand it. 
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCcSbv-18JTmDXyYi7:-NaCcSbv-18JTmDXyYi8:b-1a47ii) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*
> ```objc
>     (3) i ∉ E and j ∉ E and i < j.
> 
>    We then define M̂ = P⋅M⋅Pᵀ. In other words, M̂ is permuted from M so that
> ```

nit, 

consider defining `P` since not obvious how you obtain it from `p`, that's just a convention that I believe must be documented. I believe in this case it'd be `Pᵢⱼ = 1` for `j = p(i)` and `Pᵢⱼ = 0` otherwise? (double-check, I might have gotten it reversed)
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCbNXV7wd5iDatNmmt:-NaCbNXV7wd5iDatNmmu:b-lf5620) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L154)):*
> ```objc
>    @pre `eliminated_blocks` has all its entries in [0, M.block_cols()).
>    @post solver_mode() is SolverMode::kFactored if factorization is successful
>    and is SolverMode::kEmpty otherwise. */
> ```

`kEmpty` corresponds to "no one set a matrix yet". Here we are in the condition where someone did set a matrix but it just happens to be a bad matrix.
My preference here would be to leave the mode unchanged if the factorization failed (whether the mode was kEmpty or kAnalyzed before the call to this method). 

WDYT?



_Code quote:_
```objc
 SolverMode::kEmpty
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 228 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCdLVm9cq--YmM3a4F:-NaCdLVm9cq--YmM3a4G:b-qc9fbu) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L228)):*
> ```objc
>    the matrix. Does not set the solver mode. Unless
>    [starting_col_block, ending_col_block) is equal to [0, L.block_cols()), this
>    may leave L in an intermediate state. Call this function with care.
> ```

I haven't read the source, but these docs would seem to indicate that this "intermediate state" is the most common situation?
Bottom line, these docs don't tell me what this method actually does. Could you be more precise? In particular, what's the relationship to Factor() and CalcSchurComplementAndFactor()?
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCYTYyDigqKC7cS-zm:-NaCYTYz7Y-7-S4BaFF7:bruzqp4) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L129)):*
> ```cpp
> }
> 
> GTEST_TEST(BlockSparseCholeskySolverTest, FailureDueToNonSpdness) {
> ```

nit, maybe update either test name, setup or add minimum description.
Right now it looks like you have a zero diagonal block. Obviously that leads to non-spdness, but somehow in my mind non-spdness and having a zero diagonal block fall into two different categories. 
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCalgr77nKtIlXmJTh:-NaCalgr77nKtIlXmJTi:b-v2ybtu) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L139)):*
> ```cpp
> 
>   BlockSparseCholeskySolver<MatrixXd> solver;
>   solver.SetMatrix(A);
> ```

The precondition for `SetMatrix()` is for A to be SPD. That's impossible to verify without performing the factorization itself, and therefore the need for this precondition. That makes sense.

Now, shouldn't this method throw if we give it a matrix without a diagonal block? that's something the symbolic analysis can verify.

amcastro-tri (commented):
Before I continue further into the implementation, I'd like to understand the result of `CalcSchurComplementAndFactor()`, for which I'd need some update on the docs. Without diving into the implementation, it is not clear to me in what order the indexes in the returned `S` are. I believe it is not determined by the user, since we only supply an `unordered_set` and therefore it'd seem we allow the implementation to choose the ordering. In particular, when I use S later on, wouldn't I as the user like to know that ordering? and wouldn't I need to be able to retrieve it? (maybe next PR, but I if so, it should already show in these docs).

I think that if I understand the output of `CalcSchurComplementAndFactor()` I can quickly finish the review of its implementation.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaHloJ42l90hFev2GS5:bfx9dqd)* status: 9 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHU1of-vxz9slomvIU:-NaHU1of-vxz9slomvIV:b-3akxo4) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*
> ```objc
>     (3) i ∉ E and j ∉ E and i < j.
> 
>    We then define M̂ = P⋅M⋅Pᵀ. In other words, M̂ is permuted from M so that
> ```

comment in the notation here. In all other documentation in this file, the matrix to be factorized is called A, not M. 
Consider renamting M to A and in the block form consider other symbol like `Â = \[B, C; Cᵀ, D\]` along with `S = D - CᵀB⁻¹C`
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 228 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCdLVm9cq--YmM3a4F:-NaHHfiG2c8_lmMFqtOx:b-n4966s) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L228)):*

I haven't read the source, but these docs would seem to indicate that this "intermediate state" is the most common situation?
Bottom line, these docs don't tell me what this method actually does. Could you be more precise? In particular, what's the relationship to Factor() and CalcSchurComplementAndFactor()?
</blockquote></details>

Reading the source it's obvious what this does. Still I believe it should be documented. Maybe even rename to something like "CalcPartialPermutation()" or something that indicates that this only factorizes a portion of the matrix.

Probably the docs could say something along the lines of "Factorizes matrix A but in particular it only processes a range of columns. If the range of columns is [0, L_.block\_cols()), then this method performs a full factorization of A. This leaves the underlying factorization L_ in an intermediate state and therefore successive calls to this method must be performed with care".
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 112 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHhd5I9KXkgGr6LMPh:-NaHhd5I9KXkgGr6LMPi:b-jud2yv) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L112)):*
> ```cpp
>    We name this elimination ordering on the remaining blocks
>    `remaining_ordering`. We also store the mapping from global indices to
>    remaining indices in `global_to_remaining`. */
> ```

just to be clear, please state what these two arrays are for your example.
I believe `remaining_ordering = [2, 3, 1, 0]`, but from this description I couldn't tell what `global_to_remaining` would be, or what size it'd have. From the source it'd seem `global_to_remaining = [-1, 0, -1, 1, 2, -1, 3]`?

If so, probably a good idea to state that `global_to_remainding` assigns the set `{1, 3, 4, 6}` indexes in ascending order.

BTW, these docs with examples are great! thank you.

xuchenhan-tri (commented):
I addressed the comments. PTAL, thanks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaIOrQJB1tkmgn3Jg6e:bsx6mhw)* status: 6 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 132 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCZntH7kRjecs46hdj:-NaICDMC3utazEm-8yN2:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L132)):*

minor, consider `FactorAndCalcSchurComplement()`

> **In addition**, this function computes the Schur complement

I like how you documented this because that's how I think about it. The Schur complement is an intermediate result obtained during the main computation, the factorization.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCc1Fu2wAGLbBiJeSf:-NaICV6D80mVAFG5UVm2:bsttkym) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*

I obviously understand that `p` corresponds to a permutation so that all eliminated blocks would appear first on `M̂`.
Now, I have trouble parsing these three condition. Not sure because I am slow or because there is something not quite right. Is there another way to write this? 
In the meantime I'll reverse engineer the source to see if I understand it. 
</blockquote></details>

The permutation moves all eliminated blocks before non-eliminated blocks and do so in a way that the ordering within eliminate/non-eliminated blocks do not change.

I wanted to precisely define this permutation as concisely as possible but it looks like it failed to be clear. Let me know

1. if you now understand the definition, and
2. if you have a suggestion on how to improve.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCcSbv-18JTmDXyYi7:-NaIFkzX-kYfcJyhAp-x:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*

nit, 

consider defining `P` since not obvious how you obtain it from `p`, that's just a convention that I believe must be documented. I believe in this case it'd be `Pᵢⱼ = 1` for `j = p(i)` and `Pᵢⱼ = 0` otherwise? (double-check, I might have gotten it reversed)
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHU1of-vxz9slomvIU:-NaIG3iRC89BLWfJoRxS:b-95txdl) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*

comment in the notation here. In all other documentation in this file, the matrix to be factorized is called A, not M. 
Consider renamting M to A and in the block form consider other symbol like `Â = \[B, C; Cᵀ, D\]` along with `S = D - CᵀB⁻¹C`
</blockquote></details>


Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCbNXV7wd5iDatNmmt:-NaIG9x31muzkSgV_ctQ:brkiygy) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L154)):*

`kEmpty` corresponds to "no one set a matrix yet". Here we are in the condition where someone did set a matrix but it just happens to be a bad matrix.
My preference here would be to leave the mode unchanged if the factorization failed (whether the mode was kEmpty or kAnalyzed before the call to this method). 

WDYT?
</blockquote></details>

`kAnalyzed` means is the matrix symbolically analyzed and ready to be factored. If an factorization failed, the data in the matrix is already polluted and one has to reset the matrix.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 228 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCdLVm9cq--YmM3a4F:-NaIGeDg79etO0RTVrrN:b-4g279r) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L228)):*
> Maybe even rename to something like "CalcPartialPermutation()" or something that indicates that this only factorizes a portion of the matrix.

I suppose you meant `CalcPartialFactorization`? I'm fine with that if that's what you meant but want to confirm with you.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 112 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHhd5I9KXkgGr6LMPh:-NaIOoOb5Q7QUI6rgiby:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L112)):*

just to be clear, please state what these two arrays are for your example.
I believe `remaining_ordering = [2, 3, 1, 0]`, but from this description I couldn't tell what `global_to_remaining` would be, or what size it'd have. From the source it'd seem `global_to_remaining = [-1, 0, -1, 1, 2, -1, 3]`?

If so, probably a good idea to state that `global_to_remainding` assigns the set `{1, 3, 4, 6}` indexes in ascending order.

BTW, these docs with examples are great! thank you.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCYTYyDigqKC7cS-zm:-NaIHguV3H34ivlFvFY7:b-az7swr) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L129)):*

nit, maybe update either test name, setup or add minimum description.
Right now it looks like you have a zero diagonal block. Obviously that leads to non-spdness, but somehow in my mind non-spdness and having a zero diagonal block fall into two different categories. 
</blockquote></details>

Done

To be clear, we don't want to check numerical values of the diagonal blocks. That's too expensive. Here, symbolically, the diagonal blocks do exist. Only the numerical values are zero.

I can see where you confusion comes from though. Modified the test.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCalgr77nKtIlXmJTh:-NaIII9d7y0JIDndgtlQ:bw33qv6) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L139)):*

The precondition for `SetMatrix()` is for A to be SPD. That's impossible to verify without performing the factorization itself, and therefore the need for this precondition. That makes sense.

Now, shouldn't this method throw if we give it a matrix without a diagonal block? that's something the symbolic analysis can verify. 
</blockquote></details>

My understanding of the request is to fail fast in the case that the input A has an empty diagonal.

However, it's impossible to create such an A in the first place. See documentation of `contact_solvers::internal::BlockSparsityPattern`.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHU1of-vxz9slomvIU:-NaIbfmp4hPjPCC6P0cP:b-tgu3dd) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*

Done
</blockquote></details>

not sure if intentional, I still see M?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 154 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCbNXV7wd5iDatNmmt:-NaIcNxT3JmUj0otDJg0:b-voaot3) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L154)):*

`kAnalyzed` means is the matrix symbolically analyzed and ready to be factored. If an factorization failed, the data in the matrix is already polluted and one has to reset the matrix.
</blockquote></details>

ok, fair enough. 
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 228 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCdLVm9cq--YmM3a4F:-NaIcXX449VPsBB411P7:b9oksu0) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L228)):*

> Maybe even rename to something like "CalcPartialPermutation()" or something that indicates that this only factorizes a portion of the matrix.

I suppose you meant `CalcPartialFactorization`? I'm fine with that if that's what you meant but want to confirm with you.
</blockquote></details>

oh, yes, sorry, I meant `CalcPartialFactorization()`. 
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 129 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCYTYyDigqKC7cS-zm:-NaIdPKf36od4nibWmwS:b-zdrz9o) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L129)):*

Done

To be clear, we don't want to check numerical values of the diagonal blocks. That's too expensive. Here, symbolically, the diagonal blocks do exist. Only the numerical values are zero.

I can see where you confusion comes from though. Modified the test.
</blockquote></details>

oh, I didn't realize (or remembered, since in principle I already reviewed this :-/ ) that by default those blocks existed.
Thanks, the update is very clear.
___
*[`multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCalgr77nKtIlXmJTh:-NaIdeOI-WAmKvGX4rRv:b-9vk82g) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/test/block_sparse_cholesky_solver_test.cc#L139)):*

My understanding of the request is to fail fast in the case that the input A has an empty diagonal.

However, it's impossible to create such an A in the first place. See documentation of `contact_solvers::internal::BlockSparsityPattern`.
</blockquote></details>

ah! that's what I was missing. Thanks!

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCc1Fu2wAGLbBiJeSf:-NaIhpIrEENabFLsQqsk:b-g4m04v) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*

The permutation moves all eliminated blocks before non-eliminated blocks and do so in a way that the ordering within eliminate/non-eliminated blocks do not change.

I wanted to precisely define this permutation as concisely as possible but it looks like it failed to be clear. Let me know

1. if you now understand the definition, and
2. if you have a suggestion on how to improve.
</blockquote></details>

Not yet. I think the notation is also getting on my way. Lets see if I understand some basics:
 1. Per your definition: M̂ = P⋅M⋅Pᵀ
 2. I believe that implies that `i = p(i_permuted)`
 3. Therefore, in you axioms above, saying  `i ∈ E ` doesn't work because `i` actually belongs to the "permuted indexes" while E is the set of "original indexes"?
 4. "The permutation moves all eliminated blocks before non-eliminated blocks". That's as much as I could tell. But beyond that, the ordering is arbitrary right? If I will permute anyway, why to preserve the ordering within and outside E? maybe it'd just be easier to keep the minimum degree ordering and call it a day? not suggesting it's a good idea, but trying: 
  a. Understand your choice (I still don't)
  b. Understand if the calling code will need to know this convention to permute back solutions to equations involing the Schur-complement.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCc1Fu2wAGLbBiJeSf:-NaIk5FY1GuO8kT1U5vQ:b-omc7cz) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*
> 1. Per your definition: M̂ = P⋅M⋅Pᵀ

Yes.

> 2. I believe that implies that i = p(i_permuted)

No i_permuted = p(i). Otherwise the semantic is very confusing!

> 3. Therefore, in you axioms above, saying `i ∈ E` doesn't work because `i` actually belongs to the "permuted indexes" while E is the set of "original indexes"?

`i` belongs to original indices.

4> ... beyond that, the ordering is arbitrary right?

No, like I said earlier in this thread, the permutation is done in a way that the **ordering within eliminate/non-eliminated blocks do not change**. This uniquely defines the permutation.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 142 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaHU1of-vxz9slomvIU:-NaIibsiAAGqPLgR9ppw:b-mhus97) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L142)):*

not sure if intentional, I still see M?
</blockquote></details>

Ah sorry I misread the comment. Done now.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 228 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaCdLVm9cq--YmM3a4F:-NaIjCio4WX7phZwR0HQ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/2711e1629e412c0bc354b33901a344fc2e4c8b91/multibody/contact_solvers/block_sparse_cholesky_solver.h#L228)):*

oh, yes, sorry, I meant `CalcPartialFactorization()`. 
</blockquote></details>

Done

amcastro-tri (commented):
Thanks for the patience @xuchenhan-tri .

Just a few comments in the docs.

+@EricCousineau-TRI for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaNpJEf6ptE9MoTcFfr:b-452je1)* status: 5 unresolved discussions, LGTM missing from assignees amcastro-tri,EricCousineau-TRI(platform)

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 133 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNgQup6IpB6Hwm99ee:-NaNgQup6IpB6Hwm99ef:undefined) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L133)):*
> ```objc
> 
>    In addition, this function computes the Schur complement matrix of the input
>    matrix M in the following sense:
> ```





_Suggestion:_
```objc
matrix A
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNlbFvDFoWFyiIbx4e:-NaNlbFw8SGqvJp5pWTg:b-zdvosl) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*
> ```objc
>     (1) i ∈ E and j ∉ E or
>     (2) i ∈ E and j ∈ E and i < j or
>     (3) i ∉ E and j ∉ E and i < j.
> ```

I get it now, but it took some parsing. Consider adding an example.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 143 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNkIpA2qSLTtCZq1Jw:-NaNhgywDt-iKGOxhEQk:btodbn4) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L143)):*
> ```objc
> 
>    We then define Â = P⋅A⋅Pᵀ, where P is the permutation matrix representation
>    of p (i.e. Pᵢⱼ = 1 if j = p(i) and Pᵢⱼ = 0 otherwise). In other words, Â is
> ```

I believe this is supposed to be the opposite?
Consider a simple permutation `x̂ = P⋅x`



_Suggestion:_
```objc
   We then define Â = P⋅A⋅Pᵀ, where P is the permutation matrix representation
   of p (i.e. Pᵢⱼ = 1 iff i = p(j) and Pᵢⱼ = 0 otherwise). In other words, Â is
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 144 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNl0ge4nSWbB1E8GUI:-NaNl0ge4nSWbB1E8GUJ:b-d202it) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L144)):*
> ```objc
>    We then define Â = P⋅A⋅Pᵀ, where P is the permutation matrix representation
>    of p (i.e. Pᵢⱼ = 1 if j = p(i) and Pᵢⱼ = 0 otherwise). In other words, Â is
>    permuted from M so that blocks with indices in `eliminated_blocks` appear on
> ```

here and below we still have a few `M`



_Code quote:_
```objc
from M
```
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 78 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNmMLz6r1mIxfAtveW:-NaNmMLz6r1mIxfAtveX:b-7rzw04) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L78)):*
> ```cpp
>   }
> 
>   const MatrixX<double> permuted_S =
> ```

nit, consider adding a TODO here for a future in which S is a block sparse matrix.

amcastro-tri (commented):
ah! :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaNpUJU0I44RIxMoeIF:bpxpmd5)* status: 5 unresolved discussions, LGTM missing from assignee EricCousineau-TRI(platform)

EricCousineau-TRI (commented):
:lgtm: platform w/ uber small nits

> Reviewed 1 of 3 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NaOEMs_E8TNTsdmL-Wt:b2eoitr)* status: 7 unresolved discussions

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 147 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaOD1AYBh2Uu5326paD:-NaOD1AZCT63cRczunV7:b-rzgrbg) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L147)):*
> ```objc
>    the top left corner of the matrix M̂ and all other blocks appear on the bottom
>    right corner of the matrix. The matrix M̂ can be written in block form as Â =
>    [D, B; Bᵀ, C] where D corresponds to the blocks with indices in E and C
> ```

nit There is no explicit definition of `B`. Is it worth adding?
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 342 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaODtBf3HJU2klXZmzj:-NaODtBf3HJU2klXZmzk:bdg7lbo) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L342)):*
> ```cpp
>   DRAKE_THROW_UNLESS(solver_mode() == SolverMode::kAnalyzed);
>   for (int j = starting_col_block; j < ending_col_block; ++j) {
>     /* Update diagonal. */
> ```

nit Is it worth adding `DRAKE_ASSERT(start_col_block <= 0 && start_col_block < L_.block_cols())`, and same for `ending_col_block`?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 5 unresolved discussions

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 133 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNgQup6IpB6Hwm99ee:-NaOMEyW-255djIE9R2W:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L133)):*
> ```objc
> 
>    In addition, this function computes the Schur complement matrix of the input
>    matrix M in the following sense:
> ```

Done.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 140 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNlbFvDFoWFyiIbx4e:-NaOON6DB7MPNB4MjxTL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L140)):*

I get it now, but it took some parsing. Consider adding an example.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 143 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNkIpA2qSLTtCZq1Jw:-NaOQnCE3HlGb5Xj2Bsv:b-3zbqlk) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L143)):*

I believe this is supposed to be the opposite?
Consider a simple permutation `x̂ = P⋅x`
</blockquote></details>

I think the current spelling is correct. 

We define Pᵢⱼ = 1 iff j = p(i). If `x̂ = P⋅x`, then `x̂ᵢ  = xₚ₍ᵢ₎`, which is the permutation that we want.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 144 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNl0ge4nSWbB1E8GUI:-NaOMzAZ0sE-8z6KvcU4:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L144)):*

here and below we still have a few `M`
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 147 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaOD1AYBh2Uu5326paD:-NaONYWdDCnSJz7t7B0j:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L147)):*

nit There is no explicit definition of `B`. Is it worth adding?
</blockquote></details>

Done
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 78 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNmMLz6r1mIxfAtveW:-NaONelK6dJa5n-YR76Q:bo5njfi) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L78)):*

nit, consider adding a TODO here for a future in which S is a block sparse matrix.
</blockquote></details>

Is S ever going to be sparse? I would think the Schur complement operations destroy sparsity completely.
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 342 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaODtBf3HJU2klXZmzj:-NaOOKaZ9XGQy139dOrY:b-iry7x8) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L342)):*

nit Is it worth adding `DRAKE_ASSERT(start_col_block <= 0 && start_col_block < L_.block_cols())`, and same for `ending_col_block`?
</blockquote></details>

Done, thanks

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 5 unresolved discussions

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 5 unresolved discussions

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 143 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNkIpA2qSLTtCZq1Jw:-NaQMxa61TcbQMdxQEUX:bw3wggn) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.h#L143)):*

I think the current spelling is correct. 

We define Pᵢⱼ = 1 iff j = p(i). If `x̂ = P⋅x`, then `x̂ᵢ  = xₚ₍ᵢ₎`, which is the permutation that we want.
</blockquote></details>

Actually, I got it backwards, we want `x̂ₚ₍ᵢ₎ = xᵢ` (because the hat quantities are the permuted quantities). So you are right.

Sorry about the confusion!

amcastro-tri (commented):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 1 unresolved discussion

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 145 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaRxxpUCANjvc3biJD8:-NaRxxpV5ZtlNsSjDsCp:b-1fyd0h) ([raw file](https://github.com/RobotLocomotion/drake/blob/362bfca423fcff7956f677d969e979b7ff599cd3/multibody/contact_solvers/block_sparse_cholesky_solver.h#L145)):*
> ```objc
> 
>     [p(0), p(1), p(2), p(3), p(4), p(5), p(6)] =
>     [2,    5,    6,    0,    1,    3,    4]
> ```

I believe that per you definition of `p` below this actually is the inverse of p?
I think this should be:
`[p(0), p(1), p(2), p(3), p(4), p(5), p(6)] = [3, 4, 0, 5, 6, 1, 2]`
___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.cc` line 78 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaNmMLz6r1mIxfAtveW:-NaRxUBd0QY-LBG0esGM:bdguik7) ([raw file](https://github.com/RobotLocomotion/drake/blob/99d620959514cb8a27c0d2500c261032d5c0314c/multibody/contact_solvers/block_sparse_cholesky_solver.cc#L78)):*

Is S ever going to be sparse? I would think the Schur complement operations destroy sparsity completely.
</blockquote></details>

nvm, you are right, this is on a per-deformable for our application and yes, that Schur complement will be dense.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807)* status: 1 unresolved discussion

___
*[`multibody/contact_solvers/block_sparse_cholesky_solver.h` line 145 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-NaRxxpUCANjvc3biJD8:-NaSwTYTDUuca19Gvfkn:bayt3v2) ([raw file](https://github.com/RobotLocomotion/drake/blob/362bfca423fcff7956f677d969e979b7ff599cd3/multibody/contact_solvers/block_sparse_cholesky_solver.h#L145)):*

I believe that per you definition of `p` below this actually is the inverse of p?
I think this should be:
`[p(0), p(1), p(2), p(3), p(4), p(5), p(6)] = [3, 4, 0, 5, 6, 1, 2]`

</blockquote></details>

Done. Yes you are right. It's very easy to get confused between p and p^-1!.

amcastro-tri (commented):
Thanks @xuchenhan-tri! this is an incredibly good piece of work. Keep it up!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19807#-:-NagAFRv6ZD7F2TV5diF:b7iewfo)* status: :shipit: complete! all discussions resolved, LGTM from assignees amcastro-tri,EricCousineau-TRI(platform)

PR #19833, opened 2023-07-23
Compute s bounds for rational forward kinematics.

hongkai-dai:
This function will be used in CspaceFreeBox.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833)

### DIFF
--- a/multibody/rational/BUILD.bazel
+++ b/multibody/rational/BUILD.bazel
@@ -74,6 +74,7 @@ drake_cc_googletest(
         ":rational_forward_kinematics",
         ":rational_forward_kinematics_test_utilities",
         "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
         "//common/test_utilities:symbolic_test_util",
     ],
 )
--- a/multibody/rational/rational_forward_kinematics.cc
+++ b/multibody/rational/rational_forward_kinematics.cc
@@ -1,8 +1,12 @@
 #include "drake/multibody/rational/rational_forward_kinematics.h"
 
+#include <limits>
 #include <utility>
 
+#include <fmt/format.h>
+
 #include "drake/common/drake_assert.h"
+#include "drake/common/fmt_eigen.h"
 #include "drake/multibody/rational/rational_forward_kinematics_internal.h"
 #include "drake/multibody/tree/multibody_tree_topology.h"
 #include "drake/multibody/tree/prismatic_mobilizer.h"
@@ -13,6 +17,7 @@
 namespace drake {
 namespace multibody {
 namespace {
+const double kInf = std::numeric_limits<double>::infinity();
 
 RationalForwardKinematics::Pose<symbolic::Polynomial> GetIdentityPose() {
   RationalForwardKinematics::Pose<symbolic::Polynomial> pose;
@@ -75,10 +80,8 @@ RationalForwardKinematics::RationalForwardKinematics(
       const symbolic::Variable s_angle(fmt::format("s[{}]", s_.size()));
       s_.push_back(s_angle);
       s_angles_.push_back(s_angle);
-      cos_delta_.emplace_back(
-          fmt::format("cos_delta[{}]", cos_delta_.size()));
-      sin_delta_.emplace_back(
-          fmt::format("sin_delta[{}]", sin_delta_.size()));
+      cos_delta_.emplace_back(fmt::format("cos_delta[{}]", cos_delta_.size()));
+      sin_delta_.emplace_back(fmt::format("sin_delta[{}]", sin_delta_.size()));
       sin_cos_.emplace_back(sin_delta_.back(), cos_delta_.back());
       sin_cos_set_.insert(sin_delta_.back());
       sin_cos_set_.insert(cos_delta_.back());
@@ -132,6 +135,58 @@ RationalForwardKinematics::CalcBodyPoseAsMultilinearPolynomial(
         q_star, path[i - 1], path[i], poses[i - 1]));
   }
   return std::move(poses.back());
+}
+void RationalForwardKinematics::ComputeSBounds(
+    const Eigen::Ref<const Eigen::VectorXd>& q_star,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+    const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+    Eigen::VectorXd* s_box_lower, Eigen::VectorXd* s_box_upper) const {
+  DRAKE_DEMAND(s_box_lower != nullptr);
+  DRAKE_DEMAND(s_box_upper != nullptr);
+  s_box_lower->resize(s_.size());
+  s_box_upper->resize(s_.size());
+  for (int i = 0; i < ssize(s_); ++i) {
+    const internal::Mobilizer<double>& mobilizer =
+        GetInternalTree(plant_).get_mobilizer(
+            map_s_to_mobilizer_.at(s_[i].get_id()));
+    // the mobilizer cannot be a weld joint since weld joints don't introduce
+    // variables into s_.
+    if (IsRevolute(mobilizer)) {
+      const int q_index = mobilizer.position_start_in_q();
+      if (q_star(q_index) < q_box_lower(q_index)) {
+        throw std::invalid_argument(fmt::format(
+            "ComputeSBounds: q_star({})={} should be >= q_box_lower({})={}.",
+            q_index, q_star(q_index), q_index, q_box_lower(q_index)));
+      }
+      if (q_star(q_index) > q_box_upper(q_index)) {
+        throw std::invalid_argument(fmt::format(
+            "ComputeSBounds: q_star({})={} should be <= q_box_upper({})={}.",
+            q_index, q_star(q_index), q_index, q_box_upper(q_index)));
+      }
+      // s_box_upper = tan(min(theta_box_upper - theta_star, pi)/2)
+      // If theta_box_upper >= theta_star + pi, then the upper bound on s is
+      // inf.
+      (*s_box_upper)(i) =
+          q_box_upper(q_index) - q_star(q_index) >= M_PI
+              ? kInf
+              : tan((q_box_upper(q_index) - q_star(q_index)) / 2);
+      // s_box_lower = tan(max(theta_box_lower - theta_star, -pi)/2)
+      // If theta_box_lower <= theta_star + pi, then the lower bound on s is
+      // -inf.
+      (*s_box_lower)(i) =
+          q_box_lower(q_index) - q_star(q_index) <= -M_PI
+              ? -kInf
+              : tan((q_box_lower(q_index) - q_star(q_index)) / 2);
+    } else if (IsPrismatic(mobilizer)) {
+      const int q_index = mobilizer.position_start_in_q();
+      (*s_box_upper)(i) = q_box_upper(q_index) - q_star(q_index);
+      (*s_box_lower)(i) = q_box_lower(q_index) - q_star(q_index);
+    } else {
+      // Successful construction guarantees nothing but supported mobilizer
+      // types.
+      DRAKE_UNREACHABLE();
+    }
+  }
 }
 
 symbolic::RationalFunction
--- a/multibody/rational/rational_forward_kinematics.h
+++ b/multibody/rational/rational_forward_kinematics.h
@@ -167,6 +167,19 @@ class RationalForwardKinematics {
     return q_val;
   }
 
+  /**
+   Computes the bound of s when q is in the box q_box_lower <= q <= q_box_upper.
+   Notice that for the revolute joint we have s = tan(clamp((θ − θ*)/2, −π/2,
+   π/2). Namely we will clamp (θ − θ*)/2 at ± π/2
+   @note We assume for the revolute joint θ_box_lower <= θ_star <= θ_box_upper.
+   @throw an exception otherwise.
+   */
+  void ComputeSBounds(const Eigen::Ref<const Eigen::VectorXd>& q_star,
+                      const Eigen::Ref<const Eigen::VectorXd>& q_box_lower,
+                      const Eigen::Ref<const Eigen::VectorXd>& q_box_upper,
+                      Eigen::VectorXd* s_box_lower,
+                      Eigen::VectorXd* s_box_upper) const;
+
   const MultibodyPlant<double>& plant() const { return plant_; }
 
   Eigen::Map<const VectorX<symbolic::Variable>> s() const {
--- a/multibody/rational/test/rational_forward_kinematics_test.cc
+++ b/multibody/rational/test/rational_forward_kinematics_test.cc
@@ -1,9 +1,12 @@
 #include "drake/multibody/rational/rational_forward_kinematics.h"
 
+#include <limits>
+
 #include <gtest/gtest.h>
 
 #include "drake/common/symbolic/rational_function.h"
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/common/test_utilities/symbolic_test_util.h"
 #include "drake/math/roll_pitch_yaw.h"
 #include "drake/multibody/rational/test/rational_forward_kinematics_test_utilities.h"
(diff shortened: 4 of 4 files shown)


hongkai-dai (commented):
+@AlexandreAmice for feature review please, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-Na-cfEi3vrc0KRUSgvA:bgvexs6)* status: LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

AlexandreAmice (commented):
More of a design choice question. This function seems extremely specific to `CspaceFreeBox` and so it may be confusing/clutter the API to have it in `RationalForwardKinematics`.

What if instead we added a function to `RationalForwardKinematics` which returned the type of a desired joint as either `kRevolute` or `kPrismatic`? This seems more generally useful. Then in `CspaceFreeBox` you can compute this function.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-NaHDYJF1vA7d6j0x0Rw:b-q9mx3u)* status: 4 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/rational/rational_forward_kinematics.h` line 174 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHCQQ50yN6xJnx2RKm:-NaHCQQ50yN6xJnx2RKn:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.h#L174)):*
> ```objc
>    Notice that for the revolute joint we have s = tan(clamp((θ − θ*)/2, −π/2,
>    π/2). Namely we will clamp (θ − θ*)/2 at ± π/2
>    @note We assuem for the revolute joint θ_box_lower <= θ_star <= θ_box_upper.
> ```

nit: typo



_Suggestion:_
```objc
assume
```
___
*[`multibody/rational/rational_forward_kinematics.cc` line 152 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHCXFq6FD_6omJDzrK:-NaHCXFq6FD_6omJDzrL:b8a9ykt) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.cc#L152)):*
> ```cpp
>         GetInternalTree(plant_).get_mobilizer(
>             map_s_to_mobilizer_.at(s_[i].get_id()));
>     // the mobilizer cannot be a weld joint since weld joint doesn't introduce
> ```

nit: grammar



_Suggestion:_
```cpp
weld joints don't
```
___
*[`multibody/rational/rational_forward_kinematics.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHEVu56ApHXFZu3kT0:-NaHEVu56ApHXFZu3kT1:b1h9tqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.cc#L165)):*
> ```cpp
>             "ComputeSBounds: q_star({})={} should be <= q_box_upper({})={}.",
>             q_index, q_star(q_index), q_index, q_box_upper(q_index)));
>       }
> ```

What is the expected behavior if `q_box_upper` and `q_box_lower` are not in [q_star - pi, q_star + pi]? We should either handle the wrapping ourselves or not allow the user to do this.
___
*[`multibody/rational/test/rational_forward_kinematics_test.cc` line 446 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHDSMPBpsM_RfejyEQ:-NaHDSMQ5-M08TvJzJb4:b8a9ykt) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/test/rational_forward_kinematics_test.cc#L446)):*
> ```cpp
>   EXPECT_TRUE(CompareMatrices(s_box_upper, s_box_upper_expected));
> 
>   // Some q_box_lower is larger than q_star.
> ```

nit: grammar



_Suggestion:_
```cpp
are
```

hongkai-dai (commented):
That's a great question, here is my thought:

1. If I return the type of the joint as `kRevolute` and `kPrismatic`, then the user will need to compute `s` from `q` *outside* of `RationalForwardKinematics`. They must know that for revolute joint `s = tan((theta - theta*)/2)`, and for revolute joint `s = q - q*`.  What if we change the definition of `s` in the future? What if we add more joint types (other than revolute and prismatic) in the future, now the users need to modify not only `RationalForwardKinematics`, but also any code that also compute `s`. This doesn't seem to be very maintainable. It is more maintainable if all these functions that compute `s` with the similar math (like `ComputeSValue`) live in the same place.
2. Although the function `ComputeSBounds` is only used in CspaceFreeBox, I think as long as the math of the function is clear with nice documentation, then the future maintainers don't need to know about CspaceFreeBox to maintain this function, they just need to check if the function implements what the documentation describes.

Hence I would like to keep the function still within RationalForwardKinematics function.

What do you think?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-NaI8ZxO-dYv9sMKn_bD:b-1iyo0)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/rational/rational_forward_kinematics.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHEVu56ApHXFZu3kT0:-NaI7PTMDgs9KEjXnfW2:b-glxbg2) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.cc#L165)):*

What is the expected behavior if `q_box_upper` and `q_box_lower` are not in [q_star - pi, q_star + pi]? We should either handle the wrapping ourselves or not allow the user to do this.
</blockquote></details>

We handle the wrapping already, in the lines below
```cc
      (*s_box_upper)(i) =
          q_box_upper(q_index) - q_star(q_index) >= M_PI
              ? kInf
              : tan((q_box_upper(q_index) - q_star(q_index)) / 2);
      (*s_box_lower)(i) =
          q_box_lower(q_index) - q_star(q_index) <= -M_PI
              ? -kInf
              : tan((q_box_lower(q_index) - q_star(q_index)) / 2);
```
I added some documentation to make it more clear.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833)* status: 1 unresolved discussion, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/rational/rational_forward_kinematics.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHEVu56ApHXFZu3kT0:-NaIBBCy1_5BAdUd6CYa:bww08rm) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.cc#L165)):*

We handle the wrapping already, in the lines below
```cc
      (*s_box_upper)(i) =
          q_box_upper(q_index) - q_star(q_index) >= M_PI
              ? kInf
              : tan((q_box_upper(q_index) - q_star(q_index)) / 2);
      (*s_box_lower)(i) =
          q_box_lower(q_index) - q_star(q_index) <= -M_PI
              ? -kInf
              : tan((q_box_lower(q_index) - q_star(q_index)) / 2);
```
I added some documentation to make it more clear.
</blockquote></details>

BTW, we do want to handle the case that q_box_upper and q_box_lower are NOT in [q_star - pi, q_star + pi]. When we have q_star != 0, even if we have the assumption that -pi <= q_box_lower <= q_box_upper <= pi (as we did in the C-IRIS paper), it is still possible to have q_box_upper - q_star > pi. (For example, when q_box_upper = pi and q_star = -0.5pi).

AlexandreAmice (commented):
That is good reasoning. I think perhaps the problem I have with the name of this method is that to me `QBounds` sounds like the joint limits of our robots and so I would expect `ComputeSBounds` to return the `s` joint limits. 

I suggested a name change and am happy to mark myself as satisfied if you still disagree.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-NaRqH0W3OSgluZt4ahX:brhu6qz)* status: 3 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @hongkai-dai)

___
*[`multibody/rational/rational_forward_kinematics.h` line 175 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaRqU8w7L4qBhCcCOdI:-NaRqU8w7L4qBhCcCOdJ:bhrm9q2) ([raw file](https://github.com/RobotLocomotion/drake/blob/f35a541abe8baee483da801e979fb49d2b18917e/multibody/rational/rational_forward_kinematics.h#L175)):*
> ```objc
>    π/2). Namely we will clamp (θ − θ*)/2 at ± π/2
>    @note We assume for the revolute joint θ_box_lower <= θ_star <= θ_box_upper.
>    @throw an exception otherwise.
> ```

nit: clarity



_Suggestion:_
```objc
   Computes the bounds s_lower <= s <= s_upper when q is in the box q_box_lower <= q <= q_box_upper.
   Notice that for the revolute joint we have s = tan(clamp((θ − θ*)/2, −π/2,
   π/2). Namely we will clamp (θ − θ*)/2 at ± π/2
   @note We assume for the revolute joint that θ_box_lower <= θ_star <= θ_box_upper.
   @throw an exception otherwise.
```
___
*[`multibody/rational/rational_forward_kinematics.h` line 177 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaTGzboDhTzdBibQcTF:-NaTGzboDhTzdBibQcTG:bc5wfuo) ([raw file](https://github.com/RobotLocomotion/drake/blob/f35a541abe8baee483da801e979fb49d2b18917e/multibody/rational/rational_forward_kinematics.h#L177)):*
> ```objc
>    @throw an exception otherwise.
>    */
>   void ComputeSBounds(const Eigen::Ref<const Eigen::VectorXd>& q_star,
> ```

I think the problem I have with the name perhaps is that to me `QBounds` sounds like the joint limits of our robots and this is not computing that.

Can we change the name `ComputeSBox`
___
*[`multibody/rational/rational_forward_kinematics.cc` line 165 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-NaHEVu56ApHXFZu3kT0:-NaTFWwZ4JlMM9yclQvg:b1xdetk) ([raw file](https://github.com/RobotLocomotion/drake/blob/84f7bcb94d04dad335263cd66cbfe72cdadb071c/multibody/rational/rational_forward_kinematics.cc#L165)):*

BTW, we do want to handle the case that q_box_upper and q_box_lower are NOT in [q_star - pi, q_star + pi]. When we have q_star != 0, even if we have the assumption that -pi <= q_box_lower <= q_box_upper <= pi (as we did in the C-IRIS paper), it is still possible to have q_box_upper - q_star > pi. (For example, when q_box_upper = pi and q_star = -0.5pi).
</blockquote></details>

So my interpretation is that `q_star` is the center of the "wrap around". Consider the following example:
`q_lower = -5pi/2 = pi/2 mod [-pi,pi]`
`q_upper = 0 = 0 mod [-pi, pi]`
Here we have that `q_lower <= q_upper` linearly, but `q_lower >= q_upper` mod the wrapping WRT `q_star = 0`.

I see that throughout `RationalForwardKinematics`, we handle this is by clamping `q_lower` to `-pi/2`, so I suppose this is consistent.

However, in this case, I am not sure why we throw if `q_star <= q_lower` and conversely if `q_upper <= q_upper`

hongkai-dai (commented):
I see, I agree the name of the function isn't right.

I think the documentation of the function wasn't right, hence there was a misunderstanding on what the function tries to do.

We defined the TC variable `s` as this:
1. For revolute joint, we have `t = tan((θ−θ*)/2)`. I don't see any wrapping up at pi/2 in our current code.
2. For prismatic joint, we have `z = d - d*`.

Now we have a box region in the C-space `q_box_lower <= q <= q_box_upper`. What is the set of TC variable `s`?

For the prismatic joint, it is easy,  the range of`z` is also a box.

For the revolute joint, it is a lot harder. We are asking "what is the range of tan(x), when x is in the range [(θ_box_lower - θ*)/2, (θ_box_upper - θ*)/2]"? Without loss of generality, we can assume that -1.5π < (θ_box_lower - θ*)/2 <= (θ_box_upper - θ*)/2 < 1.5π. And there are four different cases:
1. -π/2 < (θ_box_lower - θ*)/2 <= (θ_box_upper - θ*)/2 < π/2, then tan(x) is monotonically increasing in this range, hence we have the range of t = tan(x) as the box [tan(θ_box_lower - θ*)/2, tan(θ_box_upper - θ*)/2].
2. -π/2 <  (θ_box_lower - θ*)/2 < π/2 <= (θ_box_upper - θ*)/2. Then tan(x) first goes to infinity as x approaches π/2, and then jumps to -inf and increases to tan(θ_box_upper - θ*)/2. The range of t = tan(x) is the union of the two boxes [-inf, tan(θ_box_upper - θ*)/2] and [tan(θ_box_lower - θ*)/2, inf]. Sometimes these two boxes overlap and becomes one box [-inf, inf].
3. (θ_box_lower - θ*)/2 <= -π/2 <   (θ_box_upper - θ*)/2 <= π/2. This is similar to case 2, where we might have two non-overlapping boxes, or one box [-inf, inf].
4. (θ_box_lower - θ*)/2 <= -π/2 < π/2 <= (θ_box_upper - θ*)/2. Then the range of t = tan(x) is the box [-inf, inf].

As you can see, we might end up with two boxes for the range of `t` (case 2 and 3), rather than a single box. The current API doesn't capture the two-box case.

So why I want to create this function? In the cspace box code, I care about the case when q_star is inside the box [q_box_lower, q_box_upper], hence `s` can always take the value 0. And when there are two non-overlapping boxes for the range of `s`, I care about the box that contains `s=0`. So I think this is what I want to do
```cc
/**
  Computes the range of the TC variable s when the C-space variable is in the box q_box_lower <= q <= q_box_upper. 
  We assume q_box_lower <= q_star <= q_box_upper. Depending on the range of q_box_lower and q_box_upper, sometimes the range of s are the unions of boxes. We return the box that contains s = 0 inside.
*/
void ComputeSBoxContainingOrigin(const Eigen::Ref<const Eigen::VectorXd>& q_box_lower, const Eigen::Ref<const Eigen::VectorXd>& q_box_upper, const Eigen::Ref<const Eigen::VectorXd>& q_star, Eigen::VectorXd* s_box_lower, Eigen::VectorXd* s_box_upper) const;
```
with the function name changed to `ComputeSBoxContainingOrigin`, and explicitly require that q_box_lower <=q <= q_box_upper.

What do you think?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-NagdmOx4ZLEybKMDk3X:bpjge7m)* status: 3 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

hongkai-dai (commented):
I will close this PR.

I think this function is too specific to our cspace free box code. As you said, it is better to put this code inside CspaceFreeBox. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19833#-:-Nai5SPa69Yc0di2NbKw:blkgdol)* status: 3 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes

PR #19837, opened 2023-07-24
Meshcat can specify the environment map

SeanCurtis-TRI:
Environment maps allow for a higher fidelity of the visualized geometry, particularly useful for objects defined with physically-based rendering (PBR) material properties.

This gives the Meshcat class the ability to set and clear such an environment variable (with the attendant bindings). Furthermore, model_visualizer has been extended to take an optional environment map as a parameter (implicitly testing the bindings).

The actual implementation is tested in meshcat_manual_test.cc where an environment gets loaded and then subsequently dismissed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)


SeanCurtis-TRI (commented):
+@RussTedrake for feature review, please.
+(release notes: feature) 
+(status: do not merge) This needs to merge after https://github.com/rdeits/meshcat/pull/148 merges.

The best way to see the effect is in `meshcat_manual_test`. We don't currently have many models available that directly benefit from an environment map. We'll update `drake_models` down the road to improve the fidelity of our baseline models and the impact of an environment map will be much greater.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-Na5daXF5g49e4NkdmH0:b-wcimmc)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

RussTedrake (commented):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: 8 unresolved discussions, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`geometry/BUILD.bazel` line 416 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJwbe62NRzErIIVGiq:-NaJwbe62NRzErIIVGir:b-gcnn2w) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/BUILD.bazel#L416)):*
> ```Starlark
>     name = "meshcat_resources",
>     srcs = [
>         "env_256_cornell_box.png",
> ```

this is a test resource, not a mainline meshcat resource.  i would think it belongs in a different filegroup?  and perhaps even a different directory?
___
*[`geometry/meshcat.h` line 567 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJyTTa0mAKbACYI_Xl:-NaJyTTa0mAKbACYI_Xm:b-98lkkg) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.h#L567)):*
> ```objc
>    rendering (PBR) material properties (e.g., metallic surfaces), this defines
>    the luminance environment, contributing to total illumination and appearing
>    in reflections.
> ```

in the docstrings below, you say more "can be jpg, png, etc..".  That seems worth mentioning here, too.
___
*[`geometry/meshcat.h` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxXYy0K84V5CQ6MAK:-NaJxXYy0K84V5CQ6MAL:bwo9jur) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.h#L570)):*
> ```objc
> 
>    If the path is empty, or the image can't be read, the environment map will
>    be cleared. */
> ```

btw -- perhaps that justifies using the empty string as a default argument?
___
*[`geometry/meshcat.cc` line 2335 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJx8Qg1kN0xDSQkhe7:-NaJx8Qg1kN0xDSQkhe8:b-f6spcv) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.cc#L2335)):*
> ```cpp
>                       std::istreambuf_iterator<char>());
>       image_encoding =
>           fmt::format("data:image/{};base64,", ext) +
> ```

does this work for anything other than png?  should we cover at least one more case in the meshcat_manual_test?  seems like jpg is the one you've promised in the docstrings below.
___
*[`geometry/meshcat.cc` line 2340 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJz9QLCWhCz5KCGrlB:-NaJz9QM0sUg7kjd_t3i:b-k4v41d) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.cc#L2340)):*
> ```cpp
>   }
>   impl().SetProperty("/Background/<object>", "environment_map",
>                       image_encoding);
> ```

can we confirm in the manual test that this working in the StaticHtml?
can you set the property at a certain time during an animation?  (I presume yes, but it would be nice to see). 
___
*[`geometry/test/meshcat_manual_test.cc` line 281 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxKLuDCFst_lKcAB1:-NaJxKLuDCFst_lKcAB2:bxtjk2t) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/test/meshcat_manual_test.cc#L281)):*
> ```cpp
> 
>   std::cout << "- An environment map has been loaded, the Cornell box\n"
>             << "  the dented green box should reflect it.\n";
> ```

I initially thought this was a grammar mistake: "the Cornell box the dented green box".  now I understand how you intended me to parse it.

perhaps:
An environment map (the "Cornell box") has been loaded. The dented green box should reflect it.
___
*[`bindings/pydrake/geometry/test/visualizers_test.py` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxdTECOgCz-SEQde8:-NaJxdTECOgCz-SEQde9:bzclj04) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/geometry/test/visualizers_test.py#L102)):*
> ```Python
>         self.assertIn("http", meshcat.web_url())
>         self.assertIn("ws", meshcat.ws_url())
>         meshcat.SetEnvironmentMap("")
> ```

please use the `image_path` argument name in the test.
___
*[`bindings/pydrake/visualization/_meldis.py` line 534 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJy2eG-PK4d4WjSJIw:-NaJy2eG-PK4d4WjSJIx:bgr0h5j) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L534)):*
> ```Python
>             elif environment_map.startswith("drake/"):
>                 self.meshcat.SetEnvironmentMap(
>                     FindResourceOrThrow(environment_map))
> ```

nit: it seems a little inconsistent to support this here, but not in the main Meshcat class?

SeanCurtis-TRI (commented):
The requisite meshcat PR (148) has merged, however, it introduced a bug that is being addressed in https://github.com/rdeits/meshcat/pull/149.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NaK85vOAKSdGYxWWPR0:b-7tu087)* status: 7 unresolved discussions, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`geometry/BUILD.bazel` line 416 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJwbe62NRzErIIVGiq:-NaKBq-03kxU_bJkNQnO:brrswbv) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/BUILD.bazel#L416)):*

this is a test resource, not a mainline meshcat resource.  i would think it belongs in a different filegroup?  and perhaps even a different directory?
</blockquote></details>

A fine point. I dumped it here because it guaranteed that at least one environment texture would be available any time meshcat is. For now, I've moved the image out to the test resources and if we choose to release an environment texture, we can figure out where it should go.
___
*[`geometry/meshcat.h` line 567 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJyTTa0mAKbACYI_Xl:-NaKPH86EBKmYY62CDVp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.h#L567)):*

in the docstrings below, you say more "can be jpg, png, etc..".  That seems worth mentioning here, too.
</blockquote></details>

Done.
___
*[`geometry/meshcat.h` line 570 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxXYy0K84V5CQ6MAK:-NaKOJ2VAJdpCnb7fm4a:b-2nr6mr) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.h#L570)):*

btw -- perhaps that justifies using the empty string as a default argument?
</blockquote></details>

I think there's probably a philosophical debate about default arguments. I'm going to defer default arguments for now. I can imagine possible additions to the parameter list in the future and I don't want default arguments interfering. Invoking this and passing '""' doesn't seem particularly arduous.
___
*[`geometry/meshcat.cc` line 2335 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJx8Qg1kN0xDSQkhe7:-NaKR0PDBl-K02DHNXK9:blefwer) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.cc#L2335)):*

does this work for anything other than png?  should we cover at least one more case in the meshcat_manual_test?  seems like jpg is the one you've promised in the docstrings below.
</blockquote></details>

I've updated meshcat_manual_test to load two different environments with two different file formats.
___
*[`geometry/meshcat.cc` line 2340 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJz9QLCWhCz5KCGrlB:-NaKQMUE53m94DjL5n9m:be16pv5) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.cc#L2340)):*

can we confirm in the manual test that this working in the StaticHtml?
can you set the property at a certain time during an animation?  (I presume yes, but it would be nice to see). 
</blockquote></details>

I've added the environment map to the static html (and it does appear).

As for animation, I didn't introduce the `time_in_recording` parameter for `SetEnvironmentMap` because it's not generally something that gets animated. The nature of the environment map is a static environment. Sequentially swapping the image over time seems counter-intuitive (and possibly computationally ugly, depending on how the browser handles the images). Also, a *good* environment map is usually an order of magnitude *larger* than the toy cornell box I'm using here. So, the cost will be much higher.
___
*[`geometry/test/meshcat_manual_test.cc` line 281 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxKLuDCFst_lKcAB1:-NaKdtCqBq5vPMkf264P:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/test/meshcat_manual_test.cc#L281)):*

I initially thought this was a grammar mistake: "the Cornell box the dented green box".  now I understand how you intended me to parse it.

perhaps:
An environment map (the "Cornell box") has been loaded. The dented green box should reflect it.
</blockquote></details>

Done.
___
*[`bindings/pydrake/geometry/test/visualizers_test.py` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJxdTECOgCz-SEQde8:-NaKdxgh1CdP2hteesF7:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/geometry/test/visualizers_test.py#L102)):*

please use the `image_path` argument name in the test.
</blockquote></details>

Done.
___
*[`bindings/pydrake/visualization/_meldis.py` line 534 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJy2eG-PK4d4WjSJIw:-NaKdyeX6_196nzDnTDg:b-x5so6t) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L534)):*

nit: it seems a little inconsistent to support this here, but not in the main Meshcat class?
</blockquote></details>

There is a palpable difference to me.

Meldis is an app. The environment map is set from the command line and the user has no recourse to `FindResourceOrThrow`.

`Meshcat` is a class in code. If they are exercising that API, they have full access to any mechanism Drake provides for resolving paths, so it is on them to do so. They'd be in exactly the same boat if they tried `SetObject("path", Mesh(...));`. The `Mesh` shape that is passed in must already be a path that `Meshcat` can read.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: 8 unresolved discussions, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_meldis.py` line 527 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaMqZr107KnOypc_TLW:-NaMqZr28YAIplM9ddkJ:bj6fmpu) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L527)):*
> ```Python
>                                show_stats_plot=False)
>         self.meshcat = Meshcat(params=params)
>         if (environment_map):
> ```

nit Pythonic



_Suggestion:_
```Python
if environment_map is not None:
```
___
*[`bindings/pydrake/visualization/_meldis.py` line 534 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJy2eG-PK4d4WjSJIw:-NaMqsxo3g7nz-nkk04N:ba95p00) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L534)):*

There is a palpable difference to me.

Meldis is an app. The environment map is set from the command line and the user has no recourse to `FindResourceOrThrow`.

`Meshcat` is a class in code. If they are exercising that API, they have full access to any mechanism Drake provides for resolving paths, so it is on them to do so. They'd be in exactly the same boat if they tried `SetObject("path", Mesh(...));`. The `Mesh` shape that is passed in must already be a path that `Meshcat` can read.
</blockquote></details>

The magic rewriting here is unsound.  There's no reason to think that a Drake file is always part of the declared runfiles prereqs (and in fact, mostly they would not be.)  The constructor should take a `pathlib.Path` and use it without any extra munging.

RussTedrake (commented):
will run it again locally once the meshcat sha is updated with the latest fix.

> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NaMycAN4BxdT9BpAEdR:b-53l2xi)* status: 1 unresolved discussion, LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`geometry/meshcat.cc` line 2340 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJz9QLCWhCz5KCGrlB:-NaMuwK11M5lLJx6srZY:bb73djm) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/geometry/meshcat.cc#L2340)):*

I've added the environment map to the static html (and it does appear).

As for animation, I didn't introduce the `time_in_recording` parameter for `SetEnvironmentMap` because it's not generally something that gets animated. The nature of the environment map is a static environment. Sequentially swapping the image over time seems counter-intuitive (and possibly computationally ugly, depending on how the browser handles the images). Also, a *good* environment map is usually an order of magnitude *larger* than the toy cornell box I'm using here. So, the cost will be much higher.
</blockquote></details>

SGTM

SeanCurtis-TRI (commented):
Alright, meshcat updated and comments addressed.

-(status: do not merge) 

+@ggould-tri for Monday platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NaOofl_7lJ3qV6eNsR-:bw44os5)* status: LGTM missing from assignees RussTedrake(platform),ggould-tri(platform), labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_meldis.py` line 534 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJy2eG-PK4d4WjSJIw:-NaOnC2N9MXa4ahAaLIm:b-twh1x5) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L534)):*

The magic rewriting here is unsound.  There's no reason to think that a Drake file is always part of the declared runfiles prereqs (and in fact, mostly they would not be.)  The constructor should take a `pathlib.Path` and use it without any extra munging.
</blockquote></details>

I wasn't assuming that it *would* be (I've encountered plenty that weren't). However, in the case where it was, I thought it'd be nice. That said, I'm content to make it a simple `Path`.

I've applied the same change to `_model_visualizer.py`.

RussTedrake (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NaP7IDn8_XiaE2UtzdR:bnfp4nl)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> Reviewed 1 of 8 files at r2, 1 of 4 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: 2 unresolved discussions, LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/_meldis.py` line 534 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaJy2eG-PK4d4WjSJIw:-NaPRmOj7QJyG5bS_buu:bf6f1hw) ([raw file](https://github.com/RobotLocomotion/drake/blob/52445e1b7cc6c34e1565ef5d0260d3c6a1769e80/bindings/pydrake/visualization/_meldis.py#L534)):*

I wasn't assuming that it *would* be (I've encountered plenty that weren't). However, in the case where it was, I thought it'd be nice. That said, I'm content to make it a simple `Path`.

I've applied the same change to `_model_visualizer.py`.
</blockquote></details>

FYI In the past we have tried to move the `FindResourceOrThrow` to be "under the hood" and happen within an implementation, instead of only ever at the periphery.  We always end up regretting it later.  A `Path` and/or URI should be the go-to design choice.  (In the extreme, we can have `runfiles://foo/bar.baz` as a URI.)
___
*[`bindings/pydrake/visualization/_meldis.py` line 19 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaPR3A67ujuE_gOwEyx:-NaPR3A67ujuE_gOwEyy:bfenpfz) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b4904c8498fcfa2234d2397433a997d44bfabf1/bindings/pydrake/visualization/_meldis.py#L19)):*
> ```Python
> from pydrake.common import (
>     configure_logging,
>     FindResourceOrThrow,
> ```

nit Unused `FindResourceOrThrow` import.
___
*[`bindings/pydrake/visualization/model_visualizer.py` line 148 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaPRJWsE-D6xj4Nx48B:-NaPRJWt99kJ18Q00CNd:byzll4t) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b4904c8498fcfa2234d2397433a997d44bfabf1/bindings/pydrake/visualization/model_visualizer.py#L148)):*
> ```Python
>                                   browser_new=args.browser_new,
>                                   pyplot=args.pyplot,
>                                   environment_map=Path(args.environment_map))
> ```

BTW IIRC adding `type=Path` to the `add_argument` declaration would remove the need to specify a type here, since then the parsed value would be properly typed.



_Suggestion:_
```Python
args.environment_map
```

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

___
*[`bindings/pydrake/visualization/model_visualizer.py` line 148 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NaPRJWsE-D6xj4Nx48B:-NaPwv_E6jvVoXlIu0HS:buirqu8) ([raw file](https://github.com/RobotLocomotion/drake/blob/6b4904c8498fcfa2234d2397433a997d44bfabf1/bindings/pydrake/visualization/model_visualizer.py#L148)):*

BTW IIRC adding `type=Path` to the `add_argument` declaration would remove the need to specify a type here, since then the parsed value would be properly typed.
</blockquote></details>

Oops. I forgot that. Thanks.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: LGTM missing from assignee ggould-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
+(status: commits are properly curated) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NacyDTNDWBW1y5a4zW2:b-qv2c28)* status: LGTM missing from assignee ggould-tri(platform), when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
Thanks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-:-NaiMUeS48Z2avlCfn57:b-ucfkn4)* status: 1 unresolved discussion, when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-NaiKHH_EVz54-u7JyRP:b-3ng03o) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*

minor:  Wouldn't it be better to at least warn?  A silent failure seems like a recipe for hard-to-find issues, particularly with bazel's multiple roots and rlocs and whatnot, and is unlikely to reflect caller intent.
Also this is very non-pythonic -- it invites TOCTOU errors, and indeed has one in its python test.
</blockquote></details>

Switched to warning and elaborated the documentation.
___
*[`geometry/meshcat.cc` line 2327 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagqDXPEf7dI7EqcyaT:-NaiKUnZ-lg55RqoJQoB:bgm8qs) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.cc#L2327)):*

nit: It's a bit odd that the only part of this glyphic incantation that gets an implementation comment is the bit about the file extension, and not the filesize idiom or base64 formatting.
</blockquote></details>

Because that's my novel contribution, the rest of the code is copypasta from above.

Hopefully, the new comment is more helpful.
___
*[`bindings/pydrake/visualization/_model_visualizer.py` line 303 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagsQyIEDbZkbBxEVhC:-NaiLER9Ek7AG-VGg3yf:b-120bxf) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/bindings/pydrake/visualization/_model_visualizer.py#L303)):*

BTW  This is literally the canonical example of a TOCTOU error in python, which is what the issue above is referencing.
</blockquote></details>

I believe it's implicitly resolved.

FTR  I'd argue that it is *not* a "canonical" example of TOCTOU -- it merely has the form of one. To be canonical, there has to be a viable race condition. I believe it would require exceptional circumstances for the existence of this file to be transient. (To say nothing of the incredibly low stakes at play.)

Nevertheless, I agree that there should be a clear distinction between explicitly turning off the environment map with an empty string and accidentally turning it off because the file you said couldn't be accessed.

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: 1 unresolved discussion, when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: 1 unresolved discussion, when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-Nalp5TnAvr9Uqed3yzL:bvvx3hp) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*

Switched to warning and elaborated the documentation.
</blockquote></details>

Why warn instead of throw?  It's atypical for an API function to appear to succeed when it hasn't accomplished its goal.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),RussTedrake(platform), base commit is latest master

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-NanQ07wEvfNvYGz0bey:b-opnu5p) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*

Why warn instead of throw?  It's atypical for an API function to appear to succeed when it hasn't accomplished its goal.
</blockquote></details>

Done

Although, one would be hard-pressed to describe a meshcat viewer that has no environment variable to be "successful".

And in the case where you happen to provide an image of the wrong type, the outcome is the same. Drake doesn't complain, the complaint will be in the browser console. But, meh.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: :shipit: complete! all discussions resolved, LGTM from assignees ggould-tri(platform),RussTedrake(platform), base commit is latest master

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-NanU8V4-QMlIm2Jrgs9:b-7b48kb) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*

Done

Although, one would be hard-pressed to describe a meshcat viewer that has no environment variable to be "successful".

And in the case where you happen to provide an image of the wrong type, the outcome is the same. Drake doesn't complain, the complaint will be in the browser console. But, meh.
</blockquote></details>

The Doxygen here still refers to "warns" instead of "throws"

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-NangeNNBpJb5MtfmDyn:b4b64ac) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*
Huh? I distinctly see:

> @throws if `image_path` is *not* empty and the file isn't accessible.

What are you seeing that I'm not?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19837)* status: when planning a "properly curated" merge commit the PR must always be rebased onto latest master (waiting on @SeanCurtis-TRI)

___
*[`geometry/meshcat.h` line 576 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19837#-NagrLoa6w4m9u0hhXhn:-Nani2BnBqWqAZPmZIor:b-98jcpo) ([raw file](https://github.com/RobotLocomotion/drake/blob/41fd4c132cbe3fa395234803a9eacb4d954aa955/geometry/meshcat.h#L576)):*

Huh? I distinctly see:

> @throws if `image_path` is *not* empty and the file isn't accessible.

What are you seeing that I'm not?
</blockquote></details>

Hmm.  Somehow I failed to set the File Matrix to the latest revision here.  Looks fine now.

PR #19843, opened 2023-07-25
Revert "CspaceFreeBox find separation certificates for a given box."

svenevs:
Reverts RobotLocomotion/drake#19742

Dear @hongkai-dai,

 The on-call build cop, @svenevs, believes that your PR #19742 may have
 broken one or more of Drake's continuous integration builds [1]. It is
 possible to break a build even if your PR passed continuous integration
 pre-merge because additional platforms are tested post-merge.

 The specific build failures under investigation are:

- (`FAIL: //geometry/optimization:cspace_free_polytope_with_mosek_test`) https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/linux-focal-clang-bazel-nightly-everything-leak-sanitizer/89/consoleFull
- (`FAIL: //geometry/optimization:cspace_free_polytope_with_mosek_test`) https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/linux-focal-clang-bazel-nightly-everything-thread-sanitizer/538/consoleFull
- (`FAIL: //geometry/optimization:cspace_free_polytope_with_mosek_test`) https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/linux-focal-clang-bazel-nightly-everything-undefined-behavior-sanitizer/534/consoleFull
- (`FAIL: //geometry/optimization:cspace_free_polytope_with_mosek_test`) https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/mac-arm-monterey-clang-bazel-nightly-everything-address-sanitizer/117/consoleFull
- (`FAIL: //geometry/optimization:cspace_free_polytope_with_mosek_test`) https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/mac-arm-ventura-clang-bazel-nightly-everything-release/66/consoleFull

 Therefore, the build cop has created this revert PR and started a complete
 post-merge build to determine whether your PR was in fact the cause of the
 problem. If that build passes, this revert PR will be merged 60 minutes from
 now. You can then fix the problem at your leisure, and send a new PR to
 reinstate your change.

 If you believe your original PR did not actually break the build, please
 explain on this thread.

 If you believe you can fix the break promptly in lieu of a revert, please
 explain on this thread, and send a PR to the build cop for review ASAP.

 If you believe your original PR definitely did break the build and should be
 reverted, please review and LGTM this PR. This allows the build cop to merge
 without waiting for CI results.

 For advice on how to handle a build cop revert, see [2].

 Thanks!
[...]

PR #19845, opened 2023-07-25
Fix an asan bug in CspaceFreePolytope.

hongkai-dai:
When finding the separating planes in parallel, the lambda function that actually solves the sos should return plane_count.

Resolves #19843

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845)

### DIFF
--- a/geometry/optimization/cspace_free_box.cc
+++ b/geometry/optimization/cspace_free_box.cc
@@ -304,7 +304,8 @@ CspaceFreeBox::FindSeparationCertificateGivenBox(
   // This lambda function formulates and solves a small SOS program for each
   // pair of geometries.
   auto solve_small_sos = [this, &polynomials_to_certify, &active_plane_indices,
-                          &options, &ret](int plane_count) -> bool {
+                          &options,
+                          &ret](int plane_count) -> std::pair<bool, int> {
     const int plane_index = active_plane_indices[plane_count];
     auto certificate_program = this->ConstructPlaneSearchProgram(
         polynomials_to_certify.plane_geometries[plane_count],
@@ -320,10 +321,10 @@ CspaceFreeBox::FindSeparationCertificateGivenBox(
           plane_index, separating_planes()[plane_index].a,
           separating_planes()[plane_index].b,
           separating_planes()[plane_index].decision_variables, result));
-      return true;
+      return std::make_pair(true, plane_count);
     } else {
       ret[plane_count].reset();
-      return false;
+      return std::make_pair(false, plane_count);
     }
   };
   this->SolveCertificationForEachPlaneInParallel(
--- a/geometry/optimization/cspace_free_polytope.cc
+++ b/geometry/optimization/cspace_free_polytope.cc
@@ -294,10 +294,10 @@ CspaceFreePolytope::FindSeparationCertificateGivenPolytope(
 
   // This lambda function formulates and solves a small SOS program for each
   // pair of geometries.
-  auto solve_small_sos = [this, &d_minus_Cs, &C_redundant_indices,
-                          &s_lower_redundant_indices,
-                          &s_upper_redundant_indices, &active_plane_indices,
-                          &options, &ret](int plane_count) -> bool {
+  auto solve_small_sos =
+      [this, &d_minus_Cs, &C_redundant_indices, &s_lower_redundant_indices,
+       &s_upper_redundant_indices, &active_plane_indices, &options,
+       &ret](int plane_count) -> std::pair<bool, int> {
     const int plane_index = active_plane_indices[plane_count];
     auto certificate_program = this->ConstructPlaneSearchProgram(
         this->plane_geometries_[plane_index], d_minus_Cs, C_redundant_indices,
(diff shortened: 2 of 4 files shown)


xuchenhan-tri (commented):
+@xuchenhan-tri for platform review.
+(release notes: none) 
Platform :lgtm: pending CI. 
+(priority: high) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-:-NaD8fVk5e8phljIyTR0:b-hia5eo)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 295 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-NaD6pDi7pM4a42Pdntd:-NaD6pDi7pM4a42Pdnte:b-l6hw3m) ([raw file](https://github.com/RobotLocomotion/drake/blob/01a2823ffbf9dced3be2a985cc39feb10a79b24c/geometry/optimization/cspace_free_polytope_base.cc#L295)):*
> ```cpp
>         // This call to future.get() is necessary to propagate any exception
>         // thrown during SOS setup/solve.
>         const auto [sos_success, plane_count] = operation->get();
> ```

BTW, from the usage here, it seems like the semantic is closer to a index/identifier for the planes instead of the number/count of planes.



_Suggestion:_
```cpp
plane_index
```

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 295 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-NaD6pDi7pM4a42Pdntd:-NaDAMit3TR1eKL4NRUD:b-m6p3pd) ([raw file](https://github.com/RobotLocomotion/drake/blob/01a2823ffbf9dced3be2a985cc39feb10a79b24c/geometry/optimization/cspace_free_polytope_base.cc#L295)):*

BTW, from the usage here, it seems like the semantic is closer to a index/identifier for the planes instead of the number/count of planes.
</blockquote></details>

We actually have `plane_index` variable, and it means a different thing.

For every pair of geometries in the SceneGraph, we assign a plane to separate that pair of geometries, and we given an index of the plane, we use `plane_index` for this index.

Inside this function, we don't need to find the separating planes for all pairs of geometries. Instead we only find the planes whose index is in `active_plane_indices`. And we use `plane_count` as the counting variable to loop through `active_plane_indices`.  So we want to find the plane with the index `active_plane_indices[plane_count]`, and then we record whether the plane is found or not in `is_success[plane_count]`.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 295 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-NaD6pDi7pM4a42Pdntd:-NaDBkVe6D9Uig18lFz4:b-uqzml2) ([raw file](https://github.com/RobotLocomotion/drake/blob/01a2823ffbf9dced3be2a985cc39feb10a79b24c/geometry/optimization/cspace_free_polytope_base.cc#L295)):*

We actually have `plane_index` variable, and it means a different thing.

For every pair of geometries in the SceneGraph, we assign a plane to separate that pair of geometries, and we given an index of the plane, we use `plane_index` for this index.

Inside this function, we don't need to find the separating planes for all pairs of geometries. Instead we only find the planes whose index is in `active_plane_indices`. And we use `plane_count` as the counting variable to loop through `active_plane_indices`.  So we want to find the plane with the index `active_plane_indices[plane_count]`, and then we record whether the plane is found or not in `is_success[plane_count]`. 
</blockquote></details>

I added some documentation, is it more clear now?

xuchenhan-tri (commented):
> Reviewed 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/optimization/cspace_free_polytope_base.cc` line 295 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-NaD6pDi7pM4a42Pdntd:-NaDUJUkB5iT1ZUJjAcj:br06oxi) ([raw file](https://github.com/RobotLocomotion/drake/blob/01a2823ffbf9dced3be2a985cc39feb10a79b24c/geometry/optimization/cspace_free_polytope_base.cc#L295)):*

I added some documentation, is it more clear now?
</blockquote></details>

Looks good, thanks!

xuchenhan-tri (commented):
+(status: single reviewer ok) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19845#-:-NaDUaes96j0NugYrReJ:b3dcro0)* status: :shipit: complete! all discussions resolved, LGTM from assignee xuchenhan-tri(platform)

PR #19828, opened 2023-07-21
Creates AffineSubspace subclass of ConvexSet.

cohnt:
Towards #19717. This provides an implementation of [AffineSubspace](https://en.wikipedia.org/wiki/Affine_space#Affine_subspaces_and_parallelism) as a subclass of ConvexSet.

Note that `DoAddPointInNonnegativeScalingConstraints` is not yet implemented, but within Drake, this method is only called in `GraphOfConvexSets`, and you shouldn't have unbounded sets there. (AffineSubspace is almost always unbounded.) I'm happy to come back later and implement that feature if desired.

PR #19851, opened 2023-07-26
Implement Change of Coordinates Functionality for AffineSubspace.

cohnt:
This PR implements new functionality for the AffineSubspace class:

- `AffineDimension()`, which returns the affine dimension of this AffineSubspace. For this class of sets, it's simply the number of columns in the basis.
- `Project(x)`, which computes the orthogonal projection of x onto the AffineSubspace
- `ToLocalCoordinates(x)`, which yields local coordinates of the projection onto AffineSubspace, with respect to its own basis (and zero point)
- `ToGlobalCoordinates(x)`, which lifts a point from local coordinate along AffineSubspace into the ambient space

Once #19828 lands, I'll rebase and look for a reviewer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -95,6 +95,14 @@ void DefineGeometryOptimization(py::module m) {
             cls_doc.basis.doc)
         .def("translation", &AffineSubspace::translation,
             py_rvp::reference_internal, cls_doc.translation.doc)
+        .def("AffineDimension", &AffineSubspace::AffineDimension,
+            cls_doc.AffineDimension.doc)
+        .def("Project", &AffineSubspace::Project, py::arg("x"),
+            cls_doc.Project.doc)
+        .def("ToLocalCoordinates", &AffineSubspace::ToLocalCoordinates,
+            py::arg("x"), cls_doc.ToLocalCoordinates.doc)
+        .def("ToGlobalCoordinates", &AffineSubspace::ToGlobalCoordinates,
+            py::arg("y"), cls_doc.ToGlobalCoordinates.doc)
         .def("ContainedIn", &AffineSubspace::ContainedIn, py::arg("other"),
             py::arg("tol") = 1e-15, cls_doc.ContainedIn.doc)
         .def("IsNearlyEqualTo", &AffineSubspace::IsNearlyEqualTo,
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -69,6 +69,8 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertTrue(dut.IsBounded())
         self.assertTrue(dut.PointInSet(dut.MaybeGetFeasiblePoint()))
         self.assertTrue(dut.IntersectsWith(dut))
+        self.assertTrue(dut.PointInSet(dut.Project([])))
+        self.assertEqual(dut.AffineDimension(), 0)
         self.assertTrue(dut.ContainedIn(mut.AffineSubspace()))
         self.assertTrue(dut.IsNearlyEqualTo(mut.AffineSubspace()))
 
@@ -84,6 +86,30 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertFalse(dut.IsBounded())
         self.assertTrue(dut.PointInSet(dut.MaybeGetFeasiblePoint()))
         self.assertTrue(dut.IntersectsWith(dut))
+        self.assertEqual(dut.AffineDimension(), 2)
+
+        test_point = np.array([43, 43, 0])
+        self.assertFalse(dut.PointInSet(test_point))
+        self.assertTrue(dut.PointInSet(dut.Project(test_point)))
+
+        np.testing.assert_array_equal(dut.ToGlobalCoordinates(
+            dut.ToLocalCoordinates(test_point)), dut.Project(test_point))
+        local_coords = np.array([1, -1])
+        np.testing.assert_array_equal(dut.ToLocalCoordinates(
+            dut.ToGlobalCoordinates(local_coords)),
+            local_coords.reshape(-1, 1))
+        np.testing.assert_array_equal(dut.ToLocalCoordinates(
+            dut.ToGlobalCoordinates(local_coords.reshape(-1, 1))),
+            local_coords.reshape(-1, 1))
+
+        test_point_batch = np.zeros((3, 5))
+        self.assertEqual(dut.ToLocalCoordinates(x=test_point_batch).shape,
+                         (2, 5))
+        self.assertEqual(dut.Project(x=test_point_batch).shape, (3, 5))
+        local_coords_batch = np.zeros((2, 5))
+        self.assertEqual(dut.ToGlobalCoordinates(y=local_coords_batch).shape,
+                         (3, 5))
+
         self.assertTrue(dut.ContainedIn(other=mut.AffineSubspace(
             basis, translation), tol=0))
         self.assertTrue(dut.IsNearlyEqualTo(other=mut.AffineSubspace(
--- a/geometry/optimization/affine_subspace.cc
+++ b/geometry/optimization/affine_subspace.cc
@@ -63,9 +63,7 @@ bool AffineSubspace::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
     return is_approx_equal_abstol(x, translation_, tol);
   }
   // Otherwise, project onto the flat, and compare to the input.
-  const VectorXd least_squares = basis_decomp_->solve(x - translation_);
-  const VectorXd x_sol = basis_ * least_squares + translation_;
-  return is_approx_equal_abstol(x, x_sol, tol);
+  return is_approx_equal_abstol(x, Project(x), tol);
 }
 
 std::pair<VectorX<Variable>, std::vector<Binding<Constraint>>>
@@ -119,6 +117,44 @@ std::pair<std::unique_ptr<Shape>, math::RigidTransformd>
 AffineSubspace::DoToShapeWithPose() const {
   throw std::runtime_error(
       "ToShapeWithPose is not supported by AffineSubspace.");
+}
+
+Eigen::MatrixXd AffineSubspace::Project(
+    const Eigen::Ref<const Eigen::MatrixXd>& x) const {
+  DRAKE_THROW_UNLESS(x.rows() == ambient_dimension());
+  // If the set is a point, the projection is just that point. This also
+  // directly handles the zero-dimensional case.
+  const auto maybe_point = DoMaybeGetPoint();
+  if (maybe_point) {
+    if (x.cols() == 1) {
+      return maybe_point.value();
+    } else {
+      // Outer product, which will return x.cols() copies of the feasible point.
+      return maybe_point.value() * Eigen::RowVectorXd::Ones(x.cols());
+    }
+  }
+  const Eigen::MatrixXd least_squares =
+      basis_decomp_->solve(x.colwise() - translation_);
+  return (basis_ * least_squares).colwise() + translation_;
+}
+
+Eigen::MatrixXd AffineSubspace::ToLocalCoordinates(
+    const Eigen::Ref<const Eigen::MatrixXd>& x) const {
+  DRAKE_THROW_UNLESS(x.rows() == ambient_dimension());
+  // If the set is a point, then the basis is empty, so there are no local
+  // coordinates. This behavior is handled by returning a length-zero vector.
+  // This also directly handles the zero-dimensional case.
+  auto maybe_point = DoMaybeGetPoint();
+  if (maybe_point) {
+    return Eigen::MatrixXd::Zero(0, x.cols());
+  }
+  return basis_decomp_->solve(x.colwise() - translation_);
+}
+
+Eigen::MatrixXd AffineSubspace::ToGlobalCoordinates(
+    const Eigen::Ref<const Eigen::MatrixXd>& y) const {
+  DRAKE_THROW_UNLESS(y.rows() == AffineDimension());
+  return (basis_ * y).colwise() + translation_;
 }
 
 bool AffineSubspace::ContainedIn(const AffineSubspace& other,
--- a/geometry/optimization/affine_subspace.h
+++ b/geometry/optimization/affine_subspace.h
@@ -59,6 +59,44 @@ class AffineSubspace final : public ConvexSet {
     a->Visit(MakeNameValue("translation", &translation_));
   }
 
+  /** Returns the affine dimension of this set. For an affine subspace,
+  this is simply the number of columns in the basis_ matrix. A point will
+  have affine dimension zero. */
+  int AffineDimension() const { return basis_.cols(); }
+
+  /** Computes the orthogonal projection of x onto the AffineSubspace. This
+  is achieved by finding the least squares solution y for y to x = translation_
+  + basis_*y, and returning translation_ + basis_*y. Each column of the input
+  should be a vector in the ambient space, and the corresponding column of the
+  output will be its projection onto the affine subspace.
+  @pre x.rows() == ambient_dimension() */
+  Eigen::MatrixXd Project(const Eigen::Ref<const Eigen::MatrixXd>& x) const;
+
+  /** Given a point x in the standard basis of the ambient space, returns the
+  coordinates of x in the basis of the AffineSubspace, with the zero point at
+  translation_. The component of x that is orthogonal to the AffineSubspace (if
+  it exists) is discarded, so ToGlobalCoordinates(ToLocalCoordinates(x)) is
+  equivalent to Project(x). Note that if the AffineSubspace is a point, the
+  basis is empty, so the local coordinates will also be empty (and returned as a
+  length-zero vector). Each column of the input should be a vector in the
+  ambient space, and the corresponding column of the output will be its
+  representation in the local coordinates of the affine subspace.
+  @pre x.rows() == ambient_dimension() */
+  Eigen::MatrixXd ToLocalCoordinates(
+      const Eigen::Ref<const Eigen::MatrixXd>& x) const;
+
+  /** Given a point y in the basis of the AffineSubspace, with the zero point
+  at translation_, returns the coordinates of y in the standard basis of the
+  ambient space. If the AffineSubspace is a point, it has an empty basis, so the
+  only possible local coordinates are also empty (and should be passed in as a
+  length-zero vector). Each column of the input should be a vector in the
+  affine subspace, represented in its local coordinates, and the corresponding
+  column of the output will be its representation in the coordinate system of
+  the ambient space.
+  @pre y.rows() == AffineDimension() */
+  Eigen::MatrixXd ToGlobalCoordinates(
+      const Eigen::Ref<const Eigen::MatrixXd>& y) const;
+
   /** Returns `true` if `this` AffineSubspace is contained in `other`. This is
   computed by checking if `translation()` is in `other` and then checking if
   each basis vector is in the span of the basis of `other`. The latter step
--- a/geometry/optimization/test/affine_subspace_test.cc
+++ b/geometry/optimization/test/affine_subspace_test.cc
@@ -24,6 +24,14 @@ GTEST_TEST(AffineSubspaceTest, DefaultCtor) {
   EXPECT_TRUE(dut.PointInSet(dut.MaybeGetFeasiblePoint().value()));
   EXPECT_TRUE(dut.PointInSet(Eigen::VectorXd::Zero(0)));
   EXPECT_TRUE(dut.IntersectsWith(dut));
+  EXPECT_NO_THROW(dut.Project(Eigen::VectorXd::Zero(0)));
+  EXPECT_EQ(dut.AffineDimension(), 0);
+  Eigen::VectorXd test_point(0);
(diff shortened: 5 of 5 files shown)


cohnt:
+@hongkai-dai for feature review, please!

hongkai-dai (commented):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ3tID7Hl2uP9SDUly:-NaQ3tID7Hl2uP9SDUlz:b-6kt0b9) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L71)):*
> ```objc
>   + basis_*y, and returning translation_ + basis_*y.
>   @pre x.size() == ambient_dimension() */
>   Eigen::VectorXd Project(const Eigen::Ref<const Eigen::VectorXd>& x) const;
> ```

BTW, it is possible and better to project multiple points instead of a single point. Such "batch" operation will speed up the code (especially in pydrake).
___
*[`geometry/optimization/affine_subspace.h` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4iJEB_lS5thVEW2S:-NaQ4iJEB_lS5thVEW2T:b-89a4l7) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L81)):*
> ```objc
>   length-zero vector).
>   @pre x.size() == ambient_dimension() */
>   Eigen::VectorXd ToLocalCoordinates(
> ```

Ditto
___
*[`geometry/optimization/affine_subspace.h` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4jTtEtsa6TOcpj9V:-NaQ4jTtEtsa6TOcpj9W:b-89a4l7) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L90)):*
> ```objc
>   length-zero vector).
>   @pre y.size() == AffineDimension() */
>   Eigen::VectorXd ToGlobalCoordinates(
> ```

Ditto
___
*[`geometry/optimization/affine_subspace.cc` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4KU7DgFM5oL7QV7t:-NaQ4KU7DgFM5oL7QV7u:b-enwvae) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.cc#L131)):*
> ```cpp
>     return maybe_point.value();
>   }
>   VectorXd least_squares = basis_decomp_->solve(x - translation_);
> ```

BTW, use const for the data type `const VectorXd`.

hongkai-dai (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ3tID7Hl2uP9SDUly:-NaTiLb5CdYqLmaSdTaF:b5bs03h) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L71)):*

BTW, it is possible and better to project multiple points instead of a single point. Such "batch" operation will speed up the code (especially in pydrake).
</blockquote></details>

To be more precise, I am suggesting this
```cc
// Project each column of x onto the affine subspace.
Eigen::MatrixXd AffineSubspace::Project(const Eigen::Ref<const Eigen::MatrixXd>& x) const {
  DRAKE_THROW_UNLESS(x.rows() == ambient_dimension());
  // If the set is a point, the projection is just that point. This also
  // directly handles the zero-dimensional case.
  const auto maybe_point = DoMaybeGetPoint();
  if (maybe_point) {
    if (x.cols() == 1) {
      return maybe_point.value();
    } else {
      return maybe_point.value() * Eigen::RowVectorXd::Ones(x.cols());
    }
  }
  const Eigen::MatrixXd least_squares = basis_decomp_->solve(x.colwise() - translation_);
  return (basis_ * least_squares).colwise() + translation_;
}
```
I haven't tested the code, but I think you can get a gist of it.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ3tID7Hl2uP9SDUly:-NaRnypj5G25TRprdbQo:b-9sum5i) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L71)):*

To be more precise, I am suggesting this
```cc
// Project each column of x onto the affine subspace.
Eigen::MatrixXd AffineSubspace::Project(const Eigen::Ref<const Eigen::MatrixXd>& x) const {
  DRAKE_THROW_UNLESS(x.rows() == ambient_dimension());
  // If the set is a point, the projection is just that point. This also
  // directly handles the zero-dimensional case.
  const auto maybe_point = DoMaybeGetPoint();
  if (maybe_point) {
    if (x.cols() == 1) {
      return maybe_point.value();
    } else {
      return maybe_point.value() * Eigen::RowVectorXd::Ones(x.cols());
    }
  }
  const Eigen::MatrixXd least_squares = basis_decomp_->solve(x.colwise() - translation_);
  return (basis_ * least_squares).colwise() + translation_;
}
```
I haven't tested the code, but I think you can get a gist of it.
</blockquote></details>

Done.

I also added new test cases (in C++ and Python) to verify that the batching works.
___
*[`geometry/optimization/affine_subspace.h` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4iJEB_lS5thVEW2S:-NacV24S5bDlBAPouf0A:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L81)):*

Ditto
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.h` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4jTtEtsa6TOcpj9V:-NacV4Bl55lwtE39z529:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L90)):*

Ditto
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 131 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4KU7DgFM5oL7QV7t:-NacV8bnEJwlv9LTAoYg:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.cc#L131)):*

BTW, use const for the data type `const VectorXd`.
</blockquote></details>

Done.
___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NacV9q67EAz1JLsVJJf:b-b2hh9) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*
> ```Python
>         np.testing.assert_array_equal(dut.ToLocalCoordinates(
>             dut.ToGlobalCoordinates(local_coords)),
>             local_coords.reshape(-1, 1))
> ```

Note that the Python bindings now always return column vectors, even when called with a 1d numpy vector. But I think this is okay -- VPolytope has similar behavior:
```
import numpy as np
from pydrake.all import VPolytope
v = VPolytope(np.array([1, 2, 3]))
print(v.vertices().shape)
print(v.ambient_dimension())
```

hongkai-dai (commented):
> Reviewed 1 of 5 files at r1, 3 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 71 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ3tID7Hl2uP9SDUly:-Nae7EjTAu79ScXX62EW:bnns9f7) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L71)):*

Done.

I also added new test cases (in C++ and Python) to verify that the batching works.
</blockquote></details>

BTW, in the documentation of this class, mention that each column of `x` represents a point to be projected, and column i of the returned matrix is the projected point of x.col(i).

Also the current documentation says `@pre x.size == ambient_dimension()`, this should be changed to `@pre x.rows() == ambient_dimension()`.
___
*[`geometry/optimization/affine_subspace.h` line 81 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4iJEB_lS5thVEW2S:-Nae7Ytz2FbgD7_-531N:b9suka) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L81)):*

Done.
</blockquote></details>

BTW, ditto on the documentation.
___
*[`geometry/optimization/affine_subspace.h` line 90 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaQ4jTtEtsa6TOcpj9V:-Nae7f015-C1fvfNOlPw:b9suka) ([raw file](https://github.com/RobotLocomotion/drake/blob/8b15dfbf97e898966f8287abf06c6f8657a06cc8/geometry/optimization/affine_subspace.h#L90)):*

Done.
</blockquote></details>

BTW, ditto on the documentation.
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 412 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NaeHcrf2R_wkRYQujMQ:-NaeHcrf2R_wkRYQujMR:bxhu25q) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/geometry/optimization/test/affine_subspace_test.cc#L412)):*
> ```cpp
>   // clang-format on
> 
>   Eigen::MatrixXd projected = as.Project(points);
> ```

Also check that `projected.col(i) = as.Project(points.col(i))`?
___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NaeHxqT5Amf2hjrODgL:boy1qx4) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

Note that the Python bindings now always return column vectors, even when called with a 1d numpy vector. But I think this is okay -- VPolytope has similar behavior:
```
import numpy as np
from pydrake.all import VPolytope
v = VPolytope(np.array([1, 2, 3]))
print(v.vertices().shape)
print(v.ambient_dimension())
```
</blockquote></details>

Interesting. It might surprise the user if the input dimension and the output dimension are different, that the input being a 1d numpy vector returns the output as 2-dimensional array with one dimension being 1. @EricCousineau-TRI and @jwnimmer-tri , do we have a way to resolve this in pydrake?

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NagCgzB-z8TltDpEH_I:bj2g56e) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

Interesting. It might surprise the user if the input dimension and the output dimension are different, that the input being a 1d numpy vector returns the output as 2-dimensional array with one dimension being 1. @EricCousineau-TRI and @jwnimmer-tri , do we have a way to resolve this in pydrake?
</blockquote></details>

We have `WrapToMatchInputShape`, but that's only for the output matching the same exact shape. We do not yet have anything to match input dimension.
I will try for <15min, but will have to timebox.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NagJKRR2mpgQT3zo0EU:b-9sgfnn) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

I'm not too worried about matching the input dimension. The python code seems to naturally promote 1D vectors to column vectors (as we would expect). The only ambiguity is if an nxn array is passed as input, but we clearly document that it should be column-major. I'll give `WrapToMatchInputShape` a try.
</blockquote></details>

Actually, I think this isn't a good idea, because `ToLocalCoordinates` and `ToGlobalCoordinates` are changing the shape of the array. (They reduce or increase the number of rows, based on the ambient dimension and affine dimension.)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NagLKQl5_cc_pdxZ99p:b-sgkbo3) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

yup, I'm writing something to either squeeze or unsqueeze based on the input
</blockquote></details>

I tried it here, it seems a bit fragile: #19887
I probably won't be able to move this forward any more this week, so I'd recommend keeping this as-is or doing manual reshaping in the bindings if y'all are concerned and don't want to make the general case work.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 4 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NagQl9Z59ADkDL3yylh:b-h4qjc2) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

I tried it here, it seems a bit fragile: #19887
I probably won't be able to move this forward any more this week, so I'd recommend keeping this as-is or doing manual reshaping in the bindings if y'all are concerned and don't want to make the general case work.
</blockquote></details>

My preference would be to keep it as-is for now. The output is clearly specified to be a column vector, so the user can easily reshape if need be.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: 3 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-NacV9q67EAz1JLsVJJe:-NagYaE09pPS1iUC1idO:b-bk5ipd) ([raw file](https://github.com/RobotLocomotion/drake/blob/05f5e5356d924eea49db1762aa48158b6828a114/bindings/pydrake/geometry/test/optimization_test.py#L98)):*

My preference would be to keep it as-is for now. The output is clearly specified to be a column vector, so the user can easily reshape if need be.
</blockquote></details>

Works for me!

cohnt:
Merge conflicts should be fixed now. @drake-jenkins-bot test this please

hongkai-dai (commented):
:lgtm: +@rpoyner-tri for platform review please, thanks!

> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-:-Nalt_fhBjk4-5E2_bJM:b-sjvomj)* status: LGTM missing from assignee rpoyner-tri(platform)

rpoyner-tri (commented):
:lgtm: bindings test quibble

> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-:-NalykgF9J648Fyumfk-:br3q4kz)* status: 1 unresolved discussion

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 72 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-Nalxx8mEWbL4Rl6yYJJ:-Nalxx8nCwINpXB59OXT:b-5x2ajb) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f9f130baaea75ad43d495b4adc1bad86c40be8f/bindings/pydrake/geometry/test/optimization_test.py#L72)):*
> ```Python
>         self.assertTrue(dut.PointInSet(dut.MaybeGetFeasiblePoint()))
>         self.assertTrue(dut.IntersectsWith(dut))
>         self.assertTrue(dut.PointInSet(dut.Project([])))
> ```

minor An important part of bindings tests is to make sure the `py::arg` name strings are spelled right. The way to do that in the test is just to use them (even if not really necessary). For example, `Project(x=thing)`.

Looks like `Project`, `ToLocalCoordinates`, and `ToGlobalCoordinates` have untested arg strings.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),hongkai-dai

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 72 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-Nalxx8mEWbL4Rl6yYJJ:-Nam26ek99270p_aOMjN:bishevs) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f9f130baaea75ad43d495b4adc1bad86c40be8f/bindings/pydrake/geometry/test/optimization_test.py#L72)):*

minor An important part of bindings tests is to make sure the `py::arg` name strings are spelled right. The way to do that in the test is just to use them (even if not really necessary). For example, `Project(x=thing)`.

Looks like `Project`, `ToLocalCoordinates`, and `ToGlobalCoordinates` have untested arg strings.
</blockquote></details>

Done. (I added name string tests in lines 106-110.)

hongkai-dai (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),hongkai-dai

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19851)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),hongkai-dai

___
*[`bindings/pydrake/geometry/test/optimization_test.py` line 72 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19851#-Nalxx8mEWbL4Rl6yYJJ:-NamGEuB3aGm8-kjJ8zc:b-ucfknh) ([raw file](https://github.com/RobotLocomotion/drake/blob/3f9f130baaea75ad43d495b4adc1bad86c40be8f/bindings/pydrake/geometry/test/optimization_test.py#L72)):*

Done. (I added name string tests in lines 106-110.)
</blockquote></details>

Thanks!

PR #19854, opened 2023-07-27
Deformable geometries support collision filtering.

xuchenhan-tri:
To enable collision filtering for deformable geometries, we need to modify the convention that deformable geometries are always registered to the world frame. This is because collision among geometries registered with the same frame is automatically filtered and as a result, collision between deformable geometries and anchored rigid geometries are filtered, which is usually not what we want.

This commit changes the convention so that deformable geomtries are never registered with the world frame, and it enables collision filtering for deformable bodies.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19854)

### DIFF
--- a/geometry/geometry_state.cc
+++ b/geometry/geometry_state.cc
@@ -256,11 +256,6 @@ int GeometryState<T>::NumGeometriesWithRole(Role role) const {
 
 template <typename T>
 int GeometryState<T>::NumDynamicGeometries() const {
-  return NumDeformableGeometries() + NumDynamicNonDeformableGeometries();
-}
-
-template <typename T>
-int GeometryState<T>::NumDynamicNonDeformableGeometries() const {
   int count = 0;
   for (const auto& pair : frames_) {
     const InternalFrame& frame = pair.second;
@@ -826,9 +821,10 @@ GeometryId GeometryState<T>::RegisterDeformableGeometry(
   }
 
   const GeometryId geometry_id = geometry->id();
-  if (frame_id != InternalFrame::world_frame_id()) {
+  if (frame_id == InternalFrame::world_frame_id()) {
     throw std::logic_error("Registering deformable geometry with id " +
-                           to_string(geometry_id) + " to a non-world frame");
+                           to_string(geometry_id) +
+                           " to the world frame is not allowed.");
   }
 
   ValidateRegistrationAndSetTopology(source_id, frame_id, geometry_id);
--- a/geometry/geometry_state.h
+++ b/geometry/geometry_state.h
@@ -173,10 +173,6 @@ class GeometryState {
   /** Implementation of SceneGraphInspector::NumDynamicGeometries().  */
   int NumDynamicGeometries() const;
 
-  /** Returns the total number of registered dynamic non-deformable geometries.
-   */
-  int NumDynamicNonDeformableGeometries() const;
-
   /** Returns the total number of registered deformable geometries. All
    deformable geometries are dynamic and _not_ anchored.  */
   int NumDeformableGeometries() const;
--- a/geometry/proximity/BUILD.bazel
+++ b/geometry/proximity/BUILD.bazel
@@ -183,6 +183,7 @@ drake_cc_library(
     srcs = ["deformable_contact_internal.cc"],
     hdrs = ["deformable_contact_internal.h"],
     deps = [
+        ":collision_filter",
         ":deformable_mesh_intersection",
         "//common:essential",
         "//geometry:geometry_ids",
--- a/geometry/proximity/deformable_contact_internal.cc
+++ b/geometry/proximity/deformable_contact_internal.cc
@@ -53,7 +53,8 @@ void Geometries::UpdateDeformableVertexPositions(
   }
 }
 
-DeformableContact<double> Geometries::ComputeDeformableContact() const {
+DeformableContact<double> Geometries::ComputeDeformableContact(
+    const CollisionFilter& collision_filter) const {
   DeformableContact<double> result;
   for (const auto& [deformable_id, deformable_geometry] :
        deformable_geometries_) {
@@ -61,13 +62,18 @@ DeformableContact<double> Geometries::ComputeDeformableContact() const {
         deformable_geometry.deformable_mesh().mesh();
     result.RegisterDeformableGeometry(
         deformable_id, deformable_mesh.num_vertices());
+    DRAKE_DEMAND(collision_filter.HasGeometry(deformable_id));
     for (const auto& [rigid_id, rigid_geometry] : rigid_geometries_) {
-      const math::RigidTransform<double>& X_WR = rigid_geometry.pose_in_world();
-      const auto& rigid_bvh = rigid_geometry.rigid_mesh().bvh();
-      const auto& rigid_tri_mesh = rigid_geometry.rigid_mesh().mesh();
-      AddDeformableRigidContactSurface(deformable_geometry, deformable_id,
-                                       rigid_id, rigid_tri_mesh, rigid_bvh,
-                                       X_WR, &result);
+      DRAKE_DEMAND(collision_filter.HasGeometry(rigid_id));
+      if (collision_filter.CanCollideWith(deformable_id, rigid_id)) {
+        const math::RigidTransform<double>& X_WR =
+            rigid_geometry.pose_in_world();
+        const auto& rigid_bvh = rigid_geometry.rigid_mesh().bvh();
+        const auto& rigid_tri_mesh = rigid_geometry.rigid_mesh().mesh();
+        AddDeformableRigidContactSurface(deformable_geometry, deformable_id,
+                                         rigid_id, rigid_tri_mesh, rigid_bvh,
+                                         X_WR, &result);
+      }
     }
   }
   return result;
--- a/geometry/proximity/deformable_contact_internal.h
+++ b/geometry/proximity/deformable_contact_internal.h
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "drake/geometry/geometry_ids.h"
+#include "drake/geometry/proximity/collision_filter.h"
 #include "drake/geometry/proximity/deformable_contact_geometries.h"
 #include "drake/geometry/proximity/volume_mesh.h"
 #include "drake/geometry/query_results/deformable_contact.h"
@@ -103,7 +104,8 @@ class Geometries final : public ShapeReifier {
    with respect to all registered rigid geometries. Assumes the vertex positions
    and poses of all registered deformable and rigid geometries are up to date.
   */
-  DeformableContact<double> ComputeDeformableContact() const;
+  DeformableContact<double> ComputeDeformableContact(
+      const CollisionFilter& collision_filter) const;
 
  private:
   friend class GeometriesTester;
--- a/geometry/proximity/test/deformable_contact_internal_test.cc
+++ b/geometry/proximity/test/deformable_contact_internal_test.cc
@@ -12,6 +12,16 @@
 
 namespace drake {
 namespace geometry {
+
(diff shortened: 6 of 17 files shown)


xuchenhan-tri (commented):
+@SeanCurtis-TRI could you help me take a look at my attempt to include deformable geometries in the collision filter eco-system?

Here's some context: I'm trying to enable fixed constraint between vertices of deformable bodies and rigid bodies and ran head first into the need to collision filter between deformable and rigid. We need to filter the contact constraints between deformable tets and rigid bodies if any of the vertices incident to the tet in contact is subject to fixed constraints. Otherwise, the two constraints fight each other. Ideally, we should do this on a vertex/tet level, but that seems to be complicated. So for the first pass, I want to enable collision filtering between deformable and rigid bodies on an object level.

Please take a look and see if my approach is sound and if there's anything I'm missing. If the general strategy looks good to you then I'll go ahead and round the edges for this PR. Thank you!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-:-NaJtrgiCOEB7yoSsnIZ:bwdvjy6)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

SeanCurtis-TRI (commented):
> Reviewed 13 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19854)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-NaKqNZk6UqKKRFs1eEc:-NaKqNZk6UqKKRFs1eEd:be0etow) (no related file):*
My fundamental question is about what it means to define a frame F (F != W) against which we define the deformable vertex locations. Whoever registered that deformable geometry is now obliged to output both the vertex positions *and* the frame position in perpetuity. However, I'd assume that, generally, `X_FW = I`.

Also, by defining F, one could register a rigid geometry to `F` and now the deformable geometry won't collide against that geometry. That feels more like a mistake than an intended outcome.

It seems to me that there must be a method that is less awkward for deformable bodies.

Proposal:

1. By default, deformable geometries have no collision filters defined (neither `SceneGraph` nor `MultibodyPlant` would introduce them.
2. However, they could be *added* to a collision filter explicitly. Either via the APIs or via declared collision filter groups in dmd.yaml or some such thing (can we *parse* into a deformable body?)

Implementation:

It's easy to achieve this on the `SceneGraph` side -- at the moment we apply collision filters between a geometry and all other geometries affixed to its frame, we have perfect knowledge of whether the geometry is deformable or not and we could skip the filtering act.

It'd be a bit trickier to handle things on the `MbP` side. It has the logic that defines collision filters between *adjacent* bodies. Therefore, if the deformable body is registered to the world, and some rigid body affixed to a frame is welded to the world, then that `MbP` will want to declare a collision filter. I think the solution is to slightly tweak the `CollisionFilterDeclaration` semantics.

Essentially, we'd like to say that our `CollisionFilterDeclaration` *excludes* deformable geometries.  So, a call looks like:

```c++
CollisionFilterDeclaration().ExcludeWithin(geometries);
```

What if we were to provide an optional construction argument:

```c++
CollisionFilterDeclaration(FilterDomain::kOmitDeformable).ExcludeWithin(geometries);
```

The current call would include geometries of all types. The latter would omit any of the ids in `geometries` that map to deformable geometries and `MbP` would use that when it creates its various collision filters based on adjacency and welded subgraphs. (We could also reverse those semantics and make people opt in to filters on deformable geometries. That might produce less churn.)

Thoughts?

___
___
*[`geometry/proximity/deformable_contact_internal.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-NaKvNqz-F_J3gwO4ERd:-NaKvNqz-F_J3gwO4ERe:b-q5kkez) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f58ac07e80b1163c0914ca153842386afc48f23/geometry/proximity/deformable_contact_internal.cc#L65)):*
> ```cpp
>     result.RegisterDeformableGeometry(
>         deformable_id, deformable_mesh.num_vertices());
>     DRAKE_DEMAND(collision_filter.HasGeometry(deformable_id));
> ```

BTW You might want to make these `DRAKE_ASSERT` because this is a log_2 search across all registered proximity geometries.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19854)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-NaKqNZk6UqKKRFs1eEc:-NaNYZEj1UkiAdnKYi5W:b-vesxp2) (no related file):*
> Whoever registered that deformable geometry is now obliged to output both the vertex positions and the frame position in perpetuity. 

Indeed, that is undesirable.

> Also, by defining F, one could register a rigid geometry to `F` and now the deformable geometry won't collide against that geometry. That feels more like a mistake than an intended outcome.

That's a good point. However, the same is true if we keep the deformable geometries on the same world frame with the geometries. 

> Proposal:
>1. By default, deformable geometries have no collision filters defined (neither SceneGraph nor MultibodyPlant would introduce them.
>2. However, they could be added to a collision filter explicitly. Either via the APIs or via declared collision filter groups in dmd.yaml or some such thing (can we parse into a deformable body?)

I agree with this proposal. This is exactly what I want to achieve. It's also what I tried to achieve with this draft PR, but like you pointed out, it's not quite there yet.

> Implementation: ...

I feel less ensuthasitc about the proposed implementation for two reasons

1. Having CollisionFilterDeclaration understand the concept of a geometry being deformable seems awkward to me. IMO, all geometries should appear as purely abstract indices to the collision filter.
2. It seems easy to accidentally filter deformable rigid collisions when they are registered to the same (world) frame since the convention that all geometries registered to the same frame should be filtered is so deep rooted.

How do you feel about this alternative implementation:

Instead of registering deformable geometries to the world frame like we do now, create another static frame F such that X_WF = I and register all deformable geometries to that frame. The frame F is for deformable geometries only and we keep F internal to scene graph so that we won't accidentally register rigid geometries to that frame. Then when registering geometries to SG, we omit the notion of a frame, i.e.
```cpp
GeometryId RegisterDeformableGeometry(SourceId source_id, FrameId frame_id, 
                                      std::unique_ptr<GeometryInstance> geometry,
                                      double resolution_hint);
```
With the frame being internal to SG, we don't need external sources to update its pose either.

Then for collision filtering, default no filtering for anything involving deformable geometries and users opt in to filtering based on geometry ids.

___
___
*[`geometry/proximity/deformable_contact_internal.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-NaKvNqz-F_J3gwO4ERd:-NaOLnHC9FY0-Kfv-s0z:bgzdlde) ([raw file](https://github.com/RobotLocomotion/drake/blob/5f58ac07e80b1163c0914ca153842386afc48f23/geometry/proximity/deformable_contact_internal.cc#L65)):*

BTW You might want to make these `DRAKE_ASSERT` because this is a log_2 search across all registered proximity geometries.
</blockquote></details>

Good to know, thanks!

SeanCurtis-TRI:

> Whoever registered that deformable geometry is now obliged to output both the vertex positions and the frame position in perpetuity. 

Indeed, that is undesirable.

> Also, by defining F, one could register a rigid geometry to `F` and now the deformable geometry won't collide against that geometry. That feels more like a mistake than an intended outcome.

That's a good point. However, the same is true if we keep the deformable geometries on the same world frame with the geometries. 

> Proposal:
>1. By default, deformable geometries have no collision filters defined (neither SceneGraph nor MultibodyPlant would introduce them.
>2. However, they could be added to a collision filter explicitly. Either via the APIs or via declared collision filter groups in dmd.yaml or some such thing (can we parse into a deformable body?)

I agree with this proposal. This is exactly what I want to achieve. It's also what I tried to achieve with this draft PR, but like you pointed out, it's not quite there yet.

> Implementation: ...

I feel less ensuthasitc about the proposed implementation for two reasons

1. Having CollisionFilterDeclaration understand the concept of a geometry being deformable seems awkward to me. IMO, all geometries should appear as purely abstract indices to the collision filter.
2. It seems easy to accidentally filter deformable rigid collisions when they are registered to the same (world) frame since the convention that all geometries registered to the same frame should be filtered is so deep rooted.

How do you feel about this alternative implementation:

Instead of registering deformable geometries to the world frame like we do now, create another static frame F such that X_WF = I and register all deformable geometries to that frame. The frame F is for deformable geometries only and we keep F internal to scene graph so that we won't accidentally register rigid geometries to that frame. Then when registering geometries to SG, we omit the notion of a frame, i.e.
```cpp
GeometryId RegisterDeformableGeometry(SourceId source_id, FrameId frame_id, 
                                      std::unique_ptr<GeometryInstance> geometry,
                                      double resolution_hint);
```
With the frame being internal to SG, we don't need external sources to update its pose either.

Then for collision filtering, default no filtering for anything involving deformable geometries and users opt in to filtering based on geometry ids.
</blockquote></details>

Update based on our f2f -- here's my understanding of what we agreed to.

A permutation of my proposal.

1. MultibodyPlant and SceneGraph both promise to *not automatically* apply collision filters on deformable geometries.
2. The default semantics of `CollisionFilterDeclaration` remains unchanged. Collision filter declarations on a set that includes one or more deformable geometry ids will affect that geometry.
3. However, we add an optional argument that allows someone declaring collision filters to *omit* classes of geometry -- currently, the only class under discussion is deformable.
4. SG and MbP make use of the new omission functionality to explicitly omit geometries when they declare auto-collision filters.
5. This is implemented by
    - adding an enumeration which defines the `kOmitDeformable` (or whatever it is named)
    - changing the signature of the `CollisionFilter::ExtractIds` function to include that enumeration.
    - changing `GeometryState`'s extract ids function to make use of the omission parameter.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19854)* status: 2 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19854#-NaKqNZk6UqKKRFs1eEc:-NaOgL1L82HZh6rk-Gti:b-ogpu1u) (no related file):*

Update based on our f2f -- here's my understanding of what we agreed to.

A permutation of my proposal.

1. MultibodyPlant and SceneGraph both promise to *not automatically* apply collision filters on deformable geometries.
2. The default semantics of `CollisionFilterDeclaration` remains unchanged. Collision filter declarations on a set that includes one or more deformable geometry ids will affect that geometry.
3. However, we add an optional argument that allows someone declaring collision filters to *omit* classes of geometry -- currently, the only class under discussion is deformable.
4. SG and MbP make use of the new omission functionality to explicitly omit geometries when they declare auto-collision filters.
5. This is implemented by
    - adding an enumeration which defines the `kOmitDeformable` (or whatever it is named)
    - changing the signature of the `CollisionFilter::ExtractIds` function to include that enumeration.
    - changing `GeometryState`'s extract ids function to make use of the omission parameter.

</blockquote></details>

Thanks for the nicely written summary. You beat me to it!

I'll work on another WIP PR and put it out for review.

___

SeanCurtis-TRI:
Do you want to close this one out?

xuchenhan-tri:
Superseded by #19897

PR #19857, opened 2023-07-27
Implements Hyperellipsoid::MinimumVolumeCircumscribedEllipsoid

RussTedrake:
Also known as the "outer Loewner-John" ellipsoid.

+@AlexandreAmice for feature review, please?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -198,6 +198,10 @@ void DefineGeometryOptimization(py::module m) {
             py::arg("radius"), py::arg("center"), cls_doc.MakeHypersphere.doc)
         .def_static("MakeUnitBall", &Hyperellipsoid::MakeUnitBall,
             py::arg("dim"), cls_doc.MakeUnitBall.doc)
+        .def_static("MinimumVolumeCircumscribedEllipsoid",
+            &Hyperellipsoid::MinimumVolumeCircumscribedEllipsoid,
+            py::arg("points"), py::arg("rank_tol") = 1e-6,
+            cls_doc.MinimumVolumeCircumscribedEllipsoid.doc)
         .def(py::pickle(
             [](const Hyperellipsoid& self) {
               return std::make_pair(self.A(), self.center());
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -217,6 +217,11 @@ class TestGeometryOptimization(unittest.TestCase):
         e_ball3 = mut.Hyperellipsoid.MakeUnitBall(dim=3)
         np.testing.assert_array_equal(e_ball3.A(), self.A)
         np.testing.assert_array_equal(e_ball3.center(), [0, 0, 0])
+        points = np.array([[1, 0], [-1, 0], [0, 2], [0, -2]]).T
+        e_lowner_john = mut.Hyperellipsoid.MinimumVolumeCircumscribedEllipsoid(
+            points=points, rank_tol=1e-2)
+        e_lowner_john = mut.Hyperellipsoid.MinimumVolumeCircumscribedEllipsoid(
+            points=points)
 
     def test_minkowski_sum(self):
         mut.MinkowskiSum()
--- a/geometry/optimization/BUILD.bazel
+++ b/geometry/optimization/BUILD.bazel
@@ -402,6 +402,7 @@ drake_cc_googletest(
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
         "//common/yaml",
+        "//solvers:mosek_solver",
     ],
 )
 
--- a/geometry/optimization/hpolyhedron.h
+++ b/geometry/optimization/hpolyhedron.h
@@ -106,8 +106,9 @@ class HPolyhedron final : public ConvexSet {
   negative tol means it is less likely to remote a constraint.  */
   [[nodiscard]] HPolyhedron ReduceInequalities(double tol = 1E-9) const;
 
-  /** Solves a semi-definite program to compute the inscribed ellipsoid.
-  From Section 8.4.2 in Boyd and Vandenberghe, 2004, we solve
+  /** Solves a semi-definite program to compute the inscribed ellipsoid. This is
+  also known as the inner Löwner-John ellipsoid. From Section 8.4.2 in Boyd and
+  Vandenberghe, 2004, we solve
   @verbatim
   max_{C,d} log det (C)
         s.t. |aᵢC|₂ ≤ bᵢ - aᵢd, ∀i
--- a/geometry/optimization/hyperellipsoid.cc
+++ b/geometry/optimization/hyperellipsoid.cc
@@ -28,6 +28,7 @@ using solvers::Constraint;
 using solvers::MathematicalProgram;
 using solvers::VectorXDecisionVariable;
 using std::sqrt;
+using symbolic::Expression;
 using symbolic::Variable;
 
 Hyperellipsoid::Hyperellipsoid()
@@ -175,6 +176,80 @@ Hyperellipsoid Hyperellipsoid::MakeHypersphere(
 Hyperellipsoid Hyperellipsoid::MakeUnitBall(int dim) {
   DRAKE_THROW_UNLESS(dim > 0);
   return Hyperellipsoid(MatrixXd::Identity(dim, dim), VectorXd::Zero(dim));
+}
+
+Hyperellipsoid Hyperellipsoid::MinimumVolumeCircumscribedEllipsoid(
+    const Eigen::Ref<const MatrixXd>& points, double rank_tol) {
+  DRAKE_THROW_UNLESS(points.allFinite());
+  const int dim = points.rows();
+  int rank;
+  const int n = points.cols();
+
+  // Check the numerical rank of the data matrix.
+  std::optional<Eigen::MatrixXd> U;
+  Eigen::VectorXd mean = points.rowwise().mean();
+  {
+    auto svd = (points.colwise() - mean).bdcSvd(Eigen::ComputeThinU);
+    // Eigen's SVD rank never returns zero, and their singular values are
+    // returned in decreasing order.
+    if (svd.singularValues()[0] < rank_tol) {
+      throw std::runtime_error(fmt::format(
+          "The numerical rank of the points appears to be zero. (The largest "
+          "singular value is {}, which is below rank_tol = {})",
+          svd.singularValues()[0], rank_tol));
+    }
+    svd.setThreshold(rank_tol);
+    rank = svd.rank();
+    if (rank < dim) {
+      U = svd.matrixU().leftCols(rank);
+    }
+  }
+
+  MathematicalProgram prog;
+  solvers::MatrixXDecisionVariable A =
+      prog.NewSymmetricContinuousVariables(rank, "A");
+  prog.AddMaximizeLogDeterminantCost(A.cast<Expression>());
+  solvers::VectorXDecisionVariable b = prog.NewContinuousVariables(rank, "b");
+  // TODO(russt): Avoid the symbolic computation here and write A_lorentz
+  // directly, s.t. v = A_lorentz * vars + b_lorentz, where A=Aᵀ and b are the
+  // vars.
+  VectorX<Expression> v(rank + 1);
+  v[0] = 1;
+  for (int i = 0; i < n; ++i) {
+    if (U) {  // rank < dim
+      // |AUᵀ(x-mean) + b|₂ <= 1, written as a Lorentz cone with v = [1;
+      // AUᵀ(x-mean) + b].
+      v.tail(rank) = A * U->transpose() * (points.col(i) - mean) + b;
+    } else {  // rank == dim
+      // |Ax + b|₂ <= 1, written as a Lorentz cone with v = [1; A * x + b].
+      v.tail(rank) = A * points.col(i) + b;
+    }
+    prog.AddLorentzConeConstraint(v);
+  }
+
+  solvers::MathematicalProgramResult result = Solve(prog);
+  if (!result.is_success()) {
+    throw std::runtime_error(
+        fmt::format("The MathematicalProgram was not solved successfully. The "
+                    "ambient dimension is {} and the data rank is {}, computed "
+                    "using rank_tol={}. Consider adjusting rank_tol.",
+                    dim, rank, rank_tol));
+  }
+
+  // Ax + b => A(x-c) = Ax - Ac => c = -A^{-1}b
+  const MatrixXd A_sol = result.GetSolution(A);
+  const VectorXd b_sol = result.GetSolution(b);
+  // Note: We can use llt() because know that A will be positive definite;
+  // there is a PSD constraint, but we are maximizing the eigenvalues of A and
+  // the convex hull of the points is guaranteed to be bounded.
+  const VectorXd c = A_sol.llt().solve(-b_sol);
+
+  if (U) {
+    // AUᵀ(x-mean) + b => AUᵀ(x - center), so center = -UA⁻¹b + mean.
+    return Hyperellipsoid(A_sol * U->transpose(), (*U) * c + mean);
+  } else {
+    return Hyperellipsoid(A_sol, c);
+  }
 }
 
 std::unique_ptr<ConvexSet> Hyperellipsoid::DoClone() const {
--- a/geometry/optimization/hyperellipsoid.h
+++ b/geometry/optimization/hyperellipsoid.h
@@ -98,6 +98,27 @@ class Hyperellipsoid final : public ConvexSet {
   /** Constructs the L₂-norm unit ball in `dim` dimensions, {x | |x|₂ <= 1 }. */
   static Hyperellipsoid MakeUnitBall(int dim);
 
+  /** Constructs the minimum-volume ellipsoid which contains all of the
+  `points`. This is commonly referred to as the outer Löwner-John ellipsoid.
+
+  @param points is a d-by-n matrix, where d is the ambient dimension and each
+  column represents one point.
+  @param rank_tol the singular values of the data matrix will be considered
+  non-zero if they are strictly greater than `rank_tol` * `max_singular_value`.
+  The default is 1e-6 to be compatible with common solver tolerances. This is
+  used to detect if the data lies on a lower-dimensional affine space than the
+  ambient dimension of the ellipsoid.
+  @throws std::exception if the MathematicalProgram fails to solve. If this
+  were to happen (due to numerical issues), then increasing `rank_tol` should
+  provide a mitigation.
+  @throw std::exception if points includes NaNs or infinite values.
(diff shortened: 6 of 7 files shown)


AlexandreAmice (commented):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)* status: 5 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/hyperellipsoid.h` line 105 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOUsYB26BPMcD0Xvu-:-NaOUsYB26BPMcD0Xvu0:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.h#L105)):*
> ```objc
> 
>   @param points is a d-by-n matrix, where d is the ambient dimension and each
>   colum represents one point.
> ```

nit: typo



_Suggestion:_
```objc
column
```
___
*[`geometry/optimization/hyperellipsoid.h` line 111 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOYjZl-oIlLf-V5f0E:-NaOYjZl-oIlLf-V5f0F:bb4re01) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.h#L111)):*
> ```objc
>   the resulting ellipsoid. This is particularly important for the case when the
>   points lie in a lower-dimensional subspace. We have det(A) ≥ det(Z) = ∏ᵢ Zᵢᵢ,
>   and the objective is maximizing log(det(Z)).
> ```

This isn't very clear to me. What is `Z` in this case? Is it the dual variable? What is the guidance for setting `Z_ii`?

I understand the problem if `points` are from an affine subspace, then one of the eigenvalues of `A` will go to zero causing the objective to be unbounded. So the point is that we wish to have a finite upper bound on this.

I believe a more appropriate way to test this would be to check if the numerical rank of `point` is lower than the ambient dimension and if so project to the lower dimensional manifold.



_Code quote:_
```objc
  @param Zii_ub provides some conditioning for the resulting matrix A via an
  upper bound on the appropriate slack variables.  If these constraints become
  active, they will increase the volume; all points will still be contained in
  the resulting ellipsoid. This is particularly important for the case when the
  points lie in a lower-dimensional subspace. We have det(A) ≥ det(Z) = ∏ᵢ Zᵢᵢ,
  and the objective is maximizing log(det(Z)).
```
___
*[`geometry/optimization/hyperellipsoid.cc` line 212 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOZl6w7T1Wjxk-l3Ae:-NaOZl6w7T1Wjxk-l3Af:b-mt2wk5) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.cc#L212)):*
> ```cpp
>   const VectorXd b_sol = result.GetSolution(b);
>   // Note: We can use llt() because know that A will be positive definite;
>   // there is a PSD constraint, but we are maximizing the eigenvalues of A and
> ```

As mentioned in my comment in the header file, I don't think this is the best way to handle this case. I think we should compute the `SVD` of the `point` matrix and go from there. This avoids a mysterious extra argument in this function. 

I also am wary of `Zii_ub` as it seems highly solver dependent on what this tolerance should be set to.
___
*[`geometry/optimization/hyperellipsoid.cc` line 213 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOY1e4CLIxVvfsyQpU:-NaOY1e4CLIxVvfsyQpV:b-7m6t46) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.cc#L213)):*
> ```cpp
>   // Note: We can use llt() because know that A will be positive definite;
>   // there is a PSD constraint, but we are maximizing the eigenvalues of A and
>   // the domain of the points is guaranteed to be bounded.
> ```

nit: 



_Suggestion:_
```cpp
convex hull
```
___
*[`geometry/optimization/test/hyperellipsoid_test.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaO_WwK-7Ka4TkjZfFQ:-NaO_WwK-7Ka4TkjZfFR:buxxtmz) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/test/hyperellipsoid_test.cc#L441)):*
> ```cpp
>   const double kTol = 1e-3;
>   Matrix3d A_expected;
>   A_expected << 1, 0, 0, 0, 1.0 / 2.0, 0, 0, 0, 1.0 / 3.0;
> ```

nit: Can you format this to be 3x3 to make it easier to read?

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)* status: 2 unresolved discussions, LGTM missing from assignee AlexandreAmice, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/hyperellipsoid.h` line 105 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOUsYB26BPMcD0Xvu-:-NaRXems6q13fARpJWIP:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.h#L105)):*

nit: typo
</blockquote></details>

Done.
___
*[`geometry/optimization/hyperellipsoid.h` line 111 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOYjZl-oIlLf-V5f0E:-NaQkWQ19cLdTIU2jP4R:b-lngxxa) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.h#L111)):*

This isn't very clear to me. What is `Z` in this case? Is it the dual variable? What is the guidance for setting `Z_ii`?

I understand the problem if `points` are from an affine subspace, then one of the eigenvalues of `A` will go to zero causing the objective to be unbounded. So the point is that we wish to have a finite upper bound on this.

I believe a more appropriate way to test this would be to check if the numerical rank of `point` is lower than the ambient dimension and if so project to the lower dimensional manifold.
</blockquote></details>

Done. I agree with you.  That's much better.  It over doubled the complexity of the implement and tests, but it's much better.
___
*[`geometry/optimization/hyperellipsoid.cc` line 212 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOZl6w7T1Wjxk-l3Ae:-NaRXO5T-UF9U1w5k8LG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.cc#L212)):*

As mentioned in my comment in the header file, I don't think this is the best way to handle this case. I think we should compute the `SVD` of the `point` matrix and go from there. This avoids a mysterious extra argument in this function. 

I also am wary of `Zii_ub` as it seems highly solver dependent on what this tolerance should be set to.
</blockquote></details>

Done.
___
*[`geometry/optimization/hyperellipsoid.cc` line 213 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaOY1e4CLIxVvfsyQpU:-NaRXc-X6UbsJoGZmCKG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/hyperellipsoid.cc#L213)):*

nit: 
</blockquote></details>

Done.
___
*[`geometry/optimization/test/hyperellipsoid_test.cc` line 441 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaO_WwK-7Ka4TkjZfFQ:-NaRXdF25VNvB923ncY4:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/57877c27a741d37b4f4bf61581646c9870a6ed05/geometry/optimization/test/hyperellipsoid_test.cc#L441)):*

nit: Can you format this to be 3x3 to make it easier to read?
</blockquote></details>

Done.

AlexandreAmice (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-:-NaRq16z8MPM207O6kr8:bnfp4nl)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/hyperellipsoid.h` line 109 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaRpb293WgOZ-Enekj5:-NaRpb293WgOZ-Enekj6:b-n7wihw) ([raw file](https://github.com/RobotLocomotion/drake/blob/3411621145d66f16acb8f07ef9bda042bc1adf95/geometry/optimization/hyperellipsoid.h#L109)):*
> ```objc
>   non-zero if their absolute values are strictly greater than `rank_tol`*|max
>   singular value|. The default is 1e-6 to be compatible with common solver
>   tolerances.
> ```

nit: improve clarity.



_Suggestion:_
```objc
  @param rank_tol the singular values of the data matrix will be considered
  non-zero if their absolute values are strictly greater than `rank_tol`*|max
  singular value|. The default is 1e-6 to be compatible with common solver
  tolerances. This is used to detect if the data lies on a lower-dimensional affine space
  than the ambient dimension of the ellipsoid.
```

xuchenhan-tri (commented):
+@xuchenhan-tri for platform review.
:lgtm: with some questions.

> Reviewed 1 of 6 files at r1, 3 of 6 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-:-NaSxF-17M6AUFj_LO3A:bd0ztjj)* status: 5 unresolved discussions

___
*[`geometry/optimization/hyperellipsoid.h` line 108 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT-Nj1A4YN2JjhJ3gI:-NaT-Nj1A4YN2JjhJ3gJ:b-h2tmrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.h#L108)):*
> ```objc
>   @param rank_tol the singular values of the data matrix will be considered
>   non-zero if their absolute values are strictly greater than `rank_tol`*|max
>   singular value|. The default is 1e-6 to be compatible with common solver
> ```

It seems like the actual criterion is that the singular values are deemed as 0 if their absolute value is less than rank_tol.



_Code quote:_
```objc
  @param rank_tol the singular values of the data matrix will be considered
  non-zero if their absolute values are strictly greater than `rank_tol`*|max
  singular value|. The default is 1e-6 to be compatible with common solver
```
___
*[`geometry/optimization/hyperellipsoid.cc` line 186 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaSzPF4-Kf3al-MQ-o3:-NaSzPF4-Kf3al-MQ-o4:bqg9pzb) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L186)):*
> ```cpp
>   const int dim = points.rows();
>   int rank;
>   const int N = points.cols();
> ```

nit, this is borderline GSG violation. Consider



_Suggestion:_
```cpp
const int n = points.cols();
```
___
*[`geometry/optimization/hyperellipsoid.cc` line 194 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT-7K11g0M6vF8zTRa:-NaT-7K11g0M6vF8zTRb:bpkuajl) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L194)):*
> ```cpp
>     auto svd = (points.colwise() - mean).bdcSvd(Eigen::ComputeThinU);
>     // Eigen's SVD rank never returns zero.
>     if (svd.singularValues()[0] < rank_tol) {
> ```

BTW, it'd be helpful to remind readers that Eigen's convention is to sort singular values in decreasing order.



_Code quote:_
```cpp
if (svd.singularValues()[0] < rank_tol) {
```
___
*[`geometry/optimization/hyperellipsoid.cc` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT0s2UA4f3Hzs_Lzyi:-NaT0s2UA4f3Hzs_Lzyj:b-n3d9oy) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L224)):*
> ```cpp
>       // |Ax + b|₂ <= 1
>       v.tail(rank) = A * points.col(i) + b;
>     }
> ```

I'm confused about what we are trying to achieve here. It appears that we are just thrashing the values in `v.tail(rank)`? Should we be incrementing instead? It'd be nice to clearly document what v is supposed to be.



_Code quote:_
```cpp
  for (int i = 0; i < N; ++i) {
    if (U) {  // rank < dim
      // |AUᵀ(x-mean) + b|₂ <= 1
      v.tail(rank) = A * U->transpose() * (points.col(i) - mean) + b;
    } else {  // rank == dim
      // |Ax + b|₂ <= 1
      v.tail(rank) = A * points.col(i) + b;
    }
```
___
*[`geometry/optimization/hyperellipsoid.cc` line 237 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT1m5t3LG0noD98D9e:-NaT1m5t3LG0noD98D9f:b-cxc342) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L237)):*
> ```cpp
>   }
>   log()->info("Solver {}", result.get_solver_id().name());
>   DRAKE_THROW_UNLESS(result.is_success());
> ```

BTW this should be a DRAKE_DEMAND. Failure indicates a bug as we already throw above.



_Code quote:_
```cpp
DRAKE_THROW_UNLESS(result.is_success());
```

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)* status: 2 unresolved discussions

___
*[`geometry/optimization/hyperellipsoid.h` line 108 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT-Nj1A4YN2JjhJ3gI:-NaTTSofD6ORciTf_J_B:b-g399ln) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.h#L108)):*

It seems like the actual criterion is that the singular values are deemed as 0 if their absolute value is less than rank_tol.
</blockquote></details>

Done. I think you're referring to the additional check i do on the very first singular value?  i've added that detail to the pre statement below about the rank of the data matrix, to be more precise.

but the setThreshold + rank argument in the eigen svd is doing exactly the check i've documented here.

the singular values are always positive, so I zapped the absolute value verbage`.
___
*[`geometry/optimization/hyperellipsoid.cc` line 186 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaSzPF4-Kf3al-MQ-o3:-NaTUn_u6seHILao2MvX:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L186)):*

nit, this is borderline GSG violation. Consider
</blockquote></details>

Done.
___
*[`geometry/optimization/hyperellipsoid.cc` line 194 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT-7K11g0M6vF8zTRa:-NaTUwfm8yvCjia5_lfk:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L194)):*

BTW, it'd be helpful to remind readers that Eigen's convention is to sort singular values in decreasing order.
</blockquote></details>

Done.
___
*[`geometry/optimization/hyperellipsoid.cc` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT0s2UA4f3Hzs_Lzyi:-NaTV5sP-O4mumtXFagd:b-a5bwlb) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L224)):*

I'm confused about what we are trying to achieve here. It appears that we are just thrashing the values in `v.tail(rank)`? Should we be incrementing instead? It'd be nice to clearly document what v is supposed to be.
</blockquote></details>

Done. v is the argument being passed into the AddLorentzConeConstraint just below... and is documented just above??  I've tried to make it more explicit.
___
*[`geometry/optimization/hyperellipsoid.cc` line 237 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT1m5t3LG0noD98D9e:-NaTV_j2Cm8WGQCc-fZL:b-xlp0t7) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L237)):*

BTW this should be a DRAKE_DEMAND. Failure indicates a bug as we already throw above.
</blockquote></details>

Done.  Oops.  That was just a git flub.  Those lines were only my debugging.

AlexandreAmice (commented):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)* status: 2 unresolved discussions

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19857)* status: :shipit: complete! all discussions resolved, LGTM from assignees AlexandreAmice,xuchenhan-tri(platform)

___
*[`geometry/optimization/hyperellipsoid.cc` line 224 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19857#-NaT0s2UA4f3Hzs_Lzyi:-NaclEAN-zUxRFRAlFIj:b5ybx5z) ([raw file](https://github.com/RobotLocomotion/drake/blob/fbb0e10f580f73c23c9ff55ac6d11678bfede166/geometry/optimization/hyperellipsoid.cc#L224)):*

Done. v is the argument being passed into the AddLorentzConeConstraint just below... and is documented just above??  I've tried to make it more explicit.
</blockquote></details>

I see. I missed the fact that a new constraint is added in each iteration. This is in fact obvious. Apologies.

PR #19863, opened 2023-07-28
CameraConfig can use RenderEngineParams

SeanCurtis-TRI:
Previously, the nature of the render engine was given by a single string: naming the render engine class to use. It makes the `RenderEngine` impossible to configure.

This extends the `renderer_class` field to also take instances of the various `RenderEngine*Params` structs. Now the engines can be fully specified in yaml.

This also extends the set of renderer classes to include GltfCLient.

Updated the documentation on `CameraConfig::background` to clarify its scope in the new world.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)

### DIFF
--- a/geometry/render_gl/BUILD.bazel
+++ b/geometry/render_gl/BUILD.bazel
@@ -96,6 +96,7 @@ drake_cc_library(
     hdrs = [
         "factory.h",
     ],
+    visibility = ["//visibility:public"],
     interface_deps = [
         ":render_engine_gl_params",
         "//geometry/render:render_engine",
@@ -132,6 +133,7 @@ drake_cc_library(
     hdrs = [
         "render_engine_gl_params.h",
     ],
+    visibility = ["//visibility:public"],
     deps = [
         "//common:name_value",
         "//geometry:rgba",
--- a/systems/sensors/BUILD.bazel
+++ b/systems/sensors/BUILD.bazel
@@ -83,6 +83,9 @@ drake_cc_library(
         "//common/schema:transform",
         "//geometry:rgba",
         "//geometry/render:render_camera",
+        "//geometry/render_gl:render_engine_gl_params",
+        "//geometry/render_gltf_client:render_engine_gltf_client_params",
+        "//geometry/render_vtk:render_engine_vtk_params",
     ],
 )
 
@@ -104,6 +107,7 @@ drake_cc_library(
         ":rgbd_sensor_async",
         ":sim_rgbd_sensor",
         "//geometry/render_gl",
+        "//geometry/render_gltf_client",
         "//geometry/render_vtk",
         "//math:geometric_transform",
         "//multibody/parsing:scoped_names",
@@ -358,13 +362,14 @@ drake_cc_googletest(
         "//common/test_utilities:eigen_matrix_compare",
         "//common/test_utilities:expect_throws_message",
         "//common/yaml:yaml_io",
-        "//geometry/render_gl",
+        "//geometry/render_vtk",
         "@fmt",
     ],
 )
 
 drake_cc_googletest(
     name = "camera_config_functions_test",
+    allow_network = ["render_gltf_client"],
     tags = vtk_test_tags(),
     deps = [
         ":camera_config_functions",
@@ -374,6 +379,8 @@ drake_cc_googletest(
         "//common/test_utilities:expect_throws_message",
         "//common/yaml:yaml_io",
         "//geometry/render_gl",
+        "//geometry/render_gltf_client",
+        "//geometry/render_vtk",
         "//lcm:drake_lcm",
         "//systems/lcm:lcm_publisher_system",
     ],
--- a/systems/sensors/camera_config.cc
+++ b/systems/sensors/camera_config.cc
@@ -130,13 +130,20 @@ void CameraConfig::ValidateOrThrow() const {
       },
       focal);
 
-  if (!renderer_class.empty() && !(renderer_class == "RenderEngineVtk" ||
-                                   renderer_class == "RenderEngineGl")) {
-    throw std::logic_error(fmt::format(
-        "Invalid camera configuration; the given renderer_class value '{}' "
-        "must be empty (to use the default) or be one of 'RenderEngineVtk' or "
-        "'RenderEngineGl'.",
-        renderer_class));
+  // We don't worry about the other variant alternatives; if we're here, we've
+  // constructed a set of parameters, and we'll defer to the RenderEngine to
+  // determine if the values are valid.
+  if (std::holds_alternative<std::string>(renderer_class)) {
+    const auto& class_name = std::get<std::string>(renderer_class);
+    if (!class_name.empty() &&
+        !(class_name == "RenderEngineVtk" || class_name == "RenderEngineGl" ||
+          class_name == "RenderEngineGltfClient")) {
+      throw std::logic_error(fmt::format(
+          "Invalid camera configuration; the given renderer_class value '{}' "
+          "must be empty (to use the default) or be one of 'RenderEngineVtk', "
+          "'RenderEngineGl', or 'RenderEngineGltfClient'.",
+          class_name));
+    }
   }
 
   // This throws for us if we have bad bad numerical camera values (or an empty
--- a/systems/sensors/camera_config.h
+++ b/systems/sensors/camera_config.h
@@ -9,6 +9,9 @@
 #include "drake/common/name_value.h"
 #include "drake/common/schema/transform.h"
 #include "drake/geometry/render/render_camera.h"
+#include "drake/geometry/render_gl/render_engine_gl_params.h"
+#include "drake/geometry/render_gltf_client/render_engine_gltf_client_params.h"
+#include "drake/geometry/render_vtk/render_engine_vtk_params.h"
 #include "drake/geometry/rgba.h"
 
 namespace drake {
@@ -248,21 +251,116 @@ struct CameraConfig {
    @pre `renderer_name` is not empty. */
   std::string renderer_name{"default"};
 
-  /** The choice of render engine implementation to use. The value should be
-   one of empty, '"RenderEngineVtk"', or '"RenderEngineGl"'. An `empty` string
-   signals, in essence, "don't care". If a render engine with that name has
-   already been configured, the camera will use it (regardless of type). If
-   the name doesn't already exist, the default, slower, more portable, and more
-   robust RenderEngineVtk will be instantiated. RenderEngineGl can be selected
-   if you are on Ubuntu and need the improved performance (at the *possible*
-   cost of lesser image fidelity).
+  /** The choice of render engine implementation to use. It can be specified
+   simply by providing a `string` containing the class _name_ of the
+   RenderEngine to use (i.e., `"RenderEngineVtk"`, `"RenderEngineGl"`, or
+   `"RenderEngineGltfClient"`). Or, it can be specified by providing parameters
+   for one of those engines: RenderEngineVtkParams, RenderEngineGlParams, or
+   RenderEngineGltfClientParams.
 
-   @pre `renderer_class` is empty or one of '"RenderEngineVtk"' or
-        '"RenderEngineGl"'.
-   @pre For two cameras with the same `renderer_name` value, they must also
-        specify the same `renderer_class` (either implicitly or explicitly).
+   If a `string` containing the render engine class _name_ is provided, the
+   engine instantiated will use the default parameters -- equivalent to passing
+   the set of default parameters.
+
+   It is possible for multiple cameras to reference the same `renderer_name`
+   but configure the renderer differently. This would be a configuration error.
+   The following rules will help prevent problematic configurations:
+
+     - Multiple cameras can reference the same value for `renderer_name`.
+     - If multiple cameras reference the same value for `renderer_name`, only
+       the *first* can use the engine parameters to specify it. Attempting to
+       do so with a later camera will produce an error.
+       - The later cameras can use engine class _name_.
+       - If a later camera names a *different* engine class, that will result
+         in an error.
+
+   In YAML, it can be a bit trickier. Depending on how a collection of cameras
+   is articulated, the concept of "first" may be unclear. In
+   `examples/hardware_sim/scenario.h` the collection is a map. So, the camera
+   configurations are not necessarily processed in the order they appear in the
+   YAML file. Instead, the processing order depends on the mnemonic camera key.
+   So, be aware, if you use a similar mapping, you may have to massage the key
+   names to achieve the requisite processing order. Alternatively, a vector of
+   %CameraConfig in your own scenario file, would guarantee that processing
+   order is the same as file order.
+
+   We intend to relax these restrictions in time, allowing equivalent, redundant
+   specifications of a render engine (and throwing only on inconsistent
+   specifications).
+
+   Passing the empty string is equivalent to saying, "I don't care". If a render
+   engine with that name has already been configured, the camera will use it
+   (regardless of type). If the name doesn't already exist, the default, slower,
+   more portable, and more robust RenderEngineVtk will be instantiated.
+   RenderEngineGl can be selected if you are on Ubuntu and need the improved
+   performance (at the *possible* cost of lesser image fidelity). In YAML,
+   omitting `renderer_class` from the camera specification is equivalent to
+   "passing the empty string".
+
+   <h4>Configuring in YAML</h4>
+
+   It isn't always obvious what the proper spelling in YAML is. The following
+   examples illustrate how the `renderer_class` field can be defined in YAML
+   files.
+
+   1. Use the RenderEngineVtk with all default parameters - providing the class
+      name as a string.
+   @code{yaml}
+   renderer_class: RenderEngineVtk
+   @endcode
+
+   2. Use the RenderEngineVtk with all default parameters - providing the
+      engine parameters with default values.
+   @code{yaml}
+   renderer_class: !RenderEngineVtkParams {}
+   @endcode
+
+   3. Use the RenderEngineVtk with a customized clear color (black).
+   @code{yaml}
+   renderer_class: !RenderEngineVtkParams
+     default_clear_color: [0, 0, 0]
+   @endcode
+
+   4. Use the RenderEngineGl with a customized clear color (black).
+   @code{yaml}
+   renderer_class: !RenderEngineGlParams
+     default_clear_color:
+       rgba: [0, 0, 0, 1]
+   @endcode
+
+   5. Use the RenderEngineGltfClient with fully specified properties.
+   @code{yaml}
+   renderer_class: !RenderEngineGltfClientParams
+     base_url: http://10.10.10.1
+     render_endpoint: server
+     verbose: true
+     cleanup: false
+   @endcode
+
+   Things to note:
+
+     - When providing the _parameters_ for the engine, the declaration must
+       begin with `!` to announce it as a type (examples 2, 3, and 4).
+     - A defaulted set of parameters must have a trailing `{}` (example 2).
+     - Two engine parameter sets may have the same semantic parameter but spell
+       it differently (`default_clear_color` in examples 3 and 4).
+
+   <!-- TODO(SeanCurtis-TRI): Figure out why this doesn't work.
+   ?. Explicitly default to Drake's default render engine.
+   @code{yaml}
+   renderer_class: ""
+   @endcode
+   -->
+
+   @pre `renderer_class` is a string and either empty or one of
+        `"RenderEngineVtk"`, `"RenderEngineGl"`, or `"RenderEngineGltfClient"`,
+        or, it is a parameter type of one of Drake's RenderEngine
+        implementations.
    @sa drake::geometry::SceneGraph::GetRendererTypeName(). */
-  std::string renderer_class;
+  std::variant<std::string, geometry::RenderEngineVtkParams,
+               geometry::RenderEngineGlParams,
+               geometry::RenderEngineGltfClientParams>
+      renderer_class{""};
 
   /** The "background" color. This is the color drawn where there are no objects
    visible. Its default value matches the default value for
@@ -270,10 +368,9 @@ struct CameraConfig {
    documentation for geometry::Rgba::Serialize for how to define this
    value in YAML.
 
-   N.B. If two different cameras are configured to use the same renderer (having
-   identical values for `renderer_name`) but *different* values for
-   `background`, the render engine instance will be configured with one of the
-   set of values. Which one is undefined. */
+   This value is used only if the `render_class` specifies either
+   `"RenderEngineVtk"` or `"RenderEngineGl"` by _name_ (RenderEngineGltfClient
+   doesn't have a configurable background color.) */
   geometry::Rgba background{204 / 255.0, 229 / 255.0, 255 / 255.0, 1.0};
 
   //@}
--- a/systems/sensors/camera_config_functions.cc
+++ b/systems/sensors/camera_config_functions.cc
@@ -13,6 +13,7 @@
 #include "drake/common/nice_type_name.h"
 #include "drake/geometry/render/render_camera.h"
 #include "drake/geometry/render_gl/factory.h"
+#include "drake/geometry/render_gltf_client/factory.h"
 #include "drake/geometry/render_vtk/factory.h"
 #include "drake/math/rigid_transform.h"
 #include "drake/multibody/parsing/scoped_names.h"
@@ -29,8 +30,10 @@ using drake::lcm::DrakeLcmInterface;
 using drake::systems::lcm::LcmBuses;
 using Eigen::Vector3d;
 using geometry::MakeRenderEngineGl;
+using geometry::MakeRenderEngineGltfClient;
 using geometry::MakeRenderEngineVtk;
 using geometry::RenderEngineGlParams;
+using geometry::RenderEngineGltfClientParams;
 using geometry::RenderEngineVtkParams;
 using geometry::SceneGraph;
 using geometry::render::ColorRenderCamera;
@@ -45,21 +48,82 @@ using multibody::parsing::GetScopedFrameByName;
 
 namespace {
 
+// Boilerplate for std::visit.
+template <class... Ts>
+struct overloaded : Ts... {
+  using Ts::operator()...;
+};
+template <class... Ts>
+overloaded(Ts...) -> overloaded<Ts...>;
+
+// Given a valid string containing a supported RenderEngine class name, returns
+// the full name as would be returned by SceneGraph::GetRendererTypeName().
+// @pre We already know that `class_name` is a "supported" value.
+const std::string& LookupEngineType(const std::string& class_name) {
+  using Dict = std::map<std::string, std::string>;
+  static const never_destroyed<Dict> type_lookup(
+      std::initializer_list<Dict::value_type>{
+          {"",  // Empty string defaults to RenderEngineVtk.
+           "drake::geometry::render_vtk::internal::RenderEngineVtk"},
+          {"RenderEngineVtk",
+           "drake::geometry::render_vtk::internal::RenderEngineVtk"},
+          {"RenderEngineGl",
+           "drake::geometry::render_gl::internal::RenderEngineGl"},
+          {"RenderEngineGltfClient",
+           "drake::geometry::render_gltf_client::internal::"
+           "RenderEngineGltfClient"}});
+  return type_lookup.access().at(class_name);
+}
+
+template <typename ParamsType>
+const char* GetEngineName(const ParamsType&) {
+  if constexpr (std::is_same_v<ParamsType, RenderEngineVtkParams>) {
+    return "RenderEngineVtk";
+  } else if constexpr (std::is_same_v<ParamsType, RenderEngineGlParams>) {
+    return "RenderEngineGl";
+    // NOLINTNEXTLINE(readability/braces) The line break confuses the linter.
+  } else if constexpr (std::is_same_v<ParamsType,
+                                      RenderEngineGltfClientParams>) {
+    return "RenderEngineGltfClient";
+  }
+}
+
+void MakeEngineByClassName(const std::string& class_name,
+                           const CameraConfig& config,
+                           SceneGraph<double>* scene_graph) {
+  if (class_name == "RenderEngineGl") {
+    if (!geometry::kHasRenderEngineGl) {
+      throw std::logic_error(
+          "Invalid camera configuration; renderer_class = 'RenderEngineGl' "
+          "is not supported in current build.");
+    }
+    RenderEngineGlParams params{.default_clear_color = config.background};
+    scene_graph->AddRenderer(config.renderer_name, MakeRenderEngineGl(params));
+    return;
+  } else if (class_name == "RenderEngineGltfClient") {
+    scene_graph->AddRenderer(config.renderer_name,
+                             MakeRenderEngineGltfClient({}));
+    return;
+  }
+  // Note: if we add *other* supported render engine implementations, add the
+  // logic for detecting and instantiating those types here.
+
+  // Fall through to the default render engine type (name is either empty or
+  // the only remaining possible value: "RenderEngineVtk").
+  DRAKE_DEMAND(class_name.empty() || class_name == "RenderEngineVtk");
+  RenderEngineVtkParams params;
+  const geometry::Rgba& rgba = config.background;
+  params.default_clear_color = Vector3d{rgba.r(), rgba.g(), rgba.b()};
+  scene_graph->AddRenderer(config.renderer_name, MakeRenderEngineVtk(params));
+}
+
 // Validates the render engine specification in `config`. If the specification
 // is valid, upon return, the specified RenderEngine is defined in `scene_graph`
 // (this may require creating the RenderEngine instance). Throws if the
 // specification is invalid.
-// @pre we already know that config.renderer_class has a "supported" value.
+// @pre If config.renderer_class *names* a class, it is a "supported" value.
 void ValidateEngineAndMaybeAdd(const CameraConfig& config,
                                SceneGraph<double>* scene_graph) {
-  using Dict = std::map<std::string, std::string>;
-  static const never_destroyed<Dict> type_lookup(
-      std::initializer_list<Dict::value_type>{
-          {"RenderEngineVtk",
-           "drake::geometry::render_vtk::internal::RenderEngineVtk"},
-          {"RenderEngineGl",
-           "drake::geometry::render_gl::internal::RenderEngineGl"}});
-
   DRAKE_DEMAND(scene_graph != nullptr);
 
   // Querying for the type_name of the named renderer will simultaneously tell
@@ -70,40 +134,56 @@ void ValidateEngineAndMaybeAdd(const CameraConfig& config,
 
   // Non-empty type name says that it already exists.
   bool already_exists = !type_name.empty();
-  if (already_exists && !config.renderer_class.empty()) {
-    if (type_lookup.access().at(config.renderer_class) != type_name) {
-      throw std::logic_error(
-          fmt::format("Invalid camera configuration; requested "
-                      "renderer_name = '{}' and renderer_class = '{}'. The "
-                      "name is already used with a different type: {}.",
-                      config.renderer_name, config.renderer_class, type_name));
-    }
+
+  if (already_exists) {
+    std::visit(
+        overloaded{
+            [&type_name, &config](const std::string& class_name) -> void {
+              if (LookupEngineType(class_name) != type_name) {
+                throw std::logic_error(fmt::format(
+                    "Invalid camera configuration; requested renderer_name "
+                    "= '{}' and renderer_class = '{}'. The name is already "
+                    "used with a different type: {}.",
+                    config.renderer_name, class_name, type_name));
+              }
+            },
+            [&config](auto&&) -> void {
+              throw std::logic_error(
+                  fmt::format("Invalid camera configuration; requested "
+                              "renderer_name = '{}' with renderer parameters, "
+                              "but the named renderer already exists. Only the "
+                              "first instance of the named renderer can use "
+                              "parameters.",
+                              config.renderer_name));
+            }},
+        config.renderer_class);
   }
 
   if (already_exists) return;
 
-  // Now we know we need to add one. Confirm we can add the specified class.
-  if (config.renderer_class == "RenderEngineGl") {
-    if (!geometry::kHasRenderEngineGl) {
-      throw std::logic_error(
-          "Invalid camera configuration; renderer_class = 'RenderEngineGl' "
-          "is not supported in current build.");
-    }
-    RenderEngineGlParams params{.default_clear_color = config.background};
-    scene_graph->AddRenderer(config.renderer_name, MakeRenderEngineGl(params));
-    return;
-  }
-  // Note: if we add *other* supported render engine implementations, add the
-  // logic for detecting and instantiating those types here.
-
-  // Fall through to the default render engine type (name is either empty or
-  // the only remaining possible value: "RenderEngineVtk").
-  DRAKE_DEMAND(config.renderer_class.empty() ||
-               config.renderer_class == "RenderEngineVtk");
-  RenderEngineVtkParams params;
-  const geometry::Rgba& rgba = config.background;
-  params.default_clear_color = Vector3d{rgba.r(), rgba.g(), rgba.b()};
-  scene_graph->AddRenderer(config.renderer_name, MakeRenderEngineVtk(params));
+  std::visit(
+      overloaded{
+          [&config, scene_graph](const std::string& class_name) {
+            MakeEngineByClassName(class_name, config, scene_graph);
+          },
+          [&config, scene_graph](const RenderEngineVtkParams& params) {
+            scene_graph->AddRenderer(config.renderer_name,
+                                     MakeRenderEngineVtk(params));
+          },
+          [&config, scene_graph](const RenderEngineGlParams& params) {
+            if (!geometry::kHasRenderEngineGl) {
+              throw std::logic_error(
+                  "Invalid camera configuration; renderer_class = "
+                  "'RenderEngineGl' is not supported in current build.");
+            }
+            scene_graph->AddRenderer(config.renderer_name,
+                                     MakeRenderEngineGl(params));
+          },
+          [&config, scene_graph](const RenderEngineGltfClientParams& params) {
+            scene_graph->AddRenderer(config.renderer_name,
+                                     MakeRenderEngineGltfClient(params));
+          }},
+      config.renderer_class);
 }
 
 }  // namespace
--- a/systems/sensors/test/camera_config_functions_test.cc
+++ b/systems/sensors/test/camera_config_functions_test.cc
@@ -9,6 +9,8 @@
 #include "drake/common/test_utilities/expect_throws_message.h"
 #include "drake/common/yaml/yaml_io.h"
 #include "drake/geometry/render_gl/factory.h"
+#include "drake/geometry/render_gltf_client/factory.h"
+#include "drake/geometry/render_vtk/factory.h"
 #include "drake/lcm/drake_lcm.h"
 #include "drake/systems/lcm/lcm_publisher_system.h"
 #include "drake/systems/sensors/image_to_lcm_image_array_t.h"
@@ -21,6 +23,9 @@ namespace sensors {
 namespace {
 
 using drake::geometry::FrameId;
+using drake::geometry::RenderEngineGlParams;
+using drake::geometry::RenderEngineGltfClientParams;
+using drake::geometry::RenderEngineVtkParams;
 using drake::geometry::Rgba;
 using drake::geometry::SceneGraph;
 using drake::geometry::render::ColorRenderCamera;
@@ -57,7 +62,7 @@ CameraConfig MakeConfig() {
                       .X_BC = Transform{RigidTransformd{Vector3d::UnitX()}},
                       .X_BD = Transform{RigidTransformd{Vector3d::UnitX()}},
                       .renderer_name = "test_renderer",
-                      .renderer_class = "RenderEngineVtk",
+                      .renderer_class = "RenderEngineGltfClient",
                       .background = Rgba(0.25, 0.5, 0.75),
                       .name = "test_camera",
                       .fps = 17,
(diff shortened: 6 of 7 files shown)


SeanCurtis-TRI (commented):
cc @EricCousineau-TRI and @jwnimmer-tri 

Here's my quick and dirty stab at changing `CameraConfig` as I alluded to in #19837.

I'd love to hear your thoughts.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NaPvyys44sYXPu35LYs:b-mpj8qo)* status: needs platform reviewer assigned, needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
+@jwnimmer-tri for my dashboard.

For (1), I've opened new discussions below.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NaSqoM7A462cg9omyMi:bskj7hr)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSpggB4aAoNlK-9cSV:-NaSpggC3Ea_wBq3QHI-:bij8jyy) (no related file):*
> ... `background` [is] now ignored and should be removed. TODO: simply cut? Deprecate some way? ...

See below re: keeping `renderer_class` around.

Once we do that, we can change the documentation of `background` to state that its only used when specifying the `renderer_class` by name; in case the specific Params are specified, they will trump the more generic `background`.

___
___
*[`geometry/render/render_engine.cc` line 26 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSr6f475Xw9AVzE8ZZ:-NaSr6f5CvIGaHstUDh-:binj8zq) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/geometry/render/render_engine.cc#L26)):*
> ```cpp
>   if (label_name == "unspecified") {
>     return RenderLabel::kUnspecified;
>   } else if (label_name == "don't_care") {
> ```

We need to nix the apostrophe here.

I think its important to keep enum transmogrification boring, so that we can automate it later.  If the enum constant is kFooBar, then the string encoding should be foo_bar.
___
*[`systems/sensors/camera_config.h` line 265 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSqN0-7f398kxpOdfw:-NaSqN004QSxCuE_fLAp:b8dk94y) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L265)):*
> ```objc
>         '"RenderEngineGl"'.
>    @pre For two cameras with the same `renderer_name` value, they must also
>         specify the same `renderer_class` (either implicitly or explicitly).
> ```

This raises the question of what to do when two cameras with the same name both specify some `RenderEngineFooParams` specifically.

Previously things like `background` were documented as "an arbitrary camera's config wins the engine-params battle, which one is unspecified".

Possibly we need to fail-fast in case any of the params are different at all?
___
*[`systems/sensors/camera_config.h` line 272 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSoUc3EFlFpXkpntKm:-NaSoUc4E2bWMqFMlaBq:bu4h8sp) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L272)):*
I think it would make the most sense to use the same attribute name for the new choices:

```c++
  std::variant<std::string,
               geometry::RenderEngineVtkParams,
               geometry::RenderEngineGlParams,
               geometry::RenderEngineGltfClientParams>
      renderer_class;
```

That allows for the current syntax in case they want default params:

```yaml
MyCamera:
  renderer_class: RenderEngineGl
```

Or a more similar, enhanced syntax in case they want to customize:

```yaml
MyCamera:
  renderer_class: !RenderEngineGlParams
    default_diffuse: ...
```

And removes the question of:

> ... `renderer_class` [is] now ignored and should be removed. TODO: simply cut? Deprecate some way? ...

jwnimmer-tri (commented):
I should have also mentioned that I'm fine with the current design of your question (2).  Though I might need to help add some deprecation / transition affordances.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NaTHEPcASb-Dj_E3SbY:bo9ki54)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

___
*[`geometry/render/render_engine.cc` line 26 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSr6f475Xw9AVzE8ZZ:-NacyXzt3TrIqtFQCMSr:bw7v10p) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/geometry/render/render_engine.cc#L26)):*

We need to nix the apostrophe here.

I think its important to keep enum transmogrification boring, so that we can automate it later.  If the enum constant is kFooBar, then the string encoding should be foo_bar.
</blockquote></details>

Yeah...I put that apostrophe in in the very last second. I didn't like it, so it's good to hear it strikes you the same.
___
*[`systems/sensors/camera_config.h` line 272 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSoUc3EFlFpXkpntKm:-NacylH35O8lXIGIRSu1:b-9tsov8) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L272)):*

I think it would make the most sense to use the same attribute name for the new choices:

```c++
  std::variant<std::string,
               geometry::RenderEngineVtkParams,
               geometry::RenderEngineGlParams,
               geometry::RenderEngineGltfClientParams>
      renderer_class;
```

That allows for the current syntax in case they want default params:

```yaml
MyCamera:
  renderer_class: RenderEngineGl
```

Or a more similar, enhanced syntax in case they want to customize:

```yaml
MyCamera:
  renderer_class: !RenderEngineGlParams
    default_diffuse: ...
```

And removes the question of:

> ... `renderer_class` [is] now ignored and should be removed. TODO: simply cut? Deprecate some way? ...
</blockquote></details>

This seems an elegant way forward for transitioning. I'll add that in the next version.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSpggB4aAoNlK-9cSV:-Naf6NjD85gtxw4adO00:b-q8zdrt) (no related file):*

> ... `background` [is] now ignored and should be removed. TODO: simply cut? Deprecate some way? ...

See below re: keeping `renderer_class` around.

Once we do that, we can change the documentation of `background` to state that its only used when specifying the `renderer_class` by name; in case the specific Params are specified, they will trump the more generic `background`.
</blockquote></details>

I'm marking this topic as resolved, as I agree and have implemented it.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Naf746a6cNY0iHS99Vj:-Naf746a6cNY0iHS99Vk:bvbde3d) (no related file):*
re: deprecation

We've essentially got two classes of changes that are candidates for deprecation:

1. The C++ constructor for `RenderEngine` went from taking a `RenderLabel` to taking a `string`. The python bindings omitted this. As breaking changes go, this seems *incredibly* harmless. I can't imagine there are a lot of people out there implementing `RenderEngine` implementations.
2. The three `RenderEngine*Params` classes have `default_label` changed from `RenderLabel` to `string`. This effects both python and C++.
    - It seems to me that if no one is really configuring label, it might be worth just breaking it.
    - Alternatively, turning it into a variant where we consume label *and* string and print a one-time warning when the label is used, might be a reasonable way to deprecate.

THoughts?

___
___
*[`systems/sensors/camera_config.h` line 265 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSqN0-7f398kxpOdfw:-NaeK8Kt653CV0NJijKs:boui5qo) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L265)):*

This raises the question of what to do when two cameras with the same name both specify some `RenderEngineFooParams` specifically.

Previously things like `background` were documented as "an arbitrary camera's config wins the engine-params battle, which one is unspecified".

Possibly we need to fail-fast in case any of the params are different at all?
</blockquote></details>

In the current version, I've merely extended the "in case of disagreement, a coin toss defines the winner" resolution from background to `renderer_class`.

Testing the incoming parameters against any previous parameters that might have come before can be problematic. We'd have to get access to the render engine itself and ask it about its construction parameters. The parameters would have to implement a comparison operator, etc. I started down this path and wasn't feeling particularly pleased with it.

Do you have an alternate idea for supporting the "either they match, or we throw"?

SeanCurtis-TRI (commented):
This version doesn't contain any deprecation efforts (I'm holding my aborted efforts in abeyance, pending the conversation below).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Naf7pkyE0EEVzEOyxF7:b-bj126k)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
I believe this isn't time-critical, so I'll plan to circle back here in a day or two and offer some deprecation-related code fixes.

SeanCurtis-TRI (commented):
Agreed. I'll put a more decisive stamp on the subject.

+(priority: low) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Nai6tAz0nGHbg0_12lm:b-904vs5)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
FYI So you know my agenda here. I'd like to get this in for the next release, so that `RenderEngineGltfClient` can be configured in yaml. I'm also going to investigate having more control over what kind of `RgbdSensor` gets applied (e.g., continuous, discrete, or async). Right now we've got plenty of time.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NaiPqbd4BDZp0GSV32P:b-507c9t)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
> ... control over what kind of RgbdSensor gets applied (e.g., continuous, discrete, or async).

My plan for #19437 is to add `output_delay` to CameraConfig.  When a non-zero `output_delay` is specified, it would use the async camera automatically.

I can work to front-load that to finish by the release.

SeanCurtis-TRI (commented):
I'd hope for casting a wider net.

1. For the async camera, we should also include `capture_offset` (so we can specify staggered async cameras in yaml).
2. I presume we'll never actually instantiate a `RgbdSensorDiscrete`, because that's just an async camera with a zero delay, yes?
    - That implies that the difference isn't whether the value is positive, but whether it's non-negative.
    - It would certainly be nice to be able to specify a discrete camera with zero delay (regardless of how it is implemented under the hood).
3. It would be nice to add a `label` signal to the `CameraConfig` as well.

Also, I can imagine it would be nice to be able to configure an image writer or (eventually) a video writer in yaml. But that is far down the road.

If you want to push this off of your plate and onto mine, I'm happy to grab it. I definitely want it in in time for 1.20.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NaigVYK687qlapXlDbA:bm3jdbf)* status: 5 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri:
Let's discuss in #19898.

jwnimmer-tri:
> Let's discuss in https://github.com/RobotLocomotion/drake/pull/19898.

I suppose for these less-critical features ...

- label image opt-in (`bool label`, akin to `bool depth`)
- image writer
- video writer

... maybe you can open new feature request ticket(s) for ones you/we actually need?  I wasn't planning on handling any of those in #19898 yet.

SeanCurtis-TRI:
See #19901.

jwnimmer-tri:
Back to the original thrust of the PR here...

There are two main changes happening:
(1) Adjust each `MyRenderEngineParams` to support YAML serialization, so that we can specify engine configuration in YAML.
(2) Enhance `CameraConfig` to allow specifying either just the renderer class, or else its full Params.

The deprecation / breaking change difficulties come with step (1).  Step (2) is mostly not disruptive.

Also, the MVP only strictly requires step (1).  So long as the `MyRenderEngineParams` can be serialized, we can use the `renderer_name` to link the `CameraConfig` to an engine.

For those reasons, I'd like to treat the steps as separate pull requests.

I'll work today to propose some changes for what step (1) should look like.  After some or all of those land, we can do step (2).

Note that we can also do step 2 incrementally (one engine at a time, adding more variant allowed types as we go).  I anticipate that marking the glTF engine deprecations will be easiest, so we might be able to offer that particular engine quickly.  I'll edit this post with more links as I make progress.

jwnimmer-tri (commented):
> Reviewed 1 of 18 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Naf746a6cNY0iHS99Vj:-NagW6VU6ehJ8JCymvSd:b-el2hpt) (no related file):*
> 1. The C++ constructor for `RenderEngine` went from taking a `RenderLabel` to taking a `string`. The python bindings omitted this. As breaking changes go, this seems _incredibly_ harmless. I can't imagine there are a lot of people out there implementing `RenderEngine` implementations.

The default choice is always to deprecate.  Only when the deprecation effort is large do we elect to consider breaking changes (subject to considering the impact):

> On very rare occasions it is impractical, too expensive, or logically unsound to meet the full three month deprecation window. In these cases the change will be announced in the “Breaking changes” section of the release notes.

-- https://drake.mit.edu/stable.html

In other words, considering the fallout up-front is premature.  I am happy to push the deprecation patch here myself, once this is a bit more polished.

___
___
*[`systems/sensors/BUILD.bazel` line 88 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NagYFjr7Z_P5nUY-wOu:-NagYFjr7Z_P5nUY-wOv:b-eebh88) ([raw file](https://github.com/RobotLocomotion/drake/blob/59629eb9d11d81fd4b103429304a6972eb69aea1/systems/sensors/BUILD.bazel#L88)):*
> ```Starlark
>         "//geometry/render_gl",
>         "//geometry/render_gltf_client",
>         "//geometry/render_vtk",
> ```

nit Let's limit these to the `deps` to list the only the params structs, not the full engines.  The engines have a lot of extra transitive baggage for what should be small, simple structs.
___
*[`systems/sensors/camera_config.h` line 265 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSqN0-7f398kxpOdfw:-NammyPHEnz-kUKgbs4X:b-jkcp98) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L265)):*

In the current version, I've merely extended the "in case of disagreement, a coin toss defines the winner" resolution from background to `renderer_class`.

Testing the incoming parameters against any previous parameters that might have come before can be problematic. We'd have to get access to the render engine itself and ask it about its construction parameters. The parameters would have to implement a comparison operator, etc. I started down this path and wasn't feeling particularly pleased with it.

Do you have an alternate idea for supporting the "either they match, or we throw"?
</blockquote></details>

The cop-out to get this landed quickly would be to say that if the user specifies a `FooParms` in the `renderer_class` variant, then the `renderer_name` must not already exist (and so, will necessarily be created with exactly these params).  That guarantees either they got what they asked for, or the get an error.  In the future, we could widen the contract to allow for duplicates (with whatever cross-checks we deem appropriate).

If that meets your needs at the moment, I think we should take that route and circle back later.

SeanCurtis-TRI:
> So long as the `MyRenderEngineParams` can be serialized, we can use the `renderer_name` to link the `CameraConfig` to an engine.

Can you elaborate on what you meant by this? In my attempt to infer the mental model that underlies this conclusion, I'm coming up with several different possibilities and would rather not guess.

jwnimmer-tri:
My thought for "we can use the renderer_name to link the CameraConfig to an engine" was like this:

```c++
struct Scenario {
  map<string /* renderer_name */, variant<RenderEngineGltfClientParams, ...>> render_engines;
  map<string, CameraConfig> cameras;
  ...
}
```

First, the simulation creates all the engines.  Second, the simulation adds all the sensors and publishers (using those engines).

That's the MVP.  This PR improves upon that situation by admitting the engine specification directly within the camera specification, which would be useful for engines that are only ever used by exactly one camera.

SeanCurtis-TRI:
That's what I believed you meant (with the highest probability).

re: step 1 & 2. That's a nice breakdown and I'm on board.

re: incremental step 2. Certainly, getting glTF engine through first is the primary unblocker. Currently, we have no way of specifying that render engine at all. Although, if "at all" is the key, a very small PR which appends one more string would be sufficient to unblock that. So, I'd make a counter offer: Step 2's increment should be a) add `RenderEngineGltfClient` as a viable string to the current set. b) add *all* of Drake's parameters at the same time, to maintain symmetry.

jwnimmer-tri:
I'm fine with (2a) for sure.

For (2b) the temporary asymmetry doesn't bother me, but I'm fine with however you want to roll it out; waiting for all of them at once is fine.

jwnimmer-tri:
A final passing thought for the record...

When I first looked over this PR, I had in my mind that `RenderLabel` was a C++ enum.  (It's C++ enums in particular that are difficult to serialize, due to the lack of reflection.)  However, since `RenderLabel` is actually just a regular class, it would actually be plausible to add a not-too-tricky `Serialize` for it, if we ever needed that down the road.

SeanCurtis-TRI:
The challenge in serializing `RenderLabel` is twofold.

1. There are a range of `RenderLabel` values that cannot be instantiated (using anything less than super powers -- `RenderEngine` has such super powers to convert image pixel value to `RenderLabel`): the so-called "reserved" labels.
2. That would make it difficult for someone to put something in the scene and label it as "do not render" or "don't care".

That means we need some kind of serialization that would support either assigning it an actual value. E.g., 

```yaml
label:
  value: 10
```

as well as one of the special values:

```yaml
label:
  value: do_not_render
```

jwnimmer-tri:
> That means we need some kind of serialization that would support either assigning it an actual value ... as well as one of the special values ...

Yup, that's what I had in mind.  If it were an enum, that would be especially difficult.  For a regular class, it's within the realm of possibility.

SeanCurtis-TRI:
What if `RenderLabel` always serialized as a string. The space of *valid* strings would be integers [1, kMaxUnreserve] as strings and the special strings (e.g., "do_not_render").  Then you'd always set the same value in yaml. As I understand it, the yaml text `value: 10` is compatible with a string or int value.

It reduces the complexity of authoring a yaml file (which I'm encountering more in the last couple of weeks than heretofore), by pushing the complexity into the code.

SeanCurtis-TRI:

> 1. The C++ constructor for `RenderEngine` went from taking a `RenderLabel` to taking a `string`. The python bindings omitted this. As breaking changes go, this seems _incredibly_ harmless. I can't imagine there are a lot of people out there implementing `RenderEngine` implementations.

The default choice is always to deprecate.  Only when the deprecation effort is large do we elect to consider breaking changes (subject to considering the impact):

> On very rare occasions it is impractical, too expensive, or logically unsound to meet the full three month deprecation window. In these cases the change will be announced in the “Breaking changes” section of the release notes.

-- https://drake.mit.edu/stable.html

In other words, considering the fallout up-front is premature.  I am happy to push the deprecation patch here myself, once this is a bit more polished.
</blockquote></details>

BTW, as the parameters are almost done. I'm probably going to close this PR in favor of a new PR exploiting all of those changes. It'll be cleaner that way.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Naf746a6cNY0iHS99Vj:-NasVtdS7R4yW93xLnCp:b7zs7lo) (no related file):*

BTW, as the parameters are almost done. I'm probably going to close this PR in favor of a new PR exploiting all of those changes. It'll be cleaner that way.
</blockquote></details>

Fine by me.  I'm also OK with a force-push here that nukes the geometry changes and keeps just the sensors part.

___

SeanCurtis-TRI (commented):
-(status: do not merge) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Naspvyx-DErpm0A56yb:b-b9fcpn)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Naf746a6cNY0iHS99Vj:-NasdHcH8aR8u3rnE26W:b-adf6qb) (no related file):*

Fine by me.  I'm also OK with a force-push here that nukes the geometry changes and keeps just the sensors part.
</blockquote></details>

I just updated this one.

___
___
*[`systems/sensors/camera_config.h` line 265 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaSqN0-7f398kxpOdfw:-NasdF0h1HHo2i8unSd3:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d9b762f95a3b11a7877c62bbc506002871ab85f/systems/sensors/camera_config.h#L265)):*

The cop-out to get this landed quickly would be to say that if the user specifies a `FooParms` in the `renderer_class` variant, then the `renderer_name` must not already exist (and so, will necessarily be created with exactly these params).  That guarantees either they got what they asked for, or the get an error.  In the future, we could widen the contract to allow for duplicates (with whatever cross-checks we deem appropriate).

If that meets your needs at the moment, I think we should take that route and circle back later.
</blockquote></details>

Done

jwnimmer-tri (commented):
Done with everything except the unit test.

> Reviewed 1 of 18 files at r2, 20 of 21 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NasxEy21bfin1G9kHEY:b-bda3yu)* status: 10 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`-- commits` line 12 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NasrSvN8ZRpgsWMVR_F):*
typo



_Suggestion:_
```
GltfClient
```
___
*[`geometry/render_gl/BUILD.bazel` line 89 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NasrFx12O3GZ6QPsfbQ:-NasrFx12O3GZ6QPsfbR:bw3ogo7) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/geometry/render_gl/BUILD.bazel#L89)):*
> ```Starlark
> 
> drake_cc_library(
>     name = "factory",
> ```

nit If you like, for consistency this could be `visibility = public` as well.  The other two engines already have it that way.
___
*[`systems/sensors/camera_config.h` line 259 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NassWVM2MStjOQrkoRD:-NassWVN6sqyNHkUba_F:b-lhwnbn) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L259)):*
> ```objc
>    by providing parameters for one of those engines.
> 
>    If the render engine class _name_ is provided, the engine instantiated will
> ```

Either in this sentence, or possible the prior one -- I think we should say "as a string" to clarify the exact mechanism. So like "... the RenderEngine class to use _as a string_ ..." above, or "... is provided _as a string_ ..." here.

The string vs `!` distinction is going to be somewhat subtle for users to tease apart, so we should help clue them in.

Similarly "... by providing parameters for one of those engines ..." should probably say something like "using the YAML type tag syntax (with an exclamation point) to specify a particular structure type" or the like.
___
*[`systems/sensors/camera_config.h` line 267 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NastbIu2xzQlpYWJSep:-NastbIvCMChLhwTasmd:b-duu94g) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L267)):*
> ```objc
>        renderer_class: RenderEngineVtk
>      other_camera:
>        renderer_class: !RenderEngineVtkParams
> ```

BTW I'm pretty sure in YAML with a type tag we need to say `{}` for it to parse successfully.



_Suggestion:_
```objc
       renderer_class: !RenderEngineVtkParams {}
```
___
*[`systems/sensors/camera_config.h` line 272 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NastgkR8K9nc6j1LqGu:-NastgkS6AZ1d9d5KwvM:b-agalut) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L272)):*
> ```objc
>    the camera named `my_camera` will have the same configuration as
>    `other_camera` -- the default configuration.
> ```

BTW I recommend for clarity here that we either show how to set the extra fields within a params (maybe a glTF URL?) with sample YAML, or cite a demo or test case that does so.  I'm not sure everyone will be familiar with how `std::variant<>` works here -- it's not as widely used in our YAML as other things.
___
*[`systems/sensors/camera_config.h` line 279 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nasu6PrEcme_LZ83XCL:-Nasu6PsE2rKE4DttOJs:b-62gt8i) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L279)):*
> ```objc
>      - Multiple cameras can reference the same value for `renderer_name`.
>      - If multiple cameras reference the same value for `renderer_name`, only
>        the *first* can use the engine parameters to specify it. Attempting to
> ```

I don't mind these semantics, but it will be a little bit tricky to explain a nuance to users: the file-line-number order of dictionaries in a YAML file will _usually_ not be respected while interpreting the data.  In Drake's case, we will nominally treat it like a `std::map<std::string, SubStruct>` and so we will iterate in alphabetical order, not file-line-number order.  So the "first" `!Params` in the file isn't necessarily the first engine we'll create.

I don't have any great suggestions to fix this.  The easy answer is "if you use a `!Params` then make sure its renderer_name is only used that one time" but that's not entirely satisfying.

Part of the challenge is that there is no map-valued `ApplyCameraConfig` API, the user is writing the iteration themselves.  So the order things happen is outside of our purview.  So maybe just saying "first" and let them figure it out, is all we need.



_Code quote:_
```objc
*first*
```
___
*[`systems/sensors/camera_config.cc` line 144 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NasvlFZE5M6C7TbCmds:-NasvlFZE5M6C7TbCmdt:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.cc#L144)):*
> ```cpp
>           "Invalid camera configuration; the given renderer_class value '{}' "
>           "must be empty (to use the default) or be one of 'RenderEngineVtk', "
>           "'RenderEngineGl, or 'RenderEngineGltfClient'.",
> ```

typo



_Suggestion:_
```cpp
'RenderEngineGl',
```
___
*[`systems/sensors/camera_config_functions.cc` line 89 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaswD6g7HeMuAX-aecd:-NaswD6i9AKKffkaIOjp:b3ng8z3) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config_functions.cc#L89)):*
> ```cpp
>     return "RenderEngineGltfClient";
>   } else {
>     DRAKE_UNREACHABLE();
> ```

BTW Can we just delete these two lines?  Gut feeling is that the compiler should let us do it, so long as ParamsType is valid.



_Code quote:_
```cpp
  } else {
    DRAKE_UNREACHABLE();
```
___
*[`systems/sensors/camera_config_functions.cc` line 152 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaswvHqENO3nCUIFCMc:-NaswvHr0jcbXiYHZ9TH:b-jqd3z2) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config_functions.cc#L152)):*
> ```cpp
>               }
>             },
>             [&type_name, &config](auto&&) -> void {
> ```

nit Unused



_Code quote:_
```cpp
&type_name
```
___
*[`systems/sensors/test/camera_config_functions_test.cc` line 313 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NasxA115ND_-cWKR3gI:-NasxA115ND_-cWKR3gJ:b-m1yey3) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/test/camera_config_functions_test.cc#L313)):*
> ```cpp
>   - If a name is reused and the class is specified via parameters, the
>     parameterized spec must come first.
>      */
> ```

nit Weird indentation?

jwnimmer-tri:
BTW This is still marked as a "draft".

SeanCurtis-TRI (commented):
Alrighty. I think that takes care of things.

+@sammy-tri for thursday platform review.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Natkw1O44UZWy6GOkHn:buqnijq)* status: 3 unresolved discussions, LGTM missing from assignees jwnimmer-tri(platform),sammy-tri(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`systems/sensors/camera_config.h` line 259 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NassWVM2MStjOQrkoRD:-NatgiDP8Ei5r9NvVVaI:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L259)):*

Either in this sentence, or possible the prior one -- I think we should say "as a string" to clarify the exact mechanism. So like "... the RenderEngine class to use _as a string_ ..." above, or "... is provided _as a string_ ..." here.

The string vs `!` distinction is going to be somewhat subtle for users to tease apart, so we should help clue them in.

Similarly "... by providing parameters for one of those engines ..." should probably say something like "using the YAML type tag syntax (with an exclamation point) to specify a particular structure type" or the like.
</blockquote></details>

Done.
___
*[`systems/sensors/camera_config.h` line 267 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NastbIu2xzQlpYWJSep:-Nat1Pvk1da5qSzYmHQV:btqqu5l) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L267)):*

BTW I'm pretty sure in YAML with a type tag we need to say `{}` for it to parse successfully.
</blockquote></details>

Correct.
___
*[`systems/sensors/camera_config.h` line 272 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NastgkR8K9nc6j1LqGu:-NatkE7e4ERmOVxchigd:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L272)):*

BTW I recommend for clarity here that we either show how to set the extra fields within a params (maybe a glTF URL?) with sample YAML, or cite a demo or test case that does so.  I'm not sure everyone will be familiar with how `std::variant<>` works here -- it's not as widely used in our YAML as other things.
</blockquote></details>

Done
___
*[`systems/sensors/camera_config.h` line 279 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nasu6PrEcme_LZ83XCL:-NaszfHw424I1nbw8k-D:b-95zj2r) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L279)):*

I don't mind these semantics, but it will be a little bit tricky to explain a nuance to users: the file-line-number order of dictionaries in a YAML file will _usually_ not be respected while interpreting the data.  In Drake's case, we will nominally treat it like a `std::map<std::string, SubStruct>` and so we will iterate in alphabetical order, not file-line-number order.  So the "first" `!Params` in the file isn't necessarily the first engine we'll create.

I don't have any great suggestions to fix this.  The easy answer is "if you use a `!Params` then make sure its renderer_name is only used that one time" but that's not entirely satisfying.

Part of the challenge is that there is no map-valued `ApplyCameraConfig` API, the user is writing the iteration themselves.  So the order things happen is outside of our purview.  So maybe just saying "first" and let them figure it out, is all we need.
</blockquote></details>

I struggled articulating the restriction. Ordering matters so much.

My first draft was that only one could specify `renderer_class` at all. But that isn't sufficient for exactly the same reason.

It makes me wonder about our use of a map in our example `Scenario` class. If it were simply a `vector`, order would be guaranteed. As it is, we have to distinguish between the map key (which is purely a mnemonic) vs the camera name, which is functional.  You'd think the camera name could/should serve both roles. And then we'd be conditioning our users to think about applying cameras in order.

As it is, I've tried to give a bit more guidance in this regard. PTAL.
___
*[`systems/sensors/camera_config_functions.cc` line 89 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NaswD6g7HeMuAX-aecd:-NatkY7N-YqQCe8YhHlu:b-qoy34l) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config_functions.cc#L89)):*

BTW Can we just delete these two lines?  Gut feeling is that the compiler should let us do it, so long as ParamsType is valid.
</blockquote></details>

Right. If you pass any other type, there won't be a return. So, the compiler will already catch it.

jwnimmer-tri (commented):
> Reviewed 4 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: 4 unresolved discussions, LGTM missing from assignees sammy-tri(platform),jwnimmer-tri(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`systems/sensors/camera_config.h` line 279 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nasu6PrEcme_LZ83XCL:-NavoY1J6onlF8z_EGZB:b-a03m97) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L279)):*

I struggled articulating the restriction. Ordering matters so much.

My first draft was that only one could specify `renderer_class` at all. But that isn't sufficient for exactly the same reason.

It makes me wonder about our use of a map in our example `Scenario` class. If it were simply a `vector`, order would be guaranteed. As it is, we have to distinguish between the map key (which is purely a mnemonic) vs the camera name, which is functional.  You'd think the camera name could/should serve both roles. And then we'd be conditioning our users to think about applying cameras in order.

As it is, I've tried to give a bit more guidance in this regard. PTAL.
</blockquote></details>

FYI the advantage of `std::map<>` in the example is that it allows the user to tweak the settings in the main yaml file, though a second yaml "file" (the `--scenario_text` argument).  In our YAML architecture, maps can be incrementally tweaked, while arrays can only be replace wholesale.
___
*[`systems/sensors/camera_config.h` line 255 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NavntWCAssc3lEmj9mT:-NavntWCAssc3lEmj9mU:bmj5vv9) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/camera_config.h#L255)):*
> ```objc
> 
>   /** The choice of render engine implementation to use. It can be specified
>    simply by providing a `string` containing the _name_ of the RenderEngine
> ```

nit Maybe "class name of the RenderEngine to use" flows more smoothly than "name of the RenderEngine class to use"?
___
*[`systems/sensors/camera_config.h` line 258 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NavoFAeC_WcFCBjSdjH:-NavoFAfB7quQ52lLTb4:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/camera_config.h#L258)):*
> ```objc
>    class to use (i.e., `"RenderEngineVtk"`, `"RenderEngineGl"`, or
>    `"RenderEngineGltfClient"`). Or, it can be specified by providing parameters
>    for one of those engines: RenderEngineParamsVtk, RenderEngineParamsGl, or
> ```

typo



_Suggestion:_
```objc
RenderEngineVtkParams, RenderEngineGlParams
```
___
*[`systems/sensors/camera_config.h` line 324 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NavpMndBamVLbEx0iqq:-NavpMne97WL5Uo5Loi-:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/camera_config.h#L324)):*
> ```objc
>    @endcode
> 
>    4. Use the RenderEngineVtk with a customized clear color (black).
> ```

typo



_Suggestion:_
```objc
RenderEngineGl
```
___
*[`systems/sensors/camera_config.h` line 328 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-NavpQml5laBCPGAzfz0:-NavpQml5laBCPGAzfz1:b-7ilvee) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/camera_config.h#L328)):*
> ```objc
>    renderer_class: !RenderEngineGlParams
>      default_clear_color:
>        rgba [0, 0, 0, 1])
> ```

typos



_Suggestion:_
```objc
rgba: [0, 0, 0, 1]
```

sammy-tri (commented):
:lgtm: (just typos/nits remaining)

> Reviewed 16 of 21 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-NawuX4oEcAnGYulLoOa:bhay87y)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @SeanCurtis-TRI)

jwnimmer-tri (commented):
:lgtm:

> Reviewed 2 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Nax26Y31MJFhYZpjVHM:bnfp4nl)* status: 10 unresolved discussions, missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`systems/sensors/test/camera_config_functions_test.cc` line 276 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax0N0Q3i4OUl0f1QaE:-Nax0N0Q3i4OUl0f1QaF:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_functions_test.cc#L276)):*
> ```cpp
>                 testing::EndsWith("RenderEngineGltfClient"));
>   }
>   // Add VTK by parameters.
> ```

typo



_Suggestion:_
```cpp
glTF client
```
___
*[`systems/sensors/test/camera_config_functions_test.cc` line 320 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax1Ba773Di9FBCylPV:-Nax1Ba82Gfpc2dsgJ3C:b-y6nxfn) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_functions_test.cc#L320)):*
> ```cpp
> 
>   auto perform_test = [this, &renderer_count](const auto& parameters,
>                                               const std::string& name,
> ```

nit I frequently got lost when reading this helper function.  I think a better argument name here would help a lot.

It's very much _not_ the `CameraConfig.name` nor `CamerConfig.render_name`.



_Suggestion:_
```cpp
renderer_class
```
___
*[`systems/sensors/test/camera_config_functions_test.cc` line 322 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax1nBe6ioLroVx0hZw:-Nax1nBfDjMXXSiqd-dQ:b-admevt) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_functions_test.cc#L322)):*
> ```cpp
>                                               const std::string& name,
>                                               const std::string& prefix) {
>     CameraConfig config{.renderer_name = prefix + "_1",
> ```

BTW It would have helped me ease into these test cases if `.name = ...` were set here as well, just to remind me about the concepts in play in this test case.
___
*[`systems/sensors/test/camera_config_functions_test.cc` line 342 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax1d_6BsgsYanFjLVD:-Nax1d_78PA4iJvG_jnI:bxhyl43) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_functions_test.cc#L342)):*
> ```cpp
> 
>     // Camera config using the name of a different class is angry.
>     config.name = "should've_been_" + prefix;
> ```

BTW why not just `_4`?  I had to pause and think if something special is happening here.
___
*[`systems/sensors/test/camera_config_functions_test.cc` line 356 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax23wF0TO2ERJ5nm8o:-Nax23wF0TO2ERJ5nm8p:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_functions_test.cc#L356)):*
> ```cpp
> 
>   {
>     SCOPED_TRACE("ClientGltf");
> ```

typo



_Suggestion:_
```cpp
GltfClient
```
___
*[`systems/sensors/test/camera_config_test.cc` line 23 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nax-_-W8k8CQ6Ucls7f:-Nax-_-XDuTt3RMgSibj:b-kn9xuk) ([raw file](https://github.com/RobotLocomotion/drake/blob/30713730ecfccb578f72411349ffbf32b8bf86f4/systems/sensors/test/camera_config_test.cc#L23)):*
> ```cpp
> 
> // Some quick and dirty comparison operators so we can use gtest matchers with
> // the render engine parameter types.
> ```

BTW If we don't care about canonicalizing the params (e.g., promoting options to their implied values) -- and it seems like we don't -- then a more robust way to do this would be to serialize `p1` and `p2` to YAML and do a textual comparison of the YAML strings.

SeanCurtis-TRI (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Naxm1IQ-Z6y0ukSTNwJ:b-ucfknh)* status: missing label for release notes (waiting on @SeanCurtis-TRI)

___
*[`systems/sensors/camera_config.h` line 279 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-Nasu6PrEcme_LZ83XCL:-Naxhovn2lQVTTaqF7Av:b99g0ja) ([raw file](https://github.com/RobotLocomotion/drake/blob/c1d18f61fe1432648b985b44cbc8c098917b2fff/systems/sensors/camera_config.h#L279)):*

FYI the advantage of `std::map<>` in the example is that it allows the user to tweak the settings in the main yaml file, though a second yaml "file" (the `--scenario_text` argument).  In our YAML architecture, maps can be incrementally tweaked, while arrays can only be replace wholesale.
</blockquote></details>

Interesting. I hadn't known that.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863)* status: missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19863#-:-Nay8cBW2zJMEgEd9Adh:b-l8xbmn)* status: :shipit: complete! all discussions resolved, LGTM from assignees sammy-tri(platform),jwnimmer-tri(platform)

jwnimmer-tri:
+(status: breaking change)

C++ code that was reading from the `renderer_class` field now gets a `std::variant` back, instead of a `std::string` directly.  Downstream code will need to adapt with `std::get_if<std::string>(&params.renderer_class)` to compensate.

PR #19864, opened 2023-07-28
Implement a ConvexSet Constructor to Compute the Affine Hull of another ConvexSet

cohnt:
Towards #19717 

(We're getting pretty close now!)

The procedure to compute the convex hull itself is implemented in `affine_subspace.cc`. It works by incrementally building a list of feasible line segments within the convex set that are mutually orthogonal, until no more can be found. This gives us the basis we need for the AffineSubspace, and any feasible point can serve as the translation.

Most of the additions in this PR are a comprehensive set of test cases, as we check that it can correctly produce the affine hull of all subclasses of ConvexSet, with a variety of special cases. Note that the tolerances for the default solvers are somewhat poor (especially for the intersection cases). When using commercial solvers (e.g. Gurobi), the performance is much better.

If you're up to it, +@hongkai-dai for feature review? Otherwise, I can reassign it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_optimization.cc
+++ b/bindings/pydrake/geometry/geometry_py_optimization.cc
@@ -90,7 +90,10 @@ void DefineGeometryOptimization(py::module m) {
         .def(py::init<>(), cls_doc.ctor.doc_0args)
         .def(py::init<const Eigen::Ref<const Eigen::MatrixXd>&,
                  const Eigen::Ref<const Eigen::VectorXd>&>(),
-            py::arg("basis"), py::arg("translation"), cls_doc.ctor.doc_2args)
+            py::arg("basis"), py::arg("translation"),
+            cls_doc.ctor.doc_2args_basis_translation)
+        .def(py::init<const ConvexSet&, double>(), py::arg("set"),
+            py::arg("tol") = 0, cls_doc.ctor.doc_2args_set_tol)
         .def("basis", &AffineSubspace::basis, py_rvp::reference_internal,
             cls_doc.basis.doc)
         .def("translation", &AffineSubspace::translation,
--- a/bindings/pydrake/geometry/test/optimization_test.py
+++ b/bindings/pydrake/geometry/test/optimization_test.py
@@ -122,6 +122,10 @@ class TestGeometryOptimization(unittest.TestCase):
         self.assertIsNot(dut.Clone(), dut)
         self.assertIsNot(copy.deepcopy(dut), dut)
 
+        p = np.array([11.1, 12.2, 13.3])
+        point = mut.Point(p)
+        aff = mut.AffineSubspace(set=point, tol=1e-12)
+
     def test_h_polyhedron(self):
         mut.HPolyhedron()
         hpoly = mut.HPolyhedron(A=self.A, b=self.b)
--- a/geometry/optimization/affine_subspace.cc
+++ b/geometry/optimization/affine_subspace.cc
@@ -1,6 +1,8 @@
 #include "drake/geometry/optimization/affine_subspace.h"
 
 #include "drake/common/is_approx_equal_abstol.h"
+#include "drake/geometry/optimization/spectrahedron.h"
+#include "drake/solvers/solve.h"
 
 namespace drake {
 namespace geometry {
@@ -11,6 +13,8 @@ using Eigen::RowVectorXd;
 using Eigen::VectorXd;
 using solvers::Binding;
 using solvers::Constraint;
+using solvers::LinearCost;
+using solvers::MathematicalProgram;
 using solvers::VectorXDecisionVariable;
 using symbolic::Variable;
 
@@ -28,6 +32,86 @@ AffineSubspace::AffineSubspace(const Eigen::Ref<const MatrixXd>& basis,
   } else {
     basis_decomp_ = std::nullopt;
   }
+}
+
+AffineSubspace::AffineSubspace(const ConvexSet& set, double tol)
+    : ConvexSet(0) {
+  // If the set is clearly a singleton, we can easily compute its affine hull.
+  const auto singleton_maybe = set.MaybeGetPoint();
+  if (singleton_maybe.has_value()) {
+    // Fall back to the basis and translation constructor.
+    *this = AffineSubspace(Eigen::MatrixXd::Zero(set.ambient_dimension(), 0),
+                           singleton_maybe.value());
+    return;
+  }
+
+  // If the set is not clearly a singleton, we find a feasible point and
+  // iteratively compute a basis of the affine hull. If no feasible point
+  // exists, the set is empty, so we throw an error.
+  const auto translation_maybe = set.MaybeGetFeasiblePoint();
+  if (!translation_maybe.has_value()) {
+    throw std::runtime_error(
+        "AffineSubspace: Cannot take the affine hull of an empty set!");
+  }
+  const VectorXd translation = translation_maybe.value();
+  std::vector<VectorXd> basis_vectors;
+
+  // Create the mathematical program we will use to find basis vectors.
+  MathematicalProgram prog;
+
+  // x represents the displacement within a feasible set. We add a bounding box
+  // constraint to ensure the problem is not unbounded.
+  VectorXDecisionVariable x =
+      prog.NewContinuousVariables(set.ambient_dimension(), "x");
+  prog.AddBoundingBoxConstraint(-1, 1, x);
+
+  // y and z are two feasible points in the set
+  VectorXDecisionVariable y =
+      prog.NewContinuousVariables(set.ambient_dimension(), "y");
+  VectorXDecisionVariable z =
+      prog.NewContinuousVariables(set.ambient_dimension(), "z");
+  prog.AddLinearConstraint(x == y - z);
+  set.AddPointInSetConstraints(&prog, y);
+  set.AddPointInSetConstraints(&prog, z);
+
+  // This is the objective we use. We will iteratively try to minimize the ith
+  // component of x for each dimension. Anytime this value is negative, it means
+  // we have found a new feasible direction, so we add it to the basis, and also
+  // constrain future feasible x vectors to be orthogonal to it (to ensure
+  // linear independence of the basis).
+  VectorXd new_objective_vector = VectorXd::Zero(set.ambient_dimension());
+  Binding<LinearCost> objective = prog.AddLinearCost(new_objective_vector, x);
+
+  for (int i = 0; i < set.ambient_dimension(); ++i) {
+    // Update the objective to check the ith dimension
+    new_objective_vector.setZero();
+    new_objective_vector[i] = 1;
+    objective.evaluator()->UpdateCoefficients(new_objective_vector);
+
+    // Minimize x[i]
+    auto result = solvers::Solve(prog);
+    if (!result.is_success()) {
+      throw std::runtime_error(
+          fmt::format("AffineSubspace: Failed to compute the affine hull! The "
+                      "solution result was {}.",
+                      result.get_solution_result()));
+    }
+    if (result.get_optimal_cost() < -tol) {
+      // Get the solution as a new basis vector, and add a constraint that x
+      // must now be orthogonal to that new basis vector.
+      basis_vectors.push_back(result.GetSolution(x));
+      prog.AddLinearConstraint(basis_vectors.back(), 0, 0, x);
+    }
+  }
+
+  // By construction, basis_vectors is linearly independent. Because we have
+  // checked each direction in the ambient space, it will also span the other
+  // convex set, so we now have its affine hull.
+  MatrixXd basis(set.ambient_dimension(), basis_vectors.size());
+  for (size_t i = 0; i < basis_vectors.size(); ++i) {
+    basis.col(i) = basis_vectors[i];
+  }
+  *this = AffineSubspace(basis, translation);
 }
 
 AffineSubspace::~AffineSubspace() = default;
--- a/geometry/optimization/affine_subspace.h
+++ b/geometry/optimization/affine_subspace.h
@@ -41,6 +41,21 @@ class AffineSubspace final : public ConvexSet {
   explicit AffineSubspace(const Eigen::Ref<const Eigen::MatrixXd>& basis,
                           const Eigen::Ref<const Eigen::VectorXd>& translation);
 
+  /** Constructs an affine subspace as the affine hull of another convex set.
+  This is done by finding a feasible point in the set, and then iteratively
+  computing feasible vectors until we have a basis that spans the set. If you
+  pass in a convex set whose points are matrix-valued (e.g. a Spectrahedron),
+  then the affine subspace will work over a flattened representation of those
+  coordinates. (So a Spectrahedron with n-by-n matrices will output an
+  AffineSubspace with ambient dimension (n * (n+1)) / 2.)
+
+  `tol` sets the numerical precision of the computation. For each dimension, a
+  pair of feasible points are constructed, so as to maximize the displacement in
+  that dimension. If their displacement along that dimension is larger than tol,
+  then the vector connecting the points is added as a basis vector.
+  @pre !set.IsEmpty() */
+  explicit AffineSubspace(const ConvexSet& set, double tol = 0);
+
   ~AffineSubspace() final;
 
   /** Returns the basis in an n-by-m matrix, where n is the ambient dimension,
--- a/geometry/optimization/test/affine_subspace_test.cc
+++ b/geometry/optimization/test/affine_subspace_test.cc
@@ -4,6 +4,15 @@
 
 #include "drake/common/test_utilities/eigen_matrix_compare.h"
 #include "drake/common/yaml/yaml_io.h"
+#include "drake/geometry/optimization/cartesian_product.h"
+#include "drake/geometry/optimization/convex_set.h"
+#include "drake/geometry/optimization/hpolyhedron.h"
+#include "drake/geometry/optimization/hyperellipsoid.h"
+#include "drake/geometry/optimization/intersection.h"
+#include "drake/geometry/optimization/minkowski_sum.h"
+#include "drake/geometry/optimization/point.h"
+#include "drake/geometry/optimization/spectrahedron.h"
+#include "drake/geometry/optimization/vpolytope.h"
 #include "drake/solvers/solve.h"
 
 namespace drake {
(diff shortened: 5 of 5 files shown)


hongkai-dai (commented):
Sorry for the delay, PTAL

> Reviewed 2 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-:-Nahz9OF9myHqYloKSMM:bubxuvs)* status: 11 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahoyeR3dWmLl2brMx_:-NahoyeR3dWmLl2brMxa:b-yxre1n) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.h#L49)):*
> ```objc
>   pass in a convex sets whose points are matrix-valued (e.g. a Spectrahedron),
>   then the affine subspace will work over a flattened representation of those
>   coordinates. (So a Spectrahedron with n-by-n matrices will output an
> ```

BTW, tthe ambient dimension of a Spectrahedron is (n+1) * n / 2 because it is symmetric, so should this affine space also be of dimension (n+1)*n / 2?
___
*[`geometry/optimization/affine_subspace.h` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahqpo9Ae2uNqYR590z:-Nahqpo9Ae2uNqYR591-:b-u0nbj3) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.h#L51)):*
> ```objc
>   coordinates. (So a Spectrahedron with n-by-n matrices will output an
>   AffineSubspace of R^(n^2).)
>   @pre !set.IsEmpty() */
> ```

nit, document what `tol` means?
___
*[`geometry/optimization/affine_subspace.cc` line 40 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahrcts5bEH6uhaXklM:-Nahrcts5bEH6uhaXklN:bqwr42u) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L40)):*
> ```cpp
>     : ConvexSet(0) {
>   // If the set is clearly a singleton, we can easily compute its affine hull.
>   auto singleton_maybe = set.MaybeGetPoint();
> ```

BTW, `const auto`.
___
*[`geometry/optimization/affine_subspace.cc` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahrfGC5o7ycBwOH-U9:-NahrfGC5o7ycBwOH-UA:b-409m0i) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L51)):*
> ```cpp
>   // iteratively compute a basis of the affine hull. If no feasible point
>   // exists, the set is empty, so we throw an error.
>   auto translation_maybe = set.MaybeGetFeasiblePoint();
> ```

BTW, const auto.
___
*[`geometry/optimization/affine_subspace.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahrjZs90uJ3EHsLENs:-NahrjZs90uJ3EHsLENt:b-9c7fty) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L53)):*
> ```cpp
>   auto translation_maybe = set.MaybeGetFeasiblePoint();
>   if (!translation_maybe.has_value()) {
>     throw std::runtime_error("Cannot take the affine hull of an empty set!");
> ```

Also indicate the caller function in the error message
```cc
throw std::runtime_error("AffineSubspace: Cannot take the affine hull of an empty set!");
```
___
*[`geometry/optimization/affine_subspace.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahs4iE1_aDhDjXdzpq:-Nahs4iE1_aDhDjXdzpr:b-k6zjza) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L65)):*
> ```cpp
>   VectorXDecisionVariable x =
>       prog.NewContinuousVariables(set.ambient_dimension(), "x");
>   prog.AddBoundingBoxConstraint(VectorXd::Constant(set.ambient_dimension(), -1),
> ```

BTW, you could write `prog.AddBoundingBoxConstraint(-1, 1, x)` for brevity.
___
*[`geometry/optimization/affine_subspace.cc` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahs_dvDDRaSCq-r7hc:-Nahs_dvDDRaSCq-r7hd:bv235u6) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L88)):*
> ```cpp
>   for (int i = 0; i < set.ambient_dimension(); ++i) {
>     // Update the objective to check the ith dimension
>     VectorXd new_objective_vector = VectorXd::Zero(set.ambient_dimension());
> ```

BTW, this would do heap allocation in every iteration of the for loop. You could move this `new_objective_vector` outside of the for loop, and call `new_objective_vector.setZero(); new_objective_vector(i) = 1;` within the for loop.
___
*[`geometry/optimization/affine_subspace.cc` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahspzO8joJVNclNIA1:-NahspzO8joJVNclNIA2:b-me5qml) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L96)):*
> ```cpp
>     if (!result.is_success()) {
>       throw std::runtime_error(fmt::format(
>           "Failed to compute the affine hull! The solution result was {}.",
> ```

Add the function name to the error message.
___
*[`geometry/optimization/affine_subspace.cc` line 103 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Naht4ZHEnuFaKtKAg7v:-Naht4ZHEnuFaKtKAg7w:b-mwial8) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L103)):*
> ```cpp
>       // must now be orthogonal to that new basis vector.
>       basis_vectors.push_back(result.GetSolution(x));
>       prog.AddLinearConstraint(result.GetSolution(x), 0, 0, x);
> ```

This calls `result.GetSolution(x)` again. You could change it to `prog.AddLinearConstraint(basis_vectors.back(), 0, 0, x)`.
___
*[`geometry/optimization/affine_subspace.cc` line 104 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahtHim6_gG1nnZCayt:-NahtHim6_gG1nnZCayu:bxfq1is) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L104)):*
> ```cpp
>       basis_vectors.push_back(result.GetSolution(x));
>       prog.AddLinearConstraint(result.GetSolution(x), 0, 0, x);
>       continue;
> ```

continue is redundant?
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 318 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahvz8n24VqPfpih9f7:-Nahvz8n24VqPfpih9f8:b8pf3pa) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/test/affine_subspace_test.cc#L318)):*
> ```cpp
>   set.AddPointInSetConstraints(&prog, y);
> 
>   // This is the objective we use. We will iteratively try to minimize and
> ```

I think the test here checks if the convex set `set` is a subset of the affine subspace `as`, but it doesn't check if `as` is the affine hull? Namely if the convex set in R^3 is actually on a two dimensional plane, and we choose `as` to be the full 3D space, this test would still pass?

May I suggest some additional test? Let's say the affine subspace is written as {y | y = A*x + b}, where `A` is the basis and `b` is the translation, we can create a function `bool IsSetInAffineSubspace(const ConvexSet& set, const AffineSubspace& as)` which returns true if `set` is a subset of `as`, and false otherwise. We then loop through each basis of the affine subspace, create a new affine subspace by removing one basis from `as`, denote this new affine_subspace as `as_remove_one_basis`, and we call `IsSetInAffineSubspace(set, as_remove_one_basis)`, and we should expect it returns false.

In order to check if `convex set S is a subset of affine subspace {y | y = A*x+b}`, we could test if S intersecting with the complement of the affine subspace is empty. Namely we solve the following optimization problem
```
max x_perp(i) or min x_perp(i)
s.t y in S
    y = A* x + A_perp * x_perp + b
```
where A_perp is the basis perpendicular to the affine subspace basis `A`.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864)* status: 5 unresolved discussions, LGTM missing from assignee hongkai-dai, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`geometry/optimization/affine_subspace.h` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahoyeR3dWmLl2brMx_:-NalXjHZ67ZSB3jY1TjS:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.h#L49)):*

BTW, tthe ambient dimension of a Spectrahedron is (n+1) * n / 2 because it is symmetric, so should this affine space also be of dimension (n+1)*n / 2?
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.h` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahqpo9Ae2uNqYR590z:-NalXif_CtnXZiADIWXi:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.h#L51)):*

nit, document what `tol` means?
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 40 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahrcts5bEH6uhaXklM:-NalXkJlDyaMPa3N371I:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L40)):*

BTW, `const auto`.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 51 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahrfGC5o7ycBwOH-U9:-NalXks87EWS7gj4yI-B:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L51)):*

BTW, const auto.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 53 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahrjZs90uJ3EHsLENs:-NalXlTT-o3HbtoQ5-Cs:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L53)):*

Also indicate the caller function in the error message
```cc
throw std::runtime_error("AffineSubspace: Cannot take the affine hull of an empty set!");
```
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 65 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahs4iE1_aDhDjXdzpq:-NalXm1NDOMmCuITb1Ax:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L65)):*

BTW, you could write `prog.AddBoundingBoxConstraint(-1, 1, x)` for brevity.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 88 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahs_dvDDRaSCq-r7hc:-NalXmhN-78UQSp9MLuP:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L88)):*

BTW, this would do heap allocation in every iteration of the for loop. You could move this `new_objective_vector` outside of the for loop, and call `new_objective_vector.setZero(); new_objective_vector(i) = 1;` within the for loop.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 96 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahspzO8joJVNclNIA1:-NalXnUCBpOAfZDvAUjM:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L96)):*

Add the function name to the error message.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 103 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Naht4ZHEnuFaKtKAg7v:-NalXnzq6UmSFfUDY84s:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L103)):*

This calls `result.GetSolution(x)` again. You could change it to `prog.AddLinearConstraint(basis_vectors.back(), 0, 0, x)`.
</blockquote></details>

Done.
___
*[`geometry/optimization/affine_subspace.cc` line 104 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NahtHim6_gG1nnZCayt:-NalXokyDTlVpkWFLDcW:bo1bjv6) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/affine_subspace.cc#L104)):*

continue is redundant?
</blockquote></details>

Yep. Removed it.
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 318 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahvz8n24VqPfpih9f7:-NalXrd08nTifS9_j46y:bhyz7yf) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/test/affine_subspace_test.cc#L318)):*

I think the test here checks if the convex set `set` is a subset of the affine subspace `as`, but it doesn't check if `as` is the affine hull? Namely if the convex set in R^3 is actually on a two dimensional plane, and we choose `as` to be the full 3D space, this test would still pass?

May I suggest some additional test? Let's say the affine subspace is written as {y | y = A*x + b}, where `A` is the basis and `b` is the translation, we can create a function `bool IsSetInAffineSubspace(const ConvexSet& set, const AffineSubspace& as)` which returns true if `set` is a subset of `as`, and false otherwise. We then loop through each basis of the affine subspace, create a new affine subspace by removing one basis from `as`, denote this new affine_subspace as `as_remove_one_basis`, and we call `IsSetInAffineSubspace(set, as_remove_one_basis)`, and we should expect it returns false.

In order to check if `convex set S is a subset of affine subspace {y | y = A*x+b}`, we could test if S intersecting with the complement of the affine subspace is empty. Namely we solve the following optimization problem
```
max x_perp(i) or min x_perp(i)
s.t y in S
    y = A* x + A_perp * x_perp + b
```
where A_perp is the basis perpendicular to the affine subspace basis `A`.
</blockquote></details>

I made some modifications, but I think the idea is the same. Now, checking the maximal point along each dimension of the set is used to check set containment, and I have a separate function `CheckAffineHullTightness` which iteratively removes each basis vector, and checks that the set is now *not* contained in the new affine space.

hongkai-dai (commented):
:lgtm: +@rpoyner-tri for platform review please, thanks!

> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-:-NalwDNSEvw2iCtVuYk1:b-sjvomj)* status: 1 unresolved discussion, LGTM missing from assignee rpoyner-tri(platform)

___
*[`geometry/optimization/affine_subspace.cc` line 102 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NalvGRuCwqkUNtDoMlb:-NalvGRuCwqkUNtDoMlc:bip5eot) ([raw file](https://github.com/RobotLocomotion/drake/blob/3c4f402311f4ae1ba09c1408a472f816a62e25a9/geometry/optimization/affine_subspace.cc#L102)):*
> ```cpp
>       // Get the solution as a new basis vector, and add a constraint that x
>       // must now be orthogonal to that new basis vector.
>       basis_vectors.push_back(result.GetSolution(x));
> ```

BTW, it just occured to me that `result.GetSolution(x)` is not a unit-length vector. I know that we do NOT require the basis vectors to be unit length, but would it be better to normalize the basis? I am OK with either way, but just want to mention it for discussion.
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 318 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-Nahvz8n24VqPfpih9f7:-Nalvob010lUpN0cDqYy:b-vs1prh) ([raw file](https://github.com/RobotLocomotion/drake/blob/0720ae83c1d4abe224448b40e857a7dc1a6c3c85/geometry/optimization/test/affine_subspace_test.cc#L318)):*

I made some modifications, but I think the idea is the same. Now, checking the maximal point along each dimension of the set is used to check set containment, and I have a separate function `CheckAffineHullTightness` which iteratively removes each basis vector, and checks that the set is now *not* contained in the new affine space.
</blockquote></details>

Looks good!

rpoyner-tri (commented):
:lgtm: 

> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-:-NamJAEs2LDM3dZRuu5n:bg7igkv)* status: 2 unresolved discussions

___
*[`geometry/optimization/affine_subspace.h` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NamI46X8hq5x_ElnutI:-NamI46Y6Yy58dtCCDwF:b-punf4y) ([raw file](https://github.com/RobotLocomotion/drake/blob/3c4f402311f4ae1ba09c1408a472f816a62e25a9/geometry/optimization/affine_subspace.h#L47)):*
> ```objc
>   This is done by finding a feasible point in the set, and then iteratively
>   computing feasible vectors until we have a basis that spans the set. If you
>   pass in a convex sets whose points are matrix-valued (e.g. a Spectrahedron),
> ```

maybe typo?



_Suggestion:_
```objc
a convex set
```
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 559 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NamKUtvDJXUQCJ3SSDd:-NamKUtvDJXUQCJ3SSDe:b-t509hd) ([raw file](https://github.com/RobotLocomotion/drake/blob/3c4f402311f4ae1ba09c1408a472f816a62e25a9/geometry/optimization/test/affine_subspace_test.cc#L559)):*
> ```cpp
>   AffineSubspace as4(i4);
> 
>   // The numerics are much better with commerical sovlers
> ```

typo, and should end with a period (.) .



_Suggestion:_
```cpp
solvers.
```

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19864)* status: 1 unresolved discussion

___
*[`geometry/optimization/affine_subspace.h` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NamI46X8hq5x_ElnutI:-NamQ7939sMtc-EFuDdC:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3c4f402311f4ae1ba09c1408a472f816a62e25a9/geometry/optimization/affine_subspace.h#L47)):*

maybe typo?
</blockquote></details>

Done.
___
*[`geometry/optimization/test/affine_subspace_test.cc` line 559 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19864#-NamKUtvDJXUQCJ3SSDd:-NamQ8QX0UrSNCcmaRUB:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/3c4f402311f4ae1ba09c1408a472f816a62e25a9/geometry/optimization/test/affine_subspace_test.cc#L559)):*

typo, and should end with a period (.) .
</blockquote></details>

Done.

PR #19878, opened 2023-07-30
[py visualization] Fix Meldis to recognize mesh file content changes

### LINKED ISSUE #19763: Meldis ignores repeated load commands (even if meshes change)
### What happened?

As a feature to facilitate running the same simulation over and over, a single meldis session will detect multiple load messages in a row and not load the redundant load messages. This certainly makes re-running the same simulation far nicer. It does this by looking at the contents of the load message, inferring identical load messages means no change.

However, the load message can refer to other files (specifically mesh files). If those mesh files' contents have changed, that change will not be detected by the simple load-message heuristic. Re-running the simulation will maintain the old geometries and not the current on-disk state of those additional files.

### Solution

Ideally, meldis should recognize if any of those particular files have changed. If they have changed, then the scene should reload (or, ideally, only those meshes that have changed should be updated).

### Version

_No response_

_No response_

_No response_

### Relevant log output

_No response_

jwnimmer-tri:
Closes #19763.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878)

### DIFF
--- a/bindings/pydrake/visualization/_meldis.py
+++ b/bindings/pydrake/visualization/_meldis.py
@@ -1,7 +1,9 @@
 import copy
+import hashlib
 import logging
 import numpy as np
 from pathlib import Path
+import re
 import sys
 import time
 
@@ -90,6 +92,70 @@ class _Slider:
         return value, value_changed
 
 
+class _GeometryFileHasher:
+    """Calculates a checksum of external file(s) referenced by geometry
+    messages such as lcmt_viewer_load_robot or similar.
+
+    Each "on_..." method incorporates all of the external files cited by the
+    given argument (of a specific type) into the current hash. Files that are
+    named but cannot be opened are silently skipped.
+    """
+
+    def __init__(self):
+        self._paths = []
+        self._hasher = hashlib.sha256()
+
+    def value(self):
+        return self._hasher.hexdigest()
+
+    def _read_file(self, path: Path):
+        """Reads the given file and adds its content to the current hash.
+        Returns the file content (as ``bytes``, not ``str``).
+        Remembers the filename (for unit testing).
+        If the file is missing, silently returns an empty ``bytes``.
+        """
+        try:
+            with open(path, "rb") as f:
+                content = f.read()
+            self._hasher.update(content)
+            self._paths.append(path)
+            return content
+        except IOError:
+            return b""
+
+    def on_viewer_load_robot(self, message: lcmt_viewer_load_robot):
+        assert isinstance(message, lcmt_viewer_load_robot)
+        for link in message.link:
+            for geom in link.geom:
+                self.on_viewer_geometry_data(geom)
+
+    def on_viewer_geometry_data(self, message: lcmt_viewer_geometry_data):
+        assert isinstance(message, lcmt_viewer_geometry_data)
+        if (message.type == lcmt_viewer_geometry_data.MESH
+                and message.string_data):
+            self.on_mesh(Path(message.string_data))
+
+    def on_mesh(self, path: Path):
+        assert isinstance(path, Path)
+        content = self._read_file(path)
+        if path.suffix.lower() == ".obj":
+            for mtl_names in re.findall(rb"^\s*mtllib\s+(.*?)\s*$", content,
+                                        re.MULTILINE):
+                for mtl_name in mtl_names.decode("utf-8").split():
+                    self.on_mtl(path.parent / mtl_name)
+
+    def on_mtl(self, path: Path):
+        assert isinstance(path, Path)
+        content = self._read_file(path)
+        for tex_name in re.findall(rb"^\s*map_.*?\s+(\S+)\s*$", content,
+                                   re.MULTILINE):
+            self.on_texture(path.parent / tex_name.decode("utf-8"))
+
+    def on_texture(self, path: Path):
+        assert isinstance(path, Path)
+        self._read_file(path)
+
+
 class _ViewerApplet:
     """Displays lcmt_viewer_load_robot and lcmt_viewer_draw into MeshCat."""
 
@@ -105,6 +171,7 @@ class _ViewerApplet:
         self._meshcat = meshcat
         self._path = path
         self._load_message = None
+        self._load_message_mesh_checksum = None
         self._alpha_slider = _Slider(meshcat, alpha_slider_name)
         if should_accept_link is not None:
             self._should_accept_link = should_accept_link
@@ -123,9 +190,13 @@ class _ViewerApplet:
         # performance when the user is repeatedly viewing the same simulation
         # over and over again, since reloading a scene into Meshcat has high
         # latency.
+        hasher = _GeometryFileHasher()
+        hasher.on_viewer_load_robot(message)
+        mesh_checksum = hasher.value()
         if self._load_message is not None:
             if (message.num_links == self._load_message.num_links
-                    and message.encode() == self._load_message.encode()):
+                    and message.encode() == self._load_message.encode()
+                    and mesh_checksum == self._load_message_mesh_checksum):
                 _logger.info("Ignoring duplicate load message")
                 return
 
@@ -134,6 +205,7 @@ class _ViewerApplet:
 
         self._waiting_for_first_draw_message = True
         self._load_message = message
+        self._load_message_mesh_checksum = mesh_checksum
 
     def _build_links(self):
         # Make all of our (ViewerApplet's) geometry invisible so that the
--- a/bindings/pydrake/visualization/test/meldis_test.py
+++ b/bindings/pydrake/visualization/test/meldis_test.py
@@ -13,6 +13,9 @@ You can also visualize the LCM test data like so:
 import pydrake.visualization as mut
 
 import functools
+import hashlib
+import os
+from pathlib import Path
 import sys
 import unittest
 
@@ -24,6 +27,7 @@ from drake import (
     lcmt_viewer_draw,
     lcmt_viewer_geometry_data,
     lcmt_viewer_link_data,
+    lcmt_viewer_load_robot,
 )
 
 from pydrake.geometry import (
@@ -167,23 +171,32 @@ class TestMeldis(unittest.TestCase):
         self.assertEqual(meshcat.HasPath("/DRAKE_VIEWER"), True)
         self.assertEqual(meshcat.HasPath(link_path), True)
 
-    def test_viewer_applet_robot_meshes(self):
-        """Checks _ViewerApplet support for meshes.
+    def _check_viewer_applet_on_model(self, resource):
+        """Checks that _ViewerApplet doesn't crash on the given model file.
         """
-        # Create the device under test.
         dut = mut.Meldis()
         lcm = dut._lcm
-
-        # Process the load + draw messages.
         diagram = self._make_diagram(
-            resource="drake/manipulation/models/iiwa_description/urdf/"
-                     "iiwa14_no_collision.urdf",
+            resource=resource,
             visualizer_params=DrakeVisualizerParams(),
             lcm=lcm)
         diagram.ForcedPublish(diagram.CreateDefaultContext())
         lcm.HandleSubscriptions(timeout_millis=0)
         dut._invoke_subscriptions()
         self.assertEqual(dut.meshcat.HasPath("/DRAKE_VIEWER"), True)
+
+    def test_viewer_applet_plain_meshes(self):
+        """Checks _ViewerApplet support for untextured meshes.
+        """
+        self._check_viewer_applet_on_model(
+            "drake/manipulation/models/iiwa_description/urdf/"
+            "iiwa14_no_collision.urdf")
+
+    def test_viewer_applet_textured_meshes(self):
+        """Checks _ViewerApplet support for textured meshes.
+        """
+        self._check_viewer_applet_on_model(
+            "drake/manipulation/models/ycb/sdf/004_sugar_box.sdf")
 
     def test_viewer_applet_reload_optimization(self):
         """Checks that loading the identical scene twice is efficient.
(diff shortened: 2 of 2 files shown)


jwnimmer-tri:
+@SeanCurtis-TRI for feature review (or delegation), please.

SeanCurtis-TRI (commented):
There are some OBJ corner cases. I don't think we'll ever be able to make this logic 100% robust. The current effort is good for common cases (modulo texture data). It's certainly viable to argue that it needn't be perfect to provide value (more smart auto-loading is always good). But we should at least explicitly decide whether we care about the corner cases/textures and make some note somewhere (even if it's just here).

I'd still advocate a command-line parameter where a user can say, I know I have something weird, and i just want you to reload the whole model for every load command. This still has advantage over using a `Meshcat`  as it preserves lighting and camera from load to load.

:LGTM: (With the assumption that the general feeling will be that the corner cases and textures can be addressed down the road if at all.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-NaeAfOA3t1dYYi7Bcm4:bpmrrq4)* status: 3 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/visualization/_meldis.py` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae55A75BKtvyQN46EN:-Nae55A83X6E2GqG-DsB:b9eq5rr) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L135)):*
> ```Python
> 
>         # Hash the mtl file (if there is one).
>         match = re.search(rb"^\s*mtllib\s+(\S+)\s*$", mesh_content,
> ```

BTW My first thought was that file names can have spaces. My second thought is that anyone foolish enough to do so gets what they deserve. My third thought is that even if *we* were to make the herculean effort to accommodate such foollishness, I think the odds that any downstream consumer making a similar effort are slim to none.
___
*[`bindings/pydrake/visualization/_meldis.py` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae5ITc0X9Pqb7f4saL:-Nae5ITc0X9Pqb7f4saM:b-w6kp75) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L139)):*
> ```Python
>         if not match:
>             return
>         mtllib, = match.groups()
> ```

A valid obj can have multiple `mtllib` invocations -- you can draw materials from multiple sources.

A single mttlib line can list multiple files, and, I believe, a file can have multiple mtllib lines as well.
___
*[`bindings/pydrake/visualization/_meldis.py` line 146 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae8bkvDxAT_pvFMpQ4:-Nae8bkvDxAT_pvFMpQ5:b66pjpc) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L146)):*
> ```Python
>         except IOError:
>             return
>         hasher.update(mtl_content)
> ```

Worried about textures at all?

(Aren't we glad that, currently, we only support embedded .glTF files?)

jwnimmer-tri:
I'm not happy with the unsoundness here.  I'm going to pause and rethink the design.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878)* status: 3 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/visualization/_meldis.py` line 146 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae8bkvDxAT_pvFMpQ4:-NahiW-Z4hnIq5YnKpMs:b-swg6di) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L146)):*

Worried about textures at all?

(Aren't we glad that, currently, we only support embedded .glTF files?)
</blockquote></details>

Working

Yikes.  I definitely _do_ want to hash the image files, even in this first MVP.

jwnimmer-tri (commented):
-(status: do not review) 

> I'd still advocate a command-line parameter where a user can say, I know I have something weird, and i just want you to reload the whole model for every load command.

I think for this use case, hitting Ctrl-C and re-starting Meldis is probably fine.  (Or delete the "don't reload" early return line, by hacking the source code.)  In short -- I don't think this is a common enough use case to be worth supporting with any more documentation or code.  I want it to work correctly out of the box, without extra hijinks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-NaspjbS0T8_o7fU9ItB:b-mftyp0)* status: needs at least two assigned reviewers

___
*[`bindings/pydrake/visualization/_meldis.py` line 135 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae55A75BKtvyQN46EN:-Naspdn-3EpXtTavUsB2:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L135)):*

BTW My first thought was that file names can have spaces. My second thought is that anyone foolish enough to do so gets what they deserve. My third thought is that even if *we* were to make the herculean effort to accommodate such foollishness, I think the odds that any downstream consumer making a similar effort are slim to none.
</blockquote></details>

Done
___
*[`bindings/pydrake/visualization/_meldis.py` line 139 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae5ITc0X9Pqb7f4saL:-NaspeQ17-T4dNHOfYg5:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L139)):*

A valid obj can have multiple `mtllib` invocations -- you can draw materials from multiple sources.

A single mttlib line can list multiple files, and, I believe, a file can have multiple mtllib lines as well.
</blockquote></details>

Done
___
*[`bindings/pydrake/visualization/_meldis.py` line 146 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nae8bkvDxAT_pvFMpQ4:-NaspfB6Dyj6HpkokU5e:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/fad352b1e7dfc7e80ee1b9fa1ff6ed61d0d3b15a/bindings/pydrake/visualization/_meldis.py#L146)):*

Working

Yikes.  I definitely _do_ want to hash the image files, even in this first MVP.
</blockquote></details>

Done

jwnimmer-tri:
> ... reload the whole model for every load command.

Also on the UX front: if the reload is skipped (because we think its a no-op), the user _does_ get a console message letting them know.  So if they change a model, hit reload, and see that message, they will have a clue that our change-detection code isn't working for them -- and maybe let us know with a bug report.

SeanCurtis-TRI (commented):
I don't think we have feature coverage yet (as noted below: mtllib command and texture options). The test covers what the code explicitly attempts. So, do we attempt to extend what the code covers?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-NatlcXj6aSdRtS-Tu76:bw6rqyb)* status: 7 unresolved discussions, needs at least two assigned reviewers

___
*[`bindings/pydrake/visualization/_meldis.py` line 99 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatnKv26yBaQiQ5GuZ4:-NatnKv26yBaQiQ5GuZ5:b8a9ykt) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L99)):*
> ```Python
>     messages such as lcmt_viewer_load_robot or similar.
> 
>     Each "on_..." method incorporates the all external files cited by the given
> ```

nit: grammar



_Suggestion:_
```Python
all of the
```
___
*[`bindings/pydrake/visualization/_meldis.py` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Natn6jB1cRyWMVJ0J2-:-Natn6jB1cRyWMVJ0J20:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L101)):*
> ```Python
>     Each "on_..." method incorporates the all external files cited by the given
>     argument (of a specific type) into the current hash. Files that are named
>     but cannot be opened are sliently skipped.
> ```

nit: typo



_Suggestion:_
```Python
silently
```
___
*[`bindings/pydrake/visualization/_meldis.py` line 115 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatnA_C2YN9Kq7BFOld:-NatnA_C2YN9Kq7BFOle:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L115)):*
> ```Python
>         Returns the file content (as ``bytes``, not ``str``).
>         Remembers the filename (for unit testing).
>         If the file is missing, sliently returns an empty ``bytes``.
> ```

nit: typo



_Suggestion:_
```Python
silently
```
___
*[`bindings/pydrake/visualization/_meldis.py` line 127 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatngRd4_Z3agsH4zIi:-NatngRd4_Z3agsH4zIj:b-p1ong3) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L127)):*
> ```Python
> 
>     def on_viewer_load_robot(self, message: lcmt_viewer_load_robot):
>         assert isinstance(message, lcmt_viewer_load_robot)
> ```

This exposes my python ignorance. I would've thought that the type declaration on `message` would've rendered this assert unnecessary. Now, I have to go read up on this.
___
*[`bindings/pydrake/visualization/_meldis.py` line 141 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-Nato3eh9soJH3bNcvKW:-Nato3ei6WN09zlLvSkk:b263kqz) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L141)):*
> ```Python
>         assert isinstance(path, Path)
>         content = self._read_file(path)
>         for mtl_name in re.findall(rb"^\s*mtllib\s+(.*?)\s*$", content,
> ```

btw: Given we support more than .obj, should we condition this further work based on the extension?

While doing this work has a very low probability of causing problems, it does end up suggesting the code is only focused on .objs.
___
*[`bindings/pydrake/visualization/_meldis.py` line 141 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatoZXKEb-1JYk1lIAi:-NatoZXKEb-1JYk1lIAj:b-su1zh6) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L141)):*
> ```Python
>         assert isinstance(path, Path)
>         content = self._read_file(path)
>         for mtl_name in re.findall(rb"^\s*mtllib\s+(.*?)\s*$", content,
> ```

nit: As I read this regular expression, it will cover:

```
usemtl mat1.mtl
usemtl mat2.mtl
```

and cover the horrible case of:
```
usemtl c:\User Data\my.mtl
```

It doesn't cover
```
usemtl mat1.mtl mat2.mtl
```

as documented on [this page](http://paulbourke.net/dataformats/obj/).  (Search for "mtllib"; it's the third hit.)
___
*[`bindings/pydrake/visualization/_meldis.py` line 148 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatolnL4yZrbSy54CsZ:-NatolnL4yZrbSy54Cs_:bqghyqs) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L148)):*
> ```Python
>         assert isinstance(path, Path)
>         content = self._read_file(path)
>         for tex_name in re.findall(rb"^\s*map_\S+\s+(.*?)\s*$", content,
> ```

nit: This regular expression can easily mistake texture options for file names. See [wikipedia for examples](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Texture_options). And some of these options are very important in meshcat; meshcat's default bump map intensity is usually *way* too high.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878)* status: needs at least two assigned reviewers

___
*[`bindings/pydrake/visualization/_meldis.py` line 127 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatngRd4_Z3agsH4zIi:-NavrbLBCN_0aAI8owNE:b-axllxi) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L127)):*

This exposes my python ignorance. I would've thought that the type declaration on `message` would've rendered this assert unnecessary. Now, I have to go read up on this.
</blockquote></details>

By default, type annotations are just documentation.  Duck typing is still permitted at runtime.

SeanCurtis-TRI (commented):
:LGTM: +@sammy-tri for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-Naxe5Db7wKPOgu_nZPB:b6ccful)* status: LGTM missing from assignee sammy-tri(platform)

___
*[`bindings/pydrake/visualization/_meldis.py` line 127 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-NatngRd4_Z3agsH4zIi:-NaxdTS1AAmiF54k2O6r:bwevs7o) ([raw file](https://github.com/RobotLocomotion/drake/blob/c4976ff0f022f778c8600bcd8865a320bd161999/bindings/pydrake/visualization/_meldis.py#L127)):*

By default, type annotations are just documentation.  Duck typing is still permitted at runtime.
</blockquote></details>

That makes sense. I hadn't realized it was largely decorative.

SeanCurtis-TRI (commented):
Hmmm...at this time, east coast is pretty late. I think it's going to go to +@sherm1  on monday. -@sammy-tri .

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-NaxeDkUAvpY86CToMAl:b-7fxiot)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Platform :lgtm:

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19878#-:-NbFa9BY8njFt3g5dyDU:b8qx53o)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),sherm1(platform)

PR #19886, opened 2023-07-31
Adds solvers::GetVariableInSemidefiniteRelaxation()

RussTedrake:
@hongkai-dai , @bernhardpg , @lujieyang  -- WDYT of this as a generalizable solution to the SDP relaxation variable idea?

I toyed with returning e.g. a `std::map<std::pair<Variable, Variable>, Variable>` from the `MakeSemidefiniteRelaxation` method, but this feels unscalable to more complicated settings (e.g. higher moment relaxations).  We could imagine other implementations of this -- for instance we could assign unique variable names in the relaxation and look things up by name.  We might need to pass more information between the two methods in the future. But I think as an API this is close to what we want?

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19886)

### DIFF
--- a/bindings/pydrake/solvers/solvers_py_semidefinite_relaxation.cc
+++ b/bindings/pydrake/solvers/solvers_py_semidefinite_relaxation.cc
@@ -14,6 +14,11 @@ void DefineSolversSemidefiniteRelaxation(py::module m) {
 
   m.def("MakeSemidefiniteRelaxation", &solvers::MakeSemidefiniteRelaxation,
       py::arg("prog"), doc.MakeSemidefiniteRelaxation.doc);
+  
+  m.def("GetVariableInSemidefiniteRelaxation",
+      &solvers::GetVariableInSemidefiniteRelaxation,
+      py::arg("prog"), py::arg("relaxation"), py::arg("vars_in_prog"),
+      doc.GetVariableInSemidefiniteRelaxation.doc);
 }
 
 }  // namespace internal
--- a/solvers/semidefinite_relaxation.cc
+++ b/solvers/semidefinite_relaxation.cc
@@ -259,5 +259,45 @@ std::unique_ptr<MathematicalProgram> MakeSemidefiniteRelaxation(
   return relaxation;
 }
 
+symbolic::Variable GetVariableInSemidefiniteRelaxation(
+    const MathematicalProgram& prog, const MathematicalProgram& relaxation,
+    std::vector<symbolic::Variable> vars_in_prog) {
+  DRAKE_THROW_UNLESS(vars_in_prog.size() > 0);
+  // MakeSemidefiniteRelaxation only supports quadratic problems so far.
+  DRAKE_THROW_UNLESS(vars_in_prog.size() <= 2);
+  if (vars_in_prog.size() == 1) {
+    // The original variables are always available directly in the relaxation.
+    return vars_in_prog[0];
+  }
+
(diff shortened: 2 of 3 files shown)


hongkai-dai (commented):
BTW, another approach to handle the high-degree moment matrix is to have the mapping from the monomial to the variable, i.e., `std::unordered_map<symbolic::Monomial, symbolic::Variable>` which maps xᵃ to yₐ = ∫ xᵃdμ, where `a` is a list of integer degrees.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19886#-:-NahnD0p0OMVwIfowSpg:b8mmdgz)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

RussTedrake:
It still feels a bit wrong to me to always populate one big std::map (even in higher degrees) when users will likely only use a very small fraction of it, and when the return values can be expressed very simply as a function.  The user would have to call std::map::at() multiple times. Is it really better than calling the GetVariable method multiple times?

To answer a few of the specific questions:
- fwiw -- I did actually use SortedPair, not std::pair in my original implementation of this idea. 
- unordered_map might indeed be better here.  i've learned to fear it because iterating over that type leads to inconsistent results on different systems/compilers.  So I normally try to avoid it in the public API.
- I do really like the idea of using the symbolic::Monomial.  That's better than the std::vector version and SortedPair versions, I think.

I saw your moment issue come in.  Great!

bernhardpg (commented):
In my opinion an API like this is very useful. I agree that populating a std::map seems unnecessary. Having an easy way to retrieve the relaxed variables for higher order monomials would be useful, and I think the proposed API is a good first step.

I also like @hongkai-dai s idea on having a mapping from a monomial to relaxed variables. I know that in the SOSTools package, this is called the pseudo expectation operator, and they support taking the pseudo expectation of both monomials and polynomials, to return the corresponding polynomials in the moments. Could be worth taking a look at, see i.e. Example 5 [here](https://sums-of-squares.github.io/sos/) or the implementation of the pseudo expectation operator (here)(https://github.com/yuanchenyang/SumOfSquares.py/blob/master/src/SumOfSquares/SoS.py#L89).

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19886#-:-Nauctle41S5-v-kFxKT:b-zhq6ev)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

bernhardpg (commented):
Having an API that gives you the pseudo-expectation of polynomials would also address @lujieyang concern regarding multiple calls, in the case one wants to add this constraint _after_ creating the relaxation. However I think in most cases one would probably add the trace constraint as a quadratic equality constraint in the original program, and then create the relaxation after, which would then convert this to proper trace constraint and add it for you. Either way it could be useful to have the functionality to get the pseudo-expectation of a polynomial in the case one wants to add polynomial constraints after creating the relaxation, but want to do so using the original variables.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19886#-:-Nauhvtp8MvSUzXEF1Nf:b-4iahrw)* status: LGTM missing from assignee RussTedrake(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @RussTedrake)

RussTedrake:
Closing this draft; will track in #19996.

PR #19888, opened 2023-07-31
Enable Boundedness Checks for Spectrahedra and Intersections with Unbounded Components

### LINKED ISSUE #16663: Check boundedness of `Intersection` of convex sets
As discussed with @RussTedrake and @mpetersen94, currently the boundedness check for the `Intersection` of convex sets throws unless any of the intersected sets are bounded. But it can be that even if all of the sets are unbounded the intersection is bounded. Here I report the math (might need some cleaning) that I believe we should implement for this check.

- The problem of determining if the intersection of convex sets is bounded is equal to the problem of checking if a convex set of the form `S = {x: g_i(x) <= 0 for all i}` is bounded.
- We form the recession `K` cone of `S`. This will have the form `K = {x: h_i(x) <= 0 for all i}`. To construct the functions `h_i` from the `g_i` we will need a separate method for each `ConvexSet`. For example, if a set is a polyhedron, then `g_i(x) = a' x - b` and `h_i(x) = a' x`. For an `Ellipsoid` the cone `K` is the nullspace of the Hessian. Etc.
- The set `S` is bounded iff `K` contains a nonzero point.
- First we check if `K` is pointed. If not, then `S` is unbounded. One way to do this is to check if the linearizations of the `h_i(x)` at `x = 0` form a basis for `R^n`.
- If `K` is pointed then we have `h_i(x) = 0` for all `i` iff `x = 0`.
- Finally, we then check if `K` contains a point `x != 0` by solving the convex program: `find x, y` such that ` h_i(x) <= y for all i`, `y <= 0`, and `sum_k y_k <= - 1`.
- If `K` has a nonzero point `x` in it, then the problem above is feasible. This because `x != 0` implies  `h_i(x) < 0` for some `i`.
- If `K = {0}` then the above problem is infeasible.

Note: This is essentially the dual of the check we currently do to see if a polyhedron is bounded. For a polyhedron `{x: A x <= b}` the check of pointedness of `K` becomes `ker(A) = {0}`. And the convex program becomes the LP `find x, y` such that ` A x <= y`, `y <= 0`, and `sum_k y_k <= - 1`. When we implement the stuff above, I'd also modify the check for the polyhedron since I think this primal LP is much easier to interpret than the dual check we are currently doing.

cohnt:
This implements `DoIsBoundedShortcut()` for `ConvexSet`, and also modifies the implementations for `Intersection` and `Spectahedron` to use the generic base class method (after checking some edge cases).

Fixes #16663 (although potentially not in the most efficient way). +@russtedrake for feature review.

The implementation is slightly different than what I suggested in [this comment](https://github.com/RobotLocomotion/drake/issues/16663#issuecomment-1619017030) -- instead of maximizing and minimizing each dimension, I maximize the displacement between two feasible points along each dimension. This doubles the number of decision variables, but halves the number of optimization problems solved.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)

### DIFF
--- a/geometry/optimization/spectrahedron.cc
+++ b/geometry/optimization/spectrahedron.cc
@@ -76,9 +76,19 @@ std::unique_ptr<ConvexSet> Spectrahedron::DoClone() const {
   return std::make_unique<Spectrahedron>(*this);
 }
 
-bool Spectrahedron::DoIsBounded() const {
-  throw std::runtime_error(
-      "Spectrahedron::IsBounded() is not implemented yet.");
+std::optional<bool> Spectrahedron::DoIsBoundedShortcut() const {
+  // CSDP cannot handle an unconstrained spectrahedron (i.e. a set whose only
+  // constraints are positive semidefinite constraints), so we must check for
+  // that case explicitly.
+  auto constraints = sdp_->GetAllConstraints();
+  auto psd_constraints = sdp_->positive_semidefinite_constraints();
+  if (constraints.size() == psd_constraints.size()) {
+    // If there are no constraints, then the set is bounded iff its ambient
+    // dimension is zero. If the ambient dimension is zero, it's caught in
+    // IsBounded, so we know the set will be unbounded.
+    return false;
+  }
+  return std::nullopt;
 }
 
 bool Spectrahedron::DoPointInSet(const Eigen::Ref<const VectorXd>& x,
--- a/geometry/optimization/affine_subspace.cc
+++ b/geometry/optimization/affine_subspace.cc
@@ -36,7 +36,7 @@ std::unique_ptr<ConvexSet> AffineSubspace::DoClone() const {
   return std::make_unique<AffineSubspace>(*this);
 }
 
-bool AffineSubspace::DoIsBounded() const {
+std::optional<bool> AffineSubspace::DoIsBoundedShortcut() const {
   return basis_.cols() == 0;
 }
 
--- a/geometry/optimization/affine_subspace.h
+++ b/geometry/optimization/affine_subspace.h
@@ -62,7 +62,7 @@ class AffineSubspace final : public ConvexSet {
  private:
   std::unique_ptr<ConvexSet> DoClone() const final;
 
-  bool DoIsBounded() const final;
+  std::optional<bool> DoIsBoundedShortcut() const final;
 
   bool DoIsEmpty() const final;
 
--- a/geometry/optimization/cartesian_product.cc
+++ b/geometry/optimization/cartesian_product.cc
@@ -103,7 +103,7 @@ std::unique_ptr<ConvexSet> CartesianProduct::DoClone() const {
   return std::make_unique<CartesianProduct>(*this);
 }
 
-bool CartesianProduct::DoIsBounded() const {
+std::optional<bool> CartesianProduct::DoIsBoundedShortcut() const {
   // Note: The constructor enforces that A_ is full column rank.
   for (const auto& s : sets_) {
     if (!s->IsBounded()) {
--- a/geometry/optimization/cartesian_product.h
+++ b/geometry/optimization/cartesian_product.h
@@ -81,7 +81,7 @@ class CartesianProduct final : public ConvexSet {
  private:
   std::unique_ptr<ConvexSet> DoClone() const final;
 
-  bool DoIsBounded() const final;
+  std::optional<bool> DoIsBoundedShortcut() const final;
 
   bool DoIsEmpty() const final;
 
--- a/geometry/optimization/convex_set.cc
+++ b/geometry/optimization/convex_set.cc
@@ -11,6 +11,14 @@ namespace drake {
 namespace geometry {
 namespace optimization {
 
+using Eigen::VectorXd;
+using solvers::Binding;
+using solvers::Constraint;
+using solvers::LinearCost;
+using solvers::MathematicalProgram;
+using solvers::VariableRefList;
+using solvers::VectorXDecisionVariable;
+
 ConvexSet::ConvexSet(int ambient_dimension)
     : ambient_dimension_(ambient_dimension) {
   DRAKE_THROW_UNLESS(ambient_dimension >= 0);
@@ -29,6 +37,52 @@ bool ConvexSet::IntersectsWith(const ConvexSet& other) const {
   other.AddPointInSetConstraints(&prog, x);
   solvers::MathematicalProgramResult result = solvers::Solve(prog);
   return result.is_success();
+}
+
+bool ConvexSet::GenericDoIsBounded() const {
+  // The empty set is bounded. We check it first, to ensure that the program
+  // is feasible, so SolutionResult::kInfeasibleOrUnbounded or
+  // SolutionResult::kDualInfeasible indicates unbounded, as solvers may not
+  // always explicitly return that the primal is unbounded.
+  if (IsEmpty()) {
+    return true;
+  }
+  // Let a variable x be contained in the convex set. Iteratively try to
+  // minimize or maximize x[i] for each dimension i. If any solves are
+  // unbounded, the set is not bounded.
+  MathematicalProgram prog;
+  VectorXDecisionVariable x =
+      prog.NewContinuousVariables(ambient_dimension(), "x");
+  AddPointInSetConstraints(&prog, x);
+  Binding<LinearCost> objective =
+      prog.AddLinearCost(VectorXd::Zero(ambient_dimension()), x);
+
+  VectorXd objective_vector(ambient_dimension());
+  for (int i = 0; i < ambient_dimension(); ++i) {
+    objective_vector.setZero();
+    objective_vector[i] = 1;
+    objective.evaluator()->UpdateCoefficients(objective_vector);
+    const auto result = solvers::Solve(prog);
+    if (result.get_solution_result() == solvers::SolutionResult::kUnbounded ||
+        result.get_solution_result() ==
+            solvers::SolutionResult::kInfeasibleOrUnbounded ||
+        result.get_solution_result() ==
+            solvers::SolutionResult::kDualInfeasible) {
+      return false;
+    }
+
+    objective_vector[i] = -1;
+    objective.evaluator()->UpdateCoefficients(objective_vector);
+    const auto result2 = solvers::Solve(prog);
+    if (result2.get_solution_result() == solvers::SolutionResult::kUnbounded ||
+        result2.get_solution_result() ==
+            solvers::SolutionResult::kInfeasibleOrUnbounded ||
+        result2.get_solution_result() ==
+            solvers::SolutionResult::kDualInfeasible) {
+      return false;
+    }
+  }
+  return true;
 }
 
 bool ConvexSet::DoIsEmpty() const {
--- a/geometry/optimization/convex_set.h
+++ b/geometry/optimization/convex_set.h
@@ -81,13 +81,17 @@ class ConvexSet : public ShapeReifier {
   /** Returns true iff the set is bounded, e.g., there exists an element-wise
   finite lower and upper bound for the set.  Note: for some derived classes,
   this check is trivial, but for others it can require solving an (typically
-  small) optimization problem.  Check the derived class documentation for any
-  notes. When ambient_dimension is zero, always returns true. */
+  small) optimization problem. Check the derived class documentation for any
+  notes. */
   bool IsBounded() const {
     if (ambient_dimension() == 0) {
       return true;
     }
-    return DoIsBounded();
+    const auto shortcut_result = DoIsBoundedShortcut();
+    if (shortcut_result.has_value()) {
+      return shortcut_result.value();
+    }
+    return GenericDoIsBounded();
   }
 
   /** Returns true iff the set is empty. Note: for some derived classes, this
@@ -243,9 +247,13 @@ class ConvexSet : public ShapeReifier {
   /** Non-virtual interface implementation for Clone(). */
   virtual std::unique_ptr<ConvexSet> DoClone() const = 0;
 
-  /** Non-virtual interface implementation for IsBounded().
+  /** Non-virtual interface implementation for DoIsBoundedShortcut(). Trivially
+  returns std::nullopt. This allows a derived class to implement its own
+  boundedness checks, to potentially avoid the more expensive base class checks.
   @pre ambient_dimension() >= 0 */
-  virtual bool DoIsBounded() const = 0;
+  virtual std::optional<bool> DoIsBoundedShortcut() const {
+    return std::nullopt;
+  }
 
   /** Non-virtual interface implementation for IsEmpty(). The default
   implementation solves a feasibility optimization problem, but derived
@@ -330,6 +338,10 @@ class ConvexSet : public ShapeReifier {
       std::vector<solvers::Binding<solvers::Constraint>>* constraints) const;
 
  private:
+  /** Generic implementation for IsBounded() -- applicable for all convex sets.
+  @pre ambient_dimension() >= 0 */
+  bool GenericDoIsBounded() const;
+
   // The reset_after_move wrapper adjusts ConvexSet's default move constructor
   // and move assignment operator to set the ambient dimension of a moved-from
   // object back to zero. This is essential to keep the ambient dimension in
--- a/geometry/optimization/hpolyhedron.cc
+++ b/geometry/optimization/hpolyhedron.cc
@@ -341,7 +341,7 @@ HPolyhedron HPolyhedron::MakeL1Ball(const int dim) {
   return {A, b};
 }
 
-bool HPolyhedron::DoIsBounded() const {
+std::optional<bool> HPolyhedron::DoIsBoundedShortcut() const {
   if (A_.rows() < A_.cols()) {
     return false;
   }
--- a/geometry/optimization/hpolyhedron.h
+++ b/geometry/optimization/hpolyhedron.h
@@ -231,7 +231,7 @@ class HPolyhedron final : public ConvexSet {
 
   std::unique_ptr<ConvexSet> DoClone() const final;
 
-  bool DoIsBounded() const final;
+  std::optional<bool> DoIsBoundedShortcut() const final;
 
   bool DoIsEmpty() const final;
 
--- a/geometry/optimization/hyperellipsoid.cc
+++ b/geometry/optimization/hyperellipsoid.cc
@@ -256,7 +256,7 @@ std::unique_ptr<ConvexSet> Hyperellipsoid::DoClone() const {
   return std::make_unique<Hyperellipsoid>(*this);
 }
 
-bool Hyperellipsoid::DoIsBounded() const {
+std::optional<bool> Hyperellipsoid::DoIsBoundedShortcut() const {
   if (A_.rows() < A_.cols()) {
     return false;
   }
--- a/geometry/optimization/hyperellipsoid.h
+++ b/geometry/optimization/hyperellipsoid.h
@@ -132,7 +132,7 @@ class Hyperellipsoid final : public ConvexSet {
  private:
   std::unique_ptr<ConvexSet> DoClone() const final;
 
-  bool DoIsBounded() const final;
+  std::optional<bool> DoIsBoundedShortcut() const final;
 
   bool DoIsEmpty() const final;
 
--- a/geometry/optimization/intersection.cc
+++ b/geometry/optimization/intersection.cc
@@ -55,17 +55,13 @@ std::unique_ptr<ConvexSet> Intersection::DoClone() const {
   return std::make_unique<Intersection>(*this);
 }
 
-bool Intersection::DoIsBounded() const {
+std::optional<bool> Intersection::DoIsBoundedShortcut() const {
   for (const auto& s : sets_) {
     if (s->IsBounded()) {
       return true;
     }
   }
-  // TODO(mpetersen94): Cover case where sets are unbounded but intersection is
-  // bounded.
-  throw std::runtime_error(
-      "Determining the boundedness of an Intersection made up of unbounded "
-      "elements is not currently supported.");
+  return std::nullopt;
 }
 
 bool Intersection::DoIsEmpty() const {
--- a/geometry/optimization/intersection.h
+++ b/geometry/optimization/intersection.h
@@ -45,7 +45,7 @@ class Intersection final : public ConvexSet {
  private:
   std::unique_ptr<ConvexSet> DoClone() const final;
 
-  bool DoIsBounded() const final;
+  std::optional<bool> DoIsBoundedShortcut() const final;
 
   bool DoIsEmpty() const final;
 
--- a/geometry/optimization/minkowski_sum.cc
+++ b/geometry/optimization/minkowski_sum.cc
@@ -94,7 +94,7 @@ std::unique_ptr<ConvexSet> MinkowskiSum::DoClone() const {
   return std::make_unique<MinkowskiSum>(*this);
 }
 
-bool MinkowskiSum::DoIsBounded() const {
+std::optional<bool> MinkowskiSum::DoIsBoundedShortcut() const {
   for (const auto& s : sets_) {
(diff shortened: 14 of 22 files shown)


RussTedrake:
Let's ask +@TobiaMarcucci for a concept review, too... since he opened the issue and thought about it.

TobiaMarcucci:
I think that if at some point we'll move to a unified conic description of the the sets in the GCS this check should be easy (i.e. a single cone problem with size almost equal to the size of the set). Currently it's hard to find an elegant and simple way to check boundedness of any convex set. So I guess the question is when is a good time to change basic representation of the convex sets.

I think this approach is simple and useful for debugging, although as Tommy says not very efficient.

The idea of maximizing the displacement makes sense if setting up the problems is expensive. Otherwise solving two problems with half the variables is cheaper than solving one problem with double the variables. For example, say that solving a convex problem takes n^2 time, the first approach takes 2n^2, while the second approach takes (2n)^2=4n^2. Since you are only updating the cost coefficients between one problem and the next, I'm not sure which approach works best.

cohnt:
I like Tobia's idea of moving towards a unified conic description. But I expect that will take some time to land, and being able to check the boundedness of an intersection is already relevant to Drake users. For example, we'll need it in order to add a "debug mode" to GCS (which I'm sure @shaoyuancc would like to have).

This fix "works" (pending feature review, platform review, tests, etc.) now, and we can always replace it later with a more efficient implementation.

Re maximizing and minimizing each dimension, or maximizing the displacement, I don't have a preference between the two, and I can quickly switch the code around to do one or the other. Note that this code hopefully wouldn't be called very often. If all of the convex sets in an intersection are bounded, the result is returned as bounded without going into this logic.

TobiaMarcucci:
I agree. And especially if the interface won't change when we move to the conic representation, there is no damage in merging now a less performant way to do the check.

geometry/optimization/spectrahedron.cc
```diff
@@ -77,8 +77,18 @@ std::unique_ptr<ConvexSet> Spectrahedron::DoClone() const {
 }
 
 bool Spectrahedron::DoIsBounded() const {
-  throw std::runtime_error(
-      "Spectrahedron::IsBounded() is not implemented yet.");
+  // CSDP cannot handle an unconstrained spectrahedron (i.e. a set whose only
```
TobiaMarcucci, line 80: I'm not familiar with CSDP, but this seems strange? I would imagine that the important distinction is between bounded and unbounded, rather than constrained and unconstrained. For example, does CSDP accept something like `X >= 0` and `X[0,0] == 1`? (This would be constrained but unbounded.)

TobiaMarcucci (commented):
I've some observations, but overall it looks good!

geometry/optimization/spectrahedron.cc
cohnt: There's an error in either CSDP or the Drake interface to CSDP. It will give there error:
```
Problem must have at least one constraint.
unknown file: Failure
C++ exception with description "CsdpSolver: the CSDP library exited via a fatal exception" thrown in the test body.
```

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 4 unresolved discussions, LGTM missing from assignees TobiaMarcucci,RussTedrake(platform)

___
*[`geometry/optimization/convex_set.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19888#gh-1281551277:-NarP0k35X2R2pM7QHGo:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/36a147b931ca6f2feb779941ff24e0557afd7463/geometry/optimization/convex_set.cc#L49)):*

If there is no empirical performance reason, I would say that it is better to solve 2n problems with n variables rather than n problems with 2n variables.
</blockquote></details>

Done.
___
*[`geometry/optimization/spectrahedron.cc` line 85 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19888#gh-1281561709:-NarP20LEc9CLCkIlPSZ:bcit5j3) ([raw file](https://github.com/RobotLocomotion/drake/blob/36a147b931ca6f2feb779941ff24e0557afd7463/geometry/optimization/spectrahedron.cc#L85)):*

This check seems potentially brittle? What if by mistake I added twice the constraint that `X >= 0`? Would this check fail?
</blockquote></details>

This check is purely to handle the problem where CSDP refuses to consider problems without any constraints (besides the builtin PSD constraints). As long as the optimization problem can be run, it will be caught in `ConvexSet::DoIsBounded`. (And I've added a test case to specifically examine this.)
___
*[`geometry/optimization/test/spectrahedron_test.cc` line 325 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19888#gh-1281563086:-NarPNTx-s-MUljYsqjF:bcyaomw) ([raw file](https://github.com/RobotLocomotion/drake/blob/36a147b931ca6f2feb779941ff24e0557afd7463/geometry/optimization/test/spectrahedron_test.cc#L325)):*

By the way, do we think that some tests of `ConvexSet::DoIsBounded` belong to the `convex_set_test.cc` file?
</blockquote></details>

Just added a bunch of test cases to make it a little more robust. I'm not as familiar with the corner cases of SDPs, but if you have specific suggestions, I'm happy to add them.

It's hard to put `ConvexSet::DoIsBounded` tests in `convex_set_test.cc`, since all subclasses have their own implementations, and you can't construct a pure `ConvexSet`. `Intersection` and `Spectrahedron` are the only ones that ever call the base class version, so I think we have to test it through them.

RussTedrake (commented):
:lgtm: pending @TobiaMarcucci 's review.  Thanks!

> Reviewed 4 of 6 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-Nasy8vk0Dec0G0WfKhI:bydhifa)* status: 5 unresolved discussions, LGTM missing from assignee TobiaMarcucci

___
*[`geometry/optimization/convex_set.cc` line 61 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NasxC3n9-UEnTij9-PT:-NasxC3n9-UEnTij9-PU:b-fkxbx6) ([raw file](https://github.com/RobotLocomotion/drake/blob/c96dc4c87398ad86adad9015dcb63f59ba6bc4db/geometry/optimization/convex_set.cc#L61)):*
> ```cpp
>   VectorXd objective_vector(ambient_dimension());
>   for (int i = 0; i < ambient_dimension(); ++i) {
>     objective_vector = VectorXd::Zero(ambient_dimension());
> ```

btw -- you could just call `objective_vector.setZero()` here.
___
*[`geometry/optimization/spectrahedron.cc` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19888#gh-1281559423:-Nasxdea7X0m-84tbp1e:bg054yr) ([raw file](https://github.com/RobotLocomotion/drake/blob/36a147b931ca6f2feb779941ff24e0557afd7463/geometry/optimization/spectrahedron.cc#L80)):*

There's an error in either CSDP or the Drake interface to CSDP. It will give there error:
```
Problem must have at least one constraint.
unknown file: Failure
C++ exception with description "CsdpSolver: the CSDP library exited via a fatal exception" thrown in the test body.
```
</blockquote></details>

could we please open an issue for that with a minimal test case?

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 4 unresolved discussions, LGTM missing from assignee TobiaMarcucci

___
*[`geometry/optimization/convex_set.cc` line 61 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NasxC3n9-UEnTij9-PT:-NavxgWP2hsNo5czPjiz:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/c96dc4c87398ad86adad9015dcb63f59ba6bc4db/geometry/optimization/convex_set.cc#L61)):*

btw -- you could just call `objective_vector.setZero()` here.
</blockquote></details>

Done.
___
*[`geometry/optimization/spectrahedron.cc` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19888#gh-1281559423:-Navxf0V9f6-XU7_tdun:b-cc0k4) ([raw file](https://github.com/RobotLocomotion/drake/blob/36a147b931ca6f2feb779941ff24e0557afd7463/geometry/optimization/spectrahedron.cc#L80)):*

could we please open an issue for that with a minimal test case?
</blockquote></details>

Done. #19927

TobiaMarcucci (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-NaxtlYu4V0QehteTHNj:bnfp4nl)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,RussTedrake(platform)

RussTedrake (commented):

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-Nb9TIYX8buS3tOAGzhY:bppyoqq)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Platform :lgtm: with one suggestion to consider

> Reviewed 4 of 6 files at r1, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-NbFkZYW1S2-o26tThWp:bwi5r1m)* status: 2 unresolved discussions

___
*[`geometry/optimization/convex_set.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFfBPwDn3k2V-JzaAB:-NbFfBPx3D_ADcWW7A94:bu61gb5) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L42)):*
> ```cpp
> }
> 
> bool ConvexSet::DoIsBounded() const {
> ```

BTW it's awkward to have a base class virtual implementation that has to be invoked by derived classes for correct behavior. If I've understood the intent correctly, here's an alternate design sketch for your consideration:
```c++
bool ConvexSet::IsBounded() const {
  if (std::optional<bool> result = DoIsBoundedShortcut())
    return *result;
  return IsBoundedFullCheck();
}
virtual std::optional<bool> ConvexSet::DoIsBoundedShortcut() 
  { return std::nullopt; }
```
This has the advantage that derived classes only have to think about themselves -- do I have a quick test for boundedness? If not the base class takes responsibility for the test.
___
*[`geometry/optimization/convex_set.cc` line 50 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFchK3BrJSw-dVGb-M:-NbFchK40VlkJjdit2e9:bnz6g0c) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L50)):*
> ```cpp
>     return true;
>   }
>   // Iteratively try to maximize maximize the distance between x and y along
> ```

typo: maximize maximize -> maximize

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFfBPwDn3k2V-JzaAB:-NbFnUrqC3cFf366bzvy:b6uvnri) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L42)):*

BTW it's awkward to have a base class virtual implementation that has to be invoked by derived classes for correct behavior. If I've understood the intent correctly, here's an alternate design sketch for your consideration:
```c++
bool ConvexSet::IsBounded() const {
  if (std::optional<bool> result = DoIsBoundedShortcut())
    return *result;
  return IsBoundedFullCheck();
}
virtual std::optional<bool> ConvexSet::DoIsBoundedShortcut() 
  { return std::nullopt; }
```
This has the advantage that derived classes only have to think about themselves -- do I have a quick test for boundedness? If not the base class takes responsibility for the test.

</blockquote></details>

The way it's currently set up is that `ConvexSet::DoIsBounded` is valid for any derived class of `ConvexSet`, but most classes override it with their own, more efficient checks. The only classes that ever have to call the expensive check implemented here are `Intersection` and `Spectrahedron`, but even then we can sometimes avoid it. All tests still pass if you comment out the subclass-specific behavior in `Intersection` and `Spectrahedron` and just call `ConvexSet::DoIsBounded` right away. (Note that `Spectrahedron` wouldn't pass without Mosek, due to the issue with CSDP described in #19927. This edge case is handled in `Spectrahedron`'s implementation of `DoIsBounded`.)
___
*[`geometry/optimization/convex_set.cc` line 50 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFchK3BrJSw-dVGb-M:-NbFnTiAEFrVFyr_usaL:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L50)):*

typo: maximize maximize -> maximize
</blockquote></details>

Done.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 1 unresolved discussion

___
*[`geometry/optimization/convex_set.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFfBPwDn3k2V-JzaAB:-NbFsTqf52zRZDXvl8nH:bce5tdk) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L42)):*

The way it's currently set up is that `ConvexSet::DoIsBounded` is valid for any derived class of `ConvexSet`, but most classes override it with their own, more efficient checks. The only classes that ever have to call the expensive check implemented here are `Intersection` and `Spectrahedron`, but even then we can sometimes avoid it. All tests still pass if you comment out the subclass-specific behavior in `Intersection` and `Spectrahedron` and just call `ConvexSet::DoIsBounded` right away. (Note that `Spectrahedron` wouldn't pass without Mosek, due to the issue with CSDP described in #19927. This edge case is handled in `Spectrahedron`'s implementation of `DoIsBounded`.)
</blockquote></details>

That would still be true if you replace DoIsBounded() with DoIsBoundedShortcut() but the code would better reflect the intent, IMO. As written it took me a while to figure out what was going on (derived classes are given a chance to shortcut the IsBounded check if they know how, otherwise a default expensive method is used). You could also clarify that with better comments on the DoIsBounded() virtual, directed at implementers of derived classes. But to me that doesn't seem as good as making the code clearly match the intent, and leaving responsibility for the "no shortcut available" case to the base class.

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,sherm1(platform),RussTedrake(platform)

___
*[`geometry/optimization/convex_set.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFfBPwDn3k2V-JzaAB:-NbG1QhT6lpV2OZ10QoH:b-yyu82i) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L42)):*

That would still be true if you replace DoIsBounded() with DoIsBoundedShortcut() but the code would better reflect the intent, IMO. As written it took me a while to figure out what was going on (derived classes are given a chance to shortcut the IsBounded check if they know how, otherwise a default expensive method is used). You could also clarify that with better comments on the DoIsBounded() virtual, directed at implementers of derived classes. But to me that doesn't seem as good as making the code clearly match the intent, and leaving responsibility for the "no shortcut available" case to the base class. 
</blockquote></details>

Done.

I also added some further comments on `ConvexSet::IsBounded` to clarify this behavior, and pointed them to `Intersection` and `Spectrahedron` as examples of how to use this.

Will this require another round of feature review?

sherm1 (commented):
> Reviewed 1 of 1 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 3 unresolved discussions

___
*[`geometry/optimization/convex_set.h` line 89 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGFjwy0uVNyGvdY8ld:-NbGFjwy0uVNyGvdY8le:b-igkynf) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.h#L89)):*
minor: this is a great comment, but out of place here in the public Doxygen since it is targeted to an implementer rather than an end user. Should move to the protected DoIsBoundedShortcut() since that's what the derived class writers will have to implement.

I am confused now by this comment:
> Derived classes may also use the generic base-class implementation of
  DoIsBounded. 

The derived classes should either provide a shortcut or not, right? If not then the _base class_ is responsible for the generic boundedness check, not the derived class.
___
*[`geometry/optimization/convex_set.h` line 263 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGGLf1CIYShbTjdzDR:-NbGGLf1CIYShbTjdzDS:b-djx195) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.h#L263)):*
> ```objc
>   /** Non-virtual interface implementation for IsBounded().
>   @pre ambient_dimension() >= 0 */
>   virtual bool DoIsBounded() const;
> ```

I'm confused now why this is still a virtual method. It should only be invoked by the base class so should just be a private function.

I think this should just be renamed to something like `GenericIsBoundedCheck()` and made private. Then any derived classes that previously overrode DoIsBounded() should override DoIsBoundedShortcut() instead.

I was hoping to simplify the implementation not make it more complicated -- it's not an improvement to have two ways to do the same thing.

BTW it will be easy to find the places that need changing since they will fail to compile as soon as DoIsBounded() gets a new name.
___
*[`geometry/optimization/convex_set.cc` line 42 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbFfBPwDn3k2V-JzaAB:-NbGHPuB2SIkgCmO57w2:b-rn5e2f) ([raw file](https://github.com/RobotLocomotion/drake/blob/cea99e45cd347755b7abfa2d1b0f1f1cde52518d/geometry/optimization/convex_set.cc#L42)):*
> Will this require another round of feature review?

Should be a straightforward translation that won't need more feature review. But that assumes virtual DoIsBoundedShortcut() _replaces_ the DoIsBounded() virtual. If they both exist as virtuals then the logic is too mysterious for me!
___
*[`geometry/optimization/convex_set.cc` line 50 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGI2F1ByjXzba-yeAW:-NbGI2F22hycqem8l7Jg:b-y7xdr5) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.cc#L50)):*
> ```cpp
>     return true;
>   }
>   // Iteratively try to minimize or maximize the an x in the convex set for
> ```

minor: this sentence doesn't scan any more -- some words missing?

sherm1 (commented):
:lgtm_cancel: temporarily for some logic untangling

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-NbGK4xM9bRbyHSgs3c3:b3p57l0)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform)

cohnt (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform)

___
*[`geometry/optimization/convex_set.h` line 89 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGFjwy0uVNyGvdY8ld:-NbGWJ2h3sZI-Ai2FLI4:b-hlhvct) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.h#L89)):*

minor: this is a great comment, but out of place here in the public Doxygen since it is targeted to an implementer rather than an end user. Should move to the protected DoIsBoundedShortcut() since that's what the derived class writers will have to implement.

I am confused now by this comment:
> Derived classes may also use the generic base-class implementation of
  DoIsBounded. 

The derived classes should either provide a shortcut or not, right? If not then the _base class_ is responsible for the generic boundedness check, not the derived class.
</blockquote></details>

Moved to the protected `DoIsBoundedShortcut`, and removed the note about `DoIsBounded`, since I've fully replaced it with `DoIsBoundedShortcut`.
___
*[`geometry/optimization/convex_set.h` line 263 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGGLf1CIYShbTjdzDR:-NbGWSWE2Y9Arv5FUV9X:b-77qsl0) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.h#L263)):*

I'm confused now why this is still a virtual method. It should only be invoked by the base class so should just be a private function.

I think this should just be renamed to something like `GenericIsBoundedCheck()` and made private. Then any derived classes that previously overrode DoIsBounded() should override DoIsBoundedShortcut() instead.

I was hoping to simplify the implementation not make it more complicated -- it's not an improvement to have two ways to do the same thing.

BTW it will be easy to find the places that need changing since they will fail to compile as soon as DoIsBounded() gets a new name.
</blockquote></details>

I've created a unified `DoIsBoundedShortcut` that all subclasses of `ConvexSet` use, and renamed this to `GenericIsBoundedCheck` and moved it to private.

Derived classes which had their own boundedness checks now use `DoIsBoundedShortcut`, and simply always take the shortcut.
___
*[`geometry/optimization/convex_set.cc` line 50 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-NbGI2F1ByjXzba-yeAW:-NbGWwNAC-2t5rX8rt_v:b-8dwh3h) ([raw file](https://github.com/RobotLocomotion/drake/blob/53a6c5e23f8f3df1ee0b5355d02228165fe313dc/geometry/optimization/convex_set.cc#L50)):*

minor: this sentence doesn't scan any more -- some words missing?
</blockquote></details>

Should be clearer now.

sherm1 (commented):
Beautiful! Thanks for your patience, Thomas. Platform :lgtm:.

> Reviewed 15 of 16 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19888#-:-NbGiX9m8S7Y9cmASLan:bwup0qi)* status: :shipit: complete! all discussions resolved, LGTM from assignees TobiaMarcucci,sherm1(platform),RussTedrake(platform)

cohnt:
@sherm1 this failed the CI, are you sure it was okay to merge it? The issue was somewhere in `//bindings/pydrake/visualization:py/model_visualizer_test`, which my code shouldn't have touched at all. (It also builds correctly on my machine, but I'm not on Mac.) See: https://drake-jenkins.csail.mit.edu/job/mac-arm-ventura-clang-bazel-experimental-release/1559/consoleFull

sherm1:
I saw that but interpreted it as a Mac flake since it was so remote from this work (and Mac CI is not required for merging). It might cause trouble later but I think not.

cohnt:
Got it! Just checking to make sure

Issue #19882, opened 2023-07-30
Remove dependency on PETSc


xuchenhan-tri:
We currently use PETSc for solving the linear system arising from the free motion momentum balance equation for FEM based deformable body simulations. We also depend on PETSc for performing the Schur complement calculation described in [Han et al., 2023]. The performance of PETSc for these computation, which is often the bottleneck of a deformable body simulation, is not as competitive as our bespoke solver described in the paper. We should switch to using the bespoke solver and subsequently remove dependency on PETSc.

[Han et al., 2023] Han, Xuchen, Joseph Masterjohn, and Alejandro Castro. "A Convex Formulation of Frictional Contact between Rigid and Deformable Bodies." arXiv preprint arXiv:2303.08912 (2023).

jwnimmer-tri:
I think maybe #20135 solved this issue?

PR #19890, opened 2023-07-31
[workspace] Deprecate the petsc external

jwnimmer-tri:
We anticipate no longer using this soon, so it's time to start the clock for its removal.

+@xuchenhan-tri for both reviews, please.

Towards #17231 and #19882.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19890)

### DIFF
--- a/multibody/fem/BUILD.bazel
+++ b/multibody/fem/BUILD.bazel
@@ -404,7 +404,7 @@ drake_cc_library(
     ],
     deps = [
         "//common:unused",
-        "@petsc",
+        "@petsc//:petsc_nowarn",
     ],
 )
 
--- a/tools/workspace/default.bzl
+++ b/tools/workspace/default.bzl
@@ -245,6 +245,8 @@ def add_default_repositories(excludes = [], mirrors = DEFAULT_MIRRORS):
     if "osqp_internal" not in excludes:
         osqp_internal_repository(name = "osqp_internal", mirrors = mirrors)
     if "petsc" not in excludes:
+        # The @petsc external is deprecated in Drake's WORKSPACE and will be
+        # removed on or after 2023-11-01.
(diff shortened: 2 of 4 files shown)


xuchenhan-tri (commented):
To check my understanding: this has no observable effect when linking against PETSc internally (other than switching the spelling to `petsc_nowarn`) and we can remove the dependency entirely when we don't even need PETSc for internal use. Is that right?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19890#-:-NagxnWW5xtM3tNg7ENI:bd4h0qx)* status: LGTM missing from assignee xuchenhan-tri(platform)

jwnimmer-tri:
> This has no observable effect when linking against PETSc internally (other than switching the spelling to petsc_nowarn) ...

Yes.  There is no functional effect.  The build system will run a few extra commands due to the "nowarn" shim, but nothing important.

> ... we can remove the dependency entirely when we don't even need PETSc for internal use.

We can delete `tools/workspace/petsc` after both conditions are satisfied:
(1) FEM no longer uses it, and
(2) the deprecation date has passed.

As of (1) we can drop  the `deps` line in the fem BUILD file, but we need to keep the `tools/workspace` around until the deprecation date has passed.  (Workspace rules are Stable API, unless marked `internal`.)

xuchenhan-tri (commented):
Understood. Thanks for the explanation.

:lgtm: x2

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19890#-:-NahBwFv4Ss3dL_lrej3:brwirsu)* status: :shipit: complete! all discussions resolved, LGTM from assignee xuchenhan-tri(platform)

PR #19897, opened 2023-08-01
Support collision filtering for deformable geometries

xuchenhan-tri:
Closes #19854

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897)

### DIFF
--- a/geometry/collision_filter_declaration.h
+++ b/geometry/collision_filter_declaration.h
@@ -16,6 +16,18 @@ class CollisionFilter;
 }  // namespace internal
 #endif
 
+/** Enum that defines the scope of the geometries that are affected by the
+ collision filtering mechanism. */
+enum class CollisionFilterScope {
+  /** All geometries are considered when collision filters are applied. */
+  kAll,
+  /** Deformable geometries are omitted when applying collision filters. That
+   means that all deformable geometries are not affected by the collision filter
+   declaration even if they are included in the GeometrySet when the filter is
+   declared. */
+  kOmitDeformable
+};
+
 /** Class for articulating changes to the configuration of SceneGraph's
  "collision filters"; collision filters limit the scope of various proximity
  queries.
@@ -49,6 +61,9 @@ class CollisionFilterDeclaration {
   DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(CollisionFilterDeclaration)
 
   CollisionFilterDeclaration() = default;
+
+  explicit CollisionFilterDeclaration(CollisionFilterScope scope)
+      : scope_(scope) {}
 
   /** @name  Allowing pairs in consideration (removing collision filters)
 
@@ -159,8 +174,13 @@ class CollisionFilterDeclaration {
   // it from the declaration implementation.
   const std::vector<Statement>& statements() const { return statements_; }
 
+  CollisionFilterScope scope() const { return scope_; }
+
   // The sequence of statements in this declaration.
   std::vector<Statement> statements_;
+
+  // Defines the scope of the filter. Defaults to all geometries.
+  CollisionFilterScope scope_{CollisionFilterScope::kAll};
 };
 
 }  // namespace geometry
--- a/geometry/geometry_state.cc
+++ b/geometry/geometry_state.cc
@@ -242,7 +242,7 @@ std::vector<GeometryId> GeometryState<T>::GetAllGeometryIds() const {
 template <typename T>
 unordered_set<GeometryId> GeometryState<T>::GetGeometryIds(
       const GeometrySet& geometry_set, const std::optional<Role>& role) const {
-  return CollectIds(geometry_set, role);
+  return CollectIds(geometry_set, role, CollisionFilterScope::kAll);
 }
 
 template <typename T>
@@ -993,11 +993,13 @@ void GeometryState<T>::AssignRole(SourceId source_id, GeometryId geometry_id,
       ids_for_filtering.Add(geometry.frame_id());
       // Apply collision filter between geometry id and any geometries that have
       // been identified. If none have been identified, this makes no changes.
+      // Per public documentation of SceneGraph, we exclude deformable
+      // geometries and only filter among rigid geometries.
       geometry_engine_->collision_filter().Apply(
-          CollisionFilterDeclaration().ExcludeBetween(GeometrySet(geometry_id),
-                                                      ids_for_filtering),
-          [this](const GeometrySet& set) {
-            return this->CollectIds(set, Role::kProximity);
+          CollisionFilterDeclaration(CollisionFilterScope::kOmitDeformable)
+              .ExcludeBetween(GeometrySet(geometry_id), ids_for_filtering),
+          [this](const GeometrySet& set, CollisionFilterScope scope) {
+            return this->CollectIds(set, Role::kProximity, scope);
           },
           true /* is_invariant */);
     } break;
@@ -1275,13 +1277,20 @@ GeometryState<T>::ToAutoDiffXd() const {
 
 template <typename T>
 unordered_set<GeometryId> GeometryState<T>::CollectIds(
-    const GeometrySet& geometry_set, std::optional<Role> role) const {
+    const GeometrySet& geometry_set, std::optional<Role> role,
+    CollisionFilterScope scope) const {
+  auto must_include = [scope](const InternalGeometry& g,
+                              const std::optional<Role>& r) {
+    // Must have compatible role and be part of the scope.
+    return (!r.has_value() || g.has_role(*r)) &&
+           (scope == CollisionFilterScope::kAll || !g.is_deformable());
+  };
   unordered_set<GeometryId> resultant_ids;
   for (auto frame_id : geometry_set.frames()) {
     const auto& frame = GetValueOrThrow(frame_id, frames_);
     for (auto geometry_id : frame.child_geometries()) {
       const InternalGeometry& geometry = geometries_.at(geometry_id);
-      if (!role.has_value() || geometry.has_role(*role)) {
+      if (must_include(geometry, role)) {
         resultant_ids.insert(geometry_id);
       }
     }
@@ -1295,10 +1304,11 @@ unordered_set<GeometryId> GeometryState<T>::CollectIds(
           "SceneGraph: " +
           to_string(geometry_id));
     }
-    if (!role.has_value() || geometry->has_role(*role)) {
+    if (must_include(*geometry, role)) {
       resultant_ids.insert(geometry_id);
     }
   }
+
   return resultant_ids;
 }
 
--- a/geometry/geometry_state.h
+++ b/geometry/geometry_state.h
@@ -515,8 +515,9 @@ class GeometryState {
   CollisionFilterManager collision_filter_manager() {
     geometry_version_.modify_proximity();
     return CollisionFilterManager(
-        &geometry_engine_->collision_filter(), [this](const GeometrySet& set) {
-          return this->CollectIds(set, Role::kProximity);
+        &geometry_engine_->collision_filter(),
+        [this](const GeometrySet& set, CollisionFilterScope scope) {
+          return this->CollectIds(set, Role::kProximity, scope);
         });
   }
 
@@ -632,9 +633,11 @@ class GeometryState {
   // the set that can't be mapped to known geometries or frames will cause an
   // exception to be thrown. The ids can be optionally filtered based on role.
   // If `role` is nullopt, no filtering takes place. Otherwise, just those
-  // geometries with the given role will be returned.
+  // geometries with the given role will be returned. Deformable geometries are
+  // excluded from the results if `scope == kOmitDeformable`.
   std::unordered_set<GeometryId> CollectIds(const GeometrySet& geometry_set,
-                                            std::optional<Role> role) const;
+                                            std::optional<Role> role,
+                                            CollisionFilterScope scope) const;
 
   // Sets the kinematic poses for the frames indicated by the given ids.
   // @param[in]  poses           The frame id and pose values.
--- a/geometry/proximity/BUILD.bazel
+++ b/geometry/proximity/BUILD.bazel
@@ -183,6 +183,7 @@ drake_cc_library(
     srcs = ["deformable_contact_internal.cc"],
     hdrs = ["deformable_contact_internal.h"],
     deps = [
+        ":collision_filter",
         ":deformable_mesh_intersection",
         "//common:essential",
         "//geometry:geometry_ids",
--- a/geometry/proximity/collision_filter.cc
+++ b/geometry/proximity/collision_filter.cc
@@ -152,11 +152,11 @@ bool CollisionFilter::CanCollideWith(GeometryId id_A, GeometryId id_B) const {
 
 void CollisionFilter::AddFiltersBetween(
     const GeometrySet& set_A, const GeometrySet& set_B,
-    const CollisionFilter::ExtractIds& extract_ids, bool is_invariant,
-    FilterState* state_out) {
-  const std::unordered_set<GeometryId> ids_A = extract_ids(set_A);
+    const CollisionFilter::ExtractIds& extract_ids, CollisionFilterScope scope,
+    bool is_invariant, FilterState* state_out) {
+  const std::unordered_set<GeometryId> ids_A = extract_ids(set_A, scope);
   const std::unordered_set<GeometryId>& ids_B =
-      &set_A == &set_B ? ids_A : extract_ids(set_B);
+      &set_A == &set_B ? ids_A : extract_ids(set_B, scope);
   for (GeometryId id_A : ids_A) {
     for (GeometryId id_B : ids_B) {
       AddFilteredPair(id_A, id_B, is_invariant, state_out);
@@ -166,10 +166,11 @@ void CollisionFilter::AddFiltersBetween(
 
 void CollisionFilter::RemoveFiltersBetween(
     const GeometrySet& set_A, const GeometrySet& set_B,
-    const CollisionFilter::ExtractIds& extract_ids, FilterState* state_out) {
-  const std::unordered_set<GeometryId> ids_A = extract_ids(set_A);
+    const CollisionFilter::ExtractIds& extract_ids, CollisionFilterScope scope,
+    FilterState* state_out) {
+  const std::unordered_set<GeometryId> ids_A = extract_ids(set_A, scope);
   const std::unordered_set<GeometryId>& ids_B =
-      &set_A == &set_B ? ids_A : extract_ids(set_B);
+      &set_A == &set_B ? ids_A : extract_ids(set_B, scope);
   for (GeometryId id_A : ids_A) {
     for (GeometryId id_B : ids_B) {
       RemoveFilteredPair(id_A, id_B, state_out);
@@ -233,6 +234,7 @@ void CollisionFilter::Apply(const CollisionFilterDeclaration& declaration,
                             const CollisionFilter::ExtractIds& extract_ids,
                             bool is_invariant, FilterState* filter_state) {
   using Operation = CollisionFilterDeclaration::StatementOp;
+  const CollisionFilterScope scope = declaration.scope();
   for (const auto& statement : declaration.statements()) {
     switch (statement.operation) {
       case Operation::kAllowBetween:
@@ -240,19 +242,19 @@ void CollisionFilter::Apply(const CollisionFilterDeclaration& declaration,
         // while removing collision filters.
         DRAKE_DEMAND(!is_invariant);
         RemoveFiltersBetween(statement.set_A, statement.set_B, extract_ids,
-                             filter_state);
+                             scope, filter_state);
         break;
       case Operation::kAllowWithin:
         DRAKE_DEMAND(!is_invariant);
         RemoveFiltersBetween(statement.set_A, statement.set_A, extract_ids,
-                             filter_state);
+                             scope, filter_state);
         break;
       case Operation::kExcludeWithin:
-        AddFiltersBetween(statement.set_A, statement.set_A, extract_ids,
+        AddFiltersBetween(statement.set_A, statement.set_A, extract_ids, scope,
                           is_invariant, filter_state);
         break;
       case Operation::kExcludeBetween:
-        AddFiltersBetween(statement.set_A, statement.set_B, extract_ids,
+        AddFiltersBetween(statement.set_A, statement.set_B, extract_ids, scope,
                           is_invariant, filter_state);
         break;
     }
--- a/geometry/proximity/collision_filter.h
+++ b/geometry/proximity/collision_filter.h
@@ -33,8 +33,8 @@ class CollisionFilter {
 
    This callback should throw if the GeometrySet contains FrameId values that
    cannot be mapped to GeometryIds. */
-  using ExtractIds =
-      std::function<std::unordered_set<GeometryId>(const GeometrySet&)>;
+  using ExtractIds = std::function<std::unordered_set<GeometryId>(
+      const GeometrySet&, CollisionFilterScope)>;
 
   /* Applies the collision filter declaration to the persistent state. The
    callback `extract_ids` provides a means to convert the GeometrySet into an
@@ -188,9 +188,10 @@ class CollisionFilter {
   /* Removes the geometry with the given `id` from the given filter state. */
   static void RemoveGeometry(GeometryId id, FilterState* filter_state_out);
 
-  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A, id_B ∈ set_B` to be
-   filtered. For each pair, if they are already filtered, no discernible change
-   is made.
+  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A*, id_B ∈ set_B*` to be
+   filtered, where `set_A*` and `set_B*` are the geometry ids in `set_A` and
+   `set_B`, respectively, consistent with the given `scope`. For each
+   pair, if they are already filtered, no discernible change is made.
 
    The filtered pair can be made "invariant" such that subsequent calls to
    RemoveFiltersBetween will not remove the filter. This is intended to support
@@ -202,16 +203,20 @@ class CollisionFilter {
   static void AddFiltersBetween(const GeometrySet& set_A,
                                 const GeometrySet& set_B,
                                 const ExtractIds& extract_ids,
+                                CollisionFilterScope scope,
                                 bool is_invariant, FilterState* state_out);
 
-  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A, id_B ∈ set_B` to be
-   unfiltered (if the filter isn't invariant). For each pair, if they are
-   already unfiltered, no discernible change is made.
+  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A*, id_B ∈ set_B*` to be
+   unfiltered (if the filter isn't invariant) where `set_A*` and `set_B*` are
+   the geometry ids in `set_A` and `set_B`, respectively, consistent with the
+   given `scope`. For each pair, if they are already unfiltered, no discernible
+   change is made.
 
    @pre All ids `id_A` and `id_B` are part of the system.  */
   static void RemoveFiltersBetween(const GeometrySet& set_A,
                                    const GeometrySet& set_B,
                                    const ExtractIds& extract_ids,
+                                   CollisionFilterScope scope,
                                    FilterState* state_out);
 
   /* Atomic operation in support of AddFiltersBetween().  */
--- a/geometry/proximity/deformable_contact_internal.cc
+++ b/geometry/proximity/deformable_contact_internal.cc
@@ -53,7 +53,8 @@ void Geometries::UpdateDeformableVertexPositions(
   }
 }
 
-DeformableContact<double> Geometries::ComputeDeformableContact() const {
+DeformableContact<double> Geometries::ComputeDeformableContact(
+    const CollisionFilter& collision_filter) const {
   DeformableContact<double> result;
   for (const auto& [deformable_id, deformable_geometry] :
        deformable_geometries_) {
@@ -61,13 +62,18 @@ DeformableContact<double> Geometries::ComputeDeformableContact() const {
         deformable_geometry.deformable_mesh().mesh();
     result.RegisterDeformableGeometry(
         deformable_id, deformable_mesh.num_vertices());
+    DRAKE_ASSERT(collision_filter.HasGeometry(deformable_id));
     for (const auto& [rigid_id, rigid_geometry] : rigid_geometries_) {
-      const math::RigidTransform<double>& X_WR = rigid_geometry.pose_in_world();
-      const auto& rigid_bvh = rigid_geometry.rigid_mesh().bvh();
-      const auto& rigid_tri_mesh = rigid_geometry.rigid_mesh().mesh();
-      AddDeformableRigidContactSurface(deformable_geometry, deformable_id,
-                                       rigid_id, rigid_tri_mesh, rigid_bvh,
-                                       X_WR, &result);
+      DRAKE_ASSERT(collision_filter.HasGeometry(rigid_id));
+      if (collision_filter.CanCollideWith(deformable_id, rigid_id)) {
+        const math::RigidTransform<double>& X_WR =
+            rigid_geometry.pose_in_world();
+        const auto& rigid_bvh = rigid_geometry.rigid_mesh().bvh();
+        const auto& rigid_tri_mesh = rigid_geometry.rigid_mesh().mesh();
+        AddDeformableRigidContactSurface(deformable_geometry, deformable_id,
+                                         rigid_id, rigid_tri_mesh, rigid_bvh,
+                                         X_WR, &result);
+      }
     }
   }
   return result;
--- a/geometry/proximity/deformable_contact_internal.h
+++ b/geometry/proximity/deformable_contact_internal.h
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "drake/geometry/geometry_ids.h"
+#include "drake/geometry/proximity/collision_filter.h"
 #include "drake/geometry/proximity/deformable_contact_geometries.h"
 #include "drake/geometry/proximity/volume_mesh.h"
 #include "drake/geometry/query_results/deformable_contact.h"
@@ -103,7 +104,8 @@ class Geometries final : public ShapeReifier {
    with respect to all registered rigid geometries. Assumes the vertex positions
    and poses of all registered deformable and rigid geometries are up to date.
   */
-  DeformableContact<double> ComputeDeformableContact() const;
+  DeformableContact<double> ComputeDeformableContact(
+      const CollisionFilter& collision_filter) const;
 
  private:
   friend class GeometriesTester;
--- a/geometry/proximity/test/collision_filter_test.cc
+++ b/geometry/proximity/test/collision_filter_test.cc
@@ -12,7 +12,8 @@ namespace geometry {
  ids to support the tests below. */
 class GeometrySetTester {
  public:
-  static std::unordered_set<GeometryId> geometries(const GeometrySet& s) {
+  static std::unordered_set<GeometryId> geometries(const GeometrySet& s,
+                                                   CollisionFilterScope) {
     return s.geometries();
   }
 };
--- a/geometry/proximity/test/collisions_exist_callback_test.cc
+++ b/geometry/proximity/test/collisions_exist_callback_test.cc
@@ -106,7 +106,8 @@ GTEST_TEST(CollisionsExistCallback, RespectsCollisionFilter) {
   // Filter the pair (A, B); we'll put the ids in a set and simply return that
   // set for the extract ids function.
   std::unordered_set<GeometryId> ids{data_A.id(), data_B.id()};
-  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                               CollisionFilterScope) {
     return ids;
   };
   collision_filter.Apply(CollisionFilterDeclaration().ExcludeWithin(
--- a/geometry/proximity/test/deformable_contact_internal_test.cc
+++ b/geometry/proximity/test/deformable_contact_internal_test.cc
@@ -12,6 +12,17 @@
 
 namespace drake {
 namespace geometry {
+
+/* Use GeometrySetTester's friend status with GeometrySet to leak its geometry
+ ids to support the tests below. */
+class GeometrySetTester {
+ public:
+  static std::unordered_set<GeometryId> geometries(const GeometrySet& s,
+                                                   CollisionFilterScope) {
+    return s.geometries();
+  }
+};
+
 namespace internal {
 namespace deformable {
 
@@ -60,6 +71,11 @@ ProximityProperties MakeProximityPropsWithRezHint(double resolution_hint) {
 math::RigidTransformd default_pose() {
   return math::RigidTransformd(math::RollPitchYaw<double>(1, 2, 3),
                                Vector3d(4, 5, 6));
+}
+
+/* Returns a value for the collision filter id extraction functor. */
+static CollisionFilter::ExtractIds get_extract_ids_functor() {
+  return &GeometrySetTester::geometries;
 }
 
 GTEST_TEST(GeometriesTest, AddRigidGeometry) {
@@ -309,9 +325,10 @@ GTEST_TEST(GeometriesTest, UpdateDeformableVertexPositions) {
 
 GTEST_TEST(GeometriesTest, ComputeDeformableContact) {
   Geometries geometries;
+  CollisionFilter collision_filter;
   /* The contact data is empty when there is no deformable geometry. */
   DeformableContact<double> contact_data =
-      geometries.ComputeDeformableContact();
+      geometries.ComputeDeformableContact(collision_filter);
   EXPECT_EQ(contact_data.contact_surfaces().size(), 0);
 
   /* Add a deformable unit cube. */
@@ -320,9 +337,10 @@ GTEST_TEST(GeometriesTest, ComputeDeformableContact) {
       MakeBoxVolumeMesh<double>(Box::MakeCube(1.0), 1.0);
   const int num_vertices = deformable_mesh.num_vertices();
   geometries.AddDeformableGeometry(deformable_id, std::move(deformable_mesh));
+  collision_filter.AddGeometry(deformable_id);
 
   /* There is no geometry to collide with the deformable geometry yet. */
-  contact_data = geometries.ComputeDeformableContact();
+  contact_data = geometries.ComputeDeformableContact(collision_filter);
   ASSERT_EQ(contact_data.contact_surfaces().size(), 0);
   /* Add a rigid unit cube. */
   GeometryId rigid_id = GeometryId::get_new_id();
@@ -330,9 +348,10 @@ GTEST_TEST(GeometriesTest, ComputeDeformableContact) {
   math::RigidTransform<double> X_WR(Vector3d(0, -2.0, 0));
   geometries.MaybeAddRigidGeometry(Box::MakeCube(1.0), rigid_id,
                                    rigid_properties, X_WR);
+  collision_filter.AddGeometry(rigid_id);
 
   /* The deformable box and the rigid box are not in contact yet. */
-  contact_data = geometries.ComputeDeformableContact();
+  contact_data = geometries.ComputeDeformableContact(collision_filter);
   ASSERT_EQ(contact_data.contact_surfaces().size(), 0);
 
   /* Now shift the rigid geometry closer to the deformable geometry.
@@ -356,7 +375,7 @@ GTEST_TEST(GeometriesTest, ComputeDeformableContact) {
   geometries.UpdateRigidWorldPose(rigid_id, X_WR);
 
   /* Now there should be exactly one contact data. */
-  contact_data = geometries.ComputeDeformableContact();
+  contact_data = geometries.ComputeDeformableContact(collision_filter);
   ASSERT_EQ(contact_data.contact_surfaces().size(), 1);
 
   /* Verify that the contact surface is as expected. */
@@ -389,6 +408,14 @@ GTEST_TEST(GeometriesTest, ComputeDeformableContact) {
             expected_contact_surface.num_contact_points());
   EXPECT_TRUE(contact_surface.contact_mesh_W().Equal(
       expected_contact_surface.contact_mesh_W()));
+
+  /* No contact is reported if the the pair of rigid and deformable geometries
+   are filtered in the collision filter. */
+  collision_filter.Apply(CollisionFilterDeclaration().ExcludeBetween(
+                             GeometrySet(deformable_id), GeometrySet(rigid_id)),
+                         get_extract_ids_functor());
+  contact_data = geometries.ComputeDeformableContact(collision_filter);
+  EXPECT_EQ(contact_data.contact_surfaces().size(), 0);
 }
 
 }  // namespace
--- a/geometry/proximity/test/distance_sphere_to_shape_test.cc
+++ b/geometry/proximity/test/distance_sphere_to_shape_test.cc
@@ -626,7 +626,8 @@ GTEST_TEST(Callback, ScalarSupportWithFilters) {
   // Filter the pair (A, B); we'll put the ids in a set and simply return that
   // set for the extract ids function.
   std::unordered_set<GeometryId> ids{data_A.id(), data_B.id()};
-  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                               CollisionFilterScope) {
     return ids;
   };
   collision_filter.Apply(CollisionFilterDeclaration().ExcludeWithin(
@@ -679,7 +680,8 @@ GTEST_TEST(Callback, RespectCollisionFiltering) {
   // Filter the pair (A, B); we'll put the ids in a set and simply return that
   // set for the extract ids function.
   std::unordered_set<GeometryId> ids{data_A.id(), data_B.id()};
-  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                               CollisionFilterScope) {
     return ids;
   };
   collision_filter.Apply(CollisionFilterDeclaration().ExcludeWithin(
--- a/geometry/proximity/test/find_collision_candidates_callback_test.cc
+++ b/geometry/proximity/test/find_collision_candidates_callback_test.cc
@@ -62,7 +62,8 @@ GTEST_TEST(Callback, RespectsCollisionFilter) {
   // Filter the pair (A, B); we'll put the ids in a set and simply return that
   // set for the extract ids function.
   std::unordered_set<GeometryId> ids{data_A.id(), data_B.id()};
-  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+  CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                               CollisionFilterScope) {
     return ids;
   };
   collision_filter.Apply(CollisionFilterDeclaration().ExcludeWithin(
--- a/geometry/proximity/test/hydroelastic_callback_test.cc
+++ b/geometry/proximity/test/hydroelastic_callback_test.cc
@@ -211,7 +211,8 @@ class TestScene {
     // Filter the pair (A, B); we'll put the ids in a set and simply return that
     // set for the extract ids function.
     std::unordered_set<GeometryId> ids{data_A.id(), data_B.id()};
-    CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+    CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                                 CollisionFilterScope) {
       return ids;
     };
     collision_filter_.Apply(CollisionFilterDeclaration().ExcludeWithin(
--- a/geometry/proximity/test/penetration_as_point_pair_callback_test.cc
+++ b/geometry/proximity/test/penetration_as_point_pair_callback_test.cc
@@ -169,7 +169,8 @@ class PenetrationAsPointPairCallbackTest : public ::testing::Test {
     // Filter the pair (A, B); we'll put the ids in a set and simply return that
     // set for the extract ids function.
     std::unordered_set<GeometryId> ids{id_A_, id_B_};
-    CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&) {
+    CollisionFilter::ExtractIds extract = [&ids](const GeometrySet&,
+                                                 CollisionFilterScope) {
       return ids;
     };
     collision_filter_.Apply(
--- a/geometry/proximity_engine.cc
+++ b/geometry/proximity_engine.cc
@@ -776,7 +776,8 @@ class ProximityEngine<T>::Impl : public ShapeReifier {
   void ComputeDeformableContact(
       DeformableContact<double>* deformable_contact) const {
     *deformable_contact =
-        geometries_for_deformable_contact_.ComputeDeformableContact();
+        geometries_for_deformable_contact_.ComputeDeformableContact(
+            collision_filter_);
   }
 
   // Testing utilities
--- a/geometry/scene_graph.h
+++ b/geometry/scene_graph.h
@@ -915,6 +915,13 @@ class SceneGraph final : public systems::LeafSystem<T> {
    stored in a particular Context. These methods provide access to the manager
    for the data stored in either location.
 
+   %SceneGraph implicitly filters collisions between rigid geometries affixed to
+   the same frame. This allows representation of complex shapes by providing a
+   union of simpler shapes without producing spurious collisions between those
+   overlapping shapes. %SceneGraph doesn't create *any* collision filters for
+   deformable geometries automatically. Users can add filters to deformable
+   geometries as they require after registration.
+
    Generally, it should be considered a bad practice to hang onto the instance
    of CollisionFilterManager returned by collision_filter_manager(). It is not
    immediately clear whether a particular CollisionFilterManager instance
--- a/geometry/test/geometry_state_test.cc
+++ b/geometry/test/geometry_state_test.cc
@@ -553,6 +553,15 @@ class GeometryStateTestBase {
     return source_id_;
   }
 
+  template <typename PropertyType>
+  void AssignRoleToSingleSourceTree(const PropertyType& properties) {
+    ASSERT_TRUE(source_id_.is_valid());
+    for (GeometryId id : geometries_) {
+      geometry_state_.AssignRole(source_id_, id, properties);
+    }
+    geometry_state_.AssignRole(source_id_, anchored_geometry_, properties);
+  }
+
   // This function sets up a dummy GeometryState that contains both rigid
   // (non-deformable) and deformable geometries to facilitate testing.
   SourceId SetUpWithRigidAndDeformableGeometries() {
@@ -678,15 +687,6 @@ class GeometryStateTestBase {
     // get thrown out.
     properties.AddProperty("test", "value", 17);
     AssignRoleToSingleSourceTree(properties);
-  }
-
-  template <typename PropertyType>
-  void AssignRoleToSingleSourceTree(const PropertyType& properties) {
-    ASSERT_TRUE(source_id_.is_valid());
-    for (GeometryId id : geometries_) {
-      geometry_state_.AssignRole(source_id_, id, properties);
-    }
-    geometry_state_.AssignRole(source_id_, anchored_geometry_, properties);
   }
 };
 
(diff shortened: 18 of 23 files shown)


xuchenhan-tri (commented):
+@SeanCurtis-TRI could you please take a look at this PR and see if it does what you envisioned in our discussion in #19854? Thank you.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NainV-PCba3V2LiO_wa:b8lkx9j)* status: LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

SeanCurtis-TRI (commented):
This is basically what we discussed. See my quibbles (and apologies for the redundant comments about missing tests -- I know you haven't bothered with them until we feel good about what's there).

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-Najh9Gb2OOkTwi2qbo-:b-4y3vsv)* status: 14 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/collision_filter_declaration.h` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajO_ci3bxIRv9gQ0n2:-NajO_ci3bxIRv9gQ0n3:beai3bp) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/collision_filter_declaration.h#L19)):*
> ```objc
> #endif
> 
> /** Enum that defines the scope of the geometries that are considered for the
> ```

nit: How about a stronger phrase.



_Suggestion:_
```objc
affected by
```
___
*[`geometry/collision_filter_declaration.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZmyQ6lHZ1LVtRPlj:-NajZmyQ6lHZ1LVtRPlk:b-tnoxe4) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/collision_filter_declaration.h#L21)):*
> ```objc
> /** Enum that defines the scope of the geometries that are considered for the
>  collision filtering mechanism. */
> enum class CollisionFilterCandidates {
> ```

BTW Can we come up with a shorter name?  Looking at the call sites, `geometry::CollisionFilterCandidates::kOmitGeometry` is a mouthful. Let me brainstorm here:

1. Replace "candidates" with "scope".
2. Abbreviate CollisionFilter so as to produce CFCandidates. (This is probably against the style guide).
3. Some wildly clever third thing that escapes me right now.

I suspect there's little we can do about this and will simply have to shrug. But it's worth figuring out before the PR merges, as this is part of the public API.
___
*[`geometry/geometry_state.h` line 642 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajP4s8D5g04Vfu-tZ4:-NajP4s8D5g04Vfu-tZ5:bds5j6i) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.h#L642)):*
> ```objc
>   std::unordered_set<GeometryId> CollectIds(
>       const GeometrySet& geometry_set, std::optional<Role> role,
>       bool exclude_deformable = false) const;
> ```

nit: What's the argument for not simply passing the enum through?

1. Currently a `bool` is enough, and
2. if we ever need to increase the space of `enum` such that a `bool` isn't enough, we can change signatures because they're all internal/private types?

Those are arguments for why converting to `bool` isn't wrong, but they don't seem to argue about it being *right*. As I look at the call sites, they all look a lot nice, just passing the candidates.



_Code quote:_
```objc
bool exclude_deformable = false
```
___
*[`geometry/geometry_state.cc` line 997 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXwTn0WhKE1T0VBXy:-NajXwTn0WhKE1T0VBXz:brpqbsj) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L997)):*
> ```cpp
>       // been identified. If none have been identified, this makes no changes.
>       geometry_engine_->collision_filter().Apply(
>           CollisionFilterDeclaration(CollisionFilterCandidates::kOmitDeformable)
> ```

nit: We need documentation in `SceneGraph`, promising this.
___
*[`geometry/geometry_state.cc` line 1279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajVdnu-6gmsUC5SHqS:-NajVdnu-6gmsUC5SHqT:bbenkvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1279)):*
> ```cpp
> 
> template <typename T>
> unordered_set<GeometryId> GeometryState<T>::CollectIds(
> ```

I recognize there's no unit test on `CollectIds` explicitly. But we should test this new functionality in `geometry_state_test.cc`.

We currently have the test `NonProximityRoleInCollisionFilter`. You can add an adjacent test to that which confirms the impact of candidates:
  1. kAll includes deformable.
   2. kOmitDeformable doesn't.
___
*[`geometry/geometry_state.cc` line 1306 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajSLFK440tnag8ryWY:-NajSLFK440tnag8ryWZ:b-8uomqc) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1306)):*
> ```cpp
>   }
> 
>   if (exclude_deformable) {
> ```

BTW  It seems the motivation for this approach is that, generally, the number of deformable geometries is far smaller than the number of rigid geometries. So, it makes sense to simply subtract the small number from the big number.

However, how big is the set of deformable geometries actually compared to the number of geometries included in the `GeometrySet` in this invocation? That ratio can be *much* different (and, typically will be). Here's an alternative to consider:

(I intentionally spelled the lambda the way I did because it lends itself well to changing `exclude_deformable` into `candidates`.)

```c++
template <typename T>
unordered_set<GeometryId> GeometryState<T>::CollectIds(
    const GeometrySet& geometry_set, std::optional<Role> role,
    bool exclude_deformable) const {
  auto must_include = [exclude_deformable](const InternalGeometry& g,
                                           const std::optional<Role>& role) {
    // Must have compatible role and be part of the candidates.
    return (!role.has_value() || g.has_role(*role)) &&
           (exclude_deformable == false || !g.is_deformable());
  };
  unordered_set<GeometryId> resultant_ids;
  for (auto frame_id : geometry_set.frames()) {
    const auto& frame = GetValueOrThrow(frame_id, frames_);
    for (auto geometry_id : frame.child_geometries()) {
      const InternalGeometry& geometry = geometries_.at(geometry_id);
      if (must_include(geometry, role)) {
        resultant_ids.insert(geometry_id);
      }
    }
  }

  for (auto geometry_id : geometry_set.geometries()) {
    const InternalGeometry* geometry = GetGeometry(geometry_id);
    if (geometry == nullptr) {
      throw std::logic_error(
          "Geometry set includes a geometry id that doesn't belong to the "
          "SceneGraph: " +
          to_string(geometry_id));
    }
    if (must_include(*geometry, role)) {
      resultant_ids.insert(geometry_id);
    }
  }

  return resultant_ids;
}
```
___
*[`geometry/geometry_state.cc` line 1309 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajQu3JDTL8gMFOD0vS:-NajQu3JDTL8gMFOD0vT:bkzkmd6) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1309)):*
> ```cpp
>     for (const auto [source_id, deformable_ids] :
>          source_deformable_geometry_id_map_) {
>       unused(source_id);
> ```

nit: Fun facts (well, two of them)

1. Our compiler isn't smart enough to treat Ids as ints, so it's barking about not having `auto&`.
2. There's a quick and dirty way to achieve the same effect as by your `unused` idiom.



_Suggestion:_
```cpp
    for (const auto& [_, deformable_ids] : source_deformable_geometry_id_map_) {
```
___
*[`geometry/proximity/collision_filter.h` line 194 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajTaSV2Dwm-wo0qoZO:-NajTaSV2Dwm-wo0qoZP:b-w0k140) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.h#L194)):*
> ```objc
>    filtered. For each pair, if they are already filtered, no discernible change
>    is made. Only geometries within the set described by `candidates` are
>    considered.
> ```

BTW It didn't feel like this language matched the language of the enum names. It's hard to see that `kOmitDeformables` particularly "describes" a set.

Perhaps it would be more complete to say:

```c++
  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A*, id_B ∈ set_B*` to be
   filtered. Where `set_A*` and `set_B*` are the geometry ids in `set_A` and
   `set_B`, respectively, consistent with the given `candidates`. For each
   pair, if they are already filtered, no discernible change is made.
```

The same treatment would apply to `RemoveFiltersBetween`. It might be simpler to factor it out into a group to describe the effect of candidates. However, it's an internal class that doesn't get doxygenated, so, a little copypasta probably isn't bad.
___
*[`geometry/proximity/collision_filter.cc` line 238 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajVBuuERMaHQ35nmAB:-NajVBuuERMaHQ35nmAC:b-td4gvj) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.cc#L238)):*
> ```cpp
>                             bool is_invariant, FilterState* filter_state) {
>   using Operation = CollisionFilterDeclaration::StatementOp;
>   CollisionFilterCandidates candidates = declaration.candidates();
> ```

nit `const`
___
*[`geometry/proximity/test/deformable_contact_internal_test.cc` line 412 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajWiX94NlVhNEYqtJi:-NajWiX94NlVhNEYqtJj:b-ptamf) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/test/deformable_contact_internal_test.cc#L412)):*
> ```cpp
>       expected_contact_surface.contact_mesh_W()));
> 
>   /* No contact is reported if the the pair of rigid and deformable geometries
> ```

BTW Specifically, what this proves is that the class is asking the collision filter.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1252 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajX9Tp44kv-aS8LfE_:-NajX9Tp44kv-aS8LfEa:b-aqjj4h) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/parsing/detail_mujoco_parser.cc#L1252)):*
> ```cpp
> 
>       manager.Apply(geometry::CollisionFilterDeclaration(
>                         geometry::CollisionFilterCandidates::kOmitDeformable)
> ```

Is this necessary? If so, it needs a unit test.

Arguments for why it's not necessary:
  - We can't make deformable bodies from parsing mujoco.
  - Parsing a mujoco file has no access to deformable bodies created elsewhere. So, these two `geom*_id` values cannot be deformable.
___
*[`multibody/plant/multibody_plant.cc` line 1207 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXra946kfosm1oBAX:-NajXra946kfosm1oBAY:bodly5m) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1207)):*
> ```cpp
>         scene_graph_->collision_filter_manager().Apply(
>             CollisionFilterDeclaration(
>                 geometry::CollisionFilterCandidates::kOmitDeformable)
> ```

nit: We need documentation in `MbP` promising this.
___
*[`multibody/plant/multibody_plant.cc` line 1226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZiTM33Q99QiR83k5:-NajZiTM33Q99QiR83k6:b-ksgh6y) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1226)):*
> ```cpp
>     auto geometries = CollectRegisteredGeometries(subgraph_bodies);
>     scene_graph_->collision_filter_manager().Apply(
>         CollisionFilterDeclaration().ExcludeWithin(geometries));
> ```

nit: Missed declaration. Again, unit tests.



_Code quote:_
```cpp
CollisionFilterDeclaration()
```
___
*[`multibody/plant/multibody_plant.cc` line 1261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZULwA73sS7wRbOZU:-NajZULwA73sS7wRbOZV:b-7xv7av) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1261)):*
> ```cpp
>   if (collision_filter_group_a.first == collision_filter_group_b.first) {
>     scene_graph_->collision_filter_manager().Apply(
>         CollisionFilterDeclaration().ExcludeWithin(
> ```

nit: I believe you missed a declaration.

This function is used by the parser, we definitely need some tests to show `MbP` is doing the right thing.



_Code quote:_
```cpp
CollisionFilterDeclaration()
```

xuchenhan-tri (commented):
Thanks for the prompt review! I think we agree that the PR is in the right direction. I will polish this a bit more later this week and ping back when it's ready for another look.

+(status: do not merge) +(status: do not review) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-Nam795N-haW--c9Vnov:b-nkltdv)* status: 14 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes

xuchenhan-tri (commented):

I added some unit tests and this is ready for another look. Thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NbMqinIEoN5tCDJ8OnZ:bl4j98a)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/collision_filter_declaration.h` line 19 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajO_ci3bxIRv9gQ0n2:-Nam8PwVCNy5P_Fw7WNe:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/collision_filter_declaration.h#L19)):*

nit: How about a stronger phrase.
</blockquote></details>

Done
___
*[`geometry/collision_filter_declaration.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZmyQ6lHZ1LVtRPlj:-Nam8QguEEvlF1ISW2MR:ba8ieye) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/collision_filter_declaration.h#L21)):*

BTW Can we come up with a shorter name?  Looking at the call sites, `geometry::CollisionFilterCandidates::kOmitGeometry` is a mouthful. Let me brainstorm here:

1. Replace "candidates" with "scope".
2. Abbreviate CollisionFilter so as to produce CFCandidates. (This is probably against the style guide).
3. Some wildly clever third thing that escapes me right now.

I suspect there's little we can do about this and will simply have to shrug. But it's worth figuring out before the PR merges, as this is part of the public API.
</blockquote></details>

Done

I've gone with `CollisionFilterScope`. I agree that `CFCandidates` is against GSG.
___
*[`geometry/geometry_state.h` line 642 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajP4s8D5g04Vfu-tZ4:-NamALwi1hM-lByGqycJ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.h#L642)):*

nit: What's the argument for not simply passing the enum through?

1. Currently a `bool` is enough, and
2. if we ever need to increase the space of `enum` such that a `bool` isn't enough, we can change signatures because they're all internal/private types?

Those are arguments for why converting to `bool` isn't wrong, but they don't seem to argue about it being *right*. As I look at the call sites, they all look a lot nice, just passing the candidates.
</blockquote></details>

Done
___
*[`geometry/geometry_state.cc` line 997 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXwTn0WhKE1T0VBXy:-Nam8kPAEwNjUpwdlZrx:b-fyrfe8) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L997)):*

nit: We need documentation in `SceneGraph`, promising this.
</blockquote></details>

It's not clear to me where we should document this. The promise that rigid geometries registered to the same frame are filtered seems to be an implicit one. Perhaps add a note at RegisterGeometry? Let me know if you have a better place in mind.
___
*[`geometry/geometry_state.cc` line 1279 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajVdnu-6gmsUC5SHqS:-Nam8mOT87fUiJV1GFnK:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1279)):*

I recognize there's no unit test on `CollectIds` explicitly. But we should test this new functionality in `geometry_state_test.cc`.

We currently have the test `NonProximityRoleInCollisionFilter`. You can add an adjacent test to that which confirms the impact of candidates:
  1. kAll includes deformable.
   2. kOmitDeformable doesn't.
</blockquote></details>

Done
___
*[`geometry/geometry_state.cc` line 1306 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajSLFK440tnag8ryWY:-Nam9T-n22FBY1GAcmkR:b-dtkds3) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1306)):*

BTW  It seems the motivation for this approach is that, generally, the number of deformable geometries is far smaller than the number of rigid geometries. So, it makes sense to simply subtract the small number from the big number.

However, how big is the set of deformable geometries actually compared to the number of geometries included in the `GeometrySet` in this invocation? That ratio can be *much* different (and, typically will be). Here's an alternative to consider:

(I intentionally spelled the lambda the way I did because it lends itself well to changing `exclude_deformable` into `candidates`.)

```c++
template <typename T>
unordered_set<GeometryId> GeometryState<T>::CollectIds(
    const GeometrySet& geometry_set, std::optional<Role> role,
    bool exclude_deformable) const {
  auto must_include = [exclude_deformable](const InternalGeometry& g,
                                           const std::optional<Role>& role) {
    // Must have compatible role and be part of the candidates.
    return (!role.has_value() || g.has_role(*role)) &&
           (exclude_deformable == false || !g.is_deformable());
  };
  unordered_set<GeometryId> resultant_ids;
  for (auto frame_id : geometry_set.frames()) {
    const auto& frame = GetValueOrThrow(frame_id, frames_);
    for (auto geometry_id : frame.child_geometries()) {
      const InternalGeometry& geometry = geometries_.at(geometry_id);
      if (must_include(geometry, role)) {
        resultant_ids.insert(geometry_id);
      }
    }
  }

  for (auto geometry_id : geometry_set.geometries()) {
    const InternalGeometry* geometry = GetGeometry(geometry_id);
    if (geometry == nullptr) {
      throw std::logic_error(
          "Geometry set includes a geometry id that doesn't belong to the "
          "SceneGraph: " +
          to_string(geometry_id));
    }
    if (must_include(*geometry, role)) {
      resultant_ids.insert(geometry_id);
    }
  }

  return resultant_ids;
}
```
</blockquote></details>

Done, thanks!
___
*[`geometry/geometry_state.cc` line 1309 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajQu3JDTL8gMFOD0vS:-Nam9kLr1rJ3dFa_1EX3:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L1309)):*

nit: Fun facts (well, two of them)

1. Our compiler isn't smart enough to treat Ids as ints, so it's barking about not having `auto&`.
2. There's a quick and dirty way to achieve the same effect as by your `unused` idiom.
</blockquote></details>

Done
___
*[`geometry/proximity/collision_filter.h` line 194 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajTaSV2Dwm-wo0qoZO:-Nam9vhZDpE9V06eZFo_:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.h#L194)):*

BTW It didn't feel like this language matched the language of the enum names. It's hard to see that `kOmitDeformables` particularly "describes" a set.

Perhaps it would be more complete to say:

```c++
  /* Declares pairs (`id_A`, `id_B`) `∀ id_A ∈ set_A*, id_B ∈ set_B*` to be
   filtered. Where `set_A*` and `set_B*` are the geometry ids in `set_A` and
   `set_B`, respectively, consistent with the given `candidates`. For each
   pair, if they are already filtered, no discernible change is made.
```

The same treatment would apply to `RemoveFiltersBetween`. It might be simpler to factor it out into a group to describe the effect of candidates. However, it's an internal class that doesn't get doxygenated, so, a little copypasta probably isn't bad.
</blockquote></details>

Done
___
*[`geometry/proximity/collision_filter.cc` line 238 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajVBuuERMaHQ35nmAB:-Nam9zlOAbuDS6xZnqdF:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.cc#L238)):*

nit `const`
</blockquote></details>

Done
___
*[`geometry/proximity/test/deformable_contact_internal_test.cc` line 412 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajWiX94NlVhNEYqtJi:-NamAOLO0ZrzB68tfQws:b-kl6gla) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/test/deformable_contact_internal_test.cc#L412)):*

BTW Specifically, what this proves is that the class is asking the collision filter.
</blockquote></details>

Done, I tweaked the language slightly
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1252 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajX9Tp44kv-aS8LfE_:-NamA3hZ4ullTSaZ5eFA:b-8vo8dy) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/parsing/detail_mujoco_parser.cc#L1252)):*

Is this necessary? If so, it needs a unit test.

Arguments for why it's not necessary:
  - We can't make deformable bodies from parsing mujoco.
  - Parsing a mujoco file has no access to deformable bodies created elsewhere. So, these two `geom*_id` values cannot be deformable.
</blockquote></details>

Done. Agreed that this is not necessary.
___
*[`multibody/plant/multibody_plant.cc` line 1207 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXra946kfosm1oBAX:-NamA56PD6754FHCCELt:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1207)):*

nit: We need documentation in `MbP` promising this.
</blockquote></details>

Done
___
*[`multibody/plant/multibody_plant.cc` line 1226 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZiTM33Q99QiR83k5:-NamA6051TgKV6Eg2bfB:bn50j8y) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1226)):*

nit: Missed declaration. Again, unit tests.
</blockquote></details>

Done via DeformableCollisionFilterTest
___
*[`multibody/plant/multibody_plant.cc` line 1261 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajZULwA73sS7wRbOZU:-NamAKCa5hav7XKHOu6w:bn50j8y) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/multibody/plant/multibody_plant.cc#L1261)):*

nit: I believe you missed a declaration.

This function is used by the parser, we definitely need some tests to show `MbP` is doing the right thing.
</blockquote></details>

Done via DeformableCollisionFilterTest

SeanCurtis-TRI (commented):
First pass done.

> Reviewed 18 of 19 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NbNOpY_1SZq7PwwoiyG:b-te4kzy)* status: 12 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/geometry_state.h` line 640 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN1br967VedVQs4W0y:-NbN1br967VedVQs4W0z:b-9lahkh) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/geometry_state.h#L640)):*
> ```objc
>   std::unordered_set<GeometryId> CollectIds(
>       const GeometrySet& geometry_set, std::optional<Role> role,
>       CollisionFilterScope scope = CollisionFilterScope::kOmitDeformable) const;
> ```

nit: Let's prefer *not* having a default value for `scope`. There's only one call site of `CollectIds` that relies on the default value, and we might as well pass `kAll` there so it's clear and precise.
___
*[`geometry/geometry_state.cc` line 997 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXwTn0WhKE1T0VBXy:-NbN2SV-8bryI1qlEg21:b-6zemtf) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L997)):*

It's not clear to me where we should document this. The promise that rigid geometries registered to the same frame are filtered seems to be an implicit one. Perhaps add a note at RegisterGeometry? Let me know if you have a better place in mind.
</blockquote></details>

I'd say it should go in the documentation group for `Collision filtering` in `scene_graph.h`. (around line 900). Something along the lines of the middle paragraph below:


```
  ...
  stored in a particular Context. These methods provide access to the manager
  for the data stored in either location.

   %SceneGraph implicitly filters collisions between rigid geometries affixed to
   the same frame. This allows representation of complex shapes by providing a
   union of simpler shapes without producing spurious collisions between those
   overlapping shapes. %SceneGraph doesn't create *any* collision filters for
   deformable bodies automatically. Users can add filters to deformable bodies
   as they require after registration.

   Generally, it should be considered a bad practice to hang onto the instance
   of CollisionFilterManager returned by collision_filter_manager(). It is not
  ...
```
___
*[`geometry/geometry_state.cc` line 1284 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN4H892GJ1XH_3cu4V:-NbN4H892GJ1XH_3cu4W:b-d4781v) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/geometry_state.cc#L1284)):*
> ```cpp
>     // Must have compatible role and be part of the scope.
>     return (!r.has_value() || g.has_role(*r)) &&
>            (scope == CollisionFilterScope::kAll || !g.is_deformable());
> ```

BTW The thought just occurred to me that this second line could simply be:

`g.in_scope(scope)`

I.e., a geometry knows about scope and it can tell itself if it is included in the scope. The only reason to *not* do this would be if the definition of scope includes more properties than are contained in an `InternalGeometry` -- like there were some kind of context like "only geometries ending with 'y'". 

This can certainly be kicked down the road, but it popped into my head. 

Thoughts?
___
*[`geometry/proximity/collision_filter.h` line 194 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajTaSV2Dwm-wo0qoZO:-NbN57Dx2amek9HnsWFv:b-wyu7c) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.h#L194)):*

Done
</blockquote></details>

I'd prefer that the definitions of `set_A*` and `set_B*` get pulled out of the documentation for `AddFiltersBetween` and hoisted to some higher level, or it should be included in `RemoveFiltersBetween`. `scope` plays the same role in both functions and should be treated symmetrically.
___
*[`geometry/test/geometry_state_test.cc` line 2555 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN67os8q1fo6dpEFNW:-NbN67os8q1fo6dpEFNX:b5xumfb) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/test/geometry_state_test.cc#L2555)):*
> ```cpp
> }
> 
> // Test collision filter respects the CollisionFilterScope enum.
> ```

nit: One critical piece of information is missing from this documentation, this is the closest thing that we have to a direct test of `GeometryState::CollectIds()`. That should, at the very least, be made explicit.

I'd *prefer* to push it further. I like two parts in my test documentation: what is being tested and how it's being tested. I ended up producing something like:

```c++
// Tests two aspects of GeometryState collision filter behavior:
//
//   1. No collision filters are applied to a deformable geometry (it and the
//      anchored geometry are both registered to the world frame) automatically.
//   2. GeometryState::CollectIds() respects the scope parameter in applying
//      collision filters.
//
// This is tested by calling ComputeDeformableContact(). In its initial state,
// we should report contact between a massive deformable body and *all* 
// dynamic rigid bodies (including the anchored geometry). Then applying
// a collision filter between the deformable body and the other geometries
// will eliminate all but one contacts.
```
___
*[`geometry/test/geometry_state_test.cc` line 2605 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN7eWM5HQu__beqXC1:-NbN7eWM5HQu__beqXC2:b-qlhw8f) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/test/geometry_state_test.cc#L2605)):*
> ```cpp
>       CollisionFilterDeclaration(CollisionFilterScope::kAll)
>           .ExcludeBetween(GeometrySet{deformable_id},
>                           GeometrySet(geometries_)));
> ```

BTW Why leave the anchored geometry? Simplicity of code? It *is* somewhat interesting that the one thing that got left behind is anchored. There's a *feeling* of creating filters that seem correlated with classification of the geometries that is misleading in the context of this test. Obviously, that's not the *point* of the test, but it feels like it's important that the anchored geometry got left behind.

However, I recognize that the definition of the `GeometrySet` is simpler using *either* the anchored id *or* the dynamic ids.
___
*[`multibody/plant/multibody_plant.h` line 604 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8MftE2Lg9CjpJo7D:-NbN8MftE2Lg9CjpJo7E:blcbcfi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.h#L604)):*
> ```objc
> - declare the plant's state,
> - declare the plant's input and output ports,
> - declare collision filters to ignore collisions:
> ```

nit: Here in the finalize discussion on collision filters, you should make sure this also clearly excludes deformable bodies from any collision filtering.
___
*[`multibody/plant/multibody_plant.cc` line 1246 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8fJ125x9fTcdRcae:-NbN8fJ125x9fTcdRcaf:bung48c) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.cc#L1246)):*
> ```cpp
>   scene_graph_->collision_filter_manager().Apply(
>       CollisionFilterDeclaration(
>         geometry::CollisionFilterScope::kOmitDeformable)
> ```

nit: With the `using` you no longer need the `geometry::` namespace. (It appears 3X in this file).
___
*[`multibody/plant/multibody_plant.cc` line 1253 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNMW6k5D6X_PcmuKlg:-NbNMW6k5D6X_PcmuKlh:biz3fpa) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.cc#L1253)):*
> ```cpp
> 
> template <typename T>
> void MultibodyPlant<T>::ExcludeCollisionGeometriesWithCollisionFilterGroupPair(
> ```

As near as I can tell, *this* function is only exercised by the parsing. So, to test this, you'll need to parse stuff.
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 30 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8zit3cxml9k6Ut-F:-NbN8zit3cxml9k6Ut-G:b8a9ykt) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L30)):*
> ```cpp
>  protected:
>   /* Sets up a scene with a deformable sphere intersecting two rigid bodies, one
>    weld to the world and the other free. */
> ```

nit: grammar



_Suggestion:_
```cpp
welded
```
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 78 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNL74x5hzwhMFlblAE:-NbNL74x5hzwhMFlblAF:bxs3jvi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L78)):*
> ```cpp
>   }
> 
>   /* Registers a deformable octrahedron with the given `name` and pose in the
> ```

nit: typo



_Suggestion:_
```cpp
octahedron
```
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 78 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNLJ_H1xBnJBd8b58R:-NbNLJ_I7qpddYV9hfKh:b2zq2n) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L78)):*
> ```cpp
>   }
> 
>   /* Registers a deformable octrahedron with the given `name` and pose in the
> ```

nit "...and pose".  No pose given.
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 126 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNMgbX4QTKM_08bmB8:-NbNMgbX4QTKM_08bmB9:b-r9vfz7) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L126)):*
> ```cpp
> };
> 
> /* Tests that the default collision filters applied in MbP does not affect
> ```

nit: grammar



_Suggestion:_
```cpp
MbP do not
```

xuchenhan-tri (commented):
Comments addressed, PTAL, thanks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NbQsM7V-76Us7Ww4uUU:b-4l0k1q)* status: 3 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform), needs at least two assigned reviewers, missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/geometry_state.h` line 640 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN1br967VedVQs4W0y:-NbQCHBM3mBqYYTDKKbV:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/geometry_state.h#L640)):*

nit: Let's prefer *not* having a default value for `scope`. There's only one call site of `CollectIds` that relies on the default value, and we might as well pass `kAll` there so it's clear and precise.
</blockquote></details>

Done
___
*[`geometry/geometry_state.cc` line 997 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajXwTn0WhKE1T0VBXy:-NbQI_uW39PwjAOuLRJz:bf3mav4) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/geometry_state.cc#L997)):*

I'd say it should go in the documentation group for `Collision filtering` in `scene_graph.h`. (around line 900). Something along the lines of the middle paragraph below:


```
  ...
  stored in a particular Context. These methods provide access to the manager
  for the data stored in either location.

   %SceneGraph implicitly filters collisions between rigid geometries affixed to
   the same frame. This allows representation of complex shapes by providing a
   union of simpler shapes without producing spurious collisions between those
   overlapping shapes. %SceneGraph doesn't create *any* collision filters for
   deformable bodies automatically. Users can add filters to deformable bodies
   as they require after registration.

   Generally, it should be considered a bad practice to hang onto the instance
   of CollisionFilterManager returned by collision_filter_manager(). It is not
  ...
```

</blockquote></details>

Done, thank you.
___
*[`geometry/geometry_state.cc` line 1284 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN4H892GJ1XH_3cu4V:-NbQJ6IlA2oY1fcjq3j2:b-iofvfs) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/geometry_state.cc#L1284)):*

BTW The thought just occurred to me that this second line could simply be:

`g.in_scope(scope)`

I.e., a geometry knows about scope and it can tell itself if it is included in the scope. The only reason to *not* do this would be if the definition of scope includes more properties than are contained in an `InternalGeometry` -- like there were some kind of context like "only geometries ending with 'y'". 

This can certainly be kicked down the road, but it popped into my head. 

Thoughts?
</blockquote></details>

I agree this can be done in `InternalGeometry` but I'll pass for now because it doesn't create a tremendous amount of value.
___
*[`geometry/proximity/collision_filter.h` line 194 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NajTaSV2Dwm-wo0qoZO:-NbQK-LR8WNK3xgLU8qz:becxbz6) ([raw file](https://github.com/RobotLocomotion/drake/blob/58571b057f2a208b229380e2b212cdba1e8c82cf/geometry/proximity/collision_filter.h#L194)):*

I'd prefer that the definitions of `set_A*` and `set_B*` get pulled out of the documentation for `AddFiltersBetween` and hoisted to some higher level, or it should be included in `RemoveFiltersBetween`. `scope` plays the same role in both functions and should be treated symmetrically.
</blockquote></details>

I added the definition to `RemoveFiltersBetween` because I feel the definition is easier to spot/understand locally, given context.
___
*[`geometry/test/geometry_state_test.cc` line 2555 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN67os8q1fo6dpEFNW:-NbQU_NjDbfCTCjh6Uc8:b-x5eyqi) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/test/geometry_state_test.cc#L2555)):*

nit: One critical piece of information is missing from this documentation, this is the closest thing that we have to a direct test of `GeometryState::CollectIds()`. That should, at the very least, be made explicit.

I'd *prefer* to push it further. I like two parts in my test documentation: what is being tested and how it's being tested. I ended up producing something like:

```c++
// Tests two aspects of GeometryState collision filter behavior:
//
//   1. No collision filters are applied to a deformable geometry (it and the
//      anchored geometry are both registered to the world frame) automatically.
//   2. GeometryState::CollectIds() respects the scope parameter in applying
//      collision filters.
//
// This is tested by calling ComputeDeformableContact(). In its initial state,
// we should report contact between a massive deformable body and *all* 
// dynamic rigid bodies (including the anchored geometry). Then applying
// a collision filter between the deformable body and the other geometries
// will eliminate all but one contacts.
```
</blockquote></details>

Done, thanks for the provided text.
___
*[`geometry/test/geometry_state_test.cc` line 2605 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN7eWM5HQu__beqXC1:-NbQVvScCLEd_vqo8guj:b-l3m52n) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/test/geometry_state_test.cc#L2605)):*

BTW Why leave the anchored geometry? Simplicity of code? It *is* somewhat interesting that the one thing that got left behind is anchored. There's a *feeling* of creating filters that seem correlated with classification of the geometries that is misleading in the context of this test. Obviously, that's not the *point* of the test, but it feels like it's important that the anchored geometry got left behind.

However, I recognize that the definition of the `GeometrySet` is simpler using *either* the anchored id *or* the dynamic ids.
</blockquote></details>

I modified the test. PTAL.
___
*[`multibody/plant/multibody_plant.h` line 604 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8MftE2Lg9CjpJo7D:-NbQHEVNBWWKUXKlv9bN:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.h#L604)):*

nit: Here in the finalize discussion on collision filters, you should make sure this also clearly excludes deformable bodies from any collision filtering.
</blockquote></details>

Done
___
*[`multibody/plant/multibody_plant.cc` line 1246 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8fJ125x9fTcdRcae:-NbQHL9MEHyBCrrGIEFi:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.cc#L1246)):*

nit: With the `using` you no longer need the `geometry::` namespace. (It appears 3X in this file).
</blockquote></details>

Done
___
*[`multibody/plant/multibody_plant.cc` line 1253 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNMW6k5D6X_PcmuKlg:-NbQZM8i6mWgYfwgnXvk:b-lhuee6) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.cc#L1253)):*

As near as I can tell, *this* function is only exercised by the parsing. So, to test this, you'll need to parse stuff.
</blockquote></details>

I added test coverage for this in `deformable_collision_filter_test.cc`. See if you agree.

The argument the mentioned test provides coverage is that:

1. The behavior involving only rigid geometries should be covered by existing tests (e.g. of parsers).
2. In `deformable_collision_filter_test.cc`, the new tests ensure that `MbP::ExcludeCollisionGeometriesWithCollisionFilterGroupPair` does in fact omit deformable geometries.
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 30 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8zit3cxml9k6Ut-F:-NbQHtBW819VgIdTbxc2:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L30)):*

nit: grammar
</blockquote></details>

Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 78 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNL74x5hzwhMFlblAE:-NbQI06m0qhCreutVQPV:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L78)):*

nit: typo
</blockquote></details>

Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 78 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNLJ_H1xBnJBd8b58R:-NbQI82bA13xB5BLz1JQ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L78)):*

nit "...and pose".  No pose given.
</blockquote></details>

Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 126 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNMgbX4QTKM_08bmB8:-NbQICxo5p1cJjP43dzR:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L126)):*

nit: grammar
</blockquote></details>

Done

SeanCurtis-TRI (commented):
:LGTM: +@jwnimmer-tri for platform review, please.

> Reviewed 17 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NbRV5Xr3zjNpBaWtYRi:b-yomphx)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/geometry_state.cc` line 1284 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN4H892GJ1XH_3cu4V:-NbRPmdi4UQfJn2ki3jw:bw9c9ew) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/geometry/geometry_state.cc#L1284)):*

I agree this can be done in `InternalGeometry` but I'll pass for now because it doesn't create a tremendous amount of value.
</blockquote></details>

Yeah...I had a similar feeling.
___
*[`geometry/test/geometry_state_test.cc` line 2618 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRQpA9D-6VWVY-JcNV:-NbRQpA9D-6VWVY-JcNW:b-kv804b) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/geometry/test/geometry_state_test.cc#L2618)):*
> ```cpp
>   geometry_state_.ComputeDeformableContact(&contacts);
>   EXPECT_EQ(contacts.contact_surfaces().size(),
>             single_tree_total_geometry_count() - 1);
> ```

BTW Given the change you've made, I'd suggest nudging it a bit further in this direction:

We can make sure it doesn't *just* work with dynamic geometries at almost no cost. :)



_Suggestion:_
```cpp
  // Filter with the kAll flag as the scope should have an effect. The collision
  // between the deformable geometry and one dynamic rigid geometry and one
  // anchored rigid geometry are filtered, so the number of collisions should
  // reduce by 2.
  geometry_state_.collision_filter_manager().Apply(
      CollisionFilterDeclaration(CollisionFilterScope::kAll)
          .ExcludeBetween(GeometrySet{deformable_id},
                          GeometrySet({geometries_[0], anchored_geometry_})));
  geometry_state_.ComputeDeformableContact(&contacts);
  EXPECT_EQ(contacts.contact_surfaces().size(),
            single_tree_total_geometry_count() - 2);
```
___
*[`multibody/plant/multibody_plant.h` line 614 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRRL290_SxII02d7mC:-NbRRL290_SxII02d7mD:b66gedl) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/multibody_plant.h#L614)):*
> ```objc
>   - within subgraphs of welded rigid bodies.
> Note that collisions involving deformable bodies are not affected by the
> collision filters declared in the call to Finalize().
> ```

BTW Can we make this stronger?



_Suggestion:_
```objc
Note that MultibodyPlant will *not* introduce *any* collision filters
on deformable bodies.
```
___
*[`multibody/plant/multibody_plant.cc` line 1253 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbNMW6k5D6X_PcmuKlg:-NbRUSo2AG4npciNMLFp:ba37ux5) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/multibody_plant.cc#L1253)):*

I added test coverage for this in `deformable_collision_filter_test.cc`. See if you agree.

The argument the mentioned test provides coverage is that:

1. The behavior involving only rigid geometries should be covered by existing tests (e.g. of parsers).
2. In `deformable_collision_filter_test.cc`, the new tests ensure that `MbP::ExcludeCollisionGeometriesWithCollisionFilterGroupPair` does in fact omit deformable geometries.
</blockquote></details>

I certainly support the direct execution of the function. It was silly of me to suggest parsing in place of just calling the function.  :)  My bad.
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 30 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8zit3cxml9k6Ut-F:-NbRRzUJDyuupeYdUP7K:bocxtjq) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L30)):*

Done
</blockquote></details>

I don't see the change.
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 81 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRTX0aBmuwVz4qOam4:-NbRTX0aBmuwVz4qOam5:bppjy5b) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/test/deformable_collision_filter_test.cc#L81)):*
> ```cpp
>         collision_filter_group_a, collision_filter_group_a);
>     plant_->ExcludeCollisionGeometriesWithCollisionFilterGroupPair(
>         collision_filter_group_a, collision_filter_group_b);
> ```

nit: Again, I'm going to ask that this be more explicit about its purpose and what it achieves

------

Follow up note: I wrote this before I read all the way through this file, so, before taking any action based on this comment, look at the next comment in this file. 



_Suggestion:_
```cpp
    /* ExcludeCollisionGeometriesWithCollisionFilterGroupPair() is invoked
     during parsing, but it still must adhere to MultibodyPlant's promise to
     not introduce collision filters on deformable bodies. Rather than exercise
     via parsing, we'll execute it here with some collision filters which
     *explicitly* include the deformable body's geometry id. If the dut has
     correct behavior (ignoring deformable geometries), there will be no
     collision filters involving the deformable geometry. */
    GeometrySet welded_and_deformable{deformable_geometry_id_,
                                      welded_geometry_id_};
    GeometrySet free_and_deformable{deformable_geometry_id_, free_geometry_id_};
    std::pair<std::string, GeometrySet> collision_filter_group_a = {
        "a", welded_and_deformable};
    std::pair<std::string, GeometrySet> collision_filter_group_b = {
        "b", free_and_deformable};
    plant_->ExcludeCollisionGeometriesWithCollisionFilterGroupPair(
        collision_filter_group_a, collision_filter_group_a);
    plant_->ExcludeCollisionGeometriesWithCollisionFilterGroupPair(
        collision_filter_group_a, collision_filter_group_b);
    DRAKE_DEMAND(!scene_graph_->model_inspector().CollisionFiltered(
        deformable_geometry_id_, welded_geometry_id_));
    DRAKE_DEMAND(!scene_graph_->model_inspector().CollisionFiltered(
        deformable_geometry_id_, free_geometry_id_));
```
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 139 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRTuaY70Uj3jpNdQVv:-NbRTuaY70Uj3jpNdQVw:bmm1xg1) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/test/deformable_collision_filter_test.cc#L139)):*
BTW Given my recommendation in `SetUp()`, this further elaboration is unnecessary. The advantage of the `DRAKE_DEMAND` in `SetUp()` is that we get the operation and confirmation in a nice local package.

Alternatively, you can change my recommended documentation in `SetUp()` to indicate that if

> the dut has correct behavior, then the MbPFilterIgnoresDeformable test will report all possible rigid-deformable contacts (two).

I leave it up to you.

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @xuchenhan-tri)

___
*[`geometry/test/geometry_state_test.cc` line 2618 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRQpA9D-6VWVY-JcNV:-NbRcym_B4Pq9r0QQaRu:b-dtkdrq) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/geometry/test/geometry_state_test.cc#L2618)):*

BTW Given the change you've made, I'd suggest nudging it a bit further in this direction:

We can make sure it doesn't *just* work with dynamic geometries at almost no cost. :)
</blockquote></details>

Done, thanks.
___
*[`multibody/plant/multibody_plant.h` line 614 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRRL290_SxII02d7mC:-NbRd3MN3MdpzVCwfuEM:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/multibody_plant.h#L614)):*

BTW Can we make this stronger?
</blockquote></details>

Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 30 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbN8zit3cxml9k6Ut-F:-NbRd6PTAwa9asiDHx9E:b-n8mhqn) ([raw file](https://github.com/RobotLocomotion/drake/blob/b27791be3fd4e1242f033ee5ad7e9cad350bd301/multibody/plant/test/deformable_collision_filter_test.cc#L30)):*

I don't see the change.
</blockquote></details>

Sorry. git snafu. Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 81 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRTX0aBmuwVz4qOam4:-NbRe2D6DXRp2yq33gB0:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/test/deformable_collision_filter_test.cc#L81)):*

nit: Again, I'm going to ask that this be more explicit about its purpose and what it achieves

------

Follow up note: I wrote this before I read all the way through this file, so, before taking any action based on this comment, look at the next comment in this file. 
</blockquote></details>

Done
___
*[`multibody/plant/test/deformable_collision_filter_test.cc` line 139 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbRTuaY70Uj3jpNdQVv:-NbReAq898IfBFU-bOBf:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/cad830d617871453c179cf867a4b9ad27e885676/multibody/plant/test/deformable_collision_filter_test.cc#L139)):*

BTW Given my recommendation in `SetUp()`, this further elaboration is unnecessary. The advantage of the `DRAKE_DEMAND` in `SetUp()` is that we get the operation and confirmation in a nice local package.

Alternatively, you can change my recommended documentation in `SetUp()` to indicate that if

> the dut has correct behavior, then the MbPFilterIgnoresDeformable test will report all possible rigid-deformable contacts (two).

I leave it up to you.
</blockquote></details>

Done

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897)* status: LGTM missing from assignee jwnimmer-tri(platform), missing label for release notes (waiting on @xuchenhan-tri)

jwnimmer-tri:
Timed out so => +@EricCousineau-TRI for platform.

EricCousineau-TRI (commented):
:lgtm: platform w/ minor nits

> Reviewed 2 of 19 files at r1, 5 of 19 files at r2, 14 of 17 files at r4, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-:-NbVH-tU-Lmc9VAooVOj:b7qtcjz)* status: 2 unresolved discussions

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVEMynBZ6_5gUgFvYD):*
nit Unclear what user-level takeaway is.

My understanding is that users will need to filter any deformable geometries directly through the SceneGraph API; MultibodyPlant API (and corresponding parsing) will only do rigid filtering.

If that is correct, would you be able to add that to this commit summary (or upon merge)?
___
*[`geometry/collision_filter_declaration.h` line 28 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVFokgDlhFIVt3lITJ:-NbVFokgDlhFIVt3lITK:b-8vqzwy) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/collision_filter_declaration.h#L28)):*
> ```objc
>    declaration even if they are included in the GeometrySet when the filter is
>    declared. */
>   kOmitDeformable
> ```

BTW (curiosity) Why "omit deformable" instead of "rigid only"?
Asking because most of the public documentation seems to qualify by "rigid" rather than "not deformable"
___
*[`geometry/geometry_state.cc` line 997 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVF-uY4KdMFv1xQb21:-NbVF-uY4KdMFv1xQb22:b-vqjsf4) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/geometry_state.cc#L997)):*
> ```cpp
>       // been identified. If none have been identified, this makes no changes.
>       geometry_engine_->collision_filter().Apply(
>           CollisionFilterDeclaration(CollisionFilterScope::kOmitDeformable)
> ```

BTW As naive reader, wasn't immediately clear to me why this scope was used until I read below comment.

Would it be possible to reference header mention of rigid-only filtering? e.g.
```
// Per public documentation of SceneGraph, we only filter among rigid geometries.
```
___
*[`geometry/test/geometry_state_test.cc` line 557 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVG3yOCtIwSi0Bs6V9:-NbVG3yOCtIwSi0Bs6VA:btt3q8e) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/test/geometry_state_test.cc#L557)):*
> ```cpp
> 
>   template <typename PropertyType>
>   void AssignRoleToSingleSourceTree(const PropertyType& properties) {
> ```

BTW Is `Tree` qualifier necessary?
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1252 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVGH-K8zgVBK9GcwoB:-NbVGH-K8zgVBK9GcwoC:b-oh04wo) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/multibody/parsing/detail_mujoco_parser.cc#L1252)):*
> ```cpp
> 
>       manager.Apply(geometry::CollisionFilterDeclaration()
>                         .AllowBetween(geometry::GeometrySet({geom1_id}),
> ```

nit Unnecessary white space change?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVEMynBZ6_5gUgFvYD):*

nit Unclear what user-level takeaway is.

My understanding is that users will need to filter any deformable geometries directly through the SceneGraph API; MultibodyPlant API (and corresponding parsing) will only do rigid filtering.

If that is correct, would you be able to add that to this commit summary (or upon merge)?
</blockquote></details>

Done
___
*[`geometry/collision_filter_declaration.h` line 28 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVFokgDlhFIVt3lITJ:-NbVT3hU6LWrQq7YRrS7:b-v2nzvf) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/collision_filter_declaration.h#L28)):*

BTW (curiosity) Why "omit deformable" instead of "rigid only"?
Asking because most of the public documentation seems to qualify by "rigid" rather than "not deformable"
</blockquote></details>

Sean proposed the name and I believe his consideration is that there might be more categories to come and it's easier to do subtraction than addition. My consideration is that we have tried to avoid using the term "rigid geometry" when there might be ambiguity because there's the notion of a "rigid hydroelastic geometry".
___
*[`geometry/geometry_state.cc` line 997 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVF-uY4KdMFv1xQb21:-NbVTuf44_GmJRoptc-Z:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/geometry_state.cc#L997)):*

BTW As naive reader, wasn't immediately clear to me why this scope was used until I read below comment.

Would it be possible to reference header mention of rigid-only filtering? e.g.
```
// Per public documentation of SceneGraph, we only filter among rigid geometries.
```
</blockquote></details>

Done
___
*[`geometry/test/geometry_state_test.cc` line 557 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVG3yOCtIwSi0Bs6V9:-NbVU-B_6XvvarBfAMGw:b-hkp5n) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/test/geometry_state_test.cc#L557)):*

BTW Is `Tree` qualifier necessary?
</blockquote></details>

I don't think it's necessary but it's appropriate IMO.

BTW, this function is not new to this PR. This PR moved it from the private section to the protected section.
___
*[`multibody/parsing/detail_mujoco_parser.cc` line 1252 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVGH-K8zgVBK9GcwoB:-NbVUUVh-zE6dEPsTbeP:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/multibody/parsing/detail_mujoco_parser.cc#L1252)):*

nit Unnecessary white space change?
</blockquote></details>

Done

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19897)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),EricCousineau-TRI(platform)

___
*[`geometry/collision_filter_declaration.h` line 28 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVFokgDlhFIVt3lITJ:-NbW-iwgDiuXRmogLhH2:b56uezr) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/collision_filter_declaration.h#L28)):*

Sean proposed the name and I believe his consideration is that there might be more categories to come and it's easier to do subtraction than addition. My consideration is that we have tried to avoid using the term "rigid geometry" when there might be ambiguity because there's the notion of a "rigid hydroelastic geometry".
</blockquote></details>

OK Sounds good!
___
*[`geometry/test/geometry_state_test.cc` line 557 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19897#-NbVG3yOCtIwSi0Bs6V9:-NbW-eL-1dSAn5xORBRX:biik8ec) ([raw file](https://github.com/RobotLocomotion/drake/blob/c14c26e9ef4c7be86aaa807044880e66b20627c5/geometry/test/geometry_state_test.cc#L557)):*

I don't think it's necessary but it's appropriate IMO.

BTW, this function is not new to this PR. This PR moved it from the private section to the protected section.
</blockquote></details>

OK Ah, sorry about that, and thanks!

PR #19906, opened 2023-08-01
[geometry] Deprecate RenderEngineGltfClient's default_label option

jwnimmer-tri:
Towards #19863.

For the glTF engine, I think the best path forward is to remove the problematic option, and therefore avoid serializing it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_render.cc
+++ b/bindings/pydrake/geometry/geometry_py_render.cc
@@ -3,6 +3,7 @@
  pydrake.geometry module. */
 
 #include "drake/bindings/pydrake/common/default_scalars_pybind.h"
+#include "drake/bindings/pydrake/common/deprecation_pybind.h"
 #include "drake/bindings/pydrake/common/serialize_pybind.h"
 #include "drake/bindings/pydrake/common/value_pybind.h"
 #include "drake/bindings/pydrake/documentation_pybind.h"
@@ -385,6 +386,16 @@ void DoScalarIndependentDefinitions(py::module m) {
     DefAttributesUsingSerialize(&cls, cls_doc);
     DefReprUsingSerialize(&cls);
     DefCopyAndDeepCopy(&cls);
+    // Shim in the vestigial (deprecated) attribute; it's not part of Serialize.
+    // Remove this on 2023-12-01.
+    cls.def_property("default_label",
+        WrapDeprecated(cls_doc.default_label.doc,
+            [](const Class& self) { return self.default_label; }),
+        WrapDeprecated(cls_doc.default_label.doc,
+            [](Class& self, const RenderLabel& value) {
+              self.default_label = value;
+            }),
+        cls_doc.default_label.doc);
   }
 
   m.def("MakeRenderEngineGltfClient", &MakeRenderEngineGltfClient,
--- a/bindings/pydrake/geometry/test/render_test.py
+++ b/bindings/pydrake/geometry/test/render_test.py
@@ -6,6 +6,7 @@ import unittest
 import numpy as np
 
 from pydrake.common.test_utilities import numpy_compare
+from pydrake.common.test_utilities.deprecation import catch_drake_warnings
 from pydrake.common.value import Value
 from pydrake.math import RigidTransform
 from pydrake.systems.framework import (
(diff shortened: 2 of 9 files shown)


SeanCurtis-TRI (commented):
:LGTM:

(With the possibility of a bit more love for C++ users.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-:-NanXL0o0qtBaKphuXes:b-thv374)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 317 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-NanWjO1AACLDDZoZWMM:-NanWjO1AACLDDZoZWMN:b-4oeute) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa738850ac7b11da4941f5679d19655b3dc6844f/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L317)):*
> ```cpp
>                        .default_label = parameters.default_label.value_or(
>                            render::RenderLabel::kDontCare)}),
>       render_client_{std::make_unique<RenderClient>(parameters)} {}
> ```

We could provide a deprecation-like message here. If `parameters.default_label` has a value, then we should give them a one-time warning that we're not supporting it. 
___
*[`geometry/render_gltf_client/render_engine_gltf_client_params.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-NanW79qAzMUf2s-1-q9:-NanW79qAzMUf2s-1-qA:b-hemk9x) ([raw file](https://github.com/RobotLocomotion/drake/blob/c055688fbac4649d67deef16bb777ceec3f5c923/geometry/render_gltf_client/render_engine_gltf_client_params.h#L21)):*
> ```objc
>     a->Visit(DRAKE_NVP(base_url));
>     a->Visit(DRAKE_NVP(render_endpoint));
>     a->Visit(DRAKE_NVP(default_label));
> ```

I like this. We get to excise this directly without fear of breaking any existing yaml files, because no one could actually serialize this anyways.

jwnimmer-tri (commented):
+@sherm1 for platform review per schedule, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-:-Nancxlt830aTwNZ4ZZq:bybe5xg)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

___
*[`geometry/render_gltf_client/internal_render_engine_gltf_client.cc` line 317 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-NanWjO1AACLDDZoZWMM:-NancG0TADxHjHx9LNmq:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/aa738850ac7b11da4941f5679d19655b3dc6844f/geometry/render_gltf_client/internal_render_engine_gltf_client.cc#L317)):*

We could provide a deprecation-like message here. If `parameters.default_label` has a value, then we should give them a one-time warning that we're not supporting it. 
</blockquote></details>

Done

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906)* status: LGTM missing from assignee sherm1(platform)

SeanCurtis-TRI:
BTW I suspect that we can probably treat the other engines the same way. Have you given further thought to that?

jwnimmer-tri:
I'd been going back and forth on it.  There's a very small benefit to keeping the feature enabled in the other engines, but I'm leaning towards deprecating it everywhere:
(1) It's used nowhere in Anzu nor Drake.
(2) Even if we want the config, it seems to me like it should be a SceneGraph setting, not a per-engine-instance setting.
(3) The user can always assign the labels they want.

If you're leaning towards deprecating it everywhere, I can whip up that PR now.

SeanCurtis-TRI:
It seems to me that either it's useful for all engines or no engines. We've had several years for people to care, and it appears they don't care. (No pun intended.)

That said, it's worth noting that the majority of that time has been with `RenderEngineVtk/Gl` which both have `unspecified` as the default value. So, the idea that `RenderEngineGltfClient` has a different (unconfigurable default) probably should be justified.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906)* status: 1 unresolved discussion, LGTM missing from assignee sherm1(platform)

jwnimmer-tri (commented):
After these changes merge, we should probably revisit RenderEngine itself to clean up its use of "unspecified".  I haven't studied carefully, but my guess is that we should deprecate its label constructor argument, and update its documentation to say that only if the user _explicitly_ sets `kUnspecified` as the label will the render engine reject the geometry.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-:-NanvE3h-Po_UjSAsAw2:beqxeqb)* status: LGTM missing from assignee sherm1(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-NangGS_1lPqzYCXes3C:-Nanv3YvD2iaYiE9qMrh:b-dxmzve) (no related file):*

BTW I suspect that we can probably treat the other engines the same way. Have you given further thought to that?
</blockquote></details>

#19911
#19912

___

SeanCurtis-TRI (commented):
It seems to me that if we set every concrete class to default to `kDontCare`, then we might as well make the base class conform.

Back when this was all written, the guiding philosophy was to be extra strict up front in order to give immediate feedback if things weren't configured properly -- with the idea that a "proper" configuration includes "explicit" disposition of all geometry rendered into a label image. At this point, people either haven't cared or they've been well trained. So, loosening the rope and giving people a bit more free rein may not be bad thing.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-:-Nap4IiI5UnxQxYdhH-Q:bvmz571)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
Platform :lgtm:

> Reviewed 6 of 8 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19906#-:-NartxLc29u3Gk6HgZBq:b8qx53o)* status: :shipit: complete! all discussions resolved, LGTM from assignees SeanCurtis-TRI(platform),sherm1(platform)

PR #19920, opened 2023-08-02
Pipes constraints forces through and fixes reaction forces computation

### LINKED ISSUE #13888: Fix MbP's CalcReactionForce() and make it maintainable
MBP has a joint-reaction force output port that uses a private method `CalcReactionForce()` when in need of computation. Reaction forces are essentially a side effect of our dynamics computations, so `CalcReactionForce()` begins by evaluating forward dynamics, for which there are different implementations depending whether we are in discrete or continuous mode. Currently the reaction force algorithm then tries to reproduce the forces that were used in forward dynamics, feed those to inverse dynamics, and extract the reaction forces from there.

Currently `CalcReactionForce()` fails to capture all the forces that were used in discrete mode due to recent independent changes in the discrete model, making it clear that the current structure is not maintainable. There must be a way for the forward dynamics solvers (discrete or continuous) to control the forces that are used for calculating reaction forces so that they are always correct. This is easy in continuous mode since the articulated body algorithm computes reaction forces as a side effect anyway -- we should be taking advantage of that. But we need to engineer a way for the discrete solvers to define the appropriate reaction forces for their implemented models also so this doesn't break again in the future.

Related to #12769.

### LINKED ISSUE #18749: Include arbitrary constraints in reaction forces
The computation of reaction forces essentially boils down to performing a free body diagram and therefore it requires the spatial forces applied on each body. Therefore to include arbitrary constraints in the reaction forces, constraints must be able to translate constraint multipliers to actual spatial forces we can use in this computation. 

Ideally, once the impulses (gamma) are computed as `ContactSolverResults`, we could interrogate constraints to perform this translation for us, from impulses to spatial forces. The DiscreteUpdateManager would then loop through each constraint and accumulate their effect from within `DoCalcDiscreteUpdateMultibodyForces()`.

Therefore what we need is:
  1. API to interrogate arbitrary constraints so that they can tell us the multibody forces that result from a given set of constraint multipliers.
  2. Teach `CompliantContactManager::DoCalcDiscreteUpdateMultibodyForces()` to loop through constraints and accumulate results.

### LINKED ISSUE #19068: Include hydroelastic in the computation of reaction forces
Hydroelastic contact forces are not included in the computation of reaction forces.

PR #18748 attempted to fix this, however we realize that for SAP we  must allow constraints to report multibody forces they contribute (for a given set of multiplieres) and this requires some additional infrastructure. 
For instance, fixing contact reporting will not be enough for models with joint limits (most of them).

### LINKED ISSUE #19435: SAP Infrastructure to report multibody forces
Constraints compute impulses but these need to be mapped into multibody forces in order to perform physical computations such as reaction forces. Specific constraint types are the only ones with enough information to do this.

Forces are applied to physical `objects` (these might be rigid, deformable, a particle, etc). We will introduce this concept in the API. This work will be split into two PRs

### PR 1: SAP solver level
- SapConstraint::SetObjects(), protected, meant to be called by specific constraints at construction.
- SapConstraint::AppendGeneralizedForces(), for things like coupler, stiction.
- SapConstraint::AppendSpatialForces(), for things like contact, weld, distance.
- SapContactProblem::AppendMultibodyForces(), to aggregate all multibody forces given known impulses `gamma`.

## PR 2: Teach manager and SapDriver how to use this to report forces correctly
 - CompliantContactManager::CalcDiscreteUpdateMultibodyForces().
 - Use these forces to compute reaction forces
 - Closes #19068.
   - Unit tests in #18748.
 - Closes #18749.

amcastro-tri:
This PR closes #19435 and #19068
It makes use of the infrastructure at the SAP solver level to report constraint forces within SapDriver so that we can compute the aggregate multibody forces due to constraints. With this information, we can reliably compute reaction forces including all terms in the discrete update.

With this PR we include discrete hydro forces along with every other solver constraint in the computation of reaction forces, effectively resolving #19068.

In summary
 - Closes #18749
 - Closes #19068
 - Closes #19435
 - Closes #13888

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)

### DIFF
--- a/multibody/plant/compliant_contact_manager.cc
+++ b/multibody/plant/compliant_contact_manager.cc
@@ -20,12 +20,12 @@
 #include "drake/multibody/triangle_quadrature/gaussian_triangle_quadrature_rule.h"
 #include "drake/systems/framework/context.h"
 
-using drake::multibody::contact_solvers::internal::ContactConfiguration;
 using drake::geometry::ContactSurface;
 using drake::geometry::GeometryId;
 using drake::geometry::PenetrationAsPointPair;
 using drake::math::RigidTransform;
 using drake::math::RotationMatrix;
+using drake::multibody::contact_solvers::internal::ContactConfiguration;
 using drake::multibody::contact_solvers::internal::ContactSolverResults;
 using drake::multibody::contact_solvers::internal::MatrixBlock;
 using drake::multibody::internal::DiscreteContactPair;
@@ -113,8 +113,8 @@ void CompliantContactManager<T>::DoDeclareCacheEntries() {
   // Cache contact kinematics.
   const auto& contact_kinematics_cache_entry = this->DeclareCacheEntry(
       "Contact kinematics.",
-      systems::ValueProducer(
-          this, &CompliantContactManager::CalcContactKinematics),
+      systems::ValueProducer(this,
+                             &CompliantContactManager::CalcContactKinematics),
       {systems::System<T>::xd_ticket(),
        systems::System<T>::all_parameters_ticket()});
   cache_indexes_.contact_kinematics =
@@ -366,8 +366,7 @@ void CompliantContactManager<T>::AppendDiscreteContactPairsForPointContact(
   const geometry::SceneGraphInspector<T>& inspector = query_object.inspector();
 
   const std::vector<std::vector<int>>& per_tree_unlocked_indices =
-      this->EvalJointLockingCache(context)
-          .unlocked_velocity_indices_per_tree;
+      this->EvalJointLockingCache(context).unlocked_velocity_indices_per_tree;
   const MultibodyTreeTopology& topology = this->internal_tree().get_topology();
 
   // Fill in the point contact pairs.
@@ -422,9 +421,18 @@ void CompliantContactManager<T>::AppendDiscreteContactPairsForPointContact(
       const T phi0 = -pair.depth;
       const T fn0 = k * pair.depth;  // Used by TAMSI, ignored by SAP.
 
-      contact_pairs.push_back({pair.id_A, pair.id_B, p_WC, pair.nhat_BA_W, phi0,
-                              fn0, k, d, tau, mu, {} /* no surface index */,
-                              {} /* no face index */});
+      contact_pairs.push_back({pair.id_A,
+                               pair.id_B,
+                               p_WC,
+                               pair.nhat_BA_W,
+                               phi0,
+                               fn0,
+                               k,
+                               d,
+                               tau,
+                               mu,
+                               {} /* no surface index */,
+                               {} /* no face index */});
     }
   }
 }
@@ -461,8 +469,7 @@ void CompliantContactManager<T>::
       this->EvalContactSurfaces(context);
 
   const std::vector<std::vector<int>>& per_tree_unlocked_indices =
-      this->EvalJointLockingCache(context)
-          .unlocked_velocity_indices_per_tree;
+      this->EvalJointLockingCache(context).unlocked_velocity_indices_per_tree;
   const MultibodyTreeTopology& topology = this->internal_tree().get_topology();
 
   const int num_surfaces = surfaces.size();
@@ -601,6 +608,15 @@ void CompliantContactManager<T>::
 }
 
 template <typename T>
+VectorX<T> CompliantContactManager<T>::CalcEffectiveDamping(
+    const systems::Context<T>& context) const {
+  const VectorX<T> diagonal_inertia =
+      plant().EvalReflectedInertiaCache(context) +
+      joint_damping_ * plant().time_step();
+  return diagonal_inertia;
+}
+
+template <typename T>
 void CompliantContactManager<T>::CalcAccelerationsDueToNonConstraintForcesCache(
     const systems::Context<T>& context,
     AccelerationsDueNonConstraintForcesCache<T>* forward_dynamics_cache) const {
@@ -640,9 +656,7 @@ void CompliantContactManager<T>::CalcAccelerationsDueToNonConstraintForcesCache(
   // below in terms of MultibodyTree APIs.
 
   // We must include reflected rotor inertias along with the new term dt⋅D.
-  const VectorX<T> diagonal_inertia =
-      plant().EvalReflectedInertiaCache(context) +
-      joint_damping_ * plant().time_step();
+  const VectorX<T> diagonal_inertia = CalcEffectiveDamping(context);
 
   // We compute the articulated body inertia including the contribution of the
   // additional diagonal elements arising from the implicit treatment of joint
@@ -863,7 +877,7 @@ void CompliantContactManager<T>::CalcHydroelasticContactInfo(
   const int num_surfaces = all_surfaces.size();
 
   std::vector<SpatialForce<T>> F_Ao_W_per_surface(num_surfaces,
-                                                      SpatialForce<T>::Zero());
+                                                  SpatialForce<T>::Zero());
 
   std::vector<std::vector<HydroelasticQuadraturePointData<T>>> quadrature_data(
       num_surfaces);
@@ -921,8 +935,7 @@ void CompliantContactManager<T>::CalcHydroelasticContactInfo(
 
   const MultibodyTreeTopology& topology = this->internal_tree().get_topology();
   const std::vector<std::vector<int>>& per_tree_unlocked_indices =
-      this->EvalJointLockingCache(context)
-          .unlocked_velocity_indices_per_tree;
+      this->EvalJointLockingCache(context).unlocked_velocity_indices_per_tree;
 
   // Update contact info to include the correct contact forces.
   for (int surface_index = 0; surface_index < num_surfaces; ++surface_index) {
@@ -1111,6 +1124,32 @@ void CompliantContactManager<T>::DoCalcAccelerationKinematicsCache(
       context0, plant().EvalPositionKinematics(context0),
       plant().EvalVelocityKinematics(context0), ac->get_vdot(),
       &ac->get_mutable_A_WB_pool());
+}
+
+template <typename T>
+void CompliantContactManager<T>::DoCalcDiscreteUpdateMultibodyForces(
+    const systems::Context<T>& context, MultibodyForces<T>* forces) const {
+  // Thus far only TAMSI and SAP are supported. Verify this is true.
+  DRAKE_DEMAND(
+      plant().get_discrete_contact_solver() == DiscreteContactSolver::kSap ||
+      plant().get_discrete_contact_solver() == DiscreteContactSolver::kTamsi);
+
+  // Delegate to specific solver driver.
+  if (plant().get_discrete_contact_solver() == DiscreteContactSolver::kSap) {
+    if constexpr (std::is_same_v<T, symbolic::Expression>) {
+      throw std::logic_error(
+          "Discrete updates with the SAP solver are not supported for T = "
+          "symbolic::Expression");
+    } else {
+      DRAKE_DEMAND(sap_driver_ != nullptr);
+      sap_driver_->CalcDiscreteUpdateMultibodyForces(context, forces);
+    }
+  }
+
+  if (plant().get_discrete_contact_solver() == DiscreteContactSolver::kTamsi) {
+    DRAKE_DEMAND(tamsi_driver_ != nullptr);
+    tamsi_driver_->CalcDiscreteUpdateMultibodyForces(context, forces);
+  }
 }
 
 }  // namespace internal
--- a/multibody/plant/compliant_contact_manager.h
+++ b/multibody/plant/compliant_contact_manager.h
@@ -165,6 +165,15 @@ class CompliantContactManager final
 
   void DoDeclareCacheEntries() final;
 
+  // Computes the effective damping matrix D̃ used in discrete schemes treating
+  // damping terms implicitly. This includes joint damping and reflected
+  // inertias. That is, if R is the diagonal matrix of reflected inertias and D
+  // is the diagonal matrix of joint damping coefficients, then the effective
+  // discrete damping term D̃ is: D̃ = R + δt⋅D.
+  // Since D̃ is diagonal this method returns a VectorX with the diagonal
+  // entries only.
+  VectorX<T> CalcEffectiveDamping(const systems::Context<T>& context) const;
+
   // TODO(amcastro-tri): implement these APIs according to #16955.
   // @throws For SAP if T = symbolic::Expression.
   // @throws For TAMSI if T = symbolic::Expression only if the model contains
@@ -177,9 +186,11 @@ class CompliantContactManager final
   void DoCalcAccelerationKinematicsCache(
       const systems::Context<T>&,
       multibody::internal::AccelerationKinematicsCache<T>*) const final;
-  void DoCalcContactResults(
-      const systems::Context<T>&,
-      ContactResults<T>* contact_results) const final;
+  void DoCalcContactResults(const systems::Context<T>&,
+                            ContactResults<T>* contact_results) const final;
+  void DoCalcDiscreteUpdateMultibodyForces(
+      const systems::Context<T>& context,
+      MultibodyForces<T>* forces) const final;
 
   // This method computes sparse kinematics information for each contact pair at
   // the given configuration stored in `context`.
--- a/multibody/plant/discrete_update_manager.cc
+++ b/multibody/plant/discrete_update_manager.cc
@@ -56,13 +56,33 @@ void DiscreteUpdateManager<T>::DeclareCacheEntries() {
   // See ThrowIfNonContactForceInProgress().
   const auto& non_contact_forces_evaluation_in_progress =
       this->DeclareCacheEntry(
-          "Evaluation of non-contact forces and accelerations is in progress.",
+          "Evaluation of non-contact forces and accelerations is in progress",
           // N.B. This flag is set to true only when the computation is in
           // progress. Therefore its default value is `false`.
           systems::ValueProducer(false, &systems::ValueProducer::NoopCalc),
           {systems::System<T>::nothing_ticket()});
   cache_indexes_.non_contact_forces_evaluation_in_progress =
       non_contact_forces_evaluation_in_progress.cache_index();
+
+  MultibodyForces<T> model_forces(plant());
+  const auto& multibody_forces_cache_entry = DeclareCacheEntry(
+      "Discrete update multibody forces",
+      systems::ValueProducer(
+          this, model_forces,
+          &DiscreteUpdateManager<T>::CalcDiscreteUpdateMultibodyForces),
+      {systems::System<T>::xd_ticket(),
+       systems::System<T>::all_parameters_ticket()});
+  cache_indexes_.discrete_update_multibody_forces =
+      multibody_forces_cache_entry.cache_index();
+
+  const auto& contact_results_cache_entry = DeclareCacheEntry(
+      "Contact results (discrete)",
+      systems::ValueProducer(
+          this,
+          &DiscreteUpdateManager<T>::CalcContactResults),
+      {systems::System<T>::xd_ticket(),
+       systems::System<T>::all_parameters_ticket()});
+  cache_indexes_.contact_results = contact_results_cache_entry.cache_index();
 
   // Allow derived classes to declare their own cache entries.
   DoDeclareCacheEntries();
@@ -300,6 +320,41 @@ void DiscreteUpdateManager<T>::CopyForcesFromInputPorts(
   forces->SetZero();
   MultibodyPlantDiscreteUpdateManagerAttorney<T>::AddInForcesFromInputPorts(
       plant(), context, forces);
+}
+
+template <typename T>
+void DiscreteUpdateManager<T>::CalcContactResults(
+    const systems::Context<T>& context,
+    ContactResults<T>* contact_results) const {
+  DRAKE_DEMAND(contact_results != nullptr);
+  plant().ValidateContext(context);
+  DoCalcContactResults(context, contact_results);
+}
+
+template <typename T>
+void DiscreteUpdateManager<T>::CalcDiscreteUpdateMultibodyForces(
+    const systems::Context<T>& context, MultibodyForces<T>* forces) const {
+  plant().ValidateContext(context);
+  DRAKE_DEMAND(forces != nullptr);
+  DRAKE_DEMAND(forces->CheckHasRightSizeForModel(plant()));
+  DoCalcDiscreteUpdateMultibodyForces(context, forces);
+}
+
+template <typename T>
+const MultibodyForces<T>&
+DiscreteUpdateManager<T>::EvalDiscreteUpdateMultibodyForces(
+    const systems::Context<T>& context) const {
+  return plant()
+      .get_cache_entry(cache_indexes_.discrete_update_multibody_forces)
+      .template Eval<MultibodyForces<T>>(context);
+}
+
+template <typename T>
+const ContactResults<T>& DiscreteUpdateManager<T>::EvalContactResults(
+    const systems::Context<T>& context) const {
+  return plant()
+      .get_cache_entry(cache_indexes_.contact_results)
+      .template Eval<ContactResults<T>>(context);
 }
 
 }  // namespace internal
--- a/multibody/plant/discrete_update_manager.h
+++ b/multibody/plant/discrete_update_manager.h
@@ -143,13 +143,10 @@ class DiscreteUpdateManager : public ScalarConvertibleComponent<T> {
     DoCalcDiscreteValues(context, updates);
   }
 
-  /* MultibodyPlant invokes this method to report contact results. */
-  void CalcContactResults(const systems::Context<T>& context,
-                          ContactResults<T>* contact_results) const {
-    DRAKE_DEMAND(contact_results != nullptr);
-    plant().ValidateContext(context);
-    DoCalcContactResults(context, contact_results);
-  }
+  /* Evaluates the contact results used in CalcDiscreteValues() to advance the
+   discrete update from the state stored in `context`. */
+  const ContactResults<T>& EvalContactResults(
+      const systems::Context<T>& context) const;
 
   // Computes all non-contact applied forces including:
   //  - Force elements.
@@ -165,6 +162,13 @@ class DiscreteUpdateManager : public ScalarConvertibleComponent<T> {
   // constraints, rather than raw solver results.
   const contact_solvers::internal::ContactSolverResults<T>&
   EvalContactSolverResults(const systems::Context<T>& context) const;
+
+  /* When the discrete update is performed, see CalcDiscreteValues(), multibody
+   forces applied during that updated are cached. This method returns a
+   reference to the total multibody forces applied during the discrete update.
+   These forces include force elements, constraints and contact forces. */
+  const MultibodyForces<T>& EvalDiscreteUpdateMultibodyForces(
+      const systems::Context<T>& context) const;
 
   /* Publicly exposed MultibodyPlant private/protected methods.
    @{ */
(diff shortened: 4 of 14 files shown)


amcastro-tri (commented):
+@joemasterjohn for feature review please. Thank for helping with this and the final constraints in the fam!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-Nas2rnq6AEY5SIoF_Le:b1o5gar)* status: LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 10 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 555 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NavOCEd1dzfLWwSUJSv:-NavQeKE5iv4rS4Ynccb:bmza7cd) ([raw file](https://github.com/RobotLocomotion/drake/blob/701cd5b3c07bda4aaaf4f6409593bddc6c01d66d/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L555)):*

I couldn't for the life of me find who was throwing with this message and then realized that no-one is throwing it! None of the test configurations below have `with_constraint = true` so this never happens. When added the throw actually happens deep down in whichever is the first constraint to fail to implement `DoAccumulate{Spatial/Generalized}Impulses()` with the error message:
 ```
Constraints must implement this function. See #19435.
```

Please add configurations where `with_constraint = true` and with the corrected error message.  I don't think it's worth it to fail fast up the chain in reaction forces since all of the constraints will soon be reporting forces.
</blockquote></details>

ah! thanks @joemasterjohn. What happened here is that I forgot to remove the `with_cosntraint` configuration. That was only meant for our first pass PR that meant to add hydro forces but failed to add other constraints. Now that all constraints are supported (pending merging current PRs and ball constraint), this is not necessary. I'd much rather prefer to remove it.


Now, we might want to add additional text coverage for cases with additional constraints. Maybe weld and/or joint limits, to ensure different constraints aggregate correctly. 

I'm thinking interesting variations of this test would be:
 1. Replace weld joint with weld constraint!
 2. Replace weld joint with revolute joint and joint limits.

WDYT? would that do?

**Review by joemasterjohn — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 10 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 555 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NavOCEd1dzfLWwSUJSv:-NavupFt0IAaz5kmguuF:b5ye99n) ([raw file](https://github.com/RobotLocomotion/drake/blob/701cd5b3c07bda4aaaf4f6409593bddc6c01d66d/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L555)):*

ah! thanks @joemasterjohn. What happened here is that I forgot to remove the `with_cosntraint` configuration. That was only meant for our first pass PR that meant to add hydro forces but failed to add other constraints. Now that all constraints are supported (pending merging current PRs and ball constraint), this is not necessary. I'd much rather prefer to remove it.


Now, we might want to add additional text coverage for cases with additional constraints. Maybe weld and/or joint limits, to ensure different constraints aggregate correctly. 

I'm thinking interesting variations of this test would be:
 1. Replace weld joint with weld constraint!
 2. Replace weld joint with revolute joint and joint limits.

WDYT? would that do?

</blockquote></details>

Those would be interesting, but might take a bit of effort to make sure verifying the expected forces are correct. I was thinking of something much simpler:
- Evaluate reaction forces on a candidate system without constraints
- Evaluate reaction forces on the same model but with a single constraint added.
- Check that the reaction forces for objects not involved in the constraint don't change
- Check that the reaction forces for the objects involved differ in a way that we expect (without checking physics)
  (i.e. for a weld constraint verify that the difference in reaction forces acting on body A is equal and opposite to the difference in reaction forces acting on body B, irregardless of the physical interpretation of those forces.)

What do you think about that?

**Review by amcastro-tri — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 10 unresolved discussions, LGTM missing from assignee joemasterjohn, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 555 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NavOCEd1dzfLWwSUJSv:-NavzjNK97L7sMGxc1xs:b-he1cu2) ([raw file](https://github.com/RobotLocomotion/drake/blob/701cd5b3c07bda4aaaf4f6409593bddc6c01d66d/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L555)):*

Those would be interesting, but might take a bit of effort to make sure verifying the expected forces are correct. I was thinking of something much simpler:
- Evaluate reaction forces on a candidate system without constraints
- Evaluate reaction forces on the same model but with a single constraint added.
- Check that the reaction forces for objects not involved in the constraint don't change
- Check that the reaction forces for the objects involved differ in a way that we expect (without checking physics)
  (i.e. for a weld constraint verify that the difference in reaction forces acting on body A is equal and opposite to the difference in reaction forces acting on body B, irregardless of the physical interpretation of those forces.)

What do you think about that?
</blockquote></details>

not sure, I'd need a specific example.
Also, I was thinking the variants I proposed would not be hard because in steady state the results should always be the same, regardless of the constraint we used.

In any case, regardless of the specific test, it'd seem we are on the same page.

**Review by joemasterjohn — COMMENTED** [review]
:lgtm: CI is timing out on the `DiscretePointSap` in a debug build, might need to reduce the simulation time from 10 seconds. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbQOOxJ-QHP4Ilh8_Kz:b-viub3a)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 565 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQMHZJ1X6suGFt3X8e:-NbQMHZJ1X6suGFt3X8f:bufeuym) ([raw file](https://github.com/RobotLocomotion/drake/blob/3206696f67f05c876ad63862fa8ac59740f56b0c/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L565)):*
> ```cpp
> 
> TEST_P(LadderTest, TestWithThreads) {
>   // const LadderTestConfig& config = GetParam();
> ```

nit: remove this unused line



_Code quote:_
```cpp
  // const LadderTestConfig& config = GetParam();
```

amcastro-tri (commented):
Lets see, I made the tests smaller by narrowing the sim time in AdvanceTo().

+@jwnimmer-tri for platform review please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbQXxLbEzWkb5FkicRr:b-usid32)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri (commented):
Checkpoint on the code.  I'll circle back to review the new test soon.

> Reviewed 8 of 13 files at r1, 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbQqGdfBRfQAOJS2sw1:by763vi)* status: 9 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQly6l0ScE70p2Pto0:-NbQly6l0ScE70p2Pto1:b-txb9z6) (no related file):*
BTW I want to double-check that the feature reviewers are happy with the testing strategy.

There is a lot of new code here in multiple different classes, yet the only new tests we have look to be end-to-end acceptance tests that probe only at the outermost MbP API.

Do these tests offer sufficient coverage of correctness?

Are these tests focused enough to help debug problems in one of the more narrow classes (e.g., sap_driver).  As it stands, the user needs to run the whole MbP when something goes wrong, and try to trace through where some logic bug deep down in SAP happened.  If there were for example SAP-specific test coverage

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQowpf39ZXynMLVhMI:-NbQowpf39ZXynMLVhMJ:b1pgmgp) (no related file):*
BTW You might want to ask for an extra set of feature-review eyes (Sherm) on the `multibody_plant.cc` changes to dependency tickets.  I don't know that I would be able to myself spot any defects in those declarations, and they are not enforce at runtime, rather only showing up as subtle math errors in outputs.

___
___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQoKjZA4-2cQCOmGcT):*
nit This commit title seems out-of-date with the PR title.  Which one shall we used for merging it?

Also there is a typo in this message: `MBP` is not a thing that exists in Drake; `MbP` is the thing that exists.
___
*[`multibody/plant/compliant_contact_manager.cc` line 1148 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQpkL43iAmw8dfU6dy:-NbQpkL55P82HyzzL2Ms:bly44n6) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/compliant_contact_manager.cc#L1148)):*
> ```cpp
>     tamsi_driver_->CalcDiscreteUpdateMultibodyForces(context, forces);
>   }
> }
> ```

If the contact solver is neither kSap nor kTamsi, it seems like we silently fail to calculate anything.
___
*[`multibody/plant/multibody_plant.h` line 1835 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQaAvJBwazpmCrgKed:-NbQaAvJBwazpmCrgKee:bw30yjk) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/multibody_plant.h#L1835)):*
> ```objc
> 
>   /// Sets the contact solver type used for discrete %MultibodyPlant models.
>   /// @warning no-op for continuous models (is_discrete() is false.)
> ```

BTW My preference would be to use a complete sentence here, so it reads more smoothly on our Doxygen website.
___
*[`multibody/plant/multibody_plant.cc` line 2855 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQaaek2bVw7Dz2kcUT:-NbQaaek2bVw7Dz2kcUU:bbpjxsc) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/multibody_plant.cc#L2855)):*
> ```cpp
>     return tickets;
>   }();
>   auto& contact_results_cache_entry = this->DeclareCacheEntry(
> ```

BTW It seems like for a discrete time plant, it might be plausible to not even declare this cache entry.  That would ensure that any MbP code that accidentally tried to use it would fail-fast, instead of computing wrong answers.

Another option would be to have `CalcContactResultsContinuous` fail-fast when called on a discrete plant.
___
*[`multibody/plant/sap_driver.cc` line 850 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQnNUz7-QgzXL2Cxz2:-NbQnNUz7-QgzXL2Cxz3:b-nf3h6x) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/sap_driver.cc#L850)):*
> ```cpp
>   // Include the contribution from constraints.
>   // TODO(amcastro-tri): Consider deformables.
>   if constexpr (std::is_same_v<T, double>) {
> ```

I don't understand why we need the `if constexpr` guard here.  It seems like having a non-null driver is always an error, no matter the scalar type.  This _looks_ like we are trying to silently ignore non-empty deformables when doing autodiff.
___
*[`multibody/plant/tamsi_driver.cc` line 377 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQk077-QLXYt0a-jhy:-NbQk077-QLXYt0a-jhz:b-svyx74) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/tamsi_driver.cc#L377)):*
> ```cpp
>   manager().CalcNonContactForces(context, include_joint_limit_penalty_forces,
>                                  forces);
>   auto& Fapplied_Bo_W_array = forces->mutable_body_forces();
> ```

nit For my taste, moving this statement down line line (after the contact results eval) would be a lot more readable. It would clarify that all we're doing is preparing an output alias for the Calc call, and match the argument order of the function application with the order we prepare our local variables.

jwnimmer-tri (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbQzMpE8gHoHhiH7Lxh:bnfp4nl)* status: 12 unresolved discussions

___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 260 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQyvZYCO75Z1SjYAZE:-NbQyvZYCO75Z1SjYAZF:b-bsoywg) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L260)):*
> ```cpp
>       }
>       case LadderTestConfig::WeldMethod::kRevoluteJointWithLimits: {
>         const double kInfty = std::numeric_limits<double>::infinity();
> ```

nit "Do not abbreviate by deleting letters within a word"

https://drake.mit.edu/styleguide/cppguide.html#General_Naming_Rules



_Suggestion:_
```cpp
kInf
```
___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 550 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQz6mQCof1REFa05AI:-NbQz6mQCof1REFa05AJ:b-3407gd) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L550)):*
> ```cpp
>   // is chosen consistently with the time the system is left to reach steady
>   // state and the (continuous) integration accuracy (for the continuous model).
>   const double kTolerance = 1.0e-9;
> ```

nit Be consistent



_Suggestion:_
```cpp
kTolerance{1.0e-9}
```
___
*[`multibody/plant/test/multibody_plant_reaction_forces_test.cc` line 571 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQzDx59Xh83Qg4e8Qg:-NbQzDx59Xh83Qg4e8Qh:b6za4q) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/test/multibody_plant_reaction_forces_test.cc#L571)):*
> ```cpp
> }
> 
> // Setup test cases using point and hydroelastic contact.
> ```

nit "Set up" is the verb form; "Setup" is a noun.



_Suggestion:_
```cpp
Set up
```

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 12 unresolved discussions

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQly6l0ScE70p2Pto0:-NbR1RcXEmXGhu9vr_G0:b-1jfphv) (no related file):*

BTW I want to double-check that the feature reviewers are happy with the testing strategy.

There is a lot of new code here in multiple different classes, yet the only new tests we have look to be end-to-end acceptance tests that probe only at the outermost MbP API.

Do these tests offer sufficient coverage of correctness?

Are these tests focused enough to help debug problems in one of the more narrow classes (e.g., sap_driver).  As it stands, the user needs to run the whole MbP when something goes wrong, and try to trace through where some logic bug deep down in SAP happened.  If there were for example SAP-specific test coverage
</blockquote></details>

I'm ok with the test coverage and strategy here. There is really one important internal API added `EvalDiscreteUpdateMultibodyForces()`, any other internal API added is just a helper function in the specific driver classes used to implement `CalcDiscreteUpdateMultibodyForces`. The other things are moving calc quantities into cache entries and code shuffle.

The only public facing API that interacts with the new internal API is `MbP::CalcReactionForces()` and that is thoroughly covered here. To me a test at the specific driver level would either look much like the reaction forces tests (with the additional need of making plant API calls anyway) or not sufficiently test that the reported forces from the new API actually match the physics happening in the discrete update.

___

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 3 unresolved discussions, LGTM missing from assignee sherm1(platform)

___
*[`multibody/plant/compliant_contact_manager.cc` line 1148 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQpkL43iAmw8dfU6dy:-NbpJHSUAAwMqwVOSVUF:b5yuuf7) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/compliant_contact_manager.cc#L1148)):*

If the contact solver is neither kSap nor kTamsi, it seems like we silently fail to calculate anything.
</blockquote></details>

good point. Added a DEMAND at the top.
___
*[`multibody/plant/multibody_plant.cc` line 2855 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQaaek2bVw7Dz2kcUT:-NbpK1z01IgoSUtCU9W8:b-o8ecs1) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/multibody_plant.cc#L2855)):*

BTW It seems like for a discrete time plant, it might be plausible to not even declare this cache entry.  That would ensure that any MbP code that accidentally tried to use it would fail-fast, instead of computing wrong answers.

Another option would be to have `CalcContactResultsContinuous` fail-fast when called on a discrete plant.
</blockquote></details>

godo point. I implemented both suggestions.
___
*[`multibody/plant/sap_driver.cc` line 850 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQnNUz7-QgzXL2Cxz2:-NbpK_AX7QTk_6IL_FmB:bc08ucx) ([raw file](https://github.com/RobotLocomotion/drake/blob/64bced464e889ffd3d15b0e3b8c77bab695e4abf/multibody/plant/sap_driver.cc#L850)):*

I don't understand why we need the `if constexpr` guard here.  It seems like having a non-null driver is always an error, no matter the scalar type.  This _looks_ like we are trying to silently ignore non-empty deformables when doing autodiff.
</blockquote></details>

ah.. nice catch. I just copy/pasted a similar check without thinking through the implications. You are, obviously, correct.

jwnimmer-tri (commented):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: LGTM missing from assignee sherm1(platform)

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: LGTM missing from assignee sherm1(platform)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbQly6l0ScE70p2Pto0:-NbpOd7N1VipA4iiQkvL:b-j8a7zx) (no related file):*

I'm ok with the test coverage and strategy here. There is really one important internal API added `EvalDiscreteUpdateMultibodyForces()`, any other internal API added is just a helper function in the specific driver classes used to implement `CalcDiscreteUpdateMultibodyForces`. The other things are moving calc quantities into cache entries and code shuffle.

The only public facing API that interacts with the new internal API is `MbP::CalcReactionForces()` and that is thoroughly covered here. To me a test at the specific driver level would either look much like the reaction forces tests (with the additional need of making plant API calls anyway) or not sufficiently test that the reported forces from the new API actually match the physics happening in the discrete update.
</blockquote></details>

Thanks @jwnimmer-tri. I gave your suggestion pretty good thought. However I found out that it'd pretty much require rewriting the tests pushed in this PR or already in master.

A large portion of this PR is about refactoring existing code so that we exercise the new pipeline (that we've been pushing throughout last quarter, this is just the culmination of that effort) that enables to consolidate reaction forces while making reaction forces maintainable in discrete mode. Therefore this functionality is already unit tested in `multibody_plant_reaction_forces_test.cc`, `sap_driver_test.cc`, `tamsi_driver.cc`, `deformable_driver_contact_test.cc` (and other minor tests). It'd be difficult to break our implementation without any of these tests noticing (though obviously I imagine it is not impossible).

This in turn also fixes the report of discrete hydro forces when computing reaction forces. This new bug fix is now unit tested in the new tests included in `multibody_plant_reaction_forces_test.cc`. 
The proper reporting of forces by new constraints is something that is tested at a constraint level and in the corresponding `SapContactProblem` unit tests.

Any suggestions on how to improve our testing strategy are obviously welcome.

___

amcastro-tri:
@drake-jenkins-bot mac-arm-ventura-clang-bazel-experimental-release please

joemasterjohn (commented):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: LGTM missing from assignee sherm1(platform)

sherm1 (commented):
I looked at the dependency tickets. They look fine as long as the intent is to _not_ depend on input ports at all. PTAL
(Partial) feature :lgtm: pending confirmation or fix.

> Reviewed 1 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbpjLCoEwLspBgrF4Os:by7ie26)* status: 2 unresolved discussions

___
*[`multibody/plant/multibody_plant.cc` line 2783 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-Nbpfumu6iGtBEqFkRAc:-Nbpfumu6iGtBEqFkRAd:b-zguo7z) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d832b3f68212454fb96d4f43b6a3177415d1c77/multibody/plant/multibody_plant.cc#L2783)):*
> ```cpp
>               "contact_results", ContactResults<T>(),
>               &MultibodyPlant<T>::CopyContactResultsOutput,
>               {state_ticket, systems::System<T>::all_parameters_ticket()})
> ```

BTW this says the contact result output depends on only on state x and parameters p, but not on input ports u. Is that the intent?
___
*[`multibody/plant/multibody_plant.cc` line 2859 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbpiDJC6peyM4FuAGzh:-NbpiDJC6peyM4FuAGzi:b-mr0wmd) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d832b3f68212454fb96d4f43b6a3177415d1c77/multibody/plant/multibody_plant.cc#L2859)):*
> ```cpp
>     auto& contact_results_cache_entry = this->DeclareCacheEntry(
>         std::string("Contact results (continuous)"),
>         &MultibodyPlant<T>::CalcContactResultsContinuous, {dependency_ticket});
> ```

BTW this one says q,v,p and whatever contact_info_and_body_spatial_forces depends on (which is also q,v,p). Doesn't include input ports. Should it?

amcastro-tri (commented):
Good call, thanks @sherm1. 
This makes me think we might have forgotten the input port dependencies at other places. I fixed the ones you pointed out here, but to avoid clutter I'll push related fixes separately. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-Nbtk0yJ6sLh3AU27t2F:b4xtwq8)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform),joemasterjohn

___
*[`multibody/plant/multibody_plant.cc` line 2783 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-Nbpfumu6iGtBEqFkRAc:-NbtjynH593zCPMY8Yxq:bfes19g) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d832b3f68212454fb96d4f43b6a3177415d1c77/multibody/plant/multibody_plant.cc#L2783)):*

BTW this says the contact result output depends on only on state x and parameters p, but not on input ports u. Is that the intent?
</blockquote></details>

yes, added.
___
*[`multibody/plant/multibody_plant.cc` line 2859 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbpiDJC6peyM4FuAGzh:-NbtjumQAl9WwhoeBD6a:b8tq58e) ([raw file](https://github.com/RobotLocomotion/drake/blob/1d832b3f68212454fb96d4f43b6a3177415d1c77/multibody/plant/multibody_plant.cc#L2859)):*

BTW this one says q,v,p and whatever contact_info_and_body_spatial_forces depends on (which is also q,v,p). Doesn't include input ports. Should it?
</blockquote></details>

Good call, added.

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-Nbtkcn5E9JKMPImapkH:-Nbtkcn6BgvujDexWIoF:b-vmjh8e) (no related file):*
Not sure if the `mac-arm-ventura-clang-bazel-experimental-release please` is caused by this PR (in model_visualizer_test), but it consistently happened twice. Is this a known issue or should I investigate further?
@jwnimmer-tri, do you have any pointers?

___

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920)* status: 2 unresolved discussions

___
*[`multibody/plant/test/multibody_plant_test.cc` line 1723 at r8](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-NbuTNT_C6PVVlycODFo:-NbuTNT_C6PVVlycODFp:b-ikm0ex) ([raw file](https://github.com/RobotLocomotion/drake/blob/09925bc0fd02be2f10549f99625c327d0a225c6f/multibody/plant/test/multibody_plant_test.cc#L1723)):*
> ```cpp
> // @returns `true` iff only if a closed subset of the ports is direct
> // feedthrough.
> bool VerifyFeedthroughPorts(
> ```

nit: This namechange wasn't propagated throughout this whole test file, which is causing CI failure.



_Code quote:_
```cpp
bool VerifyFeedthroughPorts(
```

jwnimmer-tri (commented):
I suggest that as a follow up, for each callback function that we register with the framework (i.e., outputs or cache entries), we consistently document atop the function which tickets its declared to use.  Then, we have a three-prong checklist to review, that gives us the full proof we need:
- For each `this->Declare(&CalcFunction, tickets)` line, the `tickets` is congruent with the comment atop `CalcFunction` implementation.
- For each `CalcFunction` body, any data it accesses from the context is listed in the comment atop it's implementation.
- For each `CalcFunction` body, any helper functions it calls has the same or subset of tickets -- compare ticket docs atop the caller and callee functions.
The checklist only applies to functions that accept the full Context.  Functions that accept narrower data (that consistent with the tickets from each call site) are OK on their own without digging in to them.

> Reviewed 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19920#-:-NbuXQd7CIErHBRjqHNI:b-jvql9p)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform),joemasterjohn

PR #19924, opened 2023-08-03
Add SchurComplement

xuchenhan-tri:
The SchurComplement class computes the Schur complement of a given 3x3 block sparse symmetric matrix `A` based on the block indices to be eliminated. It also supports solving the linear system Az = b by leveraging the Cholesky factorization obtained in the process of computing the Schur complement.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)

### DIFF
--- a/multibody/contact_solvers/BUILD.bazel
+++ b/multibody/contact_solvers/BUILD.bazel
@@ -30,6 +30,7 @@ drake_cc_package_library(
         ":newton_with_bisection",
         ":pgs_solver",
         ":point_contact_data",
+        ":schur_complement",
         ":sparse_linear_operator",
         ":supernodal_solver",
         ":system_dynamics_data",
@@ -196,6 +197,17 @@ drake_cc_library(
 )
 
 drake_cc_library(
+    name = "schur_complement",
+    srcs = ["schur_complement.cc"],
+    hdrs = ["schur_complement.h"],
+    deps = [
+        ":block_sparse_cholesky_solver",
+        ":block_sparse_lower_triangular_or_symmetric_matrix",
+        "//common:essential",
+    ],
+)
+
+drake_cc_library(
     name = "sparse_linear_operator",
     srcs = ["sparse_linear_operator.cc"],
     hdrs = ["sparse_linear_operator.h"],
@@ -336,6 +348,15 @@ drake_cc_googletest(
         ":pgs_solver",
         ":sparse_linear_operator",
         "//common/test_utilities:eigen_matrix_compare",
+    ],
+)
+
+drake_cc_googletest(
+    name = "schur_complement_test",
+    deps = [
+        ":schur_complement",
+        "//common/test_utilities:eigen_matrix_compare",
+        "//common/test_utilities:expect_throws_message",
     ],
 )
 
--- a/multibody/contact_solvers/schur_complement.cc
+++ b/multibody/contact_solvers/schur_complement.cc
@@ -1,0 +1,110 @@
+#include "drake/multibody/contact_solvers/schur_complement.h"
+
+#include <algorithm>
+#include <optional>
+#include <utility>
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+SchurComplement::SchurComplement() = default;
+
+SchurComplement::~SchurComplement() = default;
+
+SchurComplement::SchurComplement(const Block3x3SparseSymmetricMatrix& A,
+                                 const std::unordered_set<int>& D_indices)
+    : D_indices_(D_indices.begin(), D_indices.end()) {
+  /* Keep D_indices_ sorted. */
+  DRAKE_THROW_UNLESS(ssize(D_indices) <= A.block_cols());
+  std::sort(D_indices_.begin(), D_indices_.end());
+  /* If a block index doesn't belong to the D blocks, it belongs to the C
+   blocks. We step through `D_indices_` to detect the gaps in order to fill in
+   `C_indices`. */
+  /* Starting point for each potential missing segment in D_indices_. */
+  int start = 0;
+  for (int D_index : D_indices_) {
+    while (start < D_index) {
+      C_indices_.push_back(start++);
+    }
+    start = D_index + 1;
+  }
+  /* Add the remaining indices (if any) after the last D index. */
+  while (start < A.block_cols()) {
+    C_indices_.push_back(start++);
+  }
+
+  const int block_cols = C_indices_.size() + D_indices_.size();
+  DRAKE_DEMAND(block_cols * 3 == A.cols());
+  std::optional<MatrixX<double>> S =
+      A_solver_.FactorAndCalcSchurComplement(A, D_indices);
+  if (!S) {
+    throw std::runtime_error(
+        "Factorization failed when computing Schur complement. Make sure the "
+        "matrix is symmetric positive definite and not ill-conditioned.");
+  }
+  DRAKE_DEMAND(
+      A_solver_.solver_mode() ==
+      BlockSparseCholeskySolver<Matrix3<double>>::SolverMode::kFactored);
+  S_ = std::move(*S);
+}
+
+/* TODO(xuchenhan-tri): In the following implementation, we construct `a`, the
+ segment in the right-hand side corresponding to participating dofs with
+ `a = S*y`. From `a`, we construct the full right-hand side `b`, then solve the
+ entire system for `z`, and finally extract `x` from `z`. Alternatively, we
+ could directly compute `x` from the equation Dx + By = 0. This would require
+   1. a representation of `B`, and
+   2. the ability to solve the linear system `D`.
+ Both of these is avaiable, just not conveniently so:
+ We had access to `A` at construction, and we could have saved a sparse
+ representation of `B` and use it to efficiently compute `By`. From the
+ factorization of `A`, we get a factorization of D as a by-product. So far, our
+ profiling in the common case where size(D) is much smaller than size(A) shows
+ that the current implementation is not a performance issue, so we have opted
+ for simplicity in the implementation. */
+VectorX<double> SchurComplement::SolveForX(
+    const Eigen::Ref<const VectorX<double>>& y) const {
+  DRAKE_THROW_UNLESS(y.size() == 3 * ssize(C_indices_));
+  DRAKE_DEMAND(
+      A_solver_.solver_mode() ==
+      BlockSparseCholeskySolver<Matrix3<double>>::SolverMode::kFactored);
+  if (D_indices_.empty()) {
+    return VectorX<double>::Zero(0);
+  }
+  if (C_indices_.empty()) {
+    return VectorX<double>::Zero(D_indices_.size());
+  }
+
+  /* Build `a`, the rhs corrsponding to variable y. */
+  const VectorX<double> a = S_ * y;
+  /* Build the full rhs. */
+  const int block_cols = C_indices_.size() + D_indices_.size();
+  VectorX<double> b(VectorX<double>::Zero(3 * block_cols));
+  for (int i = 0; i < ssize(C_indices_); ++i) {
+    b.segment<3>(3 * C_indices_[i]) = a.segment<3>(3 * i);
+  }
+  VectorX<double>& z = b;
+  A_solver_.SolveInPlace(&z);
+  /* Extract x from the z. */
+  VectorX<double> x(3 * D_indices_.size());
+  for (int i = 0; i < ssize(D_indices_); ++i) {
+    x.segment<3>(3 * i) = z.segment<3>(3 * D_indices_[i]);
+  }
+  return x;
+}
+
+VectorX<double> SchurComplement::Solve(
+    const Eigen::Ref<const VectorX<double>>& c) const {
+  DRAKE_THROW_UNLESS(3 * (ssize(C_indices_) + ssize(D_indices_)) == c.size());
+  DRAKE_DEMAND(
+      A_solver_.solver_mode() ==
+      BlockSparseCholeskySolver<Matrix3<double>>::SolverMode::kFactored);
+  return A_solver_.Solve(c);
+}
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
--- a/multibody/contact_solvers/schur_complement.h
+++ b/multibody/contact_solvers/schur_complement.h
@@ -1,0 +1,132 @@
+#pragma once
+
+#include <unordered_set>
+#include <vector>
+
+#include "drake/common/drake_copyable.h"
+#include "drake/common/eigen_types.h"
+#include "drake/multibody/contact_solvers/block_sparse_cholesky_solver.h"
+#include "drake/multibody/contact_solvers/block_sparse_lower_triangular_or_symmetric_matrix.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+
+/* Given a symmetric linear system of equations Az = b in block form as:
+
+     Dx  + By = 0     (1)
+     Bᵀx + Cy = a     (2)
+
+ where A = [D B; Bᵀ C], z = [x; y], b = [0; a]. If C (size p-by-p),
+ D (size q-by-q) and A (size p+q-by-p+q) are positive definite, one can solve
+ the system using Schur complement. Specifically, using equation (1), we get
+
+     x = -D⁻¹By       (3)
+
+ Plugging (3) in (1), we get
+
+    (C - BᵀD⁻¹B)y = a.
+
+ After a solution for y is obtained, we can use (3) to recover the solution for
+ x. The matrix S = C - BᵀD⁻¹B is the Schur complement of the block D of the
+ matrix A. Since A is positive definite, so is S.
+
+ The decomposition of A into [D, B; Bᵀ, C] is defined by the block indices of D.
+ (See the two-argument constructor for more detail.) Given the symmetric matrix
+ A and its decomposition [D, B; Bᵀ, C], this class computes the Schur complement
+ of the block D of A. The matrix is factorized in the process and consequently,
+ one can solve the system Az = b efficiently once the Schur complement is
+ computed. */
+class SchurComplement {
+ public:
+  DRAKE_DEFAULT_COPY_AND_MOVE_AND_ASSIGN(SchurComplement);
+
+  /* Constructs an empty SchurComplement, i.e one that corresponds
+   corresponds to a linear system with no equations.*/
+  SchurComplement();
+
+  ~SchurComplement();
+
+  /* Constructs a SchurComplement for the block sparse matrix A of size
+   3*N-by-3*N consisting of blocks of size 3x3. `D_indices` determine which
+   block rows/columns make up the D diagonal blocks. The remaining block
+   rows/columns make up the C diagonal blocks. For example, if matrix A is given
+   by
+
+     x x x | u u u | w w w
+     x x x | u u u | w w w
+     x x x | u u u | w w w
+     ------+-------+------
+     u u u | y y y | v v v
+     u u u | y y y | v v v
+     u u u | y y y | v v v
+     ------+-------+------
+     w w w | v v v | z z z
+     w w w | v v v | z z z
+     w w w | v v v | z z z
+
+   and D_indices = {0, 2},  then submatrix D is given by
+
+     x x x | w w w
+     x x x | w w w
+     x x x | w w w
+     ------+------
+     w w w | z z z
+     w w w | z z z
+     w w w | z z z
+
+   submatrix C is given by
+
+     y y y
+     y y y
+     y y y
+
+   submatrix B is given by
+
+     u u u
+     u u u
+     u u u
+     -----
+     v v v
+     v v v
+     v v v
+
+   @pre D_indices is a subset of {0, ..., A.block_cols()-1}. */
+  SchurComplement(const Block3x3SparseSymmetricMatrix& A,
+                  const std::unordered_set<int>& D_indices);
+
+  /* Returns the Schur complement for the block D of the matrix A,
+   S = C - BᵀD⁻¹B. */
+  const MatrixX<double>& get_D_complement() const { return S_; }
+
+  /* Given a value of y, solves for x in the equation Dx + By = 0.
+   @pre The size of y is equal to the number of columns of B */
+  VectorX<double> SolveForX(const Eigen::Ref<const VectorX<double>>& y) const;
+
+  // TODO(xuchenhan-tri): Consider adding a SolveForY() function for
+  // completeness.
+
+  /* Given a right-hand side vector b with the same dimension as the input
+   matrix A provided at construction, solve solves for A*z = b.
+   @pre The size of b is compatible with the input matrix A provided at
+   construction. */
+  VectorX<double> Solve(const Eigen::Ref<const VectorX<double>>& b) const;
+
+ private:
+  /* Sorted block row/column indices for 3x3 blocks that belong to submatrix D.
+   */
+  std::vector<int> D_indices_;
+  /* Sorted block row/column indices for 3x3 blocks that belong to submatrix C.
+   */
+  std::vector<int> C_indices_;
+  /* Cholesky solver that factorizes the matrix and stores the factorization. */
+  BlockSparseCholeskySolver<Matrix3<double>> A_solver_;
+  /* The Schur complement of block D: S = C - BᵀD⁻¹B. */
+  MatrixX<double> S_;
+};
+
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake
--- a/multibody/contact_solvers/test/schur_complement_test.cc
+++ b/multibody/contact_solvers/test/schur_complement_test.cc
@@ -1,0 +1,134 @@
+#include "drake/multibody/contact_solvers/schur_complement.h"
+
+#include <gtest/gtest.h>
+
+#include "drake/common/test_utilities/eigen_matrix_compare.h"
+#include "drake/common/test_utilities/expect_throws_message.h"
+
+namespace drake {
+namespace multibody {
+namespace contact_solvers {
+namespace internal {
+namespace {
+
+using Eigen::Matrix3d;
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
+
+constexpr double kTolerance = 4.0 * std::numeric_limits<double>::epsilon();
+
+// clang-format off
+Matrix3d A00() {
+  return (Eigen::Matrix3d() << 11, 2, 3,
+                               2, 50, 6,
+                               3, 6, 190).finished();
+}
+Matrix3d A10() {
+  return (Eigen::Matrix3d() << 1.1, 1.0, 0.9,
+                               1.2, 1.5, 1.6,
+                               1.3, 1.8, 1.9).finished();
+}
+Matrix3d A11() {
+    return (Eigen::Matrix3d() << 110, 12, 13,
+                                 12, 150, 16,
+                                 13, 16, 190).finished();
+}
+Matrix3d A20() {
+    return (Eigen::Matrix3d() << 11, 12, 13,
+                                 14, 21, 16,
+                                 17, 18, 29).finished();
+}
+Matrix3d A22() {
+    return (Eigen::Matrix3d() << 210, 22, 23,
+                                 22, 205, 26,
+                                 23, 26, 290).finished();
+}
+// clang-format on
+
+/* Makes a block sparse symmetric matrix with 3x3 dense blocks that looks like
+   A =   A00 | A01 | A02
+        -----------------
+         A10 | A11 |  0
+        -----------------
+         A20 |  0  | A22
+ where A02 = A20.transpose(). We choose values so that A is diagonally dominant
+ and thus SPD. */
+Block3x3SparseSymmetricMatrix MakeBlockSparseMatrix() {
+  std::vector<std::vector<int>> sparsity_pattern;
+  sparsity_pattern.emplace_back(std::vector<int>{0, 1, 2});
+  sparsity_pattern.emplace_back(std::vector<int>{1});
+  sparsity_pattern.emplace_back(std::vector<int>{2});
+  BlockSparsityPattern block_pattern({{3, 3, 3}}, std::move(sparsity_pattern));
+  Block3x3SparseSymmetricMatrix A(std::move(block_pattern));
+  A.SetBlock(0, 0, A00());
+  A.SetBlock(1, 0, A10());
+  A.SetBlock(2, 0, A20());
+  A.SetBlock(1, 1, A11());
+  A.SetBlock(2, 2, A22());
+  return A;
+}
+
+/* Constructs an arbitrary block sparse matrix (see MakeBlockSparseMatrix())
+
+   A =   A00 | A01 | A02
+        -----------------
+         A10 | A11 |  0
+        -----------------
+         A20 |  0  | A22
+
+and returns the Schur complement of the A11 block. */
+SchurComplement MakeSchurComplement() {
+  Block3x3SparseSymmetricMatrix block_sparse_matrix = MakeBlockSparseMatrix();
+  const std::unordered_set<int> eliminated_blocks = {1};
+  return SchurComplement(block_sparse_matrix, eliminated_blocks);
+}
+
+GTEST_TEST(SchurComplementTest, GetDComplement) {
+  const SchurComplement schur_complement = MakeSchurComplement();
+  const MatrixXd S = schur_complement.get_D_complement();
+
+  MatrixXd C = MatrixXd(6, 6);
+  C.topLeftCorner<3, 3>() = A00();
+  C.bottomLeftCorner<3, 3>() = A20();
+  C.topRightCorner<3, 3>() = A20().transpose();
+  C.bottomRightCorner<3, 3>() = A22();
+  MatrixXd D = A11();
+  MatrixXd B = MatrixXd::Zero(3, 6);
+  B.topLeftCorner<3, 3>() = A10();
+
+  const MatrixXd expected = C - B.transpose() * D.llt().solve(B);
+  EXPECT_TRUE(CompareMatrices(S, expected, kTolerance));
+}
+
+GTEST_TEST(SchurComplementTest, SolveForX) {
+  const SchurComplement schur_complement = MakeSchurComplement();
+  const VectorXd y = VectorXd::LinSpaced(6, 0.0, 12.0);
+  const VectorXd x = schur_complement.SolveForX(y);
+
+  MatrixXd D = A11();
+  MatrixXd B = MatrixXd::Zero(3, 6);
+  B.topLeftCorner<3, 3>() = A10();
+  /* The system of equation reads
+       Dx  + By = 0 (1)
+       Bᵀx + Ay = a (2)
+     Using equation (1), we get
+       x = -D⁻¹By */
+  const VectorXd expected_x = D.llt().solve(-B * y);
+  EXPECT_TRUE(CompareMatrices(x, expected_x, kTolerance));
+}
+
+GTEST_TEST(SchurComplementTest, Solve) {
+  const SchurComplement schur_complement = MakeSchurComplement();
+  const VectorXd b = VectorXd::LinSpaced(9, 0.0, 12.0);
+  const VectorXd z = schur_complement.Solve(b);
+
+  const MatrixXd A = MakeBlockSparseMatrix().MakeDenseMatrix();
+  const VectorXd expected_z = A.llt().solve(b);
+  EXPECT_TRUE(CompareMatrices(z, expected_z, kTolerance));
+}
+
+}  // namespace
+}  // namespace internal
+}  // namespace contact_solvers
+}  // namespace multibody
+}  // namespace drake


xuchenhan-tri (commented):

Towards #19882  

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-:-Naso37b-7T-8HSJARdq:b-mpa9eg)* status: LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

RussTedrake (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/schur_complement.h` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NaswTAyAcHAwFBFX3_t:-NaswTAz7ztyfZSW_PWi:bsjc2yv) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L9)):*
> ```objc
> #include "drake/common/eigen_types.h"
> #include "drake/multibody/contact_solvers/block_sparse_cholesky_solver.h"
> #include "drake/multibody/contact_solvers/block_sparse_lower_triangular_or_symmetric_matrix.h"
> ```

btw -- why hide all these in `/multibody/contact_solvers`?  They seem general enough to go in`/math`, and potentially super useful?

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)* status: 1 unresolved discussion, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`multibody/contact_solvers/schur_complement.h` line 9 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NaswTAyAcHAwFBFX3_t:-NbFpLwD3vEMf9P6Iz8n:bdshgak) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L9)):*

btw -- why hide all these in `/multibody/contact_solvers`?  They seem general enough to go in`/math`, and potentially super useful?
</blockquote></details>

I designed the class specifically for SAP and FEM. If it turns out to be more generally useful, I'd be very happy to move it into `/math` (at which point we need to take another look at the API to ensure it's appropriately general).

amcastro-tri (commented):
what's the plan with `drake::multibody::fem::SchurComplement`. Will it go away or should we consolidate?

> Reviewed 3 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-:-NbKSLJh7RI6uUgOilyn:bjujftv)* status: 19 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKgLV26AVU0ESV4mTj:-NbKgLV30VAAcpFih5vC:b-jzqjqk) (no related file):*
Feature review done. This is great!
The only major comment is in `SolveForX()`

___
___
*[`multibody/contact_solvers/schur_complement.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXCHWDNXAhEGnGpuX:-NbKXCHWDNXAhEGnGpuY:b-nhn7ok) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L21)):*
> ```objc
>      Bᵀx + Cy = a     (2)
> 
>  where A = [D B; Bᵀ C], z = [x; y], b = [0; a], if C (size p-by-p),
> ```

minor,



_Suggestion:_
```objc
 where A = [D B; Bᵀ C], z = [x; y], b = [0; a]. If C (
```
___
*[`multibody/contact_solvers/schur_complement.h` line 33 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXWPO8TUbpWiGyFfn:-NbKXWPO8TUbpWiGyFfo:b-nhn7ok) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L33)):*
> ```objc
>  After a solution for y is obtained, we can use (3) to recover the solution for
>  x. The matrix S = C - BᵀD⁻¹B is the Schur complement of the block D of the
>  matrix A. Since A is positive definite, so is the S.
> ```

minor,



_Suggestion:_
```objc
so is S.
```
___
*[`multibody/contact_solvers/schur_complement.h` line 35 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXgPe5SGHLXFOmlXM:-NbKXgPf6P64IzboWukY:b1qzmi9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L35)):*
> ```objc
>  matrix A. Since A is positive definite, so is the S.
> 
>  Given the symmetric matrix A and information on how to decompose M into
> ```

what is M?



_Code quote:_
```objc
 decompose M into
```
___
*[`multibody/contact_solvers/schur_complement.h` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKZ7fL4ttzlEck0pra:-NbKZ7fL4ttzlEck0prb:br2xk30) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L45)):*
> ```objc
>   /* Constructs an empty SchurComplement, i.e one that corresponds
>    corresponds to a linear system with no equations.*/
>   SchurComplement() : S_(0, 0) {}
> ```

why would I want an empty SchurComplement? I don't see in this class a way to "reset" it to something useful.
___
*[`multibody/contact_solvers/schur_complement.h` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKZQo8CIZAG4FIrtOJ:-NbKZQo8CIZAG4FIrtOK:bis79r8) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L48)):*
> ```objc
> 
>   /* Constructs a SchurComplement for the block sparse matrix A of size
>   3*N-by-3*N consisting of blocks of size 3x3. `D_indices` and `C_indices`
> ```

C_indices is not an argument, but clearly implied.
Either remove from the docs, or define it, or if useful, add public methods to retrieve it.



_Code quote:_
```objc
 `C_indices`
```
___
*[`multibody/contact_solvers/schur_complement.h` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbK_Rn97ctyDe1tSTLh:-NbK_Rn97ctyDe1tSTLi:b-31boyy) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L64)):*
> ```objc
>      w w w | v v v | z z z
> 
>   and D_indices = {0, 2} and C_indices = {1}, then submatrix D is given by
> ```

ah, this is great, but the class's documentation is not stating this. 
This actually implies that the block structure `[D B; Bᵀ C]` is only obtained after a permutation of the blocks of A as specified by `D_indices`. 
See if you can convey that important detail in the class's docs.
___
*[`multibody/contact_solvers/schur_complement.h` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKaVUo6Nr7gw3EzEn9:-NbKaVUo6Nr7gw3EzEnA:b-ayojpi) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L100)):*
> ```objc
>   /* Given a value of y, solves for x in the equation Dx + By = 0.
>    @pre The size of y is equal to the number of columns of B */
>   VectorX<double> SolveForX(const Eigen::Ref<const VectorX<double>>& y) const;
> ```

minor, consider having `SolveForY()` as well.
___
*[`multibody/contact_solvers/schur_complement.h` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYf-I1zGAb_i3T-XP:-NbKYf-I1zGAb_i3T-XQ:bb7lc54) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L102)):*
> ```objc
>   VectorX<double> SolveForX(const Eigen::Ref<const VectorX<double>>& y) const;
> 
>   /* Given a right hand side vector b with the same dimension as the input
> ```

nit,



_Suggestion:_
```objc
right-hand side
```
___
*[`multibody/contact_solvers/schur_complement.h` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYpq10TTI9HRSKO6x:-NbKYpq10TTI9HRSKO6y:bl6f7b3) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L106)):*
> ```objc
>    @pre The size of b is compatible with the input matrix A provided at
>    construction. */
>   VectorX<double> Solve(const Eigen::Ref<const VectorX<double>>& b) const;
> ```

nit, consider a signature `Solve(const Eigen::Ref<const VectorX<double>>& b, VectorX<double>* z)` so that the caller can re-use pre-allocated memory.
___
*[`multibody/contact_solvers/schur_complement.h` line 116 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYAHvEr9IMRjyvr1F:-NbKYAHvEr9IMRjyvr1G:b-mq42b8) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L116)):*
> ```objc
>   std::vector<int> C_indices_;
>   /* Cholesky solver that factorizes the matrix and stores the factorization. */
>   BlockSparseCholeskySolver<Matrix3<double>> solver_;
> ```

nit, consider `A_solver_`?
___
*[`multibody/contact_solvers/schur_complement.cc` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKdQOB9kqbxg7ZmYm0:-NbKdQOB9kqbxg7ZmYm1:b8y5ku9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L34)):*
> ```cpp
>         "matrix is symmetric positive definite and not ill-conditioned.");
>   }
>   DRAKE_DEMAND(
> ```

could this ever be triggered?
___
*[`multibody/contact_solvers/schur_complement.cc` line 43 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKdXU17xvqWVJ0utly:-NbKdXU17xvqWVJ0utlz:b-8114te) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L43)):*
> ```cpp
>     const Eigen::Ref<const VectorX<double>>& y) const {
>   DRAKE_THROW_UNLESS(y.size() == 3 * ssize(C_indices_));
>   DRAKE_DEMAND(
> ```

could this ever be triggered?, ditto for `Solve()`
___
*[`multibody/contact_solvers/schur_complement.cc` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKfjjT-wtdIJWDpAH5:-NbKfjjT-wtdIJWDpAH6:bhrwf4j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L62)):*
> ```cpp
>   }
>   VectorX<double>& z = b;
>   solver_.SolveInPlace(&z);
> ```

Ideally solving for `x` would not include the eliminated blocks. That is, we'd work directly with equation `x = -D⁻¹By` (without having to compute `a` and pad it with zeros to get `b`) since someone (through the Schur complement) already computed `y`. So aren't we here doing this work twice? is there a way to avoid it?
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKadXaDMWEOHfgsAGk:-NbKadXaDMWEOHfgsAGl:b-htrxwk) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L18)):*
> ```cpp
> using Eigen::VectorXd;
> 
> constexpr double kEps = 2e-13;
> ```

nit, consider `kTolerance` for anumber different from machine epsilon.
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKcF_GDhY5sBhhrD9a:-NbKcF_GDhY5sBhhrD9b:bl4ry7k) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L18)):*
> ```cpp
> using Eigen::VectorXd;
> 
> constexpr double kEps = 2e-13;
> ```

I'd imagine you made this the tightest but, just in case, is this the lowest we can go?



_Code quote:_
```cpp
2e-13;
```
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKark0Aj98W19GD8N8:-NbKark0Aj98W19GD8N9:b-gmud3v) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L49)):*
> ```cpp
>         -----------------
>          A20 |  0  | A22
>  where A02 = A20.transpose(). We choose values so that M is diagonally dominant
> ```

consider A for consistency with docs in the class and in this test.



_Code quote:_
```cpp
 so that M is 
```
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 109 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKc-ok56hHVkR0M-uz:-NbKc-ok56hHVkR0M-v-:b-acziuj) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L109)):*
> ```cpp
>        Dx  + By = 0
>        Bᵀx + Ay = a
>      Using equation (1), we get
> ```

nit, number equation or say "first  equation" or "first block row"



_Code quote:_
```cpp
 Using equation (1), 
```

xuchenhan-tri (commented):
Yes. `fem::SchurComplement` will be eliminated along with the PETSc dependency.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-:-NbL6ZSQ9una5zjnGQNe:b23zr42)* status: 12 unresolved discussions, LGTM missing from assignee amcastro-tri, needs platform reviewer assigned, needs at least two assigned reviewers

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKgLV26AVU0ESV4mTj:-NbL6jmmEkqDc9X-J3it:b-896fix) (no related file):*

Feature review done. This is great!
The only major comment is in `SolveForX()`
</blockquote></details>

Done.

___
___
*[`multibody/contact_solvers/schur_complement.h` line 21 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXCHWDNXAhEGnGpuX:-NbL6tWzE8BEDY52R8V9:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L21)):*

minor,
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 33 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXWPO8TUbpWiGyFfn:-NbL6xs41niKG0Mb_hPI:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L33)):*

minor,
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 35 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKXgPe5SGHLXFOmlXM:-NbL70VW3trQ9ocGmD_S:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L35)):*

what is M?
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKZ7fL4ttzlEck0pra:-NbL77-6-97yk1CAjh6_:bj5ihrm) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L45)):*

why would I want an empty SchurComplement? I don't see in this class a way to "reset" it to something useful.
</blockquote></details>

A default constructor makes it easy to create a cache entry in the context.
___
*[`multibody/contact_solvers/schur_complement.h` line 48 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKZQo8CIZAG4FIrtOJ:-NbL7ktZA7Bx9wXd-A-H:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L48)):*

C_indices is not an argument, but clearly implied.
Either remove from the docs, or define it, or if useful, add public methods to retrieve it.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/schur_complement.h` line 64 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbK_Rn97ctyDe1tSTLh:-NbL8lbi-e28C7X43Wif:bp80j6r) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L64)):*

ah, this is great, but the class's documentation is not stating this. 
This actually implies that the block structure `[D B; Bᵀ C]` is only obtained after a permutation of the blocks of A as specified by `D_indices`. 
See if you can convey that important detail in the class's docs.
</blockquote></details>

I modified the class doc to xref here. PTAL.
___
*[`multibody/contact_solvers/schur_complement.h` line 100 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKaVUo6Nr7gw3EzEn9:-NbL943WD7t3WNc13uDA:b-15z3c3) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L100)):*

minor, consider having `SolveForY()` as well.
</blockquote></details>

Done via todo.
___
*[`multibody/contact_solvers/schur_complement.h` line 102 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYf-I1zGAb_i3T-XP:-NbL9DnS91NgmWKO-YGK:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L102)):*

nit,
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYpq10TTI9HRSKO6x:-NbM3zKdClH7QuJmi_eh:b9d4hl) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L106)):*

nit, consider a signature `Solve(const Eigen::Ref<const VectorX<double>>& b, VectorX<double>* z)` so that the caller can re-use pre-allocated memory.
</blockquote></details>

My callsite looks like

```
VectorX<double>& z = scratch_data.z;
z = schur_complement.Solve(b);
```

where the memory in scratch data is pre-allocated. I believe RVO is applied here so we aren't saving anything by passing into a z pointer here.
___
*[`multibody/contact_solvers/schur_complement.h` line 116 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYAHvEr9IMRjyvr1F:-NbLARnpCGkEJJsRWmkV:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L116)):*

nit, consider `A_solver_`?
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.cc` line 34 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKdQOB9kqbxg7ZmYm0:-NbLAclm1RYkCjr45W1P:b-9iz2m9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L34)):*

could this ever be triggered?
</blockquote></details>

No it shouldn't ever be triggered. This is a sanity check. Failing this is a bug.
___
*[`multibody/contact_solvers/schur_complement.cc` line 43 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKdXU17xvqWVJ0utly:-NbLAn739cRWlO8eaHk8:b-9iz2m9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L43)):*

could this ever be triggered?, ditto for `Solve()`
</blockquote></details>

No it shouldn't ever be triggered. This is a sanity check. Failing this is a bug.
___
*[`multibody/contact_solvers/schur_complement.cc` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKfjjT-wtdIJWDpAH5:-NbM7DXJE8B35_CXfhdA:b1n3o4v) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L62)):*

Ideally solving for `x` would not include the eliminated blocks. That is, we'd work directly with equation `x = -D⁻¹By` (without having to compute `a` and pad it with zeros to get `b`) since someone (through the Schur complement) already computed `y`. So aren't we here doing this work twice? is there a way to avoid it?
</blockquote></details>

Per f2f, I added a TODO to document the performance consideration.
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKadXaDMWEOHfgsAGk:-NbLAz5uDelZpeUbDl7w:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L18)):*

nit, consider `kTolerance` for anumber different from machine epsilon.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 18 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKcF_GDhY5sBhhrD9a:-NbLPtnT-94JWcHnxye2:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L18)):*

I'd imagine you made this the tightest but, just in case, is this the lowest we can go?
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKark0Aj98W19GD8N8:-NbLB3OrAtbKULjAlrpf:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L49)):*

consider A for consistency with docs in the class and in this test.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 109 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKc-ok56hHVkR0M-uz:-NbLB8Cl4OfW4J88u7G8:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/test/schur_complement_test.cc#L109)):*

nit, number equation or say "first  equation" or "first block row"
</blockquote></details>

Done

amcastro-tri (commented):
excellent, thanks. :lgtm:

+@jwnimmer-tri for platform review please

> Reviewed 1 of 4 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-:-NbPXVSA0ZTFZtv8CJ_q:b-xvw508)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

___
*[`multibody/contact_solvers/schur_complement.h` line 45 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKZ7fL4ttzlEck0pra:-NbPXoiOAteUbT4psH34:b-ox21f9) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L45)):*

A default constructor makes it easy to create a cache entry in the context.
</blockquote></details>

I see, the whole thing has value semantics.
___
*[`multibody/contact_solvers/schur_complement.h` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYpq10TTI9HRSKO6x:-NbP_B_u4AySSggRgIuJ:bbk5ygj) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L106)):*
From https://abseil.io/tips/11 :
> The compiler can’t do RVO if the calling function reuses an existing variable to store the return value (though move semantics would apply for move-enabled types in this case)

not sure about the move semantics in this case though since z has been constructed already. Maybe in this example you could do `z = std::move(schur_complement.Solve(b));`?
___
*[`multibody/contact_solvers/schur_complement.cc` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKfjjT-wtdIJWDpAH5:-NbPa23C6Va6w8VEOFK5:b-gfv2n8) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L62)):*

Per f2f, I added a TODO to document the performance consideration.
</blockquote></details>

thanks! consider adding a comment about the relative sizes of D to A considered for these statement to be true. Didn't we say this is true when size(D) is much smaller than size(A)?
___
*[`multibody/contact_solvers/schur_complement.cc` line 40 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbP_o6c3chuhKQEmIIB:-NbP_o6c3chuhKQEmIIC:b-hvv2d8) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L40)):*
> ```cpp
> }
> 
> /* TODO(xuchenhan-tri): In the following implementation, we constructs `a`, the
> ```

minor, 



_Suggestion:_
```cpp
construct 
```

amcastro-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform)

jwnimmer-tri (commented):
:lgtm:

> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-:-NbQOUM38ZvYu01RI9s_:bnfp4nl)* status: 9 unresolved discussions

___
*[`multibody/contact_solvers/schur_complement.h` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQKruQ2m-CPsh7gGWY:-NbQKruQ2m-CPsh7gGWZ:b4b5au6) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L41)):*
> ```objc
>  one can solve the system Az = b efficiently once the Schur complement is
>  computed. */
> class SchurComplement {
> ```

nit The `~SchurComplement` destructor is non-trivial, so should not be inlined.
___
*[`multibody/contact_solvers/schur_complement.h` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQK_ib8IUmUr7xUV-W:-NbQK_icAxz9CE15Se7Y:b-juajao) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L47)):*
> ```objc
>   /* Constructs an empty SchurComplement, i.e one that corresponds
>    corresponds to a linear system with no equations.*/
>   SchurComplement() : S_(0, 0) {}
> ```

BTW Is the `(0, 0)` to underline for the reader that is S is really empty?  It would be the default value, if unmentioned.



_Suggestion:_
```objc
= default;
```
___
*[`multibody/contact_solvers/schur_complement.h` line 51 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQJM_C4SvJMCtaqcPy:-NbQJM_DDIebyYBUlmhR:bgows8d) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L51)):*
> ```objc
>   /* Constructs a SchurComplement for the block sparse matrix A of size
>    3*N-by-3*N consisting of blocks of size 3x3. `D_indices` determine which
>   block rows/columns make up the D diagonal blocks. The remaining block
> ```

nit We lost the extra `+1` indentation on these few lines.
___
*[`multibody/contact_solvers/schur_complement.cc` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQLes54Su450peMk-E:-NbQLes54Su450peMk-F:b8ap2vg) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L19)):*
> ```cpp
>   /* If a block index doesn't belong to the D blocks, it belongs to the C
>    blocks. */
>   C_indices_.reserve(A.block_cols() - D_indices.size());
> ```

It seems a bit too dangerous to subtract here and end up calling `reserve()` on a negative number.  I think we need to cross-check the size of `A` vs `D` before we perform this reservation.
___
*[`multibody/contact_solvers/schur_complement.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQLxhF-5koQq0VNK42:-NbQLxhF-5koQq0VNK43:b-cx9s33) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L21)):*
> ```cpp
>   C_indices_.reserve(A.block_cols() - D_indices.size());
>   for (int j = 0; j < A.block_cols(); ++j) {
>     if (D_indices.count(j) == 0) {
> ```

BTW The set lookup _is_ constant time, but the constant factor is pretty big.

We already have `C_indices_` sorted on hand.  We could step through that vector to detect the gaps in order to fill in D, to save some time.  Possibly worth capturing as a TODO.  (Or if you've profiled already, comment that it doesn't matter).
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQMueV8Zvg7mC_PNLH:-NbQMueV8Zvg7mC_PNLI:btzwo3a) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/test/schur_complement_test.cc#L21)):*
> ```cpp
> 
> // clang-format off
> const Matrix3d A00 =
> ```

nit https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables

These data are not constant-initialized, so can't be statics.

I think the simplest fix is to make them functions:

```c++
Matrix3d A00() {
  return (Eigen::Matrix3d() <<
      11, 2, 3,
      2, 50, 6,
      3, 6, 190).finished();
}
```

xuchenhan-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)* status: 2 unresolved discussions

___
*[`multibody/contact_solvers/schur_complement.h` line 106 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKYpq10TTI9HRSKO6x:-NbRIvf2-DH6_X6WXMxc:bkzx8a6) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.h#L106)):*

From https://abseil.io/tips/11 :
> The compiler can’t do RVO if the calling function reuses an existing variable to store the return value (though move semantics would apply for move-enabled types in this case)

not sure about the move semantics in this case though since z has been constructed already. Maybe in this example you could do `z = std::move(schur_complement.Solve(b));`?
</blockquote></details>

You are right. RVO doesn't apply here. You don't need `std::move(schur_complement.Solve(b))` though, `schur_complement.Solve(b)` is already an r-value.

Returning by output parameter doesn't help too much in this case because the downstream implementation returns by value anyway.
___
*[`multibody/contact_solvers/schur_complement.h` line 41 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQKruQ2m-CPsh7gGWY:-NbRInB-9BBupsbmVkCM:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L41)):*

nit The `~SchurComplement` destructor is non-trivial, so should not be inlined.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 47 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQK_ib8IUmUr7xUV-W:-NbRInxYBW4pn8_Fj_QK:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L47)):*

BTW Is the `(0, 0)` to underline for the reader that is S is really empty?  It would be the default value, if unmentioned.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.h` line 51 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQJM_C4SvJMCtaqcPy:-NbRIuXN3nbzloMfPDeF:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.h#L51)):*

nit We lost the extra `+1` indentation on these few lines.
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.cc` line 62 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbKfjjT-wtdIJWDpAH5:-NbRQlHu8o8r7PrCTnKG:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/bcd24219eeb1dd1631042b79f21db602465c8968/multibody/contact_solvers/schur_complement.cc#L62)):*

thanks! consider adding a comment about the relative sizes of D to A considered for these statement to be true. Didn't we say this is true when size(D) is much smaller than size(A)?
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.cc` line 19 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQLes54Su450peMk-E:-NbRJVi12WB0elpgdH8G:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L19)):*

It seems a bit too dangerous to subtract here and end up calling `reserve()` on a negative number.  I think we need to cross-check the size of `A` vs `D` before we perform this reservation.
</blockquote></details>

Done.
___
*[`multibody/contact_solvers/schur_complement.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQLxhF-5koQq0VNK42:-NbRQNHt-wPuXRBGtleQ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L21)):*

BTW The set lookup _is_ constant time, but the constant factor is pretty big.

We already have `C_indices_` sorted on hand.  We could step through that vector to detect the gaps in order to fill in D, to save some time.  Possibly worth capturing as a TODO.  (Or if you've profiled already, comment that it doesn't matter).
</blockquote></details>

Done
___
*[`multibody/contact_solvers/schur_complement.cc` line 40 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbP_o6c3chuhKQEmIIB:-NbRQR5f0kHIAxMSkoej:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/schur_complement.cc#L40)):*

minor, 
</blockquote></details>

Done
___
*[`multibody/contact_solvers/test/schur_complement_test.cc` line 21 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19924#-NbQMueV8Zvg7mC_PNLH:-NbRROK5DGVg3UHFVtAD:b-iry7x8) ([raw file](https://github.com/RobotLocomotion/drake/blob/f3114251c731ea38df0f3cd161c86356b87d6694/multibody/contact_solvers/test/schur_complement_test.cc#L21)):*

nit https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables

These data are not constant-initialized, so can't be statics.

I think the simplest fix is to make them functions:

```c++
Matrix3d A00() {
  return (Eigen::Matrix3d() <<
      11, 2, 3,
      2, 50, 6,
      3, 6, 190).finished();
}
```
</blockquote></details>

Done, thanks

jwnimmer-tri (commented):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19924)* status: 1 unresolved discussion

PR #19926, opened 2023-08-03
Require unit vectors for UnitInertia functions.

mitiguy:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)

### DIFF
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -733,26 +733,54 @@ class TestPlant(unittest.TestCase):
         self.assertIsInstance(UnitInertia.SolidBox(Lx=1, Ly=2, Lz=3),
                               UnitInertia)
         self.assertIsInstance(UnitInertia.SolidCube(L=2), UnitInertia)
-        self.assertIsInstance(UnitInertia.SolidCylinder(r=1.5, L=2),
+        self.assertIsInstance(UnitInertia.SolidCylinder(
+                                  radius=1.5, length=2, unit_vector=[0, 0, 1]),
                               UnitInertia)
-        self.assertIsInstance(
-            UnitInertia.SolidCylinder(r=1.5, L=2, b_E=[1, 2, 3]), UnitInertia)
-        self.assertIsInstance(UnitInertia.SolidCapsule(r=1, L=2), UnitInertia)
-        self.assertIsInstance(UnitInertia.SolidCapsule(r=1, L=2,
-                                                       unit_vector=[0, 0, 1]),
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.SolidCylinder(r=1.5, L=2),
+                                  UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.SolidCylinder(
+                                      r=1.5, L=2, b_E=[1, 2, 3]),
+                                  UnitInertia)
+        self.assertIsInstance(UnitInertia.SolidCapsule(
+                                  radius=1, length=2, unit_vector=[0, 0, 1]),
                               UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.SolidCapsule(r=1, L=2),
+                                  UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.SolidCapsule(
+                                      r=1, L=2, unit_vector=[0, 0, 1]),
+                                  UnitInertia)
         self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(
             radius=0.1, length=0.4, unit_vector=[0, 0, 1]),
                               UnitInertia)
         with catch_drake_warnings(expected_count=1):
             self.assertIsInstance(UnitInertia.SolidCylinderAboutEnd(r=1, L=4),
                                   UnitInertia)
-        self.assertIsInstance(
-            UnitInertia.AxiallySymmetric(J=1, K=2, b_E=[1, 2, 3]), UnitInertia)
-        self.assertIsInstance(UnitInertia.StraightLine(K=1.5, b_E=[1, 2, 3]),
+        self.assertIsInstance(UnitInertia.AxiallySymmetric(
+                                  moment_parallel=1, moment_perpendicular=2,
+                                  unit_vector=[0, 0, 1]),
                               UnitInertia)
-        self.assertIsInstance(UnitInertia.ThinRod(L=1.5, b_E=[1, 2, 3]),
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.AxiallySymmetric(
+                                      J=1, K=2, b_E=[1, 2, 3]),
+                                  UnitInertia)
+        self.assertIsInstance(UnitInertia.StraightLine(
+                                  moment_perpendicular=1.5,
+                                  unit_vector=[0, 0, 1]),
                               UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.StraightLine(
+                                      K=1.5, b_E=[1, 2, 3]),
+                                  UnitInertia)
+        self.assertIsInstance(UnitInertia.ThinRod(
+                                  length=1.5, unit_vector=[0, 0, 1]),
+                              UnitInertia)
+        with catch_drake_warnings(expected_count=1):
+            self.assertIsInstance(UnitInertia.ThinRod(L=1.5, b_E=[1, 2, 3]),
+                                  UnitInertia)
 
     @numpy_compare.check_all_types
     def test_spatial_inertia_api(self, T):
--- a/bindings/pydrake/multibody/tree_py.cc
+++ b/bindings/pydrake/multibody/tree_py.cc
@@ -1152,34 +1152,32 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py::arg("Lz"), cls_doc.SolidBox.doc)
         .def_static(
             "SolidCube", &Class::SolidCube, py::arg("L"), cls_doc.SolidCube.doc)
-        .def_static("SolidCylinder", &Class::SolidCylinder, py::arg("r"),
-            py::arg("L"), py::arg("b_E") = Vector3<T>::UnitZ().eval(),
+        .def_static("SolidCylinder",
+            // TODO(jwnimmer-tri) Drop the overload_cast_explicit on 2023-12-01.
+            overload_cast_explicit<Class, const T&, const T&,
+                const Vector3<T>&>(&Class::SolidCylinder),
+            py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
             cls_doc.SolidCylinder.doc)
-        .def_static("SolidCapsule", &Class::SolidCapsule, py::arg("r"),
-            py::arg("L"), py::arg("unit_vector") = Vector3<T>::UnitZ().eval(),
+        .def_static("SolidCapsule",
+            // TODO(jwnimmer-tri) Drop the overload_cast_explicit on 2023-12-01.
+            overload_cast_explicit<Class, const T&, const T&,
+                const Vector3<T>&>(&Class::SolidCapsule),
+            py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
             cls_doc.SolidCapsule.doc)
-        // TODO(2023-11-01) Remove overload wrapping when deprecation complete.
         .def_static("SolidCylinderAboutEnd",
+            // TODO(jwnimmer-tri) Drop the overload_cast_explicit on 2023-12-01.
             overload_cast_explicit<Class, const T&, const T&,
                 const Vector3<T>&>(&Class::SolidCylinderAboutEnd),
             py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
             cls_doc.SolidCylinderAboutEnd.doc)
-        .def_static("SolidCylinderAboutEnd",
-            WrapDeprecated(cls_doc.SolidCylinderAboutEnd.doc_deprecated,
-                [](const T& r, const T& L) {
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-                  return Class::SolidCylinderAboutEnd(r, L);
-#pragma GCC diagnostic pop
-                }),
-            py::arg("r"), py::arg("L"),
-            cls_doc.SolidCylinderAboutEnd.doc_deprecated)
-        .def_static("AxiallySymmetric", &Class::AxiallySymmetric, py::arg("J"),
-            py::arg("K"), py::arg("b_E"), cls_doc.AxiallySymmetric.doc)
-        .def_static("StraightLine", &Class::StraightLine, py::arg("K"),
-            py::arg("b_E"), cls_doc.StraightLine.doc)
-        .def_static("ThinRod", &Class::ThinRod, py::arg("L"), py::arg("b_E"),
-            cls_doc.ThinRod.doc)
+        .def_static("AxiallySymmetric", &Class::AxiallySymmetric,
+            py::arg("moment_parallel"), py::arg("moment_perpendicular"),
+            py::arg("unit_vector"), cls_doc.AxiallySymmetric.doc)
+        .def_static("StraightLine", &Class::StraightLine,
+            py::arg("moment_perpendicular"), py::arg("unit_vector"),
+            cls_doc.StraightLine.doc)
+        .def_static("ThinRod", &Class::ThinRod, py::arg("length"),
+            py::arg("unit_vector"), cls_doc.ThinRod.doc)
         .def_static("TriaxiallySymmetric", &Class::TriaxiallySymmetric,
             py::arg("I_triaxial"), cls_doc.TriaxiallySymmetric.doc)
         .def(py::pickle(
@@ -1191,6 +1189,57 @@ void DoScalarDependentDefinitions(py::module m, T) {
                   I(0, 0), I(1, 1), I(2, 2), I(0, 1), I(0, 2), I(1, 2));
             }));
     DefCopyAndDeepCopy(&cls);
+    // TODO(jwnimmer-tri) Remove as of 2023-12-01.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    constexpr char doc_factory_deprecated[] = R"""(
+This overload spelling of a UnitInertia factory method is deprecated.
+
+The named arguments for the UnitInertia factory methods have been renamed:
+ - ``r`` → ``radius``
+ - ``L`` → ``length``
+ - ``J`` → ``moment_parallel``
+ - ``K`` → ``moment_perpendicular``
+ - ``b_E`` → ``unit_vector``
+
+The old argument names will be removed from Drake on or after 2023-12-01.
+
+The ``unit_vector`` argument no longer has a default value, and must already
+have been normalized by the caller. Passing in a non-normalized unit vector
+will result in a warning for now, but an exception on or after 2023-12-01.
+)""";
+    cls  // BR
+        .def_static("SolidCylinder",
+            WrapDeprecated(doc_factory_deprecated,
+                [](const T& r, const T& L, const Vector3<T>& b_E) {
+                  return Class::SolidCylinder(r, L, b_E);
+                }),
+            py::arg("r"), py::arg("L"),
+            py::arg("b_E") = Vector3<T>::UnitZ().eval(), doc_factory_deprecated)
+        .def_static("SolidCylinderAboutEnd",
+            WrapDeprecated(doc_factory_deprecated,
+                [](const T& r, const T& L) {
+                  return Class::SolidCylinderAboutEnd(r, L);
+                }),
+            py::arg("r"), py::arg("L"), doc_factory_deprecated)
+        .def_static("SolidCapsule",
+            WrapDeprecated(doc_factory_deprecated,
+                [](const T& r, const T& L, const Vector3<T>& unit_vector) {
+                  return Class::SolidCapsule(r, L, unit_vector);
+                }),
+            py::arg("r"), py::arg("L"),
+            py::arg("unit_vector") = Vector3<T>::UnitZ().eval(),
+            doc_factory_deprecated)
+        .def_static("AxiallySymmetric",
+            WrapDeprecated(doc_factory_deprecated, &Class::AxiallySymmetric),
+            py::arg("J"), py::arg("K"), py::arg("b_E"), doc_factory_deprecated)
+        .def_static("StraightLine",
+            WrapDeprecated(doc_factory_deprecated, &Class::StraightLine),
+            py::arg("K"), py::arg("b_E"), doc_factory_deprecated)
+        .def_static("ThinRod",
+            WrapDeprecated(doc_factory_deprecated, &Class::ThinRod),
+            py::arg("L"), py::arg("b_E"), doc_factory_deprecated);
+#pragma GCC diagnostic pop
   }
 
   // SpatialInertia
--- a/common/text_logging.h
+++ b/common/text_logging.h
@@ -197,7 +197,7 @@ sink* get_dist_sink();
 ///   return data;
 /// }
 /// </pre>
-struct Warn {
+struct[[maybe_unused]] Warn {
   template <typename... Args>
   Warn(const char* a, const Args&... b) {
     // TODO(jwnimmer-tri) Ideally we would compile-time check our Warn format
--- a/examples/multibody/cylinder_with_multicontact/test/populate_cylinder_plant_test.cc
+++ b/examples/multibody/cylinder_with_multicontact/test/populate_cylinder_plant_test.cc
@@ -65,8 +65,11 @@ GTEST_TEST(PopulateCylinderPlant, VerifyPlant) {
 
   EXPECT_EQ(M_Bcm_B.get_mass(), mass);
   EXPECT_EQ(M_Bcm_B.get_com(), Vector3d::Zero());
-  EXPECT_EQ(M_Bcm_B.get_unit_inertia().get_moments(),
-            Vector3d(G_perp, G_perp, G_axis));
+
+  // An empirical tolerance: two bits = 2^2 times machine epsilon.
+  const double kTolerance = 4 * std::numeric_limits<double>::epsilon();
+  EXPECT_TRUE(CompareMatrices(M_Bcm_B.get_unit_inertia().get_moments(),
+                              Vector3d(G_perp, G_perp, G_axis), kTolerance));
   EXPECT_EQ(M_Bcm_B.get_unit_inertia().get_products(), Vector3d::Zero());
 
   // TODO(amcastro-tri): Verify geometry including:
--- a/multibody/parsing/test/detail_mujoco_parser_test.cc
+++ b/multibody/parsing/test/detail_mujoco_parser_test.cc
@@ -736,9 +736,11 @@ TEST_F(MujocoParserTest, InertiaFromGeometry) {
 
   check_body("default", UnitInertia<double>::SolidSphere(0.1));
   check_body_spatial("sphere", inertia_from_inertial_tag);
-  check_body("capsule", UnitInertia<double>::SolidCapsule(0.1, 4.0));
+  check_body("capsule",
+      UnitInertia<double>::SolidCapsule(0.1, 4.0, Vector3d::UnitZ()));
   check_body("ellipsoid", UnitInertia<double>::SolidEllipsoid(0.1, 0.2, 0.3));
-  check_body("cylinder", UnitInertia<double>::SolidCylinder(0.1, 4.0));
+  check_body("cylinder",
+      UnitInertia<double>::SolidCylinder(0.1, 4.0, Vector3d::UnitZ()));
   check_body("box", UnitInertia<double>::SolidBox(0.2, 4.0, 6.0));
   check_body("box_from_default", UnitInertia<double>::SolidBox(0.2, 0.4, 0.6));
   check_body("ellipsoid_from_default",
--- a/multibody/tree/spatial_inertia.cc
+++ b/multibody/tree/spatial_inertia.cc
@@ -222,11 +222,6 @@ SpatialInertia<T> SpatialInertia<T>::ThinRodWithMass(
   ThrowUnlessValueIsPositiveFinite(mass, "mass", __func__);
   ThrowUnlessValueIsPositiveFinite(length, "length", __func__);
   ThrowUnlessVectorIsMagnitudeOne(unit_vector, __func__);
-
-  // Although a check is made that ‖unit_vector‖ ≈ 1, we normalize regardless.
-  // Note: UnitInertia::ThinRod() calls UnitInertia::AxiallySymmetric() which
-  // also normalizes unit_vector before use.
-  // TODO(Mitiguy) remove normalization in UnitInertia::AxiallySymmetric().
   const UnitInertia<T> G_BBcm_B =
       UnitInertia<T>::ThinRod(length, unit_vector);
   const Vector3<T> p_BoBcm_B = Vector3<T>::Zero();
--- a/multibody/tree/test/spatial_inertia_test.cc
+++ b/multibody/tree/test/spatial_inertia_test.cc
@@ -880,7 +880,7 @@ GTEST_TEST(SpatialInertia, Shift) {
   // First define it in frame B.
   SpatialInertia<double> M_BBcm_W(
       mass, Vector3d::Zero(),
-      UnitInertia<double>::SolidCylinder(radius, length));
+      UnitInertia<double>::SolidCylinder(radius, length, Vector3d::UnitZ()));
   // Then re-express in frame W.
   M_BBcm_W.ReExpressInPlace(R_WB);
 
--- a/multibody/tree/test/unit_inertia_test.cc
+++ b/multibody/tree/test/unit_inertia_test.cc
@@ -212,10 +212,9 @@ GTEST_TEST(UnitInertia, SolidCylinder) {
   const double I_perp = (3.0 * r * r + L * L) / 12.0;
   const double I_axial = r * r / 2.0;
   const UnitInertia<double> Gz_expected(I_perp, I_perp, I_axial);
-  // Compute the unit inertia for a cylinder oriented along the z-axis
-  // (the default).
+  // Compute the unit inertia for a cylinder oriented along the z-axis.
   UnitInertia<double> Gz =
-      UnitInertia<double>::SolidCylinder(r, L);
+      UnitInertia<double>::SolidCylinder(r, L, Vector3d::UnitZ());
   EXPECT_TRUE(Gz.CopyToFullMatrix3().isApprox(
       Gz_expected.CopyToFullMatrix3(), kEpsilon));
 
@@ -248,6 +247,25 @@ GTEST_TEST(UnitInertia, SolidCylinder) {
   EXPECT_TRUE(Gv.CopyToFullMatrix3().isApprox(
       Gv_expected.CopyToFullMatrix3(), kEpsilon));
 }
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+// 2023-12-01 Remove with the deprecated 2 argument SolidCylinder() function.
+GTEST_TEST(UnitInertia, SolidCylinderDeprecated) {
+  const double r = 2.5;
+  const double L = 1.5;
+  const double I_perp = (3.0 * r * r + L * L) / 12.0;
+  const double I_axial = r * r / 2.0;
+  const UnitInertia<double> Gz_expected(I_perp, I_perp, I_axial);
+
+  // Compute the unit inertia for a cylinder oriented along the z-axis
+  // (the default).
+  UnitInertia<double> Gz =
+      UnitInertia<double>::SolidCylinder(r, L);
+  EXPECT_TRUE(Gz.CopyToFullMatrix3().isApprox(
+      Gz_expected.CopyToFullMatrix3(), kEpsilon));
+}
+#pragma GCC diagnostic pop
 
 // Tests the static method to obtain the unit inertia of a solid cylinder B
 // computed about a point Bp at the center of its base.
@@ -286,7 +304,7 @@ GTEST_TEST(UnitInertia, SolidCylinderAboutEnd) {
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 // Tests the static method to obtain the unit inertia of a solid cylinder B
 // computed about a point Bp at the center of its base.
-// 2023-11-01 Remove with the old SolidCylinderAboutEnd test.
+// 2023-12-01 Remove with the old SolidCylinderAboutEnd test.
 GTEST_TEST(UnitInertia, SolidCylinderAboutEndDeprecated) {
   const double r = 2.5;
   const double L = 1.5;
@@ -320,7 +338,8 @@ GTEST_TEST(UnitInertia, SolidCapsule) {
 
   // The inertia properties of a capsule is calculated three ways.
   // Calculation 1: Multiply the capsule's mass with its unit inertia.
-  const UnitInertia<double> G_capsule = UnitInertia<double>::SolidCapsule(r, L);
+  const UnitInertia<double> G_capsule =
+      UnitInertia<double>::SolidCapsule(r, L, Vector3d::UnitZ());
   RotationalInertia<double> I_capsule = mass_capsule * G_capsule;
 
   // Calculation 2: Calculate the inertia analytically.
@@ -372,8 +391,8 @@ GTEST_TEST(UnitInertia, SolidCapsule) {
   Ih.ShiftToThenAwayFromCenterOfMassInPlace(mh, p_HoHcm_C, p_HcmCcm_C);
 
   // Form the cylinder's inertia about its center of mass and verify results.
-  const RotationalInertia<double> I_cylinder =
-      mass_cylinder * UnitInertia<double>::SolidCylinder(r, L);
+  const RotationalInertia<double> I_cylinder = mass_cylinder *
+      UnitInertia<double>::SolidCylinder(r, L, Vector3d::UnitZ());
   I_capsule = I_cylinder + 2 * Ih;
   EXPECT_TRUE(CompareMatrices(I_capsule.get_moments(),
                               I_capsule_expected.get_moments(), kEpsilon));
@@ -381,9 +400,9 @@ GTEST_TEST(UnitInertia, SolidCapsule) {
                               I_capsule_expected.get_products(), kEpsilon));
 
   // Ensure a bad unit vector throws an exception.
-  const Vector3<double> bad_vec(1, 0.1, 0);
+  const Vector3<double> bad_uvec(1, 0.1, 0);
   DRAKE_EXPECT_THROWS_MESSAGE(
-      UnitInertia<double>::SolidCapsule(r, L, bad_vec),
+      UnitInertia<double>::SolidCapsule(r, L, bad_uvec),
       "[^]* The unit_vector argument .* is not a unit vector.");
 }
 
@@ -393,7 +412,8 @@ GTEST_TEST(UnitInertia, SolidCapsuleDegenerateIntoSolidSphere) {
   const double r = 2.5;
   const double L = 0;
   const UnitInertia<double> G_expected = UnitInertia<double>::SolidSphere(r);
-  const UnitInertia<double> G = UnitInertia<double>::SolidCapsule(r, L);
+  const UnitInertia<double> G =
+      UnitInertia<double>::SolidCapsule(r, L, Vector3d::UnitZ());
   EXPECT_TRUE(CompareMatrices(G.get_moments(), G_expected.get_moments()));
   EXPECT_TRUE(CompareMatrices(G.get_products(), G_expected.get_products()));
 }
@@ -406,7 +426,7 @@ GTEST_TEST(UnitInertia, SolidCapsuleDegenerateIntoThinRod) {
   const UnitInertia<double> G_expected =
       UnitInertia<double>::ThinRod(L, Vector3d::UnitZ());
   const UnitInertia<double> G_degenerate =
-      UnitInertia<double>::SolidCapsule(r, L);
+      UnitInertia<double>::SolidCapsule(r, L, Vector3d::UnitZ());
   EXPECT_TRUE(
       CompareMatrices(G_degenerate.get_moments(), G_expected.get_moments()));
   EXPECT_TRUE(
@@ -414,12 +434,24 @@ GTEST_TEST(UnitInertia, SolidCapsuleDegenerateIntoThinRod) {
   // The unit inertia of the capsule should be close to that of a thin rod when
   // the radius is vanishingly small.
   const UnitInertia<double> G_close_to_degenerate =
-      UnitInertia<double>::SolidCapsule(kEpsilon, L);
+      UnitInertia<double>::SolidCapsule(kEpsilon, L, Vector3d::UnitZ());
   EXPECT_TRUE(CompareMatrices(G_close_to_degenerate.get_moments(),
                               G_expected.get_moments(), kEpsilon));
   EXPECT_TRUE(CompareMatrices(G_close_to_degenerate.get_products(),
                               G_expected.get_products()));
 }
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+// 2023-12-01 Remove with the deprecated 2 argument SolidCapsule() function.
+GTEST_TEST(UnitInertia, SolidCapsuleDeprecated) {
+  const double r = 2.5;
+  const double L = 1.5;
+  const UnitInertia<double> G_capsule = UnitInertia<double>::SolidCapsule(r, L);
+  EXPECT_TRUE(G_capsule.CopyToFullMatrix3().isApprox(
+              G_capsule.CopyToFullMatrix3(), kEpsilon));
+}
+#pragma GCC diagnostic pop
 
 // Unit tests for the factory method UnitInertia::AxiallySymmetric().
 // This test creates the unit inertia for a cylinder of radius r and length L
@@ -442,7 +474,8 @@ GTEST_TEST(UnitInertia, AxiallySymmetric) {
   const double I_axial = r * r / 2.0;
 
   // Cylinder's axis. A vector on the y-z plane, at -pi/4 from the z axis.
-  // The vector doesn't need to be normalized.
+  // The vector doesn't need to be normalized (but take note that automatic
+  // normalization is deprecated).
   const Vector3d b_E = Vector3d::UnitY() + Vector3d::UnitZ();
 
   // Rotation of -pi/4 about the x axis, from a Z frame having its z axis
@@ -456,7 +489,8 @@ GTEST_TEST(UnitInertia, AxiallySymmetric) {
 
   // The expected inertia is that of a cylinder of radius r and height L with
   // its longitudinal axis aligned with b.
-  UnitInertia<double> G_Z = UnitInertia<double>::SolidCylinder(r, L);
+  UnitInertia<double> G_Z =
+      UnitInertia<double>::SolidCylinder(r, L, Vector3d::UnitZ());
   UnitInertia<double> G_E_expected = G_Z.ReExpress(R_EZ);
 
   // Verify the computed values.
@@ -491,7 +525,8 @@ GTEST_TEST(UnitInertia, ThinRod) {
   const double I_rod = L * L / 12.0;
 
   // Rod's axis. A vector on the y-z plane, at -pi/4 from the z axis.
-  // The vector doesn't need to be normalized.
+  // The vector doesn't need to be normalized (but take note that automatic
+  // normalization is deprecated).
   const Vector3d b_E = Vector3d::UnitY() + Vector3d::UnitZ();
 
   // Rotation of -pi/4 about the x axis, from a Z frame having its z-axis
@@ -505,7 +540,8 @@ GTEST_TEST(UnitInertia, ThinRod) {
 
   // The expected inertia is that of a cylinder of zero radius and height L with
   // its longitudinal axis aligned with b.
-  UnitInertia<double> G_Z = UnitInertia<double>::SolidCylinder(0.0, L);
+  UnitInertia<double> G_Z =
+      UnitInertia<double>::SolidCylinder(0, L, Vector3d::UnitZ());
   UnitInertia<double> G_E_expected = G_Z.ReExpress(R_EZ);
 
   // Verify the computed values.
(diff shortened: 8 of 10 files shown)


mitiguy (commented):
Feature review +@sherm1 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-NbMOLArEx9QcAqpXZsz:bwuh3l8)* status: LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`examples/multibody/cylinder_with_multicontact/test/populate_cylinder_plant_test.cc` line 72 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMEhh4BR4ts0nRILHk:-NbMEhh4BR4ts0nRILHl:b-novmts) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/examples/multibody/cylinder_with_multicontact/test/populate_cylinder_plant_test.cc#L72)):*
> ```cpp
>   const double kTolerance = 4 * std::numeric_limits<double>::epsilon();
>   EXPECT_TRUE(CompareMatrices(M_Bcm_B.get_unit_inertia().get_moments(),
>                               Vector3d(G_perp, G_perp, G_axis), kTolerance));
> ```

FYI  The new change allows for a two bit difference in the results -- rather than being exactly equal.
The old test was too strict, requiring exact bit-match when it could not be guaranteed.
___
*[`multibody/tree/unit_inertia.h` line 428 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMIwko6cYDCZn99dQM:-NbMIwko6cYDCZn99dQN:bmmr46i) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L428)):*
> ```objc
>   /// @throws std::exception if moment_parallel (J) or moment_perpendicular (K)
>   /// is negative or if J > 2 K (violates the triangle inequality, see
>   /// CouldBePhysicallyValid()) or ‖unit_vector‖ is not within 1.0E-14 of 1.0.
> ```

FYI: Another name for the parameter moment_parallel could be moment_axial.
___
*[`multibody/tree/unit_inertia.cc` line 192 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMJwSS4v8CDvwGbkNE:-NbMJwSTBZi8vhnjGUqW:b-2tn4u9) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L192)):*
> ```cpp
>   // The distance dH between Hcm (the half-sphere H's center of mass) and Ccm
>   // (the cylinder C's center of mass) is from [Kane, Figure A23, pg. 369].
>   // dH = 3.0 / 8.0 * r + L / 2.0;
> ```

FYI  Change to dH = ⅜ radius + ½ length.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMLgPB8zxHRJtwQ4i-:-NbMLgPB8zxHRJtwQ4i0:bkut9yx) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/test/unit_inertia_test.cc#L215)):*
> ```cpp
>   const double I_axial = r * r / 2.0;
>   const UnitInertia<double> Gz_expected(I_perp, I_perp, I_axial);
>   // Compute the unit inertia for a cylinder oriented along the z-axis
> ```

FYI  Missing a period .
___
*[`multibody/tree/test/unit_inertia_test.cc` line 647 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMMK3S1XUrz2J5AjiY:-NbMMK3S1XUrz2J5AjiZ:b-qvfqju) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/test/unit_inertia_test.cc#L647)):*
> ```cpp
>   const double r = 2.5;
>   const double L = 1.5;
>   UnitInertia<double> G_expected = UnitInertia<double>::SolidCylinderAboutEnd(
> ```

FYI  Consider adding const, e.g.,   
const UnitInertia<double> G_expected ...
Similarly consider adding const for G2 and G3 (below).
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1159 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMNV9i7SyiLugseLQr:-NbMNV9i7SyiLugseLQs:be0h7e5) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1159)):*
> ```cpp
>                 const Vector3<T>&>(&Class::SolidCylinder),
>             py::arg("radius"), py::arg("length"), py::arg("unit_vector"),
>             cls_doc.SolidCylinder.doc)
> ```

FYI:  The code below deprecates calls to SolidCylinder() with only 2 arguments.
The default value of the optional 3rd parameter was Vector3<T>::UnitZ().

Similarly for the next block which deprecates calls to SolidCapsule() with only 2 arguments.

sherm1 (commented):
Feature review checkpoint.

> Reviewed 7 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-NbQJOBvC1jPyT7MOqU0:bfqjpd9)* status: 12 unresolved discussions, LGTM missing from assignee sherm1(platform), needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/unit_inertia.h` line 270 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8ZMv809lBMUg696z:-NbQ8ZMv809lBMUg697-:b-qkgeuf) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L270)):*
> ```objc
>   /// @param[in] r The radius of the cylinder, it must be non-negative.
>   /// @param[in] L The length of the cylinder, it must be non-negative.
>   /// @param[in] b_E
> ```

minor: this argument is no longer present in the deprecated signature
___
*[`multibody/tree/unit_inertia.h` line 286 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ7QAkBBgC-sqVJBBA:-NbQ7QAkBBgC-sqVJBBB:b-hkzupg) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L286)):*
> ```objc
>   ///   - `b_E` is the zero vector. That is if `‖b_E‖₂ ≤ ε`, where ε is the
>   ///     machine epsilon.
>   DRAKE_DEPRECATED("2023-11-01",
> ```

nit: our 3 month minimum deprecation promise means this has to be 2023-12-01 everywhere in this PR
___
*[`multibody/tree/unit_inertia.h` line 315 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8pml3-_LIjVbmqYr:-NbQ8pml3-_LIjVbmqYs:b-qkgeuf) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L315)):*
> ```objc
>   /// @param[in] r radius of the cylinder/half-sphere parts of the capsule.
>   /// @param[in] L length of the cylindrical part of the capsule.
>   /// @param[in] unit_vector direction of the cylindrical part of the capsule.
> ```

minor: this argument is no longer present in the deprecated signature
___
*[`multibody/tree/unit_inertia.h` line 348 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ9ITk7sqf7OxLZVSy:-NbQ9ITk7sqf7OxLZVSz:bli6bq8) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L348)):*
> ```objc
>   /// @param[in] L The length of the cylinder.
>   /// @throws std::exception if r or L is negative.
>   DRAKE_DEPRECATED("2023-11-01",
> ```

BTW consider changing this date to 2023-12-01 to sync it with the new related deprecations
___
*[`multibody/tree/unit_inertia.h` line 408 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQBROu6ndtxCJnU1YV:-NbQBROu6ndtxCJnU1YW:b-pq9vj1) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L408)):*
> ```objc
>   /// cylinders or cones and propellers with 3⁺ evenly spaced blades.
>   /// For a body B with axially symmetric inertia, B's unit inertia about a
>   /// point Bp on axis L can be written in terms of a unit_vector parallel to L;
> ```

BTW when I read this I was expecting to see Bp as an argument. But I see now this is just some point in the caller's mind that affects the parallel (axial) inertia value. Can you clarify that somehow?
___
*[`multibody/tree/unit_inertia.h` line 428 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMIwko6cYDCZn99dQM:-NbQAn1X1OBmXCA2zjJD:b-6shteb) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L428)):*

FYI: Another name for the parameter moment_parallel could be moment_axial.
</blockquote></details>

I could see either axial/radial or parallel/perpendicular but I think they should be a matched pair. If they are interchangeable everywhere I'd be inclined to go with the shorter words but I'm OK with either pair.
___
*[`multibody/tree/unit_inertia.cc` line 12 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQExUy0Jnbh-byUyuf:-NbQExUy0Jnbh-byUyug:b71ube) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L12)):*
> ```cpp
> // Throws unless ‖unit_vector‖ is within ≈ 5.5 bits of 1.0.
> // Note: 1E-14 ≈ 2^5.5 * std::numeric_limits<double>::epsilon();
> // Note: This function is a no-op when type T is symbolic::Expression.
> ```

BTW consider noting that 1e-14 is the threshold we use to decide whether a user's alleged unit vector is acceptable, but is not nearly as "unit" as we typically expect when normalizing, which is an extremely well-behaved calculation typically 100X better (1±1e-16). So unless the length is exactly 1 (common since coordinate axes are popular directions) we should normalize internally even if the user's vector passes this test.
___
*[`multibody/tree/unit_inertia.cc` line 25 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQDSKE9z8crt2SS3C6:-NbQDSKE9z8crt2SS3C7:b-6gqyg4) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L25)):*
> ```cpp
>     // 𝐯⋅𝐯 ≤ (1 + ε)²  and 𝐯⋅𝐯 ≥ (1 - ε)².  Distributing the square results in
>     // 𝐯⋅𝐯 ≤ 1 + 2 ε + ε²  and 𝐯⋅𝐯 ≥ 1 - 2 ε + ε². Since ε² ≪ 2 ε, this gives
>     // 𝐯⋅𝐯 - 1 ≤ 2 ε   and 𝐯⋅𝐯 - 1 ≥ -2 ε  or  |𝐯⋅𝐯 − 1| ≤ 2 ε
> ```

BTW Nice! I don't think I've seen that derivation before.
___
*[`multibody/tree/unit_inertia.cc` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQHxqn50ADlt8JIxVw:-NbQHxqn50ADlt8JIxVx:b1ks04n) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L32)):*
> ```cpp
>     // -------------------------------------------------------------
>     constexpr double kTolerance2 = 2E-14;
>     if (abs(unit_vector.squaredNorm() - 1) > kTolerance2) {
> ```

BTW consider returning `abs(unit_vector.squaredNorm() - 1` as the return value of this function if it doesn't fail. The caller could use that to decide at no cost whether a normalization is needed (e.g. if this return is zero then we're good).
___
*[`multibody/tree/unit_inertia.cc` line 82 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQIatgExirLNPgLug3:-NbQIatgExirLNPgLug4:budub8g) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L82)):*
> ```cpp
>   const T l2 = length * length;
>   const T J = 0.5 * r2;               // Axial moment of inertia J = ½ r².
>   const T K = 0.25 * r2 + l2 / 12.0;  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
> ```

BTW arggh l2/12!
___
*[`multibody/tree/unit_inertia.cc` line 112 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQFxBxA7PPEI-VdlBS:-NbQFxBxA7PPEI-VdlBT:bxszfpe) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L112)):*
> ```cpp
>   // K_Bcm (perpendicular moment of inertia about Bcm) as K_Bp = K_Bcm + dist²,
>   // where dist is the distance between points Bp and Bcm.
>   DRAKE_THROW_UNLESS(moment_parallel <= 2.0 * moment_perpendicular);
> ```

BTW should we allow some epsilon slop here or require the user to clean up their own numerics? I'm fine either way as long as it's documented. 
___
*[`multibody/tree/unit_inertia.cc` line 121 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQEVn05E_wZMCvtvNY:-NbQEVn05E_wZMCvtvNZ:b-yn83va) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L121)):*
> ```cpp
>   // Form B's unit inertia about a point Bp on B's symmetry axis,
>   // expressed in the same frame E as the unit_vector is expressed.
>   Matrix3<T> G_BBp_E =
> ```

nit: uvec and G should be const
___
*[`multibody/tree/unit_inertia.cc` line 192 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMJwSS4v8CDvwGbkNE:-NbQH8tW8WQXo7Tnkdy5:b-kgrv2h) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L192)):*

FYI  Change to dH = ⅜ radius + ½ length.
</blockquote></details>

minor: comment fix needed
___
*[`multibody/tree/unit_inertia.cc` line 217 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQHOZM1VFoTQvCigog:-NbQHOZM1VFoTQvCigoh:b-cjext8) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L217)):*
> ```cpp
>   // more efficient result by factoring on mh and mc and computing numbers as
>   const T l2 = length * length;
>   const T Ixx = mc * (l2/12.0 + 0.25*r2) + mh * (0.51875*r2 + 2*dH*dH);
> ```

BTW wow l2/12 is hard to read! Consider l2 -> length2
___
*[`multibody/tree/test/unit_inertia_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMLgPB8zxHRJtwQ4i-:-NbQIrdR0b4ZM6Pt65eJ:boynei5) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/test/unit_inertia_test.cc#L215)):*

FYI  Missing a period .
</blockquote></details>

nit comment fix needed

jwnimmer-tri (commented):
> Reviewed 5 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 16 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQXVf01-SmuWOQFilM:-NbQXVf13q3RU-XjmeTg:bsvhboh) (no related file):*
This is our deprecation policy: https://drake.mit.edu/stable.html.  In particular, it says:

> On very rare occasions it is impractical, too expensive, or logically unsound to meet the full three month deprecation window. In these cases the change will be announced in the “Breaking changes” section of the release notes.

In this PR, we have a breaking change to newly throw when a `unit_vector` is too far from unit.  I was unable to figure out why using deprecation for that (instead of a breaking change) would be impractical, expensive, or unsound.  If there is such a reason, you could reply here to help make it clear.

My assumption is that in these functions, the response to an out-of-tolerance unit vector could be to [warn once](https://drake.mit.edu/doxygen_cxx/structdrake_1_1logging_1_1_warn.html) and then re-normalize.  After the deprecation window, we would replace the warning with an exception.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-NbQZ8GB5FnOZ-FNkyzn:b-d76ycd) (no related file):*
The attempts are python deprecations are missing a lot of important pieces.  I noted a few places that I started to find, but overall it needs a lot of work.  Let me know if you'd like me to push a patch to this branch with the proper code.

___
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1169 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ34xE1zWLDl_-9ri:-NbQZ34yE3dFercytaXP:buzsptd) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1169)):*
> ```cpp
> #pragma GCC diagnostic pop
>                 }),
>             py::arg("r"), py::arg("L"), cls_doc.SolidCylinder.doc_deprecated)
> ```

We are missing the defaulted `unit_vector` kwarg argument here.
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1202 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQYnLv2_JetqC6gelr:-NbQYnLw0URf2pPNfdzZ:byz007p) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1202)):*
> ```cpp
>             cls_doc.SolidCylinderAboutEnd.doc_deprecated)
>         .def_static("AxiallySymmetric", &Class::AxiallySymmetric,
>             py::arg("moment_parallel"), py::arg("moment_perpendicular"),
> ```

It looks to me like a bunch of the argument-name-change deprecation shims are missing.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 4 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQXVf01-SmuWOQFilM:-NbSH3z61TRh1rjDRI8G:b-rnlfo9) (no related file):*

This is our deprecation policy: https://drake.mit.edu/stable.html.  In particular, it says:

> On very rare occasions it is impractical, too expensive, or logically unsound to meet the full three month deprecation window. In these cases the change will be announced in the “Breaking changes” section of the release notes.

In this PR, we have a breaking change to newly throw when a `unit_vector` is too far from unit.  I was unable to figure out why using deprecation for that (instead of a breaking change) would be impractical, expensive, or unsound.  If there is such a reason, you could reply here to help make it clear.

My assumption is that in these functions, the response to an out-of-tolerance unit vector could be to [warn once](https://drake.mit.edu/doxygen_cxx/structdrake_1_1logging_1_1_warn.html) and then re-normalize.  After the deprecation window, we would replace the warning with an exception.
</blockquote></details>

I removed the label about breaking changes.
I am unsure how to warn once.

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-NbSHTqdDOnygB0RE-2s:b-7co769) (no related file):*

The attempts are python deprecations are missing a lot of important pieces.  I noted a few places that I started to find, but overall it needs a lot of work.  Let me know if you'd like me to push a patch to this branch with the proper code.
</blockquote></details>

Assuming this new commit passes CI, that would be _great_.
Thanks Jeremy.

___
___
*[`multibody/tree/unit_inertia.h` line 270 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8ZMv809lBMUg696z:-NbRwJ-OCpQ7YPn6ORzX:b9i1s0p) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L270)):*

minor: this argument is no longer present in the deprecated signature
</blockquote></details>

Done.  Removed all references to b_E.
___
*[`multibody/tree/unit_inertia.h` line 286 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ7QAkBBgC-sqVJBBA:-NbRy1LAA9Pa4MSbpv8S:b-snh328) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L286)):*

nit: our 3 month minimum deprecation promise means this has to be 2023-12-01 everywhere in this PR
</blockquote></details>

Done here and elsewhere.
___
*[`multibody/tree/unit_inertia.h` line 315 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8pml3-_LIjVbmqYr:-NbRzUAS7895uzb64S_x:bqfmj37) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L315)):*

minor: this argument is no longer present in the deprecated signature
</blockquote></details>

Done.  Removed all references to unit_vector.
___
*[`multibody/tree/unit_inertia.h` line 348 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ9ITk7sqf7OxLZVSy:-NbRz_ML6AEQPbaddBzJ:b-snh328) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L348)):*

BTW consider changing this date to 2023-12-01 to sync it with the new related deprecations
</blockquote></details>

Done here and elsewhere.
___
*[`multibody/tree/unit_inertia.h` line 408 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQBROu6ndtxCJnU1YV:-NbS01LOC8ZPlVpF81_q:b-weam9l) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L408)):*

BTW when I read this I was expecting to see Bp as an argument. But I see now this is just some point in the caller's mind that affects the parallel (axial) inertia value. Can you clarify that somehow?
</blockquote></details>

Done.   See new @pre that was added.
___
*[`multibody/tree/unit_inertia.h` line 428 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMIwko6cYDCZn99dQM:-NbRzhqX7QFVvy93o3yg:b-x83692) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L428)):*

I could see either axial/radial or parallel/perpendicular but I think they should be a matched pair. If they are interchangeable everywhere I'd be inclined to go with the shorter words but I'm OK with either pair.
</blockquote></details>

Done.  Leaving them as-is.
___
*[`multibody/tree/unit_inertia.cc` line 12 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQExUy0Jnbh-byUyuf:-NbS94h933zhVRBzvHVr:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L12)):*

BTW consider noting that 1e-14 is the threshold we use to decide whether a user's alleged unit vector is acceptable, but is not nearly as "unit" as we typically expect when normalizing, which is an extremely well-behaved calculation typically 100X better (1±1e-16). So unless the length is exactly 1 (common since coordinate axes are popular directions) we should normalize internally even if the user's vector passes this test.
</blockquote></details>

Done.
___
*[`multibody/tree/unit_inertia.cc` line 25 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQDSKE9z8crt2SS3C6:-NbS9IYS6QuFKBd4lN8g:bkn5kkh) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L25)):*

BTW Nice! I don't think I've seen that derivation before.
</blockquote></details>

Done. Thanks Sherm.
___
*[`multibody/tree/unit_inertia.cc` line 32 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQHxqn50ADlt8JIxVw:-NbS9SapEgKoa3E6hpRh:bbwquq8) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L32)):*

BTW consider returning `abs(unit_vector.squaredNorm() - 1` as the return value of this function if it doesn't fail. The caller could use that to decide at no cost whether a normalization is needed (e.g. if this return is zero then we're good).
</blockquote></details>

Done.  Per f2f -- return magnitude squared of unit_vector.
___
*[`multibody/tree/unit_inertia.cc` line 82 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQIatgExirLNPgLug3:-NbS9ZSj9P6oAHyckk8M:b1zvu5u) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L82)):*

BTW arggh l2/12!
</blockquote></details>

Done l2 -> lsq.
___
*[`multibody/tree/unit_inertia.cc` line 112 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQFxBxA7PPEI-VdlBS:-NbSBjEpE1e5tRrvKLmj:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L112)):*

BTW should we allow some epsilon slop here or require the user to clean up their own numerics? I'm fine either way as long as it's documented. 
</blockquote></details>

Done.
___
*[`multibody/tree/unit_inertia.cc` line 121 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQEVn05E_wZMCvtvNY:-NbSCSC3EpzXPnx19Fhb:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L121)):*

nit: uvec and G should be const
</blockquote></details>

Done.
___
*[`multibody/tree/unit_inertia.cc` line 192 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMJwSS4v8CDvwGbkNE:-NbSDP2wA1fxMDG-HDRO:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L192)):*

minor: comment fix needed
</blockquote></details>

Done.
___
*[`multibody/tree/unit_inertia.cc` line 217 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQHOZM1VFoTQvCigog:-NbSDVAe0fUFcHfxQBMV:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.cc#L217)):*

BTW wow l2/12 is hard to read! Consider l2 -> length2
</blockquote></details>

Done.
___
*[`multibody/tree/test/unit_inertia_test.cc` line 215 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbMLgPB8zxHRJtwQ4i-:-NbSDd8BEFkfVHNv8hR6:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/test/unit_inertia_test.cc#L215)):*

nit comment fix needed
</blockquote></details>

Done.
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1169 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ34xE1zWLDl_-9ri:-NbSDlFd3ZcLorctoNia:b-sflx78) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1169)):*

We are missing the defaulted `unit_vector` kwarg argument here.
</blockquote></details>

Yes.  I am unsure how to handle this.
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1202 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQYnLv2_JetqC6gelr:-NbSDoFn9Ymd3VxExueZ:bt26swq) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1202)):*

It looks to me like a bunch of the argument-name-change deprecation shims are missing.
</blockquote></details>

Yes.  I agree.  I appreciate your offer to help by pushing a fix.

jwnimmer-tri (commented):
> Reviewed 3 of 3 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQXVf01-SmuWOQFilM:-NbUoY1t2kASji7Uhr_A:b-957hxm) (no related file):*
Done

> I am unsure how to warn once.

My comment above hyperlinked to the documentation about how to do it.  (Maybe the underline is a bit too subtle in Reviewable.)

In any case, I've pushed the fix:

```
[ RUN      ] UnitInertia.SolidCylinder
[2023-08-10 08:16:49.686] [console] [warning] SolidCylinder(): The unit_vector argument 1 2 3 is not a unit vector. Implicit normalization is deprecated; on or after 2023-12-01 this will become an exception.
[       OK ] UnitInertia.SolidCylinder (0 ms)
```

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-NbV4Nuk1A-JTlXYeKGg:b-6dgg5f) (no related file):*

Assuming this new commit passes CI, that would be _great_.
Thanks Jeremy.
</blockquote></details>

I'm still working on this part.

___
___
*[`multibody/tree/unit_inertia.h` line 270 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8ZMv809lBMUg696z:-NbV9oeZBPoLtB-sAqsq:b-1d192p) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L270)):*
The removal of the docs was my change, because it's a colossal pain in my ass to maintain the two copies of the documentation up to date.

Given this code in the header file ...

```c++
  static UnitInertia<T> SolidCylinder(const T& r, const T& L) {
    return SolidCylinder(r, L, Vector3<T>::UnitZ());
```

... I think the typical programmer will be able to figure out what's going on without very much difficulty at all.

> I don't think that's a good idea because the existing documentation can be helpful for a person converting from the deprecated function to the replacement.

I'm interpreting this as a suggestion, not a defect.  If there is a defect, please clarify.
___
*[`multibody/tree/unit_inertia.cc` line 70 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV6jfO6qcmnnWBQYj1:-NbV9KlI-PwQYjrH6E4N:bojl4m6) ([raw file](https://github.com/RobotLocomotion/drake/blob/e3f4e53867bd7f4ef04fec7291a24f341d149f69/multibody/tree/unit_inertia.cc#L70)):*

Our styleguide [prohibits catching exceptions](https://drake.mit.edu/styleguide/cppguide.html#Exceptions). You'll have to find a different way of doing this.
</blockquote></details>

We've previously allowed it within a small, bounded scope when the alternatives are substantially worse.

In any case, I've pushed the worse version now.
___
*[`multibody/tree/unit_inertia.cc` line 167 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV7Ysr99DkqyTzAxhw:-NbVBCDf91_msNDr5rQI:b-1j9o4v) ([raw file](https://github.com/RobotLocomotion/drake/blob/e3f4e53867bd7f4ef04fec7291a24f341d149f69/multibody/tree/unit_inertia.cc#L167)):*

BTW consider including a TODO comment with the deprecation date for each of these "Warns" to change them to "Throws"
</blockquote></details>

OK Won't do.

In December, I will not have any problem knowing what to do here.
___
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-NbV5euh96xAdxtdvO2F:b-dde9vk) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*
> ```cpp
>   const T J = 0.5 * rsq;                // Axial moment of inertia J = ½ r².
>   const T K = 0.25 * rsq + lsq / 12.0;  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
>   return AxiallySymmetric(J, K, unit_vector);
> ```

nit For my money, keeping the code clear here vastly outweights the 1 flop we might be saving introducing temporaries with cryptic names.  Do we actually expect people to call `SolidCylinder` as part of an inner loop?  (If we do care about throughput, why are we using `/ 12.0` instead of `* (1 / 12.0)`?)

My assumption is that the temporary names were intended for clarity (not performance).  To me, this is more clear:

```c++
  // Axial moment of inertia J = ½ r².
  const T J = 0.5 * radius * radius;
  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
  const T K = 0.25 * radius * radius + length * length / 12.0;
  return AxiallySymmetric(J, K, unit_vector);
```

Ditto for all of the other code throughout.

Or if you don't want to go that far, the obvious response to Sherm's "ick" on `l2` would be to named the temporary like `length2`.  That better matches up with the math in the comments.

```c++
  const T radius2 = radius * radius;
  const T length2 = length * length;
  // Axial moment of inertia J = ½ r².
  const T J = 0.5 * radius2;
  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
  const T K = 0.25 * radius2 + length2 / 12.0;
  return AxiallySymmetric(J, K, unit_vector);
```
___
*[`multibody/tree/unit_inertia.cc` line 214 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVBwXL8BxsRWppcJmM:-NbVBwXM1wd9wl1EpJxJ:b-hznhbi) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L214)):*
> ```cpp
> 
>   // Calculate vc (the volume of cylinder C) and vh (volume of half-sphere H).
>   const T rsq = radius * radius;
> ```

nit Cherry-picking one more example of a rename adjustment that made readability  _worse_ not _better_.   There are more nearby.



_Suggestion:_
```cpp
radius2
```
___
*[`multibody/tree/test/unit_inertia_test.cc` line 254 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVEU9UB1GTJVTlvXzF:-NbVEU9UB1GTJVTlvXzG:b-i1aai4) ([raw file](https://github.com/RobotLocomotion/drake/blob/92e830d0bb16d5fc6bc64a295e39b384ddcab29c/multibody/tree/test/unit_inertia_test.cc#L254)):*
> ```cpp
> // Tests the static method to obtain the unit inertia of a solid cylinder B
> // computed about a point Bp at the center of its base.
> GTEST_TEST(UnitInertia, SolidCylinderAboutEnd) {
> ```

It looks like the `UnitInertia.SolidCylinderAboutEnd` test case got deleted by mistake in 37628e8bd49ec4b1b14ead25e66baf88259a45a5.  We should put it back, right?

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 7 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/unit_inertia.h` line 270 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQ8ZMv809lBMUg696z:-NbVH8hr1lzu8C2iadZE:bub427) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/multibody/tree/unit_inertia.h#L270)):*

The removal of the docs was my change, because it's a colossal pain in my ass to maintain the two copies of the documentation up to date.

Given this code in the header file ...

```c++
  static UnitInertia<T> SolidCylinder(const T& r, const T& L) {
    return SolidCylinder(r, L, Vector3<T>::UnitZ());
```

... I think the typical programmer will be able to figure out what's going on without very much difficulty at all.

> I don't think that's a good idea because the existing documentation can be helpful for a person converting from the deprecated function to the replacement.

I'm interpreting this as a suggestion, not a defect.  If there is a defect, please clarify.
</blockquote></details>

The one thing we _could_ do is improve the deprecation message string to contain more detail(s) or suggestion(s).  If you have ideas on that front, feel free to push your improvements to the branch.

sherm1 (commented):
BTW CI didn't like this:
```
multibody/tree/unit_inertia.cc:84:32: error: unused variable 'log_once' [-Werror,-Wunused-variable]
    static const logging::Warn log_once(
```

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-NbVOIal3RkPuU_tCwVG:b-vhzg45)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

jwnimmer-tri:
Aha, thanks!  Apparently the config I use (Jammy + GCC) was the only one that passed.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`common/text_logging.h` line 200 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVXtKT6ClDdOChhfK6:-NbVXtKU89QFwWj6tE77:b-ht5xvz) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9b2df2b21a0f3ca390ddd24cb72ad7a095e15ac/common/text_logging.h#L200)):*
> ```objc
> /// }
> /// </pre>
> struct[[maybe_unused]] Warn {
> ```

FYI This is the (unfortunate) whitespace that cpplint requires.  I decided not to fight it for now.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-NbX50Og0yVH650IZ9T-:b-ky3xa9) (no related file):*

I'm still working on this part.
</blockquote></details>

OK -- I made some comments.  After you are done, I'll implement subsequent changes.

___
___
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-NbX4AuG2JFqv4IZJKez:bj1g76d) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*

nit For my money, keeping the code clear here vastly outweights the 1 flop we might be saving introducing temporaries with cryptic names.  Do we actually expect people to call `SolidCylinder` as part of an inner loop?  (If we do care about throughput, why are we using `/ 12.0` instead of `* (1 / 12.0)`?)

My assumption is that the temporary names were intended for clarity (not performance).  To me, this is more clear:

```c++
  // Axial moment of inertia J = ½ r².
  const T J = 0.5 * radius * radius;
  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
  const T K = 0.25 * radius * radius + length * length / 12.0;
  return AxiallySymmetric(J, K, unit_vector);
```

Ditto for all of the other code throughout.

Or if you don't want to go that far, the obvious response to Sherm's "ick" on `l2` would be to named the temporary like `length2`.  That better matches up with the math in the comments.

```c++
  const T radius2 = radius * radius;
  const T length2 = length * length;
  // Axial moment of inertia J = ½ r².
  const T J = 0.5 * radius2;
  // Transverse moment K = ¼ r² + ¹⁄₁₂ l².
  const T K = 0.25 * radius2 + length2 / 12.0;
  return AxiallySymmetric(J, K, unit_vector);
```
</blockquote></details>

FYI  Your suggestions sound fine.  I discussed Sherm's "ick" with him and it had more to do with l2 looking like the number 12 -- and thought the rsq and lsq were improvements.  I'll implement as you suggest (here and elsewhere).
___
*[`multibody/tree/unit_inertia.cc` line 214 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVBwXL8BxsRWppcJmM:-NbX4r1O7CLfZ280SmSc:bd5fspv) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L214)):*

nit Cherry-picking one more example of a rename adjustment that made readability  _worse_ not _better_.   There are more nearby.
</blockquote></details>

FYI Sounds good -- I'll implement.
___
*[`multibody/tree/unit_inertia.cc` line 23 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbX1mLxA3y2BRGIxfwU:-NbX1mLxA3y2BRGIxfwV:bim0ucf) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9b2df2b21a0f3ca390ddd24cb72ad7a095e15ac/multibody/tree/unit_inertia.cc#L23)):*
> ```cpp
> // even if unit_vector passes this test. Usually this is accomplished by handing
> // the unit_vector off to other function(s) that all eventually call down into
> // AxiallySymmetric() which does the normalization.
> ```

FYI:  Consider wording that is more generic.



_Suggestion:_
```cpp
// directions), the calling function should consider normalizing unit_vector
// even if unit_vector passes this test or hand the unit_vector to another
// function that eventually does the normalization.
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: 6 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-NbX6E_d389KlE1Wb0Z6:b5p38ks) (no related file):*

OK -- I made some comments.  After you are done, I'll implement subsequent changes.
</blockquote></details>

BTW I'm done with all of my C++ changes, feel free to push any C++ edits you'd like, at whatever pace.  My only remaining work is on the Python part.  Pushing C++ changes will not interfere with my work on Python.

___
___
*[`multibody/tree/unit_inertia.cc` line 23 at r6](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbX1mLxA3y2BRGIxfwU:-NbX5wnB8MpJ8zvvE_Nv:b-rrezjg) ([raw file](https://github.com/RobotLocomotion/drake/blob/a9b2df2b21a0f3ca390ddd24cb72ad7a095e15ac/multibody/tree/unit_inertia.cc#L23)):*

FYI:  Consider wording that is more generic.
</blockquote></details>

My goal here was to be *as specific as possible*.  It took me 10 minutes to chase down exactly where and how were normalizing our vectors, to convince myself that we were doing it 100% of the time and didn't miss any.

jwnimmer-tri (commented):
:lgtm: platform.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-Nb_GNPT9haeKs_7qEaJ:b-vtj4vo)* status: 3 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ8GA3O8KOl-xqcCH:-Nb_Eg_x5QOJ2OkQjuY4:b-kxlj53) (no related file):*

BTW I'm done with all of my C++ changes, feel free to push any C++ edits you'd like, at whatever pace.  My only remaining work is on the Python part.  Pushing C++ changes will not interfere with my work on Python.
</blockquote></details>

My python edits are done.  I think the deprecation shims are complete now.

___
___
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-Nb_GTko2aA7pBtmhR0X:b-51ht2x) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*

FYI  Your suggestions sound fine.  I discussed Sherm's "ick" with him and it had more to do with l2 looking like the number 12 -- and thought the rsq and lsq were improvements.  I'll implement as you suggest (here and elsewhere).
</blockquote></details>

(Yes, I had grokked that `l2` vs `12` was the complaint.)

I think the idea to carry forward to the future is to be suspicious when abbreviating that is _ever_ an improvement.  In most cases it's not.
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1169 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQZ34xE1zWLDl_-9ri:-Nb_FJLUBZ5WKt8Uic_g:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1169)):*

Yes.  I am unsure how to handle this.
</blockquote></details>

Done
___
*[`bindings/pydrake/multibody/tree_py.cc` line 1202 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbQYnLv2_JetqC6gelr:-Nb_EmzwAepg4jSQ0_ys:bpf1txc) ([raw file](https://github.com/RobotLocomotion/drake/blob/dfd55dc02a32377f4d790a4111125da3eaf2c27b/bindings/pydrake/multibody/tree_py.cc#L1202)):*

Yes.  I agree.  I appreciate your offer to help by pushing a fix.
</blockquote></details>

Done

The important clue is that when deprecating python functions, the _only_ change we should be making to the pre-existing unit test assertions is to wrap them in `catch_drake_warnings`.  The original unit test changes in this PR were editing and/or deleting pre-existing test cases, which should be a red flag that breaking changes are happening, and so a clue to ask for help.  I went back and grabbed the old test cases from git master, and copied them back into the updated unit test (modulo some whitespace reformatting).

jwnimmer-tri:
FYI the current `mac-arm-ventura-clang-bazel-experimental-release` failure is a flake, and can be ignored.  It'll automatically clear itself up after the next push.

sherm1 (commented):
Notation comments addressed, PTAL @jwnimmer-tri . If you can live with that, this is ready to merge.

> Dismissed @mitiguy from a discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-Nbv09fP5s2OA76yy3oN:b-g14a3z)* status: 2 unresolved discussions, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits) (waiting on @mitiguy)

___
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-NbuwqufAERHGUf4TmYv:b-y07w3r) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*

(Yes, I had grokked that `l2` vs `12` was the complaint.)

I think the idea to carry forward to the future is to be suspicious when abbreviating that is _ever_ an improvement.  In most cases it's not.
</blockquote></details>

OK, looking this over just now I prefer Paul's current notation because it is easier to match up with the math. To my eyes rsq looks more like r² than does radius2. r2 would be best but is precluded by the awfulness of l2 being proximal to 12!

There is an advantage to these otherwise-bad abbreviations when they match nearby math symbols. For the same reason it's nice to have the temporaries for `radius * radius` etc. (performance irrelevance notwithstanding) since spelling it out takes it further from the good notation.

The mix of full spelling in the API and terse spelling for the math is best IMO.
___
*[`multibody/tree/unit_inertia.cc` line 214 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVBwXL8BxsRWppcJmM:-NbuzUkZ-zfyiSmLaVL2:b-ywejqp) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L214)):*

FYI Sounds good -- I'll implement.
</blockquote></details>

OK as is for notation match per comments above

sherm1 (commented):
Wait, one comment I missed ...

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-Nbv2NPJ2pgfpZtQH-CD:b5n3f49)* status: 2 unresolved discussions

sherm1 (commented):
OK, PTAL now. Ready to merge pending CI.

> Reviewed 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-:-Nbv5O-31imXvGts766x:bpnt1qz)* status: 1 unresolved discussion

___
*[`multibody/tree/test/unit_inertia_test.cc` line 254 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbVEU9UB1GTJVTlvXzF:-Nbv4zdr9MslFxMzVPFP:b-376vyo) ([raw file](https://github.com/RobotLocomotion/drake/blob/92e830d0bb16d5fc6bc64a295e39b384ddcab29c/multibody/tree/test/unit_inertia_test.cc#L254)):*

It looks like the `UnitInertia.SolidCylinderAboutEnd` test case got deleted by mistake in 37628e8bd49ec4b1b14ead25e66baf88259a45a5.  We should put it back, right?
</blockquote></details>

Done. Good catch!

jwnimmer-tri (commented):
> Reviewed 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19926)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),sherm1(platform)

___
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-Nbv1JLIBHdyhhpJ1Ck9:b2wgore) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*

OK, looking this over just now I prefer Paul's current notation because it is easier to match up with the math. To my eyes rsq looks more like r² than does radius2. r2 would be best but is precluded by the awfulness of l2 being proximal to 12!

There is an advantage to these otherwise-bad abbreviations when they match nearby math symbols. For the same reason it's nice to have the temporaries for `radius * radius` etc. (performance irrelevance notwithstanding) since spelling it out takes it further from the good notation.

The mix of full spelling in the API and terse spelling for the math is best IMO.
</blockquote></details>

The only valid abbreviation per GSG is to truncate the tail of a word, but neither `rsq` nor `lsq` is prefix of any word.  The short variable here serve only to confuse rather than clarify.

I will relent and mark myself satisfied this time only out of exhaustion.  I'll not allow it next time around.

sherm1:
*[`multibody/tree/unit_inertia.cc` line 114 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19926#-NbV5eug2iQvAaX-wK8x:-NbvLVwW2uotLwLld0mg:b-i13adu) ([raw file](https://github.com/RobotLocomotion/drake/blob/54113320e7d425fd1532427a262d40d1615639e4/multibody/tree/unit_inertia.cc#L114)):*
Thanks. Next time we can argue whether this text from our styleguide applies:
> Exception: Variable and method names may violate both the capitalization standard and the “long, human-readable” standard above if a short, non-compliant name more closely matches the common conventions of the field.

PR #19939, opened 2023-08-05
[multibody] Introduce plant.RenameModelInstance()

rpoyner-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)

### DIFF
--- a/bindings/pydrake/geometry/geometry_py_scene_graph.cc
+++ b/bindings/pydrake/geometry/geometry_py_scene_graph.cc
@@ -191,6 +191,8 @@ void DoScalarDependentDefinitions(py::module m, T) {
                 &Class::RegisterFrame),
             py::arg("source_id"), py::arg("parent_id"), py::arg("frame"),
             cls_doc.RegisterFrame.doc_3args)
+        .def("RenameFrame", &Class::RenameFrame, py::arg("frame_id"),
+            py::arg("name"), cls_doc.RenameFrame.doc)
         .def("RegisterGeometry",
             py::overload_cast<SourceId, FrameId,
                 std::unique_ptr<GeometryInstance>>(&Class::RegisterGeometry),
@@ -207,6 +209,8 @@ void DoScalarDependentDefinitions(py::module m, T) {
                 &Class::RegisterAnchoredGeometry),
             py::arg("source_id"), py::arg("geometry"),
             cls_doc.RegisterAnchoredGeometry.doc)
+        .def("RenameGeometry", &Class::RenameGeometry, py::arg("geometry_id"),
+            py::arg("name"), cls_doc.RenameGeometry.doc)
         .def("ChangeShape",
             py::overload_cast<SourceId, GeometryId, const Shape&,
                 std::optional<math::RigidTransform<double>>>(
--- a/bindings/pydrake/geometry/test/scene_graph_test.py
+++ b/bindings/pydrake/geometry/test/scene_graph_test.py
@@ -33,6 +33,8 @@ class TestGeometrySceneGraph(unittest.TestCase):
         global_frame = scene_graph.RegisterFrame(
             source_id=global_source,
             frame=mut.GeometryFrame("anchored_frame1"))
+        scene_graph.RenameFrame(frame_id=global_frame, name="something")
+        scene_graph.RenameFrame(frame_id=global_frame, name="anchored_frame1")
         scene_graph.RegisterFrame(
             source_id=global_source, parent_id=global_frame,
             frame=mut.GeometryFrame("anchored_frame2"))
@@ -47,6 +49,8 @@ class TestGeometrySceneGraph(unittest.TestCase):
             geometry=mut.GeometryInstance(X_PG=RigidTransform_[float](),
                                           shape=mut.Sphere(1.),
                                           name="sphere2"))
+        scene_graph.RenameGeometry(geometry_id=sphere_2, name="else")
+        scene_graph.RenameGeometry(geometry_id=sphere_2, name="sphere2")
         props = mut.ProximityProperties()
         mut.AddRigidHydroelasticProperties(resolution_hint=1, properties=props)
         scene_graph.AssignRole(source_id=global_source, geometry_id=sphere_2,
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -232,6 +232,9 @@ void DoScalarDependentDefinitions(py::module m, T) {
             py_rvp::reference_internal, py::arg("frame"), cls_doc.AddFrame.doc)
         .def("AddModelInstance", &Class::AddModelInstance, py::arg("name"),
             cls_doc.AddModelInstance.doc)
+        .def("RenameModelInstance", &Class::RenameModelInstance,
+            py::arg("model_instance"), py::arg("name"),
+            cls_doc.RenameModelInstance.doc)
         .def(
             "AddRigidBody",
             [](Class * self, const std::string& name,
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -200,6 +200,10 @@ class TestPlant(unittest.TestCase):
         body_default_rotational_inertial = body.default_rotational_inertia()
         body_default_spatial_inertial = body.default_spatial_inertia()
         new_model_instance = plant.AddModelInstance("new_model_instance")
+        plant.RenameModelInstance(model_instance=new_model_instance,
+                                  name="something")
+        plant.RenameModelInstance(model_instance=new_model_instance,
+                                  name="new_model_instance")
         body = plant.AddRigidBody(name="new_body_2",
                                   M_BBo_B=spatial_inertia,
                                   model_instance=new_model_instance)
--- a/examples/manipulation_station/manipulation_station.cc
+++ b/examples/manipulation_station/manipulation_station.cc
@@ -232,17 +232,17 @@ multibody::ModelInstanceIndex AddAndWeldModelFrom(
     const RigidTransform<double>& X_PC, MultibodyPlant<T>* plant) {
   DRAKE_THROW_UNLESS(!plant->HasModelInstanceNamed(model_name));
 
-  // Since we need to force the model name here, exploit the fact that model
-  // directives processing can do that.
+  // We need to parse model_url into a plant model instance named model_name,
+  // ignoring the model name defined within the model_url file. We accomplish
+  // that by parsing the model first and then renaming it second. If the model
+  // name defined within the model_url file was already in use, the first step
+  // would throw; to avoid that, we must enable auto renaming during its parse.
   multibody::Parser parser(plant);
-  multibody::parsing::ModelDirectives directives;
-  multibody::parsing::ModelDirective directive;
-  directive.add_model = multibody::parsing::AddModel{
-      model_url, model_name, {}, {}};
-  directives.directives.push_back(directive);
-  const auto models = ProcessModelDirectives(directives, &parser);
+  parser.SetAutoRenaming(true);
+  const auto models = parser.AddModelsFromUrl(model_url);
   DRAKE_THROW_UNLESS(models.size() == 1);
-  const multibody::ModelInstanceIndex new_model = models[0].model_instance;
+  plant->RenameModelInstance(models[0], model_name);
+  const multibody::ModelInstanceIndex new_model = models[0];
 
   const auto& child_frame = plant->GetFrameByName(child_frame_name, new_model);
   plant->WeldFrames(parent, child_frame, X_PC);
--- a/geometry/geometry_state.cc
+++ b/geometry/geometry_state.cc
@@ -789,6 +789,34 @@ FrameId GeometryState<T>::RegisterFrame(SourceId source_id, FrameId parent_id,
 }
 
 template <typename T>
+void GeometryState<T>::RenameFrame(FrameId frame_id, const std::string& name) {
+  FindOrThrow(frame_id, frames_, [frame_id]() {
+    return "Cannot rename frame with invalid frame id: " +
+        to_string(frame_id);
+  });
+  InternalFrame& frame = frames_.at(frame_id);
+  const std::string old_name(frame.name());
+  if (old_name == name) { return; }
+
+  SourceId source_id = frame.source_id();
+
+  // Edit source_frame_name_map_.
+  FrameNameSet& f_name_set = source_frame_name_map_.at(source_id);
+  f_name_set.erase(old_name);
+  const auto& [iterator, was_inserted] =
+      f_name_set.insert(std::string(name));
+  if (!was_inserted) {
+    throw std::logic_error(
+        fmt::format("Renaming frame from '{}'"
+                    " to an already existing name '{}'",
+                    old_name, name));
+  }
+
+  // Edit internal frame object.
+  frame.set_name(name);
+}
+
+template <typename T>
 GeometryId GeometryState<T>::RegisterGeometry(
     SourceId source_id, FrameId frame_id,
     std::unique_ptr<GeometryInstance> geometry) {
@@ -854,6 +882,29 @@ GeometryId GeometryState<T>::RegisterDeformableGeometry(
   AssignAllDefinedRoles(source_id, std::move(geometry));
 
   return geometry_id;
+}
+
+template <typename T>
+void GeometryState<T>::RenameGeometry(GeometryId geometry_id,
+                                      const std::string& name) {
+  InternalGeometry* geometry = GetMutableGeometry(geometry_id);
+  if (geometry == nullptr) {
+    throw std::logic_error(
+        "Cannot rename geometry with invalid geometry id: "
+        + to_string(geometry_id));
+  }
+  if (geometry->name() == name) { return; }
+
+  // Check for name uniqueness in all assigned roles. Note: if the universe of
+  // roles grows, this iteration will need to grow as well.
+  for (Role role : {Role::kProximity, Role::kIllustration, Role::kPerception}) {
+    if (geometry->has_role(role)) {
+      ThrowIfNameExistsInRole(geometry->frame_id(), role, name);
+    }
+  }
+
+  // Edit internal geometry object.
+  geometry->set_name(name);
 }
 
 template <typename T>
--- a/geometry/geometry_state.h
+++ b/geometry/geometry_state.h
@@ -346,6 +346,9 @@ class GeometryState {
   /** Implementation of SceneGraph::RegisterFrame().  */
   FrameId RegisterFrame(SourceId source_id, const GeometryFrame& frame);
 
+  /** Implementation of SceneGraph::RenameFrame().  */
+  void RenameFrame(FrameId frame_id, const std::string& name);
+
   /** Implementation of
    @ref SceneGraph::RegisterFrame(SourceId,FrameId,const GeometryFrame&)
    "SceneGraph::RegisterFrame()" with parent FrameId.  */
@@ -369,6 +372,9 @@ class GeometryState {
   GeometryId RegisterDeformableGeometry(
       SourceId source_id, FrameId frame_id,
       std::unique_ptr<GeometryInstance> geometry, double resolution_hint);
+
+  /** Implementation of SceneGraph::RenameGeometry().  */
+  void RenameGeometry(GeometryId geometry_id, const std::string& name);
 
   /** Implementation of SceneGraph::ChangeShape().  */
   void ChangeShape(
--- a/geometry/internal_frame.h
+++ b/geometry/internal_frame.h
@@ -141,6 +141,11 @@ class InternalFrame {
     child_geometries_.erase(geometry_id);
   }
 
+  /* Changes the name of the frame to `name`. */
+  void set_name(const std::string& name) {
+    name_ = name;
+  }
+
   /* The identifier used for identifying the single world frame in all
    instances of SceneGraph. The world frame will eventually have an arbitrary
    number of child frames and geometries; but there will always only be one
--- a/geometry/internal_geometry.h
+++ b/geometry/internal_geometry.h
@@ -226,6 +226,11 @@ class InternalGeometry {
     perception_props_ = std::nullopt;
   }
 
+  /* Changes the name of the geometry to `name`. */
+  void set_name(const std::string& name) {
+    name_ = name;
+  }
+
   //@}
 
   /* Returns a pointer to the geometry's reference mesh if the geometry is
--- a/geometry/scene_graph.cc
+++ b/geometry/scene_graph.cc
@@ -181,6 +181,11 @@ FrameId SceneGraph<T>::RegisterFrame(SourceId source_id, FrameId parent_id,
 }
 
 template <typename T>
+void SceneGraph<T>::RenameFrame(FrameId frame_id, const std::string& name) {
+  return model_.RenameFrame(frame_id, name);
+}
+
+template <typename T>
 GeometryId SceneGraph<T>::RegisterGeometry(
     SourceId source_id, FrameId frame_id,
     std::unique_ptr<GeometryInstance> geometry) {
@@ -216,6 +221,12 @@ GeometryId SceneGraph<T>::RegisterDeformableGeometry(
   auto& g_state = mutable_geometry_state(context);
   return g_state.RegisterDeformableGeometry(
       source_id, frame_id, std::move(geometry), resolution_hint);
+}
+
+template <typename T>
+void SceneGraph<T>::RenameGeometry(GeometryId geometry_id,
+                                   const std::string& name) {
+  return model_.RenameGeometry(geometry_id, name);
 }
 
 template <typename T>
--- a/geometry/scene_graph.h
+++ b/geometry/scene_graph.h
@@ -446,6 +446,19 @@ class SceneGraph final : public systems::LeafSystem<T> {
   FrameId RegisterFrame(SourceId source_id, FrameId parent_id,
                         const GeometryFrame& frame);
 
+  /** Renames the frame to `name`.
+
+   This method modifies the underlying model and requires a new Context to be
+   allocated. It does not modify the model versions (see @ref
+   scene_graph_versioning).
+
+   @param frame_id  The id of the frame to rename.
+   @param name  The new name.
+   @throws std::exception if a) the `frame_id` does not map to a valid frame,
+                          or b) there is already a frame named `name` from
+                          the same source. */
+  void RenameFrame(FrameId frame_id, const std::string& name);
+
   // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument, instead using
   // the source_id associated with the given `frame_id`.
   /** Registers a new rigid geometry G for this source. This hangs geometry G on
@@ -552,6 +565,20 @@ class SceneGraph final : public systems::LeafSystem<T> {
   GeometryId RegisterDeformableGeometry(
       systems::Context<T>* context, SourceId source_id, FrameId frame_id,
       std::unique_ptr<GeometryInstance> geometry, double resolution_hint) const;
+
+  /** Renames the geometry to `name`.
+
+   This method modifies the underlying model and requires a new Context to be
+   allocated. It does not modify the model versions (see @ref
+   scene_graph_versioning).
+
+   @param geometry_id  The id of the geometry to rename.
+   @param name  The new name.
+   @throws std::exception if a) the `geometry_id` does not map to a valid
+                          geometry, or b) `name` is not unique within any
+                          assigned role of the geometry in its associated
+                          frame. */
+  void RenameGeometry(GeometryId geometry_id, const std::string& name);
 
   // TODO(jwnimmer-tri) Deprecate and remove `source_id` argument.
   /** Changes the `shape` of the geometry indicated by the given `geometry_id`.
--- a/geometry/test/geometry_state_test.cc
+++ b/geometry/test/geometry_state_test.cc
@@ -921,6 +921,39 @@ TEST_F(GeometryStateTest, GetOwningSourceName) {
       "Geometry id .* does not map to a registered geometry");
 }
 
+TEST_F(GeometryStateTest, RenameFrame) {
+  SetUpSingleSourceTree();
+  DRAKE_EXPECT_THROWS_MESSAGE(
+      geometry_state_.RenameFrame(FrameId::get_new_id(), "invalid"),
+      ".*Cannot rename.*invalid frame id:.*");
+  DRAKE_EXPECT_THROWS_MESSAGE(
+      geometry_state_.RenameFrame(frames_[0], "f1"),
+      ".*already existing.*");
+
+  // Renaming a frame to its existing name does not cause errors.
+  std::string original_name0(geometry_state_.GetName(frames_[0]));
+  geometry_state_.RenameFrame(frames_[0], original_name0);
+  EXPECT_EQ(geometry_state_.GetName(frames_[0]), original_name0);
+
+  geometry_state_.RenameFrame(frames_[0], "something");
+  EXPECT_EQ(geometry_state_.GetName(frames_[0]), "something");
+}
+
+TEST_F(GeometryStateTest, RenameGeometry) {
+  SetUpSingleSourceTree();
+  DRAKE_EXPECT_THROWS_MESSAGE(
+      geometry_state_.RenameGeometry(GeometryId::get_new_id(), "invalid"),
+      ".*Cannot rename.*invalid geometry id:.*");
+
+  // Renaming a geometry to its existing name does not cause errors.
+  std::string original_name0(geometry_state_.GetName(geometries_[0]));
+  geometry_state_.RenameGeometry(geometries_[0], original_name0);
+  EXPECT_EQ(geometry_state_.GetName(geometries_[0]), original_name0);
+
+  geometry_state_.RenameGeometry(geometries_[0], "something");
+  EXPECT_EQ(geometry_state_.GetName(geometries_[0]), "something");
+}
+
 // Compares the transmogrified geometry state (embedded in its tester) against
 // the double state to confirm they have the same values/topology.
 template <typename T>
--- a/geometry/test/internal_geometry_test.cc
+++ b/geometry/test/internal_geometry_test.cc
@@ -220,6 +220,17 @@ GTEST_TEST(InternalGeometryTest, DeformableGeometry) {
   EXPECT_FALSE(rigid_geometry.is_deformable());
 }
 
+GTEST_TEST(InternalGeometryTest, Rename) {
+  InternalGeometry geometry;
+  EXPECT_TRUE(geometry.name().empty());
+  {
+    // Scope the new name to make sure the object takes a copy.
+    std::string new_name("new_name");
+    geometry.set_name(new_name);
+  }
+  EXPECT_EQ(geometry.name(), "new_name");
+}
+
 }  // namespace
 }  // namespace internal
 }  // namespace geometry
--- a/multibody/plant/multibody_plant.cc
+++ b/multibody/plant/multibody_plant.cc
@@ -47,6 +47,7 @@ using geometry::FramePoseVector;
 using geometry::GeometryFrame;
 using geometry::GeometryId;
 using geometry::GeometryInstance;
+using geometry::GeometrySet;
 using geometry::PenetrationAsPointPair;
 using geometry::ProximityProperties;
 using geometry::render::RenderLabel;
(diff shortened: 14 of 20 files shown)


rpoyner-tri (commented):
Here is a less invasive alternative to #18666.

Closes: #18666.

The motivating issue is #18059.

FYI @jwnimmer-tri @SeanCurtis-TRI for some preliminary thoughts on this one vs. the other one.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-Nb5ulGD15WcQ_b4_ANx:b-gp5dnd)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
-(status: do not merge) -(status: do not review) +(release notes: feature) 
+@jwnimmer-tri or delegate for feature review. Note the XXX question raised in test code.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbFZHJC4CPB-6h41tWn:bup7c7)* status: LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers, labeled "do not merge" (waiting on @rpoyner-tri)

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbFZbrH0zjdmDrE6bO1:b-g1crnv) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*
> ```cpp
>   // XXX Should these be disallowed?
>   EXPECT_NO_THROW(plant.RenameModelInstance(world_model_instance(), "Mars"));
>   EXPECT_NO_THROW(plant.RenameModelInstance(default_model_instance(), "hmm"));
> ```

I'm open to changing the behavior of these to "cannot rename".

jwnimmer-tri (commented):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 11 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`geometry/geometry_state.cc` line 799 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGZkYmAtWhYVBi03yi:-NbGZkYmAtWhYVBi03yj:bzg0khu) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/geometry_state.cc#L799)):*
> ```cpp
>   });
>   std::string old_name(GetName(frame_id));
>   internal::InternalFrame old_frame = frames_[frame_id];
> ```

BTW Copying the whole frame into `old_frame` seems like somewhat of a waste?  The only part we need is the `source_id`.
___
*[`geometry/internal_frame.h` line 145 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGJnB-9nrp1z8rfZSY:-NbGJnB-9nrp1z8rfZSZ:bc20v0l) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/internal_frame.h#L145)):*
> ```objc
> 
>   /* Changes the name of the frame to `new_name`. */
>   void Rename(std::string_view new_name) {
> ```

BTW This function could be named `void set_name(std::string);` if you like.  That would be more consistent with this class being a plain struct-with-private-fields.

Ditto for `internal_geometry`.
___
*[`geometry/scene_graph.h` line 452 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGJ8kHESPRq46MLDvU:-NbGJ8kHESPRq46MLDvV:bcfavih) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/scene_graph.h#L452)):*
> ```objc
> 
>    This method modifies the underlying model and requires a new Context to be
>    allocated. It does not modify the model versions (see @ref
> ```

BTW Does `@ref` still work when it crosses a newline?

Ditto below.
___
*[`geometry/scene_graph.h` line 460 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGIhuIDeEYy_lx08eV:-NbGIhuIDeEYy_lx08eW:b-l02io6) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/scene_graph.h#L460)):*
> ```objc
>                           or b) there is already a frame named `new_name` from
>                           the same source. */
>   void Rename(FrameId frame_id, std::string_view new_name);
> ```

We need to spell both of these out (RenameFrame and RenameGeometry), not rely on overloading.  That spelling would need to uniformly carry through to GeometryState, etc.

In SceneGraph, when a function (like RemoveRole) is overloaded on FrameId or GeometryId, when called on a FrameId it applies to all of the geometries under that frame.  Here, we are renaming only the frame itself, not the geometries.



_Suggestion:_
```objc
RenameFrame
```
___
*[`geometry/test/geometry_state_test.cc` line 950 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbG_6p3AMY1_Wuaw2mK:-NbG_6p4C2oTSxr72sZ0:bui27q9) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/test/geometry_state_test.cc#L950)):*
> ```cpp
>   EXPECT_EQ(geometry_state_.GetName(geometries_[0]), "something");
> }
> ```

nit For both a frame and a geometry, I'd like to see a unit test case to prove we can rename a thing to itself, e.g., `Rename(frames_[0], "f0");`.  Self-assignment errors are a common kind of implementation defect in code like this.
___
*[`multibody/plant/multibody_plant.h` line 1158 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGCdpy4Xp9G0WgypYz:-NbGCdpy4Xp9G0WgypZ-:b-600nlb) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/multibody_plant.h#L1158)):*
> ```objc
>   }
> 
>   /// Renames an existing model instance.
> ```

nit This needs a documentation disclaimer that it's only allowed pre-Finalize.
___
*[`multibody/plant/multibody_plant.cc` line 978 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbG_rrD6AVHUeq1k6Lc:-NbG_rrD6AVHUeq1k6Ld:b-z1mery) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/multibody_plant.cc#L978)):*
> ```cpp
>   if (geometry_source_is_registered()) {
>     // Re-spam frame and geometry names, using GetScopedName().
>     auto& inspector = scene_graph_->model_inspector();
> ```

nit The `scene_graph_` is nullptr post-Finalize.  We either need to handle the possible null here, or throw-if-finalized atop this function.

I'm not a fan of deferring to MbT to throw.
___
*[`multibody/plant/multibody_plant.cc` line 979 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGa91R-Yjm0MgrXzsB:-NbGa91S1JvRutVN9kMp:bej61bc) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/multibody_plant.cc#L979)):*
> ```cpp
>     // Re-spam frame and geometry names, using GetScopedName().
>     auto& inspector = scene_graph_->model_inspector();
>     auto source_frames = inspector.FramesForSource(*source_id_);
> ```

nit The `auto source_frames = ...` here is too unclear about what the type is.

One sufficient solution would be to add "id" to the variable name, e.g., `frame_ids` or `source_frame_ids`.

Alternatively, lose the local entirely:

```c++
for (auto& frame_id : inspector.FramesForSource(*source_id_)) {
```
___
*[`multibody/plant/multibody_plant.cc` line 983 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGeHBX8zhZC1nj4IQ8:-NbGeHBX8zhZC1nj4IQ9:bfrwaat) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/multibody_plant.cc#L983)):*
> ```cpp
>     auto do_rename = [&](auto id) {
>       std::string short_name(
>           ScopedName::Parse(inspector.GetName(id)).get_element());
> ```

If we stick with this approach, then we need some kind of comment cross-check with e.g. RegisterVisualGeometry as to how the names are conventionally formed.

I think this also might fail if the geometry name itself has double-colons in it?

What I recommend instead is a pure string-prefix replace.  All of the names we find here should start with `{old_model_instance_name}::`.  We can string-cut-and-paste that to `{new_model_instance_name}::`, failing-fast in case the old name was not there anymore.

That seems easier to prove correct (as compared to scoped name parsing), and doesn't require any coupling with how RegisterVisualGeometry works.  If anything ever starts to diverge, we would fail-fast because the prefix we expected was not there.
___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbGc-Mb4iaQQ-wLKKpJ:b-yk2hqc) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

I'm open to changing the behavior of these to "cannot rename".
</blockquote></details>

I'm inclined to allow it, at least for the `default_model_instance()`.

Maybe you want to ping Sherm or Ale about whether the `world_model_instance()` should allow renaming.
___
*[`multibody/tree/multibody_tree-inl.h` line 440 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGbUS885VocNUuW-ba:-NbGbUS885VocNUuW-bb:b-e1xsoy) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/tree/multibody_tree-inl.h#L440)):*
> ```objc
>   instance_name_to_index_.erase(old_name);
>   this->SetElementIndex(name, model_instance, &instance_name_to_index_);
>   instance_index_to_name_[model_instance] = {name, false};
> ```

I am having a really tough time figuring out what this `false` is doing here.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 11 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`geometry/scene_graph.h` line 452 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGJ8kHESPRq46MLDvU:-NbGkfXf4R0e6SykcMi3:bbngqia) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/scene_graph.h#L452)):*

BTW Does `@ref` still work when it crosses a newline?

Ditto below.
</blockquote></details>

I'll double-check, but I thought I recently convinced myself that that do. There are definitely tons of other line-break hazards in doxygen, but this one I thought works. Checking.
___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbGlNhd4ypPC4Q_MRf8:b1jf788) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

I'm inclined to allow it, at least for the `default_model_instance()`.

Maybe you want to ping Sherm or Ale about whether the `world_model_instance()` should allow renaming.
</blockquote></details>

Ping @sherm1 for opinions on renaming the magic world and default model instances.
___
*[`multibody/tree/multibody_tree-inl.h` line 440 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGbUS885VocNUuW-ba:-NbGkDj58bYJ3raHK5-A:bf6wczm) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/tree/multibody_tree-inl.h#L440)):*

I am having a really tough time figuring out what this `false` is doing here.
</blockquote></details>

Ha. That is probably cruft from the Freeze/Frozen name semantics reverse engineering epoch. Why it compiles is an even more interesting question.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`geometry/scene_graph.h` line 452 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGJ8kHESPRq46MLDvU:-NbLLj9UADHalpIXgpzX:b-afbfm1) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/scene_graph.h#L452)):*

I'll double-check, but I thought I recently convinced myself that that do. There are definitely tons of other line-break hazards in doxygen, but this one I thought works. Checking.
</blockquote></details>

This case (`@ref` across a line break) does work, as seen in local testing.
___
*[`geometry/scene_graph.h` line 460 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGIhuIDeEYy_lx08eV:-NbHMEaRCxZvZVyZKyWc:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/geometry/scene_graph.h#L460)):*

We need to spell both of these out (RenameFrame and RenameGeometry), not rely on overloading.  That spelling would need to uniformly carry through to GeometryState, etc.

In SceneGraph, when a function (like RemoveRole) is overloaded on FrameId or GeometryId, when called on a FrameId it applies to all of the geometries under that frame.  Here, we are renaming only the frame itself, not the geometries.
</blockquote></details>

Done.
___
*[`multibody/plant/multibody_plant.cc` line 983 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGeHBX8zhZC1nj4IQ8:-NbL_5C74iobyT8hjJLo:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/multibody_plant.cc#L983)):*

If we stick with this approach, then we need some kind of comment cross-check with e.g. RegisterVisualGeometry as to how the names are conventionally formed.

I think this also might fail if the geometry name itself has double-colons in it?

What I recommend instead is a pure string-prefix replace.  All of the names we find here should start with `{old_model_instance_name}::`.  We can string-cut-and-paste that to `{new_model_instance_name}::`, failing-fast in case the old name was not there anymore.

That seems easier to prove correct (as compared to scoped name parsing), and doesn't require any coupling with how RegisterVisualGeometry works.  If anything ever starts to diverge, we would fail-fast because the prefix we expected was not there.
</blockquote></details>

Done.
___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbLZoA99MWicanmWk9R:b-8d3g6x) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

Ping @sherm1 for opinions on renaming the magic world and default model instances.
</blockquote></details>

Hearing no burning interest in the question, I will just convert the comment to a TODO.
___
*[`multibody/tree/multibody_tree-inl.h` line 440 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbGbUS885VocNUuW-ba:-NbLZf2i0o7bfFA7-nPK:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/tree/multibody_tree-inl.h#L440)):*

Ha. That is probably cruft from the Freeze/Frozen name semantics reverse engineering epoch. Why it compiles is an even more interesting question.
</blockquote></details>

Done.

rpoyner-tri (commented):
Comments addressed. PTAL

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbLbxeD3pVGrxRNWfBQ:byq25nx)* status: 3 unresolved discussions, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

jwnimmer-tri (commented):
> Reviewed 16 of 16 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 1 unresolved discussion, LGTM missing from assignee jwnimmer-tri(platform), needs at least two assigned reviewers

___
*[`geometry/test/geometry_state_test.cc` line 933 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbLgSa20xPnrr8q2JXa:-NbLgSa3DFpTbhqtKMQB:bl4nzfg) ([raw file](https://github.com/RobotLocomotion/drake/blob/ac5907062a708c2487a7fa149250b75931d7236c/geometry/test/geometry_state_test.cc#L933)):*
> ```cpp
>       ".*already existing.*");
> 
>   // Renaming a frame to it's existing name does not cause errors.
> ```

typo

Ditto below.



_Suggestion:_
```cpp
its
```
___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbLhFZc8dukOEJlYnea:b-tqeaxm) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

Hearing no burning interest in the question, I will just convert the comment to a TODO.
</blockquote></details>

The TODO doesn't make any sense, given the current "allow it" behavior.

There is no good way to change our mind later and decide to throw instead.  (It would be a documentation-only deprecation transition, which we should try to avoid.)

If we think that we might _ever_ want to throw in this case, then we should conservatively implement it to throw now (with API documentation to match).

If we think "allow it" is fine indefinitely, then we can lose the TODO.

jwnimmer-tri (commented):
:lgtm: feature.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbLhzsP5vUdjeHWEd4H:b-n37teh)* status: 1 unresolved discussion, needs at least two assigned reviewers

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 2 unresolved discussions, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbLjBAb3zJ_9m131dcL:b-pp2sp1) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

The TODO doesn't make any sense, given the current "allow it" behavior.

There is no good way to change our mind later and decide to throw instead.  (It would be a documentation-only deprecation transition, which we should try to avoid.)

If we think that we might _ever_ want to throw in this case, then we should conservatively implement it to throw now (with API documentation to match).

If we think "allow it" is fine indefinitely, then we can lose the TODO.
</blockquote></details>

Sorry, I missed the ping on this one. Personally I can't think of any reason we care about the names of world and default model instances as long as the ModelInstanceIndexes remain 0 and 1. However, @amcastro-tri may have assigned some significance to the names somewhere -- worth checking with him. If he doesn't care either, I'd say allow the renames.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`multibody/plant/test/multibody_plant_test.cc` line 4636 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbFZbrH0zjdmDrE6bO0:-NbLiWk9CIWsQY3uFfXU:b-lvhmu5) ([raw file](https://github.com/RobotLocomotion/drake/blob/2428b2ae736f3e01bedff16f02eaba158c6df477/multibody/plant/test/multibody_plant_test.cc#L4636)):*

Sorry, I missed the ping on this one. Personally I can't think of any reason we care about the names of world and default model instances as long as the ModelInstanceIndexes remain 0 and 1. However, @amcastro-tri may have assigned some significance to the names somewhere -- worth checking with him. If he doesn't care either, I'd say allow the renames.
</blockquote></details>

I'm leaning toward "allow it" indefinitely. Doing those renames might have unpleasant effects, but it would be the result of an affirmative step in application code, and (with these rename apis) the names are  pretty much at the discretion of the application anyway.

jwnimmer-tri (commented):
@rpoyner-tri I'm on-call for platform.  I'll let you elect how you want to assign someone.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbLk8Bo28pe16SBWNPa:bd5w78j)* status: 1 unresolved discussion, needs at least two assigned reviewers

___
*[`geometry/test/geometry_state_test.cc` line 951 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbLk1-B1HQUi5fbDH0p:-NbLk1-C5CYMaU-OozmB:bb7phvt) ([raw file](https://github.com/RobotLocomotion/drake/blob/a7b3bb2cf08c2fcba99ee2cb9ffdf82913fc61fd/geometry/test/geometry_state_test.cc#L951)):*
> ```cpp
>       ".*Cannot rename.*invalid geometry id:.*");
> 
>   // Renaming a geometry to it's existing name does not cause errors.
> ```

typo



_Suggestion:_
```cpp
its
```

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: needs at least two assigned reviewers

rpoyner-tri (commented):
+@EricCousineau-TRI for Thursday-schedule platform review; Nimmer is on-call until then and has already read it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbLkWlq7jyvKX6s2fEp:b-pncuw5)* status: LGTM missing from assignee EricCousineau-TRI(platform)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 14 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`geometry/test/geometry_state_test.cc` line 939 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMspOe3hSYH-Coqw1u:-NbN3AgL9iymEoptNqOw:b-24t8wh) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/geometry/test/geometry_state_test.cc#L939)):*
> ... adding the same comment here: scoping the name to confirm that it's getting a copy.

Or alternatively, we could change the `set_name()` functions to accept a `std::string` and move from it, instead of a `std::string_view`.  I don't see any reason to use `string_view` for those two setters, since the first thing they do is turn around and construct a string copy.  With `std::string` as the argument type, the unit testing wouldn't need to be so careful.
___
*[`bindings/pydrake/geometry/geometry_py_scene_graph.cc` line 194 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMuaxu7hwZaMLjO_sw:-NbN48jY9sLINispjU7U:bvp2ehk) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/bindings/pydrake/geometry/geometry_py_scene_graph.cc#L194)):*

BTW  Call sites like:

```py
scene_graph.RenameFrame(frame_id=my_id)
```

look painfully redundant as ("frame" X 2) compared to:

```py
scene_graph.RenameFrame(id=my_id)
```

It seems to me that if we name the *function* `Frame`, we're not obliged to name the parameter "frame" as well. (Same for geometry.)

We *do* have overloads where the function name is the same, which makes the distinction in parameter names more important.  Thoughts?
</blockquote></details>

If the user cares about brevity at the call site, they can say `scene_graph.RenameFrame(my_id)`.  For my taste, if they are specifically trying to use a named kwarg, then I don't think its a problem for it to have some amount of redundant detail.  Using `frame_id` throughout the whole API brings a level of consistency.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 14 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`geometry/test/geometry_state_test.cc` line 939 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMspOe3hSYH-Coqw1u:-NbNJqBcAi4Q5chXR83P:bsly0p) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/geometry/test/geometry_state_test.cc#L939)):*

> ... adding the same comment here: scoping the name to confirm that it's getting a copy.

Or alternatively, we could change the `set_name()` functions to accept a `std::string` and move from it, instead of a `std::string_view`.  I don't see any reason to use `string_view` for those two setters, since the first thing they do is turn around and construct a string copy.  With `std::string` as the argument type, the unit testing wouldn't need to be so careful.
</blockquote></details>

Definitely thumbs up. When I first saw that code, my first thought was, why not string by value? But I convinced myself that the string view -> string wasn't doing any more or less copying so, meh. But I think the code would be more obvious with a string value parameter.
___
*[`bindings/pydrake/geometry/geometry_py_scene_graph.cc` line 194 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMuaxu7hwZaMLjO_sw:-NbNK1ZJC5J0CGB4jIRT:b-d9kyq8) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/bindings/pydrake/geometry/geometry_py_scene_graph.cc#L194)):*

If the user cares about brevity at the call site, they can say `scene_graph.RenameFrame(my_id)`.  For my taste, if they are specifically trying to use a named kwarg, then I don't think its a problem for it to have some amount of redundant detail.  Using `frame_id` throughout the whole API brings a level of consistency.
</blockquote></details>

Consistent vs concise is an interesting qualitative judgment. (Hence the BTW.) I can see the value of favoring consistency. I'm happy to defer to people who feel more strongly than I.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 13 unresolved discussions, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`examples/manipulation_station/manipulation_station.cc` line 236 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMs5GI7HAn722EW8Q9:-NbPrvTR5vcq97wvm5aT:bkf5z9k) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/examples/manipulation_station/manipulation_station.cc#L236)):*

btw: This comment struck me as weird. This merely *moved* the risk of name collisions from parsing to renaming.
</blockquote></details>

Um, sure. What would you rather we say here?
___
*[`bindings/pydrake/geometry/geometry_py_scene_graph.cc` line 194 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMuaxu7hwZaMLjO_sw:-NbPtLVw6fCm7GRlQ0L2:bve8n22) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/bindings/pydrake/geometry/geometry_py_scene_graph.cc#L194)):*

Consistent vs concise is an interesting qualitative judgment. (Hence the BTW.) I can see the value of favoring consistency. I'm happy to defer to people who feel more strongly than I.
</blockquote></details>

In my view, if you are using param names for these call sites, concision is not your primary value.
___
*[`bindings/pydrake/multibody/plant_py.cc` line 236 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMvGTyE2VO2pyK0kF0:-NbPtZSq7q2qvQ_fYxgf:b-hj211q) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/bindings/pydrake/multibody/plant_py.cc#L236)):*

BTW SceneGraph calls this "new_name", do we want to rename this to be consistent?
</blockquote></details>

I played with "new_name" for parameters of these methods, but now I'm less convinced. I may just fall back to "name" everywhere.

rpoyner-tri (commented):
`r5` has comments so far addressed. The geometry_state code tries to better adhere to internal conventions, e.g. `GetMutableGeometry()`. Also, I tried to avoid casual use in implementations of STL `operator[]`, preferring `at()` instead where possible.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbQa_Lb6HPc-7XKHstt:b-myc0ci)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`multibody/tree/multibody_tree-inl.h` line 428 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMoaKt12bgn-E6y-yP:-NbPuJTD7kbFvDNWcNVo:b4jvdv6) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/multibody/tree/multibody_tree-inl.h#L428)):*

BTW  If you look at `AddRigidBody()`, there is an exception of the body doesn't have a unique name. It uses the language, "Body names must be unique within a given model." In that case, the "model" refers to a "model instance". Here, "model" refers to the whole tree. We've definitely got a nomenclature issue. (Although, the duplicate gravity error refers to the tree as the "model".)

The reason I came across this particular factoid, is because when I read this error message, I worried that this message is the feedback a user gets if they pass a duplicate name from an `MbP` API. Does the text sufficiently connect back to the user act?

I'm not sure there's an particular act that can/should be taken in *this* PR. Presumably, the messages have not been detrimental to understanding.


</blockquote></details>

The "this model" language is consistent with the existing documentation of plant.AddModelInstance; I'm inclined to leave well enough alone for now.

jwnimmer-tri (commented):
> Reviewed 16 of 16 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`geometry/scene_graph.cc` line 185 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbQec_34amqEqL41p5M:-NbQec_42j7HACP8xZr2:b-fmtuke) ([raw file](https://github.com/RobotLocomotion/drake/blob/08b9183bc280ac39c034e51a9e2af7fb43721dd4/geometry/scene_graph.cc#L185)):*
> ```cpp
> template <typename T>
> void SceneGraph<T>::RenameFrame(FrameId frame_id, std::string name) {
>   return model_.RenameFrame(frame_id, name);
> ```

nit It's useless to accept name by-value here when we're not moving-from it.  We might as well use `const string&`?

We wouldn't want to move-from it anyway, because GeometryState takes the name by-const-reference.

Ditto below.

I think maybe we should nix `std::string` by-value from all of the changes in this PR.  The extra bookkeeping doesn't seem worth it, for a parsing-time operation.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: 1 unresolved discussion, LGTM missing from assignee SeanCurtis-TRI(platform)

___
*[`geometry/scene_graph.cc` line 185 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbQec_34amqEqL41p5M:-NbQgSHmA6Bg7G8RYT4o:b-vng031) ([raw file](https://github.com/RobotLocomotion/drake/blob/08b9183bc280ac39c034e51a9e2af7fb43721dd4/geometry/scene_graph.cc#L185)):*

nit It's useless to accept name by-value here when we're not moving-from it.  We might as well use `const string&`?

We wouldn't want to move-from it anyway, because GeometryState takes the name by-const-reference.

Ditto below.

I think maybe we should nix `std::string` by-value from all of the changes in this PR.  The extra bookkeeping doesn't seem worth it, for a parsing-time operation.
</blockquote></details>

Bah, I thought I changed those to `const&` -- working.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939)* status: LGTM missing from assignee SeanCurtis-TRI(platform)

SeanCurtis-TRI (commented):
Out of curiosity, what's the motivation for preferring `at()` over `operator[]`? The potential for out-of-bounds feedback?

:LGTM:

> Reviewed 12 of 16 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbRWmit6GjVDLVhcQFk:bvca330)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform)

___
*[`examples/manipulation_station/manipulation_station.cc` line 236 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMs5GI7HAn722EW8Q9:-NbRVMJuEHsuHZaWtsf7:bf6k75g) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/examples/manipulation_station/manipulation_station.cc#L236)):*
That is what I inferred the intended meaning to be. I just felt that the phrasing required too much generosity from the reader.  :)

I like something like this:

> We want to add the model in `model_url` and have it named `model_name`. Changing the parser to handle auto renaming guarantees no model name collisions during parsing. We can then safely rename the model to `model_name`, because we know by construction that `model_name` has not already been used in the plant. (If we're wrong, the renaming will throw.)
___
*[`multibody/tree/multibody_tree-inl.h` line 428 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMoaKt12bgn-E6y-yP:-NbRYtlT4IkTSINmGtR9:bw9vhov) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/multibody/tree/multibody_tree-inl.h#L428)):*

The "this model" language is consistent with the existing documentation of plant.AddModelInstance; I'm inclined to leave well enough alone for now.
</blockquote></details>

Meh. It's consistent with the inconsistency. I.e., there are plenty of examples in this file where "this model" has the same meaning as used here. So, this certainly doesn't introduce any problems. It merely cast a light on it.
___
*[`multibody/tree/test/model_instance_test.cc` line 174 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMuQO6D05AnlvO06HO:-NbRZE559ERA8iVxrMfu:bq4ea0q) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/multibody/tree/test/model_instance_test.cc#L174)):*

BTW This implicitly doesn't *throw*, but I'm not sure that the test shows it's a "no op". Perhaps simply to say, "// to same name is not an error."
</blockquote></details>

Alternatively, we have a formal way of communicating that:

`EXPECT_NO_THROW(tree.RenameModelInstance(model0, "after"));
___
*[`bindings/pydrake/geometry/geometry_py_scene_graph.cc` line 194 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-NbMuaxu7hwZaMLjO_sw:-NbRZNyn8e7L6VZOXZhJ:bz9x5nu) ([raw file](https://github.com/RobotLocomotion/drake/blob/7d0164acc4d3b63ddae72d6df85a31e620827b7c/bindings/pydrake/geometry/geometry_py_scene_graph.cc#L194)):*

In my view, if you are using param names for these call sites, concision is not your primary value.
</blockquote></details>

(I'm marking myself satisfied because this is a purely subjective, philosophical response. I wouldn't be me if I could resist responding. But, in no way, does this PR need to change based on this.)

I disagree. 

Concision doesn't imply minimum characters. Concision is terseness subject to being *complete*. I may use keyword arguments to be clear about the semantics of the parameters I'm passing. Not because I like lots of characters. That means I would favor keyword arguments that are, themselves, concise. When the parameter names are not concise, then, yes, we are inflicting a cost on users who prefer clarity -- they must take needless verbosity on board.

jwnimmer-tri (commented):
The `op[]` will implicitly create a key-value pair when the key did not already exist.  The `at()` will throw if the key did not already exist.  If we expect the key to already exist, we might prefer `.at()` as better sanity check than `op[]`.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbRhebR5cmD_CznZGJB:b-5andg)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform)

SeanCurtis-TRI (commented):
D'oh! I was thinking `vector` and not `map`. Yes, of course.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19939#-:-NbSakTL3OYCguEtnLNt:brkexvq)* status: :shipit: complete! all discussions resolved, LGTM from assignees jwnimmer-tri(platform),SeanCurtis-TRI(platform)

PR #19862, opened 2023-07-27
[workspace] Uniformly use a patches directory

jwnimmer-tri:
This helps disambiguate Drake's build system commands (in the parent directory) from our set of upstream changes (in the `patches` subdirectory), and also makes it easier for users who want to do the patching themselves (using their own software supply chain infrastructure).

---
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19862)

PR #19950, opened 2023-08-07
[workspace] Use conventional filename for SCS patching

jwnimmer-tri:
I missed this one in #19862, due to the weird filename.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19950)


jwnimmer-tri:
+@sherm1 for both reviews per schedule, please.

sherm1 (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19950)* status: :shipit: complete! all discussions resolved, LGTM from assignee sherm1(platform)

___
*[`tools/workspace/scs_internal/patches/include_paths.patch` line 7 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19950#-NbFXEO448RfZssBC8js:-NbFazjB5Fn3i1z8KzZE:b-x6k9jy) ([raw file](https://github.com/RobotLocomotion/drake/blob/940a4956262f36481c53a1aa4d84d5fd6401c1f2/tools/workspace/scs_internal/patches/include_paths.patch#L7)):*

As I understand it, `---` and `+++` lines specify which filename to edit when applying the patch, so are still necessary.
</blockquote></details>

Ah, I see -- the actual patch is below. I glazed over that. Withdrawn!

Issue #19709, opened 2023-06-30
Use hydro without manual annotation of every single model file


amcastro-tri:
The problem is that we have a lot of models decorated with point spheres (and no hydro specification at all). We know these models would simulate better with hydro.
The ycb objects, for instance, would be a great place to start?  currently users have a mix of boxes + tiny sphere collision geometry. 

While we could default geometries to always have a hydroelastic representation, the tiny spheres people add to their models are incompatible with the hydroelastic model (near to or zero volume). We could ask users to remove those spheres, but this only puts the burden on them. Users would probably like to switch back and forth between models (hydro vs. simple geometry + tiny spheres).

Here are some highlights from a [long conversation in Slack](https://tri-internal.slack.com/archives/C2BMF3K28/p1680144366502789) that I believe capture the main points and challenges discussed:
- Hydro is awesome, everyone should use it a lot more. How do we get them to do that?
- Adding new Drake-specific XML boilerplate (the hydro representation) to every single SDFormat file in existence is a non-starter. The XML syntax is great for users to tune the hydro representation, but it shouldn't be required.
- Instead, we should add a new MbP ContactModel enum value. The current kHydroelasticWithFallback says that a mixed collision pair (hydro-rep + no-hydro-rep) falls back to point contact. The new enum option should prefer hydro-hydro in the mixed case, by inferring a hydro representation. Eventually, the new enum value should become our default.
- The Dynamics team needs to decide what a good inferred representation should look like.
- After all of that, there is still an edge case to consider: 
  - Models that in the past used primitive box collision with point contact often have tiny little spheres at the box corners. Currently, their extremely small volume (~attoliter) causes an exception when hydrofied. One way or another, they should be rounded down to zero (either don't participate in hydro contact, or participate but are a no-op).
- "This is the scale of things I care about." 
  - With a default value with length unit, the issue of tiny spheres seems to be easily resolved. If for hydro I care about things on the scale of 1e-3m, then a sphere with 1e-7m should be ignored when hydrofied unless otherwise noted.

rpoyner-tri:
Had a f2f discussion of this problem with @RussTedrake and @joemasterjohn today. Various techniques were discussed:
* adding code to the parser to add hydro defaults
* adding code into scene graph at some level to add hydro defaults
* various forms of plant-level contact mode hackery
* application-level property hacking after the fact

Some current possible obstacles were discussed:
[...]

PR #19956, opened 2023-08-07
wip: first sketch of hydro-inferred mode and plumbing

rpoyner-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)

### DIFF
--- a/bindings/pydrake/multibody/plant_py.cc
+++ b/bindings/pydrake/multibody/plant_py.cc
@@ -1451,6 +1451,8 @@ PYBIND11_MODULE(plant, m) {
         .value("kPoint", Class::kPoint, cls_doc.kPoint.doc)
         .value("kHydroelasticWithFallback", Class::kHydroelasticWithFallback,
             cls_doc.kHydroelasticWithFallback.doc)
+        .value("kHydroelasticInferred", Class::kHydroelasticInferred,
+            cls_doc.kHydroelasticInferred.doc)
         // Legacy alias. TODO(jwnimmer-tri) Deprecate this constant.
         .value("kHydroelasticsOnly", Class::kHydroelasticsOnly,
             cls_doc.kHydroelasticsOnly.doc)
--- a/bindings/pydrake/multibody/test/plant_test.py
+++ b/bindings/pydrake/multibody/test/plant_test.py
@@ -2614,6 +2614,7 @@ class TestPlant(unittest.TestCase):
             ContactModel.kHydroelastic,
             ContactModel.kPoint,
             ContactModel.kHydroelasticWithFallback,
+            ContactModel.kHydroelasticInferred,
             # Also test the legacy aliases.
             ContactModel.kHydroelasticsOnly,
             ContactModel.kPointContactOnly,
--- a/examples/hydroelastic/ball_plate/ball_plate_run_dynamics.cc
+++ b/examples/hydroelastic/ball_plate/ball_plate_run_dynamics.cc
@@ -18,7 +18,7 @@ DEFINE_double(simulation_time, 0.4,
 // See MultibodyPlantConfig for the valid strings of contact_model.
 DEFINE_string(contact_model, "hydroelastic",
               "Contact model. Options are: 'point', 'hydroelastic', "
-              "'hydroelastic_with_fallback'.");
+              "'hydroelastic_with_fallback', 'hydroelastic_inferred'.");
 // See MultibodyPlantConfig for the valid strings of contact surface
 // representation.
 DEFINE_string(contact_surface_representation, "polygon",
--- a/examples/hydroelastic/ball_plate/floor.sdf
+++ b/examples/hydroelastic/ball_plate/floor.sdf
@@ -39,16 +39,6 @@
             <size>3.0e-1 3.0e-1 5.0e-2</size>
           </box>
         </geometry>
-        <drake:proximity_properties>
-          <drake:compliant_hydroelastic/>
-          <drake:hydroelastic_modulus> 3.0e4 </drake:hydroelastic_modulus>
-          <drake:mesh_resolution_hint> 1.0 </drake:mesh_resolution_hint>
-          <drake:mu_dynamic> 0.3 </drake:mu_dynamic>
-          <drake:mu_static> 0.3 </drake:mu_static>
-          <drake:hunt_crossley_dissipation>
-            3.0
-          </drake:hunt_crossley_dissipation>
-        </drake:proximity_properties>
       </collision>
     </link>
   </model>
--- a/examples/hydroelastic/ball_plate/plate_8in.sdf
+++ b/examples/hydroelastic/ball_plate/plate_8in.sdf
@@ -40,13 +40,6 @@
             <scale>1.0 1.0 1.0</scale>
           </mesh>
         </geometry>
-        <drake:proximity_properties>
-          <drake:rigid_hydroelastic/>
-          <!--
-          We do not set drake:mesh_resolution_hint here because it has no
-          effect on meshes imported from obj file.
-          -->
-        </drake:proximity_properties>
       </collision>
     </link>
     <frame name="origin">
--- a/examples/multibody/rolling_sphere/rolling_sphere_run_dynamics.cc
+++ b/examples/multibody/rolling_sphere/rolling_sphere_run_dynamics.cc
@@ -24,7 +24,8 @@ DEFINE_double(simulation_time, 2.0,
 
 // Contact model parameters.
 DEFINE_string(contact_model, "point",
-              "Contact model. Options are: 'point', 'hydroelastic', 'hybrid'.");
+              "Contact model. Options are: 'point', 'hydroelastic', 'hybrid',"
+              " inferred'.");
 DEFINE_string(hydro_rep, "tri",
               "Contact-surface representation for hydroelastics. "
               "Options are: 'tri' for triangles, 'poly' for polygons. "
@@ -160,6 +161,9 @@ int do_main() {
     plant.set_contact_model(ContactModel::kHydroelasticWithFallback);
     plant.Finalize();
     plant.set_penetration_allowance(0.001);
+  } else if (FLAGS_contact_model == "inferred") {
+    plant.set_contact_model(ContactModel::kHydroelasticInferred);
+    plant.Finalize();
   } else {
     throw std::runtime_error("Invalid contact model '" + FLAGS_contact_model +
                              "'.");
--- a/geometry/geometry_state.cc
+++ b/geometry/geometry_state.cc
@@ -944,7 +944,7 @@ void GeometryState<T>::ChangeShape(SourceId source_id, GeometryId geometry_id,
     // unchecked version because we just need the engine mechanism; no
     // further GeometryState checking.
     RemoveFromProximityEngineUnchecked(*geometry);
-    AddToProximityEngineUnchecked(*geometry);
+    AddToProximityEngineUnchecked(geometry);
   }
   if (geometry->has_illustration_role()) {
     // Illustration has no "engine"; it's just the InternalGeometry. All
@@ -1016,6 +1016,7 @@ void GeometryState<T>::AssignRole(SourceId source_id, GeometryId geometry_id,
   geometry_version_.modify_proximity();
   switch (assign) {
     case RoleAssign::kNew: {
+      ProximityProperties* props = &properties;
       geometry.SetRole(std::move(properties));
       if (geometry.is_deformable()) {
         DRAKE_DEMAND(geometry.reference_mesh() != nullptr);
@@ -1027,11 +1028,11 @@ void GeometryState<T>::AssignRole(SourceId source_id, GeometryId geometry_id,
             convert_to_double(kinematics_data_.X_WGs.at(geometry_id));
         geometry_engine_->AddDynamicGeometry(geometry.shape(), X_WG,
                                              geometry_id,
-                                             *geometry.proximity_properties());
+                                             props);
       } else {
         geometry_engine_->AddAnchoredGeometry(geometry.shape(), geometry.X_FG(),
                                               geometry_id,
-                                              *geometry.proximity_properties());
+                                              props);
       }
       // The set of geometries G such that I need to introduce filtered pairs
       // (geometry_id, gᵢ) ∀ gᵢ ∈ G. Generally, it consists of those proximity
@@ -1057,7 +1058,7 @@ void GeometryState<T>::AssignRole(SourceId source_id, GeometryId geometry_id,
     case RoleAssign::kReplace:
       // Give the engine a chance to compare properties before and after.
       geometry_engine_->UpdateRepresentationForNewProperties(geometry,
-                                                             properties);
+                                                             &properties);
       geometry.SetRole(std::move(properties));
       break;
     default:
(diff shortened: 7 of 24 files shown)


rpoyner-tri (commented):
FYI @DamrongGuoy @SeanCurtis-TRI @jwnimmer-tri 

Note that this so far is just the earliest possible sketch -- no real commitments to form or spelling yet, just trying to see what can work at all. In particular the mountain of `{set,get}_hydro_inferred` methods is not awesome, but it gets me the bare minimum of initialization-time communication along the plant<=>scene-graph<=>geometry-state<=>proximity-engine<=>hydro-geometries chain.

Note there are no docs, tests, or examples yet. I may look into some minimal python binding support, so perhaps we can experiment with models/sims/scenarios via notebooks.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-NbGcsjx3VWGWQ8jWWNI:b-ih7p6s)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", missing label for release notes (waiting on @rpoyner-tri)

jwnimmer-tri:
Looks OK to me.  At first I was surprised to see the SceneGraphInspector offer a getter for this, but I suppose it makes sense because its a property of the GeometryState, not the SceneGraph.  The setter on SceneGraph should probably make it clear that it will not affect already-created contexts.

rpoyner-tri (commented):
I suppose we could have context-accepting setters and getters...but I'm happy to wait until there is a reason to want them, and we have a design we like for that communication to begin with.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-NbGj6GL3N5e57LGeKya:bxsfbza)* status: needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-NbHrxgn4xrL_dcG8v_Y:b3qcz6s) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*
> ```cpp
> template <typename T>
> void SceneGraph<T>::set_hydro_inferred(bool value) {
>     model_.set_hydro_inferred(value);
> ```

When this changes, I think we probably need to re-spin the `MaybeAddGeometry` logic, to adjust the emboded model.

It that change-handling becomes difficult to implement in the MVP, this could move to be a `SceneGraph` constructor argument, with no setter.

RussTedrake (commented):
btw -- :heart:  Thanks for pushing this!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-NbJYGhw8zx3meEry6GY:b-6ax6dx)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

rpoyner-tri (commented):
Motivating issue is #19709, FTR.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-NbKJLaSEWOrUanxVzDq:b84a847)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-Nc-6lIr98AnM8OLEP1C:b8ez9rv) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*

When this changes, I think we probably need to re-spin the `MaybeAddGeometry` logic, to adjust the emboded model.

It that change-handling becomes difficult to implement in the MVP, this could move to be a `SceneGraph` constructor argument, with no setter.
</blockquote></details>

I think the SceneGraph constructor argument is either (a) a non-starter, or (b) a clue that the new feature should not be a ContactModel enum value, but instead something else.

The difficulty is that we create mbps and scenegraphs early, and (often) only set the contact model in the last moment before calling Finalize(). Adding to this, we have any number of convenience workflows for creating mbp/sg pairs, and threading through some extra parameter all of those sounds hardly worthwhile.

rpoyner-tri (commented):
The mods to rolling_sphere have been helpful to give (non-crashing) proof of "life" for kHydroelasticInferred, but do nothing to demonstrate that we can actually operate on naive assets. That work is still ongoing.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-Nc-8-DQ0uaRZhzNN1t4:b-s3wdwe)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-Nc-9Ja5-NUtoKEbU3DC:basvtk0) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*

I think the SceneGraph constructor argument is either (a) a non-starter, or (b) a clue that the new feature should not be a ContactModel enum value, but instead something else.

The difficulty is that we create mbps and scenegraphs early, and (often) only set the contact model in the last moment before calling Finalize(). Adding to this, we have any number of convenience workflows for creating mbp/sg pairs, and threading through some extra parameter all of those sounds hardly worthwhile.
</blockquote></details>

s/before calling Finalize/even after calling Finalize/

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-Nc-B-8dD2ysQgDKtcjF:bji8i82) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*
My point is that **either** we need to carefully implement the "update anything that would be impacted" logic here, **or else** we need to demand that the user choose right away with a constructor argument (and no setter).  Possibly the "right away" caveat would just be for MVP, and we could later on implement the more complicated change tracking.

> The difficulty is that we create mbps and scenegraphs early, and (often) only set the contact model in the last moment before calling Finalize().

Sure, but ... why do people need to do that that way?  Isn't that all just historical cruft?  Are there any important use cases where they cannot decide, up front, how they plan to simulate their world?  Would it not be plausible to say "If you want nice things, you need to call the canonical function `AddMultibodyPlant(const MultibodyPlantConfig& config, systems::DiagramBuilder<double>* builder);` to instantiate your plant."?

In the original days, we expected that users would carefully write a lot of code to handle the setup boilerplate, which is where the kind habit you're talking about comes from.  That is no longer the preferred style, and we don't really care about making that easy.  The workflow we should be laser focused on is the Config / Params / YAML workflow.  Sim setup should be declarative, not imperative.  New users should always be using `Config` structs, never calling those weird legacy factory functions.

For my part, I would be satisfied if _only_ the config workflow got this nice-hydro-fallback-by-default feature.  If we can take a shortcut to make that happen faster (i.e., the constructor arg; no setter), that's my vote.  The setter function can be an add-on, if we later find a need for it.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-Nc-xB2R1AiAjJmPzQCR:b-or0092) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*

My point is that **either** we need to carefully implement the "update anything that would be impacted" logic here, **or else** we need to demand that the user choose right away with a constructor argument (and no setter).  Possibly the "right away" caveat would just be for MVP, and we could later on implement the more complicated change tracking.

> The difficulty is that we create mbps and scenegraphs early, and (often) only set the contact model in the last moment before calling Finalize().

Sure, but ... why do people need to do that that way?  Isn't that all just historical cruft?  Are there any important use cases where they cannot decide, up front, how they plan to simulate their world?  Would it not be plausible to say "If you want nice things, you need to call the canonical function `AddMultibodyPlant(const MultibodyPlantConfig& config, systems::DiagramBuilder<double>* builder);` to instantiate your plant."?

In the original days, we expected that users would carefully write a lot of code to handle the setup boilerplate, which is where the kind habit you're talking about comes from.  That is no longer the preferred style, and we don't really care about making that easy.  The workflow we should be laser focused on is the Config / Params / YAML workflow.  Sim setup should be declarative, not imperative.  New users should always be using `Config` structs, never calling those weird legacy factory functions.

For my part, I would be satisfied if _only_ the config workflow got this nice-hydro-fallback-by-default feature.  If we can take a shortcut to make that happen faster (i.e., the constructor arg; no setter), that's my vote.  The setter function can be an add-on, if we later find a need for it.
</blockquote></details>

However, if you main point was that this might not be a new `contact_mode` option but rather some kind of geometry-only option, that might be true.  I think it would probably still live in `MultibodyPlantConfig` at the top level, but how that gets plumbed down into the API might make more sense to go into SceneGraph directly.

rpoyner-tri (commented):
The mods so far to ball_plate are fairly ridiculous (plate and ball fall down to infinity). I've kinda got the hydro-internal inferred rep going, but I've clearly broken something else (probably some compliance type usage elswhere). \cc @DamrongGuoy  if you have any insight into current brokenness.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-Nc4TCQs6hvjff4XwFty:b83yip5)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
`r4` mostly works against the deliberately dis-annotated ball-plate example. So we can regard it as the "minimum blast radius" estimate of changes. Of course, many things are still wrong with it:

* deliberate `const` flouting to write back discovered proximity properties
* arbitrary hard-coded default properties -- should be configurable, or at least public?
* defective geometry construction workflow as noted in ongoing discussions

Minor points:

* ContactModel (if we go that route) probably wants some helper predicates to help tame some of the splatter of finding and patching specific hard-coded predicates (perhaps `uses_hydroelastic_contact()`, `uses_point_contact()`?).
* obviously the current deliberate hobbling of ball_plate assets should be undone.
* usual caveats about missing doc and tests.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-Nc9JKbwBwV5nDgSNrwk:bc567f3)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
`r5` addresses a major (`const` flouting) problem, and a minor (`ContactModel` predicates) problem. Everything is else is still bad. Probably the workflow problem is what to address next -- may need a more complicated example.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-:-NcDuPNJB2dR0CKcpbpT:b-9eeywm)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-NcOSDcnDUqsSYmp4MIX:btuni9s) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*

However, if you main point was that this might not be a new `contact_mode` option but rather some kind of geometry-only option, that might be true.  I think it would probably still live in `MultibodyPlantConfig` at the top level, but how that gets plumbed down into the API might make more sense to go into SceneGraph directly.
</blockquote></details>

Here is one kind of difficulty with the current sketch. Even if it's rarely done in practice, the expectation of `set_contact_model()` is that we could iteratively set different models on the same diagram/context and get more or less the same sim results for each run with a contact model. As currently hacked, the inferred model **destroys information** (the original, perhaps naive, results of model parsing or api construction), so we can't just switch to some other model and get its characteristic result afterward. 

I suppose we could treat hydro-inferred geometries as a distinct set at the lowest level (hydroelastic_internal's Geometries class), but I wonder if I'm missing a simpler plan.

\cc @SeanCurtis-TRI is you have time to consider hydro geometry metaphysics.

rpoyner-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19956)* status: 1 unresolved discussion, needs platform reviewer assigned, needs at least two assigned reviewers, labeled "do not merge", commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @rpoyner-tri)

___
*[`geometry/scene_graph.cc` line 376 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19956#-NbHrxgl6DY5h7xmIAv3:-NcSyE5A8ECzT95R0nfu:bwtvj59) ([raw file](https://github.com/RobotLocomotion/drake/blob/349426f40e78dee9664f57cb59d824c0e7ed7f96/geometry/scene_graph.cc#L376)):*

Here is one kind of difficulty with the current sketch. Even if it's rarely done in practice, the expectation of `set_contact_model()` is that we could iteratively set different models on the same diagram/context and get more or less the same sim results for each run with a contact model. As currently hacked, the inferred model **destroys information** (the original, perhaps naive, results of model parsing or api construction), so we can't just switch to some other model and get its characteristic result afterward. 

I suppose we could treat hydro-inferred geometries as a distinct set at the lowest level (hydroelastic_internal's Geometries class), but I wonder if I'm missing a simpler plan.

\cc @SeanCurtis-TRI is you have time to consider hydro geometry metaphysics.
</blockquote></details>

My thought today (no code yet), is to just make "hydro inferred" a scene graph parameter, and drop all connection with ContactModel. Then I would have no real qualms about making it constructor-only, and would have no need to modify multibody code at all, except perhaps to pass through the new configuration data.

rpoyner-tri:
Bailing on this approach. New attempt shortly.

rpoyner-tri:
new attempt is up as a draft at #20059 .

PR #19960, opened 2023-08-08
[geometry] Increase range of Meshcat's search for an open port

### LINKED ISSUE #19362: Allow increasing Meshcat kMaxPort
At the moment, we are limited to 100 meshcat ports, and it is not possible to close a port without exiting the script that opened them. This is limiting if I want to run lots of simulations in a script and save the meshcat recording for each of them. This is a quite common scenario for me, and I keep running into this issue.

From https://github.com/RobotLocomotion/drake/pull/15618, the reason for the current limit is docker:
>Yes. I've made it 7099, since that's the number of ports that the docker instances open up for meshcat (but it's all pretty arbitrary, i guess... I've rarely needed more than 2!). I've added a comment in the constructor, which I think is the right place since we may offer passing a port number as a constructor option in the future.

I would suggest adding an option to specify a custom value for `kMaxPort` to allow overriding the docker-motivated limit.

jwnimmer-tri:
Closes #19362.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960)

### DIFF
--- a/geometry/meshcat.cc
+++ b/geometry/meshcat.cc
@@ -595,7 +595,8 @@ class Meshcat::Impl {
       : prefix_("/drake"),
         main_thread_id_(std::this_thread::get_id()),
         params_(params) {
-    DRAKE_THROW_UNLESS(params.port.value_or(7000) >= 1024);
+    DRAKE_THROW_UNLESS(!params.port.has_value() || *params.port == 0 ||
+                       *params.port >= 1024);
     if (!drake::internal::IsNetworkingAllowed("meshcat")) {
       throw std::runtime_error(
           "Meshcat has been disabled via the DRAKE_ALLOW_NETWORK environment "
@@ -1726,9 +1727,6 @@ class Meshcat::Impl {
     // our code from potential implementation changes to uWebSockets.
     const std::string bind_host = (host == "*") ? "" : host;
 
-    int port = desired_port ? *desired_port : 7000;
-    const int kMaxPort = desired_port ? *desired_port : 7099;
-
     uWS::App::WebSocketBehavior<PerSocketData> behavior;
     // Set maxBackpressure = 0 so that uWS does *not* drop any messages due to
     // back pressure.
@@ -1759,20 +1757,30 @@ class Meshcat::Impl {
             .ws<PerSocketData>("/*", std::move(behavior));
     app_ = &app;
 
-    do {
-      app.listen(
-          bind_host, port, LIBUS_LISTEN_EXCLUSIVE_PORT,
-          [this](us_listen_socket_t* socket) {
-            DRAKE_DEMAND(IsThread(websocket_thread_id_));
-            if (socket) {
-              listen_socket_ = socket;
-            }
-          });
-    } while (listen_socket_ == nullptr && port++ < kMaxPort);
+    // Search for an open port.
+    int chosen_port{};
+    const int search_start = desired_port.value_or(7000);
+    const int search_end = desired_port.value_or(8000);
+    for (int port = search_start; port <= search_end; ++port) {
+      // N.B. Using `port == 0` requests an ephemeral port.
+      // https://github.com/uNetworking/uSockets/pull/136.
+      app.listen(bind_host, port, LIBUS_LISTEN_EXCLUSIVE_PORT,
+                 [this](us_listen_socket_t* socket) {
+                   DRAKE_DEMAND(IsThread(websocket_thread_id_));
+                   if (socket) {
+                     listen_socket_ = socket;
+                   }
+                 });
+      if (listen_socket_ != nullptr) {
+        chosen_port = us_socket_local_port(
+            /* ssl = */ 0, reinterpret_cast<us_socket_t*>(listen_socket_));
+        DRAKE_THROW_UNLESS(chosen_port > 0);
+        break;
+      }
+    }
 
     bool connected = listen_socket_ != nullptr;
-    app_promise.set_value(std::make_tuple(port, connected));
-
+    app_promise.set_value(std::make_tuple(chosen_port, connected));
     if (!connected) {
       return;
     }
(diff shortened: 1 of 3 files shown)


jwnimmer-tri:
+@RussTedrake for platform review, please.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960)* status: 1 unresolved discussion, LGTM missing from assignees joemasterjohn,RussTedrake(platform)

___
*[`geometry/meshcat.cc` line 1762 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-NbLOg765mtxpBvb4iWj:-NbLOg77BgSXAZblfMHn:b-e0xp8m) ([raw file](https://github.com/RobotLocomotion/drake/blob/892adabc8e266db45d0df5f921415f7a3014a539/geometry/meshcat.cc#L1762)):*
> ```cpp
>     // Search for an open port.
>     int chosen_port{};
>     for (int port = desired_port.value_or(7000); port <= 7999; ++port) {
> ```

Working

Oops.  I think this will fail is the user specifies a port like `8000`.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960)* status: LGTM missing from assignees joemasterjohn,RussTedrake(platform)

___
*[`geometry/meshcat.cc` line 1762 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-NbLOg765mtxpBvb4iWj:-NbLVD8sEP-odezavzjJ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake/blob/892adabc8e266db45d0df5f921415f7a3014a539/geometry/meshcat.cc#L1762)):*

Working

Oops.  I think this will fail is the user specifies a port like `8000`.
</blockquote></details>

Done.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960)* status: 1 unresolved discussion, LGTM missing from assignees joemasterjohn,RussTedrake(platform)

___
*[`geometry/meshcat.h` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-NbLUGNu6lhXGEAL3OnE:-NbLeJFC0P5-VSAZUgAt:b-wwlh8t) ([raw file](https://github.com/RobotLocomotion/drake/blob/892adabc8e266db45d0df5f921415f7a3014a539/geometry/meshcat.h#L44)):*

It seems odd to me to have this linear probe option AND the ephemeral port option. Are we only keeping the linear probe starting at 7000 to preserve legacy behavior? My preference would be to have `port == std::nullopt` to mean ephemeral port otherwise use the port value passed in. It seems weird that the user could create >100 meshcat instances (up to the limits of their machine) if they pass `port=0` but are hard limited to 100 if they leave out the port option.
</blockquote></details>

Having it start from 7000 (and so, in most cases actually obtain 7000) is nice for users who always keep their browser open to 7000, so that the javascript "reconnecting..." yellow box works as intended.

Also, the plain numbers (7000, 7001, ...) are also easier for humans to grok, making it easier to associate commands in certain terminals with the right browser window.  The "random" looking number like 52786 are more difficult for users to visually match up.
___
*[`geometry/meshcat.cc` line 1765 at r2](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-NbLYeXM9Xu2eWiOnWuy:-NbLfbjE3WHWTk8w2iV7:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake/blob/96b841fc9c4f54d00173ee65baa1f458f77e4e6e/geometry/meshcat.cc#L1765)):*

nit: I can't find the expected behavior of listen in the uWS docs/examples/etc. I'm assuming that if this is passed `port=0` and the `LIBUS_LISTEN_EXCLUSIVE_PORT` option, it chooses an ephemeral port? If so can we document it here before the call.
</blockquote></details>

Done

joemasterjohn (commented):
Feature :lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-:-NbLhSw64ato6X2M8y0L:b-n0m1lx)* status: LGTM missing from assignee RussTedrake(platform)

___
*[`geometry/meshcat.h` line 44 at r1](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-NbLUGNu6lhXGEAL3OnE:-NbLgiQNDono0_U_-tvW:b-m9niyo) ([raw file](https://github.com/RobotLocomotion/drake/blob/892adabc8e266db45d0df5f921415f7a3014a539/geometry/meshcat.h#L44)):*

Having it start from 7000 (and so, in most cases actually obtain 7000) is nice for users who always keep their browser open to 7000, so that the javascript "reconnecting..." yellow box works as intended.

Also, the plain numbers (7000, 7001, ...) are also easier for humans to grok, making it easier to associate commands in certain terminals with the right browser window.  The "random" looking number like 52786 are more difficult for users to visually match up.
</blockquote></details>

That's true, it would put the burden on the user to click/copy the url in the terminal and lose the "reconnecting..." feature. OK.

joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960)* status: LGTM missing from assignee RussTedrake(platform)

jwnimmer-tri (commented):
+@EricCousineau-TRI for platform review per schedule tomorrow, please.  -@RussTedrake can still weigh in, but I won't block on it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-:-NbRGhfV4lSMnb2vGZJ2:bi3g55j)* status: LGTM missing from assignee EricCousineau-TRI(platform)

EricCousineau-TRI (commented):
:lgtm: platform!

> Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19960#-:-NbUyjo43FSqdHFaXHIi:b-vtj4w1)* status: :shipit: complete! all discussions resolved, LGTM from assignees EricCousineau-TRI(platform),joemasterjohn

PR #19961, opened 2023-08-08
RenderEngineVtk consumes and renders glTF

SeanCurtis-TRI:
A glTF file can contain an arbitrary number of meshes. Those meshes are lumped together in a single assembly and mapped to the `GeometryId`. As the pose of the geometry changes, it is the pose of that assembly that changes.

Two fundamental changes:

1. `GeometryId` no longer maps to `vtkActor` and now maps to `vtkProp3D`.
    - Sometimes the `vtkProp3D` pointer points to `vtkActor` and sometimes to `vtkAssembly`.
    - It can't *always* be `vtkAssembly` as a superset, because VTK does not export `vtkAssembly` to glTF correctly. [Bug report](https://discourse.vtk.org/t/bug-in-vtkgltfexporter/12052).
    - This requires some extra work in copying because of VTK's weird "copy" semantics. It doesn't have a `Clone()` method which creates a object from an existing; it has a `ShallowCopy()` method which writes one existing object into another -- you must already know the concrete type of the copy target. Blech.
    - The protected `actors()` method has become `props()`, reflecting the change.
2. We're using the `vtkGLTFImporter`.

This also simplified the `RenderEngine` cloning logic using `ShallowCopy()` from VTK.

Incidentally, this is a minimum first step in supporting .obj files with multiple objects/materials. When we parse objs, we can return multiple meshes with single materials and map them to parts in assembly like we have done for glTFs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961)


jwnimmer-tri:
> The protected actors() method has been deprecated.

The `RenderEngineVtk` class is in `namespace internal`.  No need to deprecate anything.  It's not subclassable anywhere beyond the `RenderEngineGltfClient`.  The header file is even completely encapsulated (`internal = True`) -- not installed, and cannot be included by anyone.  The only public API is its `...Params` struct and the factory function.

SeanCurtis-TRI (commented):
All excellent points. I'll go one further. I won't just remove the deprecation, I'll remove the entire function. No one is using it and it was originally snuck in so we could paint on meshes in a hacky way. That's no longer a pressing need, so, I'll go ahead and stomp on it. If we ever have a need to do something like that, we'll take a moment to do something more disciplined.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbJFX-3Cpj8_APZd3n3:b2e6doo)* status: needs platform reviewer assigned, needs at least two assigned reviewers, commits need curation (https://drake.mit.edu/reviewable.html#curated-commits), missing label for release notes (waiting on @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
-(status: do not review) 

+@zachfang for feature review, please.
+(release notes: feature) 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NblO-Cz7Eum5TWTPMBq:b79e6qp)* status: LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

SeanCurtis-TRI (commented):
+(status: do not review) 

VTK + mac is rearing up it's ugly head again. Just like with the light configuration, mac is failing in ridiculous and stupid ways. Ways that I cannot assess without a physical mac. So, I'm going to let you finish your review so I can at least clean up the defects you've found. But I think I'm simply going to have to close the two VTK PRs until I can find some way to figure out what the hell is going on in mac land.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbqYQuG3oW2VaRkCVJu:bo7tnmt)* status: 6 unresolved discussions, LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

jwnimmer-tri:
It would not be unheard-of to land this particular feature as gated on Ubuntu-only for now.  (We could throw when given a glTF mesh and we've built for macOS, with those glTF-specific unit test cases disabled.)

Of course that assumes the refactoring part is succeeding (only the glTF importing part is what's broken), and that you want to thread that needle versus waiting until you have working hardware to test on.

SeanCurtis-TRI (commented):
Of the test failures, I'm assuming the `model_visualizer_test` is a fluke (all networky).

For the `internal_render_engine_vtk_test`, it's only failing on the newly added test `GltfSupport`.  There are some points of confusion:

1) The test results report only "5 tests from 1 test suite ran". There are certainly far more tests in that test suite. Is this an artifact of caching? If so, I'd have expected that the other tests would have been accounted for as cached. If not, what happened to the other tests? Should I be worried about that at all?
2) I would've sworn that when both lighting and glTF were combined in a single PR, I wasn't getting mac failures. The lighting PR was producing bizarre clipping in the images. I get the sense from the failure messages that weird clipping is happening here as well. It makes me wonder if simply by causing the test to be evaluated in a different order might affect the outcome.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbqaL3k14WsdGUzMXoU:b23ycai)* status: LGTM missing from assignee zachfang, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`-- commits` line 23 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbouYHXCfy9TDcIPUzy):*

BTW. Maybe it's worth mentioning the previous importer we used.
</blockquote></details>

I don't think we are. We were using `vtkGLTFImporter` already in the `server_vtk_backend.cc`. Certainly, searching for `vtkGLTFReader` in master doesn't bring up any hits for me (but `vtkGLTFImporter` is successful).
___
*[`geometry/render_vtk/internal_render_engine_vtk.h` line 256 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbpB2EP5tH_4Zr4BCOr:-NbqZt1JCr_J4aujcSIk:bnczwkd) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.h#L256)):*

nit. I think this came from the changes of configuring lights that are excluded from this PR?
</blockquote></details>

D'oh! Taking two commits and reversing their order is trickier than I expected.
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 131 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-Nbp7q-L-G9WWLkgMJqt:-Nbq_2Y54FMbw0j_7Goz:b-44pv4n) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.cc#L131)):*

BTW. To confirm my understanding, with and without the change, `RegistrationData` is a const and is immutable.
</blockquote></details>

Correct. The only reason it kind of needed to be mutable is to mark the `accepted` flag. That is now done in a more concise way, so the vast majority of the functions that deal with `RegistrationData` don't need to write anything.

jwnimmer-tri:
Ah.  We already have a ton of VTK tests disabled on ARM mac:

https://github.com/RobotLocomotion/drake/blob/0fa91afe3f0d8f9697afbbd3bcf154dc4fded82c/geometry/render_vtk/BUILD.bazel#L107-L131

Are you macOS problems ARM-specific?  If macOS x86 is passing, I'd say just add more nerfs to the list of ARM tests.

> ... causing the test to be evaluated in a different order might affect the outcome.

Yes.  The disabled test cases really somehow depend on the order things are being run.  We haven't been able to figure out what's happening at all.

SeanCurtis-TRI:
@drake-jenkins-bot  mac-x86-monterey-clang-bazel-experimental-release please.

zachfang (commented):
:lgtm_strong: Mostly cosmetics comments and some questions when I read through.

> Reviewed 1 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-Nbqvoat3sE1sf8Falhz:b-88wll3)* status: 5 unresolved discussions, needs platform reviewer assigned, needs at least two assigned reviewers

___
*[`-- commits` line 23 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbouYHXCfy9TDcIPUzy):*

I don't think we are. We were using `vtkGLTFImporter` already in the `server_vtk_backend.cc`. Certainly, searching for `vtkGLTFReader` in master doesn't bring up any hits for me (but `vtkGLTFImporter` is successful).
</blockquote></details>

Right. I think I messed that up with vtkOBJReader.
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 202 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqPwY22JaM4HS2DGmH:-NbqPwY375SGKcuZKeZH:bhjep30) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.cc#L202)):*
> ```cpp
>                                          const RigidTransformd& X_WG) {
>   vtkSmartPointer<vtkTransform> vtk_X_WG = ConvertToVtkTransform(X_WG);
>   // TODO(SeanCurtis-TRI): Perhaps provide the ability to specify actors for
> ```

BTW. Maybe we should update the comment too.



_Suggestion:_
```cpp
specify props
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 320 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqS4Tz4L9DNf770RGw:-NbqS4Tz4L9DNf770RGx:b-3kz40z) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.cc#L320)):*
> ```cpp
>                 "Geometries should all be derived from vtkProp3D.");
>   std::array<vtkSmartPointer<vtkProp3D>, N> target_props;
>   for (std::size_t i = 0; i < N; ++i) {
> ```

BTW. From the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Integer_Types), `std::size_t` is generally not desired?



_Code quote:_
```cpp
std::size_t i
```
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 583 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqpIwr3ZvfzCgbWXWx:-NbqpIws0SkiBZCS_F14:b-8p63fd) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/internal_render_engine_vtk.cc#L583)):*
> ```cpp
>   std::array<vtkSmartPointer<vtkActor>, kNumPipelines> actors{
>       vtkSmartPointer<vtkActor>::New(), vtkSmartPointer<vtkActor>::New(),
>       vtkSmartPointer<vtkActor>::New()};
> ```

BTW/question. I was wondering that would it make sense to remove `actors` given `props` is introduced now. If we only have props, we will need to do a lot of SafeDowncast for the vtkActor operations which is tedious?



_Code quote:_
```cpp
  std::array<vtkSmartPointer<vtkActor>, kNumPipelines> actors{
      vtkSmartPointer<vtkActor>::New(), vtkSmartPointer<vtkActor>::New(),
      vtkSmartPointer<vtkActor>::New()};
```
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 995 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqkfR3BMTvLaYVmgK_:-NbqkfR3BMTvLaYVmgKa:bb7lc56) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L995)):*
> ```cpp
>     // the color in the glTF's materials or textures. The PBR shader behaves
>     // differently from the phong shader in the other tests. For now, we
>     // account for this by putting the observed color in the test. If we chnage
> ```

nit.



_Suggestion:_
```cpp
change
```
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1433 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-Nbqolrd5p_bdB6ISaz4:-Nbqolrd5p_bdB6ISaz5:b-dg7jw9) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1433)):*
> ```cpp
>   bool GeometryHasColorTexture(GeometryId id,
>                                const std::string& texture_name) const {
>     vtkActor* actor = GetColorActor(id);
> ```

BTW. Should we put the const back as a const pointer?



_Code quote:_
```cpp
vtkActor* actor
```

SeanCurtis-TRI (commented):
+@rpoyner-tri for platform review, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbrPeGa0-k7tbjTTiKe:b-syrr9x)* status: 2 unresolved discussions, LGTM missing from assignee rpoyner-tri(platform)

___
*[`-- commits` line 23 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbouYHXCfy9TDcIPUzy):*

Right. I think I messed that up with vtkOBJReader.
</blockquote></details>

I did the same.  :)
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 320 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqS4Tz4L9DNf770RGw:-NbrLwIuBMdqNU6fVIg8:b-3qufny) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.cc#L320)):*
While it's strongly discouraged, it's not absolutely forbidden. Given that the templated param N *must* be `size_t` to work properly with the array, the code felt much cleaner to leave the loop variable unsigned.

My implicit argument comes from the [styleguide's treatment of integers](https://drake.mit.edu/styleguide/cppguide.html#Integer_Types):

> Unsigned types are prohibited except in the specific cases of ... -   `size_t` is allowed when a (1) non-Drake API uses unsigned types, and (2) casting to a signed type during Drake’s interactions with the non-Drake API would obscure the readability of our code, and (3) subtraction underflow below zero is obviously not at risk.

I think (1) and (3) objectively apply and I would argue that while (2) isn't a slam dunk, the for loop is clearer as is.
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 583 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqpIwr3ZvfzCgbWXWx:-NbrJqM_CFLl3Hl_al-c:b-uc2377) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/internal_render_engine_vtk.cc#L583)):*

BTW/question. I was wondering that would it make sense to remove `actors` given `props` is introduced now. If we only have props, we will need to do a lot of SafeDowncast for the vtkActor operations which is tedious?
</blockquote></details>

I'm not 100% sure about what exactly you're driving at with the question. Here's my best effort.

We've got an array of `vtkActor` and another array of `vtkProps3D`. Your question seems to be about, given the the second array can hold what's in the first array, and we eventually move them all from the actor array to the prop array, why bother with the first array?

1. It's *nice* to be clear about what it is we're creating and introducing into VTK. In this case, they are indisputably instances of `vtkActor` (that is the most derived type).
2. We are specifically exercising actor-only APIs (e.g., `GetShaderProperty`, `SetMapper`, `GetProperty()`, etc.).
3. If they are typed as `vtkActor`, we can exercise those APIs without down casting. When we're done, we then put them in the array of `vtkProp3D` (because we're not doing anything else in the function that depends on knowing they're actors).
4. Finally, we can't assign an array of `vtkActor` to an array of `vtkProp3D`. The whole `vtkSmartPointer` interferes with any intelligence that `std::array` would have about recognizing that an array of `vtkActor*` is an array of `vtkProp3D*`.
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1433 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-Nbqolrd5p_bdB6ISaz4:-NbrNSMT4ZXH8xhPO5-V:b-39j6ml) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1433)):*

BTW. Should we put the const back as a const pointer?
</blockquote></details>

I *think* what you're suggesting is that this should be.

`vtkActor* const actor = GetColorActor(id)`. That would be analogous to what was there before. Is that correct?

We could do that, but, to be perfectly frank, I don't feel that most C++ coders are used to seeing const pointers to non-const objects and, therefore, while technically correct, doesn't help legibility.

And, because the value `actor` only exists for one line of code after its declaration, making sure that no one can actually change the thing it's pointing to has low value.

P.S. In case you were thinking `const vtkActor* actor`, that can't work. `vtkActor::GetProperty()` is a non-`const` member method.

zachfang (commented):
> Reviewed 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961)* status: LGTM missing from assignee rpoyner-tri(platform)

___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 320 at r3](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqS4Tz4L9DNf770RGw:-NbtEqrg2unxWeTdvwXd:budidgt) ([raw file](https://github.com/RobotLocomotion/drake/blob/db67733171bd3d4d091bf20e9e9b5e40cd24e7e3/geometry/render_vtk/internal_render_engine_vtk.cc#L320)):*

While it's strongly discouraged, it's not absolutely forbidden. Given that the templated param N *must* be `size_t` to work properly with the array, the code felt much cleaner to leave the loop variable unsigned.

My implicit argument comes from the [styleguide's treatment of integers](https://drake.mit.edu/styleguide/cppguide.html#Integer_Types):

> Unsigned types are prohibited except in the specific cases of ... -   `size_t` is allowed when a (1) non-Drake API uses unsigned types, and (2) casting to a signed type during Drake’s interactions with the non-Drake API would obscure the readability of our code, and (3) subtraction underflow below zero is obviously not at risk.

I think (1) and (3) objectively apply and I would argue that while (2) isn't a slam dunk, the for loop is clearer as is.
</blockquote></details>

Thanks for the clarification. Good to know std::array is actually taking a `size_t`.
___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 583 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbqpIwr3ZvfzCgbWXWx:-NbtKRj5E6tFaavdnbFw:b-98th7p) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/internal_render_engine_vtk.cc#L583)):*

I'm not 100% sure about what exactly you're driving at with the question. Here's my best effort.

We've got an array of `vtkActor` and another array of `vtkProps3D`. Your question seems to be about, given the the second array can hold what's in the first array, and we eventually move them all from the actor array to the prop array, why bother with the first array?

1. It's *nice* to be clear about what it is we're creating and introducing into VTK. In this case, they are indisputably instances of `vtkActor` (that is the most derived type).
2. We are specifically exercising actor-only APIs (e.g., `GetShaderProperty`, `SetMapper`, `GetProperty()`, etc.).
3. If they are typed as `vtkActor`, we can exercise those APIs without down casting. When we're done, we then put them in the array of `vtkProp3D` (because we're not doing anything else in the function that depends on knowing they're actors).
4. Finally, we can't assign an array of `vtkActor` to an array of `vtkProp3D`. The whole `vtkSmartPointer` interferes with any intelligence that `std::array` would have about recognizing that an array of `vtkActor*` is an array of `vtkProp3D*`.
</blockquote></details>

Yes, that's precisely my question. Thanks!
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1433 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-Nbqolrd5p_bdB6ISaz4:-NbtHRyP8vfF2lkJSHw8:b-ojkmav) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1433)):*

I *think* what you're suggesting is that this should be.

`vtkActor* const actor = GetColorActor(id)`. That would be analogous to what was there before. Is that correct?

We could do that, but, to be perfectly frank, I don't feel that most C++ coders are used to seeing const pointers to non-const objects and, therefore, while technically correct, doesn't help legibility.

And, because the value `actor` only exists for one line of code after its declaration, making sure that no one can actually change the thing it's pointing to has low value.

P.S. In case you were thinking `const vtkActor* actor`, that can't work. `vtkActor::GetProperty()` is a non-`const` member method.
</blockquote></details>

Yes, I was thinking about `vtkActor* const actor`. However, now I need to re-read how `auto` is in play with all the `const` and `ptr` stuff (from the old code).

rpoyner-tri (commented):
-(status: do not review)  :lgtm: 

> Reviewed 2 of 7 files at r2, 5 of 9 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbuIU_t0l2WnNsJlgCZ:bom1oy5)* status: 2 unresolved discussions

___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 346 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbuPi77-Yt6owLd5EwA:-NbuPi783ZQxnpjH4TAD:b5xl4l1) ([raw file](https://github.com/RobotLocomotion/drake/blob/57dbcb5a73b63a9f5192b1a826e3ad97dce4a9fc/geometry/render_vtk/internal_render_engine_vtk.cc#L346)):*
> ```cpp
>       target_props = CloneProps<vtkActor>(source_props);
>     } else if (vtkAssembly::SafeDownCast(source_props[0]) != nullptr) {
>       target_props = CloneProps<vtkAssembly>(source_props);
> ```

minor This branch is not covered by unit tests -- should I worry?
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1034 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbuHQx30daXKqJRz5WE:-NbuHQx44cyKxSpNbdMM:b8gq8an) ([raw file](https://github.com/RobotLocomotion/drake/blob/57dbcb5a73b63a9f5192b1a826e3ad97dce4a9fc/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1034)):*
> ```cpp
>       {.rendered_color = Rgba(0.30980392, 0.0745098, 0.52941176),
>        .rotation = RotationMatrixd::MakeYRotation(M_PI / 2),
>        .name = "purple"},
> ```

BTW 8 sig figs per color channel seems like a lot, but I don't know that it is exactly wrong or has bad consequences here.

SeanCurtis-TRI (commented):
Test coverage updated.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-:-NbweX2D5fH4vu61m-WN:bg8jcde)* status: :shipit: complete! all discussions resolved, LGTM from assignees rpoyner-tri(platform),zachfang

___
*[`geometry/render_vtk/internal_render_engine_vtk.cc` line 346 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbuPi77-Yt6owLd5EwA:-NbwcvL20vnsrWwDh_a-:b-iejyin) ([raw file](https://github.com/RobotLocomotion/drake/blob/57dbcb5a73b63a9f5192b1a826e3ad97dce4a9fc/geometry/render_vtk/internal_render_engine_vtk.cc#L346)):*

minor This branch is not covered by unit tests -- should I worry?
</blockquote></details>

Now ostensibly covered. I'll be submitting an issue on the topic we vaguely discussed this morning.
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1433 at r4](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-Nbqolrd5p_bdB6ISaz4:-NbwdSJs847iZdaFJTbh:b-nle622) ([raw file](https://github.com/RobotLocomotion/drake/blob/175229346846644d48873f9350775b6b47add327/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1433)):*

Yes, I was thinking about `vtkActor* const actor`. However, now I need to re-read how `auto` is in play with all the `const` and `ptr` stuff (from the old code).
</blockquote></details>

Yeah, `const auto` used to resolve to `const vtkSmartPointer<vtkActor>`. That's the equivalent of `vtkActor* const`. So, your instinct was correct. I feel the old `const auto actor` spelling was misleading. The alternate `vtkActor* const actor` would be correct and not misleading, but I think it introduces its own clarity issues.
___
*[`geometry/render_vtk/test/internal_render_engine_vtk_test.cc` line 1034 at r5](https://reviewable.io/reviews/RobotLocomotion/drake/19961#-NbuHQx30daXKqJRz5WE:-NbwdOTo1wqyyI9I0D2i:bdemlx4) ([raw file](https://github.com/RobotLocomotion/drake/blob/57dbcb5a73b63a9f5192b1a826e3ad97dce4a9fc/geometry/render_vtk/test/internal_render_engine_vtk_test.cc#L1034)):*

BTW 8 sig figs per color channel seems like a lot, but I don't know that it is exactly wrong or has bad consequences here.
</blockquote></details>

Done

(Merely copy-and-paste precision from taking int / 255.0.

